PR #373, opened 2017-02-09
Feature doxygen

economon:
This PR adds an updated doxyfile (thanks Gigi Martinelli!) and a logo directly to SU2/Common/doc. Doxygen is now set up to be run directly from that directory with

$ doxygen su2_doxyfile

and an html/ directory will be created with the output. The latest Doxygen docs for v5.0 have been generated and can be found here: http://su2.stanford.edu/doxygen/.

However, we have many warnings when running doxy, and we can do a lot more to improve the documentation. We should customize further the doxyfile, update the SU2/Common/doc/docmain.hpp, and clean / improve the doxy comments throughout the code. Any volunteers to help?


hlkline:
This looks great - I notice that there's a couple of things in the doxygen file in su2code/Documentation/Doxygen/Doxyfile that might be helpful here (setting project name, there may be other things). I wonder if these two files could be merged? Perhaps we should remove su2code/Documentation/Doxygen?

For potential volunteers: all the warnings I see right now seem to come from inputs/outputs not included in the comments description - correcting this should be relatively accessible to folks just getting started, and I'd also encourage anyone who has introduced a new function recently to check and see if the comment block above the function declaration in the .hpp file is filled in. 
Other doxygen info: I recall that the last time I worked with doxygen, some warnings went away once graphviz was installed on my local machine.

Sorry that I can't commit 100% to volunteering for this right now; but I might chip away at things once in a while.

economon:
Any other comments here? Otherwise, this is a simple addition that we can put in right away, and folks can hack away on Doxygen whenever their heart desires.

PR #374, opened 2017-02-09
Move all restart reading from solver constructors to dedicated LoadRestart() routines. 

economon:
This pull request moves all restart file loading from CSolver constructors to dedicated LoadRestart() routines within each CSolver class. This is purely a refactoring of the code, so all previous functionality should stay the same. 

There are two goals:

1. Consistency for reading of restarts and minimizing duplicated code. LoadRestart() routines already existed for some solvers, but not others. They were being used in some cases, but not all. In addition, by having the restarts read in the constructor, they were tied to the instantiation of the CVariable class, but this is unnecessary and caused some odd behavior in corner cases. Now, the solver constructors always instantiate the CVariable classes with free-stream / uniform initial conditions, and the LoadRestart() routine in each solver is the only place where restart files are loaded, which is first done in CDriver::Solver_Preprocessing().

2. Now that the reading of restarts is isolated in one routine of each CSolver class (out of the CSolver constructor and separated from instantiating CVariable), new versions of the LoadRestart() routines can be written alongside the ASCII originals for reading binary restart files (e.g., LoadRestart_Binary() ), which are coming next.

~~**Important** (@fpalacios, all): the metadata that was being written to the bottom of the ASCII restart files for updating parameters has been moved to its own file for now so that we can start writing the data portion of the restarts in binary. It will be written to/read from either “restart.meta” or “restart_adj.meta” if the new option UPDATE_RESTART_PARAMS= YES is set. I would like some feedback about this. I am open to doing something different with that information, e.g., putting it in the cfg file somehow since they are config options for the most part. Please let me know what you think.~~

Please check your own cases to make sure that restarts and SU2_SOL are working for you as before.

Next up: Parallel writing of binary restart files with new LoadRestart_Binary() routines for loading.


economon:
Please let me know if y'all have comments on this one.. it's a big change to the code and impacts most people. Another PR built on top of this is imminent with fully parallel output of binary restart files, but we'd need to merge this first!

talbring:
Thanks for making these changes! So far everything looks fine for me.

One remark though regarding the metadata: maybe it is possible to have a ASCII header in the binary files ? Having for example the iteration number always in the restart file is actually not a bad idea. Addtionally, it would be also possible to make a check whether the file is binary or ascii.

economon:
Thanks, Tim. Yes, it might be best to put the iteration number (and perhaps time step, for instance), back into the restart. It's also not that hard to put a number of additional parameters at the end in binary. It's just some additional bookkeeping to keep track of them, since you won't be able to search for 'EXT_ITER=' or other cfg options. Rather, you'll have to know the exact order of the doubles in the file. 

I think we can get everything back in after the next PR with the parallel binary restarts. Btw, I am leaving all of the ASCII restart code intact, and adding the binary stuff alongside. There is a new cfg option to control ASCII or binary for restarts.

economon:
Forget the bit about the metadata. We can leave it as is in this PR, but I'm putting it all back in now that the binary restarts are ready in another branch. That PR coming very soon...

SU2_CFD/src/output_structure.cpp
```diff
@@ -3859,21 +3861,35 @@ void COutput::SetRestart(CConfig *config, CGeometry *geometry, CSolver **solver,
...
-  
+
+  /*--- Close the data portion of the restart file. ---*/
+
   restart_file.close();
-  
+
+  /*--- Also create the filename for the restart metadata ---*/
+
+  if (wrt_metadata) {
```
demanosalvas: Does this mean that we will need a "solution_flow" file and a "metadata" file to restart the flow?

economon: The metafile is only necessary if you want to include the ExtIter, AoA, etc. 

However, I have decided to put this information back into the restart files with the next PR, so this will disappear. Everything with the restarts will work as always (file names, extensions, metadata, etc.) with the change to binary being transparent to the user.

I'll leave this as is for now, since the next PR changes it again.

economon:
Thanks for the comments. Time to merge this in to make way for the next PR concerning the restarts. Please report any issues right away.

PR #375, opened 2017-02-09
Changing naming in CIteration classes: (mean flow) -> (fluid).

economon:
This is a cosmetic but important change to the naming of the CIteration class for fluids. It is more technically correct to use CFluidIteration() instead of CMeanFlowIteration(), as we drive iterations of both the mean flow and turbulence solvers at the level of the CIteration() class. Also, this is more in line with the CFluidDriver() class that sits above it.

Please let me know what you think...


talbring:
The change definitely makes sense. In general i think the term "mean flow" is misleading. What about changing also the names of the solver files ?

economon:
I'm ok with it in the solver files, since there we are actually separating the mean flow from the turbulence. Having Euler and N-S solver files would be nice, but then we'd have to separate those classes, which might be confusing since they are so tied together through inheritance.

LaSerpe:
Hi guys,
I agree with you, we should change the name of that function including the word "fluid". 
In this way the purpose of the function is stated clearly against its counterparts (structural, poisson, heat, etc)

economon:
Thanks, y'all. Merging this in now too.

PR #379, opened 2017-02-26
Parallel Binary Restarts with MPI I/O

economon:
Completes the transition from ASCII -> binary restart files. Binary restarts are now the default, and in parallel, they are written/read using MPI derived datatypes and MPI I/O. All code behavior should remain the same as before, so please check your cases and report any issues immediately.

Some notes:
* Backwards compatible with ASCII restarts using the new options 'WRT_BINARY_RESTART= NO' and 'READ_BINARY_RESTART= NO' (both are YES now by default).
* All file naming conventions remain the same, and SU2_SOL operates as before for writing visualization files. All Python scripts should also operate normally.
* Metadata (EXT_ITER, AOA, etc.) has been included at the end of the binary restarts, just as in the ASCII files, so the temporary restart.meta file has been removed.


talbring:
Hey Tom, thanks for making these changes! This is an important improvement. 

However, I just tested it with an unsteady incompressible case. When I try to convert the solutions with SU2_SOL the size of the tecplot files increases in every time step and there are a lot of NaN in the files. I'll see if I can get more information ...

talbring:
My latest commit fixes the mentioned problem.

talbring:
also used this PR to fix a small bug I just noticed.

economon:
Thanks for the fix, Tim. One area that I had to modify a bit was the SetSensitivity() routine that you just modified.. does it all look ok to you in your discrete tests? Most of the ones in the repo were working fine for me.

talbring:
Yep, I had a look at it. Thats why I came across this bug with SA_NEG that I solved above. I almost wanted to blame you for that :)

talbring:
I'll do some more tests today, but so far it's looking very good !

sametcaka:
Hope this would make writing any variables to the solution file easy :)

economon:
Quick scaling tests show major improvements with the new output for reading and writing of restart files. See attached. 

Tim, have you been testing with Intel compilers? I ran into an issue with Intel compilers on one cluster with the restarts, and I want to see if it was an isolated incident.

If all looks good, we'll merge in soon. 

output_performance

juanjosealonso:
Tom,

This looks very good in terms of absolute times to read and write input/output.  Kudos.

The scaling is a bit less exciting, although the mesh seems too small for the large numbers of cores you were using, so one cannot draw conclusions from that (unless you were to run on 32 cores or 16 cores, etc).

One interesting thing would be to compute the effective read/write bandwidth…for example, for 72 cores, what was the total (binary) size of the file being read/written?  We can compare those bandwidths to the single-stream I/O bandwidth that can be obtained in a single node.

Excellent addition…this will be a game changer, particularly for unsteady flow solutions and for time-accurate adjoints.

Best,

Juan



On Mar 2, 2017, at 6:23 PM, Thomas D. Economon <notifications@github.com<mailto:notifications@github.com>> wrote:


Quick scaling tests show major improvements with the new output for reading and writing of restart files. See attached.

Tim, have you been testing with Intel compilers? I ran into an issue with Intel compilers on one cluster with the restarts, and I want to see if it was an isolated incident.

If all looks good, we'll merge in soon.

[output_performance]<https://cloud.githubusercontent.com/assets/4896083/23535771/5d47d730-ff75-11e6-9be0-8e9542093bc1.png>

economon:
Juan,

Don't know those numbers yet for the aggregate bandwidth with Lustre or how many I/O nodes are available for this cluster. Will have to do a detailed analysis eventually. For reference, this RANS case at 27 million grid points (not cells) results in a 3 GB restart file (binary). I'll take the two orders of magnitude improvement :).

Tom

juanjosealonso:
Tom,

Ok.  So 3 Gb and 4 sec to write on 72 cores is about 10.6Mb / sec / core.  Since not each core has access to a separate I/O node (assume 8 cores / processor), that would give upwards of 80 Mb / sec / node which is quite respectable.  The reading seems a bit slower, but it is indeed a major improvement.  Later in time we should do some tests for much larger files to make sure we get some level of scalability.

Best,

Juan

On Mar 3, 2017, at 10:40 AM, Thomas D. Economon <notifications@github.com<mailto:notifications@github.com>> wrote:


Juan,

Don't know those numbers yet for the aggregate bandwidth with Lustre or how many I/O nodes are available for this cluster. Will have to do a detailed analysis eventually. For reference, this RANS case at 27 million grid points (not cells) results in a 3 GB restart file (binary). I'll take the two orders of magnitude improvement :).

Tom

economon:
These are dual socket Broadwells at 36 cores/node. The 72 core result is for 2 nodes (mesh would have trouble fitting on one node), so on a per-node basis, we have about 375 MB / sec / node for write.

talbring:
Nice work, indeed! I just tested it on our cluster (BeeGFS filesystem) and I get similar numbers (writing down from 542s to 6s on 256 cores, unfortunately I couldn't do scaling tests so far because more than half of the cluster is down atm). Furthermore I tested it with the Intel compiler (v2017, OpenMPI) and it seems to work fine for me.

economon:
Time to merge this in. Please update your branches with develop when you can and report any issues.

PR #380, opened 2017-02-28
Feature body force

economon:
Adds a general body force source term implementation to the Finite Volume flow solver, which is particularly useful for pressure-driven / gravity / channel problems, for instance. 

Also includes a simple 2D Poiseuille flow case with an analytic solution for testing the body force and periodic implementations.

### DIFF
--- a/Common/include/config_structure.hpp
+++ b/Common/include/config_structure.hpp
@@ -825,7 +825,7 @@ private:
   *default_ffd_axis,          /*!< \brief Default FFD axis for the COption class. */
   *default_inc_crit,          /*!< \brief Default incremental criteria array for the COption class. */
   *default_body_force;        /*!< \brief Default body force vector for the COption class. */
-  bool Body_Force;  			       /*!< \brief Flag to know if a body force is incuded in the formulation. */
+  bool Body_Force;            /*!< \brief Flag to know if a body force is included in the formulation. */
   su2double *Body_Force_Vector;  /*!< \brief Values of the prescribed body force vector. */
 
   /*--- all_options is a map containing all of the options. This is used during config file parsing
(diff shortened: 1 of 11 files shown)


Common/include/config_structure.hpp
```diff
@@ -823,8 +823,11 @@ class CConfig {
   *default_grid_fix,          /*!< \brief Default fixed grid (non-deforming region) array for the COption class. */
   *default_htp_axis,          /*!< \brief Default HTP axis for the COption class. */
   *default_ffd_axis,          /*!< \brief Default FFD axis for the COption class. */
-  *default_inc_crit;          /*!< \brief Default incremental criteria array for the COption class. */
-  
+  *default_inc_crit,          /*!< \brief Default incremental criteria array for the COption class. */
+  *default_body_force;        /*!< \brief Default body force vector for the COption class. */
+  bool Body_Force;  			       /*!< \brief Flag to know if a body force is incuded in the formulation. */
```
arubino: just a small typo here: ~~incuded~~ -> included

economon:
Thanks, Antonio, will fix. I'll wait for a second reviewer on this one.

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -3993,6 +3993,51 @@ void CSourceGravity::ComputeResidual(su2double *val_residual, CConfig *config) {
...
+
+    /*--- Momentum contribution ---*/
+
+    for (iDim = 0; iDim < nDim; iDim++)
+      val_residual[iDim+1] = -Volume * U_i[0] * Body_Force_Vector[iDim] / Force_Ref;
+
+    /*--- Energy contribution ---*/
+
+    for (iDim = 0; iDim < nDim; iDim++)
+      val_residual[nDim+1] = -Volume * U_i[iDim+1] * Body_Force_Vector[iDim] / Force_Ref;
```
talbring: Something is probably not correct here. val_residal[nDim+1] is overwritten in each loop cycle.

economon: Code reviews prove their worth again... yep, bug there. Should be accumulating. Fixing now, let's see how it affects the test.

PR #385, opened 2017-03-17
Feature py multiobj

hlkline:
This pull request includes:
- updates and refinement of multiple objective gradient evaluation and optimization
- per-surface evaluation of forces and moments can now be included in an optimization or gradient evaluation: this is only used when more than one objective is specified. 
- the multiobjective test case has been added back into the regressions
- FIN_DIFF_STEP can now be specified in the config file (the option was already added in the develop branch; a small modification was needed to the python code in order to use it). If this option is not specified the previous behavior is used. 
- a few minor corrections are also included
- penalty functions can now be used as a portion of the objective. The penalty function is defined as the square of the difference between the function and its constraint - this is hard-coded, but can be modified by simply changing obj_p and obj_dp in the python scripts. 

A tutorial around multiobjective optimization may be added later based on the inv_wedge_ROE_multiobj.cfg case.

### DIFF
--- a/SU2_PY/SU2/eval/design.py
+++ b/SU2_PY/SU2/eval/design.py
@@ -272,13 +272,14 @@ def obj_dp(config,state,this_obj,def_objs):
     value = su2func(this_obj,config,state)
     valuec = float(def_objs[this_obj]['CVAL'])
     dpenalty = 0.0
+
     # Inequalities will be 0 or a positive value
     if ((def_objs[this_obj]['CTYPE']=='>' and value < valuec)  or\
          (def_objs[this_obj]['CTYPE']=='<' and value > valuec )):
         dpenalty=2.0*abs(valuec - value)
-    # Equalities will be positive if value>constraint, negative if value<constraint
     elif (def_objs[this_obj]['CTYPE']=='='):
-        depenalty=2.0*(value -valuec)
+        # Equalities will be positive if value>constraint, negative if value<constraint
+        dpenalty=2.0*(value -valuec)
 
     return dpenalty
 
@@ -326,8 +327,6 @@ def obj_df(dvs,config,state=None):
                 # For a penalty function, the term is scaled by the partial derivative
                 # d p(j) / dx = (dj / dx) * ( dp / dj)  
                 scale[i_obj]*=obj_dp(config, state, this_obj, def_objs)
-
- 
             
         config['OBJECTIVE_WEIGHT']=','.join(map(str,scale))
         
--- a/SU2_PY/SU2/eval/design.py
+++ b/SU2_PY/SU2/eval/design.py
@@ -225,8 +225,9 @@ def obj_f(dvs,config,state=None):
     state = su2io.State(state)
     
     def_objs = config['OPT_OBJECTIVE']
-    objectives = def_objs.keys()
     
+    objectives = def_objs.keys()
+
 #    if objectives: print('Evaluate Objectives')
     # evaluate each objective
     vals_out = []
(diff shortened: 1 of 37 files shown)


hlkline:
After a correction to the regression tests, this is now ready for review.

TestCases/cont_adj_euler/wedge/inv_wedge_ROE_multiobj.cfg
```diff
@@ -362,7 +377,9 @@ VISUALIZE_DEFORMATION= YES
 %
 % Optimization objective function with scaling factor
 % ex= Objective * Scale
-OPT_OBJECTIVE=LIFT*-1.0E4;OUTFLOW_GENERALIZED*0.01 
+OPT_OBJECTIVE=AVG_TOTAL_PRESSURE*1E-8;(DRAG = 0.045)*1E4
```
fpalacios: The format is not clear for me. What is the reason for the parenthesis? and the = 0.045?

hlkline: This is an example of setting the optimization objective function to include a penalty function. 
In general, I was trying to set it up so that the user can specify the constraint in the same way that they would if the constraint was passed explicitly to the optimizer. 

Using (Drag = constraint)*scale means that rather than optimizing to minimize drag, the optimization function will include scale*(CD - constraint)^2. Using an inequality would do something similar, except that when the inequality constraint is satisfied the penalty function is 0.

See obj_p and obj_dp in eval/design.py for where the penalty function is implemented. The chain rule is used to obtain the gradients needed by the optimizer.

hlkline: In other words, the goal is to be able to cut-and-paste directly from the OPT_CONSTRAINT line in order to switch between using penalty functions inside the objective vs using the optimization algorithm.

SU2_PY/SU2/eval/design.py
```diff
@@ -226,25 +226,64 @@ def obj_f(dvs,config,state=None):
...
-    
+    # If evaluating the combined function is desired, update it here
+    if state.FUNCTIONS.has_key('COMBO'):
+        state['FUNCTIONS']['COMBO'] = func
+        
     return vals_out
 
 #: def obj_f()
+        
+def obj_p(config,state,this_obj,def_objs):
```
fpalacios: I like the possibility of adding penalty functions... sometimes is quite difficult to make the constraint optimizer works.

SU2_PY/SU2/eval/design.py
```diff
@@ -226,25 +226,64 @@ def obj_f(dvs,config,state=None):
...
     func = 0.0
     for i_obj,this_obj in enumerate(objectives):
         scale = def_objs[this_obj]['SCALE']
         sign  = su2io.get_objectiveSign(this_obj)
-        
         # Evaluate Objective Function
         # scaling and sign
-        func += su2func(this_obj,config,state) * sign * scale
-        
+        if def_objs[this_obj]['CTYPE']=='NONE':
```
hlkline: 'CTYPE' is short for 'constraint type' - >, <, =, or 'NONE', where if none it is just the normal objective evaluation - but looking back at this that is not abundantly clear, so I will try to think up a better name for that option.

SU2_PY/SU2/eval/gradients.py
```diff
@@ -467,7 +467,10 @@ def findiff( config, state=None ):
         log_findiff = None
 
     # evaluate step
-    step = 0.001 * float(config.REF_LENGTH_MOMENT)
+    if config.has_key('FIN_DIFF_STEP'):
+        step = float(config.FIN_DIFF_STEP)
+    else:
+        step = 0.001 * float(config.REF_LENGTH_MOMENT)
```
hlkline: Ok, will do (both on removing the REF_LENGTH_MOMENT option and making sure FIN_DIFF_STEP is in the config file)

hlkline:
@fpalacios - For the continuous adjoint it is NOT required to run multiple adjoints, you can actually combine them. Whether or not to do so is controlled by the OPT_COMBINE_OBJECTIVE option - I will try to add in some more comments as you suggest to make this clearer, along with other updates next week.

fpalacios:
Hi Heather, I think we are getting closer. Something that is not clear for me is why we can only combine objective functions with the continuous adjoint. In principle, it is "easier" to combine objective functions with he discrete adjoint. Isn't it? Is there a particular reason for not combining the objective functions with he discrete adjoint? Thanks! Francisco

hlkline:
Thanks
The discrete adjoint can also be used with combined objectives, and uses the same config specifications. It may not be clear from this pull request since this functionality was added in a previously - when multiple objectives (limited to objectives defined inside the CFD solver) are specified, they are added together (with specified weights) in the CFD solver into the 'COMBO' output to history. This variable is registered for the discrete adjoint solver. 
An example of this can be seen in the serial_regression_AD.py, test name 'discadj_multi_py'.

fpalacios:
Thanks for the clarification!

Best,
Francisco

hlkline:
@fpalacios, @economon - There's a comment in the config files that says that 'OBJECTIVE_FUNCTION' sets the objective function for the optimizer - however it seems that this option is only used with controlling the adjoint solver and output, and whatever the user specifies there will be overwritten by the optimizer. I put an alternative comment in the config_templates here; I can either revert that back or change the comment in all the config files (I have a script).

fpalacios:
You are completely right... the description for OBJECTIVE_FUNCTION

% Objective function in optimization problem (DRAG, LIFT, SIDEFORCE, MOMENT_X,
%                                             MOMENT_Y, MOMENT_Z, EFFICIENCY,
%                                             EQUIVALENT_AREA, NEARFIELD_PRESSURE,
%                                             FORCE_X, FORCE_Y, FORCE_Z, THRUST,
%                                             TORQUE, TOTAL_HEATFLUX,
%                                             MAXIMUM_HEATFLUX, INVERSE_DESIGN_PRESSURE,
%                                             INVERSE_DESIGN_HEATFLUX, AVG_TOTAL_PRESSURE, 
%                                             MASS_FLOW_RATE, ELLIPTIC_SPANLOAD, MAX_SECTIONAL_CL)
OBJECTIVE_FUNCTION= DRAG

Is not accurate and introduces some level of confusion. 

From the “average" user point of view my experience is that they only use OPT_OBJECTIVE and thy don’t know that OBJECTIVE_FUNCTION exists. Could you please change "Objective function in optimization problem” by something more accurate? In fact, we should maybe think on changing the keyword OBJECTIVE_FUNCTION itself.

Best,
Francisco

Issue #347, opened 2016-12-19
Problem with regression?


jschuele:
Fetched developed branch and ran serial_regression.py. Got case FAILED with GCC for ls89_sa, centrifugal_blade, centrifugal_stage.
With INTEL compiler on same platform get passed for ls89_sa, centrifugal_blade, centrifugal_stage but failed for statbeam3d. Strange are the differences for naca0012_def and most of the brick-tests with intel. They converge in significant fewer steps and are marked FAILED.
Is it worthwhile to investigate the differences between GCC and Intel for the brick-cases?
I'am asking, because I'm gone look to fasten calculations up with optimizations in the existing code for single core and perhaps MPI. Far goal, but really far, may be using GPUs as well. Currently I have a 15% improvement single core for RANS from the Comp&Fluids_129 paper. But of course the calculations should not only be faster but correct as well.
Regards, Josef

economon:
Hi Josef: glad to see you getting involved!

I am very interested in this. We have seen differing behavior sometimes between the two compiler sets, in particular when high levels of optimization are being used. In fact, you might check if the same issues appear without compiler optimization. Often, it is an initialization problem or some optimization that affects correctness.

If you have the time to explore different settings and perhaps dig into the code a bit, I would very much like to isolate and remove these compiler-dependent problems with your help.

jschuele:
Fine, I will do so and check what the compilers are doing in these cases. 
In between a comment perhaps to the memory leak issue:

In
CUpwCUSP_Flow destructor
memory leak. *Jacobian is not deleted. Just the elements.

fpalacios:
Hi Josef,

This is an important issue that we need to solve, I going to double check
the problem and I'll keep you posted.

Thanks for your interest in SU2,
Best,
Francisco

economon:
Ah, nice catch. Yes, in that particular case we are using periodic BCs, and in serial, the only way nPoint does not equal nPointDomain is when we have periodic grid points.

jschuele:
Ok, the remaining FAILED test I had after the change mentioned above was fea_fsi/StatBeam_3d. I broke down the issue to my changes to Gauss_Elimination in matrix_structure. The major change in terms of a difference is exchanging 
weight = block[..]/block[..]
against
reciproque=1.0f/block[..] 
weight = block[..] * reciproque
which is of course faster as it removes some costly divisions.
[...]

PR #386, opened 2017-03-23
Perfomance improvements

jschuele:
Gauss-Elimination with reciprocal values, some optimization in limiter calc. ...
A suspected bug in Line 230 solver_structure.cpp fixed


LaSerpe:
I can see that only the Gauss_Elimination_core.hpp was changed, could you please give more information about the bug in solver_structure.cpp you are referring to?

jschuele:
Sorry,

files were missing.

As I mentioned in my issue, there is a mismatch in some cases between 
MPI and non parallel version in the output. Thomas Econom commented on 
it and referred to elements missed in that case.

Regards,

Josef

fpalacios:
Hi @jschuele thanks a lot for your contribution to SU2. Would you be so kind to update this branch and give us an estimate of the performance improvement that the changes are introducing. 

@economon, @talbring  we need to carefully look at them, they seems reasonable but... maybe to specific. Is it worth it the lost of generality? some numbers would help a lot!

In my opinion the line 230 is not a bug... we need to divide by the total number of points, not the number of points in a particular core of the MPI decomposition. But I'll double check that. What do you think?

Thanks a lot,
Francisco

jschuele:
So far I run the turb_oneram case with different number of points. There the chances made up to around 15% reduction of CPU time used. Currently I'm changing the MatrixVectorProduct for that case and find another 15% reduction. I will update this branch if there is a chance that that work is not wasted.

Line 230: In the nicf/LS89 testcase the MPI and sequential code gives different results. The only reason is that line. I discussed that with Tom in #347

fpalacios:
Hi @jschuele, definitively a 10% or 15% improvement in performance is totally worth it. Be sure that we we'll approve the changes and the work is not going to be wasted. 

Thanks again for your contribution! I'm going to add some comments to the source code to be sure that we understand why the performance improvement.

Best,
Francisco

Common/include/Gauss_Elimination_core.hpp
```diff
@@ -0,0 +1,141 @@
+    su2double d0,d1,d2,d3,d4;
```
fpalacios: Does this subroutine have the traditional SU2 headers?

Common/include/Gauss_Elimination_core.hpp
```diff
@@ -0,0 +1,141 @@
...
+    if (nVar<6) {
+#define BLOCK(i,j) block[(i)*nVar+(j)]
+    d0 = 1.0000000000000000 / BLOCK(0,0);
```
fpalacios: Why 1 is defined ass 1.00000000 . I´m sure there is a good reason for that. Thanks!

Common/src/matrix_structure.cpp
```diff
@@ -732,11 +734,52 @@ void CSysMatrix::ProdBlockVector(unsigned long block_i, unsigned long block_j, c
   
   su2double *block = GetBlock(block_i, block_j);
   
+  switch (nVar)
```
fpalacios: Do you know why this change improve performance? Thanks!

SU2_CFD/src/numerics_structure.cpp
```diff
@@ -332,6 +333,35 @@ void CNumerics::GetInviscidProjJac(su2double *val_velocity, su2double *val_energ
   for (iDim = 0; iDim < nDim; iDim++)
     val_Proj_Jac_Tensor[nDim+1][iDim+1] = val_scale*(val_normal[iDim]*a1-a2*val_velocity[iDim]*proj_vel);
   val_Proj_Jac_Tensor[nDim+1][nDim+1] = val_scale*Gamma*proj_vel;
+
```
fpalacios: This change affects the generality of the source code. Do you know how much is worth it in computational time? Thanks!

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -6591,35 +6657,60 @@ void CEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config) {
 
         /*--- Calculate the interface left gradient, delta- (dm) ---*/
         
-        dm = 0.0;
-        for (iDim = 0; iDim < nDim; iDim++)
-          dm += 0.5*(Coord_j[iDim]-Coord_i[iDim])*Gradient_i[iVar][iDim];
+        switch(nDim)
```
fpalacios: Again... just curiosity... why this change improve the performance? Thanks!

SU2_CFD/src/transport_model.cpp
```diff
@@ -87,15 +87,19 @@ CSutherland::~CSutherland(void) { }
...
+  Mu = Mu_ref*(aux1*aux1*aux1)*((T_ref + S)/(T + S));
 
 }
 
 void CSutherland::SetDerViscosity(su2double T, su2double rho) {
 
   dmudrho_T = 0.0;
-  dmudT_rho = Mu_ref*( (3.0/2.0)*pow( (T/T_ref),(1.0/2.0) )*( (T_ref + S)/(T + S) )
-          -pow( (T/T_ref),(3.0/2.0) )*(T_ref + S)/(T + S)/(T + S) );
+  su2double aux1=sqrt(T/T_ref);
```
fpalacios: Thanks!

SU2_CFD/src/solver_structure.cpp
```diff
@@ -226,7 +226,8 @@ void CSolver::SetResidual_RMS(CGeometry *geometry, CConfig *config) {
       
     }
     
-    SetRes_RMS(iVar, max(EPS*EPS, sqrt(rbuf_residual[iVar]/Global_nPointDomain)));
+    SetRes_RMS(iVar, max(EPS*EPS, sqrt(rbuf_residual[iVar]/ geometry->GetnPoint() )));
```
fpalacios: I´m not sure about this change. @economon What do you think?

Common/include/Gauss_Elimination_core.hpp
jschuele: I wrote it as include file to be include below the header. This provides the advantage to use the same modifications for the LU-Decomp. and the ILU while not blowing up the original source code by that many lines.

Common/include/Gauss_Elimination_core.hpp
jschuele: The rounding error plays a role for Gauss-Elimination. So I just wanted to assure, that 1. indeed has fullest possible precision (double) and not just the (most common) float-precision (8 digits) with some random numbers behind those 8 digits.

Common/src/matrix_structure.cpp
jschuele: Loops introduce some overhead which can not be compensated for if the loops are short. The compiler has no chance to optimize register usage, introduces a condition each iteration to test for the iteration count. All not important if the loop is long.

SU2_CFD/src/numerics_structure.cpp
jschuele: I optimized this function even more and for the moment I would like to postpone a comment to this.

SU2_CFD/src/solver_direct_mean.cpp
jschuele: Two reasons - first like above removing a very frequent called short loop. Next the precalculation of Coordiv (independent of the nVar-loop) and with a change in sign to be used for i and j limiter as well.

SU2_CFD/src/solver_structure.cpp
economon: I think this will only work when the code is built in serial. This is a special case when nPoint = Global_nPointDomain. Otherwise, we should use MPI to gather the total number of grid nodes in the mesh (Global_nPointDomain above), excluding ghost nodes, so that we are always dividing by the same number of grid nodes regardless of number of MPI ranks.

jschuele: You know better than I which value to use there. But in serial the result is wrong - that's my only point. It should be correct and the same in serial and in parallel.

jschuele:
I am not sure whether all is on in this request like it should.
- I did an update of my request with respect to current develop.
- I selected some of my changes and collected once again Timings for turb_oneram with 8k elements, 20 iterations in serial runs.
1. Use NUMACTL if more than 1 CPU - may save a lot (without 654sec)
2. Original develop branch: 557 sec
3. Gauss-elimination down to 546, prodblockvector 533, matrixvectorproduct 525, diag_upper in preconditioner 513, transportmodel 509, Changes to CUpwRoe_Flow::ComputeResidual and called funtions 485. Thus in total 72sec, appr. 15% savings.
4. My - not commited, still in test - version runs in 390 sec.

SU2_CFD/src/solver_structure.cpp
fpalacios: Could you be a little bit more specific about why the result is wrong in serial? I have run a quick test with the QuickStart test and I think I always have the same number of points 5233 (for serial of parallel simulations). Thanks!

fpalacios:
Thanks for your work, We are really looking forward to accept these changes, could you please take a look at the regression test? Thanks again, 
Best,
Francisco

oleburghardt:
Dear @jschuele
As this PR is open now for more than a year and possibly outdated, I will close it for now. We emphasize that you are very welcome to open a new one anytime.

PR #389, opened 2017-05-05
Feature sliding interface

LaSerpe:
Hi guys,
this is to merge in the sliding mesh approach as proposed in the PhD thesis
"Models and simulation of non-ideal fluid flows in unconventional turbomachinery: Toward highly efficient next-generation green power systems"

This request mostly includes additions to the interpolator and to the transfer classes.
The config option string to identify zone interface markers was changed from MARKER_FSI_INTERFACE to MARKER_ZONE_INTERFACE.
When the domain involve more than one interface,  boundaries sharing the same interface should be listed as pairs in the config string, the order of the pairs doesn't matter.
Currently the code works in parallel and it is able to treat a random number of fluid zones.
Periodic support is still missing though I am planning to work on it in a future push.


Common/include/dual_grid_structure.hpp
```diff
@@ -54,81 +54,81 @@ using namespace std;
  */
 class CDualGrid{
 protected:
-	static unsigned short nDim; /*!< \brief Number of dimensions of the problem. */
-	
+    static unsigned short nDim; /*!< \brief Number of dimensions of the problem. */
```
talbring: This file for example.

Common/src/config_structure.cpp
```diff
@@ -176,7 +176,7 @@ unsigned short CConfig::GetnZone(string val_mesh_filename, unsigned short val_fo
   /*--- For harmonic balance integration, nZones = nTimeInstances. ---*/
 
   if (config->GetUnsteady_Simulation() == HARMONIC_BALANCE && (config->GetKind_SU2() != SU2_DEF)   ) {
-  	nZone = config->GetnTimeInstances();
+    nZone = config->GetnTimeInstances();
```
talbring: Also look at the white spaces in this file.

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -650,6 +650,7 @@ void CUpwCUSP_Flow::ComputeResidual(su2double *val_residual, su2double **val_Jac
   /*-- Face area ---*/
   
   Area = 0.0;
+
```
talbring: Same again here.

LaSerpe:
Hi Tim,
yep, you're right, sorry about that.
I reverted the whitespace changes and now the diff output should be fine.
Let me know if you think it could be further improved.

economon:
Giulio: this is very exciting! Thanks for the hard work (the commits date back to almost a year ago). I will submit a review soon...

LaSerpe:
Thanks Tom, actually most of the work was already done in September, lately I've been busy with other stuff and that's why it took so long before this pull request.

Regarding the upcoming release of the sliding meshes, before the AIAA, we can surely think about including this one too. 
It still has some limitations, for instance it doesn't work with periodic boundaries, though Antonio and I are working on it in separate branches, but I've tested it for a bunch of different flow configurations and it seems to be quite robust.
Basically, after this pull request the code will be able to carry out computations using an arbitrary number of zones, both Euler/RANS, serial/parallel, using the nearest neighbour approach and the one from Rinaldi.

By the way, it doesn't really matter to me if this feature is included in the upcoming release or in the one later, I just would like to have this on the develop so that I don't have to merge and fix conflicts any time the branch gets updated.
So some of the modification I made will be stable and I can move to the next step (the adaptive sliding meshes).

Common/src/interpolation_structure.cpp
```diff
@@ -1352,7 +1361,1770 @@ void CMirror::Set_TransferCoeff(CConfig **config) {
   }
 
   #ifdef HAVE_MPI
-  if (rank == MASTER_NODE) 
+  if (rank == MASTER_NODE && Buffer_Recv_mark != NULL) 
     delete [] Buffer_Recv_mark;
   #endif
 }
+
+CSlidingmesh::CSlidingmesh(CGeometry ***geometry_container, CConfig **config, unsigned int iZone, unsigned int jZone)  :  CInterpolator(geometry_container, config, iZone, jZone){
```
economon: This constructor is getting quite long (it's hard to follow). It's ok for now, but can you think about how to decompose into smaller methods that can be called for the different portions? More comments describing the various sections would be useful too.

Common/src/interpolation_structure.cpp
```diff
@@ -1352,7 +1361,1770 @@ void CMirror::Set_TransferCoeff(CConfig **config) {
...
+      return fabs( dotA2 - dotB2 );
+  }
+
+  if( ( dotB1 <= 0 && dotA2 <= dotB2 ) || ( dotB2 <= 0 && dotA2 <= dotB1 ) )
+    return fabs( dotA2 );
+
+  return 0.0;
+}
+
+su2double CSlidingmesh::Compute_Triangle_Intersection(su2double* A1, su2double* A2, su2double* A3, su2double* B1, su2double* B2, su2double* B3, su2double* Direction){
```
economon: All of these geometry routines look nice. Do you think we could use them for other purposes? I am thinking for things like probes, cutting slices of data or geometry, etc. We have some routines for these things in other locations for SU2_GEO. Maybe we need to consolidate these routines in a general location somewhere...

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -15234,6 +15269,30 @@ CNSSolver::~CNSSolver(void) {
     delete [] Inlet_FlowDir;
   }
 
+  if ( SlidingState != NULL ) {
```
economon: Great, thanks for deleting the memory!

TestCases/sliding_interface/uniform_flow/uniform_SM.cfg
```diff
@@ -0,0 +1,267 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
economon: Thanks for adding so many test cases. Do you have the critical ones also added to the regression tests in Travis CI? It will help protect against things breaking in the future.

Also, are there any general options/comments that you can add to the config_template.cfg file for the sliding mesh capability? We try to keep that up to date the best we can.

Common/src/interpolation_structure.cpp
```diff
@@ -276,6 +276,19 @@ int CInterpolator::Find_InterfaceMarker(CConfig *config, unsigned short val_mark
   return -1;
 }
 
+su2double CInterpolator::PointsDistance(su2double *point_i, su2double *point_j){
```
LaSerpe: The plan here was to change the brute force algorithm with the ADT structure that Edwin included some while ago. That is why this is not an inline function yet.
I'd rather leave this change for the next push, but we can surely consider to exploit ADTs from now on already.
It will just take some more time to implement such mod within the code (I'm busy for a coming conference in early June) but it won't be big deal, I guess.

Common/src/config_structure.cpp
```diff
@@ -3845,11 +3837,11 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
       cout<< endl;
     }
     
-    if (nMarker_FSIinterface != 0) {
+    if (nMarker_ZoneInterface != 0) {
```
LaSerpe: Done.

Common/include/interpolation_structure.hpp
```diff
@@ -253,5 +260,98 @@ class CMirror : public CInterpolator {
    * \param[in] config - Definition of the particular problem.
    */
   void Set_TransferCoeff(CConfig **config);
+  
+};
+
+/*!
+ * \brief Sliding mesh approach
+  */
+class CSlidingmesh : public CInterpolator {
```
LaSerpe: Yep, sure.

Common/src/interpolation_structure.cpp
LaSerpe: Well, routines that make such operations are always useful and one could definitely use them for other purposes.
I definitely agree with you then, we should have a library or a general location where to collect all these "geometrical" routines. I can move the ones I put in the interpolation class into the CGeometry class for now on.
For instance, I noticed that also vectorial operation are re-written in different places around the code (cross/scalar product, vector magnitude, point distance...). 
We can surely talk about it and create a general location where to collect all these routines. 

Moreover, here in Milano we have a guy who is working on a particle tracking solver (which acts now as a post-processing tool) and therefore we have a lot of routines that work on the geometry, like walking search algorithm, line-plane intersection and so on...we could think to migrate those algorithms into SU2 to attach what is missing.
I am also thinking about my next step, which is an adaptive sliding mesh approach which will require a lot of function to carry out geometrical operations like splitting elements and faces.
I went through the geometry_adaptation_structure.cpp and I noticed that we could probably do something similar with these mesh operation too. 

Perhaps we can have a chat and speak about it.

TestCases/sliding_interface/uniform_flow/uniform_SM.cfg
LaSerpe: Yep.
By the way I have other reg test that I run locally on my desktop, I didn't put all of those to avoid making checks too long (I remember we had some issues with timeout limits in the past).
The crucial ones are there already.
If timeout is no longer a problem,  I can include the 5 zone pipe in place of the single stage (2zone only) so to have a more demanding test on multi-zone fluid problems.

economon: We have up to 75 minutes now for the regression tests, and we're usually around 50 minutes, so I think adding a few more test cases would be good. We'll ask the Travis CI folks for more help if we need it, and I'd rather we have good regression test coverage.

Common/src/interpolation_structure.cpp
economon: Very interesting. Yes, a discussion soon would probably be good. There are several folks now working on some of these very same things for interpolating solutions between two arbitrary volume grids, and we are also beginning to think about generic probes in the solution mesh, extracting planes/surfaces/lines of solution data for post-processing (or for computing things like noise), etc.

We should plan to consolidate all of these similar tools then.

TestCases/sliding_interface/uniform_flow/uniform_SM.cfg
LaSerpe: Agree, I added a few more, including the one with 5 zones which I believe is critical.
I was thinking that, right now, parallel regression works using only 2 processes for each test.
We should have some of them testing the code using also a different number of threads, as some bugs may not arise with just 2.

Common/src/interpolation_structure.cpp
LaSerpe: Yep, you're right.
I created a couple of new methods, now the code should much more readable, let me know what you think.

economon:
Thanks for all of the updates, Giulio. Let us know when you think it's ready, and we can take a final pass.

LaSerpe:
Hi guys, I made some last updates. 
It is now ready for a final pass.

Common/include/option_structure.hpp
```diff
@@ -297,16 +297,17 @@ static const map<string, ENUM_MAT_COMPRESS> MatComp_Map = CCreateMap<string, ENU
  * \brief types of interpolators
  */
 enum ENUM_INTERPOLATOR {
-  NEAREST_NEIGHBOR = 0,   /*!< \brief Nearest Neigbhor interpolation */
-  ISOPARAMETRIC = 1, /*!< \brief Isoparametric interpolation */
-  CONSISTCONSERVE = 2,/*!< \brief Consistent & Conservative interpolation (S.A. Brown 1997). Utilizes Isoparametric interpolation. */
+  NEAREST_NEIGHBOR 	= 0,   	/*!< \brief Nearest Neigbhor interpolation */
+  ISOPARAMETRIC 	= 1,	/*!< \brief Isoparametric interpolation */
+  CONSISTCONSERVE 	= 2,	/*!< \brief Consistent & Conservative interpolation (S.A. Brown 1997). Utilizes Isoparametric interpolation. */
+  SLIDING_MESH		= 3, 	/*!< \brief Sliding Mesh Approach E. Rinaldi 2015 */
```
arubino: Instead of "SLIDING_MESH", would a name more specific of the interpolation type be more appropriate for this, here? The sliding mesh would work also with the other interpolation methods.

Common/include/interpolation_structure.hpp
```diff
@@ -253,5 +285,98 @@ class CMirror : public CInterpolator {
    * \param[in] config - Definition of the particular problem.
    */
   void Set_TransferCoeff(CConfig **config);
+  
+};
+
+/*!
+ * \brief Sliding mesh approach
```
arubino: Again here, for the name CSlidingMesh would a name focused on the interpolation type be more appropriate than "SlidingMesh"? This method seems to be more a specific type of interpolation.

Common/src/config_structure.cpp
```diff
@@ -757,8 +757,8 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   addStringListOption("MARKER_FLUID_INTERFACE", nMarker_Fluid_InterfaceBound, Marker_Fluid_InterfaceBound);
   /*!\brief MARKER_INTERFACE\n DESCRIPTION: Zone interface boundary marker(s) \ingroup Config*/
   addStringListOption("MARKER_INTERFACE", nMarker_InterfaceBound, Marker_InterfaceBound);
-  /*!\brief MARKER_FSI_INTERFACE \n DESCRIPTION: FSI interface boundary marker(s) \ingroup Config*/
-  addStringListOption("MARKER_FSI_INTERFACE", nMarker_FSIinterface, Marker_FSIinterface);
+  /*!\brief MARKER_FSI_INTERFACE \n DESCRIPTION: ZONE interface boundary marker(s) \ingroup Config*/
+  addStringListOption("MARKER_ZONE_INTERFACE", nMarker_ZoneInterface, Marker_ZoneInterface);
   /*!\brief MARKER_DIRICHLET  \n DESCRIPTION: Dirichlet boundary marker(s) \ingroup Config*/
```
arubino: Both MARKER_ZONE_INTERFACE and MARKER_FLUID_INTERFACE seem to be needed, is that going to change later? We could maybe think to just use one marker. What do you think?

Common/src/interpolation_structure.cpp
```diff
@@ -277,50 +283,352 @@ int CInterpolator::Find_InterfaceMarker(CConfig *config, unsigned short val_mark
...
+#endif
+  
+  if( Buffer_Send_Coord              != NULL) {delete [] Buffer_Send_Coord;            Buffer_Send_Coord            = NULL;} 
+  if( Buffer_Send_GlobalPoint        != NULL) {delete [] Buffer_Send_GlobalPoint;      Buffer_Send_GlobalPoint      = NULL;}
+  if( Buffer_Send_LinkedNodes        != NULL) {delete [] Buffer_Send_LinkedNodes;      Buffer_Send_LinkedNodes      = NULL;}
+  if( Buffer_Send_nLinkedNodes       != NULL) {delete [] Buffer_Send_nLinkedNodes;     Buffer_Send_nLinkedNodes     = NULL;}
+  if( Buffer_Send_StartLinkedNodes   != NULL) {delete [] Buffer_Send_StartLinkedNodes; Buffer_Send_StartLinkedNodes = NULL;}
+}
+
+bool CInterpolator::CheckInterfaceBoundary(int markDonor, int markTarget){
```
arubino: Very useful and clean! Thanks!

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -12901,81 +12917,106 @@ void CEulerSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_cont
...
+              PrimVar_j[nDim+6] = GetSlidingState(iMarker, iVertex, nDim+6, jVertex); 
 
-            PrimVar_j[nDim+5] = GetSlidingState(iMarker, iVertex, nDim+5);
-            PrimVar_j[nDim+6] = GetSlidingState(iMarker, iVertex, nDim+6);
-
-            /*--- Set the normal vector and the coordinates ---*/
-
-            visc_numerics->SetNormal(Normal);
-            visc_numerics->SetCoord(geometry->node[iPoint]->GetCoord(), geometry->node[Point_Normal]->GetCoord());
+              coeff = GetSlidingState(iMarker, iVertex, nPrimVar, jVertex);
```
arubino: Maybe, can you add a comment about "coeff", adding why for more then one vertex is needed? What about a different name?

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -3271,6 +3305,8 @@ void CTurbSSTSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
 void CTurbSSTSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics,
     CNumerics *visc_numerics, CConfig *config){
 
+///////////// This has to be changed for Rinaldi approach
+
```
arubino: Here it just works for one vertex, if I am correct. Maybe we can add an error for the moment when used with the approach of Enrico? What would be its behavior, with this current implementation?

Common/include/option_structure.hpp
LaSerpe: I see your point, WEIGHTED_AVERAGE perhaps? But I think it would be confusing, do you have any better idea?

Common/src/config_structure.cpp
LaSerpe: Yep, indeed they're both needed as they have a different meaning.
MARKER_ZONE_INTERFACE is needed to tell SU2 how the interfaces are made up, i.e which boundaries share a particular interface, between two different zones.
MARKER_FLUID_INTERFACE is instead needed to specify the type of boundary condition to be applied over the markers (it could have been also one heatflux and one normal load in place of the fluid interface, as for the fsi cases).

SU2_CFD/src/solver_direct_mean.cpp
LaSerpe: perhaps "weight"?

SU2_CFD/src/solver_direct_turbulent.cpp
LaSerpe: Yes, I missed this one, now it is fixed.

Common/src/config_structure.cpp
rsanfer: Hi,
I can see the generality point in renaming MARKER_FSI_INTERFACE into MARKER_ZONE_INTERFACE (although this will force me to change all my configs...). 
However, would it be possible, then, to rename MARKER_FLUID_INTERFACE in a more general way as well? Something like MARKER_BC_INTERFACE? Because otherwise it will be limited to fluid BCs and will likely need to be changed in the future.
Cheers,
Ruben

LaSerpe: Hi Ruben,
the MARKER_FLUID_INTERFACE was indeed designed to be used for fluids only.
That is why it is currently owned by the solver_direct_mean.
How do you think it might need to change in the future?

rsanfer: Yep, I understand, however also FSI_INTERFACE was originally designed for FSI interfaces and has been proven to be useful in other contexts...
As an example, and for the same purpose as you have define above, you could want to define boundary conditions on different zones which indeed have different physics. An example, you may want to transfer heat, instead of fluid tractions, into a structural domain. I was thinking of it as a way to generalize the way the transfer (child) classes are initialized.
Another question, I don't understand what do you mean with "owned by solver_direct_mean". When you define a config option, it is owned by the config class, that means, it should be accessible anywhere that config is available. Am I right?

LaSerpe: I see your point which is fair, but I don't think that moving to a general BC is going to be straight forward, perhaps it would be better to define a new BC to treat such hybrid cases.
We should carefully think about it to get to manage all the possibilities, and maybe we can have a chat in the future (cause I'm not really into structural stuff so I may not see the bigger picture), but right now I would keep things separated.

Sorry, I was unclear, I meant that the method that corresponds to the fluid_interface marker is included in the solver_direct_mean.cpp.

Common/include/option_structure.hpp
arubino: perhaps FLUX_CONSERVATIVE or CONSERVATIVE_AVERAGE or  something like this? In that case I would slightly change CONSISTCONSERVE into ISOPARAMETRIC_CONSCONSERVATIVE.

SU2_CFD/src/solver_direct_mean.cpp
arubino: Good idea, just maybe also add a comment to explain what you are doing for the calculation of the residual in the loop.

Common/include/option_structure.hpp
LaSerpe: Truth be told the method is flux conservative for a set special cases only (planar surfaces basically, and actually not even for them in general).
I would avoid the term CONSERVATIVE then, probably WEIGHTED_FLUX is the most appropriate. What do you think?

arubino: I didn't know that.  Yes, WEIGHTED_FLUX sounds very clear to me.

Common/include/option_structure.hpp
```diff
@@ -300,7 +300,7 @@ enum ENUM_INTERPOLATOR {
   NEAREST_NEIGHBOR 	= 0,   	/*!< \brief Nearest Neigbhor interpolation */
   ISOPARAMETRIC 	= 1,	/*!< \brief Isoparametric interpolation */
   CONSISTCONSERVE 	= 2,	/*!< \brief Consistent & Conservative interpolation (S.A. Brown 1997). Utilizes Isoparametric interpolation. */
-  SLIDING_MESH		= 3, 	/*!< \brief Sliding Mesh Approach E. Rinaldi 2015 */
+  WEIGHTED_AVERAGE  = 3, 	/*!< \brief Sliding Mesh Approach E. Rinaldi 2015 */
 };
```
arubino: I see you opted for WEIGHTED_AVERAGE eventually. I would then also change the corresponding class name (and maybe the comment for the description?).

Common/include/interpolation_structure.hpp
```diff
@@ -253,5 +285,98 @@ class CMirror : public CInterpolator {
...
+  */
+class CSlidingMesh : public CInterpolator {
+public:
```
arubino: This is the class name I was referring in the other comment.

LaSerpe: I would leave that name since I'll probably need to re-use the very same class as a starting point for the implementation of the adaptive approach. 
So I'd keep it as general as possible for the time being, at least until the adaptive approach is ready.
We can still change the class name afterwards though.

arubino: OK, I do see an inconsistency in the name that can be misleading as this class is a specific interpolation method that can be used also with sliding meshes, but not the only one. I agree, if you prefer we can change this later on.

talbring:
All my requests were also addressed.Thanks for this feature!

LaSerpe:
Thanks guys, I'll merge this in right after Tom gives his approval.

economon:
Giulio, all: thanks for taking the time to add more comments and reviews. This is exactly the type of collaboration that we need to keep the quality high! LGTM!

LaSerpe:
Thanks to all of you guys! Merging this in now!

PR #390, opened 2017-05-09
Deprecate outflow generalized

hlkline:
This pull request deprecates the generalized outflow adjoint. This functionality is still available in an alternate branch, however I won't be able to maintain it for future updates. 
This means that some of the regression tests had to be updated, since they previously included testing this functionality. 

This pull request uses a matching branch in the TestCases repository to ease that update - the .travis.yml file will can be switched back to the TestCases/develop branch when needed.


SU2_CFD/src/solver_adjoint_mean_inc.cpp
```diff
@@ -3490,51 +3489,11 @@ void CAdjIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_contain
       case MASS_FLOW_RATE:
         Psi_outlet[0]+=1;
         break;
-      case OUTFLOW_GENERALIZED:
```
economon: I noticed this is in the incompressible adjoint solver. did these outlet options work within that solver too? Perhaps we need to trim down more?

TestCases/cont_adj_euler/wedge/inv_wedge_ROE.cfg
```diff
@@ -196,16 +196,14 @@ MG_ADJFLOW= NO
 %                                             INVERSE_DESIGN_HEATFLUX, AVG_TOTAL_PRESSURE, 
 %                                             MASS_FLOW_RATE)
 % For a weighted sum of objectives: separate by commas, add OBJECTIVE_WEIGHT and MARKER_MONITORING in matching order.
-OBJECTIVE_FUNCTION= OUTFLOW_GENERALIZED
+OBJECTIVE_FUNCTION= AVG_TOTAL_PRESSURE
```
economon: Great - thanks for updating the test to check the total pressure / 1D stuff.

TestCases/serial_regression_AD.py
```diff
@@ -135,7 +135,7 @@ def main():
     # test continuous_adjoint.py, with multiple objectives
     discadj_multi_py            = TestCase('discadj_multi_py')
     discadj_multi_py.cfg_dir    = "cont_adj_euler/wedge"
-    discadj_multi_py.cfg_file   = "inv_wedge_ROE_disc_multiobj.cfg"
+    discadj_multi_py.cfg_file   = "inv_wedge_ROE_multiobj.cfg"
```
economon: Just to be clear: the same config file now works for both continuous and discrete, correct?

SU2_CFD/src/solver_adjoint_mean_inc.cpp
hlkline: I will remove this shortly. I think these were copied over when someone that *_inc file was created - these boundary conditions would NOT be expected to work with the incompressible solver; there is literature on incompressible adjoints for a few objectives, and the equations will be different.

TestCases/serial_regression_AD.py
hlkline: Yes.

Issue #377, opened 2017-02-19
Unsteady optimization in V5.0


mghaffari:
Unsteady optimization in V5.0 has some issues. The problem is that after direct iterations SU2 cannot calculate adjoint for each time step properly.
It just keep calculating adjoint equation for one time step. For example a part of the log_adjoint file for NACA64A010 optimization is as follow:

Loading flow solution from direct iteration 250.
Begin direct solver to store flow data (single iteration).
Compute residuals to check the convergence of the direct problem.
log10[Maximum residual]: -2.77153, located at point 1018.
End direct solver, begin adjoint problem.

Min DT: 0.000323558.Max DT: 1.20067.Dual Time step: 0.677451.
log10[Maximum residual]: -3.34769.
Maximum residual point 134, located at (0.63649, 0.0371117).

 IntIter  ExtIter   Res[Psi_Rho]     Res[Psi_E]      Sens_Geo      Sens_AoA
       0       0      -4.444881      -4.159977    4.4996e+00   -6.5737e-04
      10       0      -4.488648      -4.764869    4.4053e+00   -6.5737e-04
      20       0      -4.433964      -4.864120    2.7194e+00   -6.5737e-04
      30       0      -4.449170      -4.851557    1.7148e+00   -6.5737e-04
      40       0      -4.669063      -4.954710    1.1593e+00   -6.5737e-04
      50       0      -4.689637      -5.116023    9.2103e-01   -6.5737e-04
      60       0      -4.797857      -5.289007    8.5846e-01   -6.5737e-04
      70       0      -4.922113      -5.428807    8.9570e-01   -6.5737e-04
      80       0      -5.021316      -5.534515    9.5224e-01   -6.5737e-04
      90       0      -5.102257      -5.604468    9.9700e-01   -6.5737e-04
     100       0      -5.169358      -5.654203    1.0424e+00   -6.5737e-04

Min DT: 0.0003236.Max DT: 1.201.Dual Time step: 0.6775.
log10[Maximum residual]: -4.172.
Maximum residual point 693, located at (0.03989, -0.0254).

 IntIter  ExtIter   Res[Psi_Rho]     Res[Psi_E]      Sens_Geo      Sens_AoA
       0       0      -5.221491      -5.704096    1.0838e+00   -6.5737e-04
      10       0      -5.262729      -5.764520    1.1144e+00   -6.5737e-04
      20       0      -5.306548      -5.835208    1.1305e+00   -6.5737e-04
[...]

PR #407, opened 2017-07-10
Adjoint Improvements

talbring:
This pull request introduces major structural changes to the discrete adjoint solver. I moved some of the routines to the driver level so that multi-zone problems work with the discrete adjoint. The basic implementation works for Fluid-Fluid zones. Other drivers follow soon (e.g. turbomachinery).

Along with that there also some other minor changes:

- The low memory tecplot output (which was not working properly anyway) is now removed.
- The issue with the unsteady cont. adjoint referenced in #377 is solved.
- Restart for the discrete adjoint works properly (as long as the same number of procs is used)
- I added options for frozen eddy viscosity and frozen limiter. This can be enabled with the options FROZEN_VISC_DISC/FROZEN_LIMITER_DISC. The default is NO for both.


talbring:
I also thought about unifying the options. However, for example for the frozen viscosity option we have different default values (NO for disc. adj. and YES for cont. adj.). That's why I did it that way for now. But I'll think of some solutions for that problem for the future.

Best,
Tim

fpalacios:
Thanks Tim,

From the practical view of point, it is important to decide which numerical spatial integration do you want to use for the AD stuff… even if it is not consistent with the direct solver.

Excellent job!

Thanks,
Francisco

PR #408, opened 2017-07-10
Collection of improvements

JSmith36:
Hi, I have been working on SU2 for a while and I have some improvements (some of them are just cosmetic improvements):

- SU2 is now able to identify several engines with multiple intakes and exhaust (this is an important change).
- Renaming of the keywords REF_LENGTH_MOMENT (the word MOMENT was really confusing) or RefAreaCoeff.
- This is a interesting one. There were two wing slicing subroutines... I have I have picked the best one for the entire code and improve a little bit its usability.
- The grid deformation was not perfect. It was computing only the distance from the moving surfaces. I have added the possibility to compute distance from any solid surface (that is the default that works better).
- Removing some options that look like incomplete and adding some extra info like CoP.

Thanks!


talbring:
Hi Jsmith36,

I just quickly went through some of the changes and they do look useful. But I would suggest that you split these things into several commits, otherwise the review gets quite messy (and for other git-related reasons).

Tim

fpalacios:
I concur, this is a really very useful set of implementations but... too much together. If you can split this in maybe in two pull request (e.g. engine stuff, and miscellanea) would be great. We want to minimize your work but we should keep the code organized.

Really very sorry for the extra work. And thanks again for contributing.

Best,
Francisco

PR #412, opened 2017-07-12
Flow solver improvements

JSmith36:
Hi again, I have cleaned my previous pull request. In this one you can find the following:

Renaming of the keywords REF_LENGTH_MOMENT (the word MOMENT was really confusing) or RefAreaCoeff.
In the original SU2 there were two geometry slicing subroutines... I have picked the best one for the entire code and improved its usability.
The original grid deformation subroutine only computes the distance from the moving surfaces. I have added the possibility to compute distance from any solid surface (now by default).
Removing some options that look like incomplete.

Thanks!


talbring:
Hi JSmith86,

I really appreciate your effort in cleaning up the changes. But it looks like as if there are still a lot of changes in other parts that are not related to the things you describe. Furthermore I really request you to split this up in multiple commits so that it is immediately clear what you did in each single one (this can be done quite simple with a proper diff tool like [meld](http://meldmerge.org/)).

Let me emphasize that this is not to bother you in any way but rather to ease understanding and maintainability. I know from my own experience that this requires some additional work, but in the end it certainly pays off. 

Thanks!
Tim

talbring:
PS: if you need any help with doing that, let me know.

JSmith36:
Thanks Tim,

Unfortunately, currently, I really don't have the time to divide all the contributions. Just creating this pull request from the previous one took me too much time. And, if we really want to make it right (in terms of tagging the individual contributions) we are speaking about a time that I frankly don't have right now.

I appreciate your help but, for the time being, I'll close the pull request.

Thanks,

fpalacios:
Hi @talbring, I cannot agree more with you, 

@JSmith36 has introduced too many changes in one pull request which has, as you pointed out, some important problems.

On the other hand, @JSmith36 situation is not so strange. It is quite common to find contributors who are working on their own branch for a long time and at a certain point decide to contribute (which is very generous of them).

My suggestion is the following... let's keep the pull request open (we don't want to lose this contribution) and meanwhile, @JSmith36 why don't you create pull request with the most-easy to- divide parts of your contribution. e.g. Let's start with 
"Renaming of the keywords REF_LENGTH_MOMENT (the word MOMENT was really confusing) or RefAreaCoeff."

That is easy to reimplement and it will simplify the main pull request.

For the future, please, let's all of us to keep the contributions the most simple and modular as possible. Thank you very much to each of you who contribute to this great community and great code!

Best,
Francisco

talbring:
@fpalacios I also think it is a good idea to keep the request open. Like I said, I can also help to simplify it. In fact, splitting it up in multiple commits is probably easier then you might think @JSmith36 :)

Common/src/geometry_structure.cpp
```diff
@@ -579,301 +579,616 @@ bool CGeometry::SegmentIntersectsTriangle(su2double point0[3], su2double point1[
...
-      delete [] Coord_Variation[iPoint];
-    delete [] Coord_Variation;
-    
-  }
-  
-  
+
+	if (original_surface == false) {
+
+		Coord_Variation = new su2double *[nPoint];
```
talbring: Please make sure to set the tab settings in your editor to spaces.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -1308,15 +1308,8 @@ void CDriver::Numerics_Preprocessing(CNumerics ****numerics_container,
             default : cout << "Centered scheme not implemented." << endl; exit(EXIT_FAILURE); break;
           }
           
-          if (!config->GetLowFidelitySim()) {
```
talbring: Does this mean the low fidelity sim. is removed or is this just an improvment ? I found it was a useful feature (although it was not really working properly in the past).

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -4115,60 +4115,74 @@ CSourceAxisymmetric_Flow::~CSourceAxisymmetric_Flow(void) { }
...
-    val_residual[0] = yinv*Volume*U_i[2];
-    val_residual[1] = yinv*Volume*U_i[1]*U_i[2]/U_i[0];
-    val_residual[2] = yinv*Volume*(U_i[2]*U_i[2]/U_i[0]);
-    val_residual[3] = yinv*Volume*Enthalpy_i*U_i[2];
-  }
-  if (incompressible) {
-    val_residual[0] = yinv*Volume*U_i[2]*BetaInc2_i;
-    val_residual[1] = yinv*Volume*U_i[1]*U_i[2]/DensityInc_i;
-    val_residual[2] = yinv*Volume*U_i[2]*U_i[2]/DensityInc_i;
+  if (Coord_i[1] > EPS) {
```
talbring: Here you introduced also tabs. Sorry about that, but we have to be strict on that one to avoid future conflicts.

SU2_CFD/src/output_tecplot.cpp
```diff
@@ -925,10 +925,200 @@ void COutput::SetTecplotASCII_Mesh(CConfig *config, CGeometry *geometry, bool su
...
-  if (surf_sol) delete [] LocalIndex;
+  if (surf_sol) {
+  	delete [] LocalIndex;
+  	delete [] SurfacePoint;
+  }
   
 }
 
+void COutput::SetSTL_MeshASCII(CConfig *config, CGeometry *geometry) {
+
```
talbring: Here are again tabs.

talbring:
@JSmith36 , can you enable the editing for maintainers ? Then I would push a version which is merged with the current develop.

JSmith36:
Sorry @talbring I have just read your message and I did the merging by myself. Thanks for the help.

talbring:
@JSmith36 , I am wondering why the direct differentiation test case fails. Do you have any idea what the reason could be ? My guess would be something in the grid movement has changed that causes that.

JSmith36:
Good catch! yep! I implemented an interesting/important change: Before, WALL_DISTANCE was computed using only the surfaces that you have identified as moving surfaces... as you can imagine that only works when you have a very simple problem (maybe an airfoil) but... if you have a problem with more Navier-Stokes markers together and you are moving only one of them the method doesn't work (e.g. wing-fuselage). For that reason I reimplemented WALL_DISTANCE which now is computed from all the solid surfaces and I also created DEF_WALL_DISTANCE that computes the distance from the surfaces that we are moving (as before). Frankly, I haven't found a situation in which DEF_WALL_DISTANCE outperforms the new solid WALL_DISTANCE... And, my suggestion is to eliminate DEF_WALL_DISTANCE in the future. For the time being I have added DEFORM_STIFFNESS_TYPE= DEF_WALL_DISTANCE to the config files to see if that solves the problem. Thanks!

Common/src/config_structure.cpp
```diff
@@ -142,9 +147,9 @@ unsigned short CConfig::GetnZone(string val_mesh_filename, unsigned short val_fo
       strcpy (cstr, val_mesh_filename.c_str());
       mesh_file.open(cstr, ios::in);
       if (mesh_file.fail()) {
-        cout << "cstr=" << cstr << endl;
-        cout << "There is no geometry file (GetnZone))!" << endl;
-
+        if (rank == MASTER_NODE) {
+          cout << "There is no geometry file called " << cstr << "." << endl;
```
JSmith36: No problem :-) ... As a side note, I have noticed that when you run the software in parallel the message output doesn't always  work. e.g. when you have an error in the config file. In other words, you need to run SU2_CFD (no parallel) to catch the errors...

Common/src/config_structure.cpp
```diff
@@ -4633,11 +4687,14 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
       case INVERSE_VOLUME:
         cout << "Cell stiffness scaled by inverse of the cell volume." << endl;
         break;
-      case WALL_DISTANCE:
-        cout << "Cell stiffness scaled by distance from the deforming surface." << endl;
+      case DEF_WALL_DISTANCE:
+        cout << "Cell stiffness scaled by distance to nearest deforming surface." << endl;
+        break;
+      case SOLID_WALL_DISTANCE:
```
JSmith36: My suggestion is to eliminate in the future the DEF_WALL_DISTANCE... there is not a good reason to keep that option. I kept that option because the regression test... but, for the next release my recommendation is to update all the regression tests and use only the distance to the SOLID_WALL. And yes... there is a WALL_distance in the limiter stuff, in the future it could be important to rename all the limiters adding the "_LIM" keyword to the name.

Common/src/grid_movement_structure.cpp
```diff
@@ -608,6 +609,136 @@ void CVolumetricMovement::ComputeDeforming_Wall_Distance(CGeometry *geometry, CC
...
+
+  unsigned long nLocalVertex_SolidWall = 0, nGlobalVertex_SolidWall = 0, MaxLocalVertex_SolidWall = 0;
+  unsigned long *Buffer_Send_nVertex    = new unsigned long [1];
+  unsigned long *Buffer_Receive_nVertex = new unsigned long [nProcessor];
+
+  /*--- Count the total number of nodes on deforming boundaries within the
+   local partition. ---*/
+
+  for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++)
+    if (config->GetMarker_All_KindBC(iMarker) != SEND_RECEIVE &&
```
JSmith36: I have changed SOLID_WALL_DISTANCE to BOUNDARY_DISTANCE in the C++ implementation but not in the config file option to keep the compatibility with previous versions.

Common/src/primal_grid_structure.cpp
```diff
@@ -612,4 +608,12 @@ CPyramid::~CPyramid() {
   
 }
 
-void CPyramid::Change_Orientation(void) { cout << "Not defined orientation change" << endl; }
+void CPyramid::Change_Orientation(void) {
```
JSmith36: No problem

SU2_CFD/src/output_structure.cpp
```diff
@@ -7779,132 +7948,281 @@ void COutput::SetForceSections(CSolver *solver_container, CGeometry *geometry, C
...
         
-        /*--- Basic conectivity ---*/
-        
-        for (iVertex = 1; iVertex < Xcoord_Airfoil.size(); iVertex++) {
-          Cp_File << iVertex << "\t" << iVertex+1 << "\n";
-        }
-        
         Cp_File.close();
         
+        //				/*--- Write Cp at each section (n3d format) ---*/
```
JSmith36: Gone!

SU2_CFD/src/solver_structure.cpp
```diff
@@ -2250,53 +2254,53 @@ void CSolver::Read_SU2_Restart_Metadata(CGeometry *geometry, CConfig *config, st
 
     /*--- Compute (negative) displacements and grab the metadata. ---*/
 
-    fseek(fhw,-(sizeof(int) + 5*sizeof(passivedouble)), SEEK_END);
+		fseek(fhw,-(sizeof(int) + 5*sizeof(passivedouble)), SEEK_END);
```
JSmith36: Got catch! In the future, it should be more easy to add extra metadata... maybe the 8 should be a variable that you can change only in one place

TestCases/cont_adj_euler/oneram6/inv_ONERAM6.cfg
```diff
@@ -369,14 +369,12 @@ FFD_CONTINUITY= 2ND_DERIVATIVE
...
-%    1/4_THICKNESS_SEC1, 1/4_THICKNESS_SEC2, 1/4_THICKNESS_SEC3, 1/4_THICKNESS_SEC4, 1/4_THICKNESS_SEC5, 
-%    1/2_THICKNESS_SEC1, 1/2_THICKNESS_SEC2, 1/2_THICKNESS_SEC3, 1/2_THICKNESS_SEC4, 1/2_THICKNESS_SEC5, 
-%    3/4_THICKNESS_SEC1, 3/4_THICKNESS_SEC2, 3/4_THICKNESS_SEC3, 3/4_THICKNESS_SEC4, 3/4_THICKNESS_SEC5, 
-%    AREA_SEC1, AREA_SEC2, AREA_SEC3, AREA_SEC4, AREA_SEC5, 
-%    AOA_SEC1, AOA_SEC2, AOA_SEC3, AOA_SEC4, AOA_SEC5, 
-%    CHORD_SEC1, CHORD_SEC2, CHORD_SEC3, CHORD_SEC4, CHORD_SEC5
+%    AIRFOIL_AREA, AIRFOIL_THICKNESS, AIRFOIL_CHORD, AIRFOIL_TOC, AIRFOIL_AOA,
+%    WING_VOLUME, WING_MIN_THICKNESS, WING_MAX_THICKNESS, WING_MAX_CHORD, WING_MIN_TOC, WING_MAX_TWIST, WING_MAX_CURVATURE, WING_MAX_DIHEDRAL
+%    FUSELAGE_VOLUME, FUSELAGE_WETTED_AREA, FUSELAGE_MIN_WIDTH, FUSELAGE_MAX_WIDTH, FUSELAGE_MIN_HEIGHT, FUSELAGE_MIN_HEIGHT, FUSELAGE_MAX_CURVATURE
+%    STATION#_WIDTH, STATION#_WATERLINE_WIDTH, STATION#_HEIGHT,
```
JSmith36: # is the index of the station defined in GEO_LOCATION_STATIONS I have updated the config file with that extra info. Thanks!

Common/src/geometry_structure.cpp
JSmith36: Sorry about that. Now it should be OK

SU2_CFD/src/driver_structure.cpp
JSmith36: It is gone... I agree, it is the typical feature that any user would like to try but... if it is not fully operative turns out to be a big source of frustration.

SU2_CFD/src/numerics_direct_mean.cpp
JSmith36: Agree, now it should be OK

SU2_CFD/src/output_tecplot.cpp
JSmith36: Wow, sorry about that

JSmith36:
Thanks a lot for the reviews, I have added all your comments to the implementation. 
I had some issues with the metadata in the old restart files but, now, everything is working great.

SU2_DEF/src/SU2_DEF.cpp
```diff
@@ -1,319 +1,329 @@
-/*!
```
talbring: Can you have a look what happened here ? Apparently git thinks the whole file has changed.

JSmith36:
Thanks for letting me know. It should be OK now... I have change the files from DOS (CR + LF) to UNIX (LF) format

talbring:
Thanks! For me everything looks fine now.

economon:
Thanks for the contribution and for working to wrap up the review changes. LGMT! Merging now.

PR #413, opened 2017-07-13
first release of the turbomachinery capabilities

salvovitale:
This pull request enlarges the capability of SU2 to simulate internal flows in turbomachinery. The flow solver is now capable of performing 3D steady multi-zone simulations via a 3D Non-Reflecting Mixing Plane. The implementation is based on the work done by Saxer and Giles (available at https://doi.org/10.2514/3.23618). To adapt the implementation of Saxer and Giles to an unstructured vertex-based solver such as SU2,  a new data structure, called **_CTurboVertex_**,  was created for the inflow and outflow boundaries.  The **_CTurboVertex_** class is a child of the class **_CVertex_**; thus, it maintains all the standard vertex properties. However, more than a standard vertex, the turboVertex posses:

1. the possibility of accessing the inflow and outflow boundaries on a span-wise manner (to impose different values of the mixing-plane condition along the blade-height)  and on a pitch-wise ordered manner (to compute the Spatial Fourier Transformation of the outgoing characteristics at the boundary for the non-reflectivity).

2. embedded  information on what kind of turbomachinery architecture is being simulated, so that the values of velocity at the boundaries can be imposed using the correct frame of reference.

So in a nutshell the flow solver is now capable of:

- simulating single 2D/3D turbomachinery blades  using 1D characteristic-based BC or Non reflecting BC
- simulating multi-stage 2D/3D turbomachinery via a Mixing-Plane interface, including also the mixing of the turbulent quantities
- handling any kind of turbomachinery architecture (AXIAL, CENTRIFUGAL, CENTRIPETAL and mixed such as CENTRIPETAL_AXIAL or AXIAL_CENTRIFUGAL)
- selecting 2 different kind of interpolation methods for the mixing-plane interface: LINEAR_INTERPOLATION and NEAREST_NEIGHBOR
- handling an unlimited number of blades/stages —> (zones)
- simulating turbomachinery which works on the NICFD regime such as supercritical CO2 compressors or Organic Rankine Cycle turbines.
- printing history_x.dat file  for multi-zone convergence monitoring
- printing some post-processing thermodynamic, kinematic and geometric quantities in some files inside a TURBOMACHINERY folder

The limitations of the turbomachinery capabilities are:

- the axis of rotation should be always the Z one
- MGL acceleration technique is not working because of some un-resolved issues with the periodic BC.

This pull request introduce also some limited design capability for turbomachinery applications. Using the AD tool, it is now available on the suite also a single zone discrete adjoint to compute the sensitivity  for 2D/3D turbomachinery blade with respect to certain performance parameters (ENTROPY_GENERATION, TOTAL_PRESSURE_LOSS, KINETIC_ENERGY_LOSS, ecc.). However, the optimization script only works for 2D blades since the FFD and the mesh deformation routines need still to be adapted to handle 3D blades (especially for problem on the deformation of the hub and shroud surfaces).


Common/include/config_structure.hpp
```diff
@@ -177,8 +180,11 @@ class CConfig {
   nMarker_SymWall,				/*!< \brief Number of symmetry wall markers. */
   nMarker_Pressure,				/*!< \brief Number of pressure wall markers. */
   nMarker_PerBound,				/*!< \brief Number of periodic boundary markers. */
-  nMarker_MixBound,				/*!< \brief Number of mixing boundary markers. */
-  nMarker_TurboPerf,				/*!< \brief Number of mixing boundary markers. */
+  nMarker_MixingPlaneInterface,				/*!< \brief Number of mixing plane interface boundary markers. */
+  nMarker_Turbomachinery,				/*!< \brief Number turbomachinery markers. */
+  nMarker_TurboPerformance,				/*!< \brief Number of turboperformance markers. */
+  nSpanWiseSections_User,			/*!< \brief Number of spanwise sections to compute 3D BC and Performance for turbomachinery   */
+  nMarker_Shroud,/*!< \brief Number of shroud markers to set grid velocity to 0.*/
```
fpalacios: What is a shroud in this context?

Common/include/config_structure.hpp
```diff
@@ -828,9 +847,26 @@ class CConfig {
   *default_htp_axis,          /*!< \brief Default HTP axis for the COption class. */
   *default_ffd_axis,          /*!< \brief Default FFD axis for the COption class. */
   *default_inc_crit,          /*!< \brief Default incremental criteria array for the COption class. */
-  *default_body_force;        /*!< \brief Default body force vector for the COption class. */
+  *default_extrarelfac;       /*!< \brief Default extra relaxation factor for NRBC in the COption class. */
+  unsigned short nSpanWiseSections; /*!< \brief number of span-wise sections */
+  unsigned short nSpanMaxAllZones; /*!< \brief number of maximum span-wise sections for all zones */
```
fpalacios: Once we accept this pull request we need to start thinking about how to unify wing geometrical analysis with blade geometrical analysis.

Common/include/dual_grid_structure.hpp
```diff
@@ -1245,4 +1245,118 @@ class CVertex : public CDualGrid {
 
 };
 
+/*!
+ * \class CTurboVertex
+ * \brief Class for vertex definition for turbomachinery (equivalent to edges, but for the boundaries).
+ * \author S. Vitale
+ * \version 5.0.0 "Raven"
+ */
+class CTurboVertex : public CVertex {
```
fpalacios: What is the main reason for not using CVertex?

Common/include/option_structure.hpp
```diff
@@ -625,13 +625,15 @@ enum ENUM_LIMITER {
   VENKATAKRISHNAN = 0,	/*!< \brief Slope limiter using Venkatakrisnan method. */
   BARTH_JESPERSEN = 1,  /*!< \brief Slope limiter using Barth-Jespersen method. */
   SHARP_EDGES = 2,       /*!< \brief Slope limiter using sharp edges. */
-  SOLID_WALL_DISTANCE = 3       /*!< \brief Slope limiter using wall distance. */
+  SOLID_WALL_DISTANCE = 3,       /*!< \brief Slope limiter using wall distance. */
+	VAN_ALBADA = 4         /*!< \brief Slope limiter using wall distance. */
```
fpalacios: Could you please update the description. Have your experimented with this limiter? I think Venkat's limiter was an improvement over the Van Albada. Anyway, it is great to count with another limiter implementation.

Common/include/option_structure.hpp
```diff
@@ -2765,120 +2863,128 @@ class COptionRiemann : public COptionBase {
...
+  unsigned short* & field; // Reference to the fieldname
+  string name; // identifier for the option
+  su2double * & var1;
+  su2double * & var2;
+  su2double ** & flowdir;
+  su2double * & relfac1;
+  su2double * & relfac2;
 
 public:
-	  COptionNRBC(string option_field_name, unsigned short & nMarker_NRBC, string* & Marker_NRBC, unsigned short* & option_field,
```
fpalacios: NR stands for Non-reflecting? isn't it? Maybe we should be a little more explicit and use something like COptionNonReflectingBC

Common/include/option_structure.hpp
```diff
@@ -2765,120 +2863,128 @@ class COptionRiemann : public COptionBase {
...
+      istringstream ss_4th(option_value[9*i + 5]);
+      if (!(ss_4th >> this->flowdir[i][1])) {
+        return badValue(option_value, "NRBC", this->name);
+      }
+      istringstream ss_5th(option_value[9*i + 6]);
+      if (!(ss_5th >> this->flowdir[i][2])) {
+        return badValue(option_value, "NRBC", this->name);
+      }
+      istringstream ss_6th(option_value[9*i + 7]);
+			if (!(ss_6th >> this->relfac1[i])) {
```
fpalacios: Is there something wrong with the number os spaces?

Common/src/config_structure.cpp
```diff
@@ -799,15 +822,58 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   /*!\brief MARKER_RIEMANN \n DESCRIPTION: Riemann boundary marker(s) with the following formats, a unit vector.
    * \n OPTIONS: See \link Riemann_Map \endlink. The variables indicated by the option and the flow direction unit vector must be specified. \ingroup Config*/
   addRiemannOption("MARKER_RIEMANN", nMarker_Riemann, Marker_Riemann, Kind_Data_Riemann, Riemann_Map, Riemann_Var1, Riemann_Var2, Riemann_FlowDir);
-  /*!\brief MARKER_NRBC \n DESCRIPTION: Riemann boundary marker(s) with the following formats, a unit vector. \ingroup Config*/
-  addNRBCOption("MARKER_NRBC", nMarker_NRBC, Marker_NRBC, Kind_Data_NRBC, NRBC_Map, NRBC_Var1, NRBC_Var2, NRBC_FlowDir);
-  /*!\brief MIXING_PROCESS_TYPE \n DESCRIPTION: types of mixing process for averaging quantities at the boundaries.
+  /*!\brief MARKER_NRBC \n DESCRIPTION: Riemann boundary marker(s) with the following formats, a unit vector. */
+  addNRBCOption("MARKER_NRBC", nMarker_NRBC, Marker_NRBC, Kind_Data_NRBC, NRBC_Map, NRBC_Var1, NRBC_Var2, NRBC_FlowDir, RelaxFactorAverage, RelaxFactorFourier);
```
fpalacios: Could you please update the description

Common/src/config_structure.cpp
```diff
@@ -799,15 +822,58 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
...
+   * the 2nd coeff is the number of iterations for the update, 3rd is the number of total iteration till reaching the final outlet pressure value */
+  addDoubleArrayOption("RAMP_OUTLET_PRESSURE_COEFF", 3, RampOutletPressure_Coeff, default_rampOutPres_coeff);
   /*!\brief MARKER_MIXINGPLANE \n DESCRIPTION: Identify the boundaries in which the mixing plane is applied. \ingroup Config*/
-  addMixingPlaneOption("MARKER_MIXINGPLANE", nMarker_MixBound, Marker_MixBound, Marker_MixDonor);
+  addStringListOption("MARKER_MIXINGPLANE_INTERFACE", nMarker_MixingPlaneInterface, Marker_MixingPlaneInterface);
+  /*!\brief SUBSONIC_ENGINE\n DESCRIPTION: Engine subsonic intake region \ingroup Config*/
+  addBoolOption("TURBULENT_MIXINGPLANE", turbMixingPlane, false);
   /*!\brief MARKER_MIXINGPLANE \n DESCRIPTION: Identify the boundaries in which the mixing plane is applied. \ingroup Config*/
-  addTurboPerfOption("MARKER_TURBO_PERFORMANCE", nMarker_TurboPerf, Marker_TurboBoundIn, Marker_TurboBoundOut, Kind_TurboPerformance, TurboPerformance_Map);
+  addTurboPerfOption("MARKER_TURBOMACHINERY", nMarker_Turbomachinery, Marker_TurboBoundIn, Marker_TurboBoundOut);
```
fpalacios: Please update the description

Common/src/geometry_structure.cpp
```diff
@@ -9515,167 +9666,1804 @@ void CPhysicalGeometry::SetVertex(CConfig *config) {
...
+              }
+            }
+          }
+        }
+      }
+    }
+
+#ifdef HAVE_MPI
+
+    MyTotalArea
```
fpalacios: Is there any change in CPhysicalGeometry::SetCoord_CG? It looks like that the diff tool is not working properly

fpalacios: Same comment here, are you modifying CPhysicalGeometry::SetBoundControlVolume?

Common/src/grid_movement_structure.cpp
```diff
@@ -238,61 +238,63 @@ void CVolumetricMovement::SetVolume_Deformation(CGeometry *geometry, CConfig *co
 
     CSysSolve *system  = new CSysSolve();
     
-    switch (config->GetKind_Deform_Linear_Solver()) {
+    if (LinSysRes.norm() != 0.0){
+      switch (config->GetKind_Deform_Linear_Solver()) {
         
         /*--- Solve the linear system (GMRES with restart) ---*/
         
-      case RESTARTED_FGMRES:
```
fpalacios: As before, is there any relevant change in this subroutine?

SU2_CFD/include/output_structure.hpp
```diff
@@ -47,7 +47,7 @@
 #include <time.h>
 #include <fstream>
 
-#include "solver_structure.hpp"
+//#include "solver_structure.hpp"
```
fpalacios: What is the reason for commenting out this line?

SU2_CFD/include/solver_structure.hpp
```diff
@@ -876,21 +876,18 @@ class CSolver {
...
-   * \param[in] conv_numerics - Description of the numerical method.
-   * \param[in] visc_numerics - Description of the numerical method.
    * \param[in] config - Definition of the particular problem.
-   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   * \param[in] marker_flag - Surface marker flag where the function is applied.
    */
-  virtual void BC_Inlet(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics, CNumerics *visc_numerics,
-                        CConfig *config, unsigned short val_marker);
-  
+  virtual void PreprocessBC_NonReflecting(CGeometry *geometry, CConfig *config, CNumerics *conv_numerics, unsigned short marker_flag);
```
fpalacios: I like the fact that you are using NonREflecting instead of NR

SU2_CFD/include/solver_structure.hpp
```diff
@@ -947,7 +944,8 @@ class CSolver {
    * \param[in] config - Definition of the particular problem.
    * \param[in] val_marker - Surface marker where the boundary condition is applied.
    */
-  virtual void BC_Engine_Inflow(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics, CNumerics *visc_numerics, CConfig *config, unsigned short val_marker);
+  virtual void BC_Supersonic_Inlet(CGeometry *geometry, CSolver **solver_container,
```
fpalacios: Yep... the diff tool is confused.

economon:
Salvo: thanks for the pull request, very exciting stuff! I will also add a review soon, but perhaps I will let you comment on some of the review by @fpalacios first.

Common/include/config_structure.hpp
salvovitale: When you simulate rotor blades with tip clearance, the top case of the machine is not rotating. So to make the simulations more close to "reality", with this option you can set the velocity to 0.0 for the wall upper boundary while the rest of the mesh is still steady-rotating. Of course this is only valid with the ROTATING_FRAME grid movement option.

Common/include/config_structure.hpp
salvovitale: I agree!

Common/include/dual_grid_structure.hpp
salvovitale: As I said in the description of the pull request, to impose Non reflecting BC and Mixing plane you need to access the boundary information on a span-wise and pitch wise ordered manner. Moreover, you need to convert the velocities in different frame of references  w.r.t. the turbomachinery architecture you are simulating. The turboVertex takes care of all of these needs.

Common/include/option_structure.hpp
salvovitale: I have a bunch of test-cases in which the Van Albada works much better than the Venkatakrishnan. Usually what happens is that the residuals stall with the Venka, but instead
they keep going down with the Van Albada.

Common/include/option_structure.hpp
salvovitale: I have changed everything from NRBC --> Giles_BC.

Common/include/option_structure.hpp
salvovitale: fixed

Common/src/geometry_structure.cpp
salvovitale: No I did not touch that routine. I think the diff tool is confused.

salvovitale: same here.

SU2_CFD/include/output_structure.hpp
salvovitale: I was experimenting some stuff. In truth this header file does not need to be included. For the moment, I have included again.

SU2_CFD/include/solver_structure.hpp
salvovitale: Now everything is BC_Giles so this one becomes PreprocessBC_Giles.

SU2_CFD/include/solver_structure.hpp
salvovitale: very confused!

Common/src/grid_movement_structure.cpp
salvovitale: Personally, I did not change this routine. @talbring did you modify this one?

talbring: the whole switch statement is wrapped in an if condition now, thats all (that is needed for multizone deformations). I dont know why github has problems even here ...

salvovitale:
Dear @fpalacios , 
thank you very much for taking the time to go trough the review process of the pull request.
I 've made all the change you requested. 

Regarding splitting in smaller pull requests I agree. However, I believe that usually there are 2 kind of pull requests:

1.  improving existing features
2. creating new features

While for the type 1 I think is possible to keep pull requests small, for the type 2 this is not always the case. In any case I agree we should always keep the review process in mind.

best

Salvo

Common/src/grid_movement_structure.cpp
salvovitale: Thanks @talbring  for the prompt answer!

Common/include/config_structure.hpp
LaSerpe: Sounds great! 
So is this, basically, a marker to specify a specific value of fluid speed over a certain boundary (which is exactly the speed of the grid)?
Cause I think that such possibility could be very useful also in applications other than turbomachinery.
So, maybe in the future, we could think to create a general BC to allow the user to specify the velocity over the boundary.

talbring:
Just to clarify things, I am not against pull requests with a lot of changes, @fpalacios. For me the problem in #412 is simply the fact that everything is included in one commit.

Common/include/config_structure.hpp
salvovitale: dear @LaSerpe,
for the moment, this option just set to 0.0 the velocity of the grid in the specified marker, but, again, only in case the ROTATING FRAME option is specified. I agree that there can be more application for it; perhaps in the future we can generalized this option for multiple uses.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -3642,6 +3886,244 @@ void CFluidDriver::SetInitialMesh() {
...
+  }
+
+  if (rank == MASTER_NODE){
+    SetTurboPerformance(ZONE_0);
+  }
+
+
+}
+
+void CTurbomachineryDriver::SetMixingPlane(unsigned short donorZone){
```
LaSerpe: Perhaps we could think about moving the mixing plane into the interpolator_class.
In this way we'll have all the routines that manage the communication between zones in the same place.

salvovitale: Indeed, in fact the Mixing-Plane interface is already implemented within the Transfer structure, in which we exchange info between zones. However, I did not use the the interpolator_class because it does not occur an interpolation between the nodes of the interfaces.

LaSerpe: I see, so maybe we should change the name of the class to something else.
Indeed "interpolation" (in a strictly mathematical sense) doesn't actually occur even for the nearest neighbour approach (zones are loosely coupled and the interface is basically treated as a non-homogeneous BC).
Having all the possible way of communicating between zones in the same src  would make the code much more clear, and possibly we could avoid creating the new driver class just by modifying the fluid driver a little bit.
By the way, we may do this in a later pull request (so nevermind) I'm just trying to figure out if that would make sense.

salvovitale: Yeah I see your point. However the new turbo driver is a child of the new FluidDriver, and only few routines are overwritten:
- the Run() routine because i need to use the iterator container in a slightly different way 
- the Monitor() routine in which I implemented ad hoc methods for ramping up the outlet pressure and the rotational speed to facilitate the convergence of multi-stage simulations. 

Indeed I could have implemented on the CFluidDriver, but there are two reason why I did not do it:
1. I wanted to avoid many if statements to customize the driver
2. (and more important) a clear structure of the turbo Driver also facilitate the implementation of his adjoint counterpart.

I' m pretty sure that when the structure is there we can always improve the code with future pull requests,  as we have been doing so far.  Anyway thanks for your valuable comments Giulio.

regards

Salvo

Common/include/config_structure.hpp
fpalacios: Thanks for the clarification!

Common/include/dual_grid_structure.hpp
fpalacios: Perfect. thanks!

Common/include/option_structure.hpp
fpalacios: Mike is going to be happy :-)

SU2_CFD/include/solver_structure.hpp
fpalacios: For the time being... no problem... let's try to do it different next time. Thanks!

SU2_CFD/src/driver_structure.cpp
fpalacios: Very interesting discussion. Please, after this pull request, do your best to revisit the issue to be sure that we have the best solution possible. Thanks

fpalacios:
Thanks @salvovitale for this contribution, the open-source community is going to appreciate a lot this new capability. 

In fact, I think it is a perfect timing to prepare a tutorial about turbo machinery  in https://github.com/su2code/SU2/wiki/User-Tutorials

As you know, the tutorial is a really important step. Otherwise, for an average users, it is difficult to know that this awesome capability is available. And... a large number of users will help us to improve any implementation.

Best,
Francisco

salvovitale:
Dear @fpalacios
thank you very much for merge this in. 

I agree that one or two tutorials are fundamental to let the user exploit the new capabilities. I am certainly counting  on the help of the community to make this new feature as much reliable and general as possible.  


regards

Salvo

PR #420, opened 2017-07-27
Travis CI Fix

economon:
Updates to .travis.yml to get the tests working after the Travis CI move to Ubuntu Trusty. Script is a bit cleaner, but there were also a few important changes to fix problems related to compilers and Python packages.


economon:
Merging this in immediately so we can finish the standing PRs.

PR #422, opened 2017-07-31
Numerics refactoring for CAvgGrad- and CUpwSca- classes

clarkpede:
I noticed there was a lot of duplicate code in the turbulent numerics files (`numerics_structure.hpp` and `numerics_direct_turbulent.cpp`).  Since I'm working with new model equations for hybrid RANS/LES, I didn't want to add to the technical debt.  So I decided to refactor the turbulence numerics using the [template design pattern](https://sourcemaking.com/design_patterns/template_method).

Basically, I moved common behavior for the viscous and the upwinding residuals to respective  base classes and then left the specifics of each model to derived classes.  I did my best to not change any of the underlying behavior.

The source residual calculations for SST and SA have little in common.  I did not apply the template design pattern to them.

I named the template classes `CUpwSca_Abstract` and `CAvgGrad_Abstract` because the `_Template` suffix was already taken.

I also noticed that there were almost no difference between the old `CAvgGrad_` classes and the `CAvgGradCorrected_` classes.  The gradient correction is such a small change that it's better to have a single class with a switch than two classes that are about 95% similar.  So I added a boolean property and logical switching if correction is to be applied to the gradients.  This boolean parameter is now a parameter for the constructor.

Overall, these changes trim down a lot of duplicate code (40% of the lines in `numerics_direct_turbulent.cpp`).  These changes could be extended to other areas (such as the transition model), but I left that to future pull requests.


SU2_CFD/include/numerics_structure.hpp
```diff
@@ -66,13 +66,13 @@ class CNumerics {
   su2double *Enthalpy_formation;
   su2double Prandtl_Lam;        /*!< \brief Laminar Prandtl's number. */
   su2double Prandtl_Turb;    /*!< \brief Turbulent Prandtl's number. */
-
```
fpalacios: Thanks for fixing these extra spaces

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2071,34 +2071,64 @@ class CUpwLin_AdjTurb : public CNumerics {
 };
 
 /*!
- * \class CUpwSca_TurbSA
- * \brief Class for doing a scalar upwind solver for the Spalar-Allmaral turbulence model equations.
+ * \class CUpwSca_Abstract
```
fpalacios: Makes a lot of sense to me. By the way, we are using a very basic linear upwind scheme based on the velocity. Have you ever tried to use a more elaborate upwind? (e.g. something similar to the upwind scheme that we use for the mass conservation equation like using Roe's averaged variables).

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2110,42 +2140,85 @@ class CUpwSca_TurbSA : public CNumerics {
 };
 
 /*!
- * \class CUpwSca_TurbML
- * \brief Class for doing a scalar upwind solver for the Spalar-Allmaral turbulence model equations.
+ * \class CUpwSca_TurbSA
+ * \brief Class for doing a scalar upwind solver for the Spalar-Allmaras turbulence model equations.
```
fpalacios: Thanks for fixing the typo in Allmaras!

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2110,42 +2140,85 @@ class CUpwSca_TurbSA : public CNumerics {
...
  * \ingroup ConvDiscr
  * \author A. Bueno.
  * \version 5.0.0 "Raven"
  */
-class CUpwSca_TurbML : public CNumerics {
+class CUpwSca_TurbSA : public CUpwSca_Abstract {
```
fpalacios: This is a good opportunity to fix this problem. If you have a while could you please eliminate all classes and subroutines with "TurbML": CUpwSca_TurbML, CAvgGrad_TurbML, CAvgGradCorrected_TurbML, CTurbMLVariable. ML stands for Machine Learning, that was an old implementation that was eliminated from the main SU2 distribution... but it seems that there are still some residuals. Thanks!

SU2_CFD/src/driver_structure.cpp
```diff
@@ -1525,9 +1525,9 @@ void CDriver::Numerics_Preprocessing(CNumerics ****numerics_container,
     /*--- Definition of the viscous scheme for each equation and mesh level ---*/
     
     for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
-      if (spalart_allmaras) numerics_container[iMGlevel][TURB_SOL][VISC_TERM] = new CAvgGradCorrected_TurbSA(nDim, nVar_Turb, config);
-      else if (neg_spalart_allmaras) numerics_container[iMGlevel][TURB_SOL][VISC_TERM] = new CAvgGradCorrected_TurbSA_Neg(nDim, nVar_Turb, config);
-      else if (menter_sst) numerics_container[iMGlevel][TURB_SOL][VISC_TERM] = new CAvgGradCorrected_TurbSST(nDim, nVar_Turb, constants, config);
+      if (spalart_allmaras) numerics_container[iMGlevel][TURB_SOL][VISC_TERM] = new CAvgGrad_TurbSA(nDim, nVar_Turb, true, config);
```
fpalacios: Thanks for simplifying the structure. And yes, a boolean works better... in fact, in the future we should probably only use the corrected version which is quite standard.

chaos-dev:
In reply to:
> What is the subroutine "FinishResidualCalc" is it a new name to call "ComputeResidual" if that is the case, for consistency with the rest of the code we should probably use "ComputeResidual"

I would have liked to do that, but there's two reasons why it has to be named differently.  The first is philosophical, the second is due to programming restrictions.
1. The `FinishResidualCalc` serves a different purpose.  Yes, this is where the computing of the residual happens.  But it is only a step in the larger, `ComputeResidual` method in a parent class (i.e. `CAvgGrad_Abstract::ComputeResidual`).
2. In the template method, you want the user to call the method from the parent class.  That's where the skeleton of your calculation is stored.  That means that you have to be able to call the `ComputeResidual` method from a derived class, such as `CAvgGrad_TurbSA`.  If both the base class method `ComputeResidual` and the derived class method `FinishResidualCalc` are named the same, then the derived class overrides the base class.

I'm not fixed on the name.  Honestly, I named it somewhat arbitrarily.  I just know that we can't name it "ComputeResidual" if we want to keep the interfaces consistent with other CNumerics classes. Do you have any other suggestions for a better name?

SU2_CFD/include/numerics_structure.hpp
chaos-dev: No, I haven't looked at that.  I've used the higher order upwind schemes in the mass conservation equations, but I haven't tried to apply them to the turbulence transport equations.

SU2_CFD/include/numerics_structure.hpp
chaos-dev: I'll make that change one of the next commits.  I had no idea what those extra classes were for.  Thanks for the clarification!

clarkpede: I had no idea what those classes were for.  Thanks for the clarification.  I'll make that change one of my next commits.

clarkpede:
In reply to:
> What is the subroutine "FinishResidualCalc" is it a new name to call "ComputeResidual" if that is the case, for consistency with the rest of the code we should probably use "ComputeResidual"

I would have liked to have named it the way you suggested.  But unfortunately, it's not possible. There's two reasons for this: one is more philosophical, the other is due to a programming restriction.

1. The real residual calculation is a larger method in the parent class (i.e. `CAvgGrad_Abstract::ComputeResidual`).  The method currently called `FinishResidualCalc` is just a step in that larger process.
2. Since the real residual calculation is in the parent class, we need to be able to call it from derived classes (i.e. `CAvgGrad_TurbSA::ComputeResidual`).  If we also name a separate method in the derived class `ComputeResidual`, then that method now overrides the method in the parent class.  Instead of performing the whole calculation, callers would only get the final step in the calculation.

I'm not fixed on the name.  Honestly, I picked it somewhat arbitrarily.  I just know that "ComputeResidual" isn't an option. What do you suggest as a better name?

SU2_CFD/include/numerics_structure.hpp
clarkpede: No, I haven't looked into using the higher order upwinding in the turbulence classes.  For the most part, our group has had other priorities (like implementing a higher order time-stepping).

clarkpede:
In reply to:
> You are adding the prefix "_Abstract" to the new parent classes. For the time being, as we are applying this only for the turbulence model, would it be possible to add also the key word "Turb". And... are you completely sure about using the word "Abstract" does it makes sense to use something like for example CAvgGrad_TurbBaseline instead of CAvgGrad_Abstract. What do you think?

I'm not fixed on "Abstract".  In fact, I wanted to use the suffix "Template" instead, and only decided on "Abstract" as a less satisfactory second choice. As the code currently stands, I totally agree with you.  "Turb" would be a better suffix.  But I was also thinking about possible changes in the future.

I can see these classes (`CAvgGrad_Abstract` and `CUpwSca_Abstract`) being used for the transition model (in `numerics_direct_transition.cpp`) as well. Do you agree? Do you think this refactoring should be applied to the transition model?

I didn't picture the transition model as part of this pull request's scope.  I wanted to include the transition model as a separate pull request.  If we do want to refactor the transition model to eliminate the duplicate code there, then I don't think "Turb" or "TurbBaseline" are good suffixes.

If you do want the duplicate code in the transition model eliminated, then which would you prefer: "Baseline", "Abstract", "Outline", or "Scalar"? Or do you have another idea?

SU2_CFD/include/numerics_structure.hpp
talbring: I don't think that this is intended. Usually in some editors the default is to remove empty lines when saving the file. In fact, I would suggest to revert these changes as they give the potential to lead to conflicts in the future. We should be consistent regarding the handling of white spaces and empty lines.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2865,138 +2897,168 @@ class CAvgGradArtComp_Flow : public CNumerics {
...
+ * \author C. Pederson, A. Bueno, and F. Palacios
  * \version 5.0.0 "Raven"
  */
-class CAvgGrad_TurbSA : public CNumerics {
-private:
-  
-  su2double **Mean_GradTurbVar;
-  su2double *Proj_Mean_GradTurbVar_Kappa, *Proj_Mean_GradTurbVar_Edge;
-  su2double *Edge_Vector;
+class CAvgGrad_Abstract : public CNumerics {
```
talbring: I would suggest to use a different name for this class. Simply CAvgGrad_Turb would be ok I think.

clarkpede: Did you get a chance to look at my earlier reply to Dr. Palacios' similar comment? What are your thoughts?  Should we put the suffix as "Turb", and change it in the future, or have the name be more general, such as "Baseline", "Abstract", "Outline", or "Scalar"?

economon:
@clarkpede, @talbring, @fpalacios: I think this is a great discussion, and a good opportunity to think about future expansion of the code with this naming (the changes to other classes could come later). I envision that we will use the same high-level conv/visc schemes abstracted here for the transition model but also for general scalar transport equations in the future, such as passive scalars, combustion models, etc. I am fine with "Turb" but "Scalar" could be the most general.

SU2_CFD/include/numerics_structure.hpp
talbring: Oops sorry. Didn't read that.

talbring:
I agree that if we plan to have it even more general in the future, then we should probably go with the term "Scalar".

fpalacios:
Yep, "Scalar" sounds great. As for me, just implement that small change and I would very happy with your implementation.

Thanks again,
Best,
Francisco

SU2_CFD/include/numerics_structure.hpp
fpalacios: Thanks for removing the TurbML stuff!

SU2_CFD/include/numerics_structure.hpp
clarkpede: I just reverted the unnecessary whitespace changes.  I agree that they should be eliminated to reduce the number of "false conflicts," and I also agree that SU2 could benefit from more consistent use of whitespace.

clarkpede:
All right.  I changed the suffixes from "_Abstract" to "_Scalar".  I think that resolves all previous issues.

I noticed Codacy flagged some code quality issues that predate this pull request.  I can change those in a separate pull request, unless any of you want me to do that here.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2071,34 +2071,64 @@ class CUpwLin_AdjTurb : public CNumerics {
...
+ *   are implemented by derived classes.  In order to add a new residual
+ *   calculation for a convection residual, extend this class and implement
+ *   the pure virtual functions with model-specific behavior.
  * \ingroup ConvDiscr
- * \author A. Bueno.
+ * \author C. Pederson, A. Bueno., and A. Campos.
  * \version 5.0.0 "Raven"
  */
-class CUpwSca_TurbSA : public CNumerics {
+class CUpwSca_Scalar : public CNumerics {
```
economon: For me, CUpwSca_Scalar is a bit redundant... CUpw_Scalar or CUpwScalar might be better. Perhaps the latter is best, as later child classes can be more specific to the model, e.g., CUpwScalar_TurbSA, etc.

fpalacios:
Hi @talbring are you happy with this version of the pull request? Thanks!

talbring:
Looks good now ! Thanks !

economon:
Looks good to me too now, thanks for the changes, merging...

PR #425, opened 2017-08-11
Bug Fixes

economon:
The history file and the continuous_adjoint.py, discrete_adjoint.py, and shape_optimization.py scripts are currently broken for many cases (importantly, the Navier-Stokes/RANS equations). This PR fixes bugs to remedy the issues.


talbring:
Thanks @economon for fixing this. That was indeed something we overlooked when adding the NZONE option ...

PR #426, opened 2017-08-19
Adjusting software style

IndianaStokes:
I have seen that depending on the developer. SU2 includes (or not) a space between "if" and "(" or "for" and "(", etc. 

From the coding style point of view I think it is better to have a common criteria:
"if (" "for (", ") {" and "else {"

I have implemented the previous convention in the software... it is the most popular one in order to differentiate "if(" or "for(" from function calls, it also improves the code readability.


fpalacios:
We should unify the coding strategy but... this change affects too many files. What do you think @economon. Should we unify the coding style? or just leave it as it is

petebachant:
By the way, there's also this cool tool called vera++ that allows you to check the style of the entire codebase. This could be integrated with Travis so make sure PRs are consistent with the coding style guide.

talbring:
Although we should be more strict about coding style, as @fpalacios already noted this is a change that affects too much code.  In my opinion this change does not contribute to readability and is not worth all the conflicts we will have probably have when this is merged ...

fpalacios:
Yep... I agree... we need to do something with style problem but, this is too much change.

economon:
Agreed that the style needs some attention. I am open to ideas about this. You may have noticed that I integrated Codacy recently to test it out: https://www.codacy.com. We have the style guide, but honestly, it's not getting much attention. We should automate checks or have scripts that automatically enforce style constraints. Vera++ sounds interesting in this regard. A little while back, @rsanfer and I had been working on indentation issues with the uncrustify tool.

Should we open this as an issue and get some input? Find volunteers to look into these things?

talbring:
Indeed that we should have a closer look on that. Especially indentation and whitespaces at beginning/end of line and blank lines are really something we should deal with soon. We can set up for example cpplint in travis to check for Googles C++ code style. I don't think that we should modify the code automatically, though ... (maybe just running uncrustify or astyle once to have everthing in the correct style and then we should enforce that in pull requests)

PR #427, opened 2017-08-20
Bug fixes...

economon:
for the incompressible discrete adjoint and python scripts, as well as new test cases for the incomp. primal and discrete adjoint.

### DIFF
--- a/SU2_CFD/include/solver_structure.hpp
+++ b/SU2_CFD/include/solver_structure.hpp
@@ -6442,6 +6442,7 @@ protected:
   Total_MaxHeat, /*!< \brief Maximum heat flux on all boundaries. */
   Total_CpDiff,      /*!< \brief Total Equivalent Area coefficient for all the boundaries. */
   Total_HeatFluxDiff,      /*!< \brief Total Equivalent Area coefficient for all the boundaries. */
+  Total_Custom_ObjFunc,        /*!< \brief Total custom objective function for all the boundaries. */
   Total_MassFlowRate;     /*!< \brief Total Mass Flow Rate on monitored boundaries. */
   su2double *Surface_CL,   /*!< \brief Lift coefficient for each monitoring surface. */
   *Surface_CD,          /*!< \brief Drag coefficient for each monitoring surface. */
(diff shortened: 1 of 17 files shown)


SU2_CFD/include/solver_structure.hpp
```diff
@@ -7195,7 +7202,21 @@ class CIncEulerSolver : public CSolver {
...
    */
   void SetTotal_CLift(su2double val_Total_CLift);
-  
+
+  /*!
+   * \brief Set the value of the custom objective function.
+   * \param[in] val_Total_Custom_ObjFunc - Value of the total custom objective function.
+   * \param[in] val_weight - Value of the weight for the custom objective function.
+   */
+  void SetTotal_Custom_ObjFunc(su2double val_total_custom_objfunc, su2double val_weight);
```
fpalacios: Thanks for adding the Custom_ObjFunc to the incompressible solver

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -689,7 +689,7 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config, unsigned short
   Total_CFx     = 0.0;    Total_CFy          = 0.0;    Total_CFz          = 0.0;
   Total_CT      = 0.0;    Total_CQ           = 0.0;    Total_CMerit       = 0.0;
   Total_MaxHeat = 0.0;    Total_Heat         = 0.0;    Total_ComboObj     = 0.0;
-  Total_CpDiff  = 0.0;    Total_HeatFluxDiff = 0.0;
+  Total_CpDiff  = 0.0;    Total_HeatFluxDiff = 0.0;    Total_Custom_ObjFunc=0.0;
```
fpalacios: Good catch!

SU2_PY/direct_differentiation.py
```diff
@@ -48,15 +48,19 @@ def main():
     parser.add_option("-n", "--partitions", dest="partitions", default=1,
                       help="number of PARTITIONS", metavar="PARTITIONS")
     parser.add_option("-q", "--quiet",      dest="quiet",      default='False',
-                      help="output QUIET to log files", metavar="QUIET")    
-    
+                      help="output QUIET to log files", metavar="QUIET")
+    parser.add_option("-z", "--zones", dest="nzones", default="1",
+                      help="Number of Zones", metavar="ZONES")
```
fpalacios: Why the number of zones is now an input of these scripts? I thought it was stored in the mesh information.

talbring:
Thanks for fixing this. We can merge this in right away.

SU2_PY/direct_differentiation.py
talbring: We have this because the python scripts also need to know the number of zones in order to have to correct file names.

economon:
Thanks for the quick reviews. Merging now.

PR #432, opened 2017-08-31
Output reorganization

IndianaStokes:
In this pull request I have reorganized the existing non-standard output. 

Before, it was not really clear how to create a new output and each developer had a different strategy. Now all the special outputs have the same structure and have been implemented in the same part of the software for the direct and adjoint problem: CDriver::Output() for direct analysis and CDiscAdjFluidDriver::SetObjFunction() for the AD.

The current "special" outputs are 

output->SpecialOutput_AnalyzeSurface
output->SpecialOutput_Distortion
output->SpecialOutput_SonicBoom
output->SpecialOutput_SpanLoad
output->SpecialOutput_HarmonicBalance
output->SpecialOutput_Turbo
output->SpecialOutput_ForcesBreakdown

I have also merged the existing "1D output" and "surface analysis" in a more functional subroutine called output->SpecialOutput_AnalyzeSurface.

In short, this is a very important improvement that clarify a lot how to implement a new non-standard output, including its relation with the AD.


LaSerpe:
Hi, this sounds very interesting!
I'll take the chance to raise a point: we still don't have a clear strategy to manage the output of multi-zone simulations.

Indeed, earlier we used to have the residuals plotted for each zone (so we used to have a line for each zone), currently we just get the residuals corresponding to ZONE_0 (this was probably changed in one of the latest pull request, I think it could be the one from @salvovitale ).

We should decide if to restore the previous "multi-residual" output or if to sum up residuals from each zone and print out the results.
Cause right now we can't know how the residuals are behaving in the rest of the domain.
We could perhaps create an additional "special" output to manage multi-zone computations.

cheers,
Giulio

SU2_CFD/src/output_structure.cpp
```diff
@@ -4845,30 +4781,21 @@ void COutput::SetConvHistory_Body(ofstream *ConvHist_file,
...
-        /*--- Get Mass Flow at the Monitored Markers ---*/
-        
-        
-        if (output_massflow) {
-          Total_Mdot = solver_container[val_iZone][FinestMesh][FLOW_SOL]->GetOneD_MassFlowRate();
+        /*--- Get flux-averaged values at the specified surface ---*/
+
+        if (output_surface) {
+
+          unsigned short iMarker_Analyze = 0;
```
fpalacios: This should be improved... at least the solver should compute an average of all the markers to be analyzed

PR #435, opened 2017-09-07
Specification of wall function treatment per individual marker

vdweide:
This PR implements the specification of a different wall function treatment per viscous wall marker. In the configuration file you do this as follows, where the default is NO_WALL_FUNCTION.

% Viscous wall markers for which wall functions must be applied. (NONE = no marker)

You can determine the type of wall function treatment for a given marker by calling the function CConfig::GetWallFunction_Treatment. I put a dummy call in CNSSolver::BC_HeatFlux_Wall and CNSSolver::BC_Isothermal_Wall, as an example.

It is the intention that the branches feature_Wall_Functions and feature_hom_wallModel merge with this version and work out the actual implementation for the FV solver and DG-FEM solver, respectively.


vdweide:
Paul and I talked about this yesterday and it could be that additional input parameters are required for at least some of the wall models. Those can be added to the input data for the wall function markers. @GomerOfDoom, and others, could you specify what additional input data is needed for some of the wall models?

Edwin

GomerOfDoom:
Hi All,

We will at least need a way to specify wall model thickness (distance away from the wall at which we exchange information with the flow solver).

I think it would also be good to be able to specify the number of wall model points (for a 1-D model) between the exchange location and the wall, the stretching factor, near-wall thickness, etc.

-Paul

talbring:
Maybe it makes more sense to have this changes in a separate branch for now ? This way we avoid potential problems for others in case there are some modifications of this config option necessary during the final implementation of the wall functions.

vdweide:
Tim,

You're right. I thought it was an easy thing to do, such that both the FVM and DG-FEM solvers could use it. But it seems to be more complicated than I originally thought. So let's close this PR for now.

Edwin

vdweide:
Added the possibility to add additional information for some of the wall models as requested by @GomerOfDoom. I hope this is general enough.

SU2_CFD/src/output_structure.cpp
```diff
@@ -4726,12 +4726,12 @@ void COutput::SetConvHistory_Body(ofstream *ConvHist_file,
           Total_IDC     = solver_container[val_iZone][FinestMesh][FLOW_SOL]->GetTotal_IDC();
         }
         
-        if (inv_design) {
+    /*  if (inv_design) {
```
economon: Everything looks good, other than a couple things that are commented, like this 'inv_design' section. Was this intended?

SU2_CFD/src/output_structure.cpp
```diff
@@ -8038,12 +8038,12 @@ void COutput::SpecialOutput_SpanLoad(CSolver *solver, CGeometry *geometry, CConf
         if (iSection == 0) {
           Dihedral_Leading = atan((Zcoord_LeadingEdge_ - Zcoord_LeadingEdge) / (Ycoord_LeadingEdge_ - Ycoord_LeadingEdge))*180/PI_NUMBER;
           Dihedral_Trailing = atan((Zcoord_TrailingEdge_ - Zcoord_TrailingEdge) / (Ycoord_TrailingEdge_ - Ycoord_TrailingEdge))*180/PI_NUMBER;
-          Dihedral = 0.5*(Dihedral_Leading + Dihedral_Trailing);
+     //   Dihedral = 0.5*(Dihedral_Leading + Dihedral_Trailing);
```
economon: Comment intended?

SU2_CFD/src/output_structure.cpp
```diff
@@ -8038,12 +8038,12 @@ void COutput::SpecialOutput_SpanLoad(CSolver *solver, CGeometry *geometry, CConf
...
         }
         else {
           Dihedral_Leading = atan((Zcoord_LeadingEdge - Zcoord_LeadingEdge_) / (Ycoord_LeadingEdge - Ycoord_LeadingEdge_))*180/PI_NUMBER;
           Dihedral_Trailing = atan((Zcoord_TrailingEdge - Zcoord_TrailingEdge_) / (Ycoord_TrailingEdge - Ycoord_TrailingEdge_))*180/PI_NUMBER;
-          Dihedral = 0.5*(Dihedral_Leading + Dihedral_Trailing);
+     //   Dihedral = 0.5*(Dihedral_Leading + Dihedral_Trailing);
```
economon: Comment intended?

SU2_CFD/src/output_structure.cpp
vdweide: Yes, because this variable is not used and leads to a compiler warning. This was merged in earlier, but apparently was reverted by another PR.

vdweide: I can remove it, if you prefer.

SU2_CFD/src/output_structure.cpp
vdweide: Same here. This is to avoid a compiler warning, which was merged in earlier and reverted again. Again, it can be removed.

SU2_CFD/src/output_structure.cpp
vdweide: Same here. This is to avoid a compiler warning, which was merged in earlier and reverted again. Again, it can be removed.

economon: Great. Can you please remove these extra commented lines to keep it clean? Then it's all good to merge.

PR #440, opened 2017-09-13
Fixing Warnings

VivaanKhatri:
My compiler is detection tens of warnings, I have fixed most of them.

Best,
Vivaan

### DIFF
--- a/Common/src/config_structure.cpp
+++ b/Common/src/config_structure.cpp
@@ -4458,7 +4458,7 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
                 cout << "The reference element size is: " << RefElemLength <<". "<< endl;
                 break;
               case VENKATAKRISHNAN_WANG:
-                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: " << LimiterCoeff <<". "<< endl;
+                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: 0.03."<< endl;
                 break;
               case BARTH_JESPERSEN:
                 cout << "Barth-Jespersen slope-limiting method." << endl;
@@ -4489,7 +4489,7 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
                 cout << "The reference element size is: " << RefElemLength <<". "<< endl;
                 break;
               case VENKATAKRISHNAN_WANG:
-                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: " << LimiterCoeff <<". "<< endl;
+                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: 0.03."<< endl;
                 break;
               case BARTH_JESPERSEN:
                 cout << "Barth-Jespersen slope-limiting method." << endl;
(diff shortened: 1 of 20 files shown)


SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -4226,10 +4226,11 @@ void CIncEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config)
 
       for (iVar = 0; iVar < nPrimVarGrad; iVar++) {
         
-        if (config->GetKind_SlopeLimit_Flow() == VENKATAKRISHNAN_WANG)
-          eps2 = LimK *( GlobalMaxPrimitive[iVar] - GlobalMinPrimitive[iVar] );
-        else
-          eps2 = eps1*eps1*eps1;
+        if (config->GetKind_SlopeLimit_Flow() == VENKATAKRISHNAN_WANG) {
+          eps1 = 0.03 * (GlobalMaxPrimitive[iVar] - GlobalMinPrimitive[iVar]);
```
economon: Looks like the VEKATAKRISHNAN_WANG limiter coefficient is hard-coded. Any reason to make this variable, or does the tuning within this new limiter take care of any required scaling?

VivaanKhatri: Thanks!

There is a good reason for that... my frustration using the original Venkat's limiter. The freedom to choose parameters is a double-edged sword... you have control but you can affect the results. With the original method (2 parameters)... I was able to obtain any result! 

For that reason, in this new implementation I decided to hardcode the recommended value for the parameter. So... users will have consisten solutions.

By the way, I have some results that point out to the fact that this version is much better than the original Venkat's limiter (convergence and accuracy)... maybe in the near future we should replace the original one by this one.

Best,
Vivaan

EduardoMolina: Dear all,

Thanks for the implementation @VivaanKhatri 

However, the eps value in VEKATAKRISHNAN_WANG can be view as percentage of the maximum range of the flow field, oscillations exceeding this value are limited.  In my opinion, this value should not be hard-coded and we can use the limiter coefficient from the config file. According to Wang's original "A fast nested multi-grid viscous flow solver", the value of 0.05 is a good starting candidate. 

Best,

Eduardo

VivaanKhatri: I have added the option LIMITER_COEFF to the VENKATAKRISHNAN_WANG limiter.

Best,
Vivaan

Common/src/config_structure.cpp
```diff
@@ -4458,7 +4458,7 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
                 cout << "The reference element size is: " << RefElemLength <<". "<< endl;
                 break;
               case VENKATAKRISHNAN_WANG:
-                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: " << LimiterCoeff <<". "<< endl;
+                cout << "Venkatakrishnan-Wang slope-limiting method, with constant: 0.03."<< endl;
```
economon: Thanks for adding the coefficient back in.. can you please update these cout statements to print the value of LimiterCoeff again?

VivaanKhatri: Done!

PR #441, opened 2017-09-13
Compiler warnings

vdweide:
This PR removes most of the compiler warnings from both the parallel and sequential executables when all compiler warnings are enabled. In this way it is much easier to check for compiler warnings when the source code is modified.


economon:
Couldn't agree more. In fact, I would recommend that all developers build with the "-Wall" and "-Wextra" compiler flags in order to see all warnings and to correct them as you go.

Thanks for cleaning these up.

vdweide:
There are actually quite a few warnings in the cgns and metis libraries as well. I didn't touch those, but would it be useful to take a look at those too?

Edwin

economon:
Yes, there are some warnings there. I have mostly avoided the Tecplot ones with "-Wno-unused-parameter." We could also contact the developers of those two packages directly (CGNS is now on GitHub).

PR #443, opened 2017-09-22
Preconditioning added to harmonic balance source term

jhowison:
Harmonic balance source term is subject to numerical instability. Source is preconditioned in StabilizeHarmonicBalance() using matrix presented by Thomas et al. in AIAA 2009-4270. This allows for higher CFL numbers in harmonic balance cases than current code.

### DIFF
--- a/SU2_CFD/src/driver_structure.cpp
+++ b/SU2_CFD/src/driver_structure.cpp
@@ -4,8 +4,8 @@
  * \author T. Economon, H. Kline, R. Sanchez, F. Palacios
  * \version 5.0.0 "Raven"
  *
- * SU2 Lead Developers: Dr. Francisco Palacios (Francisco.D.Palacios@boeing.com).
- *                      Dr. Thomas D. Economon (economon@stanford.edu).
+ * SU2 Original Developers: Dr. Francisco D. Palacios.
+ *                          Dr. Thomas D. Economon.
  *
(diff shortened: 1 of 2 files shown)


fpalacios:
Hi, thanks a lot for contributing, this is a very interesting and useful implementation. 

Unfortunately something strange has happened with the git diff tool and the merging is not possible... it seems that is trying to update the driver subroutine with an old version of SU2.

I'm very sorry about this but, would you be so kind to reimplement the changes in a clean copy of the developer's branch? so... only the relevant changes of your implementation will be added to SU2?

Just to check that nobody if going to break your implementation. In this case, I think it is also important to add a regression test. Could you please add one example that travis can control?

Thanks a lot for your hard work!

Best,
Francisco

jhowison:
I will try to get that updated in the next few days. I think I am pulling from develop branch, but the build keeps failing. I originally was modifying an older version of SU2 on my computer then pasting changes into development branch so I will make sure the changes work with the current develop branch and retry. I am new to github.

I will work on regression test after that. I had been testing the NACA 64a010 test case currently used for harmonic balance. The method improves max CFL from about 1.0 to at least 4.0. I have not checked higher numbers. Is it better to add a different case for regression, pitching NACA 0012 for example? 

Jason

jhowison:
I am going to close this pull request and attempt changes from a new request with clean version of develop branch.

PR #444, opened 2017-09-23
Grid deformation improvements

IndianaStokes:
In this pull request I have introduced several improvements in the grid deformation subroutines:

1. ComputeDeforming_Wall_Distance has been eliminated (no reason to keep that method).
2. ComputeSolid_Wall_Distance is now only computing distance to Euler and Navier-Stokes walls.
3. It is possible to specify the extension of the grid deformation based on the distance to the walls.
4. Non-dimensionalization of the distance has been eliminated.
5. Improvements in the I/O
6. Final residual is recalculated at the end of the deformation.

Best,
Indiana


talbring:
Thanks for the improvements! Just a quick suggestion: maybe we should use for the distance computation the already available ADT structure ? Just have a look at CPhysicalGeometry::ComputeWall_Distance on how to use it. Should be more or less a simple copy/paste.

IndianaStokes:
I have added the ComputeWall_Distance using the ADT. It looks great.

By the way, it was not my objective but... trying to improve the deformation capabilities of SU2 I found that the ILU0 implementation had a bug, I fixed the problem and now ILU0 preconditioned is way better than LU_SGS.

Fixing the bug had implications on the SU2 CFD solver... that is the reason why some simulations provide a slightly different results (in particular the discrete adjoint that it was using ILU0 by default).

Common/src/matrix_structure.cpp
```diff
@@ -1425,17 +1425,12 @@ void CSysMatrix::BuildILUPreconditioner(bool transposed) {
...
-            if (kPoint >= jPoint) {
-              
-              // WARNING: here we have a left multiply by Block_jk, should it
-              // be a right multiply to give Aik' = Aik - Aij*inv(Ajj)*Ajk?
-              
-              MatrixMatrixProduct(Block_jk, block_weight, block);
-              SubtractBlock_ILUMatrix(iPoint, kPoint, block);
-              
-            }
+            MatrixMatrixProduct(block_weight, Block_jk, block);
```
economon: Thanks for the fix for ILU. In your experience, is ILU performing much better than LU_SGS for both SU2_DEF and SU2_CFD too (flow problems)?

Common/src/config_structure.cpp
```diff
@@ -1562,16 +1562,18 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   addDoubleOption("DEFORM_TOL_FACTOR", Deform_Tol_Factor, 1E-6);
   /* DESCRIPTION: Deform coefficient (-1.0 to 0.5) */
   addDoubleOption("DEFORM_COEFF", Deform_Coeff, 1E6);
-  /* DESCRIPTION: Type of element stiffness imposed for FEA mesh deformation (INVERSE_VOLUME, DEF_WALL_DISTANCE, WALL_DISTANCE, CONSTANT_STIFFNESS) */
-  addEnumOption("DEFORM_STIFFNESS_TYPE", Deform_Stiffness_Type, Deform_Stiffness_Map, BOUNDARY_DISTANCE);
+  /* DESCRIPTION: Deform limit in m or inches */
+  addDoubleOption("DEFORM_LIMIT", Deform_Limit, 1E6);
```
economon: When is this new option for the deform limit useful? Is this what you meant by the 3rd bullet in your PR description, basically controlling the max value that will be used when far from a solid wall (to make the convergence behave better)?

talbring:
You just have to update the files in the TestCases repository (`/home/travis/build/su2code/SU2/TestCases/cont_adj_euler/naca0012/of_grad_directdiff.dat.ref` and `home/travis/build/su2code/SU2/TestCases/cont_adj_euler/naca0012/of_grad_cd_disc.dat.ref`)

Common/src/matrix_structure.cpp
IndianaStokes: Yes, ILU0 is working better (or much better) in both problems. Just try the following... RAE2822 turbulent simulation and
LINEAR_SOLVER_ERROR= 1E-10
LINEAR_SOLVER_ITER= 100

Now ILU requires ~20% less iterations than the LU_SGS to achieve the same level of convergence. Before, ILU was worst than LU_SGS... same situation with the grid deformation. 

Since ILU is working I have also changed my way to run SU2... Now I'm always running the solver with a CFL ~10 before it was ~1... the price is that I'm performing more LINEAR_SOLVER_ITER iterations but it is totally worth it and the CFL numbers are competitive with other solvers.

Now that I understand the ILU implementation I'm planning to work on a ILU1 implementation.

Common/src/config_structure.cpp
IndianaStokes: I found some particular problems where it was useful to limit the extension of the deformation. In fact, there was something similar in the solver (you can define a box and the solver only will move the points  that are inside the box). I found the box system a little bit complicated to define and I thought that something based on the distance to the surface was easier to understand... e.g. DEFORM_LIMIT=1 only moves the points that are 1m away from the solid surface.

PR #445, opened 2017-09-25
HB source term preconditioning

jhowison:
Harmonic balance source term is preconditioned using matrix presented in Thomas et al. AIAA 2009-4270. This allows stable CFL number to go from near 1.0 up to at least 6.0. I have not tested beyond that. Once merge is successful, I will create a regression test.

Jason


jhowison:
It looks like changes have gone through for driver_structure.hpp and driver_structure.cpp. I am not sure how to remove the previous commits that failed. It also appears a .hpp got placed in the src directory that needs to be removed.

Preconditioning is called automatically for HB cases in this setup. The current HB regression test may therefore be sufficient. If preconditioning should be an option in the cfg file instead I will leave open to discussion.

talbring:
Just do a `git rm SU2_CFD/src/driver_structure.hpp ` in the main directory and add/push this change.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -4811,6 +4818,174 @@ void CHBDriver::SetHarmonicBalance(unsigned short iZone) {
...
+  }
+
+  /*--- Loop over all grid levels ---*/
+  for (iMGlevel = 0; iMGlevel <= config_container[ZONE_0]->GetnMGLevels(); iMGlevel++) {
+
+    /*--- Loop over each node in the volume mesh ---*/
+    for (iPoint = 0; iPoint < geometry_container[ZONE_0][iMGlevel]->GetnPoint(); iPoint++) {
+
+      /*--- Get time step for current node ---*/
+      Delta = solver_container[ZONE_0][iMGlevel][FLOW_SOL]->node[iPoint]->GetDelta_Time();
```
arubino: I am actually working on a non-unifrom time sampling approach. Can we generalize this? It is not necessary for this PR, but maybe ```solver_container[ZONE_0][iMGlevel][FLOW_SOL]->node[iPoint]->GetDelta_Time()``` can be different for each zone?

jhowison: Antonio,

I do not believe the preconditioning matrix can be stored at setup, unfortunately. It includes the time step at each node, each iteration. It may be possible to invert matrix symbolically and maybe only a multiplicative factor of the time step would be needed at each node at each time step instead of inverting whole matrix. I will look into that.

I will add user option for preconditioning to configuration file. If this is done, however, will another regression case be necessary? One would need preconditioning on and one would need it off. 

I agree about a matrix inversion subroutine. I was surprised to see one was not already in the code. 

For your non-uniform time approach, the impact of time step in each sub-time level is not obvious to me. I have simply used the time step from the first sub-time level in the preconditioning matrix, and it is effective. That may need to be explored further for what you are proposing.

Jason

arubino: Jason,

Thanks. I see your point and actually it does not seem trivial to separate the 'delta' contribution. It is fine then, just leave it like this. About a new reg test, have you tested the implementation for more than one frequency (i.e. more than the 3 time instances of current reg test)? If so, perhaps it would be nice to have a regtest with more than one frequency and preconditioning.
Please, when you can, remove the hpp file as @talbring suggested.

A

jhowison: I have tested one and three harmonics for pitching NACA 64a010 and NACA 0012. More harmonics would be a better regression test since solution becomes stiffer as number of harmonics increases. I think three harmonics with preconditioning will be the way to go. I will work on that. Getting used to github is my biggest hurdle, but I'll get there.

Jason

jhowison: Antonio,

I am new to C++ so getting the option to turn preconditioning on/off is not as straightforward as I hoped. I am trying to follow same structure as CFL_ADAPT, for example, but it does not quite work. If it is easier, I can push my work back, but I'll summarize the new lines here since it is only a few.

In config_structure.hpp I am adding:
bool HB_Precondition
bool GetHB_Precondition(void);

In config_structure.cpp I am adding (where HB_PRECONDITION is in .cfg file and turned on with YES):
addBoolOption("HB_PRECONDITION", HB_Precondition, false);

In driver_structure.cpp I am adding:
if (config_container[ZONE_0]->GetHB_Precondition() == YES) {
  StabilizeHarmonicBalance();
}

The driver_structure command seems to be the issue. 

Thanks,
Jason

arubino: Jason, 

that is a good "template" to follow. Did you add the corresponded inline function in configure_structure.inl ?
e.g. inline bool CConfig::GetCFL_Adapt(void) { return CFL_Adapt; }

jhowison: That is the missing piece. I had been wondering how that information actually got to the container. Let me give that a shot. Thanks.

Jason

fpalacios: Hi Jason,

Thanks a lot for this contribution... this is great. We are looking forward to see this implemented in the software and hopefully with the regression test. Just two small comments:
- As you have noticed the files that you have committed have changed the whitespacing... it doesn't seem a big deal but the problem is that it could affect future merging... What editor are you using to modify SU2? it seems that it automatically removed any white space in the empty lines. @talbring, @economon  what do you recommend? The change goes in the correct direction, isn't it? Should we accept the change or reimplement the subroutines in a clean copy (I think we should accept the change... but I would like to know your opinion).
- SU2 exists thanks to contributors like you,  we really want to acknowledge your work... so feel free to add your name and or institutions to the subrouines or files that you have created/developed.

Thanks again for your contribution.
Best,
Francisco

PS.- @arubino thanks for your review, please let us know if you are done with your review so we can accept the implementation once we decide what to do with the white spacing problem. Thanks!

arubino: The changes look good to me. There is indeed the issue of white spaces as outlined by @fpalacios  ( which also makes difficult to follow the actual changes in the code ). However, they seem to improve the files by removing necessary spaces. What do you think?

To approve the changes I was waiting for the regression test we discussed. Please, let me know.

talbring: We should find a definite solution of the whitespace issues. The best would be to simply agree on not having any whitespaces at the end of a line or in empty lines. This is also the default for almost all editors (they remove them upon saving the file) and also the reason why we always see this changes in pull requests. We should do that for all files in a separate pull request. 

Although there might be a lot of conflicts for people when they try to merge that, you can tell `git merge` with ` -Xignore-space-change` to ignore whitespace changes altogether.

rsanfer: Hi all,

I am currently working on the merge of the coupled AD solver for FSI, and definitely the changes in spaces are a big issue. I am manually going over all the files after doing the merge with -Xignore-space-change, and it has been (and still is) a massive task. It's taking me forever, but I am doing it this way to minimize impact on other's contributions. 

I would ask Jason to please correct the spacing in this pull request and maintain the previous structure, otherwise it will generate big problems when merging. Particularly, driver_structure.cpp and config_structure.cpp are two of the files in which I have had most problems.

My suggestion would be to implement any whitespace change as the last change in develop before the new release of the code (the next merge of develop into master). Then, I would establish a clear, fixed criteria (possibly provide the settings for different IDEs), and then enforce it from then on.

Cheers,
Ruben

talbring: I fully agree with @rsanfer on that and would also suggest the approach he mentioned.

jhowison: I am using Atom, which is the editor that comes with GitHub desktop. Normally, I would just use gedit or vi, but I switched to GitHub desktop for this pull request when I had the initial issues with passing regression tests. I have not noticed the whitespacing issue, but I can go back and address that for this pull.

Does the regression test have special requirements? For example, the preconditioning helps significantly with the current HB test case, particularly with more harmonics. There is also the classic pitching NACA 0012 case, but that case is not nearly as stiff as the NACA 64a010 so preconditioning does not offer much improvement. Any preference?

Once we decide on a case, do I just push the .cfg and .su2 files to appropriate new folders and that is it?

Thank you all for your input and patience. I have a few more options I can offer the HB piece of the code now that I am over the GitHub learning curve (almost).

Jason

jhowison: I see all the whitespace issues now. I have no idea why it is doing that. I'll redo those files with a different editor.

Jason

arubino: Jason, many thanks! I would suggest to go for the NACA 64a010 with more than one frequency. I I was actually wondering, does the preconditioning improve turbulent calculations? I don't see any preconditioning of turbulent source term. Is it possible to add that as well? If so the NACA 64a010 could be added as RANS reg test. Let us know what you prefer.

rsanfer: Thank you very much, Jason, greatly appreciated. I normally use meld to visualize if I have introduced this kind of changes changes, it works quite well. 

This spacing issue is definitely something that we will need to address soon, @talbring @economon @fpalacios . I will be participating more in the coming weeks than I could these last months, in which it was impossible for me, so I can probably help in sorting this for the next release.

jhowison: It should be fine for turbulent calculations. I will play around with it and make that the regression test then.

Jason

economon: @rsanfer yes, it would be great to standardize and correct this issue. If you have time, maybe we can come back to the branch where we were playing with uncrustify, so that we can ship a script that can always be run by developers to correct spacing/indentation. We can even put this into the regressions tests to automatically detect issues ahead of time.

rsanfer: @economon yep, I can probably give it another iteration to the uncrustify script that I built a while ago. We need to make a consistent way of merging in the changes, though, I remember I was having trouble with that.

economon: @rsanfer : we'll probably have to play around more with the formatting of branches using a script.. I expect there is some combination we can find that works to avoid conflicts for everyone. Either everyone can apply the same script before their next merge with develop to avoid conflicts, or we work some magic by reformatting develop and creating new feature branch copies that get merged with folks' existing versions.

jhowison:
I believe this request is fixed with requested changes and white space removed. I have added a grid file and .cfg file to the respective test case repositories. I am not sure where to add a solution file if this is to become a regression test.

Jason

arubino:
Thanks, Jason! Can you please add the test-case in the python scripts for the regression test? You can find them in the TestCases folder.

jhowison:
I think I see how the regressions work now. I was thinking the cases were restarted from a converged solution. It looks as though the regression works by comparing the results after a set number of iterations from a non-restart case? That means the values I need to enter into the python script are the residuals and lift/drag coefficients after that number of iterations I set in the regression test?

Jason

arubino:
Hi Jason, Travis is failing with the following error:

```  
File "serial_regression.py", line 1116, in <module>
    main()
  File "serial_regression.py", line 508, in main
    test_list.append(turbulent_hb)
NameError: global name 'turbulent_hb' is not defined
```

are you addressing this?

jhowison:
Antonio,

I have been trying to figure the error message out, but I have been unsuccessful. I've only modified serial_regression.py and parallel_regression.py. Is there a file where turbulent_hb needs to be defined or does the code read the names from the test case repository?

We may want to change structure of HB test case folder since there are now two cases. The turbulent case has a sub-directory in there but the inviscid does not.

Jason

TestCases/serial_regression.py
```diff
@@ -496,6 +496,17 @@ def main():
...
+    # Turbulent pitching NACA 64a010 airfoil
+    harmonic_balance           = TestCase('turbulent_hb')
+    harmonic_balance.cfg_dir   = "harmonic_balance/turbulent_hb"
+    harmonic_balance.cfg_file  = "davis.cfg"
+    harmonic_balance.test_iter = 25
+    harmonic_balance.test_vals = [-1.784954,-5.471762, 0.007973, 0.077757] #last 4 columns
+    harmonic_balance.su2_exec  = "SU2_CFD"
+    harmonic_balance.timeout   = 1600
+    harmonic_balance.tol       = 0.00001
+    test_list.append(turbulent_hb)
```
arubino: Jason, 
I see two issues here:
1. please add the mesh file in the corresponding TestCases repository https://github.com/su2code/TestCases . At the moment your input mesh is not given.
2. you named your testcase *turbulent_hb* but here I still see the old name *harmonic_balance*. That's why I would expect an error at line 508. Try to change the names accordingly (e.g.  `harmonic_balance           = TestCase('turbulent_hb')` -> `turbulent_hb = TestCase('turbulent_hb')` and so on).

Maybe we could choose a different name to include the use of preconditioning; something like *hb_rans_preconditioning* ?

jhowison:
Antonio,

We will see if this check passes with corrections to .py files, but the mesh file is sitting in a pull request in TestCases repository. I could not add to that repository as part of this pull request. I have a feeling both checks are going to fail unless they can access each other.

Jason

arubino:
@economon for the regression test of this PR travis in failing in the TestCases repository, I guess for time limit. Do they have different time limits?. I would propose to merge that PR https://github.com/su2code/TestCases/pull/17) so that we can close this one once the reg test is passing.

jhowison:
I will caution that the regression test that I added related to this pull has yet to actually run. I don't know why it is running out of time, but it would be good to see it actually pass before merging. I don't know how to force a retry.

Jason

SU2_CFD/src/driver_structure.cpp
```diff
@@ -4621,13 +4621,23 @@ void CHBDriver::Run() {
...
+
+  }
+
+  for (iZone = 0; iZone < nZone; iZone++) {
+
+    /*--- Update the harmonic balance terms across all zones ---*/
     iteration_container[iZone]->Update(output, integration_container, geometry_container,
-        solver_container, numerics_container, config_container,
-        surface_movement, grid_movement, FFDBox, iZone);
+                                       solver_container, numerics_container, config_container,
```
economon: Something funny happened with the white space here. Can you please check this (make it adhere to normal style)?

SU2_CFD/src/driver_structure.cpp
```diff
@@ -4796,6 +4806,174 @@ void CHBDriver::SetHarmonicBalance(unsigned short iZone) {
...
+  unsigned short nVar = solver_container[ZONE_0][MESH_0][FLOW_SOL]->GetnVar();
+  unsigned long iPoint;
+  bool implicit = (config_container[ZONE_0]->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+  bool adjoint = (config_container[ZONE_0]->GetContinuous_Adjoint());
+  if (adjoint) {
+    implicit = (config_container[ZONE_0]->GetKind_TimeIntScheme_AdjFlow() == EULER_IMPLICIT);
+  }
+
+  /*--- Retrieve values from the config file ---*/
+  su2double *Source     = new su2double[nZone];
```
economon: Please check that this array is deleted.

arubino:
Any update or help needed here to close this PR?

arubino:
Jason, many thanks! looks good to me, can you just address the changes that @economon was suggesting? We are ready to merge this in.

economon:
@jhowison : thanks for the extra effort to clean things up. It is very important to have community involvement, and it's a very nice contribution! Nice to see you and @arubino work together on this too. We'll merge once the tests pass.

PR #448, opened 2017-10-02
Feature ILU(n)

IndianaStokes:
ILU preconditioner implementation with an arbitrary fill-in level (n).

In my opinion ILU(1) is a good compromise between time and accuracy. As I pointed out in my previous pull request a fully working ILU preconditioner changes the way SU2 can be (should be) used... now CFL>>10 is a standard in almost any run with a decent grid.

The fill-in level is controlled with LINEAR_SOLVER_ILU_FILL_IN and, instead of the ILU0 option, we are using just ILU.

### DIFF
--- a/Common/include/config_structure.hpp
+++ b/Common/include/config_structure.hpp
@@ -481,6 +481,7 @@ private:
   unsigned long Linear_Solver_Iter;		/*!< \brief Max iterations of the linear solver for the implicit formulation. */
   unsigned long Linear_Solver_Iter_FSI_Struc;		/*!< \brief Max iterations of the linear solver for FSI applications and structural solver. */
   unsigned long Linear_Solver_Restart_Frequency;   /*!< \brief Restart frequency of the linear solver for the implicit formulation. */
+  unsigned short Linear_Solver_ILU_n;		/*!< \brief ILU fill=in level. */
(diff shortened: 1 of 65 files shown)


talbring:
That sounds very nice! Thanks for this update. 

Btw: I fixed the failing regression test in the TestCases repo. There was a whitespace missing in the reference file. That's why the file diff returned an error.

salvovitale:
Dear @IndianaStokes,

First of all, I would like to thank you for the many contributions you are giving to SU2.

I have just a question? Did you test this improvement in convergence rate with both JST and Roe 2nd order limiter?

regards   

sv

IndianaStokes:
Thanks @salvovitale and @talbring

As you know, the main consequence of a better linear solver is that it is possible to increase the CFL number (independently of the spatial discretization). In other words... the benefits are for JST and Roe. By the way, I have found that a fill-in level of 1 is too expensive for 3D problems (great for 2D)... I have reduced the default value from 1 to 0... but feel free to increase the number if you are looking for very large CFL numbers.

I think that regression tests are taking too much time and they don't finish on the assigned time... hopefully the change from 1 to 0 fill-in level will resolve the problem, otherwise I'll reduce the time on some of the tests.

PR #451, opened 2017-10-09
Spatial discretization renaming

JSmith36:
The objective of this pull request is to clarify the name of some parameters. I found that:
1. LIMITER_COEFF was confusing because: A) it is a parameters linked only to the Venkat's limiter B) Venkat's original and Wang's version use the same parameter but with different value! (0.5 vs. 0.05)
2. I found AD_COEFF_FLOW very confusing too: A) You can think that AD stands for Automatic Differentiation B) AD is too convoluted... it is just JST!
3. REF_ELEM_LENGTH is mainly linked to Venkat's limiter... and a constant value is recommended (0.1 or 5.0 depending on the units). In the method REF_ELEM_LENGTH is multiplied by LIMITER_COEFF... so why two parameters to define a single constant (EPS)?... too many options for the user...

The current config file to define the limiters or the pressure sensor is as follow:

% ----------- SLOPE LIMITER AND DISSIPATION SENSOR DEFINITION -----------------%
%
% Coefficient for the Venkat's limiter (upwind scheme). A larger values decrease
%             the extent of limiting, values approaching zero cause
%             lower-order approximation to the solution (0.05 by default)
VENKAT_LIMITER_COEFF= 0.05
%
% 1st, 2nd and 4th order artificial dissipation coefficients for
%             the JST ( 0.15, 0.5, 0.02 by default )
JST_SENSOR_COEFF= ( 0.15, 0.5, 0.02 )
%
% Coefficient for the adjoint sharp edges limiter (3.0 by default).
ADJ_SHARP_LIMITER_COEFF= 3.0
%
% 1st, 2nd, and 4th order artificial dissipation coefficients for
%             the adjoint JST method ( 0.15, 0.5, 0.02 by default )
ADJ_JST_SENSOR_COEFF= ( 0.15, 0.5, 0.02 )

This config file is much more clear for people that are not familiar with SU2.

Best,
Jonathan


economon:
I agree with the proposed changes, but I also think this could go farther. Some of the most common issues for users involve misuse of the *_ORDER options, dissipation coefficients, and limiters.

Overall, it is still not very clear for a user how the centered schemes work. For instance, Lax is always first order and JST is always second order, but sometimes there are not error messages when trying to change the "ORDER" option when using these schemes. Also, if we rename the coefficients for JST as proposed, shouldn't we also separate the first coefficient that is only used for Lax? It would make treating the dissipation more clear.

fpalacios:
Lax vs. JST is a very good point... As Tom pointed out we should add JST_SENSOR_COEFF and LAX_SENSOR_COEFF... to be consistent... that for sure.

I don't like the ORDER stuff either... it is very confusing! My suggestion is to eliminate the SPATIAL_ORDER_FLOW option and use a boolean UPWIND_MUSCL I think that Monotonic Upwind Scheme for Conservation Laws is the correct term for what we are trying to say with first or second order. As you know MUSCL implies the use of a limiter (by definition) but we can use SLOPE_LIMITER_FLOW = NONE to obtain a 2nd order without limiter if we really need it.

What do you think?
Best,
Francisco

JSmith36:
I going to proceed with the implementation of a

% Monotonic Upwind Scheme for Conservation Laws (TVD). Required for
%           2nd order upwind schemes (NO, YES)
MUSCL_FLOW= YES

instead of the existing option... SPATIAL_ORDER_FLOW

fpalacios:
This is a big change that will affect all the config files... could you please add a meaningful error message if somebody uses the old options? Thanks!

PR #472, opened 2017-11-21
Some improvements to the incompressible solver

talbring:
We did some modifications to the incompressible solver. In particular:

- Leftovers from the compressible solver are removed in the non-dimensionalization routine, the time-step computation and the Jacobian computation in numerics.
- The surface analysis now works correctly for incompressible cases and the mass flux average finally works (only uses the modulus of the mass flux now). 
- There is now again a BC_Isothermal_Wall() routine. It just features as a wrapper for the BC_HeatFlux_Wall... until we found a better solution (maybe adding MARKER_WALL/MARKER_NO_SLIP or something).


koodlyakshay:
Can we set Velocity_Ref to 1.0 and Alpha and Beta to 0.0 in case ModVel_Freestream is zero in SetNondinemsionalisation routine? There might some cases of pure rotation with zero free stream velocity which makes many variables in Pressure_Forces/Momentum_Forces/Friction_Forces undefined because Velocity_Ref goes to zero.

talbring:
@koodlyakshay i think in that case you can use the MARKER_INLET which does the same as MARKER_FAR, except that you can provide a velocity that is different from FREESTREAM_VELOCITY,

talbring:
Merging this in.

PR #480, opened 2017-11-29
Temporarily reduce regression test runtime

talbring:
I reduced the number of iterations for some of the regression tests. They need ~ 45 min now. This should at least temporarily help.

Along with that I found one or two bugs ...

Nevertheless, all of this shouldn't affect a lot of people, so I hope we can merge this in asap, to avoid any more travis runs to fail :)


rsanfer:
Hi Tim,
thanks, this is really useful. Let's wait until the test cases finish, but I agree, we should get this in sooner than later. The regression test issue is slowing us all down, definitely something to get fixed.

TestCases/serial_regression.py
```diff
@@ -93,8 +93,8 @@ def main():
     fixedCL_naca0012           = TestCase('fixedcl_naca0012')
     fixedCL_naca0012.cfg_dir   = "fixed_cl/naca0012"
     fixedCL_naca0012.cfg_file  = "inv_NACA0012.cfg"
-    fixedCL_naca0012.test_iter = 500
-    fixedCL_naca0012.test_vals = [-5.856803,-0.469459,0.299943,0.019385] #last 4 columns
+    fixedCL_naca0012.test_iter = 100
```
economon: You may have already checked this, but does 100 iterations cover at least one change of the AoA with fixed Cl mode? Just want to make sure we're covering that.

talbring: I didn't think of this ... but fortunately in the config file the UPDATE_ALPHA is set to 5, so it should be covered.

economon: I think that is actually the number of times to update over the length of the calculation. Can you please double check this? I think we can just add a few iterations to this one if necessary...

talbring: I just checked, the interval when alpha is updated is computed as EXT_ITER/UPDATE_ALPHA in the code.

PR #488, opened 2017-12-15
Feature py wrapper

tobadavid:
This PR improves the capabilities of the Py wrapper by exposing new members of the driver:

- Extracting information (tag, type, number of vertices, ...) on the different markers involved in the computation.
- Adding more Set/Get functions for vertices values.
- Adding the possibility to customize the computation with non-uniform and unsteady boundary conditions, currently implemented for solid boundary motion (FSI applications) and solid thermal quantities (temperatures or heat fluxes, CHT applications). This could be extended to other type of BC's (e.g. user-defined inlet velocity profile).
- Adding proper reg tests for the Python wrapper (see also PR#19 from TestCases repo that has to merged before.)

EDIT: Reg tests still to be committed once PR in TestCases is merged.


talbring:
Hey David,

to define custom BCs inside python is an amazing feature! I can see a lot of potential there. Thanks for that! 

Since the python wrapper functionality is growing, what do you think about moving all of these wrapper routines to a separate file cpp file? 
Tim

tobadavid:
Hi Tim,

Thanks!

I totally agree on the fact that Python-related functionalities could be in a separate file for code clarity. Anyway, considering only the C++ structure, those functionalities are still members of the CDriver (and all child classes). So if you are OK with the fact that having members of the same C++ class in different cpp files, I am OK too. 

However, if you mean creating a new class for the Python wrapper (like CPyWrapper or whatever), and leave the CDriver "clean", this might be more tricky. Indeed, since the wrapper has to be a top-level class, it is better to have access to all the other main classes (config, geometry, solver, ...) and the CDriver was the perfect candidate (it instanciates all of them). So now if we create a new top-level class next to the CDriver, the direct link with all those classes is lost. Basically we will have to create accessors like CDriver::GetConfig(), CDriver::GetGeometry(), ... that will return the different containers to be used by the wrapper. Obviously this is possible but not in a short time, besides this might be a significant change in the top-level code structure.

We could also think about exposing thoses CConfig, CGeometry, ... classes to Python in addition to just  the driver, I already made some local tests and it worked great but this approach is less user-friendly.

David

config_template.cfg
```diff
@@ -587,6 +587,9 @@ MARKER_MONITORING = ( airfoil )
 % Format: ( marker name, wall function type, ... )
 MARKER_WALL_FUNCTIONS= ( airfoil, NO_WALL_FUNCTION )
 %
+% Marker(s) of the surface where custom thermal BC's are defined.
+MARKER_CHT = ( NONE )
```
economon: The new Python CHT capability is very interesting.

Just too make sure we avoid conflicts: @talbring @oleburghardt, do you see any potential conflicts with the options here for doing CHT through the Python wrapper as opposed to the native implementation you are working on?

Common/src/geometry_structure.cpp
```diff
@@ -14743,6 +14766,27 @@ void CPhysicalGeometry::SetPeriodicBoundary(CConfig *config) {
   
 }
 
+void CPhysicalGeometry::SetCustomBoundary(CConfig *config){
```
economon: It is a little peculiar for me to see physical BCs set in the geometry class.. Is there a way to put these into the solver class(es), or is there currently an issue with data encapsulation or access that disallows this?

tobadavid: I put it there for easy access to "node" and "vertex" arrays. However I don't see any issue if I put it into the CSolver and CEulerSolver classes. I just have to pass an additional CGeometry as argument in order to keep the access to node and vertex.

Note that this routine is for now called only at SU2 initialization, just to guarantee that if the user do not provide BC values from Python, the computation can still run with the default values read from the config file.

config_template.cfg
oleburghardt: Concerning the options: No, not really.

We can go for the boundary marker 'MARKER_CHT' (ours is called 'MARKER_CHT_INTERFACE' but that's just a renaming - including all the internal config variables of course - ... I can adapt ours, no trouble whatsoever).

The differences in both implementations could be more problematic.
Both will have some things in common on the solver's level (above, it's either controlled by the python calls or by a special native CHT driver).
I just had a very brief look into the pyWrapper-branch. If I got it correctly, it's "just" that custom boundary values are read in BC_Isothermal_Wall or BC_HeatFlux_Wall when a boundary is (additionally?) marked as CHT.
I would recommend an own BC routine for that. Main reason is that things might be more difficult than either setting one or the other. And additionally, we could set it up in a way that it fits both needs without blowing up the standard solver routines too much.

One other difference is that in this branch the interface values are stored within the geometry class (and not as solver member variables in my case). That's a legitimate way to go, too, I guess. I could imagine to adapt our code here as well so that it makes the boundary condition implementation more generic :-).

Regards, Ole

tobadavid: Thank you for the comments @oleburghardt  !

You are right, when the BC_HeatFlux or the BC_Isothermal are *additionally* marked as CHT ( see MARKER_CHT in the config file) they will just read the custom values stored in the geometry points instead of the one specified in the config.

My implementation is very light and does not need to stay the way it is. It is much easier to me to change the implementation than you changing yours. Originally, I am working on the Py wrapper and not on CHT in particular. So if there is a full CHT implementation the wrapper should not interfere with it.

I will have a look to your branch and see how you proceed. However some hints from you might be a valuable help in order to identify the relevant routines:

- How/where do you store the temperature/heat fluxes used for the fluid part of the computation ?
- When/where do you propagate the (fluid) interface temperatures/heat fluxes from the finest to the coarser grids (multi-grid) ?
- You leave the BC_HeatFlux_Wall and BC_Isothermal_Wall as they are and you use a dedicated BC type for CHT, is that right ?

Thanks,

David

talbring:
We can keep the wrapper routines as part of the CDriver class, but moving them to a separate file could already improve readability in my opinion.

config_template.cfg
oleburghardt: (Sorry for the late reply - I've been occupied with all these christmas duties... :D)

From a very practical point of view, your branch will be merged into develop first. And from this point on it's me who should adapt (possibly dublicate) variables and especially config options since the user should not be bothered with it. It's not so much stuff anyway, so I'm pretty sure that there will be no interference from your branches side :-).

1. So far, I am only using a heat equation solver that I wrote myself which is - in the fluid domain - weakly coupled to the flow solver. Within the heat solver, the CHT variables are simply stored as class variables. (I am planning to also adding a coupling to the compressible solver. I guess I will do the same thing there, too.) I like the idea to store them in the geometry class.

2. I have not used multigrid technique so far. I don't even know how this is done for fluid-fluid-couplings (does someone?). In either case, the routines in transfer_structure.cpp only see the different solvers, so they shouldn't be bothered if they are instantiated on the finest or a multigrid level I guess... Hm.

3. Yes. There seem to be some primitive methods for CHT which work quite stable but slow, but also slightly more complicated ones. Too much code in the end for putting it into the standard ones, I assume. This way you can also add the BC routine nicely into integration_structure.cpp (case CHT_WALL: ...).  

Regards, Ole

Common/src/geometry_structure.cpp
economon: I see.. yes, you could pass in the geometry object to achieve this.

My main concerns are 1) adding extra doubles to the CPoint class (remember that every node in the grid then has more data, when we only need this information on the boundaries) and 2) having PDE-related information in the CGeometry classes when it is more natural in the CSolver classes.

This is a really nice feature, and I have been thinking more about it. Is it possible to generalize it by putting the "CustomBC" routines up in the CSolver parent class, so that all of the child solver classes can inherit the ability to have custom BCs from Python? You would have to allocate some memory to store the values, ideally just for the boundary vertices, like the data structures at the end of the mean flow solver constructor. The data could be stored in generic "custom" structures and then the child classes could access them in the BC routines when necessary for solver-specific tasks (for CHT, custom profiles, etc.). That way, you really only need to implement it one time. What do you think?

tobadavid:
Some updates:

- Python wrapper functionalities have been moved into a dedicated python_wrapper_structure.cpp file for code clarity.
- The data structures for the customized values at boundaries have been removed from the CPoint class. Data structures are stored in the geometry class, they are initialized only when customization is set ON and only for involved vertices. There was no restriction to put the data structures in the solver class, except that when multi-grids needs to be updated (required for non uniform field), the geometry class needed an access to the solver class, which is not the case since everything that is in the Common structure seems to be independent of what it is in SU2_CFD, and I wanted to keep that. Those data structures should not be seen as true PDE-related variables, they should be seen as a generic support for boundary customization.
- CHT related labels have been removed and replaced by "Python customizable" in order to not introduce any confusion with the native CHT development. The basic purpose of this work is to enable existing boudary conditions to be customized from Python and not to develop a CHT interface. CHT is just an example of application (since only the HEAT_FLUX and ISOTHERMAL walls are customizable for now).
- Based on the previous point, no new specific boundary condition was introduced. For now, I keep the idea to just modify (one single line) the existing boundary conditions to take into account the customization. New dedicated boundary conditions (for example BC_HEAT_FLUX_CUSTOM or BC_ISOTHERMAL_CUSTOM) would be a 99% copy of the originals... Dedicated boundary conditions for native CHT development can still be defined independently (no condlfict).

Cheers,

David

talbring:
It looks like the regression tests fail because there is an additional `------------------------------ Begin Solver -----------------------------
` at the end of the output of the python wrapper test cases.

tobadavid:
This is not an _additional_ one... Instead of being printed at the beginning of the computation as it should be, it is printed at the true end and it does not make any sense to me... Right now I just can't understand why but it seems to be a Python 3 trick again.

talbring:
Maybe a flush is required (`sys.stdout.flush()`) ? But I agree that is strange ...

tobadavid:
Yep, my first attempt to fix this will use sys.stdout.flush()... Let's see. If it does not work, I will try other options.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -16410,6 +16410,9 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
     /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
     
     if (geometry->node[iPoint]->GetDomain()) {
+
+      /* If it is a customizable patch, retrieve the specified wall heat flux. */
+      if (config->GetMarker_All_PyCustom(val_marker)) Wall_HeatFlux = geometry->GetCustomBoundaryHeatFlux(val_marker, iVertex);
```
economon: It would be great to get these features into the incompressible solver too, but we can do that in a future PR (some changes are coming there too).

SU2_PY/FSI/FSIInterface.py
```diff
@@ -1544,26 +1553,26 @@ def setFluidInterfaceVarCoord(self, FluidSolver):
...
               DispX, DispY, DispZ = self.haloNodesDisplacements[GlobalIndex]
-              if posY0 == 0.0:
-                posX = posX0
-                posY = posY0
-                posZ = posZ0
-              else:
-                posX = posX0 + DispX
-                posY = posY0 + DispY
-                posZ = posZ0 + DispZ
+              #if posY0 == 0.0:
```
economon: Are these commented lines still needed?

SU2_PY/FSI/FSIInterface.py
```diff
@@ -1544,26 +1553,26 @@ def setFluidInterfaceVarCoord(self, FluidSolver):
...
             else:
-              if self.localFluidInterface_array_Y_init[localIndex] == 0.0:				# !!! This is temporary and case dependent, it should be removed ASAP !!!
-                posX = self.localFluidInterface_array_X_init[localIndex]
-                posY = self.localFluidInterface_array_Y_init[localIndex]
-                posZ = self.localFluidInterface_array_Z_init[localIndex]
-              else:
-                posX = self.localFluidInterface_array_DispX[localIndex] + self.localFluidInterface_array_X_init[localIndex]
-                posY = self.localFluidInterface_array_DispY[localIndex] + self.localFluidInterface_array_Y_init[localIndex]
-                posZ = self.localFluidInterface_array_DispZ[localIndex] + self.localFluidInterface_array_Z_init[localIndex]
+              #if self.localFluidInterface_array_Y_init[localIndex] == 0.0:				# !!! This is temporary and case dependent, it should be removed ASAP !!!
```
economon: And these?

SU2_CFD/src/solver_direct_mean.cpp
tobadavid: Sure ! In the future, the features should be extended to other solvers like the incompressible and also be extended to other kinds of BCs (typically inlet velocity profile).

SU2_PY/FSI/FSIInterface.py
tobadavid: I did some clean up here. Originally these old FSI scripts should have been removed, but then I realized that some people actually wanted to use it or were already using it like this. So, at the end, I just keep them until the coupling with our environment can be officially achieved.

SU2_CFD/src/solver_direct_mean.cpp
clarkpede: Any updates on this? (Specifically, custom BCs for Euler/Navier-Stokes solvers). A python-based implementation would be much simpler than the C++ based implementation I was creating.

If there aren't any updates, I could work on the implementation myself.  It looks fairly straightforward.

tobadavid: There is no update since the last merge. I won't be able to work on it intensively before a long time (projects and a thesis to finish...) so if you are interested you could definitely work on it. As you said, it should be straightforward, at least up to a certain point.

PR #490, opened 2017-12-26
Feature ddes

EduardoMolina:
Dear SU2 developers,

This PR is a long effort to add Hybrid RANS/LES capabilities, specially the Delayed Detached Eddy Simulation with Grey area mitigation and low dissipation schemes proper to unsteady flows.

Summary of the contribution:

- Delayed DES:
  - Standard SGS
  - Vorticity SGS
  - Shear Layer Adapted SGS

- Several low dissipation functions:
  - DDES FD
  - Ducros' shock sensor
  - NTS adaptive

- Variants of SA turbulence model:
  - Compressibility correction
  - Edwards and Chandra
  - QCR 2000 (Quadratic constitutive relation)

- Low dissipation schemes:
  - Low dissipation Roe.
  - Low Mach Roe.
  - Simple low dissipation AUSM (SLAU and SLAU2)

I added only a small regression test for 2D DDES, travis can not handle even the smallest 3D test case that I have. Suggestions are welcome.

Best,

Eduardo


SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -2407,23 +3385,36 @@ void CUpwRoe_Flow::ComputeResidual(su2double *val_residual, su2double **val_Jaco
...
   Energy_j = Enthalpy_j - Pressure_j/Density_j;
   SoundSpeed_j = sqrt(fabs(Pressure_j*Gamma/Density_j));
+  Mach_j = sqrt(Mach_j)/SoundSpeed_j;
+  
+  /*--- Modification according to Thornber et al. 2016-01-22 ----*/
+  /*--- An improved reconstruction method for compressible flows with low Mach number features. JCP V227 I10 2008 ---*/
+  /*--- L2Roe: a low dissipation version of Roe's approximate Riemann solver for low Mach numbers. IJNMF 2015 ---*/
+  
+  zeta = min(1.0,max(Mach_i,Mach_j));
+  //zeta = max(zeta,0.05);
```
economon: Remove this comment if unnecessary.

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -2614,194 +3654,190 @@ CUpwGeneralRoe_Flow::~CUpwGeneralRoe_Flow(void) {
...
-  Kappa_j = S_j[1]/Density_j;
-  Chi_j = S_j[0] - Kappa_j*StaticEnergy_j;
-  SoundSpeed_j = sqrt(Chi_j + StaticEnthalpy_j*Kappa_j);
+  SoundSpeed_j = sqrt(fabs(Pressure_j*Gamma/Density_j));
+  Mach_j = sqrt(Mach_j)/SoundSpeed_j;
+  
+  /*--- Rieper, A low-Mach number fix for Roe's approximate Riemman Solver, JCP 2011 ---*/
+  
+  zeta = min(1.0,max(Mach_i,Mach_j));
+  zeta = max(0.05,zeta);
```
economon: Remove the previous line if unnecessary.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -479,6 +479,480 @@ void CSourcePieceWise_TurbSA::ComputeResidual(su2double *val_residual, su2double
...
+        for(jDim=0;jDim<nDim;++jDim){
+            Sbar+= (PrimVar_Grad_i[1+iDim][jDim]+PrimVar_Grad_i[1+jDim][iDim])*(PrimVar_Grad_i[1+iDim][jDim]);}}
+    for(iDim=0;iDim<nDim;++iDim){
+        Sbar-= ((2.0/3.0)*pow(PrimVar_Grad_i[1+iDim][iDim],2.0));}
+    
+    Omega= sqrt(max(Sbar,0.0));
+    
+    /*--- Evaluate Omega ---*/
+    
+    //Omega = sqrt(Vorticity_i[0]*Vorticity_i[0] + Vorticity_i[1]*Vorticity_i[1] + Vorticity_i[2]*Vorticity_i[2]);
```
economon: Remove commented line if unnecessary.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -479,6 +479,480 @@ void CSourcePieceWise_TurbSA::ComputeResidual(su2double *val_residual, su2double
...
+        //Shat = S + TurbVar_i[0]*fv2*inv_k2_d2;
+        Shat = max(S*((1.0/max(Ji,1.0e-16))+fv1),1.0e-16);
+        
+        Shat = max(Shat, 1.0e-10);
+        inv_Shat = 1.0/Shat;
+        
+        /*--- Production term ---*/;
+        
+        //    Original SA model
+        //    Production = cb1*(1.0-ft2)*Shat*TurbVar_i[0]*Volume;
```
economon: In general, please remove the commented lines from the original S-A model to keep the implementation as clean as possible (especially for new users).

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -6761,7 +6874,140 @@ void CEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config) {
...
+        Coord_i    = geometry->node[iPoint]->GetCoord();
+        Coord_j    = geometry->node[jPoint]->GetCoord();
+        
+        
+        AD::StartPreacc();
+        AD::SetPreaccIn(Gradient_i, nPrimVarGrad, nDim);
+        AD::SetPreaccIn(Gradient_j, nPrimVarGrad, nDim);
+        AD::SetPreaccIn(Coord_i, nDim); AD::SetPreaccIn(Coord_j, nDim);
+        
+        /*--- Simple and Parameter-Free Second Slope Limiter for Unstructured Grid Aerodynamic Simulations---*/
```
economon: Having a parameter-free option is interesting.. folks still have trouble tuning limiters and avoiding limit cycles/stalled convergence. Have you had a positive experience with this limiter in terms of convergence?

EduardoMolina: There is a trick with this parameter-free second limiter. If you have a high quality 3D grid, this limiter will greatly improve convergence. However for complex geometries (most of the time in aerospace industry), this second limiter will diverge the solution because it will turn off the primary limiter in regions of poor quality cells when M<1.0.
There is a measure of the cell quality (Angle, Skewness, etc) in SU2? If we have,  I would like to use  here in order to avoid turning off the limiter in this regions.

talbring:
Hm why are the regression tests failing ? It looks like is has something to do with the changes in the limiter in a recent commit.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -700,16 +700,23 @@ class CSolver {
    * \param[in] solver_container - Container vector with all the solutions.
    * \param[in] config - Definition of the particular problem.
    */
-  virtual void SetDissipation_Switch(CGeometry *geometry, CConfig *config);
+  virtual void SetPressure_Sensor(CGeometry *geometry, CConfig *config);
```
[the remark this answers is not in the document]
talbring: I changed it to be consistent with other sensor values (e.g. the Ducros sensor), I didn't know that something else than the pressure is used in some schemes :) We can very well undo that renaming.

talbring:
Ok I think I found the problem, you accidentally removed the Set_MPI_Limiter call. Let's see what travis says.

SU2_CFD/src/solver_direct_mean.cpp
talbring: I see some problems in the current way it is implemented. First of all doing a search involving neighbors shouldn't be done inside an edge loop as you compute the same values twice (just an efficiency thing). Furthermore, in order for the computation to give the same result using different number of processors (which we should aim for) a communication of the variable Mach_Local is necessary.

Unless you think that this limiter greatly improves convergence, I would suggest to wait with this limiter implementation until we have figured out a way of generalizing this.

SU2_CFD/include/solver_structure.hpp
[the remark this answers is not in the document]
talbring: Now I can remember why I did it that way. The Ducros sensor (for Roe) and the Pressure sensor (for JST) are mutually exclusive. So I wanted to avoid having an extra variable for both of them so I store them both in the `Sensor` member variable (that also saves an extra MPI routine). So we could either name them SetDissipation_Sensor/SetDucros_Sensor or SetDissipation_Switch/SetDucros_Switch.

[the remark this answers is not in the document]
economon: Ah, I see. Then, the main difference is centered vs upwind for the dissipation sensors. I am fine with either proposal, or we could even add a "centered" or "upwind" mention if you'd like.

PR #491, opened 2018-01-05
Fix bug in Jacobian for SST Source Term

### LINKED ISSUE #369: SST destruction term jacobian
SST Jacobian for the destruction term in the k-equation w.r.t. omega set to zero.  Should be:
val_Jacobian_i[0][1] = -beta_star*TurbVar_i[0]*Volume;

clarkpede:
This fixes #369.  An extra term is added to the Jacobian for the source term, as suggested by @shaering. Repeating his comment:

The error is in the Jacobian of the destruction term in the k-equation for SST (Beta^* * rho * k * omega). The off-diagonal d/d(rho*omega) term is non-zero. It looks like it was just an omission.


clarkpede:
The change in the numerics of the Jacobian calculation changed the results of the regression tests.   That was expected, since improvements to the linear solve will change the residuals at each time step. If you look at the values from the tests, you'll see something like:

```
ERROR: Difference between computed input and test_vals exceeded tolerance. TOL=0.000001
test_iter=50
test_vals (stored): -6.459421, -4.595755, 1.201830, -0.007146, 0.080516
sim_vals (computed): -6.459444, -4.595710, 1.201844, -0.007146, 0.080517
```

The values using the new numerics are very close, but not within tolerance.

economon:
Thanks for looking into this.. it could be an important change. Regression changes are expected, of course, but does the convergence of the code improve with this change, in your experience? 

Since it is an off-diagonal term, it may be hurting diagonal dominance and thus possibly convergence of the linear system. The final solution does not depend on the Jacobian approximation at convergence, but if this noticeably helps overall convergence rate, we should make the change.

clarkpede:
So I tested the convergence on a couple of test cases, to see its behavior.  I used the following three different preconditioner/linear solver combinations:

+ LU SGS with FGMRES
+ ILU with FGMRES
+ ILU with BCGSTAB

I looked at three different test cases from the repo: the zero-pressure gradient flat-plate turbulent boundary layer, the NACA 0012 airfoil, and the RAE 2822. In each case, I ran without a restart file. For the flat plate and the NACA 0012, I used adaptive CFL and no multigrid.  For the RAE 2822, I used constant CFL and multigrid.

I looked at both the number of iterations of the linear solver (the inner iterations) and the overall convergence over time (the outer iterations). Here's some sample results and discussion:

For the flat-plate boundary layer, there was no noticeable difference between the develop branch and the fix:
flat_plate_ilu

For the NACA 0012, the linear solver converged in fewer iterations with the fix from this pull request. This was seen for all combinations of linear solver/preconditioner tested.  The residual in turbulent kinetic energy did lag in the middle, but both cases converged in a similar number of outer iterations.
naca0012_lu_sgs

For the RAE 2822, the overall convergence was identical (same number of outer iterations). The linear solvers also converged in a similar number of iterations. Occasionally, the two versions would differ by an iteration, but this depended on the case:

+ LU SGS with FGMRES: Current develop branch converged in fewer linear solver iterations
+ ILU with FGMRES: Fix branch converged in fewer linear solver iterations
+ ILU with BCGSTAB: Convergence was almost identical

rae2822_lu_sgs

### Conclusions
__There doesn't seem to be any large difference in the convergence with or without the fix__.  Occasionally this fixed branch will outperform the existing develop branch (seen in the NACA 0012).  Aside from that, the differences are usually only one or two iterations of the linear solver (seen in the RAE 2822).

economon:
Thanks for the detailed analysis... out of curiosity, how tightly did you converge the linear solve with each nonlinear iteration? 

Lately, we have seen that the ILU preconditioner performs best, and, when solving the linear systems to a smaller tolerance, one can often use a higher CFL number. It could be interesting to look at the convergence of the linear solve for a single nonlinear iteration (at different points of the calculation) for GMRES+ILU to a near machine precision tolerance both with and without the term in the Jacobian. That might best show the impact of this term.

All of that being said, I think we can add this term, as performance is the same or better, and it is technically correct.

clarkpede:
So I realized I made a mistake with my previous post.  The plots previously showed the iterations of the linear solver for the Navier-Stokes equations.  I apologize for the mistake. I updated them with the iterations of the linear solver for the turbulence solver.  The results aren't very different.

clarkpede:
My previous post used a linear solver precision of 1E-6.

I ran a couple of test cases at near machine precision (a linear solver tolerance of 1E-14) and higher CFL number (CFL=150), with no CFL adaptation. As suggested, I used ILU+FGMRES. I capped the maximum number of linear solver iterations at 200, and did not raise any errors if the linear solver stopped at 200 iterations with less than 1E-14. The linear solver didn't exceed 200 iterations for the RANS equations, but it did for the Navier-Stokes equations. Once again, there doesn't appear to be a huge difference.

For the flat-plate, the results were almost identical:

flat_plate_improved

For the NACA 0012 airfoil, the current `develop` branch of the code outperformed this pull request:

naca0012_improved

I believe overall convergence was similar because the Navier-Stokes solver, not the turbulence-solver, is the limiting factor when it comes to precision. The linear solver always converged on a turbulence solution with high precision in under 50 linear solver iterations. Meanwhile, the linear solver sometimes failed to find a Navier-Stokes solution within 1E-14 even after 200 iterations. The only way to make the Navier-Stokes solver converge to 1E-14 was to use small time steps (CFL ~= 1) at difficult points in the overall convergence.

talbring:
@economon that is indeed an interesting point. Since long I have the plan to use AD to set up the exact Jacobian. This way we can what we can expect in the best case. Unfortunately I haven't had the time to that yet ...

PR #492, opened 2018-01-05
Fixed possible infinite loop in mesh adaptation

### LINKED ISSUE #487: Infinite Loop When Running Grid Adaptivity on Hexahderons
I've encountered a bug where SU2 will get stuck in an infinite loop.  I've been trying out the grid adaptation code recently. Tets, rectangles, and triangles all worked fine, but hexahedra threw the code into an infinite loop.  Basically, the code is missing the proper code for hexahedra in the `CGridAdaptation::SetSensorElem` function.

I've attached the [offending lines 3640-3653](https://github.com/su2code/SU2/blob/22fa337afd26a44234344483142deff0f9524ae0/Common/src/grid_adaptation_structure.cpp#L3640) of `Common/src/grid_adaptation_structure.cpp` below:

	/*--- Selection of the elements to be adapted ---*/
	threshold = 0.999;
	nElem_real = 0;
	while (nElem_real <= max_elem) {
		for (iElem = 0; iElem < geometry->GetnElem(); iElem ++)
			if ( Sensor[iElem] >= threshold && !geometry->elem[iElem]->GetDivide() ) {
				if (geometry->elem[iElem]->GetVTK_Type() == TRIANGLE) nElem_real = nElem_real + 3;
				if (geometry->elem[iElem]->GetVTK_Type() == QUADRILATERAL) nElem_real = nElem_real + 3;
				if (geometry->elem[iElem]->GetVTK_Type() == TETRAHEDRON) nElem_real = nElem_real + 7;
				geometry->elem[iElem]->SetDivide(true);
				if (nElem_real >= max_elem) break;
			}	
		threshold = threshold - 0.001;
	}

As you can see, if all of the elements are hexahedral, the threshold will keep decreasing but `nElem_real` will never increase, and the loop will never end.

This is an easy fix, and I've already fixed it on a local branch.  But I'm also wondering why grid adaptation doesn't work for hexahedra.  It's not just this section of code that doesn't work.  There's other sections too.  One of the biggest ones is at [lines 2369-2386](https://github.com/su2code/SU2/blob/22fa337afd26a44234344483142deff0f9524ae0/Common/src/grid_adaptation_structure.cpp#L2369) of `Common/src/grid_adaptation_structure.cpp`.  In that section, all hexahedra are explicitly _removed_ from the grid adaptivity process.

This is rather confusing, because there is a **lot** of code in the CGridAdaptation class that is set up for adaptation of hexahedra.  The [AIAA paper from 2013](https://arc.aiaa.org/doi/abs/10.2514/6.2013-287) even dedicates a whole paragraph to explaining the hexahedral grid adaptation.  I understand that a pure hex grid can be a poor case for adaptive grid refinement, but I would like to try the grid adaptivity on a hybrid mesh (with hexes and tets).

Is there some reason that grid adaptivity for hexahedra is "turned off"?

clarkpede:
As noted in issue #487, the mesh adaptation routine can fall into an infinite loop if there aren't enough cells to adapt.  For example, if you run the mesh adaptation on a grid with pure hexahedra, the program will look for a certain number of cells to adapt but find none.  I added an additional break where the adaptivity loop will break if all suitable cells have been marked for adaptation.

I also added an additional warning that will pop up if mesh adaptation is run without specifying a NEW_ELEMS percentage.

Edit: This fixes #487.

### DIFF
--- a/Common/src/grid_adaptation_structure.cpp
+++ b/Common/src/grid_adaptation_structure.cpp
@@ -3594,6 +3594,13 @@ void CGridAdaptation::SetSensorElem(CGeometry *geometry, CConfig *config, unsign
 	su2double Max_Sensor, threshold;
 	su2double *Sensor = new su2double[geometry->GetnElem()];
 	unsigned long ip_0, ip_1, ip_2, ip_3, iElem, nElem_real;
+  
+	if (max_elem > geometry->GetnElem()) {
+		cout << "WARNING: Attempted to adapt " << max_elem << " cells," << endl;
+		cout << "  which is greater than the total number of cells, ";
+		cout << geometry->GetnElem() << "." << endl;
+		cout << "  Did you set the option NEW_ELEMS in the *.cfg file?" << endl;
+  }
 	
 	/*--- Compute the the adaptation index at each element ---*/
 	Max_Sensor = 0.0;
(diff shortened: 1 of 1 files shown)


clarkpede:
@economon Dr. Alonso emailed me saying that "We used [grid adaptation] successfully in tetrahedral meshes, but I do not believe the hexahedral adaptation was ever fully functional." That matches my experience with a few tests I ran.

economon:
Thanks, @clarkpede, merging now. We can keep the extended convo going with issue #487.

PR #493, opened 2018-01-08
Fixed bug in ADT

vdweide:
This PR fixes a bug in the ADT that was introduced to circumvent the Allgatherv functionality. Allgatherv is now available in SU2_MPI, hence the original code has been restored. Due to this correction the convergence values of the 2D axial stage regression test had to be altered.


talbring:
Perfect, thank you Edwin. As soon as Travis succeeds we will merge it in.

PR #494, opened 2018-01-10
Feature multi obj update

hlkline:
Fix to multi-functional optimization, addition of regression test. 
This re-introduces functionality that was previously added in https://github.com/su2code/SU2/pull/385. 

(it was accidentally effected by someone else's pull request that appears to have been based on an outdated version of develop - if work you merged ~Feb-July has mysteriously disappeared I can privately point you to the relevant pull request to aid in fixing that)


hlkline:
Why don't I add additional regression cases of multiple objective on a single surface and a single objective on multiple surfaces, I think otherwise the situations you mention are covered (both of those should already work, but may not be covered by regressions, and they can easily be covered by single-iteration tests). 
There is a tutorial page here: 
https://github.com/su2code/SU2/wiki/Multi-Objective-Optimization
Although that's from a while ago I don't think much needs to be changed, except for broken links including images.

economon:
Ah, that tutorial page is very important.. I don't think it was ever linked in the main wiki menu, but it is critical for making the options clear to the user (especially the surface handling). We need to make the design features as easy to use as possible - it is already hard enough for experts to use adjoints :).

This is a good opportunity to move it over to the new repo where we are placing the tutorials here: https://su2code.github.io/Tutorials/docs/home/. The files, including the markdown, mesh, config, and images, can be moved to that repo. The markdown portion goes into the _docs directory, and we should make sure it's linked under the shape design tutorials. Do you have some time for this? Let me know if you need some help.

hlkline:
In theory I can update the tutorial, but most likely I won't be able to get to completing that for a couple of weeks. If you have the time to work on it, that would be based on the files in optimization_euler/multiobjective_wedge/; since that's a separate that would be in a different pull request. 
There may be an issue with the single objective/multi-surface case: I'm going to comment that out from the regression script but leave the files in to make that easy to add back in in the future - it looks like someone in another pull request was working on similar things (https://github.com/su2code/SU2/pull/497) so that should probably wait until their changes are merged in as well. Otherwise this is ready.

hlkline:
The single objective/multi surface case was added back in along with a regression test by these previous commits; I do still think it is reasonable to wait until after #497 is merged in so that I can be sure to pass their tests as well, and I will be working on updating the tutorial separately. 
In short, this can be reviewed at any time, and reviewers should expect an additional update if #497 is merged in first.

hlkline:
closing temporarily for finishing tutorial/anticipating merging with the previously mentioned pull request/bug fix/generally avoiding confusion. Will reopen when those are complete.

hlkline:
Reopening. 
See tutorial here: https://github.com/su2code/Tutorials/pull/2

hlkline:
Thanks - although I think maybe the recent compiler warning request should be considered as well.

PR #495, opened 2018-01-10
Update codi/medi

talbring:
This is just a small update of the codi/medi submodules to the newest versions.


fpalacios:
LGTM Thanks for the improvement!

PR #497, opened 2018-01-12
Small changes

VivaanKhatri:
Collection of small improvements:
- Discrete Adjoint as default.
- New options in the config file: FROZEN_LIMITER_DISC, FROZEN_VISC_DISC, INCONSISTENT_DISC
- Fixing some issues with the multi objective design.
- Updating Xcode project


Common/src/linear_solvers_structure.cpp
```diff
@@ -87,9 +87,12 @@ void CSysSolve::SolveReduced(const int & n, const vector<vector<su2double> > & H
 void CSysSolve::ModGramSchmidt(int i, vector<vector<su2double> > & Hsbg, vector<CSysVector> & w) {
   
   bool Convergence = true;
+  
+#ifdef HAVE_MPI
```
economon: Could you please remove the ifdef's? These should be unnecessary now that the wrapper takes care of it internally. Or was there a problem?

SU2_CFD/src/output_structure.cpp
```diff
@@ -3998,6 +3998,11 @@ void COutput::SetRestart(CConfig *config, CGeometry *geometry, CSolver **solver,
   bool dual_time = ((config->GetUnsteady_Simulation() == DT_STEPPING_1ST) ||
                     (config->GetUnsteady_Simulation() == DT_STEPPING_2ND));
 
+  int rank = MASTER_NODE;
```
economon: same here

SU2_CFD/src/output_structure.cpp
```diff
@@ -11073,6 +11079,14 @@ void COutput::SetResult_Files_Parallel(CSolver ****solver_container,
                                        unsigned long iExtIter,
                                        unsigned short val_nZone) {
   
+  int rank = MASTER_NODE;
+  int size = SINGLE_NODE;
+  
+#ifdef HAVE_MPI
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
economon: Same here.. you can use SU2_MPI::GetRank() and SU2_MPI::GetSize()

SU2_IDE/Xcode/SU2_CFD.xcodeproj/project.pbxproj
```diff
@@ -69,6 +69,7 @@
 		05F1089C1978D2AE00F2F288 /* fluid_model_pvdw.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 05F108981978D2AE00F2F288 /* fluid_model_pvdw.cpp */; };
 		05F1089D1978D2AE00F2F288 /* fluid_model.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 05F108991978D2AE00F2F288 /* fluid_model.cpp */; };
 		05F108A41978D2F200F2F288 /* transport_model.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 05F108A31978D2F200F2F288 /* transport_model.cpp */; };
+		05F8F2682008A1C8000FEA01 /* python_wrapper_structure.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 05F8F2672008A1C7000FEA01 /* python_wrapper_structure.cpp */; };
```
economon: Were these updates to the Xcode project intended (looks like it)?

SU2_CFD/src/output_structure.cpp
talbring: You can also remove the declaration of the 'rank' variable as it is now also a member variable of most of the classes (including COutput).

config_template.cfg
```diff
@@ -1143,19 +1148,19 @@ MULTIPOINT_WEIGHT= (0.33333, 0.33333, 0.33333)
 % Optimization objective function with scaling factor, separated by semicolons.
 % To include quadratic penalty function: use OPT_CONSTRAINT option syntax within the OPT_OBJECTIVE list.
 % ex= Objective * Scale
-OPT_OBJECTIVE= DRAG * 0.001
+OPT_OBJECTIVE= DRAG
```
talbring: Why is that removed ?

SU2_PY/shape_optimization.py
```diff
@@ -49,7 +49,7 @@ def main():
                       help="try to restart from project file NAME", metavar="NAME")
     parser.add_option("-n", "--partitions", dest="partitions", default=1,
                       help="number of PARTITIONS", metavar="PARTITIONS")
-    parser.add_option("-g", "--gradient", dest="gradient", default="CONTINUOUS_ADJOINT",
+    parser.add_option("-g", "--gradient", dest="gradient", default="DISCRETE_ADJOINT",
```
talbring: I think changing default values is a big step. I would actually recommend to not change it for now.

config_template.cfg
```diff
@@ -779,6 +773,17 @@ THERMAL_DIFFUSIVITY= 1.0
...
+% Frozen the slope limiter in the discrete adjoint formulation (NO, YES)
+FROZEN_LIMITER_DISC= NO
+%
+% Frozen the turbulent viscosity in the discrete adjoint formulation (NO, YES)
+FROZEN_VISC_DISC= NO
+%
+% Use an inconsistent spatial integration (primal-dual) in the discrete
+% adjoint formulation. The AD will use the numerical methods in
+% the ADJOINT-FLOW NUMERICAL METHOD DEFINITION section (NO, YES)
+INCONSITENT_DISC= NO
```
talbring: I have never tried to use a different numerical method for the discrete adjoint. Will do that soon :+1:  Thanks for adding the other options to the config template.

BTW: There is a typo in the name of this option.

SU2_PY/shape_optimization.py
hlkline: I agree - unfortunately most users will not have compiled with AD and may find this confusing.

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -4331,16 +4331,27 @@ void CIncEulerSolver::SetFarfield_AoA(CGeometry *geometry, CSolver **solver_cont
...
 
-  unsigned short iMarker_Monitoring;
+  unsigned short iMarker_Monitoring, Kind_ObjFunc;
   su2double Weight_ObjFunc;
 
+  Total_ComboObj = 0.0;
+  
   /*--- Loop over all monitored markers, add to the 'combo' objective ---*/
 
   for (iMarker_Monitoring = 0; iMarker_Monitoring < config->GetnMarker_Monitoring(); iMarker_Monitoring++) {
```
hlkline: I have been working on updating related code (https://github.com/su2code/SU2/pull/494) - other mismatches between markers and objectives in the combined objective case are taken care of in config_structure.cpp ~ line 2268. I think that a fix for this case can be added there instead, both for consistency and to avoid repeated code (especially since this problem effects the continuous adjoint as well).
I don't mind adding that in to my current pull request; however just a note that that will make checking for the number of markers here redundant. (They won't conflict with each other, just be doing the same thing twice)

SU2_PY/shape_optimization.py
VivaanKhatri: Based on my experience DISCRETE_ADJOINT is the only one that really works with turbulent flows... and that is what most of the users are expecting. No reason for not compiling the software with AD... it should be the default.

config_template.cfg
VivaanKhatri: If you use OPT_RELAX_FACTOR and OPT_GRADIENT_FACTOR instead of the objective function scale factor you obtain much better results.

SU2_CFD/src/solver_direct_mean_inc.cpp
VivaanKhatri: We need a regression test with several surfaces to be sure that everything is working fine. You can remove this line in your pull request after checking that multiple surfaces are working just fine.

hlkline: I agree, could you please add one that will fail if the functionality addressed by this section of code is removed? Based on other discussion on this pull request, you may be able to hit several birds with one stone by also making that an optimization case using a turbulent flow and the discrete adjoint, and which uses both the OPT_GRADIENT_FACTOR and OPT_RELAX_FACTOR. Trust me, it's better to go through the time to add a regression test early rather than try to figure out what happened to your features months or years down the line.

SU2_PY/shape_optimization.py
talbring: I am the last person who has something against this kind of change :) Still, in order for that to be default I should perhaps update the preconfigure script to work more reliably (especially when SU2 is not cloned using github ...)

SU2_IDE/Xcode/SU2_CFD.xcodeproj/project.pbxproj
VivaanKhatri: Yep. I have updated the Xcode files

SU2_CFD/src/solver_direct_mean_inc.cpp
VivaanKhatri: Done!

config_template.cfg
fpalacios: OPT_RELAX_FACTOR and OPT_GRADIENT_FACTOR are the correct variables to use in optimization problems for the new release

SU2_PY/shape_optimization.py
fpalacios: Discrete adjoint is going to be the default for the tutorials. Continuous adjoint is working great for inviscid problems but... from the user point of view you are looking for something that works in any circumstance. The community should start using AD by default and we need to push that. In the tutorial we will explicitly indicate DISCRETE or CONTINUOUS

PR #500, opened 2018-01-17
Overhauled node sorting in periodic geometry to be less buggy

### LINKED ISSUE #431: Segfault when running problem with periodic boundary conditions
When I run a mixing-layer problem with periodic boundary conditions, I get a segfault while saving the file. I checked out the most recent develop branch, and I get the same problem.

It may be that I messed up the periodic boundary conditions.  I'm still learning how to use them in SU2, and I may have made a mistake.  But even if my *.cfg files are set up wrong, I should get a runtime error message, not a segfault.

I did a brief check and it looks like `SU2_MSH` is doing its job in generating the correct periodic BCs.

Valgrind says that there's a lot of invalid reads in `COutput::SortOutputData`, which terminate in a segfault. I'm attaching the valgrind output, in case you want to look at it.

#### Steps to reproduce
1. Build SU2 with parallel support.
1. Put the attached grid, `mixing_MSH.cfg` and `mixing_CFD.cfg` files into the  same folder.
2. Run `SU2_MSH mixing_MSH.cfg`.
3. Run `SU2_MSH mixing_CFD.cfg`.

I've also attached my config output, so you can see what my setup is like.

Here's the files: 
[PeriodicSegfault.tar.gz](https://github.com/su2code/SU2/files/1257883/PeriodicSegfault.tar.gz)

clarkpede:
As described in Issue #431, using `SU2_MSH` for periodic mesh construction can be buggy due to its sorting of nodes.  Under certain conditions, nodes can be duplicated and/or missing from the output *.su2 file.

This pull request implements a more robust sorting process, where the sorting occurs in the following steps:

1. Send/receive nodes are explicitly identified
2. The sorted list is created without any send/receive nodes.
3. The receive and then send nodes are added to the sorted list.

Here's a few comments about the pull request:
+ Since this routine is only run once, I felt the usability of std::vector outweighed the computational efficiency of raw arrays.
+ I had to create two lists, NewSort and ReverseSort, instead of the original one list (NewSort).  That's because two different mappings between node numbers need to occur.  In the old code, where the list was sorted by switching elements, the two mappings _should_ have been identical.  Since the improved method does not use switching, two lists are necessary.
+ There's a lot of memory leaks/uninitialized values when checking this pull request with Valgrind.  I checked them, and this pull request doesn't create any _new_ memory issues.  Apparently both `SU2_CFD` and `SU2_MSH` have memory issues.

#### Verification
Since none of the existing regression tests run `SU2_MSH`, I created a separate case to verify both the bug and the fix.  It's a simple 3x3 cube that's periodic in the z-direction.  I've attached the files below. Here's the steps:

1. Run `SU2_MSH MSH.cfg`
2. Run `SU2_CFD per_CFD_dev.cfg`

Using the existing develop branch, the verification case runs into a segfault when writing the output.  This pull request allows the verification case to complete successfully.

You can also check the original test case from Issue #431 to verify the bug fix. If any of you know of additional verification tests I could run, then please let me know.

Test case: [cube.tar.gz](https://github.com/su2code/SU2/files/1639835/cube.tar.gz)

This fixes #431

### DIFF
--- a/Common/src/geometry_structure.cpp
+++ b/Common/src/geometry_structure.cpp
@@ -18794,30 +18794,55 @@ void CPeriodicGeometry::SetMeshFile(CGeometry *geometry, CConfig *config, string
   /*--- Ghost points, look at the nodes in the send receive ---*/
   iMarkerReceive = nMarker - 1;
   GhostPoints = nElem_Bound[iMarkerReceive];
-  
+
   /*--- Change the numbering to guarantee that the all the receive
-   points are at the end of the file ---*/
-  unsigned long OldnPoint = geometry->GetnPoint();
-  unsigned long *NewSort = new unsigned long[nPoint];
-  for (iPoint = 0; iPoint < nPoint; iPoint++) {
-    NewSort[iPoint] = iPoint;
-  }
-  
-  unsigned long Index = OldnPoint-1;
+   points are at the end of the file. ---*/
+  std::vector<unsigned long> receive_nodes;
+  std::vector<unsigned long> send_nodes;
   for (iMarker = 0; iMarker < nMarker; iMarker++) {
     if (bound[iMarker][0]->GetVTK_Type() == VERTEX) {
       if (config->GetMarker_All_SendRecv(iMarker) < 0) {
         for (iElem_Bound = 0; iElem_Bound < nElem_Bound[iMarker]; iElem_Bound++) {
-          if (bound[iMarker][iElem_Bound]->GetNode(0) < geometry->GetnPoint()) {
-            NewSort[bound[iMarker][iElem_Bound]->GetNode(0)] = Index;
-            NewSort[Index] = bound[iMarker][iElem_Bound]->GetNode(0);
-            Index--;
+          if (bound[iMarker][iElem_Bound]->GetRotation_Type() == 1) {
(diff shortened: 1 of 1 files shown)


Common/src/geometry_structure.cpp
```diff
@@ -18794,45 +18794,72 @@ void CPeriodicGeometry::SetMeshFile(CGeometry *geometry, CConfig *config, string
...
+    if (!isSend && !isReceive) {
+      NewSort.push_back(iPoint);
+    }
+  }
+  NewSort.insert(NewSort.end(), receive_nodes.begin(), receive_nodes.end());
+  NewSort.insert(NewSort.end(), send_nodes.begin(), send_nodes.end());
   
+  ReverseSort.resize(NewSort.size());
+  for (iPoint = 0; iPoint < nPoint; iPoint++) {
+    unsigned short jPoint;
```
LaSerpe: This is defined as short unsigned.
It causes problems when the mesh has a higher number of nodes, changing it to long solves the issue in my case.

LaSerpe:
Hi guys,
I found a small flaw in the new code lines (line 18835). the counter should be defined as a long unsigned I guess.

clarkpede:
@LaSerpe thanks for catching that! Yeah, that definitely seems like a problem.  I'll submit a patch.

PR #505, opened 2018-02-07
Native CHT functionality

oleburghardt:
This branch adds a native CHT functionality to SU2.

Main contents are:

1. A new driver (CMultiPhysicalZonesDriver) to specify the physics (e.g. which zones are fluids, which are solids) without a bunch of a priori or 'hard coded' assumptions. I intend to develop it further in near future which basically means to incorporate the existing sorts of interfaces (mixing plane, FSI, .. maybe it works out).
1. A new finite-volume-based heat solver (CHeatSolverFVM). It can be run stand-alone in solids for pure heat conduction or weakly coupled to the incompressible mean flow solver.
1. A new CTransfer child class (CTransfer_ConjugateHeatVars) to exchange CHT-related interface information.

Most of these commits do not touch existing code but just use it. Exceptions are:
1. Quite a few config options.
1. A small new data structure in CDriver to store types of interfaces that is used by CMultiPhysicalZonesDriver. Some additions in **Interface_Preprocessing**.
1. A change of **SetCFL_Number** so that the CFL adaption works for multiple zones. @salvovitale : One of your test cases is a multi-zones one that uses CFL adaption (the Jones one). This commit of course changes the solver behaviour in that case. I saw that you are averaging CFL numbers - maybe these problems are connected? Could you check if the test case is still alright? For now, I had to change the regression test values.
1. Some more else if(...)-cases in output_structure.cpp...

As this is my first pull request, it's very likely that I missed some good-practices. I would like to point out that many of the commits contain _purely copied code_.
There was always some reason when it could not re-use it, but please be a bit indulgent if I missed to reference the credits. I just don't know where to.. if something attracts your attention, I will change it asap :-) Furthermore I should mention that
1. This branch contains are very **naive couling procedure**. I will add Robin-like BC's and conservative energy correction routines when I feel comfortable about them.
1. The **discrete adjoint version** of all this works fine. I will add it, too, if you like these additions.


.travis.yml
```diff
@@ -16,7 +16,7 @@ notifications:
   
 branches:
     only:
-        - develop
+        - feature_cht
```
talbring: Revert that back develop.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -6952,3 +7113,313 @@ void CDiscAdjFSIStatDriver::Postprocess(unsigned short ZONE_FLOW,
...
+          || (config_container[iZone]->GetKind_Solver() == RANS))
+        checkConvergence += (int) integration_container[iZone][FLOW_SOL]->GetConvergence();
+      else if ((config_container[iZone]->GetKind_Solver() == DISC_ADJ_EULER)
+               || (config_container[iZone]->GetKind_Solver() == DISC_ADJ_NAVIER_STOKES)
+               || (config_container[iZone]->GetKind_Solver() == DISC_ADJ_RANS))
+        checkConvergence += (int) integration_container[iZone][ADJFLOW_SOL]->GetConvergence();
+      else if ((config_container[iZone]->GetKind_Solver() == HEAT_EQUATION)
+               || (config_container[iZone]->GetKind_Solver() == HEAT_EQUATION_FVM))
+        checkConvergence += (int) integration_container[iZone][HEAT_SOL]->GetConvergence();
+      else { }
```
talbring: This else statement can be removed I guess.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -17180,3 +17198,312 @@ void CNSSolver::SetRoe_Dissipation(CGeometry *geometry, CConfig *config){
...
+  string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
+
+  /*--- Retrieve the specified wall function treatment.---*/
+
+  Wall_Function = config->GetWallFunction_Treatment(Marker_Tag);
+  if(Wall_Function != NO_WALL_FUNCTION) {
+
+    cout << endl << "Wall function treament not implemented yet" << endl << endl;
+#ifndef HAVE_MPI
+    exit(EXIT_FAILURE);
```
talbring: This error message/exit should be adapted to the new format with SU2_MPI::Error.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -11119,6 +11199,284 @@ class CWaveSolver : public CSolver {
   
 };
 
+/*! \class CHeatSolverFVM
+ *  \brief Main class for defining the finite-volume heat solver.
+ *  \author O. Burghardt
+ *  \version 5.0.0 "Raven"
```
economon: Check that the versions are updated.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -11119,6 +11199,284 @@ class CWaveSolver : public CSolver {
...
+ *  \date January 19, 2018.
+ */
+class CHeatSolverFVM : public CSolver {
```
economon: Again, I am fine with this becoming the only heat solver (it is ok to remove old features that folks aren't using). Of course, it also sounds like we can also reuse some things between this and the new FVM Poisson solver written by @koodlyakshay (numerics routines come to mind, as one example). We can coordinate more on this.

SU2_CFD/include/transfer_structure.hpp
```diff
@@ -756,4 +756,58 @@ class CTransfer_SlidingInterface : public CTransfer {
 
 };
 
+/*!
+ * \class CTransfer_ConjugateHeatVars
+ * \brief Transfer temperature and heatflux density for conjugate heat interfaces between structure and fluid zones.
+ * \author O. Burghardt
+ * \version 4.0.1 "Cardinal"
+ */
+class CTransfer_ConjugateHeatVars : public CTransfer {
```
economon: Check version number. Nice, good to see more usage of the CTransfer classes :)!

SU2_CFD/include/variable_structure.hpp
```diff
@@ -2482,6 +2482,39 @@ class CWaveVariable : public CVariable {
   
 };
 
+/*!
+ * \class CHeatFVMVariable
+ * \brief Main class for defining the variables of the finite-volume heat equation solver.
+ * \author O. Burghardt
+ * \version 5.0.0 "Raven"
```
economon: Same, version number. Same comments apply to CVariable - we can eventually make these the only Heat classes.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -6952,3 +7113,313 @@ void CDiscAdjFSIStatDriver::Postprocess(unsigned short ZONE_FLOW,
...
+                       unsigned short val_nZone,
+                       unsigned short val_nDim,
+                       SU2_Comm MPICommunicator) : CDriver(confFile,
+                                                          val_nZone,
+                                                          val_nDim,
+                                                          MPICommunicator) { }
+
+CMultiPhysicalZonesDriver::~CMultiPhysicalZonesDriver(void) { }
+
+void CMultiPhysicalZonesDriver::Run() {
```
economon: I really like the idea of abstracting the multiphysics driver to the point that it can (in theory) include almost any set of physics in any zone with modular coupling at the interfaces. This is the right vision, I think, and we've been working in this direction. As you already know :) it will also make the construction of multiphysics adjoints with AD even more straightforward. Looking forward to the discrete adjoint of this when you have time to push it (and future multiphysics adjoints). Nice job here.

SU2_CFD/src/numerics_direct_heat.cpp
```diff
@@ -37,3 +37,318 @@
...
+}
+
+CSourceHeating::~CSourceHeating(void) { }
+
+void CSourceHeating::ComputeResidual(su2double *val_residual, CConfig *config) {
+  unsigned short iVar;
+
+  //su2double external_heat = config->GetExternalHeating();
+
+  //have to finish the implementation...
```
economon: Yep, we will want to clean this up so the user can specify volumetric heating with some access in the config file. Maybe we need to discuss a good design for this to make it easy for the user to specify the spatial bounds for volumetric heating. Any ideas?

SU2_CFD/src/solver_direct_heat.cpp
```diff
@@ -578,3 +578,1978 @@ void CHeatSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_
...
+
+  nVarGrad = nVar;
+
+  /*--- Define geometry constants in the solver structure ---*/
+
+  nDim = geometry->GetnDim();
+  node = new CVariable*[nPoint];
+  nMarker = config->GetnMarker_All();
+
+  CurrentMesh = iMesh;
```
economon: Is multigrid working well for the new FVM heat solver?

SU2_CFD/src/solver_direct_heat.cpp
```diff
@@ -578,3 +578,1978 @@ void CHeatSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_
...
+          for (iDim = 0; iDim < nDim; iDim++) Area += Normal[iDim]*Normal[iDim]; Area = sqrt(Area);
+
+          dist = 0.0;
+          for (iDim = 0; iDim < nDim; iDim++) dist += (Coord_Normal[iDim]-Coord[iDim])*(Coord_Normal[iDim]-Coord[iDim]);
+          dist = sqrt(dist);
+
+          dTdn = (Twall - node[iPointNormal]->GetSolution(0))/dist;
+
+          if(flow) {
+            // eddy viscosity should be added?
```
economon: On the wall, eddy viscosity should be zero. In the volume, it should be included to get the effective conductivity (unless this is set constant by the user).

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -17180,3 +17198,312 @@ void CNSSolver::SetRoe_Dissipation(CGeometry *geometry, CConfig *config){
...
+        LinSysRes.SetBlock_Zero(iPoint, iDim+1);
+      node[iPoint]->SetVel_ResTruncError_Zero();
+
+      /*--- Get transport coefficients ---*/
+
+      laminar_viscosity    = node[iPoint]->GetLaminarViscosity();
+      eddy_viscosity       = node[iPoint]->GetEddyViscosity();
+      thermal_conductivity = Cp * ( laminar_viscosity/Prandtl_Lam + eddy_viscosity/Prandtl_Turb);
+
+      // work in progress on real-gases...
```
economon: I have a fix for this type of issue in my incompressible branch.

Common/include/option_structure.hpp
```diff
@@ -220,6 +222,7 @@ static const map<string, ENUM_SOLVER> Solver_Map = CCreateMap<string, ENUM_SOLVE
 ("ADJ_NAVIER_STOKES", ADJ_NAVIER_STOKES)
 ("ADJ_RANS", ADJ_RANS )
 ("WAVE_EQUATION", WAVE_EQUATION)
+("HEAT_EQUATION_FVM", HEAT_EQUATION_FVM)
```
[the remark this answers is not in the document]
oleburghardt: Alright.. as I'm new to "real" contributing, I have no clear opinion on that yet - I would prefer to leave the decision up to you (and @talbring , if he likes?). 
My main concern is that a FEM solver may offer something a FVM cannot provide, regardless of me being not aware of an example. Still, i would be a pity once the FEM solver finds a new developer in charge.

Please just still me, I will adapt the naming.

SU2_CFD/src/solver_direct_heat.cpp
oleburghardt: It did in all in cases where it worked for the flow solver, too.
So I like to think that all failing cases were due to the general partitioning problem.

As a interim solution, only a CSingleGridIntegration is used which still performs well.
(But I saw that you just opened up a topic on that matter, so maybe I can make the multigrid usage up to a user preference soon!)

SU2_CFD/src/solver_direct_heat.cpp
oleburghardt: Agreed. Will remove these questions!

TestCases/parallel_regression.py
```diff
@@ -626,7 +626,7 @@ def main():
     Jones_tc.cfg_dir   = "turbomachinery/APU_turbocharger"
     Jones_tc.cfg_file  = "Jones.cfg"
     Jones_tc.test_iter = 5
-    Jones_tc.test_vals = [-5.352425, 0.417103, 82.840150, 0.882252] #last 4 columns
+    Jones_tc.test_vals = [-5.294493, 0.429399, 80.085030, 1.034808] #last 4 columns
```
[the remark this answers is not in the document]
oleburghardt: Could provide a collection.. ;-)
I will add a CHT-version of the prominent 2d-cylinder-in-laminar-flow-case to the repository.

SU2_CFD/src/transfer_physics.cpp
```diff
@@ -719,3 +719,133 @@ void CTransfer_SlidingInterface::SetTarget_Variable(CSolver *target_solution, CG
...
+    conductivity_over_dist  = thermal_conductivity/dist;
+  }
+  else {
+
+    thermal_diffusivity     = donor_config->GetThermalDiffusivity_Solid();
+    heat_flux_density       = (thermal_diffusivity*dTdn)*rho_cp_solid;
+    conductivity_over_dist  = thermal_diffusivity*rho_cp_solid/dist;
+  }
+
+  Donor_Variable[0] = Twall;
```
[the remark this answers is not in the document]
oleburghardt: To my knowledge (which is going to be a big restriction for this answer),
there is a way to abstract this to a certain point.

Generally, it would be too hard for a driver to figure out the best choises for the CHT couplings. Even worse, there are some ("mathematical") pitfalls that it _had to_ avoid (e. g., assigning only Neumann/heatflux-BC to a solid solver is not well-posed).

However, for some types of interfaces (independent of the complete multi zone problem), there are coupling procedures one can use as a default.
*So far*, I only added the most important compressible and incompressible fluid-solid coupling (I will add solid-solid, ... subsequently soon), for which there seems to be one preferred method which is the only one I committed. (Dirichlet on the flow side, Neumann or Robin on the solid side.)

If one likes to specify other coupling methods, there should be different BC implementations within the solver classes, respectively.

Within the transfer class, one could also use this information to reduce the overhead, as only parts of the data at the interface are needed.
For now, it's only a development comfort for me that I'm always transferring all four informations that might be used by the solver.

SU2_CFD/include/driver_structure.hpp
```diff
@@ -1408,3 +1409,51 @@ class CDiscAdjFSIStatDriver : public CFSIStatDriver {
...
+
+public:
+
+  /*!
+   * \brief Constructor of the class.
+   * \param[in] confFile - Configuration file name.
+   * \param[in] val_nZone - Total number of zones.
+   * \param[in] MPICommunicator - MPI communicator for SU2.
+   */
+  CMultiPhysicalZonesDriver(char* confFile,
```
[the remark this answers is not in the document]
oleburghardt: Great, thanks!

SU2_CFD/src/driver_structure.cpp
oleburghardt: It's ready. I just wanted to avoid writing the clean version of "CDiscAdjMultiphysicsZonalDriver" before this pull request as you may suggest/request changes (like the naming of the thing :D).

My local discrete adjoint driver (so far only for CHT, but following this approach arbitrary couplings should be no problem, as long as we can formulate them as a sequential one to fit the taping/pseudo-time framework) gives 100% accurate results.

oleburghardt:
@talbring @economon Thanks to you both!!

I could realise most of your remarks. To summarize the leftovers:

1. Concerning the intersecting code with your heat extension to the incompressible solver - should we already find common function names or do we align them once they will be merged?
1. Concerning the numerics class CSourceHeating: I think I will remove it for now, as I don't have any use or test case for it anyway.
1. I will wait for your agreement on the FEM solver, if that's alright.

talbring:
For me everything looks good now. In my opinion we can merge that it and deal with the naming when @economon opens the pull request with his implementation. About the old FEM heat solver: I am going to do a sweeping PR soon to remove some old stuff anyway ...

PR #506, opened 2018-02-10
Removed compiler warnings

vdweide:
This PR removes quite a few compiler warnings that you get when you compile with the following flags for the g++ compiler
-g -Wall -Wextra -Wno-unused-parameter -Wno-empty-body

I would advice people to use these flags before merging into develop to keep the code as clean as possible. Warnings are very useful to catch faulty behavior, but if there are already quite a few present in working code, it is easy to miss important ones.

With these flags there is still one file that gives warnings in the develop version, namely Common/src/geometry_structure.cpp. The reason is that on line 745 of this file there is the following statement

if ((jPoint > iPoint) // && (CrossProduct >= 0.0)

The commented part is new and leads to a few compiler warnings about unused variables. The question therefore is whether the comment should indeed be in there (and hence CrossProduct is not relevant) or this is just something for debugging. In the former case some more cleaning up can be done. In the latter case the comment should be removed again.

### DIFF
--- a/SU2_CFD/src/output_structure.cpp
+++ b/SU2_CFD/src/output_structure.cpp
@@ -16327,7 +16327,7 @@ void COutput::SpecialOutput_AnalyzeSurface(CSolver *solver, CGeometry *geometry,
   unsigned short iDim, iMarker, iMarker_Analyze;
   unsigned long iVertex, iPoint;
   su2double Mach = 0.0, Pressure, Temperature = 0.0, TotalPressure = 0.0, TotalTemperature = 0.0,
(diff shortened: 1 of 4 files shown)


fpalacios:
LGTM

if ((jPoint > iPoint) // && (CrossProduct >= 0.0) seems to me debugging... I'll eliminate the 
// && (CrossProduct >= 0.0) part.

By the way... alter this pull request we are moving to the release_v6.0.0 branch... so we are not going to accept more changes in the develop branch. Thanks!

vdweide:
@fpalacios, you misunderstood me. What I meant was that the CrossProduct part should be uncommented in order to remove the compiler warnings, not removed. When I read this back, this was not very clear from my side. In any case there are still two compiler warnings in geometry_structure.cpp due to this issue in the develop branch.

PR #507, opened 2018-02-13
Fix a few memory issues in 6.0.0

clarkpede:
I've noticed while using Valgrind that some of the recent pull requests have introduced a few memory issues.

Here's a summary of the changes in this PR:                                                                       

+ Sets default for `ilu_fill_in` to 0 to prevent uninitialized value bugs on destruction of `CSysMatrix`
+ Adds deallocation of some turbomachinery variables
+ Adds NULL initialization and deallocation of some of the new markers

I also found a memory issue in a weird corner case with `RefOriginMoment_X`, `_Y`, and `_Z`.  If the reference origin is defined exactly once, but no surfaces are marked for monitoring, then `RefOriginMoment_X` (and `_Y` and `_Z`) is set as an empty array.  I think that's a bug.  That bug creates problems in `CConfig::SetOutput`, (line 4284) where SU2 tries to print out the reference origin.

### DIFF
--- a/Common/src/dual_grid_structure.cpp
+++ b/Common/src/dual_grid_structure.cpp
@@ -659,8 +659,6 @@ CTurboVertex::CTurboVertex(unsigned long val_point, unsigned short val_nDim) : C
 
 CTurboVertex::~CTurboVertex() {
 
-	if (Normal != NULL) delete[] Normal;
-	if (Nodes != NULL) delete[] Nodes;
 	if (TurboNormal != NULL) delete [] TurboNormal;
 
 }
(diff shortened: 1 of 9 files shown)


Common/src/dual_grid_structure.cpp
```diff
@@ -659,8 +659,6 @@ CTurboVertex::CTurboVertex(unsigned long val_point, unsigned short val_nDim) : C
 
 CTurboVertex::~CTurboVertex() {
 
-	if (Normal != NULL) delete[] Normal;
```
talbring: Why is that removed ? Is is deleted somewhere else ?

clarkpede: Thanks for reading through my changes.  I like having a second set of eyes on the code.

`CTurboVertex` is derived from `CVertex`, so the destructor for `CTurboVertex` is called, then the destructor for `CVertex`.  Both destructors have these two lines (`delete [] Nodes` and `delete [] Normal`).  So I got a double-free when I tested a turbomachinery problem.

The simple solution is just to remove the lines from `CTurboVertex` and let the base class destructor handle the deallocation.

clarkpede:
I noticed that the `release_v6.0.0` branch isn't set up with Travis CI. Are there plans to have Travis CI track this branch (v6.0.0), or are the regression tests only happening on other branches?

PR #513, opened 2018-03-11
Fix partitioning

economon:
This Pull Request is the long-awaited overhaul of the front-end partitioning routines for the finite volume solver. It is both a complete refactoring and a streamlining of the routines to remove several bottlenecks (memory and superfluous MPI calls) that were limiting the performance at high rank counts. The rewrite now makes it possible to fix the root issues for the periodic BCs, down to the mesh input level, which is next on the list. 

For now, the old geometry routines are used for processing meshes with periodic BCs. Once the periodic BCs are all updated, we will remove the old CPhysicalGeometry constructor and associated routines in order to clean up.

**Note that this PR affects everyone using the finite volume solver.** The regressions are going to fail :). As there are some changes to the order in which element types are read, stored, and partitioned, some cases can show small changes in the residual output, even in serial (especially with multigrid). I have tested everything successfully up to roughly 4000 cores locally, achieving the same final solutions, but it is important that everyone please check your own cases. If we agree that everything is ok, we will have to modify the stored residual values for many regression cases.


Common/src/geometry_structure.cpp
```diff
@@ -4978,6 +4985,243 @@ CPhysicalGeometry::CPhysicalGeometry(CGeometry *geometry, CConfig *config) {
   
 }
 
+CPhysicalGeometry::CPhysicalGeometry(CGeometry *geometry,
+                                     CConfig *config,
```
fpalacios: Is this the new constructor for CPhysicalGeometry? Is the old one still there?

economon: Yes, this is a new constructor for CPhysicalGeometry that overloads the original. We are keeping the original *only* for periodic cases for the time being, until we can finish that update too. We'll completely remove the original after that.

economon:
Time to merge this one. Please just let me know if you run into any issues with the new partitioning routines.

PR #514, opened 2018-03-15
Feature preconditioning

economon:
This PR is a major upgrade to the existing density-based incompressible solver (Euler, N-S, and RANS). The solver is now outfitted with a custom preconditioning that can handle a range of incompressible problems, including isothermal, decoupled energy, Boussinesq, and variable density flows (low Mach N-S). The solver uses a conservative formulation and is primitive variable-based.

Features:
* Energy equation has been integrated into the incompressible solver. Can be disabled, run in decoupled mode, or coupled for variable density.
* Customized upwind and centered convective schemes in space and time integration schemes that account for preconditioning.
* 2 new fluid models: constant density fluid, incompressible ideal gas (variable density as a function of temperature). Temperature dependent transport coefficients.
* New non-dim. and initialization that mirrors the compressible solver.
* New set of configuration file options specific for incompressible problems (see config_template.cfg).
* Boussinesq approximation for problems with small temperature differences (e.g., natural convection).
* Axisymmetric mode updated, including viscous terms.

Some notes:
* Discrete adjoint is available for new solver, but continuous incompressible adjoint has been removed for now. 
* Method of handling non-constant thermal conductivity in transport model (constant Prandtl w/ RANS) could be reused with compressible code.
* Heat flux reference value for non-dim. is missing from compressible code.
* I will do some clean up after this PR to remove some old “incompressible” flags and code that are no longer needed (output and source terms, especially).
* I have many V&V cases for this solver that I will be adding as additional regressions and tutorials.
* Some effort now needed to sync the CHT work. @oleburghardt @talbring , let's chat about this.


.travis.yml
```diff
@@ -80,11 +80,11 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone -b feature_preconditioning https://github.com/su2code/TestCases.git ./TestData
```
hlkline: I'm sure you intend to change these; just adding a reminder

Common/include/config_structure.hpp
```diff
@@ -509,7 +517,9 @@ class CConfig {
   Kind_DV_FEA;				/*!< \brief Kind of Design Variable for FEA problems.*/
   unsigned short Kind_Turb_Model;			/*!< \brief Turbulent model definition. */
   unsigned short Kind_Trans_Model,			/*!< \brief Transition model definition. */
-  Kind_ActDisk, Kind_Engine_Inflow, Kind_Inlet, *Kind_Data_Riemann, *Kind_Data_Giles;           /*!< \brief Kind of inlet boundary treatment. */
+  Kind_ActDisk, Kind_Engine_Inflow, Kind_Inlet, *Kind_Inc_Inlet, *Kind_Data_Riemann, *Kind_Data_Giles;           /*!< \brief Kind of inlet boundary treatment. */
+  unsigned short nInc_Inlet;  /*!< \brief Number of inlet boundary treatment types listed. */
+  bool Inc_Inlet_UseNormal;    /*!< \brief Falg for whether to use the local normal as the flow direction for an incompressible pressure inlet. */
```
hlkline: typo (Falg)

Common/include/option_structure.hpp
```diff
@@ -371,13 +371,17 @@ enum ENUM_KIND_NONDIM {
   DIMENSIONAL = 0,			    /*!< \brief Dimensional simulation. */
   FREESTREAM_PRESS_EQ_ONE = 1, /*!< \brief Non-dimensional simulation. */
   FREESTREAM_VEL_EQ_MACH = 2, /*!< \brief Non-dimensional simulation. */
-  FREESTREAM_VEL_EQ_ONE = 3 /*!< \brief Non-dimensional simulation. */
+  FREESTREAM_VEL_EQ_ONE = 3,
```
hlkline: I know this is how we've always done it, but enums don't need to have explicit values set unless there's somewhere that it's converted to an int for a function that doesn't have access to the enum definition. Should probably leave it this way, since it's consistent with the rest of the code.

TestCases/cont_adj_incomp_navierstokes/cylinder/lam_incomp_cylinder_disc.cfg
```diff
@@ -33,16 +33,25 @@ WRT_BINARY_RESTART= NO
 % Read binary restart files (YES, NO)
 READ_BINARY_RESTART= NO
 
-% -------------------- INCOMPRESSIBLE FREE-STREAM DEFINITION ------------------%
+% ---------------- INCOMPRESSIBLE FLOW CONDITION DEFINITION -------------------%
 %
-% Free-stream density (1.2886 Kg/m^3 (air), 998.2 Kg/m^3 (water))
-FREESTREAM_DENSITY= 998.2
+% Initial density for incompressible flows (1.2886 kg/m^3 by default)
```
hlkline: I think the note somewhere in this file that the settings are for water would be nice - in the header if not here.

Common/include/config_structure.hpp
```diff
@@ -723,6 +733,7 @@ class CConfig {
   Wrt_Limiters,              /*!< \brief Write residuals to solution file */
   Wrt_SharpEdges,              /*!< \brief Write residuals to solution file */
   Wrt_Halo,                   /*!< \brief Write rind layers in solution files */
+  Wrt_Slice,                   /*!< \brief Write 1D slice of a 2D cartesian solution */
```
fpalacios: What is the reason for this?

Common/include/config_structure.hpp
```diff
@@ -1878,6 +1947,66 @@ class CConfig {
...
+   * \brief Get the value of the reference velocity for custom incompressible non-dimensionalization.
+   * \return Reference velocity for custom incompressible non-dimensionalization.
+   */
+  su2double GetInc_Velocity_Ref(void);
+
+  /*!
+   * \brief Get the value of the reference temperature for custom incompressible non-dimensionalization.
+   * \return Reference temperature for custom incompressible non-dimensionalization.
+   */
+  su2double GetInc_Temperature_Ref(void);
```
fpalacios: Does we need to use the Inc_ prefix? If we need to use it... are we using it for everything? Pressure, etc...

Common/include/option_structure.hpp
```diff
@@ -371,13 +371,17 @@ enum ENUM_KIND_NONDIM {
   DIMENSIONAL = 0,			    /*!< \brief Dimensional simulation. */
   FREESTREAM_PRESS_EQ_ONE = 1, /*!< \brief Non-dimensional simulation. */
```
fpalacios: Could you please update these comments... I'm worried about the fact that SU2 produce different results depending on the non-dimensionalization. And I think is because in the Venkat limiter we are not careful with the epsilon. Have you ever looked at that? I think that DIMENSIONAL is the default for compressible. Are you using the same for incompressible?

SU2_CFD/include/solver_structure.hpp
```diff
@@ -10116,68 +10297,32 @@ class CAdjEulerSolver : public CSolver {
 };
 
 /*!
- * \class CAdjIncEulerSolver
- * \brief Main class for defining the incompressible Euler adjoint flow solver.
- * \ingroup Euler_Equations
- * \author F. Palacios, T. Economon, T. Albring
- * \version 4.2.0 "Cardinal"
+ * \class CAdjNSSolver
```
fpalacios: This header doesn't make too much sense

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -74,8 +74,8 @@ void CUpwScalar::ComputeResidual(su2double *val_residual,
   ExtraADPreaccIn();
 
   if (incompressible) {
-    Density_i = V_i[nDim+1];
```
fpalacios: Is there a place were we describe what are the primitive variables?

config_template.cfg
```diff
@@ -85,8 +85,6 @@ REYNOLDS_NUMBER= 6.5E6
 %
 % Reynolds length (1 m, 1 inch by default)
 REYNOLDS_LENGTH= 1.0
-
-% -------------------- INCOMPRESSIBLE FREE-STREAM DEFINITION ------------------%
 %
```
fpalacios: Are we still using this?

config_template.cfg
```diff
@@ -131,21 +169,22 @@ REF_AREA= 1.0
 % Aircraft semi-span (0 implies automatic calculation) (m or in)
 SEMI_SPAN= 0.0
 %
-% Flow non-dimensionalization (DIMENSIONAL, FREESTREAM_PRESS_EQ_ONE,
+% Compressible flow non-dimensionalization (DIMENSIONAL, FREESTREAM_PRESS_EQ_ONE,
 %                              FREESTREAM_VEL_EQ_MACH, FREESTREAM_VEL_EQ_ONE)
 REF_DIMENSIONALIZATION= DIMENSIONAL
```
fpalacios: To be consistent with INC_NONDIM should we call this option COMP_NONDIM?

oleburghardt:
Hi @TobiKattmann - thanks for that.

> From my (limited) perspective the change could be reverted without any harm
Unlikely... I didn't check, but just have a look at a few lines below that one. But that's maybe also due to the different compiler (or you started a calculation and killed it).
I thought of it as a more elegant way but one can alternatively copy the address initialisations. I think it's for sure in @economon 's interest if you feel free changing it yourself (if you like) ..?

TobiKattmann:
Hey @oleburghardt , 
the error I got can be fixed (using the same setup as described above) via enforcing the C++11 standard (CXXFLAGS=-std=c++11) as I am an older standard (not sure which though). C++11 explicitly introduces constructors that are allowed to call other constructors of the same class.
(see here: https://en.wikipedia.org/wiki/C%2B%2B11#Object_construction_improvement)

For now I commited changes in the constructor as you suggested, after having talked to @economon . If compatibility with older C++ standards is desired I think this is a viable solution.
Sorry for not noticing that right away (Tom suggested to try C++11),
Regards, Tobi

oleburghardt:
@TobiKattmann 
> If compatibility with older C++ standards is desired I think this is a viable solution.

I think so - that's why I agreed with the "hard-coded" alternative. (Although C++11 is not such a vanguard anymore, one will also encouter problems elsewhere not using it. But I think it's perfectly fine that way and it's a minor change anyway.)

oleburghardt:
@economon 
I had a look at the only test case that was still failing - looks that everything is alright. This is the pressure delta after 200 iterations:
schubauer_klebanoff

So it seems that it's still a transition model regression test and didn't turn into something else :smiley:  I will change the values in a minute.

Common/include/option_structure.hpp
talbring: Maybe we should remove this everywhere. Often this leads to unnecessary conflicts.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1095,8 +1095,8 @@ CUpwSca_TurbSST::~CUpwSca_TurbSST(void) {
 
 void CUpwSca_TurbSST::ExtraADPreaccIn() {
   if (incompressible) {
-    AD::SetPreaccIn(V_i, nDim+2);
-    AD::SetPreaccIn(V_j, nDim+2);
+    AD::SetPreaccIn(V_i, nDim+3);
```
talbring: Have you validated the gradient already for the discrete adjoint using the incompressible solver ?

SU2_CFD/src/solver_adjoint_discrete.cpp
```diff
@@ -329,10 +329,41 @@ void CDiscAdjSolver::RegisterVariables(CGeometry *geometry, CConfig *config, boo
     config->SetTotalTemperatureIn_BC(Temperature);
   }
 
+  /*--- Register incompressible initialization values as input ---*/
 
-    /*--- Here it is possible to register other variables as input that influence the flow solution
-     * and thereby also the objective function. The adjoint values (i.e. the derivatives) can be
-     * extracted in the ExtractAdjointVariables routine. ---*/
+  if ((config->GetKind_Regime() == INCOMPRESSIBLE) &&
```
talbring: Finally ;) Thanks for adding that. I always wanted to change that ...

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -4724,46 +5470,47 @@ void CIncEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
...
+
 //      if (viscous) {
-//
-//        /*--- Set laminar and eddy viscosity at the inlet ---*/
-//
-//        V_inlet[nDim+3] = node[iPoint]->GetLaminarViscosity();
-//        V_inlet[nDim+4] = node[iPoint]->GetEddyViscosity();
+//        
+//        /*--- Set transport properties at the inlet ---*/
+//
```
talbring: Maybe we can remove that commented block ?

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -4865,43 +5621,45 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
-      
+
 //      if (viscous) {
 //
-//        V_outlet[nDim+3] = node[iPoint]->GetLaminarViscosity();
-//        V_outlet[nDim+4] = node[iPoint]->GetEddyViscosity();
+//        /*--- Set transport properties at the outlet. ---*/
 //
-//        /*--- Set the normal vector and the coordinates ---*/
+//        V_outlet[nDim+4] = node[iPoint]->GetLaminarViscosity();
```
talbring: Maybe here, too ?

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -6641,7 +7457,7 @@ void CIncNSSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_container,
     
     if (Vol != 0.0) {
       Local_Delta_Time = config->GetCFL(iMesh)*Vol / node[iPoint]->GetMax_Lambda_Inv();
-      Local_Delta_Time_Visc = config->GetCFL(iMesh)*K_v*Vol*Vol/ node[iPoint]->GetMax_Lambda_Visc();
+      Local_Delta_Time_Visc = config->GetCFL(iMesh)*K_v*Vol*Vol/ node[iPoint]->GetMax_Lambda_Visc(); //TDE check
```
talbring: Is it checked ? :D

config_template.cfg
talbring: We should agree in general on a consistent naming scheme for these options. And we should maybe move that option to the compressible free stream definition section.

.travis.yml
economon: Yep, thanks for the reminder. We've got the classic issue where we need to update both the TestCases repo and code repo simultaneously to avoid failure.. I'll push to TestCases right before we merge this one.

Common/include/config_structure.hpp
economon: It's meant to be a hidden feature :). If you have a cartesian 2D grid (structured), I have put in a routine that will extract profiles from lines of nodes with a constant X or Y coordinate. This is particularly helpful for V&V cases, like the NASA turbulence modeling cases for flat plates, etc., where you want to extract the various profiles for comparison with other codes and experiment.

Common/include/config_structure.hpp
economon: This is only for the reference values in the non-dim scheme. For the incompressible solver, I have allowed the user to give directly the reference density, velocity, and temp (if not isothermal), like we originally did for the compressible code. The incompressible solver now uses almost the same non-dim scheme as the compressible code, other than the definition of the reference pressure (rho_ref*V_ref*V_ref).

config_template.cfg
economon: It has been replaced by "INCOMPRESSIBLE FLOW CONDITION DEFINITION" instead, since we are more typically dealing with internal flows. The initialization no longer uses "FREESTREAM" values, rather we simply have *_INIT values.

config_template.cfg
economon: While the compressible and incompressible both use the same enumerated type (it will throw an error if we redefine DIMENSIONAL, for instance), I have used a separate option, INC_NONDIM, for the incompressible non-dim options, which are only DIMENSIONAL, INITIAL_VALUES (default), or specified REFERENCE_VALUES. I separate the option because we may need to supply the INC_*_REF values, and I wanted all of the incompressible options in the config file to be consistent with the INC_* prefix. 

I also did not change the name of the compressible option, REF_NONDIMENSIONALIZATION, since this would break many config files. I did, however, move this option into the compressible section of the config template.

I am open to more comments on this one.. but these were my thoughts.

SU2_CFD/src/solver_adjoint_discrete.cpp
economon: Of course! It is very useful to be able to add these things. It is an initial implementation geared toward internal flows, but could potentially be improved.

SU2_CFD/src/solver_direct_mean_inc.cpp
economon: I have actually decided to uncomment and keep these blocks. However, we should discuss the original change in the compressible code and turb models, which may have been made by @fpalacios? I find that the performance is ok with the incompressible solver, and that this might be important for internal flows or truncated domains.

SU2_CFD/src/numerics_direct_turbulent.cpp
economon: It's looking pretty good so far. Attaching some laminar cylinder checks for 50 FFD vars for both drag and heat flux. I have activated the custom objective and heat objectives for the direct differentiation. I have also added some new objectives, including flow uniformity, secondary flow strength, pressure drop.

drag
heat

economon:
Thanks to all for the detailed comments - very valuable. Please let me know if you have any final thoughts. I would like to get this one merged in the next couple of days. Thanks!

PR #520, opened 2018-03-25
Fix some discrepancies in compute_polar.py

erangit:
Proper restart at each sweep member from the previous member
EXT_ITER from the cfg file is used unless overwritten in script command
More careful folders renaming to avoid inconsistencies in parallel run
Listing case details on the polar-sweep concluding file
Semantic and cosmetic fixes to make pylint less verbal


erangit:
Hi
Kind of strange that Travis tests failed. Modifications were performed only in the script
compute_polar.py and the pull request was set following git pull from develop branch.
Could it be that there is a problem with the online develop branch?
How can I check what caused Travis to fail?
Thanks,
Eran

hlkline:
There are two causes for the failed test: your branch is currently failing the compute_polar test case, and an unrelated issue involving time-outs that is effected tests for everyone. 

You can click the 'details' link to the right of the failed test in section just below the comments, or click on the red x next to any failed test. This takes you to the travis page, where through following through the links you can find which tests failed, and access the test logs. 

The tests with exclamation points on the travis page are the ones that are failing due to other issues, and for now I would suggest focusing on the compute_polar test case.

hlkline:
Please see the end of the SU2_PY/SU2/io/config.py file - this is where defaults for python are currently defined. If you see  any additional ones that are needed for the python scripts to operate, or any that have the incorrect default values, please do update those. 
I would recommend against adding an exhaustive list of defaults, and only adding those that are actually used by the python scripts. The reason is that the default values are not automatically synced between the c++ and python codes, and so this would add unneeded overhead for developers and room for future errors.

erangit:
Heather, thanks for your response.
I do agree that it is preferable to avoid an exhaustive list of defaults. However, currently SU2 runs double booking: For the C++ code, defaults are set in SetRunTime_Options (part of config_structures.cpp) and for the Python scripts, a partial list of defaults is defined in config.py.  If we can automatically sync between the c++ and python codes, it will make developers life much easier. 
I suggest creating a single ASCII file that contains all the defaults values. This file can be read both by the c++ code and by the Python scripts, before overwitting parameters, reading from the case cfg file. 
Such a file can be part of the distribution. Whenever a new parameter is defined in SetRunTime_Options, its default value should be written in the defaults' file. Actually, we can prepare a rendering script that would read through SetRunTime_Options and prepares the defaults in the ASCII file, synced with the c++, for the Python. What do you think?
Best,
Eran

hlkline:
You've reminded me that there is an old python script (parse_config.py) which parses the c++ config structure code - I couldn't get it to run right away, but maybe you can update that script to generate your list automatically from the c++ code, and so not require much change (and also not require you to manually write out all the defaults). 

Personally I think that we should keep the default definitions in the C++ as it is now, as this keeps the executables self-contained among other things. 
By the way, I think you meant SetConfig_Options().

erangit:
Reviving the parsing script can be a simple and effective solution for syncing the defaults.
I shall be out of the country for the next three weeks. I shall try my hand at your suggestion once I return. Meanwhile, I hope the current PR will complete its approval cycle. Current Travis failures are marked by ! only.

fpalacios:
Hi, thanks for this interesting discussion. Just a reminder... there is a third place where the default values are stored! the config_template.cfg file.

I have had similar problems to what @erangit is describing with the multiple definition of the default values in different places, in fact, sometimes is not clear what is the minimum number of parameters that you can use in a config file.

I think that the ideal scenario would be to use the config_template.cfg file as the default value keeper and create subroutines in C++ and python that update the defaults with the existing information in that file. Remember that config_template.cfg is always required otherwise the user doesn't know that are the existing options.

By the way... this discussion reminds me that we should update SetRunTime_Options(void). This is an incredible useful small subroutine to modify the software parameters during runtime. EXT_ITER is the only parameters currently accepted but in the near future we should add all or most of them.

PR #530, opened 2018-05-06
Feature wall functions

VivaanKhatri:
## Proposed Changes
This is an update of the feature_Wall_Functions. In fact, it is only a minor contribution to the original implementation by @economon based on the article "Wall Function Boundary Conditions Including Heat Transfer and Compressibility"
 
## Related Work
The original implementation was 99% done but, it was not working because of 
- two typos on the original article (eq. 11 and eq. 16). 
- not accounting for a negative nu_tilde (very small value... but, negative). 
- not setting tau_wall to -1 on all the BC but non-slip ones.

Apart from those changes, a Newton method has been introduced to compute nu_tilde (it would be important to do the same with the tau_wall computation... it will be done -time depending-), some minor renaming of the functions has been introduced as well as the use of MARKER_WALL_FUNCTIONS (implemented after the original implementation).

## PR Checklist


rsanfer:
Thanks for this pull request! LGTM. I have just updated the branch with develop and the tests are passing, however I will wait for @economon to give his opinion as it was originally his implementation.

Only a minor comment: there are some files in which white spaces are introduced in blank lines, probably done automatically by the editor (e.g., numerics_direct_mean.cpp and solver_direct_mean.cpp). This is not a problem in itself, but I have experienced this can generate problems with future merges.

rsanfer:
Fixing the indentation once and for all is something that we should aim soon. However, we need to be careful on how to do it, as it can generate some conflicts with ongoing branches. We should have a chat about that.

vdweide:
If the indentation is going to be fixed, I strongly suggest the use g++ version 7.3 with the compiler flags -g -Wall -Wextra -Wno-unused-parameter -Wno-empty-body. This gives some additional warnings about indentation.

Common/src/config_structure.cpp
```diff
@@ -2333,6 +2332,23 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+  Wall_Functions = false;
+  if (nMarker_WallFunctions > 0) {
+    for (iMarker = 0; iMarker < nMarker_WallFunctions; iMarker++) {
+      if (Kind_WallFunctions[iMarker] != NO_WALL_FUNCTION)
+        Wall_Functions = true;
+      
+      if ((Kind_WallFunctions[iMarker] == STANDARD_WALL_FUNCTION) || (Kind_WallFunctions[iMarker] == SCALABLE_WALL_FUNCTION)
+        || (Kind_WallFunctions[iMarker] == NONEQUILIBRIUM_WALL_MODEL))
+
+        SU2_MPI::Error(string("For RANS problems, use NO_WALL_FUNCTION, ADAPTIVE_WALL_FUNCTION or EQUILIBRIUM_WALL_MODEL.\n"), CURRENT_FUNCTION);
```
economon: How are you defining the current WF approach in terms of enumerated type? I would think it is a STANDARD_WALL_FUNCTION, and if we allow automatic switching on/off based on y^+, I think it would become a SCALABLE_WALL_FUNCTION. At least, I think this is how it is defined in some other software...

SU2_CFD/include/solver_structure.hpp
```diff
@@ -3830,6 +3830,23 @@ class CSolver {
    */
   virtual void SetFreeStream_Solution(CConfig *config);
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   */
+  virtual void SetTauWall_FW(CGeometry *geometry, CSolver** solver_container, CConfig* config);
```
economon: Did you mean "WF" instead of "FW" for Wall Function?

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -4413,10 +4413,18 @@ void CAvgGrad_Flow::ComputeResidual(su2double *val_residual, su2double **val_Jac
     }
   }
   
+  /*--- Wall shear stress values (wall functions) ---*/
+  
+  if (TauWall_i > 0.0 && TauWall_j > 0.0) Mean_TauWall = 0.5*(TauWall_i + TauWall_j);
+  else if (TauWall_i > 0.0) Mean_TauWall = TauWall_i;
+  else if (TauWall_j > 0.0) Mean_TauWall = TauWall_j;
+  else Mean_TauWall = -1.0;
```
economon: One concern that I had in my original implementation was that this check for positive/negative numbers was not very elegant... don't remember what my original motivation was.. perhaps we should just use a boolean?

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -8058,12 +8058,12 @@ void CIncNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_contai
...
-
-  Wall_Function = config->GetWallFunction_Treatment(Marker_Tag);
-  if (Wall_Function != NO_WALL_FUNCTION) {
-    SU2_MPI::Error("Wall function treatment not implemented yet.", CURRENT_FUNCTION);
-  }
+//  /*--- Get wall function treatment from config. ---*/
+//
+//  Wall_Function = config->GetWallFunction_Treatment(Marker_Tag);
+//  if (Wall_Function != NO_WALL_FUNCTION) {
+//    SU2_MPI::Error("Wall function treament not implemented yet", CURRENT_FUNCTION);
```
economon: I will update the incompressible solver after this is merged.

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -2952,6 +2966,279 @@ void CTurbSASolver::BC_NearField_Boundary(CGeometry *geometry, CSolver **solver_
...
+  /*--- Get the freestream velocity magnitude for non-dim. purposes ---*/
+  
+  su2double *VelInf = config->GetVelocity_FreeStreamND();
+  VelInfMod = 0.0;
+  for (iDim = 0; iDim < nDim; iDim++)
+    VelInfMod += VelInf[iDim];
+  VelInfMod = sqrt(VelInfMod);
+  
+  /*--- Compute the recovery factor ---*/
+  // su2double-check: laminar or turbulent Pr for this?
```
economon: I think it should be the turbulent Pr.

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -2952,6 +2966,279 @@ void CTurbSASolver::BC_NearField_Boundary(CGeometry *geometry, CSolver **solver_
...
+  /*--- Get the specified wall heat flux from config ---*/
+  
+  Wall_HeatFlux = config->GetWall_HeatFlux(Marker_Tag);
+  
+  /*--- Loop over all of the vertices on this boundary marker ---*/
+  
+  for(iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+    iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
+    
+    /*--- We can use also GetNormal_Neighbor, and eliminate the following loop ---*/
```
economon: I think the original intention was to make sure all points that share an edge with the wall are well updated (it's possible to miss points if we just use the normal neighbor that is stored). We could check for repeats though in order to save some effort.

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -2952,6 +2966,279 @@ void CTurbSASolver::BC_NearField_Boundary(CGeometry *geometry, CSolver **solver_
...
+                                              + kappa*kappa*U_Plus*U_Plus/2.0)
+                                             - Lam_Visc_Normal/Lam_Visc_Wall);
+        
+        /*--- Eddy viscosity should be always a positive number ---*/
+        
+        Eddy_Visc = max(0.0, Eddy_Visc);
+        
+        /*--- Solve for the new value of nu_tilde given the eddy viscosity and using a Newton method ---*/
+        
+        nu_til_old = 0.0; nu_til_4 = 0.0; nu_til = 0.0; cv1_3 = 7.1*7.1*7.1;
```
economon: Is it possible to separate this routine into 2 routines for the eddy viscosity calculation (first) and the nu_tilde calculation (second)? The reason is that the first part for eddy viscosity could be put in the parent class for CTurbSolver and be used for all eddy viscosity models, and with a simple short follow up routine, the eddy viscosity can be used to compute nu_tilde for S-A or k and omega in SST within the child classes.

EduardoMolina:
@VivaanKhatri 
Please let me know if you need any help.
Thanks.
Eduardo.

SU2_CFD/src/solver_direct_turbulent.cpp
EduardoMolina: According to the paper is the laminar Pr.

SU2_CFD/src/solver_direct_turbulent.cpp
EduardoMolina: Ok! I will do that.

Common/src/config_structure.cpp
EduardoMolina: So, what is ADAPTIVE_WALL_FUNCTION?

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -2952,6 +2966,279 @@ void CTurbSASolver::BC_NearField_Boundary(CGeometry *geometry, CSolver **solver_
...
+        WallDistMod = sqrt(WallDistMod);
+        
+        /*--- Compute mach number ---*/
+        
+        M_Normal = VelTangMod / sqrt(Gamma * Gas_Constant * T_Normal);
+        
+        /*--- Compute the wall temperature using the Crocco-Buseman equation ---*/
+        
+        T_Wall = T_Normal * (1.0 + 0.5*Gamma_Minus_One*Recovery*M_Normal*M_Normal);
+
```
EduardoMolina: I think it should be:
T_Wall = T_Normal + Recovery*pow(VelTangMod,2.0)/(2.0*Cp);
Following Eq(7) in Eq(10).

EduardoMolina: @economon 
I will add a regression test case similar to the paper once we agree with the modifications.

economon:
I agree, @fpalacios, it's time to merge this and build upon it. I will start with the incompressible implementation. Thanks again @VivaanKhatri

PR #532, opened 2018-05-09
Correct NTS central/upwind blending function

clarkpede:
The equations previously did not match the paper by Travin et al, and
are corrected in this commit.

## Proposed Changes
The hybrid upwind/central blending introduced for DES has the option to use the blending by Travin _et al._ (2002).  It's referred to as the "NTS adaptive" blending or the "NTS" blending for short, and was labeled as a "low dissipation function". But the what's implemented in the code doesn't match the paper, and it doesn't even make sense. You can compare the code to the paper by Travin _et al.__ at: https://doi.org/10.1007/0-306-48383-1_16 or the related AIAA presentation at: https://arc.aiaa.org/doi/pdf/10.2514/6.2001-879

There's multiple issues that I believe are present:

+ There's an extra square root in the calculation of `Baux`, which leads `Baux` to have units of 1/time.  That throws off the nondimensionalization of the blending constant.
+ The square root in the calculation of `Lturb` is in the wrong place.  Travin only takes the square root of the denominator.
+ Travin _et al._ took the variables `Lturb` and `g` and used them to build a new variable, `A`, that was used to set the blending parameter.  In the code, `LTurb` is simply multiplied by `Gaux`. This doesn't match the paper, and it's not going to scale correctly either.  `Lturb` should have units of length, and multiplying a length by a nondimensional parameter does not give you a nondimensional parameter.
+ The variable `Omega` isn't consistent with `StrainMag`.  `StrainMag` has a factor of sqrt(2) built in, but `Omega` doesn't.  The original paper actually doesn't specify if the factor of sqrt(2) should be included, but the code should at least be consistent.
+ The value of the variable `ch3` is 2 in the code, but 0.5 in the paper.

I've corrected these mistakes, but I wanted to get feedback on the results.  This section of code was not covered by any existing regression tests, and the validation cases in the [SU2-DDES paper](http://dx.doi.org/10.2514/6.2017-4284) don't use the NTS blending.  Meaningful verification and validation of hybrid RANS/LES is both tricky and computationally expensive (compared to RANS).  So any input (especially from @EduardoMolina) on what tests to use would be appreciated.  Otherwise, I'll come up with my own test.

## Related Work
This hybrid upwind/central blending was introduced by @EduardoMolina as part of PR #490

## PR Checklist


EduardoMolina:
@clarkpede 
Thank you for catching this. I think a messed up something here while I was merging my implementation to the current version! Please be aware that the papers of Travin et al. and Strelets et al. also have some typographical errors, see the papers of:

Zhixiang Xiao, Jian Liu, Jingbo Huang, and Song Fu.  "Numerical Dissipation Effects on Massive Separation Around Tandem Cylinders", AIAA Journal, Vol. 50, No. 5 (2012), pp. 1119-1136. 
https://arc.aiaa.org/doi/abs/10.2514/1.J051299 
Mockett, "A Comprehensive Study of Detached Eddy Simulation" PhD Thesis.

Regarding the DDES test cases, I agree that Travis will not handle any meaningful 3D test case. Do you have any suggestion? By now, all I can do is to test it outside. 

Best,

Eduardo.

SU2_CFD/src/variable_direct_mean.cpp
```diff
@@ -612,35 +612,55 @@ bool CNSVariable::SetStrainMag(void) {
...
+  Baux = (ch3 * Omega * max(StrainMag, Omega)) / max((pow(StrainMag,2)+Omega_2)*0.5, 1E-20);
   Gaux = tanh(pow(Baux,4.0));
   
   Kaux = max(sqrt((Omega_2 + StrainMag)*0.5), 0.1 * inv_TimeScale);
   
-  Lturb = sqrt((GetEddyViscosity() + GetLaminarViscosity())/(cnu*Kaux));
+  const su2double nu = GetLaminarViscosity()/GetDensity();
+  const su2double nu_t = GetEddyViscosity()/GetDensity();
+  Lturb = (nu + nu_t)/sqrt(cnu*Kaux);
+
```
EduardoMolina: According to Mockett and Xiao, Lturb = sqrt((nu + nu_t)/(cnu*Kaux)). Do you tested both definitions?

SU2_CFD/src/variable_direct_mean.cpp
```diff
@@ -612,35 +612,55 @@ bool CNSVariable::SetStrainMag(void) {
...
   
-  Lturb = sqrt((GetEddyViscosity() + GetLaminarViscosity())/(cnu*Kaux));
+  const su2double nu = GetLaminarViscosity()/GetDensity();
+  const su2double nu_t = GetEddyViscosity()/GetDensity();
+  Lturb = (nu + nu_t)/sqrt(cnu*Kaux);
+  
+  Aaux = ch2*max((C_DES*delta/Lturb)/Gaux -  0.5, 0.0);
   
-  Roe_Dissipation = Lturb*Gaux;
+  Roe_Dissipation = sigma_max * tanh(pow(Aaux, ch1));
```
EduardoMolina: Thank you for clearing the mess here.

clarkpede:
Ah, I wasn't aware of the corrected version.  Thanks for pointing out my mistake.  The corrected version does make more sense, and the main changes seem to be:

+ Changing the location of the sqrt in the definition of `Lturb`
+ Changing `ch3` from 0.5 to 2.0
+ Clarifying that S and Omega should have a factor of sqrt(2)

As for code verification, I would propose using one of the cases examined by Travin, Stretlets, Mockett, or Xiao for comparison.  We know that they got a nice range of values for the blending constant on their tests with DDES.  The cases I can see where they explicitly show the results of the blending function are:

+ Circular cylinder at Re = 50,000 (Travin _et al._ 2002)
+ Tandem cylinders at Re = 166,000 (Xiao _et al._ 2012)
+ NACA0021 airfoil in deep stall (Mockett 2009)

The tricky thing is that the value of the blending constant is determined by the flow solution, which is determined by the solver.  So the values we obtain will be dependent both on the version of DDES used and the nature of the code.  We can't expect to match any of these papers exactly.

We could also use a simpler flow, such as fully developed channel flow or a flat plate boundary layer.  The tradeoff would be simplicity for ease of comparison.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -17320,7 +17320,15 @@ void CNSSolver::SetRoe_Dissipation(CGeometry *geometry, CConfig *config){
     }
     
     if (kind_roe_dissipation == NTS || kind_roe_dissipation == NTS_DUCROS){
-      node[iPoint]->SetRoe_Dissipation_NTS();
+      /*--- XXX: This grid length does not match the original paper.
```
economon: Remove XXX?

clarkpede: What is the SU2 convention for commenting a "hack"? I wanted to denote that the grid length used here does *not* match the grid length from the original papers.

economon: I see.. no standard "hack" convention. Unless you need something searchable, I think your comment is clear enough.

EduardoMolina:
Thanks @clarkpede 

I will merge it. We can talk more about regressions tests in another topic.

Best,

Eduardo

PR #554, opened 2018-07-18
Bugfix DiscAdjFluid: DualTimeStepping 1st order

TobiKattmann:
## Proposed Changes
Hi all,
there is a little bug in the loading of the primal solution for the DiscAdjFluid Iteration for DualTimeStepping 1st order. 
1. The solution for timestep n-1 should be loaded instead of n-2 -> I added an if statement there.
2. solution_old is used as a temporary container for the freshly loaded solution. Then the solution which is currently at solution_time_n is pushed to be the "main" solution n+1. Afterwards the solution_old is put in the solution_time_n container -> That last step was wrong such that the loaded solution never gets involved.

I used the regression test in disc_adj_rans/cylinder which covers DualTimeStepping 2nd order and reused it for a new Testcase which covers 1st order. 
For now, the .travis.yml has a different su2code/Testcases branch in it such that other are not hindered by this PR when running travis. If this PR is merged with the new Testcase, the TestCases repo has to be updated and the .travis.yml needs to be reverted. 
Let me know if adding a new TestCase is OK. I am aware that not every detail can get its own regression test. 

I checked my changes against finite differences for that Testcase. I'll post the result below.

Additionally I suppressed the output of .csv files in driver_structure.cpp for unsteady simulations if `WRT_CSV_SOL= NO` is set. 

Regards, Tobi

## Related Work
Nothing I know of.

## PR Checklist


TobiKattmann:
As promised (drag) gradient comparison:
The blue curve represent the current develop branch with the bug. Bugfix and FinDiff are hopefully self-explanatory. The results are taken from my Testcase, so it is really small (10 timesteps), i.e. for a longer running case the gradient deviates a lot more.
Let me know if you want some other comparisons :)

image

image

.travis.yml
```diff
@@ -80,7 +80,7 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone -b feature_bugfix_DT1st_DiscAdjFluid https://github.com/su2code/TestCases.git ./TestData
```
rsanfer: You can do a pull request to the develop branch of the TestCases first to incorporate the new case.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -3719,7 +3719,7 @@ void CDriver::Output(unsigned long ExtIter) {
 
   /*--- Export Surface Solution File for Unsteady Simulations ---*/
   /*--- When calculate mean/fluctuation option will be available, delete the following part ---*/
-  if ((config_container[ZONE_0]->GetUnsteady_Simulation() == DT_STEPPING_2ND) && (ExtIter % config_container[ZONE_0]->GetWrt_Surf_Freq_DualTime() == 0)) {
+  if ((config_container[ZONE_0]->GetUnsteady_Simulation() == DT_STEPPING_2ND) && (ExtIter % config_container[ZONE_0]->GetWrt_Surf_Freq_DualTime() == 0) && config_container[ZONE_0]->GetWrt_Csv_Sol()) {
```
rsanfer: Nice catch, thanks! We'll need to rework some of this, but it's good that csv files will no longer be forced to be written.

TobiKattmann:
Hi @economon @rsanfer and @talbring ,

There are no additional warnings generated. In addition to a check of the build history in travis (@rsanfer gave me the hint and checked that too) (not including any AD-build), I compared the stderr-output of this branch with the latest develop 58892c9. I used all the additional compiler flags and enabled mpi, autodiff and direct-diff and ran the build on a single core (i.e. make install 2> error.out 1> output.out). Otherwise (i.e. multiple cores) the order gets mixed up and it is impossible to compare the warning/error output. 
A diff on the "error.out" of the two builds showed the files are identical (...like completely identical), although the files have ~18.000 lines. The stdout deviates only in the specified build dirs.

For now I checked the box above for this PR. If there's s.th. else to check let me know. 

@economon I guess you're right in demanding to check the compiler warnings to ensure a nice code. Being strict in this regard probably helps everybody in the long run.

Thanks you all for the support although it is just a small contribution! It helps a lot.
Thanks! Tobi

PR #555, opened 2018-07-19
Feature rbf

pcarruscag:
## Proposed Changes
Radial Basis Function interpolation for zone to zone transfer without matching meshes.
Main purpose is to use it with the FSI solvers, direct and adjoint. 

## Related Work
Also:
Adds under-relaxation to the adjoint FSI solver to facilitate convergence of strongly coupled problems.
Fixes missing turbulence solver dependencies.
Adds fluid node sensitivities to the output (before they were not set).

## PR Checklist


LaSerpe:
Hi, 
thanks for this contribution, I will get a more detailed feedback soon.

A first comment about the CSymmetricMatrix class came up to my mind, from a first brief review.
There is already a CSysMatrix class and a few routines added in this pull request look redundant (matvect product for instance).
On the other hand, many of those routines (cholesky decomposition and so on) could turn out to be very useful also for other future code developments.
That said, perhaps it would be nice to collect all the matrix-related routines into the CSymmetricMatrix class (removing redundancies) to make them available at an higher level.

Cheers,
Giulio

pcarruscag:
Hi Giulio,
Thank you for the feedback. I inherited some of this code from a previous student of my PhD supervisor and I thought the same when I saw the CSymmetricMatrix class. However, as I understand it, CSysMatrix implements a block sparse format and here we have a simpler dense format. Another big difference is that CSymmetricMatrix does not need to be used in parallel. But I agree that making these two classes related somehow would be better, so I am open to suggestions.
Cheers,
Pedro

rsanfer:
Hi Pedro,
thanks for cleaning up Joel's code and creating the pull request. The disc_adj_fsi testcase is failing, I believe because of the relaxation. Probably you need to update the value of the test_case config and set the relaxation to 1.0 to recover the previous test. 
About the symmetric matrix class: Joel's implementation was indeed intended to populate the matrix needed for the RBF; the CSysMatrix would not be able to accommodate some terms due to the sparsity. It's well explained in the comments. Also, his implementation was only single-core. I agree that it would be very interesting to have some methods such as the cholesky decomposition available, but I think they would require some reworking (at least for parallelization).
I think this is a very valuable contribution, so it would be good to get more people on board to push it forward.
Cheers,
Ruben

pcarruscag:
Hi Ruben,
Thanks for the tip it is passing the test now.
Cheers,
Pedro

LaSerpe:
Hi,
I just had a few (very minor) comments.
Regarding the main contribution, everything makes sense to me and I didn't find any relevant issue.
This is a very nice contribution and the implementation is quite clear, we can go ahead and merge this in soon.

Regarding the matrix stuff, perhaps we could start having a new, general, matrix class in Common and then extend it to parallel in future pushes.
It may not be ideal, but at least we'll get started.
I am just afraid that if we leave those potentially useful routines there, in the interpolator, we will soon forget about them (of course it doesn't have to be addressed in this PR but we should really discuss about this)

ciao,
Giulio

pcarruscag:
Hi Giulio,
Great news, thanks!
After looking at CSysMatrix more, one possibility would be to move dense matrix routines to a class that would replace/augment this CSymmetricMatrix and then CSysMatrix could use it for operations on its blocks (which are dense). But since this involves refactoring a core class of SU2 I agree more people should be involved in the discussion.
Cheers,
Pedro

economon:
Any final comments/requests on this one? @pcarruscag : I noticed that the check box for warnings in the PR template is empty. Can you verify that the changes result in no new compiler warnings? Thanks!

pcarruscag:
That should do.
@economon by the way, what is the right way to add a regression test/example? Do I just add the cfg and appropriate instructions to the scripts, and the mesh and initialization to the Testcases repo? If so does the order matter? Thanks.

rsanfer:
Any further comments on this? I think we could get it merged in during this week.

LaSerpe:
I have some very minor comments, but it looks good to me.

Cheers,
Giulio

Common/include/interpolation_structure.hpp
```diff
@@ -180,7 +189,8 @@ class CInterpolator {
    * \param[in] nDim - number of physical dimensions.
    */
   void Collect_VertexInfo(bool faces, int markDonor, int markTarget, unsigned long nVertexDonor, unsigned short nDim);
-
+  
+  void Collect_VertexInfo(bool faces, int markDonor, int markTarget, unsigned long nVertexDonor, unsigned short nDim, unsigned long &nLocalVertex_Donor);
```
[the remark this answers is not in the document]
pcarruscag: That parameter is meant as an output, there are ways to give it a default value but it makes that intent less obvious, I could however change the return type of the original function but that kind of breaks the pattern. Thoughts?

Common/src/interpolation_structure.cpp
```diff
@@ -2719,3 +2729,1240 @@ bool CSlidingMesh::CheckPointInsideTriangle(su2double* Point, su2double* T1, su2
...
+  delete PPT;
+  delete [] coeff;
+  if (max_diff<max_diff_tol) {
+    delete [] write_row;
+    delete [] P_tmp;    
+  }
+  
+}
+
+void CRadialBasisFunction::Get_Distance(su2double *Coord_i, su2double *Coord_j, unsigned short nDim, su2double &dist)
```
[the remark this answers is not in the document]
pcarruscag: Specifically, I think someone was distracted XD. Thanks for spotting this.

rsanfer:
@pcarruscag done, it's merged in already so you can uncomment your test case.

Common/include/interpolation_structure.hpp
```diff
@@ -384,4 +394,106 @@ class CSlidingMesh : public CInterpolator {
...
+   */
+  ~CRadialBasisFunction(void);
+
+  /*!
+   * \brief Set up transfer matrix defining relation between two meshes
+   * \param[in] config - Definition of the particular problem.
+   */
+  void Set_TransferCoeff(CConfig **config);
+
+  void Check_PolynomialTerms(int m, unsigned long n, su2double *P, int *skip_row, int *keep_row, su2double max_diff_tol_in, bool P_transposed, int &n_polynomial);
```
talbring: Some comments would be good here in order to be consistent.

Common/src/interpolation_structure.cpp
```diff
@@ -2719,3 +2729,1240 @@ bool CSlidingMesh::CheckPointInsideTriangle(su2double* Point, su2double* T1, su2
...
+  /*--- Initialize transfer coefficients between the zones ---*/
+  Set_TransferCoeff(config);
+
+}
+
+CRadialBasisFunction::~CRadialBasisFunction() {}
+
+void CRadialBasisFunction::Set_TransferCoeff(CConfig **config) {
+
+  int rank=MASTER_NODE, nProcessor=SINGLE_NODE;
```
talbring: rank and size are already member variables of the base class and properly initialized. So you can remove them here.

Common/src/interpolation_structure.cpp
```diff
@@ -2719,3 +2729,1240 @@ bool CSlidingMesh::CheckPointInsideTriangle(su2double* Point, su2double* T1, su2
...
+  if (!initialized) {
+    throw invalid_argument("Matrix not initialized.");
+  }
+
+  /*--- Point to correct vector ---*/
+  if (overwrite) {
+    vec = val_vec;
+  }
+  else {
+    decompose_vec = new double [num_val];
```
talbring: If you have to use double here, use the typedef passivedouble. That eventually avoids some problems.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -6386,6 +6408,11 @@ void CDiscAdjFSIDriver::Fluid_Iteration_Direct(unsigned short ZONE_FLOW, unsigne
 
   solver_container[ZONE_FLOW][INST_0][MESH_0][FLOW_SOL]->Preprocessing(geometry_container[ZONE_FLOW][INST_0][MESH_0],solver_container[ZONE_FLOW][INST_0][MESH_0], config_container[ZONE_FLOW], MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, true);
 
+  if(turbulent){
```
talbring: Maybe it is better to have these changes in a separate PR ? Or are they related to the RBF implementation ?

pcarruscag: Not strictly, they are fixes/features I ended up doing while testing the interpolation.

Common/src/interpolation_structure.cpp
pcarruscag: Most of the code uses su2double but the areas that use LAPACK use double explicitly in case someone decides to make passivedouble a float (the code only calls the double version of the routines).

pcarruscag:
@LaSerpe @talbring all done except the passivedouble issue mentioned above, but if you think my concern is unfounded I'll do a replace all.
Cheers,
Pedro

hlkline:
@talbring  it looks like your requested changes were addressed - do you agree?

talbring:
@pcarruscag sorry for being picky :D but I would like to have the double changed to passivedouble. In case someone needs a float we can name it passivefloat. Sometimes I use automated scripts to check for occurences of double variables, so we should use passivedouble whenever we can to avoid future confusions on whether the use of double is intentionally or not.

pcarruscag:
@talbring fair enough, have a look now, I hope I did not make it worse xD.

talbring:
Do we really need the templates right now ? Or this is just in case ? I think if we make su2double a float we get also problems somewhere else (furthermore using float makes actually no sense on current architectures regarding performance)

pcarruscag:
No. Yes. Plenty of other places, I checked. (Not true)

economon:
Just another comment when it comes to single precision: this could be interesting from an HPC perspective in the future, or at least for some flexibility for research for future architectures. 

Some time back, I made the changes necessary to run with floats, and you can see those changes in this branch: https://github.com/su2code/SU2/tree/feature_single_precision. It actually wasn't too much work. We know from previous experience that in some cases you can get away with it for fluid simulations (Euler, N-S, and maybe S-A turbulence model.. prob not SST).

talbring:
Thanks for addressing these things @pcarruscag. LGTM

PR #558, opened 2018-08-06
Tecplotszl

ScottImlay:
Modified the Tecplot binary output to use the latest TecIO library and to write Tecplot szplt (SZL) files instead of Tecplot plt files. Tecplot szplt files are smaller than plt files and are utilized more efficiently in Tecplot. For large files visualization within Tecplot can be much faster and use far less memory with szplt files. To make this change work correctly, the TecplotBinary DomainMesh and DomainSolution files were modified to write a single mixed-element zone instead of separate zones for each element type.


economon:
@ScottImlay : it is wonderful to have your support from Tecplot, and we really appreciate the contribution for improvements through the open source!

I have a first question... I noticed that there are many new files, especially headers from the Boost library. Is it possible to reduce to a smaller set of files that get included, or are all of those that have been added necessary?

ScottImlay:
Hi Tom,
I did a Boost extract (bcp) which theoretically only includes the required header files. I'll look and see if it can be reduced further.
Scott
    On ‎Wednesday‎, ‎August‎ ‎8‎, ‎2018‎ ‎09‎:‎10‎:‎15‎ ‎PM‎ ‎PDT, Thomas D. Economon <notifications@github.com> wrote:  
 
 
@ScottImlay : it is wonderful to have your support from Tecplot, and we really appreciate the contribution for improvements through the open source!

I have a first question... I noticed that there are many new files, especially headers from the Boost library. Is it possible to reduce to a smaller set of files that get included, or are all of those that have been added necessary?

talbring:
Like @economon already said, this is a interesting and useful contribution! Thanks! 

I don't know why travis is not reporting. I am gonna look into it tmrw.

ScottImlay:
Thanks Tim!
Tom, I have not found any easy (and repeatable) way to reduce the BOOST header files further. Unless it is a major problem, I suggest we go with them as they are.
Thanks!Scott Imlay
    On Monday, August 13, 2018, 1:56:19 PM PDT, Tim Albring <notifications@github.com> wrote:  
 
 
Like @economon already said, this is a interesting and useful contribution! Thanks!

I don't know why travis is not reporting. I am gonna look into it tmrw.

talbring:
@ScottImlay Travis is probably not running because the pull request was opened with the master branch as base branch. It is not recognizing the change to develop ... Can you reopen it with directly with develop as the base branch ?

ScottImlay:
OK, I'll try closing it and reopening on the develop branch.

economon:
It looks like our regression tests are failing during the compilation of TecIO on the Travis system (they use pretty standard Ubuntu builds). @ScottImlay: could you please take a look at the build errors when you have some time?

ScottImlay:
Hi Tom,
I've consulted with our main TecIO developer, Dave Taflin, and the two of us are having a hard time finding the problem. There are a bunch of compile warnings in TecIO, but none of them look like they would cause the linker error. The actual error seems to be in the link of SU2, not the TecIO library. Could it be something I changed in the config stuff (configure.ac, m4/tecio.m4, or Makefile.am)?
Is it possible to get a more verbose error report for the linker?
Thanks!Scott Imlay
    On Sunday, August 19, 2018, 2:44:12 PM PDT, Thomas D. Economon <notifications@github.com> wrote:  
 
 
It looks like our regression tests are failing during the compilation of TecIO on the Travis system (they use pretty standard Ubuntu builds). @ScottImlay: could you please take a look at the build errors when you have some time?

economon:
For updating the build, I would recommend the following: after adding the headers and sources in the correct directories, modify only configure.ac to do the correct checks for the tecio option, m4/tecio.m4 to configure the compilation of tecio, and externals/tecio/Makefile.am to make sure the proper headers/files are being included in the build. 

Then, as a final step, run the ./bootstrap script in the root directory of SU2/, which will build and run autoreconf with an older, standard version of the autotools that we ship for portability. This will update the makefiles throughout the build for you, and from there, you can run configure - make - install. Just let me know if you're still having trouble.

ScottImlay:
Hi Tim,

I think I fixed the problem yesterday. I'll do a final check this morning and then redo the pull request.

ScottImlay:
Looks like it passed the Travis checks.

talbring:
Thanks @ScottImlay. Seems to work now. Just a quick question. You only adapted the volume mesh/solution routines. Is it also possible to use the new format for the surface files (i.e. in the routines SetTecplotBinary_SurfaceMesh/SetTecplotBinary_SurfaceSolution) ?

ScottImlay:
Thanks Tim @talbring. The new format offers little benefit for purely surface files. I could make the change, but I wanted to focus on the volume data first. If you think it is important, I can make the change to the surface files as well.

talbring:
Thanks for the answer. If it has no benefit we can leave it like that. I was just worried that is not possible to load szplt and plt files together into tecplot. But I tried it and it works :+1: 

I am happy with the change. As soon as there is further approval, we can merge it in.

talbring:
Hm I agree with @economon that this is a problem. We can certainly add it to the preconfigure.py and/or have it as a git submodule. But then it should be possible to deactivate the use of szplt and use the old plt format instead ...

ScottImlay:
Thanks Tom and Tim. I agree that the BOOST include files make the library much larger and that isn't ideal. If we wish to download it on the fly there are a couple things to consider:

1. The extracted include files are much smaller than the full BOOST library. I'm traveling today so I don't have the sizes handy, but I will look up the details when I'm in the office tomorrow.
2. There is a bug in the latest release of BOOST that requires us to include an earlier version of BOOST. Hopefully they will release a new version that fixes this bug, and this problem will go away.

I will double check the sizes tomorrow and post an update.

talbring:
Another possibility would be to compress the boost files and uncompress it during configure/preconfigure. With tar.gz this reduces the size down to 15 MB.

economon:
Good idea, @talbring. 

@ScottImlay : I think we should try the compression to a tarball, and then hopefully we can add a command in the m4 file to unpack the boost source before build. If you have time can you please take a look at this? Let me know if you need any help.

talbring:
I compressed the boost sources and added a call to extract it to configure.ac (its much easier than in the m4 scripts, since you can use shell commands). I also added a check whether the boost folder exists, so that it is only done the first time you call configure.

economon:
I was just giving this a try on a mac, and while the configure/unpack of boost seems to work fine, I get the following error:

teciosrc/auxdata.cpp:14:2239: error: cast from pointer to smaller type '___90' (aka 'int') loses information

This is with LLVM. @ScottImlay : anything obvious I am missing w.r.t. data types or the build steps for mac? Maybe there is something with 32 v 64 bit. Have others given this a try yet?

PR #563, opened 2018-08-20
Bug fix

fpalacios:
## Proposed Changes
Collection of improvements to fix SU2 bugs:
    - Inconsistency between NO_DEFORMATION (index 24) and NONE (index = 0). I have updated and sorted the numbering of the design variables and set NO_DEFORMATION to index 0 (for consistency with the rest of the code).
   - SU2_GEO was not working properly for 2D cases and the wing slicing routine was extremely expensive for the AD.
    - VISUALIZE_DEFORMATION was very expensive and not useful, I have divided VISUALIZE_DEFORMATION between surface VISUALIZE_SURFACE_DEF and volume VISUALIZE_VOLUME_DEF.
   - The actuator disk was not working properly when a pressure ratio (total pressures) was prescribed (the restart was not working either).
   - The FFD twist design variable was not working properly

## Related Work
This commit resolves the issues (bugs) described above without adding any extra capabilities to the software. It should be approved as soon as possible because some of the changes directly impact most of the the user's community.

## PR Checklist


SU2_CFD/src/output_structure.cpp
```diff
@@ -8593,21 +8598,25 @@ void COutput::SetMesh_Files(CGeometry **geometry, CConfig **config, unsigned sho
...
+        
+        /*--- Write an stl surface file ---*/
+        
+        if (rank == MASTER_NODE) cout <<"Writing .stl surface file." << endl;
+        
+        SetSTL_MeshASCII(config[iZone], geometry[iZone]);
+        
+        /*--- Write a binary file with the grid coordinates alone. ---*/
+        
+        if (Wrt_Crd) {
```
economon: Could you please move the "if (Wrt_Crd) { }" conditional outside of the "if (su2_file) { }" conditional?

TestCases/aeroelastic/aeroelastic_NACA64A010.cfg
```diff
@@ -254,8 +254,11 @@ DEFORM_TOL_FACTOR = 0.001
 %                                          WALL_DISTANCE, CONSTANT_STIFFNESS)
 DEFORM_STIFFNESS_TYPE= INVERSE_VOLUME
 %
-% Visualize the deformation (NO, YES)
-VISUALIZE_DEFORMATION= NO
+% Visualize the surface deformation (NO, YES)
+VISUALIZE_SURFACE_DEF= NO
```
economon: Do you think it is cheap enough to turn the surface file writing by default? I think at least the surface visualization can be very important...

SU2_CFD/src/output_structure.cpp
fpalacios: Thanks for the review! Done!

TestCases/aeroelastic/aeroelastic_NACA64A010.cfg
fpalacios: Agree! Done!

PR #570, opened 2018-08-29
Feature sst uq

jayantmukho:
## Proposed Changes
Adds epistemic uncertainty quantification capabilities for the SST turbulence model. Details on methodology, implementation, and performance on test cases, are available here: https://arxiv.org/abs/1803.00725
 

## Related Work
NA

## PR Checklist

- [ X ] I am submitting my contribution to the develop branch.
- [ X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [ X ] I have added a test case that demonstrates my contribution, if necessary.


jayantmukho:
The TravisCI checks with autodiff and direct-diff seemed to have failed. Not sure why, since I didn't change any of the files that throw the errors. All the errors seem to originate from the codi package. Any suggestions on how to fix that?

jayantmukho:
@talbring Is this an issue you have encountered in the past, or would you have any suggestions on how to fix the issue in this branch?

economon:
@jayantmukho : I am seeing two possible reasons for the test failure.. 

1. There are changes noted to some of the autotools files that I don't normally expect. Did you run the ./bootstrap script locally in order to update the Python script install list? If not you may want to try again, since we should see minimal changes recorded in the commit.

2. The logs report that externals/codi changed, but I don't think anything should be modified from there. Can you please revert any changes in that directory?

jayantmukho:
@economon Fixed the both the issues that you mentioned. Now the files changed reflect only the changes I intended to make.

jayantmukho:
Looking at some of the other pull requests, it seems like if I add a test case to the repository, I should also add it to the regression tests. For this test case, SU2 has to run 5 different simulations. So I am assuming I should write a regression test for each of the simulations. 

Just wanted to confirm this before I change the regression scripts

juanjosealonso:
Jayant,

Indeed: we want the regression tests to provide “coverage” of the code. In this case, each of your 5 sims runs a slightly modified version of the turbulence model and therefore requires a different test case. Of course, you only need to run each case for a short time and compare with a pre-recorded reference outcome.

Best,

Juan

On Sep 10, 2018, at 11:43 AM, Jayant Mukhopadhaya <notifications@github.com<mailto:notifications@github.com>> wrote:


Looking at some of the other pull requests, it seems like if I add a test case to the repository, I should also add it to the regression tests. For this test case, SU2 has to run 5 different simulations. So I am assuming I should write a regression test for each of the simulations.

Just wanted to confirm this before I change the regression scripts

Common/src/config_structure.cpp
```diff
@@ -3656,6 +3671,20 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
   	}
   }
 
+  /* --- Throw error if UQ used for any turbulence model other that SST --- */
+
+  if (Kind_Solver == RANS && Kind_Turb_Model != SST && using_uq){
+      cout << "UQ capabilities only implemented for SST turbulence model" << endl;
+      exit(EXIT_FAILURE);
```
economon: Please switch these to use the SU2_MPI::Error() routine instead of exit(). There are examples in this routine.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -821,6 +827,25 @@ class CNumerics {
...
+   * \param[in] val_gradprimvar - Gradient of the primitive variables.
+   * \param[in] val_turb_ke - Turbulent kinetic energy
+   * \param[in] val_normal - Normal vector, the norm of the vector is the area of the face.
+   * \param[in] val_laminar_viscosity - Laminar viscosity.
+   * \param[in] val_eddy_viscosity - Eddy viscosity.
+   * \param[in] val_reynolds_stress - Reynolds stress.
+   * \param[in] val_perturbed_rsm - Perturbed Reynolds Stress Matrix
+   */
+
+  void GetViscousProjFlux(su2double *val_primvar, su2double **val_gradprimvar,
```
economon: I think this routine is overloaded multiple times.. any unexpected use cases that might break things? For example, if someone tries to run the UQ module with a non-ideal fluid model?

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -3454,6 +3511,32 @@ class CAvgGradCorrected_Flow : public CNumerics {
 	 * \param[in] val_tauwall_j - Value of the wall shear stress at point j.
 	 */
   void SetTauWall(su2double val_tauwall_i, su2double val_tauwall_j);
+
+  /*!
+   * \brief Initialize the Reynolds Stress Matrix
+   * \param turb_ke turbulent kinetic energy of node
```
economon: Very nice to see all the doxy comments 👍 !

Don't forget to add the '[in]' to 'param' here in many of these, or I think it will cause problems with doxygen later. Might want to do a sweep through your PR for this.

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -4716,6 +4728,223 @@ void CAvgGradCorrected_Flow::ComputeResidual(su2double *val_residual, su2double
...
+
+  GetMeanRateOfStrainMatrix(S_ij);
+
+  /* --- Using rate of strain matrix, calculate Reynolds stress tensor --- */
+
+  for (iDim = 0; iDim < 3; iDim++){
+    divVel += S_ij[iDim][iDim];
+  }
+
+  for (iDim = 0; iDim < 3; iDim++){
```
economon: A number of loops and mem allocations are running up to '3' in the dimension. Is this safe even for 2D cases? We typically run the loops up to nDim so that it is not hard-coded. Please double-check these throughout the PR, just in case.

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -4716,6 +4728,223 @@ void CAvgGradCorrected_Flow::ComputeResidual(su2double *val_residual, su2double
...
+
+  for (iDim = 0; iDim < 3; iDim++)
+    delete [] S_ij[iDim];
+  delete [] S_ij;
+}
+
+void CAvgGradCorrected_Flow::SetPerturbedRSM(su2double turb_ke, CConfig *config){
+
+  unsigned short iDim,jDim;
+  su2double **A_ij;
```
economon: You might consider moving some of these allocations to be class data and then only allocate/delete them once in the constructor destructor, only if they are active, of course. This will be more efficient than doing in each call to the numerics routines.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1222,42 +1222,320 @@ void CSourcePieceWise_TurbSST::ComputeResidual(su2double *val_residual, su2doubl
...
+     SetPerturbedStrainMag(TurbVar_i[0]);
+     pk = Eddy_Viscosity_i*PerturbedStrainMag*PerturbedStrainMag
+     - 2.0/3.0*Density_i*TurbVar_i[0]*diverg;
+   }
+   else {
+     pk = Eddy_Viscosity_i*StrainMag_i*StrainMag_i - 2.0/3.0*Density_i*TurbVar_i[0]*diverg;
+   }
+
+
+   pk = min(pk,20.0*beta_star*Density_i*TurbVar_i[1]*TurbVar_i[0]);
```
economon: One thing you and I discussed some time ago was exposing the production limiter as a config parameter. Some later variants of SST call for small changes to that value. Are you interested in adding this here?

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1222,42 +1222,320 @@ void CSourcePieceWise_TurbSST::ComputeResidual(su2double *val_residual, su2doubl
...
+  for (iDim = 0; iDim < 3; iDim++)
+    delete [] S_ij[iDim];
+  delete [] S_ij;
+}
+
+void CSourcePieceWise_TurbSST::SetPerturbedRSM (su2double turb_ke, CConfig *config){
+
+  unsigned short iDim,jDim;
+  su2double **A_ij;
+  su2double **Eig_Vec;
```
economon: Same comment about mem allocations w.r.t. being class data and hard-coded '3' values.

SU2_CFD/src/numerics_structure.cpp
```diff
@@ -2810,6 +2920,285 @@ void CNumerics::SetRoe_Dissipation(const su2double Dissipation_i,
...
+  for (i = 0; i < 3; i++){
+    delete [] tmp[i];
+  }
+  delete [] tmp;
+}
+
+void CNumerics::tred2(su2double **V, su2double *d, su2double *e) {
+
+    unsigned short i,j,k;
+    /*  This is derived from the Algol procedures tred2 by        */
```
economon: Attribution here is good to have, but do they give specific guidance about the usage of this code (is it open)? Is there a specific license or statement we need to include? Or did you write it yourself with only inspiration from their work?

SU2_PY/compute_uncertainty.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python
+
+## \file Compute_Mpolar.py
+#  \brief Python script for performing sweep of Mach number.
```
economon: Please update header

TestCases/parallel_regression.py
```diff
@@ -468,6 +468,65 @@ def main():
     contadj_rans_rae2822.tol       = 0.00001
     test_list.append(contadj_rans_rae2822)
 
+    #############################
```
economon: 👍

pcarruscag:
Overall some solid looking code @jayantmukho, thanks!
I have a few suggestions:
1 - Some of the config variables have fairly broad names, "permute" for example, I think prefixing them with UQ would more clearly identify what they are meant for.
2 - There are a lot of dynamic allocations of static size, if the current implementation is compatible with 2 and 3 dimensions I would allocate those variables statically, makes for cleaner and faster code (stack allocations are much faster and subsequent loops would likely be unrolled by the compiler).
3 - It looks like some of the new methods in CNumerics are for generic "Eigenvalue-stuff" for re-usability I suggest moving those methods to a helper class, even if for now that class becomes just a collection of static methods. In the implementation of those methods the dimension of the inputs is being assumed, if you do move them, make them generic.
Cheers,
Pedro

SU2_CFD/src/numerics_direct_mean.cpp
jayantmukho: So the reason we need 3 dimensions, regardless of the dimension of the problem, is that we need 3 ordered eigenvalues and eigenvectors (even if one of them is 0). It makes the rest of the code a lot cleaner to just have everything be of dimension 3. 

I checked through the PR and it is consistent throughout, so it shouldn't cause an issue. That being said, if you want these built to nDim, that can be done. It would require more if statements in the code and I don't know which would be more efficient. 

Currently the eigen-decomposition functions are hard coded to work with symmetric tensors of order 3. But I am working on making that general to n order. But I think I would prefer having the tensors be of order 3 regardless.

SU2_CFD/src/numerics_structure.cpp
jayantmukho: This was essentially converted from fortran code (https://netlib.sandia.gov/eispack/tql2.f) to be used with c++. I attributed it according to how it was attributed in the fortran code. Not sure what exactly is the policy on that. Since the code was adapted, not exactly copied. 

This is from EISPACK which is now superseded by LAPACK

Common/src/config_structure.cpp
jayantmukho: Fixed it! Thanks, I didn't realize we had shifted to that error routine.

SU2_CFD/include/numerics_structure.hpp
jayantmukho: That's a good point. I tweaked the implementation and removed this overloaded function. Now it uses the two existing GetViscousProjFlux functions. This was possible by incorporating one of the other suggestions; the one about moving some of the data structures to the CNumerics class and initializing them in the constructor. 

This way, the Perturbed Reynolds Stress Tensor is a variable in the CNumerics class and doesn't need to be passed into the GetViscousProjFlux function. 

The UQ methodology is now implemented for the general fluid, and for both the CorrectedGrad and AvgGrad classes as well.

SU2_CFD/src/numerics_direct_turbulent.cpp
jayantmukho: I think I might do a different pull request for that, just so it is treated as a separate issue

SU2_CFD/src/numerics_direct_mean.cpp
jayantmukho: This was a great suggestion. The code works 25% faster now

SU2_CFD/src/numerics_direct_turbulent.cpp
jayantmukho: Addressed in other comments

SU2_PY/compute_uncertainty.py
jayantmukho: fixed it!

jayantmukho:
Hey @pcarruscag !

Thanks for the suggestions. 

1) That's a good point. I will change the option names to be more specific. 

2) Not a 100% certain what you mean by this. To be clear, instead of allocating memory using the keyword new (MeanReynoldsStress  = new su2double* [3];), I should be declaring them statically (su2double MeanReynoldsStress[3][3])? I was following the variable declaration norms I saw in the code. But I guess those were usually for allocations to nDim. Is that what you are suggesting?

3) I have generalized the Eigen-value functions to n order matrices now. But I think I am going to keep them in the numerics class. Since it seems like that is where they would be most useful. Would you suggest otherwise?

Cheers,

pcarruscag:
Hi @jayantmukho 
1. Thanks!
2. I am not saying you should (because I am no authority in these matters) but I would xD, making those static would also make initialization easier e.g. su2double MeanReynoldsStress[3][3] = {{0, 0, 0},{...},{...}}; and it would not require explicit deallocation with operator delete. Just be aware that you would not be able to pass a static 2D array to a function taking a dynamic 2D array e.g. su2double **A.
3. As a general remark about SU2 and not just this PR, I think these matrix-like objects and associated operations (construction, destruction, addition, decomposition, etc.) should have their own class, and be defined in some kind of math library. So I think anything that moves us in that direction is positive. That being said I am ok with the "eigen functions" living in numerics for now, but make them static so that people can use them without an instance of numerics (if they are already general it is just a matter of adding the keyword).
Cheers,
Pedro

jayantmukho:
Good point about making the eigen-functions static. That makes it a lot more accessible in other parts of the code. Will include that in the changes along with the option name changes. 

I like the idea about the static allocation, but I foresee a problem that you actually mention. If I convert the variables to be statically defined, I would have to make the eigen-functions accept statically defined arrays. Since most arrays in SU2 are dynamically allocated, this would cause some compatibility issues. 

In general, I see the value in making a math library associated with matrix operations. Maybe you could pitch the idea in an issue and get feedback on it?

jayantmukho:
Made the last of the changes that I wanted too in connection to the comments provided by @economon and @pcarruscag . Just as a review of the things I changed: 

1) Changed the error message
2) Implemented the UQ methodology for a non-ideal fluid and for both gradient methods (AvgGrad and AvgCorrectedGrad). 
3) generalized the eigen-decomposition functions for order n matrices. Also made the functions static so that they can be used without the need for an instantiation of the CNumerics class
4) Improved performance by moving variable declarations to the class constructor for CNumerics
5) Changed UQ config option names to be more specific. Added an output message for when people use the previous names.
6) This instigated some changes to the TestCases config files, and to the compute_uncertainty.py script.
7) Some general code clean up

Let me know if there are any other improvements that I should include.

SU2_CFD/src/numerics_structure.cpp
jayantmukho: I couldn't find the license explicitly for EISPACK. It seems like they might not have thought about it when they wrote those routines and put them online. But they have the modified BSD license on the LAPACK website: http://www.netlib.org/lapack/LICENSE.txt 

As for the modified C++ routines, they would require some changes as well as it has an unnecessary output that requires additional computations. In the end, it would look identical to what is in the code currently, perhaps with a couple of more helper variables. 

Since the code would end up being the same, I would prefer attributing to the original EISPACK implementation. Does that mean that I should provide the link to the license file in the code?

economon: Yep. We should provide the original license, or some written permission to use the code from the author/copyright holder. If you can find the original license, we can embed it in the comments in this file, or we can ship the file with the rest of the code.

jayantmukho: So I have repeatedly emailed the LAPACK email list about the EISPACK license but haven't gotten a reply back from them. So I am not sure how to proceed. The only license I have found is still for LAPACK. 

I also found this thread: http://icl.cs.utk.edu/lapack-forum/archives/lapack/msg00202.html 

It seems to say that there is no licensing restrictions on EISPACK code. We could reference it in the way it is referenced in the c++ version of the code by jburkardt that you linked in the previously, which I have copied here: 

//  Author:
//
//    Original FORTRAN77 version by Smith, Boyle, Dongarra, Garbow, Ikebe,
//    Klema, Moler.
//    C version by John Burkardt.
//
//  Reference:
//
//    Martin, Reinsch, Wilkinson,
//    TRED2,
//    Numerische Mathematik,
//    Volume 11, pages 181-195, 1968.
//
//    James Wilkinson, Christian Reinsch,
//    Handbook for Automatic Computation,
//    Volume II, Linear Algebra, Part 2,
//    Springer, 1971,
//    ISBN: 0387054146,
//    LC: QA251.W67.
//
//    Brian Smith, James Boyle, Jack Dongarra, Burton Garbow,
//    Yasuhiko Ikebe, Virginia Klema, Cleve Moler,
//    Matrix Eigensystem Routines, EISPACK Guide,
//    Lecture Notes in Computer Science, Volume 6,
//    Springer Verlag, 1976,
//    ISBN13: 978-3540075462,
//    LC: QA193.M37.

talbring:
@economon Do you accept the changes so that we can merge it in soon ?

SU2_CFD/src/numerics_structure.cpp
economon: Ok, please add a statement similar to the reference that you have pasted here.. at least there is some precedent for that from the linked website. If you hear anything more from email, we can update accordingly.

economon:
Yep, time to get this one in. After updating the statement and the tests pass, let's merge.

SU2_CFD/src/numerics_structure.cpp
economon: Maybe one last comment: you could also try reaching out personally to one of the original authors to ask for explicit permission and document that.

Issue #568, opened 2018-08-28
Segfault in SU2_SOL for multiple zones


TobiKattmann:
Hi all, hi @rsanfer,

I get a segfault when running SU2_SOL on problems with multiple Zones for the develop branch. Specifically from the Testcases I tried:
 
- `turbomachinery/APU_turbocharger/Jones_rst.cfg`
- `sliding_interface/rotating_cylinders/rot_cylinders_WA.cfg` (I saw a regression test failing for @economon which I was able to reproduce)

I did some debugging and I think the code is wrong in `CBaseLineSolver::LoadRestart(...)` in solver_structure.cpp starting in line 3925. Inside that routine the `geometry` container is called as `geometry[iZone]` to address the final `C(Physical)Geometry` object. I think that the addressable level is the instance level i.e. it should be `geometry[iInst]`. 
As the Jones_rst case has two zones but only one instance the object `geometry[iZone=1]` in LoadRestart refers to what is `geometry_container[iZone=1][iInst=1]` in SU2_SOL.cpp - which does not exists and should be `geometry_container[iZone=1][iInst=0]`. Unfortunately it does not fail there directly for as it reaches (I guess) valid memory and simply interprets what it finds which leads to wrong evaluation of e.g. `GetnDim()=2376` or so. 

I extracted a patch (apply with `patch -p1 < _patchfile_ ` in the root of the code directory) for the current develop branch which fixes the segfault problem for the above mentioned cases but I don't know whether it works correctly if multiple Instances are computed. 
[SU2_SOL_segfault_patch.txt](https://github.com/su2code/SU2/files/2328459/SU2_SOL_segfault_patch.txt) (github does not accept *.patch files...)

Is there a testcase which covers multiple zones and multiple instances. At best with nZones!=nInstances? Can somebody confirm that this is actually a bug and if the patch fixes it?  I probably don't know if it occurs somewhere else, so if somebody thinks their problems are related let me know. It is no problem to do a PR if I get some feedback.

The (possible) bug was introduced here https://github.com/su2code/SU2/commit/1a75ebd0343694977a98c945833928bcc96932c1 

I hope my explanation of the bug above is somewhat understandable :)
Cheers Tobi

economon:
@TobiKattmann : nice catch, thanks for the effort to debug this one. I have also been seeing this in the regressions. 

@rsanfer or @arubino : do you have a case that you used for testing the new instance index, possibly with multiple zones including harmonic balance?

[...]

PR #572, opened 2018-08-30
Bugfix SU2_SOL segfault for multiple zones

TobiKattmann:
## Proposed Changes
Hi all,
 
a wrong index (iZone vs iInst) caused the SU2_SOL module to segfault if nZone > nInst.  The geometry container is called with the wrong index in the `CBaseLineSolver::LoadRestart` routine. In SU2_SOL.cpp `&geometry_container[iZone][INST_0]` to `geometry_container[iZone]` are interchangeable, I just thought the 2nd formulation is used more often throughout the code. I cannot guarante though that it now runs for multiple instances, but as far as I know there is no such testcase.

Testcase: The PR does not come with a Testcase which would catch the error in the future. In testcases started with parallel_computation.py the SU2_SOL module is called and fails for multizone cases (e.g. Jones_rst.cfg which I tested on) but if the residuals fit the case passes. That circumstance is not tackled in this PR but I heard that this is a know issue. I briefly chatted with @rsanfer about that and I offer my help if this is approached. 

Additionally I fixed an error message to contain the correct config-option and added a --disable-tecio option to the preconfigure.py script as it is convenient for me as I build on compute nodes which cannot access the necessary X11 libraries and I couldn't get it to work with tecio enabled :(   

If there are comments or requests for changing s.th. let me know. 

Best Regards, Tobi

## Related Work
Please see issue https://github.com/su2code/SU2/issues/568 for more information. There the error is described more detailed there.

## PR Checklist

- [ X ] I am submitting my contribution to the develop branch.
- [ X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ compiler:
 notifications:
     email:
         recipients:
-            - su2code-dev@lists.stanford.edu
+            - tobias.kattmann@de.bosch.com
   
 branches:
     only:
-        - develop
+        - fix_SU2SOL_segfault
 
 python:
     - 2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ compiler:
 notifications:
     email:
         recipients:
-            - tobias.kattmann@de.bosch.com
+            - su2code-dev@lists.stanford.edu
   
 branches:
     only:
-        - fix_SU2SOL_segfault
+        - develop
 
 python:
     - 2.7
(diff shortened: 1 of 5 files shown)


rsanfer:
Thanks a lot, @TobiKattmann, for this fix,

I was able to reproduce the error locally as well and the fix makes perfect sense to me. I've only added one small statement:
`config_container[iZone]->SetiInst(INST_0);`
to SU2_SOL in cases with single instance, to enforce that iInst = config->GetiInst() in CBaselineSolver::LoadRestart() receives the correct value.

@economon as of now, I think there are no multizone-multiinstance cases available in develop. When that comes through, it would only require an extra loop on iInst and the correct setting of
`config_container[iZone]->SetiInst(iInst);`
in the very same fashion as it is done for harmonic balance (lines 357-380 in SU2_SOL.cpp). Then, inside the LoadRestart routine, the correct container position is read in line 3939 of solver_structure.cpp,
`unsigned short iInst = config->GetiInst();`
and used onwards. 

I'll wait until the tests pass for this latest commit and then merge it in by Monday morning if there are no further comments.

PR #574, opened 2018-09-04
Feature error message

vdweide:
## Proposed Changes
This small PR modifies the error handling, such that the error message is also written when the call to the error handler is not collective. The function CBaseMPIWrapper::Error is not so small anymore, so it may be an idea to move it to the .cpp file instead of inlining it.
 
## Related Work
N/A

## PR Checklist

- [ X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [ X] My contribution is commented and consistent with SU2 style.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,11 @@ compiler:
 notifications:
     email:
         recipients:
-            - su2code-dev@lists.stanford.edu
+            - e.t.a.vanderweide@utwente.nl
   
 branches:
     only:
-        - develop
+        - feature_error_message
 
 python:
     - 2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # Continous Integration setup for SU2.
 # Tests on the develop branch in both serial and parallel.
 
-dist: bionic
+dist: xenial
 sudo: required
 
 language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # Continous Integration setup for SU2.
 # Tests on the develop branch in both serial and parallel.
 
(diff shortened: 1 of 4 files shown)


vdweide:
I forgot to mention that MPI-3 functionality is required for these changes. Can we assume that all practical MPI implementations support the MPI-3 standard or should we add a test in configure to check?

vdweide:
The MPI version used by travis is too old to support all the new MPI functions. Canceling this one for now.

vdweide:
I ifdeffed the availability of MPI_Ibarrier for checking whether the error call is collective. If MPI_Ibarrier is not present, there is a delay in the writing of the error message of 1 second (to make sure that the buffers for the one-sided communication are updated). If MPI_Ibarrier should be used when available (this is the cleaner solution), an additional check in configure must be added.

talbring:
Thanks @vdweide. This looks like a good solution. I think we could determine whether MPI_IBarrier is available or not from the version macro that the MPI implementations provide.

vdweide:
@talbring, is it easy to add such a test to configure? I have no idea how to do this to be honest?

vdweide:
According to the MPI standard the macro MPI_VERSION should be defined. However, even if it is not defined, the MPI_Ibarrier will not be called. This is exactly what you want. I'll push these changes shortly and see if the regression tests go through.

talbring:
Perfect, that looks good! No further comments, will merge that in soon.

vdweide:
Alright, merging in then.

vdweide:
Looks like I screwed up the merge

PR #575, opened 2018-09-05
Change dynamic viscosity in FD hybrid central / upwind blending to kinematic viscosity

clarkpede:
## Proposed Changes
In both [the original paper on DDES](https://doi.org/10.1007/s00162-006-0015-0) and [the paper](https://doi.org/10.2514/6.2017-4284) by @EduardoMolina on hybrid RANS/LES in SU2, the calculation for `r_d` uses kinematic viscosity.  In SU2, the dynamic viscosity is currently used.  This leads to poor behavior if density is not close to 1.  This PR exchanges the dynamic viscosity for the kinematic viscosity.

## Related Work
This is part of the changes originally proposed in PR #552

## PR Checklist

### DIFF
--- a/SU2_CFD/src/variable_direct_mean.cpp
+++ b/SU2_CFD/src/variable_direct_mean.cpp
@@ -677,7 +677,7 @@ void CNSVariable::SetRoe_Dissipation_FD(su2double val_wall_dist){
   
   static const su2double k2 = pow(0.41,2.0);
   
-  su2double uijuij = 0, r_d;
+  su2double uijuij = 0;
   unsigned short iDim, jDim;
   
   AD::StartPreacc();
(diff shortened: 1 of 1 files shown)


clarkpede:
The behavior can be seen in the backward facing step case, as shown in a snapshot below.  As currently set up, `Roe_Dissipation` is equal to `f_d` from the original DDES paper.  So `Roe_Dissipation` should:

+ Go to 0 in attached boundary layers
+ Go to 1 in regions with resolved turbulent fluctuations

Here's a comparison of the `Roe_Dissipation` before and after this PR:

viscosity_comparison

Here's the same case, but zoomed in at the step:

viscosity_comparison_zoom

With this PR, the model does indeed go to 0 near the walls and go to 1 in the highly unsteady ("LES") region after the step.

vdweide:
I don't know too much about this topic, but the first thing that comes to my mind when I read this is dimensional consistency. So did somebody do a proper dimensional analysis on the implementation?

Edwin

clarkpede:
@vdweide It's easy to do, so I'll show the dimensional analysis here.

I'll use L for length, T for time, and M for mass.

The original paper defined `r_d`, which is supposed to be a nondimensional parameter.  (`r_d` gets raised to the 3rd power and put into a hyperbolic tangent, so it would be messy if it were dimensional).

In the code you'll see:

     r_d = (nu + nu_t)/(uijuij*k2*pow(val_wall_dist, 2.0));

where `nu` is the kinematic viscosity, `nu_t` is the kinematic eddy viscosity, `uijuij` is the magnitude of the velocity gradient tensor, k2 is the square of the von-Karman constant, and `val_wall_dist` is the wall distance.

+ `nu` has units of L^2/T
+ `nu_t` has units of L^2/T
+ `uijuij` has units of 1/T
+ `val_wall_dist` has units of L

So if you do the math, `r_d` has units of L^/T in the numerator and L^2/T in the denominator.  `r_d` is nondimensional.

If you switch `nu` and `nu_t` from kinematic viscosities to dynamic viscosities, you get that `r_d` has units matching density, or M/L^3.  That's obviously not correct.

vdweide:
Merging things in. I kept the branch, but delete it if you don't need it anymore.

PR #577, opened 2018-09-08
Fix NTS duplicate calc

clarkpede:
## Proposed Changes

This pull request is two related changes:

1. While "fixing" the NTS central/upwind blending function in PR #532, I accidentally introduced a bug.  Some steps in the calculation were repeated twice. The main purpose of this pull request is to remedy the bug.
2. I moved the calculation of the maximum cell width to the `CPhysicalGeometry` class.  This makes inclusion in the central/upwind blending easier. For DES and DDES, the cell lengthscale is not dependent on the flow.  So there's no need to compute it every iteration in the flow solver.
 

## Related Work

This is a subset of the changes originally proposed in PR #552. The discussion and validation there still apply here.  This pull request is not a new set of changes, but rather a smaller portion of PR #552.

## PR Checklist

### DIFF
--- a/Common/src/geometry_structure.cpp
+++ b/Common/src/geometry_structure.cpp
@@ -1381,7 +1381,7 @@ void CGeometry::UpdateGeometry(CGeometry **geometry_container, CConfig *config) 
   geometry_container[MESH_0]->SetCoord_CG();
   geometry_container[MESH_0]->SetControlVolume(config, UPDATE);
   geometry_container[MESH_0]->SetBoundControlVolume(config, UPDATE);
-  geometry_container[MESH_0]->SetMaxLength();
+  geometry_container[MESH_0]->SetMaxLength(config);
   
   for (iMesh = 1; iMesh <= config->GetnMGLevels(); iMesh++) {
     /*--- Update the control volume structures ---*/
@@ -14958,7 +14958,7 @@ void CPhysicalGeometry::SetBoundControlVolume(CConfig *config, unsigned short ac
   
 }
 
-void CPhysicalGeometry::SetMaxLength(void) {
+void CPhysicalGeometry::SetMaxLength(CConfig* config) {
 
   for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++){
     const unsigned short nNeigh = node[iPoint]->GetnPoint();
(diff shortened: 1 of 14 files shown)


clarkpede:
#### Verification

The NTS method sets the value of the `Roe_Dissipation`, or the central/upwind blending constant.  In this context, 1 = upwind scheme, 0 = central scheme. The NTS method was explicitly designed to make `Roe_Dissipation`:

+ Go to 1 in attached boundary layers
+ Go to 1 in irrotational regions, even if strain is present (i.e., the freestream)
+ Go to 0 in regions with high vorticity.

Below is an example result from the backward facing step problem.  The figure compares the current develop branch with the branch from the PR.  Something definitely looks wrong with the current develop branch, but the PR looks much better.  All three goals seem to be met.

nts_comparison

Common/src/geometry_structure.cpp
```diff
@@ -14957,6 +14958,34 @@ void CPhysicalGeometry::SetBoundControlVolume(CConfig *config, unsigned short ac
   
 }
 
+void CPhysicalGeometry::SetMaxLength(void) {
```
talbring: The computation in this routine has two major issues. First of all, if run in parallel it is not taking into account that there are neighbors on other processes. So a communication is necessary. Second, if using AD, the computation of a maximum generates a lot of unnecessary overhead since we store all computations of delta_aux, although we only need the maximum value. I know it is not very intuitive, but one way to solve that is to do the computation of the maximum first using the passivedouble datatype to find out the iNeigh index. Then, redo the computation of the delta_aux using su2double datatype.

Let me know if you need help.

clarkpede: @talbring Thanks for pointing out my mistakes. This PR is even more important then, because those problems are also present in the original code (i.e., the `develop` branch).  The max calculations are just moved to a new place in this PR--I didn't change the underlying algorithm.

I should be able to fix the MPI communication myself.

For the AD portion, I can commit some changes and then have you review them.  I haven't worked much with the AD side of SU2, so my knowledge is limited.

clarkpede: @talbring Can you take a look and see if my recent commits have fixed the two problems you identified?

talbring:
Thanks @clarkpede for addressing the things I mentioned.  Let me know if you think that this branch is ready to be merged.

clarkpede:
@talbring Last I heard, @EduardoMolina was going to run a test on this branch to check it.  Aside from that, there's nothing left to complete on my side.

PR #579, opened 2018-09-12
Fix Commands for SOL_FSI, SOL and GEO in SU2_PY/SU2/run/interface.py

Patschke:
## Proposed Changes
Like for any other command in this file, build_command will add a quote on the left hand side of our command path. So we'll always need another quote on the right hand side. For some reason SOL_FSI, SOL and GEO were missing them, resulting in Errors when attempting to execute them.

## Related Work
Same as #578 but now against develop branch. Sorry for creating a PR against master...

## PR Checklist

- [ x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).


Patschke:
Those changes in option_structure.hpp were actually done to the main branch by somebody else. I forked from main, and merged develop. I can of course undo them, if you want (or rebase my commit on develop instead of merging).

economon:
Whichever is easier for you. After that, we can merge in.

Issue #584, opened 2018-09-21
Autoconf/automake options


rsanfer:
In link to a comment from @jayantmukho in  #573, I would like to raise the issue of the current versions of autotools which we maintain in SU2. In particular, in my local machine I can not longer run ./bootstrap, as the compilation of automake fails at:

  GEN      doc/automake-1.12.1
help2man: can't get `--help' info from automake-1.12
Try `--no-discard-stderr' if option outputs to stderr

Searching around the web, this appears to be a known issue when the local machine runs a newer version of Perl (in my case 5.26), see e.g.
http://gnu-automake.7480.n7.nabble.com/bug-29178-Failure-to-build-automake-gt-11-td22684.html 
so I foresee that as we move on this kind of problems will start appearing more often. 

We are still working with automake 1.12 from 2012...

vdweide:
I'm running into problems as well on some machines. So I think it is a good idea to update the autotools to a more recent version. Don't know how much work this is though.

economon:
Resolved in #586

PR #586, opened 2018-09-22
Fix configure

economon:
## Proposed Changes
 
This PR is a clean up of the configure options and bindings. I have removed unused bindings: HDF5, ZLIB, SZIP, JSON, and LAPACK.  If anyone needs these, just let me know and we can put the necessary bindings back into configure.

## Related Work

We can also think about changing the autotools here as discussed in #584.

## PR Checklist


vdweide:
My preference would be to change the autotools in this PR as well.

talbring:
What if we remove all the generated files (Makefile.in) from the repository except for the master branch ? In my opinion that makes perfectly sense since developers (so all people which use branches other than master) usually have autotools installed.

erangit:
Tim,
I think the Makefile.in are useful also in the developer branch. First for
cases where you install a new machine. Second, for new users who are
already opensource-expirienced and tend to work with develop branch.
Best
Eran

rsanfer:
Hi @economon,
thanks for cleaning this. One question, I think that LAPACK may be used by the RBF feature in #555, but to confirm, @pcarruscag, is this the case?
Regarding the autotools, I agree with @vdweide, I think we should use this PR to fix them. I think that @talbring proposal is a good idea, it would avoid the constant conflict that we have when we develop in different machines. I would also update the version of autotools provided in externals, so when bootstrap is run a recent version is compiled.

vdweide:
@rsanfer, you can always specify the lapack libraries by setting the environment variables LDFLAGS and LIBS. This is what I do and it gives you more control than specifying it in configure.

pcarruscag:
@rsanfer I've only done it through the environment variables like Edwin suggests, and that is how I documented it. That way the libraries can be anywhere which is handy when you can't "install" them in the machine you want to run on.

rsanfer:
@vdweide @pcarruscag ah ok, I wasn't sure how it was working on the RBF but seeing that there is no need for it to be in configure, it's better to clean it up as well.

talbring:
@rsanfer and me had a discussion on it. We both came to the agreement that it might the best if we just remove the generated files from the repo in general. Since buildtools are already required to have 'make' it shouldn't be much of a burden to also require autotools/automake to be installed. Furthermore, we eventually could provide a simple way of installing/compiling it within the preconfigure.py script if necessary.

economon:
I tend to agree w Eran on this topic. When it is working as it should, it is very convenient to have the makefiles already available in the repo. I switch between Mac, Ubuntu, and a red hat cluster without needing to update anything or call the bootstrap script. I rarely use it.. basically only when adding or removing files to the build. Also, it is good to minimize dependencies, as sometimes old machines or clusters have limited tool sets.

But, clearly we should at a minimum fix the issue that some of you are finding with the autotools, which should just need an update. What systems are you having trouble with? It’s important that we don’t jump too far ahead for portability reasons (that’s why we have kept the old versions for so long). Do you have an idea of the minimum autotools you need for the failing systems? We could start by updating to that version and see if things stabilize.

vdweide:
@economon, I agree with you that you only need to run the bootstrap when you add new files. At least that is the case for me, I don't know about other users. Unfortunately, I had to do that relatively often, which didn't always run smoothly with the autotools shipped with SU2. Sometimes I had to use the autotools installed on the system, which now runs Ubuntu 18.05 (automake 1.15.1). 

I also agree with you that we should make sure that the entire build process works on as many systems as possible. Therefore I would say that Tim's suggestion to remove all the automatically generated files from the repository is a good one, because then you also remove all possible conflicts. Of course we have to make sure that .gitignore is updated accordingly, but that is just a detail.

hlkline:
It looks like replacing autotools is fairly straightforward - while I'm at it I'll go ahead and remove the .in and .m4f files from the repo. 
I'll push that up to this branch once I test it on my own system - however I will only have tested this on an ubuntu system so it would be good to double check it on a different OS.

economon:
Thanks for taking a shot at this. Just let us know if you need some help, and I am sure several of us can help with the testing. I can try on multiple systems here.

vdweide:
I can test it on both Centos 7.5 and Ubuntu 18.04, if needed.

hlkline:
Two options: one (most recent commit, 23ed092) which requires running ./bootstrap for all new installs, and in commit ce9cd2b that updates automake but doesn't require ./bootstrap (through including updated Makefile.in's, etc). 
updated automake found here: https://ftp.gnu.org/gnu/automake/

vdweide:
I tested this on Ubuntu 18.04 and Centos 7.5. On Ubuntu both options work fine. However, on Centos only the most recent commit (i.e. running ./bootstrap) works. The commit ce9cd2b fails, because the system runs automake 1.13 and therefore it cannot find aclocal-1.15 when running configure. This can be solved by either changing am__api_version in configure to 1.13 (dirty) or running bootstrap. However, I don't know if either option is desired. BTW, this problem will also show up for the master branch, whatever option will be chosen, as for the master branch the configure and Makefile.in's will be in the distribution.

In short, the option with bootstrap works fine on the operating systems I tested (and I assume on all Linux and Unix distributions). When configure is present in the distribution, you could run into problems if the system does not have the required version of autotools.

vdweide:
And believe it or not, but the bootstrap script even works under Windows (with MSys2).....

hlkline:
It sounds like the version requiring bootstrap is the way to go - given that this is a choice between either not working on 1.15 or not working on 1.13. Assuming that updating .gitignore doesn't cause any problems that's it for now from me. 
We will probably need to either add a line about running bootstrap first to the installation directions, and/or have it taken care of inside the preconfigure.py script prior to this getting merged to master at the next release.

economon:
I can report that both seem to work ok for me on macOS 10.13, including the bootstrap script.

It would be good to work out our strategy so that this doesn't cause issues for those that download the source distribution, keeping in mind that many users still build from source (the binaries are only serial).

vdweide:
@erangit, it looks like we are converging to the option where the automatically generated files are removed from the repository. Is that OK for you?

vdweide:
@economon, I think you forget one thing. The configure script requires aclocal. When a newer version of the autotools is used to create the configure script, it does not work (as far as I know) and you still need the bootstrap in order to get things to work. The solution to this would be to use a less recent version of autotools, but that is not what we want either, I guess.

hlkline:
If you'd like me to revert the changes updating the autotools version and/or removal of auto-generated files, just let me know (you can also simply revert the last few commits for the same effect).

economon:
Thanks, @hlkline.

Any final comments/suggestions from the group? Otherwise, we will make a decision and move forward with this in the next days.

talbring:
How do we proceed on this one ? A democratic vote ? :D I still think we save ourselves lot of troubles  if we just remove the files. And executing one further command (./bootstrap) should not hurt anyone.

talbring:
And a VCS is in general not used for releasing code, but for development. So everybody who clones the repo should be considered as a (soon-to-be) developer with the proper tool-chain installed on their machines.

vdweide:
I am in favor of removing.

talbring:
Anyone else has an opinion ?

talbring:
Anyone else has an opinion ?

rsanfer:
I'd also go for removing, I think it would make life much easier for developers.

talbring:
Anyone else has an opinion on that topic ?

talbring:
Anyone else has an opinion on that topic ?

economon:
LGTM. Thanks for the discussion. We need to keep moving, so let's go for the removal! We can always reevaluate if we have any issues in the future. For now, let's get this merged asap.

One last request: @talbring, is it quick for you to add a ./bootstrap call to the preconfigure.py script, just to save a step? Not critical, but would be nice. Once happy, please approve and merge this one (I can't approve my own PR :) ).

I have updated the README.md. I am also creating an issue to remind us to update the docs in all locations with build instructions for the next release.

vdweide:
@economon, I am happy to approve it, but at the moment there are a few conflicts due to the merging with the new tecplot output.

PR #590, opened 2018-09-27
Let's add the contributor covenant

economon:
As discussed at the 3rd SU2 Developers Meeting, we would like to make sure that the SU2 project remains as it has been since the start: an open and welcoming place for any and all contributors. 

Let's add the contributor covenant as a proactive way to set our expectations as a community (which is growing each day!). I have listed my own email address as a contact point initially.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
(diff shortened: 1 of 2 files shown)


talbring:
Thanks Tom! To me that looks like an important step. We can merge that in asap!

hlkline:
Just added a couple of quick edits (some of it was punctuation). 
Regarding adding a 'consequences' list: I've seen this done in other policies, and I think that it is a good idea to explicitly list consequences not only because it makes it a bit more 'real' for those who might need some convincing, but also because uncertainty about the consequences can lead to unnecessary fear and defensiveness on both sides (of a situation that hopefully will never come about).
Please feel free to revert those changes and/or change the consequences to fit what you think is appropriate.

vdweide:
Let's merge it in then.

vdweide:
@economon, I am not sure if you still need this branch, so I didn't delete it.

PR #592, opened 2018-10-03
Feature pre-accumulation for the structural solver

pcarruscag:
## Proposed Changes
Use of pre-accumulation in the structural solver numerics to provide order of magnitude reduction in the adjoint structural solver memory footprint.

Using the fact that only the RHS of linear systems is differentiated, we can use pre-accumulation on the nodal stress residual computed in CFEAElasticity::Compute_Tangent_Matrix and CFEAElasticity::Compute_NodalStress_Term.
The implementation accounts for material properties and nodal coordinates, the latter are not strictly required at the moment but this future proofs the implementation for structural shape derivatives. 

The above routines where chosen because "everything" happens inside them: computation of element gradients, constitutive terms, stress tensors, etc. So it is where the technique would be most effective.

## Related Work

## PR Checklist


pcarruscag:
Test case discadj_fea is failing due to a change of residuals (the derivatives are the same up to the displayed precision), the change in residuals is smaller than that caused by running on 2 cores rather than serial. Further investigation needed nevertheless.

pcarruscag:
I updated the reference values.
Running the case for more iterations and with tighter tolerances on the linear solver, and comparing preaccumulation on vs off, in serial and parallel, the conclusion is that the derivatives are the same to the 10th decimal place. Which is of the order of the residual reduction, and of the order of the variation between consecutive iterations at convergence.
The data on which I base the conclusion -> [investigation.zip](https://github.com/su2code/SU2/files/2446386/investigation.zip)

talbring:
@pcarruscag To me it looks ok!

rsanfer:
Thanks @pcarruscag! LGTM. I had a go at the results that you shared and as you mentioned the changes in the sensitivities are in the order of the convergence level. Happy to get this merged in.

pcarruscag:
Great! In it goes then.

PR #595, opened 2018-10-06
Structural topology optimization

pcarruscag:
## Proposed Changes
Plain vanilla approach to topology optimization, element-wise constant density parameterization and SIMP interpolation scheme. It is working nicely for 2D cases of moderate size (10 to 100k elements), over that the linear solvers will struggle.
Practical issues with the feature:
- Requires ad-hoc scripting to use with an optimizer, so far the L-BFGS-B from Scipy together with a penalty formulation has served me well.
Technical issues with the implementation:
- Scalability. A filter is applied to the densities to avoid numerical issues such as hinges, and checkerboards, and to ensure minimum feature size and mesh independence. My implementation, although parallel, requires all ranks to get a copy of global information, namely the connectivity and the element centroid coordinates. The ranks get this in compact format, so it is not super expensive (100MB per rank per million hexahedrons) and definitely not an issue for the problem size the linear solvers and preconditioners can currently handle.
- Output. Currently this is done at every iteration of the adjoint solver into a file (only the master rank writes the file) mainly because I did not see an easy way to output element values using the output class, everything else is computed at the nodes.

So not perfect but a nice feature to have (I think). Feedback is welcome on any of these issues (or others you see!). I added an example case to explain the new options, the large files will be in the Testcases repo.

## PR Checklist


Common/include/gauss_structure.hpp
```diff
@@ -160,6 +162,18 @@ class CElementProperty {
   unsigned long GetElectric_Prop(void);
 
   unsigned long GetDV(void);
+  
+  void SetDesignDensity(su2double valDensity);
+  
+  su2double GetDesignDensity(void);
+  
+  void SetPhysicalDensity(su2double valDensity);
```
economon: Please add comments to the class methods.

Common/include/geometry_structure.hpp
```diff
@@ -1535,6 +1535,44 @@ class CGeometry {
...
+                               const su2double *input_values, su2double *output_values) const;
+  
+  /*!
+   * \brief Build the global (entire mesh!) adjacency matrix for the elements in compressed format.
+   *        Used by FilterValuesAtElementCG to search for geometrically close neighbours.
+   * \param[out] neighbour_start - i'th position stores the start position in "neighbour_idx" for the immediate
+   *             neighbours of global element "i". Size nElemDomain+1
+   * \param[out] neighbour_idx - Global index of the neighbours, mush be NULL on entry and free'd by calling function.
+   */
+  void GetGlobalElementAdjacencyMatrix(vector<unsigned long> &neighbour_start, long *&neighbour_idx) const;
```
economon: We have to build the adjacency info for all points (not elems) during preprocessing that we then send to ParMETIS for graph partitioning. This is done in parallel and then the missing info is communicated around to the other procs. I know it is different, but perhaps the code there can give you some ideas for how you can adapt for element adjacency and parallelize/communicate. Just in case it is useful, see the routines related to Partition/Distribute/Load in geometry_structure.cpp, starting with CPhysicalGeometry::DistributeColoring(). If the filters don't change during runtime, at least you will only need to do this once at the beginning.

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -997,6 +1001,9 @@ void CFEASolver::Set_ElementProperties(CGeometry *geometry, CConfig *config) {
     if (rank == MASTER_NODE){
       cout << "There is no element-based properties file." << endl;
       cout << "The structural domain has uniform properties." << endl;
+      
+      if (topology_mode)
+        SU2_MPI::Error("Topology mode requires an element-based properties file.",CURRENT_FUNCTION);
```
economon: Are the property files in ASCII? If they are becoming a bottleneck, you can switch to binary and use MPI IO like we do for the restarts. I can help you with this, and there are some examples you can follow from the restart writer/readers.

pcarruscag:
Thanks for the reviews guys.
@economon I added comments to the area of the code you highlighted.
The file IO is not a bottleneck at the moment, and that filtering step is only done once.
I did have a look at anything with "adjacency" in the name but I took the easy way out (shamelessly) as those routines looked quite intimidating.
If you have plans to make the output structure capable of handling element values I will piggy-back/collaborate on that and output the derivatives via the output instead of writing a dedicated file.
Since we have two approvals, I will turn the example case I've included into a regression and then this can be merged.

economon:
@pcarruscag: is this all ready now that the regression is in place? We can merge in if so.

Also, you mentioned that ad-hoc scripting is needed to run optimization. Is that script provided anywhere in this PR so that folks can try out your implementation?

pcarruscag:
@economon, yeah it is ready. The regression is a file diff that has 10k lines (only 100kB though), is it ok to have the reference file in the code repo or should I move it to testcases?
No that script is not in the PR, I can try to put a minimal version together.

economon:
@pcarruscag : great! I think that is ok to have that file in the repo. If you can put together a very minimal version of the script and add to the PR in SU2_PY/ when you have time, we will merge right after that.

Issue #596, opened 2018-10-11
FSI fluid traction transfer


pcarruscag:
Currently we transfer forces from the fluid side to the structural side, this is fine for matching meshes and for non-matching (interpolated) if the conservative approach is used (i.e. transpose the displacement interpolation matrix to transfer the loads).

For non-matching and the consistent approach (i.e. use the same interpolation scheme both ways) this creates "extra force", to avoid that, stress needs to be transferred instead of force and integrated on the structural side.

pcarruscag:
Fixed in #597

PR #597, opened 2018-10-11
Fix   interpolation

pcarruscag:
## Proposed Changes
1 - Generalize the conservative interpolation approach for the following interpolation schemes:
NEAREST_NEIGHBOR, ISOPARAMETRIC, and RADIAL_BASIS_FUNCTION
The therefore redundant CONSISTCONSERVE option can be recovered with:
KIND_INTERPOLATION= ISOPARAMETRIC
CONSERVATIVE_INTERPOLATION= YES
If you guys think the conservative option also makes sense for the WEIGHTED_AVERAGE (sliding mesh) scheme I will put that in too (I am not familiar so left it alone).

2 - In CTransfer_FlowTraction, if not matching mesh and not conservative interpolation:
  Compute the donor variables as stresses instead of forces;
  Integrate the target variables on the structural side before transferring them.

3 - Removed a bit of code duplication in transfer_physics.cpp while I was there.

I have yet to test this for the FSI adjoint but for the direct problem it looks ok.

## Related Work
#596 

## PR Checklist

### DIFF
--- a/Common/include/option_structure.hpp
+++ b/Common/include/option_structure.hpp
@@ -311,8 +311,7 @@ static const map<string, ENUM_MAT_COMPRESS> MatComp_Map = CCreateMap<string, ENU
  */
 enum ENUM_INTERPOLATOR {
   NEAREST_NEIGHBOR 	= 0,   	/*!< \brief Nearest Neigbhor interpolation */
-  ISOPARAMETRIC 	= 1,	/*!< \brief Isoparametric interpolation */
-  CONSISTCONSERVE 	= 2,	/*!< \brief Consistent & Conservative interpolation (S.A. Brown 1997). Utilizes Isoparametric interpolation. */
+  ISOPARAMETRIC 	= 1,	/*!< \brief Isoparametric interpolation, use CONSERVATIVE_INTERPOLATION=YES for conservative interpolation (S.A. Brown 1997).*/
   WEIGHTED_AVERAGE  = 3, 	/*!< \brief Sliding Mesh Approach E. Rinaldi 2015 */
   RADIAL_BASIS_FUNCTION= 4, /*!< \brief Radial basis function interpolation. */
 };
@@ -320,7 +319,6 @@ enum ENUM_INTERPOLATOR {
 static const map<string, ENUM_INTERPOLATOR> Interpolator_Map = CCreateMap<string, ENUM_INTERPOLATOR>
 ("NEAREST_NEIGHBOR", NEAREST_NEIGHBOR)
 ("ISOPARAMETRIC",    ISOPARAMETRIC)
-("CONSISTCONSERVE",  CONSISTCONSERVE)
 ("WEIGHTED_AVERAGE", WEIGHTED_AVERAGE)
 ("RADIAL_BASIS_FUNCTION", RADIAL_BASIS_FUNCTION);
 
--- a/Common/src/config_structure.cpp
+++ b/Common/src/config_structure.cpp
@@ -1905,7 +1905,7 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   
   /*  DESCRIPTION: Use conservative approach for interpolating between meshes.
   *  Options: NO, YES \ingroup Config */
-  addBoolOption("CONSERVATIVE_INTERPOLATION", ConservativeInterpolation, true);
+  addBoolOption("CONSERVATIVE_INTERPOLATION", ConservativeInterpolation, false);
 
   /*!\par KIND_INTERPOLATION \n
    * DESCRIPTION: Type of interpolation to use for multi-zone problems. \n OPTIONS: see \link Interpolator_Map \endlink
--- a/SU2_CFD/src/driver_structure.cpp
+++ b/SU2_CFD/src/driver_structure.cpp
@@ -2991,18 +2991,35 @@ void CDriver::Interface_Preprocessing() {
         }
         /*--- Else: interpolate ---*/
         else {
-        switch (config_container[donorZone]->GetKindInterpolation()) {
+          bool conservative_interp = config_container[donorZone]->GetConservativeInterpolation();
+          
+          /*--- Conditions for conservative interpolation are not met, we cannot fallback on the consistent approach
+                because CTransfer_FlowTraction relies on the information in config to be correct. ---*/
+          if ( conservative_interp && targetZone == 0 && structural_target )
+            SU2_MPI::Error("Conservative interpolation assumes the structural model mesh is evaluated second, somehow this has not happened.",CURRENT_FUNCTION);
+        
+          switch (config_container[donorZone]->GetKindInterpolation()) {
 
           case NEAREST_NEIGHBOR:
-            interpolator_container[donorZone][targetZone] = new CNearestNeighbor(geometry_container, config_container, donorZone, targetZone);
-            if (rank == MASTER_NODE) cout << "using a nearest-neighbor approach." << endl;
-
+            if ( conservative_interp && targetZone > 0 && structural_target ) {
+              interpolator_container[donorZone][targetZone] = new CMirror(geometry_container, config_container, donorZone, targetZone);
+              if (rank == MASTER_NODE) cout << "using a mirror approach: matching coefficients from opposite mesh." << endl;
+            }
+            else {
+              interpolator_container[donorZone][targetZone] = new CNearestNeighbor(geometry_container, config_container, donorZone, targetZone);
+              if (rank == MASTER_NODE) cout << "using a nearest-neighbor approach." << endl;
+            }
             break;
 
           case ISOPARAMETRIC:
-            interpolator_container[donorZone][targetZone] = new CIsoparametric(geometry_container, config_container, donorZone, targetZone);
-            if (rank == MASTER_NODE) cout << "using an isoparametric approach." << endl;
-
+            if ( conservative_interp && targetZone > 0 && structural_target ) {
+              interpolator_container[donorZone][targetZone] = new CMirror(geometry_container, config_container, donorZone, targetZone);
+              if (rank == MASTER_NODE) cout << "using a mirror approach: matching coefficients from opposite mesh." << endl;
+            }
+            else {
+              interpolator_container[donorZone][targetZone] = new CIsoparametric(geometry_container, config_container, donorZone, targetZone);
+              if (rank == MASTER_NODE) cout << "using an isoparametric approach." << endl;
+            }
             break;
 
           case WEIGHTED_AVERAGE:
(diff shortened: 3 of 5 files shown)


pcarruscag:
(I seem to have a knack to find cans of worms).
What I did in a55ca26 is not accurate, the integral of the fluid stress needs to be done in current coordinates and the normal available trough CVertex is in reference (undeformed) coordinates.

I've implemented this in the structural solver as a preprocessing step. For each surface element of all FSI markers an average stress is computed based on its nodal values, it is then integrated in current (deformed) coordinates. A second pass over the elements scatters the force on the element to the nodes, this is to account for nodes that participate in multiple markers.
The mechanism by which the structural solver takes the flow tractions and updates the Residual vector was not changed, so this changes are fairly local.
The only downside I see is that knowledge of whether the interpolation is consistent or conservative is now required in two locations...

Verification of the FSI adjoint is still ongoing.

pcarruscag:
I have tested the adjoint FSI solver, results from matching, conservative, and consistent interpolation all agree. I am attaching only the bare minimum [files](https://github.com/su2code/SU2/files/2475798/testing.zip): meshes, configs, histories and logs. As to avoid uploading large files, I can obviously share the solutions.
To me the original issue is fixed.

pcarruscag:
Hi Ruben, no worries.

Integrating on the fluid side is what makes sense for matching meshes, less hassle given the availability of dual grid information, no need for MPI boiler plate, etc.

The adjoint implications are being accounted for, this integration step is being recorded and overall the dependencies have not changed, the fluid nodes are still the ones defining the direction of the stress.
As you can see from the files I attached, the sensitivities are very similar for conservative/consistent interpolation cases. 

It is not possible to generalise this for conservative interpolation (i.e. using the transpose of the displacement interpolation matrix to interpolate tractions) because that approach implicitly takes care of the (possible) different areas associated with fluid and solid nodes. If you transfer stresses with conservative interpolation and then integrate you get the wrong force. The conservative approach requires forces to be transferred, otherwise we start taking liberties with the meaning of the word :)

Cheers,
Pedro

pcarruscag:
@rsanfer I am attaching [some results](https://github.com/su2code/SU2/files/2612501/FFD_verification.pdf) for FFD derivatives for FSI cases (that made use of this fix) to rekindle the discussion.

I am taking this directly from my early stage so apologies if not all details are clear, the conclusion is that the fix does not break the adjoint.

SU2_CFD/src/transfer_physics.cpp
```diff
@@ -130,17 +138,25 @@ void CTransfer_FlowTraction::GetDonor_Variable(CSolver *flow_solution, CGeometry
...
+  su2double area = 0.0;
 
   su2double Pinf = flow_solution->GetPressure_Inf();
 
   Point_Flow = flow_geometry->vertex[Marker_Flow][Vertex_Flow]->GetNode();
   // Get the normal at the vertex: this normal goes inside the fluid domain.
   Normal_Flow = flow_geometry->vertex[Marker_Flow][Vertex_Flow]->GetNormal();
+  
+  if (consistent_interpolation)
+    for (iVar = 0; iVar < nVar; ++iVar) area += Normal_Flow[iVar]*Normal_Flow[iVar];
```
LaSerpe: Should be nDim in place of nVar, should't it?

SU2_CFD/src/transfer_physics.cpp
```diff
@@ -73,12 +69,24 @@ void CTransfer_FlowTraction::GetPhysical_Constants(CSolver *flow_solution, CSolv
 
   Velocity2_Real = 0.0;
   Velocity2_ND   = 0.0;
-  for (iVar = 0; iVar < nVar; iVar++) {
+  for (unsigned short iVar = 0; iVar < nVar; iVar++) {
```
LaSerpe: Should be nDim in place of nVar, should't it?

SU2_CFD/src/driver_structure.cpp
```diff
@@ -3414,18 +3414,35 @@ void CDriver::Interface_Preprocessing() {
...
+          if ( conservative_interp && targetZone == 0 && structural_target )
+            SU2_MPI::Error("Conservative interpolation assumes the structural model mesh is evaluated second, somehow this has not happened.",CURRENT_FUNCTION);
+        
+          switch (config_container[donorZone]->GetKindInterpolation()) {
 
           case NEAREST_NEIGHBOR:
-            interpolator_container[donorZone][targetZone] = new CNearestNeighbor(geometry_container, config_container, donorZone, targetZone);
-            if (rank == MASTER_NODE) cout << "using a nearest-neighbor approach." << endl;
-
+            if ( conservative_interp && targetZone > 0 && structural_target ) {
```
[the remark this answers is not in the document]
pcarruscag: Hi @LaSerpe , thank you for your comments.

Hence my first question, do you agree then that it makes sense to generalize the conservative interpolation for all interpolation types?

Common/include/option_structure.hpp
```diff
@@ -341,16 +341,14 @@ static const map<string, ENUM_MAT_COMPRESS> MatComp_Map = CCreateMap<string, ENU
  */
 enum ENUM_INTERPOLATOR {
   NEAREST_NEIGHBOR 	= 0,   	/*!< \brief Nearest Neigbhor interpolation */
-  ISOPARAMETRIC 	= 1,	/*!< \brief Isoparametric interpolation */
```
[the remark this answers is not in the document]
pcarruscag: Ok, I did not know what was the policy in these cases.

SU2_CFD/src/transfer_physics.cpp
pcarruscag: The whole CTransfer family works like that I think, nVar gets the value of nDim in the constructor I believe. I kept the convention.

SU2_CFD/src/driver_structure.cpp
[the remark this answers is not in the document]
LaSerpe: Well, thing is that the mirror class is instantiated only in case a structural target is involved and conservative interpolation is true.
The code looks more like an override to the instantiation of the other interpolation classes than a generalisation.
I am not very familiar with fsi simulations but I do not see any reason why the user could specify an interpolation method which is then override.
Perhaps the instantiation of a CMirror class could remain an option of the switch statement?

Giulio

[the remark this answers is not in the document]
LaSerpe: Ok, then I think it won't have sense to generalise the conservative interpolation to the weighted average.
Weighted average computes the numerical fluxes between nodes and scales them by the fraction of the interface area overlapped by the neighbouring cell.
The current implementation just can't work if one of the two zones is a structure, they both have to be fluid zones.

Common/include/option_structure.hpp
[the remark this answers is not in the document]
rsanfer: The numbering should not be necessary, and it's normally a source of conflicts during merging processes. I would be keen to remove the numbers, then the enumerated variables would be just sequential. What do you guys think @pcarruscag @LaSerpe @economon @talbring ?

TestCases/fea_fsi/WallChannel_2d/configFSI.cfg
```diff
@@ -22,6 +22,8 @@ RESTART_ITER = 2
 
 OUTER_ITER = 5
 
+MATCHING_MESH = YES
```
rsanfer: This might be the reason why tests are not passing.

rsanfer: Hi @pcarruscag 
ok, then, why is it necessary to specify MATCHING_MESH in the config?

Also, given that for the test cases the meshes are matching, I don't understand why the results should change (independently of the mesh being coarser or finer). It should not matter to do either
- Compute stresses -> integrate forces -> transfer data -> apply loads
- Compute stresses -> transfer data -> integrate forces -> apply loads

as in both sides the areas are exactly the same.

pcarruscag: It does not matter at convergence, complementing what I wrote on Oct 12:
On the fluid side the area vector is available at the node.
On the structural side I am first integrating nodal stresses which results in a force at face centers, then those need to be put back on the nodes. Which means, that for a surface mesh of quads, each nodal force depends on 9 nodal stresses instead of 1.

pcarruscag: For clarity, I mean grid convergence.

rsanfer: I understand what you mean. However, the tests were originally passing with these changes, and the same mesh and settings, see, e.g., here:
https://travis-ci.org/su2code/SU2/builds/460157253
and then they failed in the next build:
https://travis-ci.org/su2code/SU2/builds/461388095
which only included the commit
https://github.com/su2code/SU2/commit/b4638d639c3a93a95b4b1d8647a40c6c36b64a39
into the develop after the new Driver structure was merged in.

Provided that the new driver is recovering the same behaviour as before (and that's the case, because the tests passed), I still don't understand what's the fundamental change that's made the regression tests fail now and wasn't making them fail before. There should be possible to make them pass without any modification of the test values.

pcarruscag: They were passing before the changes to the driver, it started failing here:
https://travis-ci.org/su2code/SU2/builds/460835123
Because SU2 is using nearest neighbour consistent interpolation and the option matching mesh is not taking effect on the multiple config cases.
You can see that before b4638d6 two cases were failing (one adjoint one direct) and after I put the matching mesh option in the configs the adjoint passed (single config) while the other did not.

rsanfer: I see. Doesn't that mean though that setting
`CONSERVATIVE_INTERPOLATION = YES`
should do the trick? It should probably be the default anyway.

rsanfer: I totally agree with that, it's also on my list to remove some options that might be unnecessary and/or duplicate, this is a good example. Could you remove `MATCHING_MESH=YES` in the FSI regression tests by making CONSERVATIVE_INTERPOLATION=YES as the default? They should pass like that. Maybe adding a consistent, non-matching simple FSI test case (direct and adjoint) would be useful for completeness. I'll also have a look as soon as I have a minute to other test cases to remove that keyword, and maybe some other confusing ones in a separate PR.

pcarruscag: @rsanfer I made the changes to the settings and if you agree I changed the airfoil RBF testcase to consistent interpolation.
I will open a PR in the TestCase repo to update the initial conditions for the case.

Common/src/config_structure.cpp
```diff
@@ -2130,7 +2130,7 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   
   /*  DESCRIPTION: Use conservative approach for interpolating between meshes.
   *  Options: NO, YES \ingroup Config */
-  addBoolOption("CONSERVATIVE_INTERPOLATION", ConservativeInterpolation, true);
+  addBoolOption("CONSERVATIVE_INTERPOLATION", ConservativeInterpolation, false);
```
rsanfer: I would still make `true` the default here.

TestCases/fea_fsi/WallChannel_2d/configFSI.cfg
rsanfer: I think it's a good idea to use the airfoil RBF as testcase for consistent interpolation.
However, the original test case keeps failing, and with conservative interpolation enabled and a matching mesh, that should not be the case. Could you check it out?

pcarruscag: I did not put the option in all the configs, still getting to grips with the multi-config process. In any case the change to the default value should do the trick.
I merged the TestCases PR so the RBF case should pass too.
Please advise on the enumerator subject as this should be good to go.

PR #603, opened 2018-10-31
avoid calls to blas routines when the data type is not compatible

pcarruscag:
## Proposed Changes
Avoid calls to external functions when compiling in AD and the datatype becomes incompatible.
(The HAVE_LAPACK flag is compatible with AD in the RBF interpolation but not in the HOM feature).
@vdweide please have a look at this.

## PR Checklist

- [X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).

### DIFF
--- a/Common/include/blas_structure.hpp
+++ b/Common/include/blas_structure.hpp
@@ -101,7 +101,7 @@ public:
 
 private:
 
-#if !defined(HAVE_LIBXSMM) && !defined(HAVE_BLAS) && !defined(HAVE_MKL)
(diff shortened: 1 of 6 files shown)


vdweide:
Yep, this is not tested in the regression tests, so it slipped through.

pcarruscag:
@vdweide ok I think I safeguarded the required areas, if you could test this it would be great as I do not have these libraries.

vdweide:
@pcarruscag, everything looks to be working as it should. As soon as the last regression test goes through (it failed due to a connection time out), it can be merged in.

Issue #609, opened 2018-11-14
Sign Error in `GetViscousProjJacs`


clarkpede:
I believe there's a sign error in how the Jacobian of the viscous fluxes is computed for ideal gases.  If you're not familiar with the code, there are two variants: one for ideal gases, and one for arbitrary equations of state. The code for ideal gases is not consistent with the code for arbitrary gas laws.

In ``CNumerics::GetViscousProjJacs`` for the ideal gas law you'll find:
 
    val_Proj_Jac_Tensor_i[4][0] += factor*proj_viscousflux_vel;
    val_Proj_Jac_Tensor_j[4][0] += factor*proj_viscousflux_vel;

In ``CNumerics::GetViscousProjJacs`` for arbitrary gas laws you'll find:

    val_Proj_Jac_Tensor_i[4][0] -= factor*proj_viscousflux_vel;
    val_Proj_Jac_Tensor_j[4][0] -= factor*proj_viscousflux_vel;

The term is added for ideal gases, but subtracted for non-ideal gases.  After deriving the Jacobian myself and having a colleague check my work, I believe that the form for arbitrary gas laws is correct.  The ideal-gas law variant seems to be incorrect.

Moreover, the code in `GetViscousProjJacs` could use some cleanup.  I propose both correcting the sign error and cleaning up the code to be readable.

hlkline:
This sounds like an important thing to fix, and it looks like you've done some serious work to confirm it. 
Would you mind putting your changes into a pull request?

Since the ideal gas version has been around for longer, I would initially suspect that there is another sign flip somewhere else in the code for some reason (which may be hard to find... especially if the code needs some cleanup). Just in case, I suggest running a verification case with both versions, using the ideal gas law as the 'arbitrary' equation of state, and comparing against the built-in ideal gas, under both versions of the code, in order to confirm that the result changes to be the same when the sign is changed as you suggest. Ideally, something that can be compared to results from another code or to experiment to confirm that the sign-corrected version is correct, especially since I'm guessing this will effect a lot of regression tests.

Code cleanup is of course always appreciated as well.

@salvovitale & @LaSerpe since it looks like you've worked on the real gas version: any comments, and do you have a viscous test case?

clarkpede:
[...]

PR #612, opened 2018-11-16
Fix sign error in viscous Jacobian

clarkpede:
## Proposed Changes

There seems to be a sign error in the Jacobian calculation for the viscous flux of ideal gases.

This has been checked in two ways.  First, the Jacobian was derived by hand and compared to the actual code.  Second, the code has been compared with the code for arbitrary equations of state.  While the ideal gas law variant adds the inner product of the viscous flux and the velocity, the general gas law variant subtracts this term.

## Related Work

This error has been previously discussed in issue #609.

## PR Checklist


hlkline:
From a quick skim it looks like it's failing due to small differences in residual values, which is not unexpected here.
Those can be updated in the regression test scripts in SU2/TestCases/ in order to pass the tests, otherwise LGTM

PR #614, opened 2018-11-22
Feature buffet

bmunguia:
## Proposed Changes
Add the Kenway-Martins separation sensor for buffet-onset detection (see [doi.org/10.2514/1.J055172](https://doi.org/10.2514/1.J055172) for details). Config options include:

1. BUFFET_MONITORING= YES will compute
a. The sensor value (between 0 and 1) on NS marker vertices, stored in CNSSolver as Buffet_Sensor
b. The integral of the sensor values normalized by reference area (percent of reference area over which separation occurs), stored in CNSSolver as Buffet_Metric (per-surface) and Total_Buffet_Metric
2. BUFFET_K and BUFFET_LAMBDA specify the sharpness and offset of the Heaviside function used to compute the nodal sensor value
3. Objective function option BUFFET, which uses the integrated metric as a cost function
 

## Related Work
N/A

## PR Checklist


pcarruscag:
Hi @bmunguia,
Overall this looks good, clean implementation and consistent with the other functions.
My only note is regarding "Evaluate_ObjFunc", it looks like a copy paste from the EulerSolver with the buffet function added, could you not have modified just the Evaluate_ObjFunc method of the EulerSolver?
Cheers,
Pedro

bmunguia:
Thanks for the feedback @pcarruscag. @jayantmukho and I discussed this when we implemented the objective function. The main motivation was that since our implementation of the separation sensor is computed using the skin friction coefficient, we only defined the variables Buffet_Sensor, Buffet_Metric, etc. for the NSSolver class. Since the NSSolver needs to be able to evaluate all the same objectives as the EulerSolver, but the buffet objective is exclusive to the NSSolver, I copy pasted the function and added the new objective.

Let me know if you have any suggestions for a cleaner implementation. We were basically choosing between defining all the Buffet_* variables in EulerSolver which would just add unnecessary variables to the constantly growing list of variables, or copying EvaluateObj_Func to allow for NS-specific objectives.

pcarruscag:
If you want to keep what is of the NS solver in the NS solver (which I praise) I would implement NSSolver::EvaluateObj_Func as a call to what is common to Euler and NS (implemented in Euler) plus whatever NS can do extra.
I think you can actually call the method of a parent class from a derived one by disambiguating the call.
If that leads to unclear implementation create a new method in Euler that both EvaluateObj_Func's call but the one in NS adds special sauce to it.

bmunguia:
Good call, that's definitely a lot cleaner. Just made the changes.

EduardoMolina:
Hi @bmunguia and @jayantmukho, 
Thanks for the implementation. This feature will help a lot in the design of more realistic transonic aircrafts. 
LGTM! It is ready to merge.

PR #615, opened 2018-11-25
Fix spurious values in FSI adjoint

pcarruscag:
## Proposed Changes
This is one I have been trying to solve for a while and now found a solution that is not a hack.
 Under some circumstances that I do not understand (very case dependent) one of the cross terms (structure w.r.t. fluid) creates spurious values, e.g.:
image
The fix I found is to change the recording strategy, instead of making 3 recordings of the structural iteration (one for each set of input variables) this is merged into a single recording.
So from this:
image
To this:
image
So far this has fixed the issue for a range of cases, large to small.
[Here](https://github.com/su2code/SU2/files/2613239/spurious_values.zip) you can see the difference between running the case in the example picture with and without the fix.
 

## PR Checklist


SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2421,7 +2421,7 @@ void CDiscAdjFluidIteration::RegisterInput(CSolver *****solver_container, CGeome
 
     solver_container[iZone][iInst][MESH_0][ADJFLOW_SOL]->RegisterSolution(geometry_container[iZone][iInst][MESH_0], config_container[iZone]);
 
-    if (turbulent){
+    if ((Kind_Solver == DISC_ADJ_RANS) && !frozen_visc){
```
rsanfer: In driver_structure.cpp you replaced all the Kind_Solver checks by a turbulent boolean, why is the opposite case here?

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -3023,27 +3023,16 @@ void CDiscAdjFEAIteration::SetRecording(COutput *output,
...
-
-    /*--- Register only structural displacements as input ---*/
-
-    solver_container[iZone][iInst][MESH_0][ADJFEA_SOL]->RegisterSolution(geometry_container[iZone][iInst][MESH_0], config_container[iZone]);
-
-  }
+  solver_container[iZone][iInst][MESH_0][ADJFEA_SOL]->RegisterVariables(geometry_container[iZone][iInst][MESH_0], config_container[iZone]);
 
+  /*--- Both need to be registered regardless of kind_recording for structural shape derivatives to work properly.
+        Otherwise, the code simply diverges as the FEM_CROSS_TERM_GEOMETRY breaks! (no idea why) for this term we register but do not extract! ---*/
```
rsanfer: Do you have any new idea of why is this happening? It should not be the case, all the variables registered should be extracted. What happens if you add 
`solver_container[ZONE_STRUCT][INST_0][MESH_0][ADJFEA_SOL]->ExtractAdjoint_Variables()`
in 
`CDiscAdjFSIDriver::ExtractAdjoint()`
under
`(kind_recording == FEM_CROSS_TERM_GEOMETRY)`
does it still diverge then?

SU2_CFD/src/driver_structure.cpp
```diff
@@ -6260,21 +6260,54 @@ void CDiscAdjFSIDriver::Run( ) {
...
   Preprocess(ZONE_FLOW, ZONE_STRUCT, ALL_VARIABLES);
 
-  switch (Kind_Objective_Function){
-  case FLOW_OBJECTIVE_FUNCTION:
-    Iterate_Block_FlowOF(ZONE_FLOW, ZONE_STRUCT, ALL_VARIABLES);
-    break;
-  case FEM_OBJECTIVE_FUNCTION:
-    Iterate_Block_StructuralOF(ZONE_FLOW, ZONE_STRUCT, ALL_VARIABLES);
-    break;
+
```
rsanfer: This is a good idea, thanks.

SU2_CFD/src/iteration_structure.cpp
pcarruscag: Nope, no idea.
I do not know what would happen but if I extract those values I think they will be wrong, i.e. not the FEM_VARIABLE derivatives we want, due to the variables registered and iteration recorded at the time.

pcarruscag:
Hi @rsanfer,
Regarding patterns, larger cases are more likely to have the issue than small ones, I did not test boundary conditions.
I have not tested the forward mode because it differentiates linear solvers and in the past I have found that to be very limiting so I'd rather not go there. The most detailed verification we have is that present by @cvencro at scitech (finite differences). @cvencro do you have any before/after cases where this problem showed?
On the recording of variables, I like the FEM_VARIABLES as a convergence monitor. But for larger sets of variables, e.g. grid nodes, topology, a "post processing" recording step makes sense, that increase in memory footprint is why I have not created a PR for FSI shape derivatives yet. I talked about this briefly with @oleburghardt during the annual meeting and the new adjoint driver should have hooks for this type of post processing.
Cheers,
Pedro

SU2_CFD/src/iteration_structure.cpp
pcarruscag: I noticed some places where I had forgotten frozen_visc, I'll sift through it to make sure all is consistent, thanks for spotting it.

SU2_CFD/src/iteration_structure.cpp
rsanfer: I agree that they will probably be wrong, but maybe it's worth a try, given that the code diverges if that statements is not there.

pcarruscag:
Hi @rsanfer,
I've updated those booleans.
The results of that verification are [these](https://github.com/su2code/SU2/files/2612501/FFD_verification.pdf), which I had linked to in #597.
I do not think the case will run without this PR, in any case [here](https://github.com/su2code/SU2/files/2792370/results.zip) are the config/mesh and solution files.
Cheers,
Pedro

rsanfer:
LGTM, thanks for updating the booleans. I think you can still add the test case to the PR right?

pcarruscag:
Sure, would you prefer I add the small case that is enough to demonstrate the spurious values, or this larger one I attached last?

SU2_CFD/src/iteration_structure.cpp
oleburghardt: Without having read through the code, here's a guess what might be the reason.
The current AD implementation relies on one very important assumption: Derivatives/adjoints have to be extracted **in the very same order** as they have been registered.
So if you're obtaining wrong derivatives (that cause some divergence) and this can apparently be corrected by changing the registration, that's a big hint into that direction.

Just at a first glance: Could this reasoning apply here? If so, I'll have a more thorough look at it.

pcarruscag: Hi @oleburghardt,
I don't think so. These changes I'm proposing don't change the order of any extraction.
What's more the problem was in the cross terms and:
- These had one set of inputs and one set of outputs, so it would be hard to extract in the incorrect order;
- The spurious values appear in areas of the domain where we know everything has to be zero (those cross-terms only depend on boundary values) and when I tried setting those areas to zero manually I got correct derivatives, i.e. the values that are not spurious are correct and in the correct place.
That being said you are more than welcome to have a look.

oleburghardt:
Hi @pcarruscag , just a small follow-up to my comment in the code section concerning the new adjoint implementation that might also help here (either by using it directly or just getting some ideas).

What I would like to change is in fact not the `CDiscAdjSolver ` (which I find is very good) - but I'd like to propose some changes for the *interface* to the AD tool. And some subsequent stuff.
It will allow for a more direct control of derivatives which is favourable in case one deals with multiple zones, multiple sets of conservative variables, geometry coordinates in case of FSI, and so on. And of course in terms of performance as one needs only one tape that one can keep during for all iterations.

In case you are interested, take a look at the implementation in [su2code:sc_develop](https://github.com/su2code/SU2/tree/sc_develop). I'll open the PR as soon as @talbring opened his one for his great rework on input/ouput.

As for this PR, let me wait a second if we can clear up the issue concerning the registration and I'll leave a small review afterwards.

pcarruscag:
For some reason SU2_SOL doesn't like discrete adjoint cases now... last push will probably fail again

PR #617, opened 2018-11-26
Fix multizone def

cvencro:
## Proposed Changes
Small fixes for FFD deformation of FSI problem using SU2_DEF. 

- SetMesh_Files which writes the output .su2 file was called only for single zone, which has now been changed to the number of zones in the problem

- When writing the output mesh, SetSU2_MeshASCII  closes the file after the first zone but it doesn't get reopened for the second zone. This function is only called by SetMesh_Files in COutput, where the the output file is opened and closed by the master node, therefore it is not necessary to the close the output file inside SetSU2_MeshASCII. 

- When writing the deformed surface for the structural mesh, the same file name as the original was used thus overwriting the original file. '_def' has been appended to the filename similar to the fluid only case.

 

## Related Work

## PR Checklist

- [x ] I am submitting my contribution to the develop branch.


vdweide:
Let's merge it in then.

PR #618, opened 2018-11-28
Fix bug in restart of dynamic meshes with turbulence models

rsanfer:
## Proposed Changes
This PR fixes a bug in the restart of problems which involve dynamic meshes and turbulence.
Given that the Grid Velocities are written in the restart files after the turbulence variables, the indices must me incremented when the restart file is being read in.
At this moment that does not happen, as the variable index is incremented but later reset, and therefore the GridVel container receives incorrect values.

## Related Work
-

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solver_direct_mean.cpp
+++ b/SU2_CFD/src/solver_direct_mean.cpp
@@ -13486,6 +13486,7 @@ void CEulerSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig 
                      (config->GetFSI_Simulation()));
   bool steady_restart = config->GetSteadyRestart();
   bool time_stepping = config->GetUnsteady_Simulation() == TIME_STEPPING;
+  bool turbulent     = (config->GetKind_Solver() == RANS) || (config->GetKind_Solver() == DISC_ADJ_RANS);
 
   string UnstExt, text_line;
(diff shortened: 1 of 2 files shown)


SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -13508,6 +13508,12 @@ void CEulerSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig
   
   unsigned short skipVars = geometry[MESH_0]->GetnDim();
 
+  /*--- Store the number of variables for the turbulence model
+   (that could appear in the restart file before the grid velocities). ---*/
+  unsigned short turbVars = 0;
+  if (turb_model == SA || turb_model == SA_NEG) turbVars=1;
```
economon: Since there are multiple SA variants and only the SST model has 2 eqns., to be safe, I would recommend something like

if(turb){
if(sst) vars = 2;
else vars = 1;
}

Long term, I would like to remove all index jumping as we have here in favor of either a method that accesses the correct index automatically, or accessing it by string tag instead, which we have in the restart file.

talbring: I updated the if statement accordingly. I also agree that we should change that soon by accessing it by strings ...

Issue #620, opened 2018-11-30
su2Deform python 3.7 error


galbramc:
I get the following error from su2Deform when using python 3.7 and SU2 6.1.0:

```
su2Deform( "testing.cfg", args.numberProc) # Run SU2 mesh deformation
  File "$SU2_RUN/mesh_deformation.py", line 83, in mesh_deformation
    info = SU2.run.DEF(config)
  File "$SU2_RUN/SU2/run/interface.py", line 158, in DEF
    konfig.dump(tempname) 
  File "$SU2_RUN/SU2/io/config.py", line 125, in dump
    dump_config(filename,self)
  File "$SU2_RUN/SU2/io/config.py", line 961, in dump_config
    write_config(filename,config)    
  File "$SU2_RUN/SU2/io/config.py", line 775, in write_config
    n_lists = len(new_value)
TypeError: object of type 'map' has no len()
```

Any help fixing this would be greatly appreciated.

galbramc:
@economon it does work with python 2.7, I have not tried multiple versions of python 3, just 3.7. The failure occurs with the python script and does not get to the point where it starts executing SU2_DEG. The minimum example to reproduce the error is:

```
from mesh_deformation import mesh_deformation as su2Deform
su2Deform( "testing.cfg" )
```
using [testing.cfg](https://github.com/su2code/SU2/files/2635991/testing.txt), but with a txt extension.

rsanfer:
Hi all, I am also seeing this error with Python 3.7.2 in the finite_difference.py script. I replaced 'len(new_value)' with 'len(list(new_value))' here and there but then the DV_VALUE_NEW doesn't get correctly set. Any new hints?

petebachant:
@rsanfer can you post the traceback?

talbring:
I fixed that bug in #641. (https://github.com/su2code/SU2/pull/641/commits/c22a5c5cfe7daec4deacbec621d278d26362c5d3)

PR #622, opened 2018-12-01
Add Python 3.7 to Travis config

petebachant:
Adding Python 3.7 to test environments in light of #620.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 dist: trusty
 sudo: required
 
-language: python
+language: c++
 
 compiler:
     - gcc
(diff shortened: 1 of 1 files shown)


petebachant:
Hmmm this didn't work because Travis doesn't have a Python 3.7 binary to download, even though it wouldn't be used. Going to require some more thought if this is worth testing for.

talbring:
We should also think of a smarter way of setting up the travis system. Running 12 independent jobs, each of them having to compile the whole code multiple times, just to test python is a little bit of an overhead.

petebachant:
Not sure if it's preferable, but I changed the environment to use the latest available Python 3, rather than 3.6, and the tests still pass with Python 3.7.

economon:
@petebachant : do you have any ideas about the failure in #620 posted by @galbramc ?

petebachant:
The use of SU2deform in #620 is apparently not part of the test suite, so testing with Python 3.7 doesn't replicate the issue.

petebachant:
Yep, I don't know of any changes to `map` from Python 3.6 to Python 3.7.

PR #623, opened 2018-12-02
Convergence criteria for mesh deformation

economon:
## Proposed Changes
 
This PR changes the convergence criteria for mesh deformation with SU2_DEF. The existing tolerance is related to the smallest cell in the grid, but this results in a criteria that depends on the grid scale and can be hard to tune. Instead, we switch to a standard relative residual reduction criteria for the linear solver during deformation, which brings it in line with other solvers.

The change will break existing config files with 'DEFORM_TOL_FACTOR', so a clear error message has been added informing users of the change.

Changes to the su2code.github.io repo for the tutorial configs need to be merged simultaneously.

## Related Work

N/A

## PR Checklist


pcarruscag:
Thanks @economon this makes a lot of sense. And if it goes ahead maybe the options DEFORM_LINEAR_SOLVER_ITER and DEFORM_LINEAR_ITER should also be fused?

This is has a smallish negative implication to FSI cases. There because the deformation is incremental it is acceptable to converge the solver to 1e-6 or only 1e-5. But in shape optimization (@cvencro and I are working on that) we would be limited by the lowest required tolerance. @rsanfer this is something to consider in the mesh deformation refactoring.

economon:
I think combining options where possible would be great. Also, won’t the move to multiple config files for multizone avoid the linear solver tolerance issue you mention for FSI, as the elasticity config can again reuse Linear_Solver_Error once it is not simultaneously being used by the fluid? Therefore, the deform option can be solely for grid deformation either during shape design or for the mesh equations in FSI problems.

rsanfer:
I agree with @pcarruscag , we should try to merge options. I have been thinking for quite some time that we should make the mesh deformation a full solver on its own and homogenise all the options and procedures, be able to define different boundary conditions, etc. It should be based on the linear elasticity but have it's own variable definition, so it's usable out of the box with the adjoint solver. I have a preliminary implementation in [`feature_mesh_solver` ](https://github.com/su2code/SU2/tree/feature_mesh_solver), but I would need some help with that.

economon:
Thanks @rsanfer. Just a reminder that we need to merge the tutorial update in the website repo too simultaneously if we agree to merge this one.

rsanfer:
I've just opened the PR to su2code.github.io, @economon. Is that the correct one?

economon:
@rsanfer : that is the correct branch with my changes, but we should submit it to the develop branch first. We will push it all to master with v6.2.0

economon:
Thanks all.. time to merge this one. @cvencro @pcarruscag, just let us know if this change causes issues. As @rsanfer says, we can work on the bigger development for v7.

PR #625, opened 2018-12-03
SU2_DOT Interface

economon:
## Proposed Changes
 
Adds an interface to SU2_DOT_AD for volume -> surface sensitivity projection. New option to write out sensitivities at each surface vertex to a generic ASCII file for use with an external parameterization.

Includes some minor code clean up and fixes some compiler warnings.

## Related Work

N/A

## PR Checklist

### DIFF
--- a/Common/src/config_structure.cpp
+++ b/Common/src/config_structure.cpp
@@ -1818,6 +1818,8 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   addDVValueOption("DV_VALUE", nDV_Value, DV_Value, nDV, ParamDV, Design_Variable);
   /* DESCRIPTION: Provide a file of surface positions from an external parameterization. */
   addStringOption("DV_FILENAME", DV_Filename, string("surface_positions.dat"));
+  /* DESCRIPTION: Provide a file of sensitivities from an external adjoint. */
+  addStringOption("DV_SENS_FILENAME", DV_Sens_Filename, string("external_sens.dat"));
 	/* DESCRIPTION: Hold the grid fixed in a region */
   addBoolOption("HOLD_GRID_FIXED", Hold_GridFixed, false);
 	default_grid_fix[0] = -1E15; default_grid_fix[1] = -1E15; default_grid_fix[2] = -1E15;
@@ -5144,6 +5146,7 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
           (Design_Variable[iDV] != SCALE_GRID) &&
           (Design_Variable[iDV] != TRANSLATE_GRID) &&
           (Design_Variable[iDV] != ROTATE_GRID) &&
+          (Design_Variable[iDV] != EXTERNAL_SENSITIVITY) &&
           (Design_Variable[iDV] != SURFACE_FILE)) {
         
         if (iDV == 0)
--- a/Common/src/config_structure.cpp
+++ b/Common/src/config_structure.cpp
@@ -1637,6 +1637,8 @@ void CConfig::SetConfig_Options(unsigned short val_iZone, unsigned short val_nZo
   addBoolOption("WRT_PERFORMANCE", Wrt_Performance, false);
     /* DESCRIPTION: Output a 1D slice of a 2D cartesian solution \ingroup Config*/
   addBoolOption("WRT_SLICE", Wrt_Slice, false);
+  /* DESCRIPTION: Write raw sensitivities (dJ/dx) on surfaces to ASCII file. \ingroup Config*/
+  addBoolOption("WRT_EXTERNAL_SENSITIVITY", Wrt_External_Sensitivity, false);
   /*!\brief MARKER_ANALYZE_AVERAGE
    *  \n DESCRIPTION: Output averaged flow values on specified analyze marker.
    *  Options: AREA, MASSFLUX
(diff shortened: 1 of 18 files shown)


SU2_DOT/src/SU2_DOT.cpp
```diff
@@ -267,9 +267,16 @@ int main(int argc, char *argv[]) {
...
-      if (rank == MASTER_NODE) cout << "Reading volume sensitivities at each node from file." << endl;
+
+      if (rank == MASTER_NODE)
+        cout << "Reading volume sensitivities at each node from file." << endl;
       grid_movement[iZone] = new CVolumetricMovement(geometry_container[iZone][INST_0], config_container[iZone]);
-      geometry_container[iZone][INST_0]->SetSensitivity(config_container[iZone]);
+
+      /*--- Read in sensitivities from file. ---*/
+      if (config_container[ZONE_0]->GetDesign_Variable(0) == EXTERNAL_SENSITIVITY)
+        geometry_container[iZone][INST_0]->ReadExternalSensitivity(config_container[iZone]);
```
oleburghardt: Could you give some explanation where the main difference lies between _ReadExternalSensitivity_ and _SetSensitivity_? From looking through the code, I see that there are some changes concerning the filename, but I guess I'm missing something important here.
Where should I focus on?

oleburghardt: Alright, so it's primarily about the file format, thanks for clarifying.

Still I find the term "external sensitivities" to describe the sensitivity input (in the config file as well as in the code) a bit misleading. It is somehow suggesting that the sensitivities in SU2_CFD's output files were no external sensitivities to SU2_DOT in that regard (which they are).

Something like "UNORDERED_SENS_FILE" would be more appropriate, I think. It's not as elegant, though. Maybe someone comes up with a good name?
I could also think about one myself if we would generally agree on changing it (I'm just proposing it :)).

economon: Yep, basically just another format that SU2_DOT can handle for the same process we already perform with our native ASCII and binary files.

I am open name changes, of course, if there are any other ideas!

oleburghardt: What do you think about "PROJECT_SENSITIVITIES" or "SURFACE_SENSITIVITY"?

Maybe @talbring has an idea, too?
As it's just up to a name, I'd like to give it a small push so that we can merge it in :+1:

economon: @oleburghardt : you are right, this is a subtle but important change. I actually like your first suggestion for 'UNORDERED_SENS_FILE' as opposed to the SU2 native format. I am going to slightly change the interface here too in another commit soon so that it is more clear.

economon: I have adjusted the interface now and think this is even more useful. References to 'external sensitivity' have been removed in favor of reusing the same 'SURFACE_FILE' DV option that already exists for SU2_DEF. 

The envisioned usage is that one can use the DV type 'SURFACE_FILE' to create their own workflow with a custom parameterization outside of the current options in SU2. In this case, the user would export the projected surface sensitivities at each surface node from SU2_DOT, use their own parameterization/optimizer to decide the displacements for each surface node, and then provide those surface node displacements to SU2_DEF for mesh deformation before running the next primal/adjoint.

I have also separated the idea of the unordered sensitivity file. Now, for any parameterization in SU2_DOT, users can decide whether to use the NATIVE_BINARY format for the volume sensitivity reader, or to use the new UNORDERED_ASCII. There is a new option for this, but the default is NATIVE_BINARY.

Thanks for the feedback and let me know what you think, Ole.

Lastly, I have updated the Xcode projects for the other modules (thanks again for the reminder, Jessie!)

SU2_DOT/src/SU2_DOT.cpp
```diff
@@ -267,9 +267,16 @@ int main(int argc, char *argv[]) {
...
     } else {
-      if (rank == MASTER_NODE) cout << "Reading volume sensitivities at each node from file." << endl;
+
+      if (rank == MASTER_NODE)
+        cout << "Reading volume sensitivities at each node from file." << endl;
       grid_movement[iZone] = new CVolumetricMovement(geometry_container[iZone][INST_0], config_container[iZone]);
-      geometry_container[iZone][INST_0]->SetSensitivity(config_container[iZone]);
+
+      /*--- Read in sensitivities from file. ---*/
+      if (config_container[ZONE_0]->GetSensitivity_Format() == UNORDERED_ASCII)
```
oleburghardt: Nice, I think that's the better approach. Thanks for the update :+1:

SU2_CFD/src/output_su2.cpp
```diff
@@ -312,3 +312,77 @@ void COutput::WriteCoordinates_Binary(CConfig *config, CGeometry *geometry, unsi
...
+                                        unsigned short val_iZone,
+                                        unsigned short val_nZone) {
+  
+  unsigned short iVar;
+  unsigned long iPoint, iElem, iNode;
+  unsigned long *LocalIndex = NULL;
+  bool *SurfacePoint = NULL;
+  string filename, fieldname;
+  
+  filename = config->GetDV_Sens_Filename();
```
oleburghardt: Why would want to overwrite the input file? We usually have distinct file names for input and for output (as e.g. for meshes and solutions). Maybe it's more generic if we have it like this here, too?

Common/src/config_structure.cpp
```diff
@@ -4286,6 +4290,14 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
       SU2_MPI::Error("Different number of topology filter kernels and respective radii.", CURRENT_FUNCTION);
     }
   }
+  
+  /*--- If we are executing SU2_DOT in surface file mode, then
+   force the projected surface sensitivity file to be written. ---*/
+  
+  Wrt_Projected_Sensitivity = false;
+  if ((Kind_SU2 == SU2_DOT) && (Design_Variable[0] == SURFACE_FILE)) {
+    Wrt_Projected_Sensitivity = true;
```
oleburghardt: Usually by default, we have a file written out called surface_sens.dat (e.g. Tecplot ASCII) when calling SU2_DOT. Why is this routine needed? Is it also for keeping a certain format? From looking through the code, it is very similar to `SetTecplotASCII`.

SU2_CFD/src/output_su2.cpp
economon: Agreed. I am going to submit one or two small final changes to this PR, and I will update this.

Common/src/config_structure.cpp
economon: Yep, it is for the format. Basically, the visualization files contain other information like headers and the connectivity for the mesh. In the simple ASCII output here, it is simply the x,y,z, and sens values (unordered), which is smaller/easier to handle downstream. Also, depending on whether you write ParaView or Tecplot files for the surface sens, it will also be different and require special handling downstream. Basically, it would be good to have a fixed format for this if it will be relied upon in optimization loops.

oleburghardt: Ok.

SU2_CFD/src/output_su2.cpp
oleburghardt: Nice :)

PR #626, opened 2018-12-03
Refactor viscous numerics

clarkpede:
## Proposed Changes

I have refactored some of the viscous residual calculation.  This should not change any of the underlying numerics.

+ The stress tensor due to molecular viscosity and turbulence is now separated out from the residual calculation.  This removes the implicit assumption that eddy-viscosity models are used.  Now other models, such as explicit algebraic stress models, can be added.
+ The "heat flux vector" was also separated out from the residual calculation.
+ The Jacobian calculation for `CAvgGrad_Flow` has been rewritten so as to be more readable.
+ The three main viscous numerics classes (`CAvgGrad_Flow`, `CGeneralAvgGrad_Flow`, and `CAvgGradInc_Flow`) are now all children of a base class, `CAvgGrad_Base`.
+ Instead of two sets of classes for the viscous residual, normal and "corrected", I made a boolean switch in the constructor.  This eliminates the duplication between normal and "corrected" viscous numerics.

There's definitely some more work that could be done to generalize the base class and optimize the speed, but I'm content with the code for now.

## Related Work

This piggybacks off of PR #612, though the two PRs are separate and distinct.

## PR Checklist


economon:
@clarkpede : I am not too concerned about differences on the order of what you are seeing with the CHT case. It is down in the range of the changes you would see just due to switching hardware / OS / etc. One option is to loosen the tolerance in the python regression script, e.g., increase 'cht_incompressible.tol       = 0.00001' to a higher val so it is not as sensitive.

clarkpede:
So I've changed the CHT test case to have a higher tolerance for errors, and I've merged in the changes from PR #570.  Everything looks ready to go.

On a side note, I've noticed that SU2 has some internal inconsistencies with the viscous/turbulent stresses.  Some new options, such as QCR, SST UQ, and wall functions, have been added to the stress tensor definition over the past two years.  But there's still some parts of the code, such as `CNSSolver::BC_Isothermal_Wall` and `CSourceIncAxisymmetric_Flow::ComputeResidual` where those changes aren't reflected.  The stress tensor is still defined the old way.  Is this intentional? Or was it an oversight?

clarkpede:
> Thanks for pointing that out. I dont think it is intentional. Maybe we can use this PR also to fix this ?

That depends on how atomic we want these PRs to be.  This PR is largely a refactoring PR, and shouldn't change any underlying performance.  But adding extra calculations on source terms and wall conditions might change the performance of the code.  The two changes are completely compatible, so combining them wouldn't be hard.  Nevertheless, I propose we keep these changes separate, so that the purpose of each PR is clear and easy to see.  In my mind, two small PRs are easier to check and review than one large PR.

@talbring @economon What do you think? One big PR or two small PRs?

talbring:
Does it need some updates now that #631 has been merged in ?

clarkpede:
@economon Sure, I can submit another PR to address the code duplication (and possible problems) in the stress tensor definition. My work in hybrid RANS/LES methods requires changing the viscous stress, so I would like to have a good base to work from.

@talbring If I'm not mistaken, #631 addresses only the convective numerics, while this PR addresses only the viscous numerics.  So there doesn't seem to be any overlap.

talbring:
@clarkpede Yep you are right :+1: I like this PR. Removing code by still keeping the same functionality. As soon as the regression tests pass we can merge it in.

Do you think you can submit the other PR some time next week ? So that we have it for the 6.2.0 release ?

PR #628, opened 2018-12-06
Ausmplusup

aeroamit:
## Proposed Changes
AUSM+ -up scheme of Liou (A sequel to AUSM, Part II: AUSM + -up for all speeds) has been implemented . It is an all speed scheme and is more robust and accurate in comparison to the AUSM scheme. Also It has relatively fewer issues in comparison to the Roe scheme.
(I tried it over a relatively complex geometry (unstructured hybrid mesh), implementation looks fine). 

## Related Work
**

## PR Checklist


cjz667:
Hello. Can you add a test case about AUSMPLUSUP or AUSMPLUSUP2? I installed SU2 6.2.0 and used it to simulate two-dimensional lid-driven cavity flow. When I set 'CONV_NUM_METHOD_FOLW' to be 'AUSMPLUSUP' or 'AUSMPLUSUP2', it shows 'Error in ''void CSoLver::SetResidual_RMS(CGeometry*, CConfig*):' and 'MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD'. But when I set it to be other schemes like AUSM, ROE or so, it runs. Maybe if there is a test case, I can find where is my mistake. Thank you.

aeroamit:
Hi, @cjz667

When I tried lid driven cavity case with default setting and switching ROE to AUSMPLUSUP, I got the following error - Error in “void CSyssolve::ModGramSchmidt(int.........
SU2 has diverged (It ran for 2 iterations).

Reduced the CFL to 0.5 (originally 2.0), it runs fine.
But it is going to take more number of iterations.
This issue was noted earlier as well and reason behind that is inconsistent Jacobian (Roe here) with AUSM family of schemes. This restricts the allowable CFL number to a lower values for these scehemes. For low speed flows u may have to use smaller value of CFL with these two schemes. 

Dr. Padro has done some work in this direction and implemented hybrid Jacobian for some of these schemes (available in develop branch) , which improves this situation and u can try higher CFL. I have also worked out analytical Jacobian and will be pushing in future release.

I hope you can go with reasonable CFL for majority of the other cases, where Mach number is not too small.
I have added one Testcase for high speed flow past blunt body in testcase folder /euler/bluntbody, which uses AUSMPLUSUP. Give a try with other cases, I have tried many in RANS, NS, EULER.
Hope this help. 

Regards 
Amit

cjz667:
Hello, @aeroamit 

Thank you very much. I try for some times and find that mach number matters. When I simulate two-dimensional lid-driven cavity flow, the flow starts from static. So this case always fails. I try laminar boundary layer case with 0.1 incoming mach number. This time SU2 runs well with AUSMPLUSUP, but even with a very low CFL number(0.01 or lower), it doesn't converge. I learn that this scheme is perfect for high speed flow, but it may not be good at low mach number case.

Perhaps there are some mistakes when I use it. I think I am not familiar with this scheme enough and that I know SU2 not very well. Before using it in practice, I should read more papers and codes.

By the way, SU2 6.2.0 doesn't have the option 'USE_ACCURATE_FLUX_JACOBIANS'.

Thanks again.

Regards
Cao J. Z.

aeroamit:
Hi @cjz667,
Download and install the develop branch, you can use the  USE_ACCURATE_FLUX_JACOBIANS option. This option was added by @pcarruscag.
It will allow reasonably good CFL.

cjz667:
Hello, @aeroamit 
Thanks a lot. I will have a try.

PR #638, opened 2019-01-29
Sliding mesh for incompressible flows + Cleaning of transfer routines

rsanfer:
## Proposed Changes
The current PR serves two purposes. First, it extends the sliding interface capabilities for multi-zone fluid-fluid interfaces to incompressible flows. Until now, these features were only available for compressible flows. For example, take the following case of a viscous flow enclosed by diathermal walls and discretized using two different zones with a permeable, non-matching interface:

testcase
discretization

With the current SU2 implementation, the solution diverges due to the lack of interface boundary conditions. After only 10 iterations, the contours of velocity are:

former

With this PR, the problem can be successfully solved with a multizone solution process. The contours of velocity and temperature are now:

velocity
temperature

and also for unsteady problems

unsteady

For the latter case I haven’t yet rotated the inner circle, because there are some missing features for dynamic meshes in the incompressible solver. The previous test case for steady-state flows has been added. One for unsteady flows will be enabled once the new output is in place, as the current output structure makes the regression scripts fail.

The previous, however, opened a can of worms, as there were some physics based calls in the parent transfer structure (e.g. `GetnSlidingStates()`, `SetSlidingState()` and  `SetnSlidingStates()`) that made the incompressible CHT cases run into problems. 

So I’ve also taken the chance to move some initialization and some calls to the specific physics-based transfer structures, making use of the polymorphism of the transfer code. This way the parent class remains opaque to the physics. At the same time, I cleaned up some long-time unused routines, made the Broadcast_Data the default for zone interfaces, and removed the MATCHING_MESH option as it is just a particular case of Nearest Neighbour.

## Related Work

- Complements the sliding interface feature by @LaSerpe in #389 by adding the corresponding functions to the incompressible flow solver.

- Removes MATCHING_MESH option as discussed with @pcarruscag in #597.

- Removes some compilation warnings due to unused variables.

## PR Checklist


LaSerpe:
Hi Ruben,
thanks for this contributions.
The implementation seems to be ok at first glance, and also the cleaning up is very nice.
It should pass without any problem, btw I'll have a more accurate look at it tomorrow!

cheers,
Giulio

SU2_CFD/src/driver_structure.cpp
```diff
@@ -7975,173 +7751,40 @@ void CMultiphysicsZonalDriver::DynamicMeshUpdate(unsigned long ExtIter) {
...
-      }
-      else if ((transfer_types[donorZone][targetZone] == NO_TRANSFER)
-               || (transfer_types[donorZone][targetZone] == ZONES_ARE_EQUAL)
-               || (transfer_types[donorZone][targetZone] == NO_COMMON_INTERFACE)) { }
-      else {
-        cout << "WARNING: One of the specified interface transfer routines is not known to the chosen driver and can't be executed." << endl;
-      }
-    }
-    break;
+  if (transfer_types[donorZone][targetZone] == SLIDING_INTERFACE) {
```
LaSerpe: perhaps we could put a switch statement in place of else ifs.

SU2_CFD/src/driver_structure.cpp
```diff
@@ -3413,31 +3413,23 @@ void CDriver::Interface_Preprocessing() {
         /*--- Match Zones ---*/
       if (rank == MASTER_NODE) cout << "Setting coupling ";
 
-        /*--- If the mesh is matching: match points ---*/
-      if ( config_container[donorZone]->GetMatchingMesh() ) {
-        if (rank == MASTER_NODE) 
-            cout << "between matching meshes. " << endl;
-        geometry_container[donorZone][INST_0][MESH_0]->MatchZone(config_container[donorZone], geometry_container[targetZone][INST_0][MESH_0], config_container[targetZone], donorZone, nZone);
```
pcarruscag: You can probably delete the MatchZone method?

SU2_CFD/src/driver_direct_multizone.cpp
```diff
@@ -638,44 +638,44 @@ bool CMultizoneDriver::Transfer_Data(unsigned short donorZone, unsigned short ta
   /*--- Select the transfer method according to the magnitudes being transferred ---*/
 
   if (transfer_types[donorZone][targetZone] == SLIDING_INTERFACE) {
-    transfer_container[donorZone][targetZone]->Broadcast_InterfaceData_Interpolate(solver_container[donorZone][INST_0][MESH_0][FLOW_SOL],solver_container[targetZone][INST_0][MESH_0][FLOW_SOL],
-                                                                                   geometry_container[donorZone][INST_0][MESH_0],geometry_container[targetZone][INST_0][MESH_0],
```
pcarruscag: In the interest of keeping the drivers independent of physics, would it make sense to make "transfer_types" a matrix of a little class (instead of enum) that stores also the donor/target solvers? You could then avoid all these if's.

SU2_CFD/src/driver_structure.cpp
rsanfer: Sure, I forgot to do so. I'll do it in a minute.

SU2_CFD/src/driver_direct_multizone.cpp
rsanfer: I agree this could be done better. I'll look into it soon and put a new PR request, probably with some other changes towards generalisation on the driver that are also needed.

SU2_CFD/src/driver_structure.cpp
rsanfer: As I mentioned below to @pcarruscag, I'll have a look into this and maybe create a small class for the transfer types to avoid so many switch/if statements. I aim to remove this from driver_structure and move it all to driver_direct_multizone.cpp soon, I'll try to include it there.

rsanfer:
Tests are passing, so if there are no further comments I'll merge this one in. I'll further tackle driver generalisation in PRs to come.

PR #647, opened 2019-02-05
Update issue templates

economon:


economon:
This PR adds issue templates from GitHub to our repo so that when folks would like to submit (a) bug reports or (b) feature requests, there will already be a standard template form that can be easily filled out. It is similar to the PR template.

talbring:
Thanks @economon. Good idea. Looks good. We can merge it any time. At some point we have to merge master into develop anyway, since they are out of sync.

economon:
Yep. Let's do it right after this one by merging the commits in master to the open PR for v6.2.0 if possible

PR #641, opened 2019-01-30
Release v6.2.0

economon:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
 
This PR will close the release of v6.2.0. The goal is to tag it within the next week or two once this is merged.

Soon, folks in the community will be submitting PRs for major changes to the I/O, drivers, MPI, and more, as we prepare for a major v7 release (scheduled around the time of the developers meeting in May). This v6.2.0 is a minor/maintenance release that will serve as a good snapshot before the major changes.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

We will merge some of the currently open PRs before the release, especially smaller PRs and clean up PRs. Feel free to comment below if there are any outstanding items that you would like to add to develop via PR before release. For significant changes, please wait for the v7 release.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.

PR #649, opened 2019-02-06
Minor bugfixes

clarkpede:
## Proposed Changes

This is a collection of minor bug-fixes that I've found while working with the codebase.  Each is separated into a separate commit, with a description of the purpose of each bug fix in the commit description.

## Related Work

N/A

## PR Checklist


clarkpede:
If you look at the testing history, commit 4f5f3ed doesn't pass the regression tests, while commit 8551cac does.  Only two tests are failing on 4f5f3ed: `discadj_topol_optim` and `discadj_fsi_airfoil`.

The difference between the two commits is a simple change.  I changed the `SU2_MPI::Error` routine to give a return status of `EXIT_FAILURE` instead of `0`.  Since `0` is conventionally defined as a successful program exit, returning `EXIT_FAILURE` makes more sense than returning `0` when `SU2_MPI::Error` is called.

So why did that change cause the regression tests to fail? My suspicion is that these two tests have been exiting with an error for some time, but these regressions were not picked up by Travis since SU2 kept on returning `0` (i.e. success).  Changing the exit code of `SU2_MPI::Error` doesn't cause any errors, but it will expose errors that are (sometimes silently) occurring.

@pcarruscag I think you added these tests.  Any idea why they're failing?

pcarruscag:
They should be due to SU2_SOL not playing nice with the discrete adjoint for a while.
I think @economon put a fix for that in #641?
@clarkpede can you apply the EXIT_FAILURE change to #641? Those tests should pass there.

clarkpede:
> Thanks for the PR! I just have a single question: Why Set_MPI_MaxLength needed to be called 2 times for periodic boundaries?
> 
> LGTM! If Travis pass, we are ready to merge.
> Eduardo

To be honest, I don't know.  I just know that that:

+ Without calling it twice, the max length is not initialized on some of the periodic halo nodes.
+ When you do call it twice, the max length is initialized on all periodic halo nodes.

Something similar happens when loading a solution from a restart file, where the solver needs to do something like this:

    solver[MESH_0][FLOW_SOL]->Set_MPI_Solution(geometry[MESH_0], config);
    solver[MESH_0][FLOW_SOL]->Set_MPI_Solution(geometry[MESH_0], config);

@economon Do you know why the MPI calls need to be made twice?

economon:
@clarkpede @EduardoMolina : yep, the issue is that the periodic and MPI communications are tangled up together in the current implementation. The main issue occurs when there are periodic points that are also ghost nodes. 

Ideally the periodic and MPI would be separated, and the periodic communications would happen first so that all periodic BCs are synchronized before trying to send across partitions. Right now, the comms are mixed, so sometimes, we send old data at periodic points to their matching pair, which then gets updated in a later MPI comm. Therefore, the second call you have added to the MPI resends the periodic update to with the correct data that was updated with MPI in the first call. I hope this is clear..

Anyway, please go with this change as you have it. All of the MPI and periodic comms will be replaced with a cleaner/separated version right after v6.2.0 (you can see what I am working on in feature_mpi_periodic). I will likely need some help testing that soon :).

EduardoMolina:
Hi @clarkpede and @economon .
Thanks for the clear explanation.
Merging now.

Eduardo

PR #652, opened 2019-02-15
MPI Point-to-Point Refactoring + New Periodic BC

### LINKED ISSUE #252: Order of parallel communication
While working on the turbo features Salvo and I noticed that there is currently a problem with the order of the communication (i.e. the order in which we loop through the SEND_RECEIVE markers.)

In the code we always have loops like this

```
for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {

    if ((config->GetMarker_All_KindBC(iMarker) == SEND_RECEIVE) &&
        (config->GetMarker_All_SendRecv(iMarker) > 0)) {

...
```

This way the send/receive involving the periodic boundaries is always done **before** the send/receive  involving the boundaries related to the parallel partitioning (this is because the periodic structure is created before the partitioning is done).  Hence the wrong values are send to the periodic ghost cells.

An easy solution is to change the order of the loop, i.e.

```
for (iMarker = config->GetnMarker_All() - 1; iMarker >= 0; iMarker--) {

    if ((config->GetMarker_All_KindBC(iMarker) == SEND_RECEIVE) &&
        (config->GetMarker_All_SendRecv(iMarker) > 0)) {

...
```

Since I am not completely sure whether this is a valid solution (I dont know if this gives performance issues or will break something else) I opened this as an issue.

Related to this I would also suggest to maybe implement a general send/recv routine that every class can use, because all these SendReceive_\* and Set_MPI_\* routines do essentially the same at the moment (just with different data).

### LINKED ISSUE #416: Current status of periodic BCs
Dear developers,

I would like to know about the current status of Periodic BCs in SU2? Anyone working on this topic? For instance, it is working for single periodic BCs, for example, spanwise direction of a 3D channel?

Regards,

Eduardo Molina

### LINKED ISSUE #431: Segfault when running problem with periodic boundary conditions

### LINKED ISSUE #550: MPI Failure when Running with 16+ Cores
I've recently run into a problem with periodic geometry when I run a RANS problem on 16 cores or more (256+ MPI tasks).  While initializing the Jacobian structure for the turbulence model, I run into one of two errors, depending on the core count.

The first error results in the following error message:
```
Fatal error in MPI_Sendrecv: Message truncated, error stack:
MPI_Sendrecv(249).................: MPI_Sendrecv(sbuf=0x2ee74f0, scount=10, MPI_DOUBLE, dest=19, stag=0, rbuf=0x2ee68e0, rcount=385, MPI_MPIDI_CH3U_Receive_data_found(144): Message from rank 25 and tag 0 truncated; 3200 bytes received but buffer size is 3080
aborting job
```

The second error just leads to the solver hanging indefinitely at the `Initialize Jacobian structure (SA model)` step.  I'm guessing that an MPI send/receive is left dangling.

I have not seen these problems at lower core counts (2-4 cores with 2-32 MPI tasks).

**The errors seem to be tied to the way the periodic send/receives are set up.** If I change the periodic boundaries to far-field boundaries, the error vanishes.

I've also done a lot of work to weed out possible causes:

+ I've generated the meshes using both `SU2_MSH` and the `su2perio` Fortran tool.
+ I've run this on two different supercomputers, with different MPI builds.
+ I've tested multiple different meshes with different resolutions.
+ I've tried changing the RANS model and steady/unsteady options.
+ I've even used a different solver (our hybrid solver) that's completely independent of the RANS solver classes.  Same error.
+ The problem occurs whether you're restarting or starting without a restart file.

I've got a minimal example that you can use to test this for yourself, in the attached files.  It should be self-explanatory.

[MPI_Failure_Example.tar.gz](https://github.com/su2code/SU2/files/2167185/MPI_Failure_Example.tar.gz)

economon:
## Proposed Changes
 
This is one for v7 and needs some testing. This PR contains a complete refactoring of the point-to-point MPI infrastructure and an entirely new implementation of the periodic BC for the FV solvers.

For the MPI refactoring:
- All calls are abstracted to parent classes (geometry, solver, matrix), meaning we will remove all Set_MPI_* routines.
- Point-to-point comms are all non-blocking calls that have been separated into InitiateComms() & CompleteComms(), so that, in theory, they can be overlapped with computation where possible. It would also be possible now to group the comms into larger messages easily. These optimizations are a next step (not for this PR).
- Persistent communication schedules and memory are used throughout the calculation.
- Results are the same, this is just a refactoring to improve performance, scalability, etc. while making the code much simpler and separating the periodic part.

For the periodic BC:
- No longer need a preprocessing with SU2_MSH - use the original grid and just call SU2_CFD.
- Adjacent periodic faces are now possible, e.g., triply-periodic cube.
- Concept is now based on completing the residuals from partial control volumes on either side of a periodic face instead of creating extra halo cells.
- All periodic comms have been separated from the MPI point-to-point comms and are performed first for consistency.

In a follow-up PR, I will remove many lines of code that are no longer needed (all Set_MPI_* routines, many outdated geometry routines, etc.)

## Related Work

~~I could use some help with testing this one, in particular the rotational periodicity, which might still need some changes. Unfortunately, I do not have the original meshes for the periodic cases in the repo that are failing (other than Poiseuille)..~~

~~@salvovitale @LaSerpe @arubino: do you have the original grids for the periodic cases before adding the halos with SU2_MSH? I would like to make sure that everything is working ok for all previous capability.~~

All periodic meshes in the regressions have been updated for the new periodic BC, and a python script has been added to remove halo cells from old meshes that were preprocessed with SU2_MSH for the previous periodic BC implementation.

Resolves #550 
Resolves #431 
Resolves #416 
Resolves #252 

## PR Checklist


LaSerpe:
Hi Tom,

I just have the mesh for the LS89 test case in the NICFD folder.
(Apparently, I had to change the extension to attach the file here but just revert it to .su2).
Let me know.

Giulio

[mesh_vki_turbine_prepbc.txt](https://github.com/su2code/SU2/files/2871791/mesh_vki_turbine_prepbc.txt)

EduardoMolina:
Hi Tom, 

I will be very happy to review and run a couple of cases! When do you plan to merge this PR? I might need some days.
Thanks.
Eduardo

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -4560,20 +4751,29 @@ void CIncEulerSolver::SetPrimitive_Gradient_LS(CGeometry *geometry, CConfig *con
...
     
-    AD::SetPreaccOut(node[iPoint]->GetGradient_Primitive(), nPrimVarGrad, nDim);
-    AD::EndPreacc();
+    for (iVar = 0; iVar < nPrimVarGrad; iVar++) {
+      for (iDim = 0; iDim < nDim; iDim++) {
+        node[iPoint]->SetGradient_Primitive(iVar, iDim, Cvector[iVar][iDim]);
+      }
+    }
+    
+    //    AD::SetPreaccOut(node[iPoint]->GetGradient_Primitive(), nPrimVarGrad, nDim);
```
pcarruscag: Any issues with the pre accumulation?

economon:
@LaSerpe : thanks for the mesh!

@EduardoMolina : no huge rush on merging this one.. we should make sure everything is working ok. It would be very helpful if you can take a look at the Taylor-Green Vortex case again, like we looked at before, now that the parallelization is improved. The option to run is still there (TGV = YES), and I think you already have the meshes..

@pcarruscag : thanks for the comments (will address soon). Also, thanks for the reminder about the pre-accumulation. I disabled that since I had to rework the least squares routines for the periodic BC and I did not turn it back on. The variables may need to be adjusted now though

TobiKattmann:
Small thing I just noticed: 
In config_structure.cpp l. 439+400 Periodic_Translate = NULL; etc. are set. This exact initialization to NULL is repeated 2 times l. 448-450 and l. 542+543. 
As this code is not part of this PR I couldn't commend directly in the code.

SU2_CFD/src/output_structure.cpp
```diff
@@ -13389,6 +13514,30 @@ void COutput::LoadLocalData_IncFlow(CConfig *config, CGeometry *geometry, CSolve
...
+      nVar_Par += 3;
+      Variable_Names.push_back("error_vel_x");
+      Variable_Names.push_back("error_vel_y");
+      Variable_Names.push_back("error_pressure");
+      
+    }
+    
     if (wrt_cp) {
       nVar_Par += 1;
       Variable_Names.push_back("Specific_Heat");
```
TobiKattmann: This push_back of cp and thermal conductivity should move up 25 lines below the variable_density conditional in order to match with the order below where the values are written.

economon:
> Small thing I just noticed:
> In config_structure.cpp l. 439+400 Periodic_Translate = NULL; etc. are set. This exact initialization to NULL is repeated 2 times l. 448-450 and l. 542+543.
> As this code is not part of this PR I couldn't commend directly in the code.

Thanks @TobiKattmann : fixed.

Common/src/geometry_structure.cpp
```diff
@@ -16064,6 +17403,288 @@ void CPhysicalGeometry::MatchActuator_Disk(CConfig *config) {
...
+          
+          if (node[iPoint]->GetDomain()) {
+            
+            /*--- Coordinates of the boundary point ---*/
+            
+            Coord_i = node[iPoint]->GetCoord(); mindist = 1E6; pProcessor = 0; pPoint = 0;
+            
+            /*--- Loop over all the boundaries to find the pair ---*/
+            for (iProcessor = 0; iProcessor < nProcessor; iProcessor++)
+              for (jVertex = 0; jVertex < Buffer_Receive_nVertex[iProcessor]; jVertex++) {
```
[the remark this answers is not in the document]
economon: I cleaned up this entire routine and added better comments. The cost is not a bottleneck for the foreseeable future. The data gathered for the periodic markers on each rank is typically very small relative to the volume grid data. In addition, we also perform this search only for two matching periodic markers at a time, rather than all markers at once, which also helps keep the cost down.

Common/src/geometry_structure.cpp
```diff
@@ -16064,6 +17403,288 @@ void CPhysicalGeometry::MatchActuator_Disk(CConfig *config) {
...
+                  dist = 0.0; for (iDim = 0; iDim < nDim; iDim++) {
+                    Coord_j[iDim] = Buffer_Receive_Coord[(iProcessor*MaxLocalVertex_Periodic+jVertex)*nDim+iDim];
+                    dist += pow(Coord_j[iDim]-rotCoord[iDim],2.0);
+                  } dist = sqrt(dist);
+                  
+                  if (((dist < mindist) && (iProcessor != rank)) ||
+                      ((dist < mindist) && (iProcessor == rank) && (jPoint != iPoint))) {
+                    mindist = dist; pProcessor = iProcessor; pPoint = jPoint; pPointGlobal = jPointGlobal;
+                    pVertex = jVertex_; pMarker = jMarker;
+                    if (dist == 0.0) break;
```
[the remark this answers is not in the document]
economon: I ended up removing that check entirely, since the cost is not prohibitive (it wasn't doing much), and we don't want to have an arbitrary tolerance that might cause us to find the wrong points if it is too large.

SU2_CFD/src/solver_direct_mean_inc.cpp
economon: For now, I have removed the preaccumulation calls entirely. The reason is that I have reworked the least-squares gradient calculation in order to be compliant with the new periodic BCs. In doing so, I have divided the routine into multiple loops over the nodes, and I do not think it is suitable for using preaccum. as we were before.

economon:
This PR is ready from my point of view. All periodic / turbomachinery meshes and regressions have been updated (thanks @salvovitale), and I can confirm they converge with the new periodic implementation. Only a few other regression tests needed to be updated due to the refactoring of the MPI, but they were very minor.

If the tests pass, then I recommend we merge very soon. Any final comments or reviews are most welcome.

SU2_CFD/src/solver_direct_mean_inc.cpp
pcarruscag: Do you know what is the effect on the memory footprint of the discrete adjoint?

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
pcarruscag: Yes I suppose one could even break the periodic boundaries in multiple markers, in any case it is nice when quadratic complexity can be avoided.

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
pcarruscag: It does not need to be arbitrary, it can be a fraction of the local cell sizes.

EduardoMolina:
Hi Tom,

Thanks for this great PR. I have two notes:
a) I have one grid (flow around a high lift airfoil) where the periodic boundaries are not placed  entirely on fix plane, like y=0.0, but on a very small value and fluctuates around, i.e. y~1.e-5. Now, many bad match points are found in the last commit of this PR. Previously, I was able to run the same mesh in a different commit. Do you change the tolerance of the search algorithm? Or I need to make sure now that all the periodic BC nodes must be placed on a fix plane.
b) I would like to share the results of the Taylor Green Vortex with the community. Indeed, the results are good and they are comparable with the reference DNS solution. 
SU2_TGV_DissipationRate

Thank you one more time.

Best,

Eduardo

SU2_CFD/src/driver_structure.cpp
```diff
@@ -186,7 +186,30 @@ CDriver::CDriver(char* confFile,
...
       }
-
+      
+      /*--- If we have any periodic markers in this calculation, we must
+       match the periodic points found on both sides of the periodic BC.
+       Note that the current implementation requires a 1-to-1 matching of
+       periodic points on the pair of periodic faces after the translation
+       or rotation is taken into account. ---*/
+      
+      if (config_container[iZone]->GetnMarker_Periodic() != 0) {
```
EduardoMolina: This part will create a segmentation fault when using the DG solver.  I think we can add "&& !fem_solver" in the if statement of line 196.

economon:
@EduardoMolina : Nice results! Thanks for the comments. You are seeing more warnings bc I did in fact change the tolerance that controls the warning. But, note that this tolerance only controls when to print the warnings, so the matches should have still be the same (nearest neighbor after transformation). 

The reason is that the new implementation assumes a 1-to-1 matching of the nodes and surface elements on either side of the periodic face. To be consistent, the points/faces on either side should match exactly so that they can be summed into a complete dual control volume (this assumption is made in the implementation and the partial residuals are summed as such).

Nice catch on the fem_solver check. I will add that to the conditional.

Thanks for the review @rsanfer. I will submit the final version shortly and if the tests pass, I would propose we merge this one first, if alright with @pcarruscag (the reason being that I would likely need your help to update the implementation for templating in the end anyway :) ).

economon:
Thanks for the input, everyone. Time to finally merge this one.

@pcarruscag : let's finish up #653 as a next step and get that merged asap. I can help fix up any conflicts in the matrix classes.

Issue #594, opened 2018-10-05
Discrete Adjoint Memory Usage


jayantmukho:
Hey everyone, 

We (@bmunguia and I) ran the discrete adjoint on the ONERAM6 wing mesh with ~500, 000 elements and noticed the following memory usage:

MG_LEVEL | DIRECT | DISCRETE_ADJOINT
-| ------------ | -------------
0| 1.3GB | 16GB
2| 1.5GB | 33GB

Is this increase in memory consumption expected? 

I ran the computation on the develop branch. The mesh is in TestCases/cont_adj_rans/oneram6/ 

Cheers, 
Jayant

erangit:
We have also encountered the significant memory requirements of the discrete adjoint. Actually, this requirement prevented the use of fine mesh (of 10-20M cells) in optimizations process. This is not the case while using continuous adjoint but the effectiveness of the discrete adjoint is higher by far.
I think this is a very important issue in the actual use of the method. Unfortunately, I do not have any solution to suggest.
Best,
Eran

bmunguia:
Another potential issue is unnecessary recording of computations. Jayant and I noticed that on his fix_wall_distance branch, the MG 2 case mentioned above was using 8GB more than on develop because during the tree search, the distance computations to all possible bounding boxes were being recorded, when the only necessary computation for recording is the final computation (once the minimum distance node is determined).

So I wonder if there are other places in the code where we can turn off recording to reduce memory usage. I'm curious what @talbring's thoughts are.

Best,
Brian

hlkline:
In terms of improvements on a larger scale, this work by Tom Taylor et al. describes a hybrid approach that could take down that memory requirement quite a bit more:
https://arc.aiaa.org/doi/abs/10.2514/6.2012-3342
In short, there are theoretically ways to get at least some of the advantages of both worlds (continuous vs discrete). I recall seeing some work that got only 4x more memory required (I think it was out of Jaoquim Martin's group? And I think they did it in part by implementing boundary conditions manually and using auto diff for jacobians) but I wasn't able to find the citation as quickly as I'd like. I hope that helps. If I manage to find that paper again I'll post again.

talbring:
@all, I agree that it might be a problem for bigger cases and that we have to think of some ways to reduce the memory requirements for the discrete adjoint. 

[...]

PR #543, opened 2018-06-13
Integrate MKL option for performance improvement on Intel Architectures.

costat:
This PR integrates new MKL features into SU2 to accelerate performance on Intel architectures. Changes 1) and 2) below are protected by a "HAVE_MKL" preprocessor flag. To observe performance improvement from change 2), the "DIRECT_CALL_MKL_SEQ" compiler flag must be enabled. These changes require MKL 2019 or newer, as the JIT GEMM feature is a brand new feature.

1) Integrate MKL JIT GEMM to accelerate MatrixMatrix and MatrixVector Products. 
2) Use LAPACK DGETRF + DGETRS in place of Gaussian Elimination ILU when MKL is present. 
3) Use memcpy in Gaussian Elimination ILU. Source/dest overlap is not a concern and this is faster.

The changes improve Broadwell performance by up to 18% and Skylake performance by up to 28%. These improvements were measured on the Inviscid_ONERA_M6 tutorial.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.

Issue #643, opened 2019-02-01
Linear Algebra library


pcarruscag:
Hi Folks,

I would like to collect the views of the community about something (and I apologize if this has been debated already).

Context:
Once again I find myself working on a feature that will require a bit of dense algebra.
I think implementing (i.e. paraphrasing from Numerical Recipes...) that kind of routine is a waste of time because someone has done already, and they have done it well.
Moreover, we have a very inefficient implementation of small matrix-like objects (arrays of arrays) and associated operations, which also leads to code bloat due to the required allocation / initialization / destruction.

Proposal:
Start using a linear algebra library such as [Eigen](http://eigen.tuxfamily.org/index.php?title=Main_Page).
For this particular library:
- The learning curve is minimal;
- Their license (MPL2) is less restrictive than SU2's;
- No link time dependencies (it is all header files);
- Compatible with AD (everything is templated);
- I like their logo...

I look forward to hear your opinions on this.

Best regards,
Pedro

bmunguia:
@EduardoMolina and I have discussed this over the past few weeks and are also in favor of using an external library. I don't have a strong opinion on the library we choose, but he seems to be in favor of [PETSc](https://www.mcs.anl.gov/petsc/) from ANL, which has a 2-clause BSD license and is used by ADflow (formerly SUmb), among other solvers. Eduardo could probably provide more details.

Another one that's come up in our discussions is [HYPRE](https://computation.llnl.gov/projects/hypre-scalable-linear-solvers-multigrid-methods) from LLNL which has a GNU LGPL.

juanjosealonso:
All,

I would agree that we need to look at an existing library and not re-develop one.

The standard for dense linear algebra is LAPACK. It has been around forever, relies on BLAS routines, is currently being used in the SU2 higher-order DG-FEM solver. Various computer vendors (including GPU manufacturers) have spent many years optimizing BLAS and LAPACK calls on their own computer architectures.  This means that your code is (a) portable and, by linking to the appropriate vendor-provided version of the library (say like Intel’s MKL), (b) you get highest performance.

I am not familiar with Eigen (just looked at the web page), but there is an interesting comparison between Eigen and LAPACK (written by the developers of Eigen…take it with a grain of salt) here<http://eigen.tuxfamily.org/index.php?title=FAQ#How_does_Eigen_compare_to_BLAS.2FLAPACK.3F>.

[...]

PR #653, opened 2019-02-16
Templated Linear Solvers

pcarruscag:
## Proposed Changes
This is the continuation to #650, so please compare these changes to the ones therein.

The linear systems solved in the discrete adjoint are now in passivedouble which speeds the discrete adjoint by a factor of about 1.5 (less for low CFL, more for stiff Jacobians like those in structural FEM). This is possible because we only carry the derivatives of the residual.

I tried to keep the design simple. CSysSolve and all related classes (vector, matrix, preconditioner, and matrix-vector product) have a single template parameter, the data type (which can be passivedouble or su2double). There are no provisions for "mixed" arithmetic, except in CSysSolve where through ::Solve (and only through ::Solve) one can ask for the solution of a system with passive Jacobian and active RHS and solution (this is possible at the expense of temporaries that are allocated only once). Passive-Passive and Active-Active work without temporaries and Active-Passive is not supported as it does not make sense (see end of previous paragraph).

This is to keep the need for template specialization to a minimum. Wherever mixing types was necessary small helper methods were defined to provide the compatibility instead of specializing larger methods, I think this keeps the code readable.

The place where passive and active types mix the most is CSysMatrix. This happens because the blocks are prepared by the numerics in **su2double and are then "Set", "Add", or "Subtract", on a CSysMatrix<passivedouble>. The solution was to inline those routines and template them also for the data type of the block (or diagonal).

I only tested on one fluid adjoint and one FSI adjoint case (fingers cross not to fail too many tests).

## Related Work
#594 Does not help with memory much but helps with speed.
#648 Makes it easier to interface with an external solver and still work with the discrete adjoint.
#650 Builds on top of what is proposed there.
#543 These MKL optimizations will now be possible for the discrete adjoint but I have not made them available yet.
Branch *feature_template_linear_solver*

## PR Checklist


talbring:
@pcarruscag Thanks for taking a lead on that and picking that up! I wanted to do that for a very long time, but never had enough time :+1: I will have a look into it soon.

rsanfer:
This has been sitting here for way too long; my apologies, @pcarruscag . Thanks for taking the lead and making the effort. The changes look good to me, but I'm no expert in templating myself, so I'd rather have somebody else with more experience in the topic approving it. Any comments, @economon @talbring ? However, I think this should go in soon, so if there are no further comments I'll be merging in the changes by next week.

Common/src/linear_solvers_structure.cpp
```diff
@@ -221,14 +232,14 @@ unsigned long CSysSolve::CG_LinSolver(const CSysVector & b, CSysVector & x, CMat
 
   mat_vec(x, A_x);
   r = b; r -= A_x;
-  su2double norm_r = r.norm();
-  su2double norm0 = b.norm();
+  ScalarType norm_r = r.norm();
```
economon: You will see some minor changes in the linear solver routines in #652 that introduce a COMM_LEVEL check to decide whether or not we compute residuals and print (save some minor overhead). Just fyi, in case we have conflicts, but they should be very minor

Common/src/linear_solvers_structure.cpp
```diff
@@ -657,52 +728,54 @@ unsigned long CSysSolve::Solve(CSysMatrix & Jacobian, CSysVector & LinSysRes, CS
...
         Jacobian.BuildJacobiPreconditioner();
-        precond = new CJacobiPreconditioner(Jacobian, geometry, config);
+        precond = new CJacobiPreconditioner<ScalarType>(Jacobian, geometry, config);
         break;
     }
     
     switch (KindSolver) {
       case BCGSTAB:
-        IterLinSol = BCGSTAB_LinSolver(LinSysRes, LinSysSol, *mat_vec, *precond, SolverTol, MaxIter, &Residual, ScreenOutput);
+        IterLinSol = BCGSTAB_LinSolver(*LinSysRes_ptr, *LinSysSol_ptr, *mat_vec, *precond, SolverTol, MaxIter, &Residual, ScreenOutput);
```
economon: Same here.. might be a conflict w #652 since I now send in the config container at the end of the argument list. Should be minor.

Common/src/matrix_structure.cpp
```diff
@@ -975,11 +892,13 @@ void CSysMatrix::DiagonalProduct(CSysVector & vec, unsigned long row_i) {
   
 }
 
-void CSysMatrix::SendReceive_Solution(CSysVector & x, CGeometry *geometry, CConfig *config) {
+template<class ScalarType>
+template<class OtherType>
+void CSysMatrix<ScalarType>::SendReceive_Solution(CSysVector<OtherType> & x, CGeometry *geometry, CConfig *config) {
```
economon: #652 replaces this communication routine as well as the transpose version. We will need to update those new MPI communication routines with the templated data types you have here..

Common/src/matrix_structure.cpp
```diff
@@ -1329,14 +1259,14 @@ void CSysMatrix::InverseDiagonalBlock(unsigned long block_i, su2double *invBlock
       invBlock[jVar*nVar+iVar] = aux_vector[jVar];
   }
   
-  //  su2double Det, **Matrix, **CoFactor;
-  //  su2double *Block = GetBlock(block_i, block_i);
+  //  ScalarType Det, **Matrix, **CoFactor;
```
economon: Can this commented code be removed?

Common/src/matrix_structure.cpp
```diff
@@ -1380,14 +1310,14 @@ void CSysMatrix::InverseDiagonalBlock_ILUMatrix(unsigned long block_i, su2double
       invBlock[jVar*nVar+iVar] = aux_vector[jVar];
   }
   
-  //  su2double Det, **Matrix, **CoFactor;
-  //  su2double *Block = GetBlock_ILUMatrix(block_i, block_i);
+  //  ScalarType Det, **Matrix, **CoFactor;
```
economon: Same.. can this commented code be removed?

pcarruscag:
Thanks guys.
@economon, regarding the comments, I am happy to remove them if that is ok with who put them there.
Regarding the conflicts I noticed them when first reading through 652, the first two you mention are easy to solve, the one about moving the comms will probably require templating the associated method, I can help with the merge.
Cheers,
Pedro

economon:
Looks like the commented code was added here (https://github.com/su2code/SU2/commit/b5db893bcc57a67abbbe43ddd5a87753ec61d23b) but never activated. The Matrix* routines are not being used anywhere at the moment. Do you see some value in testing it out? Otherwise, might be best to remove so we aren't carrying around dead code.

pcarruscag:
I see, I do not know what is the quickest way to invert a 5x5 matrix, most robust would probably be LU with pivoting (for which we have the code in the RBF class). Since that relates to how we handle small dense matrices I would say it relates to #643 so it would be good if the community got to a conclusion there.
In any case I want this PR to be only about templating, I can do that kind of cleanup when I:
- Try to activave the MKL optimizations for the discrete adjoint.
- Move the row/col elimination tasks from the structural solver and mesh deformation to CSysMatrix (as you suggested in #658).

economon:
Sounds good to me. Then, I suggest we leave the comments for now, and you can come back to it when considering #643 further (or when some performance issues are considered) in a later PR.

Common/src/linear_solvers_structure.cpp
```diff
@@ -463,7 +476,7 @@ unsigned long CSysSolve::FGMRES_LinSolver(const CSysVector & b, CSysVector & x,
   if (monitoring) {
```
pcarruscag: @economon, I noticed that here you did not add the FULL_COMMS condition, do you want me to?

economon: If you have the time, sure. That would be great.

The framework is not complete yet, but as I am sure you have figured, the idea is to have COMM_NONE, COMM_MINIMAL, COMM_FULL. COMM_FULL is default, COMM_MINIMAL reduces the MPI comms slightly by disabling things like residual reductions during runtime that do not affect correctness, and COMM_NONE will eventually disable all comms during runtime (purely for "ideal" performance and scalability benchmarking, which will obviously be incorrect in terms of the solution).

economon:
W.r.t. the translation in IntiateComms() and CompleteComms(): appears to be a pretty straightforward typecasting to take care of the templating. LGTM.

Common/src/linear_solvers_structure.cpp
pcarruscag: There is quite a lot of work on low communication Krylov methods, things like not reducing dot products for example, I am not using enough cores to make that an issue yet (I think) but it could be a nice feature.
I'll just make FGMRES consistent with CG and BCGSTAB.

pcarruscag:
Thank you, I am putting this to rest now then.

PR #654, opened 2019-02-19
adding implicit check to prevent segfault

WallyMaier:
## Proposed Changes
Adding an implicit flag to prevent segfault while running Roe scheme explicitly.
 
## Related Work
N/A

## PR Checklist

- [x ] I am submitting my contribution to the develop branch.

### DIFF
--- a/SU2_CFD/src/numerics_direct_mean.cpp
+++ b/SU2_CFD/src/numerics_direct_mean.cpp
@@ -1,4 +1,4 @@
-/*!
+﻿/*!
  * \file numerics_direct_mean.cpp
  * \brief This file contains the numerical methods for compressible flow.
  * \author F. Palacios, T. Economon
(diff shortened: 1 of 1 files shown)


SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -3274,9 +3274,11 @@ void CUpwRoe_Flow::ComputeResidual(su2double *val_residual, su2double **val_Jaco
   if (RoeSoundSpeed2 <= 0.0) {
     for (iVar = 0; iVar < nVar; iVar++) {
       val_residual[iVar] = 0.0;
-      for (jVar = 0; jVar < nVar; jVar++) {
-        val_Jacobian_i[iVar][iVar] = 0.0;
-        val_Jacobian_j[iVar][iVar] = 0.0;
+      if (implicit){
+        for (jVar = 0; jVar < nVar; jVar++) {
+          val_Jacobian_i[iVar][iVar] = 0.0;
```
pcarruscag: Why do we loop through jVar but access only the diagonal here? This is either a bug or an unnecessary loop.

vdweide: It is a bug, also present in the original code. One of the iVar should be jVar. 

The question is whether this treatment of negative square of the speed of sound (setting the flux to zero) is the best one possible (if there already is a correct treatment). Maybe a dissipation term based on a^* would be better in this case (if it is only used during the iteration process).

pcarruscag: Agreed. Would clamping RoeSoundSpeed2 to some safe lower bound be a solution? Setting Jacobian contributions to 0 may make it ill-conditioned.

vdweide: At least better than what is currently in there, although these tricks usually only postpone the crash.

PR #658, opened 2019-02-22
Fix given displacement BC's of FEA solver and CElasticityMovement

pcarruscag:
## Proposed Changes
Problem: If a given displacement boundary was cut during partitioning the solver would give wrong results.
Reason: When "de-singularizing" the stiffness matrix the column was only being deleted by the rank that owned the corresponding node.
Fix: All ranks delete the column (and update the load vector).

Before:
image

After:
image

## PR Checklist

### DIFF
--- a/Common/src/grid_movement_structure.cpp
+++ b/Common/src/grid_movement_structure.cpp
@@ -9315,14 +9315,43 @@ void CElasticityMovement::UpdateMultiGrid(CGeometry **geometry, CConfig *config)
 
 void CElasticityMovement::SetBoundaryDisplacements(CGeometry *geometry, CConfig *config){
 
-  unsigned short iMarker;
-
   /*--- Get the SU2 module. SU2_CFD will use this routine for dynamically
    deforming meshes (MARKER_FSI_INTERFACE). ---*/
 
   unsigned short Kind_SU2 = config->GetKind_SU2();
 
-  /*--- First of all, move the FSI interfaces. ---*/
+  /*--- Share halo vertex displacements across ranks using the solution vector.
+   The transfer routines do not do this, i.e. halo vertices have wrong values. ---*/
+
+  unsigned short iMarker, iDim;
+  unsigned long iNode, iVertex;
+
+  su2double VarIncrement = 1.0/su2double(config->GetGridDef_Nonlinear_Iter());
+
+  for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
+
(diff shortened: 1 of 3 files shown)


vdweide:
Did you check whether the results are identical, independent of the number of MPI ranks?

pcarruscag:
I sure did. What I did not do was reproduce the problem in CElasticityMovement, that is only used for FSI and I only have 2D FSI cases, for which it is less probable that the boundaries will be cut.

pcarruscag:
This had to be more intricate than first impressions suggested...
The displacements applied by CElasticityMovement to FSI interfaces come from the vertices, but halo vertices do not have the correct (i.e. communicated) coordinates which means for those the force vector contributions were not being computed correctly.
The known blocks (prescribed displacement) of the displacement vector are now set and communicated before "de-singularizing" the matrix, that process then grabs displacements from the displacement vector instead of from the vertices.

Common/src/grid_movement_structure.cpp
```diff
@@ -9386,36 +9413,25 @@ void CElasticityMovement::SetClamped_Boundary(CGeometry *geometry, CConfig *conf
...
-        if (valJacobian_ij_00 != 0.0 ){
-          /*--- Set the rest of the row to 0 ---*/
-          if (iNode != jPoint) {
-            StiffMatrix.SetBlock(iNode,jPoint,matrixZeros);
-          }
-          /*--- And the diagonal to 1.0 ---*/
-          else{
-            StiffMatrix.SetBlock(iNode,jPoint,matrixId);
-          }
+        if (iNode != jPoint) {
```
economon: There are several places where you do a similar operation as to what is done here.. have you considered reusing something like StiffMatrix.DeleteValsRowi() or creating a new version that suits what you are doing here within the matrix class to save some code by calling it instead?

pcarruscag: Yes definitely, especially since these operations can be performed more efficiently if the sparsity pattern is know. And there are other areas too, for example we currently add the contribution of the mass matrix to the stiffness matrix in a similar block by block way. But I would prefer to address all of those in one PR and avoid creating conflicts between ongoing PR's.

rsanfer: I think creating a StiffMatrix.DeleteValsColi() would do the job. There is some room for code saving in all of this implementation, but I agree with @pcarruscag, that could be done in an additional PR.

PR #663, opened 2019-03-25
Reduce memory footprint FEA problems

rsanfer:
## Proposed Changes
This is a small one, just saving some memory in the FEA solver by not allocating some unnecessary double structures outside the boundaries. 

## Related Work
General maintenance and improvements.

## PR Checklist


pcarruscag:
Hi @rsanfer,
Did you consider using a derived class of CFEAVariable that augments it with the fluid tractions?
This would save you the memory of the data, the pointer, and the boolean (which due to alignment issues is probably not just 1 byte). It would also avoid the if statements.
Cheers for removing the methods returning pointers by the way.
Pedro

rsanfer:
Hi @pcarruscag 
in fact it was the boolean what I originally wanted to have there, so that it could easily be checked whether a node is a surface vertex without having to loop through the markers. Getting the memory savings this way just came naturally, but it's true that it could be done more efficiently. 
As it's not just the fluid tractions the ones involved, but also the surface pressures, I've reused an old CFEABoundVariable class that was there sitting unused (so, 2 for the price of 1!) to store those containers. It should lead to some additional memory savings.
Cheers,
Ruben

PR #670, opened 2019-04-18
Add missing term to strain magnitude for 2D

clarkpede:
## Proposed Changes

For 2D problems, the calculation for the magnitude of the rate-of-strain tensor is missing a term.

When summing over the diagonal components of the deviatoric rate-of-strain tensor, the code loops over `nDim` diagonal components.  In 2D, that means it only adds the first two diagonal components. This creates problems, because the third component is *not* zero, even for 2D flow.  While third component of the rate-of-strain tensor in 2D is zero (i.e. S_33 = 0), the deviatoric rate-of-strain tensor has a nonzero third diagonal component.  So in 2D, the third diagonal component must be explicitly added.

This should only make significant impacts on problems where the divergence of velocity is non-negligible when compared to the diagonal components of the rate-of-strain tensor (i.e. for higher Mach cases).

## Related Work
None that I know of.

## PR Checklist


economon:
Thanks, @clarkpede, it might be that we are missing this in other places too, such as in the numerics classes possibly. Did you already have a look elsewhere in the code?

clarkpede:
I've done a brief search.  There's an almost exact duplicate of the function in the incompressible solver (`SetStrainMag`). But as far as I am aware, stating that a flow is incompressible means that the divergence of velocity is zero.  So the extra term would be zero, and no change is necessary.  But that leaves the question, "Why is the divergence of velocity included there in the first place?"

economon:
Thanks for checking, @clarkpede. The reason that the full stress tensor (and divergence of velocity) appear still in the incompressible solver is for the situation when the density is allowed to vary with the ideal gas law (low Mach formulation). Here, the density can vary with temperature, which leads to the divergence of velocity being a function of temperature (instead of zero).

clarkpede:
So I want to update the `transonic_stator_2D` test case with a new restart file that matches the new numerics.  Since it's a transonic case, the difference in production is significant.

But what was the stopping criteria used while generating the restart file in [TestCases/turbomachinery/transonic_stator_2D](https://github.com/su2code/TestCases/tree/develop/turbomachinery/transonic_stator_2D)? Was it the same stopping criteria used in the [non-restart cfg file](https://github.com/su2code/SU2/blob/master/TestCases/turbomachinery/transonic_stator_2D/transonic_stator.cfg)? For that case, the solver stops at 2001 external iterations.

clarkpede:
For my part, this is ready to be merged in once su2code/TestCases#34 goes through. Only the transonic stator restart case is failing.

economon:
Just merged the test. Let's get this merged next once it passes.

PR #671, opened 2019-04-23
Single-zone adjoint driver

rsanfer:
## Proposed Changes
This PR goes in line with the driver reorganisation started in #585. At this time, it is the single-zone adjoint driver that is introduced, which inherits from the primal single-zone driver class. Problem properties are set at time of driver instantiation, and routines are mostly opaque to the physics (not yet completely as some further work is required in input/output). The functions are exposed to python so in a later PR the adjoint can also be used with the AD mode enabled. Some old driver classes are already removed.

## Related Work
- Further work by @oleburghardt to follow on the multizone adjoint driver.
- Some features awaiting the input/output restructuring PR by @talbring.

## PR Checklist


TestCases/cont_adj_euler/wedge/of_grad_combo.dat.refdiscrete
```diff
@@ -1,5 +1,5 @@
 VARIABLES="VARIABLE"      , "GRADIENT"      , "FINDIFF_STEP"  
-          0               , 0.00146785      , 0.001           
-          1               , 0.0015213       , 0.001           
-          2               , 0.00177652      , 0.001           
-          3               , 0.00222003      , 0.001           
+          0               , 0.00651231      , 0.001
```
pcarruscag: Do you know why these have changed?

SU2_CFD/src/driver_adjoint_singlezone.cpp
```diff
@@ -0,0 +1,521 @@
...
+  }
+
+}
+
+void CDiscAdjSinglezoneDriver::MainRecording(){
+
+  /*--- SetRecording stores the computational graph on one iteration of the direct problem. Calling it with NONE
+   *    as argument ensures that all information from a previous recording is removed. ---*/
+
+  SetRecording(NONE);
```
pcarruscag: Is this required? I ask because it takes time (especially for FEA+Topology) and in the Adj FSI driver I have it commented out and all seems fine.

TestCases/cont_adj_euler/wedge/of_grad_combo.dat.refdiscrete
rsanfer: Yes, in fact I had to chase this one down for a bit. In principle this is a bug in the current code. It comes because the adaptive CFL is enabled and any time we do a recording after the first one (so, after there has been any geometry recording, for example) it uses the adapted CFL, so the path is changed a bit. With this version, that's no longer the case as the CFL adaptation is not called for the adjoint driver.

SU2_CFD/src/driver_adjoint_singlezone.cpp
rsanfer: Actually, I have not tested it without, so it might be the case that it is not needed. Let me comment it out and see what it does.

rsanfer: I just tested it, flow adjoint tests are all failing 
https://travis-ci.org/su2code/SU2/builds/526434537
so I reverted the changes. I think that @talbring can explain better why that's necessary.

SU2_CFD/include/driver_structure.hpp
```diff
@@ -1583,6 +1406,106 @@ class CSinglezoneDriver : public CDriver {
...
+  /*!
+   * \brief Constructor of the class.
+   * \param[in] confFile - Configuration file name.
+   * \param[in] val_nZone - Total number of zones.
+   * \param[in] MPICommunicator - MPI communicator for SU2.
+   */
+  CDiscAdjSinglezoneDriver(char* confFile,
+             unsigned short val_nZone,
+             unsigned short val_nDim,
+             bool val_periodic,
```
economon: Can you please remove references to 'val_periodic' in all driver constructors? They were used before to check for periodic meshes so that the old CPhysicalGeometry() constructor could be instantiated. This is no longer necessary now that the periodic BCs have been fixed. That variable can simply be removed from the argument list and deleted.

SU2_CFD/include/iteration_structure.hpp
```diff
@@ -1515,6 +1533,23 @@ class CDiscAdjFEAIteration : public CIteration {
                       unsigned short val_iInst,
                       unsigned short kind_recording);
 
+  /*!
+   * \brief Record a single iteration of the direct mean flow system.
```
economon: Thanks for keeping the doxy up to date.. double-check some of these comments though (should mean flow -> elasticity here?)

SU2_CFD/src/SU2_CFD.cpp
```diff
@@ -161,19 +152,7 @@ int main(int argc, char *argv[]) {
...
-        driver = new CDiscAdjTurbomachineryDriver(config_file_name, nZone, nDim, periodic, MPICommunicator);
-
-      } else {
-
-        driver = new CDiscAdjFluidDriver(config_file_name, nZone, nDim, periodic, MPICommunicator);
-        
-      }
-
-    } else if (turbo) {
+    if (turbo) {
```
economon: Will the remaining capability from all other drivers (turbo, HB, etc) be absorbed into the primal and adjoint single/multizone drivers such that they can be removed eventually?

SU2_CFD/src/driver_adjoint_singlezone.cpp
```diff
@@ -0,0 +1,521 @@
...
+  bool compressible = (config->GetKind_Regime() == COMPRESSIBLE);
+  bool heat         = (config->GetWeakly_Coupled_Heat());
+  bool turbo        = (config->GetBoolTurbomachinery());
+
+  ObjFunc = 0.0;
+
+  /*--- Specific scalar objective functions ---*/
+
+  switch (config->GetKind_Solver()) {
+  case EULER:                    case NAVIER_STOKES:                   case RANS:
```
economon: You mention moving some items to the new output classes (I have seen this in the IO branch), but is it possible to move all of these switches and conditionals to the output too for holding the obj functions, such that one can simply call a virtual function in the base output class like output->SetObjFunc() here and not worry about the particular solver? We have discussed this before and maybe this is the longer term plan..

SU2_CFD/src/driver_structure.cpp
```diff
@@ -4069,65 +4069,6 @@ void CDriver::Output(unsigned long ExtIter) {
 
 CDriver::~CDriver(void) {}
 
-CGeneralDriver::CGeneralDriver(char* confFile, unsigned short val_nZone,
```
economon: Nice clean up.. feels good to remove lines :).

TestCases/disc_adj_euler/arina2k/Arina2KRS.cfg
```diff
@@ -48,6 +48,9 @@ DISCARD_INFILES= NO
 %                                       Speed = ft/s, Equiv. Area = ft^2 )
 SYSTEM_MEASUREMENTS= SI
 
+SINGLEZONE_DRIVER = YES
```
economon: Is this option required or will it be deduced automatically?

SU2_CFD/src/driver_adjoint_singlezone.cpp
pcarruscag: Interesting, thanks for trying it. The FSI adjoint seems to work fine without it though, it definitely does not diverge, another AD mystery.

SU2_CFD/src/driver_adjoint_singlezone.cpp
```diff
@@ -0,0 +1,521 @@
...
+      cout << "Direct iteration to store the primal computational graph." << endl;
+      cout << "Compute residuals to check the convergence of the direct problem." << endl;
+    }
+    iteration->RegisterInput(solver_container, geometry_container, config_container, ZONE_0, INST_0, kind_recording);
+
+  }
+
+  /*--- Set the dependencies of the iteration ---*/
+
+  iteration->SetDependencies(solver_container, geometry_container, numerics_container, config_container, ZONE_0, INST_0, kind_recording);
```
pcarruscag: I have a request, can you break some of these lines? I find myself horizontal scrolling too many times, maybe 120 characters long?

rsanfer: Sure, just limited the lines to 118 characters, which I think it's the limit of the github editor. Will push it in a bit.

talbring:
In general everything looks fine to me! I will do some smaller adjustments to this new driver for the new output structure soon (but in the other branch, not here). We can merge this in as long as there are no other comments.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2794,165 +2850,33 @@ void CDiscAdjFEAIteration::Iterate(COutput *output,
...
-  nIntIter = config_container[val_iZone]->GetDyn_nIntIter();
-
-  for(IntIter = 0; IntIter < nIntIter; IntIter++){
-
-    /*--- Set the internal iteration ---*/
-
-    config_container[val_iZone]->SetIntIter(IntIter);
-
-    /*--- Set the adjoint values of the flow and objective function ---*/
+  unsigned long nIntIter = config_container[val_iZone]->GetnIter();
```
TobiKattmann: I just noticed GetnIter() returns the value of the variable Iter and not nIter (which does not exist) from the CConfig class. 
It would be nice to have the S/Get<VarName> right? 
More a general question than a change request.

TestCases/disc_adj_rans/cylinder_DT_1ST/cylinder.cfg
```diff
@@ -36,6 +36,11 @@ WRT_CSV_SOL= NO
 WRT_VOL_SOL= NO
 %
 WRT_SRF_SOL= NO
+
+SINGLEZONE_DRIVER = YES
```
TobiKattmann: I cant find any of these 4 new config options in the config_template. Please add those with a description of how and when to use it.

rsanfer: The idea is that SINGLEZONE_DRIVER goes away after the driver refactoring is complete, that's why it wasn't added so far. I've just tentatively added them with a brief explanation, but further description will be needed once the driver is completely refactored and ITER / TIME_ITER become the standard.

Issue #581, opened 2018-09-19
Validation Suite for SU2


jayantmukho:
Hey everyone, 

Following the discussions at the SU2 Developers meeting this week, I wanted to start a conversation about compiling a comprehensive set of V&V cases for SU2 that can showcase it's performance in comparison to other solvers. 

I think the the [NASA TMR website](https://turbmodels.larc.nasa.gov/index.html) is a good model to base it on. The idea would be to present the V&V case, provide working configuration and mesh files, and provide results comparing performance to other solvers and to higher fidelity data (when available). This allows people to see the performance of SU2 and replicate it, if need be. 

The first step would be compiling a list of cases that should be covered. The [SU2 2014 SciTech paper](https://su2code.github.io/documents/SU2_AIAA_SciTech2014.pdf) would be a good starting point as it already had a couple of validation cases. This list can be bolstered with some of the NASA TMR cases, and with grid convergence studies. I would like to propose an initial list that the community can talk through and make changes as we see fit. I am mostly only familiar with canonical CFD flows that are used in these cases. But it would be great to have other cases, such as Turbo-machinary or FSI cases, that show the full breadth of SU2's abilities. This is by no means an exhaustive list:

1. Zero Gradient Flat Plate
2. 2D and 3D Bump in Channel
3. Asymmetric Diffuser
4. Backward facing step
5. Unsteady Square cylinder
6. NACA0012
7. NACA4412 Trailing Edge Seperation
8. Joukowski Airfoil
9. 30P30N High Lift airfoil
10. ONERAM6 Wing
11. NASA CRM
12. Subsonic and Supersonic Jets

It would be ideal for these test cases to have either high-fidelity data (wind tunnel tests, or LES/DNS data), and/or published results of other solvers so that comparisons can be made. 

I also want to point out the difference between this and the TestCases repository. The TestCases repo is used in regression tests to ensure that parts of the code don't break when changes are made. This is more a test of SU2's fidelity. It would feature large grids and configuration files that can be run to convergence.

erangit:
Hi Jayant
Thanks for being so prompt and fast in handling this very important issue.
I’m currently in vacation at Scotland so I’ll be brief and just add few
comments:
a. As you wrote, validation is very different from test cases. Mesh,
configuration and solutions should be optimized for accuracy, experiment
reproduction and performance.
b. I think it is important to start the validation project, create a link
from the main SU2 web page to it, and than let it grow with the existing
[...]

PR #672, opened 2019-04-23
Solution Verification Framework

economon:
## Proposed Changes
 
This is a joint effort with @vdweide. 

We have added an extensible library of classes for the purposes of solution verification. We hope that folks will use it and continue growing the library to increase the formal verification activity for the code. Our design goals for the classes were:

* lightweight & small memory footprint.
* reusable by different solvers (so far: all DG and FV fluid solvers).
* require almost no changes to the solver classes when switching between different solutions by using polymorphism (other than maybe where they get instantiated).
* applicable for any of the following: imposing initial conditions, analytic solutions, or manufactured solutions.

There is also now a user-defined solution class that folks can use to prescribe initial conditions as a function of space for any type of calculation. It should be easy to add new solutions to the library by creating new child classes.

We are placing a few cases in the regressions, but we will also add full cases with analysis to the V&V repo. It will also be discussed in a presentation at the developers meeting. You will find a few other small bug fixes in the PR as well.

As an example, here are some (particularly nice) sample observed order of accuracy results for the compressible FV solver with a laminar N-S MMS (to be fair, not all look this clean):

accuracy_rms_rho
slope_rms_rho

## Related Work

Cases to be added as part of the V&V repo discussed in #581 

## PR Checklist


jayantmukho:
I am not a 100% clear on all the things this new feature can do. From what I gather we can now set an initial flow field that is not just freestream condition everywhere? If this is the case, this is a hugely useful feature so thank you guys for doing that. 

How exactly is this allowing for solution verification? Is it allowing you to run the same case with a bunch of different solver schemes? Can it run the solvers on a set of meshes or do you still have to run on each mesh refinement individually?

juanjosealonso:
Good point, Jayant. Maybe a brief tutorial and an example can be created so people know how to use this new feature? There will be a presentation by Edwin and Tom at the annual meeting that might also be helpful.

Best,
Juan

On Apr 24, 2019, at 11:32 AM, Jayant Mukhopadhaya <notifications@github.com<mailto:notifications@github.com>> wrote:


I am not a 100% clear on all the things this new feature can do. From what I gather we can now set an initial flow field that is not just freestream condition everywhere? If this is the case, this is a hugely useful feature so thank you guys for doing that.

How exactly is this allowing for solution verification? Is it allowing you to run the same case with a bunch of different solver schemes? Can it run the solvers on a set of meshes or do you still have to run on each mesh refinement individually?

vdweide:
@jayantmukho, as @juanjosealonso indicated the verification efforts will be presented at the annual developers meeting. However, we may not have time to answer all questions there, so a tutorial may be a good thing to have.

To answer your question about setting different initial conditions (as well as space and time dependent boundary conditions), yes you can do that with the class CUserDefinedSolution, although this was not the primary intention. The intention of these classes is to run a verification case, where you add source terms to the governing equations, such that a given solution is an exact solution to the modified equations. In this way you can do a thorough accuracy check of your discretization, as shown in the figures by @economon.

Hope this answers your questions a bit.

Common/include/toolboxes/solution_toolbox.hpp
```diff
@@ -0,0 +1,1084 @@
...
+#include <cmath>
+#include "../config_structure.hpp"
+
+/*!
+ * \class CVerificationSolution
+ * \brief Class for holding verification PDE solutions, e.g., phi = phi(x,y,z,t),
+ *        used for initial conditions, analytic solutions, manufactured solutions.
+ * \author T. Economon, E. van der Weide
+ */
+class CVerificationSolution {
```
talbring: Since we start reorganizing the file structure soon, it would be maybe good to already start with that here. Separating all classes into individual header/source files inside of an MMS directory. What do you say ?

vdweide: Fine with me. Do you want that MMS directory then directly in Common/include or in Common/include/toolboxes? Same for the sources of course.

talbring: We can put the MMS directory in the toolboxes directory. I don't mind having additional sub-directories.

vdweide:
@talbring, I created separate include and source files for the verification solution classes. Is this what you had in mind?

talbring:
@vdweide Yep that is exactly what I had in mind. Thank you. Looks good. What do the others think ?

economon:
Completely in support of dividing the files!

I think it is time though to then also decide on an objective criteria for this for all developers to adhere to in the future. Shall we then formalize it as a single C++ class per file, i.e., each class has a .cpp and an .hpp (inline files merged into the .hpp files as in the new output)? Merging the inlines with the header may slow down compile times, but not sure how much. Would be good to also standardize the naming convention for these files. 

Thoughts from others?

talbring:
Lets first merge this one in I would say. Then we open an issue to discuss it further.

vdweide:
I also prefer one .cpp and .hpp per class. However, that will lead to significant conflicts with other branches. Any idea how to handle that?

economon:
@vdweide : I think we have no choice but to do it piece-by-piece over time by opening PRs and alerting folks that the change will be made so that they can merge anything critical beforehand into develop.

I am fine with merging this as is. If there are no additional comments from the community, then we can merge shortly and I'll open a new issue for discussion of file division.

economon:
This PR is ready to go, but we need some reviews from the community before officially merging. If you have some time, please give it a look.

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
```diff
@@ -0,0 +1,402 @@
...
+                                                    const su2double val_t,
+                                                    su2double       *val_source) {
+
+  /*--- Abbreviate Pi and the coordinates. ---*/
+  const su2double Pi = PI_NUMBER;
+  const su2double x = val_coords[0];
+  const su2double y = val_coords[1];
+  const su2double z = val_coords[2];
+
+  /*--- The source code for the source terms is generated in Maple. ---*/
```
pcarruscag: If small enough, can you include the steps used to generate this kind of machine generated code?

SU2_CFD/src/driver_direct_multizone.cpp
```diff
@@ -207,6 +207,15 @@ void CMultizoneDriver::Preprocess(unsigned long TimeIter) {
     /*--- TODO: This should be generalised for an homogeneous criteria throughout the code. --------*/
     config_container[iZone]->SetExtIter(TimeIter);
 
+    /*--- Store the current physical time in the config container, as
+     this can be used for verification / MMS. This should also be more
+     general once the drivers are more stable. ---*/
+    
+    if (unsteady)
+      config_container[iZone]->SetPhysicalTime(static_cast<su2double>(TimeIter)*config_container[iZone]->GetDelta_UnstTimeND());
```
pcarruscag: Why static_cast?

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -3078,6 +3089,26 @@ void CEulerSolver::SetInitialCondition(CGeometry **geometry, CSolver ***solver_c
...
+
+    /*--- Loop over the multigrid levels. ---*/
+    for (iMesh = 0; iMesh <= config->GetnMGLevels(); iMesh++) {
+      
+      /*--- Loop over all grid points. ---*/
+      for (iPoint = 0; iPoint < geometry[iMesh]->GetnPoint(); iPoint++) {
+        
+        /* Set the pointers to the coordinates and solution of this DOF. */
+        const su2double *coor = geometry[iMesh]->node[iPoint]->GetCoord();
+        su2double *solDOF     = solver_container[iMesh][FLOW_SOL]->node[iPoint]->GetSolution();
```
pcarruscag: From an idiomatic point of view it would be clearer if you put the manufactured solution into a tmp variable and then called SetSolution on the node (also good practice not to get pointers to private variables). Three calls to "getter" methods makes it look you are not doing anything.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -4093,6 +4124,40 @@ void CEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_contain
...
+      for (iPoint = 0; iPoint < nPointDomain; iPoint++) {
+        
+        /*--- Get control volume size. ---*/
+        su2double Volume = geometry->node[iPoint]->GetVolume();
+        
+        /*--- Get the current point coordinates. ---*/
+        const su2double *coor = geometry->node[iPoint]->GetCoord();
+        
+        /*--- Get the MMS source term. ---*/
+        vector<su2double> sourceMan(nVar,0.0);
```
pcarruscag: Good, I like stl containers and const *

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -12848,6 +12997,105 @@ void CEulerSolver::UpdateSolution_BGS(CGeometry *geometry, CConfig *config){
...
+            cout << setw(20) << "Max Error [RhoW]: " << setw(12) << GetError_Max(3);
+            cout << endl;
+          }
+        
+          cout << setw(20) << "RMS Error [RhoE]: " << setw(12) << GetError_RMS(nDim+1) << "     | ";
+          cout << setw(20) << "Max Error [RhoE]: " << setw(12) << GetError_Max(nDim+1);
+          cout << endl;
+        
+          cout << "-------------------------------------------------------------------------" << endl << endl;
+          cout.unsetf(ios_base::floatfield);
```
pcarruscag: Nice, this is often forgotten.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -466,6 +475,64 @@ class CSolver {
    */
   su2double* GetPoint_Max_Coord_BGS(unsigned short val_var);
   
+  /*!
```
pcarruscag: Could this methods and associated variables be put in CVerificationSolution? The CSolver namespace is very cluttered and with this naming these variables and methods are asking for trouble.

SU2_CFD/src/driver_direct_multizone.cpp
vdweide: I guess this is to avoid a compiler warning, but @economon is better suited to comment on this.

SU2_CFD/src/solver_direct_mean.cpp
vdweide: I did that to avoid an additional copy, but that just comes from my old habit that copies must be avoided, if possible. We can easily change that, if that improves readability.

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
vdweide: Yes, that is possible. However, any symbolic package can do this, and we did not want to make the impression that this has to be done with Maple or python sympy. If people think it will clarify matters, we can certainly add this information.

SU2_CFD/src/solver_direct_mean.cpp
vdweide: Thanks. It can avoid bugs, if you know beforehand that a variable is constant in a piece of code.

SU2_CFD/src/solver_direct_mean.cpp
vdweide: Thank @economon for that.

SU2_CFD/include/solver_structure.hpp
vdweide: I suppose you mean the methods AddError_RMS, AddError_Max, SetError_RMS and SetError_Max? Yes, I think we can move those to CVerificationSolution, but I am not 100 percent sure. @economon, is there a compelling reason to keep these methods in CSolver?

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
pcarruscag: I think it is clear, it is just in case someone wants to reproduce the same steps, serves as mini tutorial almost.

SU2_CFD/src/driver_direct_multizone.cpp
pcarruscag: I only ask because most other casts in the code are C style

SU2_CFD/src/solver_direct_mean.cpp
pcarruscag: Yeah performance wise this is probably better, virtual methods are involved so I doubt the compiler could optimize away the copy. In any case in the call to the "Get" in the verification class maybe say it will write directly to the node.

SU2_CFD/include/solver_structure.hpp
pcarruscag: Yes, this entire new section line 478 to 535

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
vdweide: Would it be an idea to include the Maple and Python scripts in a separate directory then?

SU2_CFD/src/solver_direct_mean.cpp
vdweide: Performance is not really an issue here, because a) it is in the initialization phase and b) performance is of secondary, it not less, importance for the verification cases. Readability is much more important. So if it is clearer with the additional copy, we should do this.

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
pcarruscag: A good one I think

rsanfer: Should we wait for this before we merge it in, then?

economon: I think adding the scripts to the VandV repo where our initial MMS cases have already been placed could be ideal. Fits in nicely, and we will eventually be able to explain everything with those examples (note that there is a working group to discuss what the docs should look like there).

SU2_CFD/src/driver_direct_multizone.cpp
economon: No particular reason from what I recall. We can go with C style for consistency

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
vdweide: @economon, I just pushed the Maple files that generate the source code for the MMS source terms to Common/src/toolboxes/MMS/CreateMMSSourceTerms. The reason for putting in the source tree and not in the VandV repo is that it generates actual source code. However, if you prefer the VandV repo for these scripts, that is fine with me as well as long as it is properly referenced in the source code. 

BTW, I just realized that I have not added a regression test case for this feature, so do not merge yet.

SU2_CFD/src/solver_direct_mean.cpp
vdweide: Should we change this to Get and Set?

SU2_CFD/include/solver_structure.hpp
economon: Yep, I like this idea. I think that the Set/Add/GetError_*() can all move to CVerificationSolution, as well as the SetVerificationError() routine, which just does a final communication and stores the value.

Prob need to keep the ComputeVerificationError() in the solvers though, since this routine holds the loop over the DOFs to compute the local error, and these loops have different structures in the different solver (FV vs DG, for instance).

SU2_CFD/src/driver_direct_multizone.cpp
pcarruscag: Actually that seems to be the safe way to do it (https://stackoverflow.com/questions/332030/when-should-static-cast-dynamic-cast-const-cast-and-reinterpret-cast-be-used) keep it please.

SU2_CFD/src/solver_direct_mean.cpp
pcarruscag: I would, but a comment saying "Set" on that last "Get" would do just as well

SU2_CFD/include/solver_structure.hpp
vdweide: I moved these methods to CVerificationSolution. Indeed, this makes a lot more sense. @pcarruscag, thanks for the suggestion.

SU2_CFD/src/solver_direct_mean.cpp
vdweide: It is now mentioned clearly in the comments.

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
economon: Aok @vdweide, this is also very logical and just fine for me. Since we are still working on the VandV structure anyway, let's go with posting the scripts that generate the sources directly in the folder you have created. I will also add the python scripts for the incompressible MMS cases with sympy there. Would be good to wrap this one up asap and adjust later only if necessary.

vdweide: @economon, if you add those python scripts we addressed all concerns/remarks as far as I can see it. So it is ready to be merged in.

vdweide: And after I fix an error in the regression tests.....

economon:
Thanks for the feedback and suggestions. Nice teamwork! I am also plenty happy with this now.

Last comment: it was mentioned at the meeting (maybe in the V&V working group), that it would be good to have a separate option for the user-defined solution, in the case of setting a custom initial condition or BC, that sits outside the KIND_VERIFICATION_SOLUTION option list. Doesn't have to necessarily be acted upon now, but want it on record.

vdweide:
Merging this one in.

Issue #148, opened 2015-01-22
Implement/store FFD box info in CGNS mesh files


economon:

WallyMaier:
#674

PR #674, opened 2019-04-26
Feature ffd box in CGNS format

MicK7:
## Proposed Changes
- This PR proposes to add CGNS format as a possible writing format for FFD box
 

## Related Work
- This should address #148 

## PR Checklist


pcarruscag:
Hi @MicK7,
Thank you for the feature, it looks ok to me. But cleaning the indentation may cause a lot of merge issues to others, so I will give it time to see if anyone has issues.
Cheers,
Pedro

Common/src/grid_movement_structure.cpp
```diff
@@ -8441,40 +8457,144 @@ void CFreeFormDefBox::SetSphe2Cart_CornerPoints(CConfig *config) {
...
+  if (nDim==2){
+    std::fill_n(buffer, pointlen, 0.0);
+    cgns_err = cg_coord_write(FFDBox_cgns_file, cgns_base, cgns_zone, CGNS_ENUMV(RealDouble), coord_names[2], buffer, &dummy);
+    if (cgns_err) cg_error_print();
+  }
+
+  delete [] buffer;
+  cgns_err = cg_close(FFDBox_cgns_file);
+  if (cgns_err) cg_error_print();
+#endif
```
pcarruscag: Maybe throw an error if the code was not compiled with CGNS support?

MicK7: Thanks for the review, I added a message as it is done elsewhere in the code.

Common/src/grid_movement_structure.cpp
```diff
@@ -8441,40 +8457,146 @@ void CFreeFormDefBox::SetSphe2Cart_CornerPoints(CConfig *config) {
...
+     dims[2] = nDegree+1;
+  }
+  cgsize_t pointlen = 1;
+  for(int ii=0; ii<cell_dim; ii++){
+    dims[ii+cell_dim] = dims[ii] -1;
+    dims[ii+2*cell_dim] = 0;
+    pointlen *= dims[ii];
+  }
+
+  double *buffer = new double[pointlen];
```
pcarruscag: About the compilation error you are getting, in general in SU2 you should use the typedef "su2double" instead of double or float for floating point variables. The exception is when those variables are then passed to external libraries (which seems to be the case) in which case you should use "passivedouble". Then when assigning to those variables, and the left hand side is of type "su2double" you should enclose the expression in SU2_TYPE::GetValue(), i.e. buffer[pos] = SU2_TYPE::GetValue(Coord_Control_Points[iDegree][jDegree][kDegree][iDim]);

Common/src/grid_movement_structure.cpp
pcarruscag: Thanks. I am not sure if this area of the code is executed in parallel, if it is, the message should only be printed by the MASTER_NODE, see e.g. line 211 of this file.

MicK7: This area of the code is only called by MASTER_NODE and thus not parallel see line 2739 of the file.

pcarruscag: Cool, thanks.

PR #681, opened 2019-05-16
Complete FE (structural) support for 3D + refactoring of CElement + bug fixes

pcarruscag:
## Proposed Changes
In a joint collaboration with @rsanfer we discovered that the Prism and Pyramid shape elements used for mesh deformation had some issues, namely:
- The Prism (6 nodes) element had incorrect integration points, and perhaps due to some inconsistency between the ordering of the nodes and the shape function it had "negative" stiffness (deform opposite to force). The latter issue was remedied by using negative weights... I tried permuting the integration points but that did not help.
- The Pyramid (5 nodes) had inconsistent integration points and shape functions, it was using pyramid points but degenerate hex functions, this made the element some 20 times more flexible than it should be, pyramid shape functions are used now.

My main reason to look into this was to support all element types in the structural solver both for linear and nonlinear analysis, in principle that should be working now, but I still have tests to run.

I took the opportunity to refactor the CElement family of classes, now the derived classes only deal with the shape specifics:
- Integration/extrapolation points, shape function and derivatives, are computed and stored at construction.
- Area/Volume computation in undeformed (reference) or deformed (current) coordinates.

Generic actions like allocating data and computing gradients are done by the base class. For the latter I included one method for 2D and another for 3D, they only differ in the inversion of the Jacobian but being specialized nDim can be "hardcoded" which should allow for more compile time optimization.

I expect some regressions to fail, and I will still run more tests to verify the revised elements. I will post those results and add a regression.

## PR Checklist


Common/include/element_structure.hpp
```diff
@@ -114,12 +134,6 @@ class CElement {
 	 */
 	unsigned short GetnGaussPoints(void);
 
-	/*!
```
pcarruscag: @rsanfer I noticed that everything dealing with pressure subintegration was not being used so I removed it.
I guessed that the initial idea was to have dedicated per-element methods, but eventually was replaced by instantiating lower order elements in the structural solver to deal specifically with the pressure term?
If this was not the case I can put it back.

Common/include/element_structure.hpp
```diff
@@ -415,48 +392,45 @@ class CElement {
...
-  virtual su2double ComputeVolume(void);
+  virtual su2double ComputeVolume(const FrameType mode = REFERENCE);
 
   /*!
-   * \brief Compute the value of the area of the element in current coordinates
+   * \brief Compute the value of the area of the element in current coordinates (wrapper to ComputeArea(CURRENT)).
    * \param[out] val_Area - Area of the element
    */
-  virtual su2double ComputeCurrentArea(void);
+  su2double ComputeCurrentArea(void);
```
pcarruscag: This is now just short-hand to calling ComputeArea with "CURRENT" as argument, this keeps the change from spreading to other areas of the code but one could always drop the ComputeCurrent* methods...

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -157,22 +157,26 @@ CFEASolver::CFEASolver(CGeometry *geometry, CConfig *config) : CSolver() {
...
-        element_container[DE_TERM][EL_HEXA] = new CHEXA8(nDim, config);
+        element_container[DE_TERM][EL_HEXA]  = new CHEXA8 (nDim, config);
+        element_container[DE_TERM][EL_PYRAM] = new CPYRAM5(nDim, config);
+        element_container[DE_TERM][EL_PRISM] = new CPRISM6(nDim, config);
       }
 
       if (incompressible) {
         element_container[INC_TERM][EL_TETRA] = new CTETRA1(nDim, config);
-        element_container[INC_TERM][EL_HEXA] = new CHEXA1(nDim, config);
+        element_container[INC_TERM][EL_HEXA]  = new CHEXA1 (nDim, config);
```
pcarruscag: Incompressible will not work with Pyramids and Prisms for now but an error is thrown if someone tries it.

Common/src/grid_movement_structure.cpp
```diff
@@ -1477,13 +1477,15 @@ void CVolumetricMovement::SetFEA_StiffMatrix3D(CGeometry *geometry, CConfig *con
...
+          that causes "negative" stiffness, the remedy is to use negative weights. ---*/
     nGauss = 6;
-    Location[0][0] = 0.5;                 Location[0][1] = 0.5;                 Location[0][2] = -0.577350269189626;  Weight[0] = 0.166666666666666;
-    Location[1][0] = -0.577350269189626;  Location[1][1] = 0.0;                 Location[1][2] = 0.5;                 Weight[1] = 0.166666666666666;
-    Location[2][0] = 0.5;                 Location[2][1] = -0.577350269189626;  Location[2][2] = 0.0;                 Weight[2] = 0.166666666666666;
-    Location[3][0] = 0.5;                 Location[3][1] = 0.5;                 Location[3][2] = 0.577350269189626;   Weight[3] = 0.166666666666666;
-    Location[4][0] = 0.577350269189626;   Location[4][1] = 0.0;                 Location[4][2] = 0.5;                 Weight[4] = 0.166666666666666;
-    Location[5][0] = 0.5;                 Location[5][1] = 0.577350269189626;   Location[5][2] = 0.0;                 Weight[5] = 0.166666666666666;
+    Location[0][0] = -0.577350269189626;  Location[0][1] = 0.166666666666667;  Location[0][2] = 0.166666666666667;  Weight[0] = -0.166666666666667;
```
pcarruscag: These integration points are based on Calculix v2.15 instead of the original reference mentioned in the code.

Common/src/grid_movement_structure.cpp
```diff
@@ -799,35 +799,35 @@ su2double CVolumetricMovement::ShapeFunc_Pyram(su2double Xi, su2double Eta, su2d
   
   /*--- Shape functions ---*/
   
-  DShapeFunction[0][3] = 0.125*(1.0-Xi)*(1.0-Eta)*(1.0-Zeta);
-  DShapeFunction[1][3] = 0.125*(1.0+Xi)*(1.0-Eta)*(1.0-Zeta);
-  DShapeFunction[2][3] = 0.125*(1.0+Xi)*(1.0+Eta)*(1.0-Zeta);
-  DShapeFunction[3][3] = 0.125*(1.0-Xi)*(1.0+Eta)*(1.0-Zeta);
-  DShapeFunction[4][3] = 0.5*(1.0+Zeta);
+  DShapeFunction[0][3] = 0.25*(-Xi+Eta+Zeta-1.0)*(-Xi-Eta+Zeta-1.0)/(1.0-Zeta);
```
pcarruscag: These shape functions are now consistent with the reference cited in the code:
https://www.code-aster.org/V2/doc/v13/en/man_r/r3/r3.01.01.pdf

Common/src/element_linear.cpp
```diff
@@ -621,1526 +203,260 @@ su2double CQUAD4::ComputeArea(void){
...
+    NodalExtrap[iNode][4] = 0.125*(1.0-Xi)*(1.0-Eta)*(1.0+Zeta);
+    NodalExtrap[iNode][5] = 0.125*(1.0+Xi)*(1.0-Eta)*(1.0+Zeta);
+    NodalExtrap[iNode][6] = 0.125*(1.0+Xi)*(1.0+Eta)*(1.0+Zeta);
+    NodalExtrap[iNode][7] = 0.125*(1.0-Xi)*(1.0+Eta)*(1.0+Zeta);
   }
 
 }
 
-su2double CHEXA1::ComputeVolume(void){
+su2double CHEXA8::ComputeVolume(const FrameType mode){
```
pcarruscag: @rsanfer would it be worth computing the volume the "finite element way" instead of using this "finite volume way"? I think these methods are only used by the mesh deformation but nevertheless it would be the generic way to do it (I have no idea about the performance - accuracy trade-off here though).

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -1261,10 +1265,13 @@ void CFEASolver::Compute_StiffMatrix_NodalStressRes(CGeometry *geometry, CSolver
     if (geometry->elem[iElem]->GetVTK_Type() == TRIANGLE)      {nNodes = 3; EL_KIND = EL_TRIA;}
     if (geometry->elem[iElem]->GetVTK_Type() == QUADRILATERAL) {nNodes = 4; EL_KIND = EL_QUAD;}
     if (geometry->elem[iElem]->GetVTK_Type() == TETRAHEDRON)   {nNodes = 4; EL_KIND = EL_TETRA;}
-    if (geometry->elem[iElem]->GetVTK_Type() == PYRAMID)       {nNodes = 5; EL_KIND = EL_TRIA;}
-    if (geometry->elem[iElem]->GetVTK_Type() == PRISM)         {nNodes = 6; EL_KIND = EL_TRIA;}
+    if (geometry->elem[iElem]->GetVTK_Type() == PYRAMID)       {nNodes = 5; EL_KIND = EL_PYRAM;}
+    if (geometry->elem[iElem]->GetVTK_Type() == PRISM)         {nNodes = 6; EL_KIND = EL_PRISM;}
     if (geometry->elem[iElem]->GetVTK_Type() == HEXAHEDRON)    {nNodes = 8; EL_KIND = EL_HEXA;}
     
+    if (incompressible) if (element_container[INC_TERM][EL_KIND] == NULL)
```
WallyMaier: I am a fan of these error messages!

pcarruscag: Thanks! They save some time sometimes.

pcarruscag:
Verification results [here](https://github.com/su2code/SU2/files/3192519/verif.pdf) along with pictures of the failed testcases, 5/6 new reference residuals are lower than before.

Funny story the prism elements do not need negative weights in CVolumetric movement... Only the elasticity solver seems to need this, I tested both with a mesh with mixed elements. I have no explanation for this... the code is the "same" in both places.

Common/src/element_linear.cpp
```diff
@@ -2609,118 +772,75 @@ CPRISM6::CPRISM6(unsigned short val_nDim, CConfig *config)
...
-      /*--- dN/d xi ---*/
+  su2double ExtrapCoord[6][3];
 
-      dNiXj[0][0] = -0.5*Eta;
-      dNiXj[1][0] = -0.5*Zeta;
-      dNiXj[2][0] = -0.5*(1.0-Eta-Zeta);
-      dNiXj[3][0] = 0.5*Eta;
-      dNiXj[4][0] = 0.5*Zeta;
-      dNiXj[5][0] = 0.5*(1.0-Eta-Zeta);
+  ExtrapCoord[0][0] = -1.732050807568877;  ExtrapCoord[0][1] = -1.0/3.0;  ExtrapCoord[0][2] = -1.0/3.0;
```
rsanfer: As in other places in the code, a su2double Sqrt3 variable could be useful here.

Common/include/element_structure.hpp
pcarruscag: Yep that's right

SU2_CFD/src/solver_direct_elasticity.cpp
pcarruscag: Ok, I can do that, there are no implications to the other (than neo-Hooken) material models right?

rsanfer: No, I think you would only have to remove the NeoHookean_Incomp. The Knowles model didn't use the under-integrated elements and is listed under NEARLY_INCOMPRESSIBLE_MAT, so it doesn't use the INC_TERM elements. That should simplify the code - and if someone wants to take it from there, we always have the older versions.

pcarruscag:
Thank you for the review @rsanfer.
I put the extra comments and "centralized constants" in the code.
Everything relating to the incompressible neo-Hookean material has been deleted, let me know if it was too much.
I'll put a testcase together with a healthy mix of elements, probably next week.
Cheers,
Pedro

Common/include/element_structure.hpp
pcarruscag: For the mesh solver it would be very efficient to compute the volume the FE way (sum Ji*Wi) since the gradients need to be computed anyway.

rsanfer: Yeah I also think so. I hope to be able to open the PR on the mesh solver soon(ish - next month and a half I will be mostly travelling), and maybe then we can give it a shot.

rsanfer:
Thanks for the changes, @pcarruscag. I like PRs like this with 797 additions and 3220 deletions maintaining functionality :+1: The incompressible part of the FEA problem was, like some other features in the code, 90% ready but never totally finished, so to avoid confusion in users and developers and ease maintenance I think it's better for it to be removed until someone can take over and finalize the implementation.
Let me know when you've had some time to put that testcase, and we can get this merged in.
Cheers,
Ruben

pcarruscag:
@rsanfer testcase is in. I used the Knowles material model, I don't think there was a regression for it.
Cheers,
Pedro

rsanfer:
Hi @pcarruscag,
thanks for adding the test case. This LGTM, from my side is ready. Let's allow as common practice some 24 h to see if anyone has additional comments, then we can merge it in.
Best,
Ruben

Issue #81, opened 2014-10-01
Rename CNumerics to CSpatialScheme


aniketaranake:
The word "numerics" applies to most of what SU2 does including time integration. For this reason, a class called CNumerics that only relates to spatial discretization is confusing. 

Considering the routines which inherit from this parent class (CUpwRoe_Flow, CAvgGradCorrected_Flow, etc.), I suggest as a new name CSpatialSchemes

WallyMaier:
@economon is this still something we should pursue?  If so, I can work on it.

economon:
This is related to #691. Perhaps we can bring any renaming there for discussion, including the potential division of files.

PR #691, opened 2019-05-27
Refactor convective numerics

pcarruscag:
## Proposed Changes
Hi Folks,
I would like your views on something. Many of the convective schemes we have are small variations of each other, and so I was thinking we can maybe have some intermediate numerics classes to make things a bit easier to maintain, quicker to compile etc.
So far I experimented with having a base class for central schemes, the ComputeResidual method is moved to this class and the children classes (Lax, JST, and JST_KE) only implement the artificial dissipation part specific to them.
Similarly AUSM+up(1/2) and SLAU(1/2) only differ in how the mass and pressure fluxes are computed so again a common base for these 4 could be devised. Isolating the computation of mass and pressure fluxes could be an interesting way to compute the Jacobians of these schemes in a hybrid way (currently the Roe Jacobian is used).
There is a very small performance penalty (maybe 1-2% for explicit solution methods) but my reason to be looking at the numerics in the first place is that with some small tweaks to the Jacobians we can probably run most schemes at higher CFL (namely increase the dissipative part of the Jacobian to make the system matrix more diagonal dominant) so for implicit solution methods there would be a speedup.
So, what do you think?
Cheers,
Pedro

## PR Checklist


WallyMaier:
Hi Pedro, 
Thanks for this! I think eliminating duplicated code is a good idea.  That being said, would this make adding new numerical schemes more difficult? (Thinking from a new developer standpoint).

pcarruscag:
Hi Wally,
I think that would depend, if the new scheme is part of a family that already exists it should be easier to implement, if not you do it from scratch... There is also the possibility of making life hard by trying to find common ground where there is none, e.g. the way CUSP is implemented in SU2 has something in common with the central schemes, but it did not feel right to attach an upwind scheme to the central family.

aeroamit:
Hi Pedro,

It is indeed a good idea to keep similar schemes (with minor variations) under one umbrella,  especially  central scheme as only the dissipation term calculation differs. 

- In case of AUSM+-up1/2 and SLAU1/2, each can be reduced by having a separate pressure flux definition.

- But if we try to combine AUSM+-up and SLAU, will it be a clear/helpful implementation as mass flux definitions/calculations (which is substantial portion) are different (unlike the central scheme, where only dissipation term varies)?

- Also could you please shed some light on Jacobian part for these schemes (as you mentioned - “...Isolating the computation of mass and pressure fluxes could be an interesting way to compute the Jacobians of these schemes in a hybrid way ...”).

Best
Amit

pcarruscag:
Hi Amit,
The idea is basically this:
IMG_20190605_224411
F is the general form of the flux per unit area for those 4 schemes (according to the AUSM+-up2 paper).
The Jacobians of PSI with respect to the conservatives (U) are known analytically, the derivatives of the mass and pressure fluxes are harder to derive, so if we encapsulated their calculation in a function ( [m,p] = f(U) ) we could approximate them by finite differences (or even use AD).
So the base class would deal with the general form and Jacobians, the derived classes would implement this function for m and p.
Cheers,
Pedro

aeroamit:
Hi Pedro,
Thanks for sharing the nice idea. If it helps in ramping up the CFL than really good....
One query - generally how Jacobian computation by finite difference or AD tend to be in comparison to analytical Jacobians in terms of cost (time). Probably Analytical jacobians are more efficient but they are difficult to compute for a given scheme (is this mostly true?). 

Thanks
Amit

pcarruscag:
Hi Amit,
That is also my intuition, I guess the only way to know if it is worth the extra cost is by doing.
Do you have a good supersonic case I can use for testing (all my work is subsonic)? Preferably something that converges well without initialization.
Cheers,
Pedro

aeroamit:
Hi Pedro,

SU2 Testcase repo has 3 supersonic TestCases under euler folder (wedge, biparabolic and bluntbody), they all go well without initialization (you may be already aware of these euler cases).

But I do not have any specific supersonic case for RANS or Laminar flow problems (some of them ideally be like nozzle flow or SHOCK-BL interaction).
May be need to have a look at NASA CFD site for validation and verification cases.

Best
Amit

pcarruscag:
Ok, I will start with those when I have a minute.
What about you @WallyMaier, do you have any reference supersonic RANS case?
Pedro

WallyMaier:
@pcarruscag @aeroamit I do not have a good test case.  But am currently looking for one.   Let me know if y'all find a good.

economon:
All for improving the abstractions if possible. We did something similar for the scalar upwinding and the viscous schemes already. 

My only comments, which you can probably guess already, are to make sure we don't take a large performance hit by adding more layers (for example, one can make an argument to remove the entire set of CNumerics classes and implement the methods more efficiently directly in the solver class with vectorization, etc) and that we maintain the flexibility for folks to quickly add new convective schemes which was the original motivation for the existing structure (and more layers could complicate this). Sounds like you're already considering these things, but it is important to find the right balance.

pcarruscag:
Right, that is the AUSM-family refactoring done with the hybrid Jacobians, Roe approximation is still available and it is the default.
So far:
- It was not practical to do finite differences of the mass flux and pressure by perturbing the conservatives, I had to do it on the primitives which then requires an extra Jacobian... So the approach is more expensive per iteration, which could be partially offset by using vector-mode forward-AD.
- However for some cases convergence is much better for the same CFL (wedge case with CFL=25) and in others (transonic airfoil, SST, low Re mesh) it allows much higher (4x) CFL to be used, which again results in lower time overall.
- The performance penalty from the extra layers does not seem to be big, but I have to measure that in a machine that is not doing other things at the same time.

WallyMaier:
Another possible thing to consider with this pull request is a possible future renaming of numerics.  issue #81 mentions this but I wanted to bring it up here to discuss whether we think this a sold plan, or should it be disregarding.

pcarruscag:
Yeah, I saw that... Will spacial scheme still be appropriate when other physics become involved?

WallyMaier:
Maybe not, but possibly a renaming to CFluxes or something a little more descriptive?

pcarruscag:
Yes that would be descriptive of fluid applications, but the numerics of the structural solver are a mixture of kinematics and constitutive law (and there are no fluxes there).
I do not think it is worth making changes to the entire code because of an issue opened 5 years ago that no one else has complained about... It is an abstract base class let it have an abstract name.

WallyMaier:
Great! Sounds good, just figured id bring it up to get some of these ancient issue cleared up. :)

aeroamit:
Cool job Pedro,
This (higher CFL) will be really helpful with these latest schemes. 

Cheers
Amit

pcarruscag:
[Here](https://github.com/su2code/SU2/files/3291497/report.pdf) are some promising results, I took a transonic turbulent airfoil case I had, and a couple supersonic ones from the testcase repo (biparabolic and wedge). I kind of random selected the numerical scheme for each one and tried to make the settings fair. [These](https://github.com/su2code/SU2/files/3291494/files.zip) are the required files should anyone want to replicate the findings.
@aeroamit Thank you. I do not have much interest in low Mach applications and I am not familiar with the required preconditioning techniques so I am sorry but I will not be running that test myself.

I have yet to test the discrete adjoint now that preaccumulation is available for these AUSM-family schemes.

aeroamit:
Good to see the results. It will really help with schemes having inconsistent discretization for the implicit part. 

One of the interesting thing (slide 3) is allowable CFL change with change in number of cores. 
What could be the possible logic behind that?

Regards
Amit

pcarruscag:
Hi Amit,
That was my initial feeling too, all our linear preconditioners (except Jacobi) suffer with number of partitions, but the linear solver was still able to converge, and to even lower tolerance if needed.
The case would not blow up immediately, it converges nicely 4 orders of magnitude and then diverges, maybe some oscilations develops somewhere, but I do not see any point in the domain that stands out... I tried using GMRES instead of BCGSTAB, and converging the linear solution more, and using more MG residual smoothing cycles... I blame this mostly on the grid being too stretched.
Cheers,
Pedro

EduardoMolina:
Hi @pcarruscag ,

I will run a low-Mach case and I will post the results here.

Eduardo

aeroamit:
Hi @pcarruscag , All



I have done a test case for NACA 0012 airfoil at Mach 0.01 to check the effectiveness of hybrid jacobians (as requested by me earlier).
There are some interesting observations. Check out the attachment.
[Low_Speed_test.pdf](https://github.com/su2code/SU2/files/3303753/Low_Speed_test.pdf)
[cfg_files.zip](https://github.com/su2code/SU2/files/3303763/cfg_files.zip)

Regards
Amit

pcarruscag:
Hi Amit,
Thank you for testing the feature on those cases, I would never have guessed the convergence rate could be higher with Roe Jacobians. I got curious and played with the case too, I tried reducing the finite difference step and changing linear solvers but it made no difference.
It is especially intriguing that for two of the schemes Roe needs CFL of 0.01 (at which point it would possibly better to use explicit RK) and for the other two schemes of the same family the sky is the limit. I guess one of the conclusions is that it is good to have options.
Pedro

economon:
@pcarruscag: if I have not missed something in the changes you propose, it looks like you are including a numerically exact Jacobian via finite differences that is consistent with the corresponding schemes (instead of using Roe w/ AUSM, for instance). I would definitely expect this to give an improvement in convergence for the majority of cases.

Though, I think we will still end up with a more accurate first-order approximation to the Jacobian, unless the neighbors of neighbors are considered due to the gradient terms in the upwind reconstruction. Have you considered an exact Jacobian consistent with our 2nd-order schemes? This could require a larger memory footprint, or it could be done with matrix-free methods. We have thrown this idea around several times before and even considered using AD. Not recommending it for this PR, but it could be interesting to have in our toolbox eventually.

pcarruscag:
@economon that is what I propose for AUSM+up(2) and SLAU(2). Unfortunately their predecessor AUSM does not fit into the same generic form and so it will not have access to these numerical Jacobians.

On the subject of accurate Jacobians, IMO matrix free is definitely the way to go (not just because of storage but because it would possibly require changing the edge loops, I did something similar in the past and it got messy).
Using forward AD to linearize the entire iteration and create some sort of Newton-nonlinear-GMRES monster solver preconditioned by the current 1st order Jacobians would indeed be interesting... I have no feeling for potential speed-up though.

aeroamit:
Hi,

@pcarruscag , I tried few more runs (Roe Jacobians with SLAU (2)) with this case and they ran well at CFL 500 for Mach 0.1 to 0.3 at M=0.4, it dips to ~400 and I think that trend will continue downward. I ran it for Mach 0.8 and CFL was restricted to ~25.0.

So it’s a very surprising/peculiar behaviour observed with this case and as you rightly said that it is good to keep both the options. 

@economon , can you share any experience of first vs second order analytical Jacobians (personal work, references/papers). Working out a second order Jacobian will be a really arduous task (and there can be other specific issues related to specific schemes as well). Consistent discretization with 1st order analytical Jacobian is likely to be better than inconsistent discretization for both 1st and 2nd  order accurate simulations (in general). I was curious to look at any reference showing the real/comparative gains one will have with second order accurate analytical Jacobians.

Another thing, many of the commercial codes use different strategies coupled with good linear solvers to have really good convergence rates (probably I mentioned this earlier as well).

Thanks and Regards
Amit

aeroamit:
One of the interesting paper -

" Comparison of numerical and Analytical Jacobians", Kirk J. Vanden, Paul D. Orkwis
AIAA, Vol 34, No. 6, June 1996.

They computed the exact analytical Jacobian with symbolic manipulation.

In conclusion they are showing that both analytical and numerical Jacobians showed similar performance and suggesting that for simpler numerical fluxes, analytical Jacobians should be the best way to go and for complex numerical fluxes, numerical Jacobian can be preferable choice (but if one can work out analytical, that should be good as well, I guess (one time effort) ).

Best
Amit

pcarruscag:
I am glad you mention that @aeroamit, I experimented with analytical Jacobians for the AUSM+up scheme (latest commits). Indeed there does not seem to make much difference for convergence, but it is computationally cheaper than the finite difference approach.

aeroamit:
@pcarruscag, Did you work out the analytical Jacobian for AUSM+-up? (holy cow).

I have also worked out and implemented in my local machine  for AUSM, up and up2 (with minor simplifactions/assumptions, wherever seemed okay) . I was planning to consolidate things appropriately and refine it before pushimg them to repo.
By the way, I am also observing the similar behavior with little time advantage for analytical part (a bit more refinement/cleaning in implementation from my side will be done) 

So what should be the next plan ???

Cheers
Amit

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -327,1435 +468,1013 @@ void CCentJST_KE_Flow::ComputeResidual(su2double *val_residual, su2double **val_
...
-   
   /*--- Face area (norm or the normal vector) ---*/
+
   Area = 0.0;
   for (iDim = 0; iDim < nDim; iDim++)
     Area += Normal[iDim]*Normal[iDim];
   Area = sqrt(Area);
-  
+
   /*-- Unit Normal
```
pcarruscag: @aeroamit I did and it was more troublesome than I expected. The differentiation of the mass flux and pressure start here and end in line 1299, unfortunately git is quite confused with the changes, look only at the green chunks.
I modified the "framework" such that derived classes can set the derivatives of mdot and pressure with respect to the velocities, pressures, densities, and enthalpies. By doing so the finite differences are bypassed.

If you have those quantities worked out for the other schemes feel free to push to this PR, I am not planning to do the other schemes manually :)

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -2613,1427 +2348,954 @@ if (sM > 0.0) {
...
   grid_movement = config->GetGrid_Movement();
   kappa = config->GetRoe_Kappa(); // 1 is unstable
-  
+
   Gamma = config->GetGamma();
   Gamma_Minus_One = Gamma - 1.0;
   
+  roe_low_dissipation = val_low_dissipation;
+  
   Diff_U = new su2doub
```
pcarruscag: The Bogeyman came for the Roe classes. Roe, L2Roe, and LMRoe now share a parent that implements the `fc_{1/2} = kappa*(fc_i+fc_j)*Normal` part, the rest is then done by the derived classes. And with this I concluded this type of refactoring.

SU2_CFD/src/numerics_direct_mean.cpp
aeroamit: Hi @pcarruscag ,

Things look good here (+bonus fruitful learning). I have worked out the Jacobians with respect to conservative variables.  I will do the needful and take a while to push to the online repository (some work and local formalities etc.)

Cheers
Amit

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -599,190 +359,151 @@ CUpwCUSP_Flow::CUpwCUSP_Flow(unsigned short val_nDim, unsigned short val_nVar, C
...
+    else if (Beta < 0.0) Nu_c = (1.0-Beta)*LamdaPos;
+    /*--- Limit the minimum scalar dissipation ---*/
+    else Nu_c = max(fabs(ProjVelocity), config->GetEntropyFix_Coeff()*MeanSoundSpeed);
+  }  
   
   /*--- Compute the residual ---*/
   
   for (iVar = 0; iVar < nVar; iVar++)
-    val_residual[iVar] += (0.5*Nu_c*Diff_U[iVar] + 0.5*Beta*Diff_Flux[iVar])*Area;
+    val_residual[iVar] = 0.5*((1.0+Beta)*ProjFlux_i[iVar] + (1.0-Beta)*ProjFlux_j[iVar] + Nu_c*Diff_U[iVar])*Area;
```
pcarruscag: As discussed in #705, according to the literature the CUSP flux should be expressed as the average of the left and right fluxes plus the artificial diffusion part, and not as a flux of averages + diffusion. This way the scheme becomes 100% upwind above Mach 1, and as a bonus the implementation also becomes more efficient.

Whatever I could do to reduce duplication and improve convergence is done, this PR is ready for reviews.

EduardoMolina:
Hi Pedro,

Do you know why this PR is failing during Travis? I looked at the test cases and both are unsteady using Roe? 
BTW, I will be happy to review this PR.
Eduardo

pcarruscag:
Hi Eduardo,
The channel 2D and supersonic vortex shedding cases are failing due to small changes in residuals, the operations in Roe were re-ordered a bit, and these cases run for many iterations, I am guessing it is just due to numerical precision. But I will have a look before updating the residuals.
The rotating cylinders case is the infamous one that appears to be sensitive to minute, and seemingly insignificant, changes as reported in #700, at the reference residuals the case is actually diverged.
Finally there is an AD case using JST that also fails because it is a file diff and we output 15 significant digits to that file. The change is due to the refactoring of central schemes that required changing the order in which variables are registered for pre-accumulation.
I do appreciate you taking the time to review this PR, especially since you are "first author" of the low Mach Roe schemes and there are no regressions for them, nevertheless I tested them before and after and it seems I did not break anything.
Cheers,
Pedro

pcarruscag:
@EduardoMolina 
I ran the [channel_2D](https://github.com/su2code/SU2/files/3346960/channel_2D.zip) and the
[supersonic_vortex_shedding](https://github.com/su2code/SU2/files/3346962/supersonic_vortex_shedding.zip) testcases with a tighter tolerance on the linear solver to make sure the differences were not due to some change in number of iterations.
Some differences do accumulate over time but they are very small, for the channel_2D the final density field differs by 1e-6 at most (which is the output precision).
For the supersonic vortex shedding differences are greater, 2e-4 on average and up to 0.01 maximum, but the case is also larger and runs for longer. If I restart the case for a 6th time-step (from the same restart files) the initial residuals are the same.
I will update the other two cases when #700 gets merged as they changed there too.
Cheers,
Pedro

EduardoMolina:
Hi @pcarruscag 
I am also running a couple of test cases, mainly the RANS NASA Hump. Do you think it is a good idea to add since we don't have any test case covering the low Mach Roe implementation? If so, I am happy to add.
Best,
Eduardo

pcarruscag:
Hi @EduardoMolina ,
Thanks! Let me know how it goes.
I think having coverage is important but maybe it would be better to re-purpose an existing case since in our regressions Roe/JST are over-sampled and the regressions already take quite a bit of time to run.
Cheers,
Pedro

EduardoMolina:
Hi @pcarruscag ,

Sorry for the late response. 

I went through all your implementation and performed some tests. 

Although, I didn't see an improvement using the accurate jacobians in the subsonic test cases, as already mentioned here, and even in the transonic OneraM6 case from the repo. This implementation is very clean and in my opinion is a great improvement. 

I hope that in other cases, the use of accurate jacobian will have a positive effect in the convergence.

The only thing that I would like to bring is that in the future, you open a PR from su2 repo instead from your personal account. It is easier for the reviewers to compile and test.

I will wait Travis pass to merge this PR.

Thanks again.

Eduardo

pcarruscag:
Thank you @EduardoMolina.
In the future I will make sure the branch exists in the SU2 repo.
Yes unfortunately the accurate Jacobians for SLAU and AUSM+ are not silver bullets and their success seems to be very case dependent.
The rotating_cylinders case stopped diverging after merging the changes from #700 so I have updated all reference residuals.

aeroamit:
Hi @pcarruscag, All, 

It's good to see one of this important PR coming to completion. It was a fruitful interaction, and I got to know more stuff.
I am not aware of detailed review process for SU2 PRs, and would like to know more details (I certainly go along with the fact that any member should be eligible for specific case/topic to review at first place, but useful details and ideas can be shared at any point - correct me if I missed something).
I could see the code changes for this PR in bits and pieces as I didn't have my lappy with me for nearly last two weeks and it's difficult/unpleasant on smartphone. Anyway, I will go through it in detail and extract the juice from it :) . 

Thanks and Regards 
Amit

EduardoMolina:
Hi @pcarruscag and @aeroamit ,

Thanks for the discussion and for share the ideas/results. I think this is the beauty of open-source, everyone is welcome to jump in and review all the pull requests. Certainly, as you said @aeroamit, we will all learn something fruitful when we review and merge each PR.

Best regards,

Eduardo

Issue #38, opened 2014-06-27
SU2 can keep running with ridiculus residuals


btracey:
Sometimes, the code effectively diverges, but does not actually nan. In this case, the code continues to run. There should be some sort of check that halts execution when this occurs.

For example, here the code has a residual increase of 25 orders of magnitude and then continues for a long time afterward. 

  514   0.858170     -3.810382     -0.818090       0.534918       0.011557
  515   0.858110     -3.806673      1.276861       0.535673       0.011433
  516   0.858248     -3.523170      2.017555       0.533807       0.010890
  517   0.858833     -3.146589      2.453353       0.529337       0.009891
  518   0.859031     -2.980286      3.070260       0.523952       0.009211
  519   0.859039     -2.637185      3.740419       0.455339      -0.003290
The solution contains 100 non-physical points.

 Maximum residual: -0.0373748, located at point 4582.

 Iter    Time(s)      Res[Rho]       Res[nu]   CLift(Total)   CDrag(Total)
  520   0.859013     -1.437953      4.373667      -0.099686      -0.094575

The solution contains 127 non-physical points.
  521   0.859003     -1.111074      5.056160      -4.065337      -0.780464
The solution contains 284 non-physical points.
  522   0.858853     -0.693782      5.183287     -15.761187      -2.268335
The solution contains 514 non-physical points.
  523   0.858725     -0.072830      5.147720     -60.615908      -9.156773
The solution contains 691 non-physical points.
  524   0.858711      0.730028      4.462988    -358.683565     -69.148449
The solution contains 1434 non-physical points.
  525   0.858921      1.373102      4.241595    -914.744092    -187.067270
The solution contains 2138 non-physical points.
  526   0.859282      1.713954      3.814762   -1000.000000    -358.772646
The solution contains 1918 non-physical points.
  527   0.859916      2.236290      4.340435   -1000.000000   -1000.000000
The solution contains 3154 non-physical points.
  528   0.860132      2.291044      4.470125   -1000.000000   -1000.000000
The solution contains 2146 non-physical points.
  529   0.860356      2.371124      3.845344   -1000.000000   -1000.000000
The solution contains 1962 non-physical points.
  530   0.860386      2.282541      4.230209   -1000.000000   -1000.000000
The solution contains 2372 non-physical points.
  531   0.860488      2.346939      4.654921   -1000.000000   -1000.000000
The solution contains 2496 non-physical points.
[...]

PR #697, opened 2019-06-04
adding large residual check

WallyMaier:
## Proposed Changes
This adds a check to prevent SU2 continuing diverge/reach extremely large residuals.  This seems to be a feature request in both the SU2 issues (#38) and on the CFD-Online forum.   

This is similiar to the divergence criteria but is activated at residuals greater than 10^12 . This number can change, it was arbitrarily chosen. 
 

pcarruscag:
Hi @WallyMaier,
In most of my cases the initial residual for SST is ~10 and sometimes it goes over 15 before starting to drop, probably a normalization thing... anyway, can you make it 20?
Cheers,
Pedro

WallyMaier:
@pcarruscag done deal

SU2_CFD/src/solver_structure.cpp
```diff
@@ -1,4 +1,4 @@
-/*!
+﻿/*!
  * \file solver_structure.cpp
  * \brief Main subrotuines for solving direct, adjoint and linearized problems.
```
economon: Could you please fix that old typo for ‘subroutines’ in the \brief? Also, we don’t solve any linearized gov eqns any longer, so perhaps adjust to ‘primal and adjoint problems’

WallyMaier: Done

PR #700, opened 2019-06-05
CSysMatrix cleanup and performance improvements

pcarruscag:
## Proposed Changes
I made this in micro steps and the commit messages are fairly detailed, so here I will give only the highlights. Every change is mathematically equivalent.
I claim performance improvements based on the type of cases I run (steady RANS implicit, elasticity), results may vary, so please give this a try.

**Cleanup**
- Iterative smoothers (Jacobi, ILU, etc.) are now implemented in CSysSolve as one generic method with the same interface as the Krylov solvers.
- Dead code was removed from CSysMatrix (as requested in #653).
- Duplicated code, mostly dealing with block-block/vector multiplication, was merged.
- Row/column elimination tasks implemented in CSysMatrix (as requested in #658).
- Linelet uses fewer working variables.
- Make a lot of methods private (the diff of matrix_structure.hpp is not going to be pretty...).

**Performance**
- Inlined small methods.
- No calls to "Get/SetBlock" when the block location is already known (helps ILU and LU_SGS).
- Block inversion done in one go instead of multiple Gaussian eliminations (which used to require repeated upper matrix transformations) (helps Jacobi, Linelet, and ILU).
- Inverted ILU diagonal blocks are stored during "build" for use in "compute".

**Bugs**
- Linelet preconditioner was not working for multiple wall markers.

**ToDo**
- I will try to make the MKL optimizations work for the discrete adjoint (I say try because it may require too much creativity with the templates...).
- Get some benchmarks (I don't want you to just take my word for it).

## Related Work
#650, #653, #658

## PR Checklist


SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -2009,82 +2009,35 @@ void CFEASolver::BC_Clamped(CGeometry *geometry, CSolver **solver_container, CNu
...
+      node[iPoint]->SetSolution_Accel(Solution);
+      node[iPoint]->SetSolution_time_n(Solution);
+      node[iPoint]->SetSolution_Vel_time_n(Solution);
+      node[iPoint]->SetSolution_Accel_time_n(Solution);
     }
-    
+
+    LinSysSol.SetBlock(iPoint, Solution);
+    LinSysReact.SetBlock(iPoint, Solution);
+    Jacobian.EnforceSolutionAtNode(iPoint, Solution, LinSysRes);
```
economon: Nice - thanks for following up on this w/ the EnforeSolutionAtNode() method

Common/src/matrix_structure.cpp
```diff
@@ -1457,327 +1049,35 @@ void CSysMatrix<ScalarType>::MatrixVectorProductTransposed(const CSysVector<Scal
...
-  
-  /*--- MPI Parallelization ---*/
-  
-  InitiateComms(prod, geometry, config, SOLUTION_MATRIX);
-  CompleteComms(prod, geometry, config, SOLUTION_MATRIX);
-  
-}
-
-template<class ScalarType>
-unsigned long CSysMatrix<ScalarType>::Jacobi_Smoother(const CSysVector<ScalarType> & b, CSysVector<ScalarType> & x, CMatrixVectorProduct<ScalarType> & mat_vec, ScalarType tol, unsigned long m, ScalarType *residual, bool monitoring, CGeometry *geometry, CConfig *config) {
```
economon: Thanks for collapsing all of my old smoothers into a unified implementation. Can some of the comments here be preserved in the new implementations under the Compute*Preconditioner() routines? I recall adding many of them to explain step by step what each of the smoothers were doing...

Common/src/linear_solvers_structure.cpp
```diff
@@ -651,6 +652,95 @@ unsigned long CSysSolve<ScalarType>::BCGSTAB_LinSolver(const CSysVector<ScalarTy
   return (unsigned long) i;
 }
 
+template<class ScalarType>
+unsigned long CSysSolve<ScalarType>::Smoother_LinSolver(const CSysVector<ScalarType> & b, CSysVector<ScalarType> & x,
+                                                        CMatrixVectorProduct<ScalarType> & mat_vec, CPreconditioner<ScalarType> & precond,
+                                                        ScalarType tol, unsigned long m, ScalarType *residual, bool monitoring, CConfig *config) {
+
+  int rank = SU2_MPI::GetRank();
+  ScalarType norm_r = 0.0, norm0 = 0.0, omega = 1.0; // hardcoded relaxation factor
```
economon: Have you investigated the effects of the relaxation factor? When I first put this in, I included a variable for it, but I am not sure I tested it much. Would be interested to see.

Also might be worth exposing this to the config and/or making it consistent with the relaxation factors appearing in other linear solvers.

Common/include/matrix_structure.hpp
```diff
@@ -280,127 +248,90 @@ class CSysMatrix {
...
-  template<class OtherType>
-  void SetVal2Diag(unsigned long block_i, OtherType val_matrix);
+  void UpperProduct(const CSysVector<ScalarType> & vec, unsigned long row_i);
   
   /*!
-   * \brief Calculates the matrix-vector product
-   * \param[in] matrix
-   * \param[in] vector
-   * \param[out] product
+   * \brief Performs the product of i-th row of the lower part of a sparse matrix by a vector.
```
economon: Thanks for attention to detail in the comments

Common/src/matrix_structure.cpp
```diff
@@ -1457,327 +1049,35 @@ void CSysMatrix<ScalarType>::MatrixVectorProductTransposed(const CSysVector<Scal
...
-    
-    /*--- Compute the i-th column of the inverse matrix ---*/
-    
-    Gauss_Elimination(block_i, aux_vector, transpose);
-    for (jVar = 0; jVar < nVar; jVar++)
-      invBlock[jVar*nVar+iVar] = aux_vector[jVar];
-  }
+void CSysMatrix<ScalarType>::BuildILUPreconditioner(bool transposed) {
   
-  //  ScalarType Det, **Matrix, **CoFactor;
```
economon: Thanks for the clean up here as requested too

Common/include/matrix_structure.hpp
pcarruscag: Borderline ocd... but oh well.

Common/src/matrix_structure.cpp
pcarruscag: Sure can, I'll see if there were comments in the smoothers that are not in the preconditioner.

Common/src/linear_solvers_structure.cpp
pcarruscag: Not in SU2, but the theory should be that if the preconditioner does not approximate the inverse of the matrix well you need smaller relaxation, you would want the relaxation that minimizes the spectral radius but that is not easy to compute. I also do not know what is the effect on the "quality" of the smoothing...

I can expose it to the config but I'm not too keen on running parameters tests on this.
By consistent with the other solvers do you mean computing it every iteration with some formula (e.g. such that the norm of the residual is minimized)? With this I have played with and it very easily stalls (omega=0).

Thank you for starting the review this quickly by the way.

economon: Aok, no need to test rigorously. I would be interested to have this exposed to the config though.

As for consistency with other relaxation factors, it is probably beyond this PR, and it can wait for later. Basically, what I meant is that it would be nice to see a more thought-out strategy for relaxation factors that affect the linear solvers. The current approach seems disjointed, as we have RELAXATION_FACTOR_FLOW that multiplies the output increment of the linear solve in the FV flow solver, for instance, and then other potential relaxation factors inside the linear solver, such as the one in question, and maybe other physics solvers that don't apply relaxation at all in the linear solver. Just food for thought for now.

Common/src/linear_solvers_structure.cpp
```diff
@@ -738,7 +828,17 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, CS
     ScreenOutput = config->GetDeform_Output();
   }
 
-  CMatrixVectorProduct<ScalarType> *mat_vec = NULL;
+  /*--- For smoother-type solvers set the associated preconditioner type. ---*/
+
+  switch (KindSolver) {
+    case SMOOTHER_LINELET: KindPrecond = LINELET; break;
```
pcarruscag: Should we just have KindSolver SMOOTHER in the config now?

Common/include/matrix_structure.inl
```diff
@@ -160,25 +175,364 @@ template<class ScalarType>
...
+    for (j = 0; j < n; j++) {
+      c[i*n+j] = 0.0;
+      for (k = 0; k < n; k++)
+        c[i*n+j] += a[i*n+k] * b[k*n+j];
+    }
+  }
+}
+
+#define __MATVECPROD_SIGNATURE__(TYPE,NAME) \
+inline void CSysMatrix<TYPE>::NAME(const TYPE *matrix, const TYPE *vector, TYPE *product)
```
pcarruscag: I hope this does not look too bad, but this MKL stuff does work very well for the discrete adjoint (maybe 10% faster than native code).

talbring:
Thanks @pcarruscag! This is a nice cleanup! :+1: 

So while you are already at it, you can already separate it in folders/files and remove the *inl (if all agree with these changes). Maybe adding a linear algebra subfolder makes sense.

pcarruscag:
[Here](https://github.com/su2code/SU2/files/3302370/benchmark.zip) are some comparisons for a case where the linear solver does little work (so not much benefit from this PR) and another where the linear solver has to work harder (18 iterations per iteration on average) and this makes a substantial difference.
I have started updated the residuals of things that seem consequence of round-off error, only one case requires closer inspection where things change several orders of magnitude (for the best).

pcarruscag:
Thanks @talbring,
I moved the preconditioner and matrix-vector product wrapper classes to separate files, these are so light weight that I was thinking of leaving them in one file.
I also moved some inlines to the hpp but I kept the private inlines in the inl file, these are only needed in the cpp and so by including the inl from the cpp (instead of bottom of hpp) we might avoid triggering recompilation of more units when working on implementation details of CSysMatrix.
Finally I would like to move/rename the larger files on a separate PR, that way it will be easier to track changes.
What do you think?

pcarruscag:
The case with large change of residuals (3 orders) is the rotating cylinders case (sliding interface).
With this PR I get the following flow field at the last iteration:
image
With develop I get this one at time iter 2:
image
And I think it is fair to say the case was actually blowing up before:
image
If I change the linear preconditioner to the ILU the develop results are basically the same as with this PR. Why does the previous LU_SGS diverge and the current doesn't? No idea, maybe for a different limit condition it would go the other way...

talbring:
@pcarruscag Good that you mention it. I saw exactly the same behavior while working on #715. If I disable grid movement for the first zone (where grid velocity is 0 anyway because there is no rotation) it works. In my opinion having grid movement disabled or enabled and specifying zero movement should result in exactly the same convergence behavior ...

Common/src/linear_solvers_structure.cpp
economon: I like this idea, yes!

pcarruscag: Done

pcarruscag:
I tested the MKL integration with the discrete adjoint, and everything looks ok, turns out to be only about 5% faster on a per iteration basis (i.e. excluding recording times). I had to grab some work from my other ongoing PR's to run a case with reasonable CFL settings so I am not going to upload files for this test.

@economon , @talbring , if you do not mind me moving the files with significant changes on a separate PR, I think this is ready to go.

economon:
@talbring : as you know, the difference between disabled and active grid movement with 0 velocity is that the former case does not even allocate the memory for the grid velocity at each node, and many conditionals checking for grid movement throughout the solver (fluxes, BCs) are avoided. This was to save memory and effort when grid motion is not needed, however, maybe we need to now change the strategy for multizone problems which may have both fixed and moving zones (perhaps always active with 0 as default). 

I am a little surprised they are not the same as well, but somewhere in the code path there must be an issue with this.. my guess is something related to BC_Fluid_Interface() or the transfer structure when grid movement is active on both sides but has a value of 0 on one of the interfaces.

talbring:
@pcarruscag For me its fine if you do it in a separate PR.

@economon I also assume that is has something to do with how the interfaces are handled. We definitely have to check it. For now in PR #715 I enabled grid movement (in the config) also in fixed zones.

pcarruscag:
If that case is so sensitive to be affected by the changes in this PR (it started converging without any change of options), it does not surprise me that other sources of numerical noise can have a similar influence.

economon:
Ok for me to wait for another PR to break the other files.

Would be good to get some additional reviews on this one from the community if folks have some time.

Common/src/matrix_structure.cpp
```diff
@@ -1656,9 +886,9 @@ void CSysMatrix<ScalarType>::BuildILUPreconditioner(bool transposed) {
            upper triangular part, then multiply and modify the matrix.
            Here, Aik' = Aik - Aij*inv(Ajj)*Ajk. ---*/
           
-          if ((kPoint >= jPoint) && (jPoint < (long)nPointDomain)) {
+          if (kPoint > jPoint) {
```
pcarruscag: This was not a bug, but we overwrite the ij block completely so it is not necessary to update it here.

pcarruscag:
@economon no one seems to have time, let's get going please.

.travis.yml
```diff
@@ -81,7 +81,7 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone -b feature_refactor_lin_solvers https://github.com/su2code/TestCases.git ./TestData
```
EduardoMolina: Please revert before we merge.

Common/include/linear_solvers_structure.hpp
```diff
@@ -242,19 +248,35 @@ class CSysSolve {
...
+   * \param[in] tol - tolerance with which to solve the system
+   * \param[in] m - maximum number of iterations
+   * \param[in] residual - norm of final residual
+   * \param[in] monitoring - turn on priting residuals from solver to screen.
+   * \param[in] config - Definition of the particular problem.
+   */
+  unsigned long Smoother_LinSolver(const VectorType & b, VectorType & x, ProductType & mat_vec,
+                                   PrecondType & precond, ScalarType tol, unsigned long m,
+                                   ScalarType *residual, bool monitoring, CConfig *config);
+
```
EduardoMolina: Thanks for the comments.

pcarruscag:
Thank you @WallyMaier and @EduardoMolina for the reviews.

PR #702, opened 2019-06-06
Eliminate Mac compiler warnings.

davetaflin:
## Proposed Changes
Removing a couple unused variables to quash compiler warnings on Mac.
 

## PR Checklist


vdweide:
@davetaflin, could you also try to use the following compiler options for the g++ compiler?

-g -Wall -Wextra -Wno-unused-parameter -Wno-empty-body

Then I get a lot of warnings coming from the TecIO functionality. Quite a few of these warnings come from boost, so I don't think you can do a lot about those, but there are also a few in the TecIO files themselves.

I run this on Ubuntu 18.04, g++ version 7.4.0, but I don't think the actual operating system and g++ version matter a lot.

davetaflin:
@vdweide, our (Tecplot’s) development compiler has been GCC 5, which gives no warnings with those flags. With GCC 9, ignoring warnings from Boost and OpenMPI (on Fedora 30), I see three types of warnings: (1) clause does not guard… (misleading indentation),  (2) Implicitly-declared operator= is deprecated, and (3) writing to an object of type…use copy-assignment or copy-initialization instead.

(1) Is a side-effect of code obfuscation (the reason management lets us release this code), and is probably unavoidable.
(2) and (3) are probably addressable. I’ll put these on my to-do list.

Thanks,

Dave

Issue #465, opened 2017-10-29
Test CUSP scheme in QuickStart's mesh


LeyiChai:
I'm a junior student and a new CFDer.
I'v been testing upwind scheme in QuickStart and found that CUSP scheme diverged. Below is some important configuration information
Firstly I just changed convective scheme of "inv_NACA0012.cfg" in QuickStart to CUSP and SU2 diverged directly. Configuration information lies below.(CFL = 4.0)
>% Mach number (non-dimensional, based on the free-stream values)
MACH_NUMBER= 0.8
%
% Angle of attack (degrees)
AOA= 1.25
%
% -------------------------- MULTIGRID PARAMETERS -----------------------------%
%
% Multi-Grid Levels (0 = no multi-grid)
MGLEVEL= 3
%
% Multi-grid cycle (V_CYCLE, W_CYCLE, FULLMG_CYCLE)
MGCYCLE= W_CYCLE
%
% Multi-Grid PreSmoothing Level
MG_PRE_SMOOTH= ( 1, 2, 3, 3 )
%
% Multi-Grid PostSmoothing Level
MG_POST_SMOOTH= ( 0, 0, 0, 0 )
%
% Jacobi implicit smoothing of the correction
MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0 )
%
% Damping factor for the residual restriction
MG_DAMP_RESTRICTION= 1.0
%
% Damping factor for the correction prolongation
MG_DAMP_PROLONGATION= 1.0
% -------------------- FLOW NUMERICAL METHOD DEFINITION -----------------------%
%
% Convective numerical method (JST, LAX-FRIEDRICH, CUSP, ROE, AUSM, HLLC,
%                              TURKEL_PREC, MSW)
CONV_NUM_METHOD_FLOW= CUSP
%
% Spatial numerical order integration (1ST_ORDER, 2ND_ORDER, 2ND_ORDER_LIMITER)
%
SPATIAL_ORDER_FLOW= 2ND_ORDER_LIMITER
%
% Slope limiter (VENKATAKRISHNAN, MINMOD)
SLOPE_LIMITER_FLOW= VENKATAKRISHNAN
%
% 1st, 2nd and 4th order artificial dissipation coefficients
AD_COEFF_FLOW= ( 0.15, 0.5, 0.02 )
%
% Time discretization (RUNGE-KUTTA_EXPLICIT, EULER_IMPLICIT, EULER_EXPLICIT)
[...]

PR #705, opened 2019-06-10
fixing CUSP numerics

WallyMaier:
## Proposed Changes
This addressed problems with flux computation in the CUSP scheme, mentioned in issue #465. 
The implementation was checked in reference to "Comparison of Several Dissipation Algorithms for Central Difference Schemes" Swanson.


WallyMaier:
@pcarruscag I updated the jacobian.  Regarding the flux, I am a bit confused as well.  The paper referenced above says flux implemented is a higher-order scheme that employs the SLIP scheme's limiter.  That being said, I do not have access to Blazek's book at the moment.

pcarruscag:
I followed up on two of the references the book cites:
http://aero-comlab.stanford.edu/Papers/AIAA-12422-882.pdf
https://ntrs.nasa.gov/archive/nasa/casi.ntrs.nasa.gov/19950005461.pdf
Both express the face flux as the average of left and right fluxes, which I guess makes sense since those are used for the dissipation part anyway.

PR #709, opened 2019-06-13
Removing python 2.7 from test matrix

talbring:
## Proposed Changes
This PR removes python2.7 from the test matrix in travis. We have a lot of development right now and until we have found a better way to speed up regression tests, this is, I think, the best approach to remove that roadblock at the moment:

- There is no real development done in the python framework at the moment
- Usually the incompatibilities between python 2 and 3 can be resolved quite easily.
- Python 2 is marked as deprecated anyway.

If you have other thoughts let me know.

There might be other ways like the caching feature of travis, but it takes effort and time to modify our system.

## PR Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ branches:
         - develop
 
 python:
-    - 2.7
     - 3.6
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ notifications:
   
 branches:
     only:
-        - master
+        - develop
 
 python:
     - 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ notifications:
   
 branches:
     only:
-        - develop
+        - master
 
 python:
     - 3.6


pcarruscag:
Hi @talbring 
This makes sense to me.
Also, would it make sense to merge the serial/parallel AD and direct regressions? So compile always with AD support which I guess also generates the non-AD, and then run the two regression scripts (e.g. serial_regression.py and serial_regression_AD.py)
Pedro

talbring:
I also thought about that merging but unfortunately we are already at the time-limit of 90mins with only the normal compilation/regression tests. I will explore more the caching feature. I already did some tests in the meson_build branch on that.

talbring:
Another thing related to that: Maybe it makes sense to disable travis for develop. Since we can only push to this branch through PRs, there is no point in running it again after a merge ... at least I don't see a reason.

talbring:
Following my proposal in the last comment:

I disabled travis to run on develop, instead I set it master. BUT: in addition I set up a cron job in travis, so that at least once every week develop is tested. What do you say ?

talbring:
Enabled it for develop again. Otherwise no test runs. Changed settings in travis to not build on pushes, only on PRs. Should save some computational work.

talbring:
Any further comments on this ? Otherwise I assume that there is no objection and I will merge it in today. Btw. as you might have seen, I enabled codacy for the project in order to check the code for security, errors and code style.

PR #715, opened 2019-06-19
Multizone Config Update 

talbring:
## Proposed Changes
This is the first part of a series of PRs which completely refactor the output (and parts of the input).

In this one I introduce the sub-config file feature for multizone/multiphysics problems. More information on how it works can be found here:

https://su2code.github.io/docs/Multizone

Note that the automatic deduction of the correct interface type is *not* working at the moment. So the `MARKER_FLUID_INTERFACE` or `MARKER_CHT_INTERFACE` still has to be there at the moment ...

Furthermore this changes the way grid movement is specified in the config files. Examples can be found also using the link above or by simply looking at the `sliding_interface` test cases. 

## PR Checklist


Common/include/option_structure.hpp
```diff
@@ -626,38 +626,48 @@ static const map<string, ENUM_CONDUCTIVITYMODEL_TURB> TurbConductivityModel_Map
...
-  EXTERNAL_ROTATION = 5,  /*!< \brief Arbitrary grid motion specified by external files at each time step with rigid rotation. */
-  AEROELASTIC = 6,    /*!< \brief Simulation with aeroelastic motion. */
-  MOVING_WALL = 7,    /*!< \brief Simulation with moving walls (translation/rotation). */
   ROTATING_FRAME = 8,    /*!< \brief Simulation in a rotating frame. */
   ELASTICITY = 9,    /*!< \brief Linear Elasticity. */
-  AEROELASTIC_RIGID_MOTION = 10, /*!< \brief Simulation with rotation and aeroelastic motion. */
   STEADY_TRANSLATION = 11,    /*!< \brief Simulation in a steadily translating frame. */
   GUST = 12, /*!< \brief Simulation on a static mesh with a gust. */
   MOVING_HTP = 13,    /*!< \brief Simulation with moving HTP (rotation). */
-  FLUID_STRUCTURE_STATIC = 14 /*!< \brief Fluid structure deformation with no grid velocity. */
```
WallyMaier: Should the identifiers be updated?

SU2_CFD/src/numerics_direct_mean.cpp
```diff
@@ -5768,9 +5768,9 @@ void CSourceRotatingFrame_Flow::ComputeResidual(su2double *val_residual, su2doub
   
   /*--- Retrieve the angular velocity vector from config. ---*/
   
-  Omega[0] = config->GetRotation_Rate_X(config->GetiZone())/config->GetOmega_Ref();
-  Omega[1] = config->GetRotation_Rate_Y(config->GetiZone())/config->GetOmega_Ref();
-  Omega[2] = config->GetRotation_Rate_Z(config->GetiZone())/config->GetOmega_Ref();
+  for (iDim = 0; iDim < 3; iDim++){
+    Omega[iDim] = config->GetRotation_Rate()[iDim]/config->GetOmega_Ref();
```
WallyMaier: These changes are great, thanks!

Common/include/config_structure.inl
```diff
@@ -937,71 +937,51 @@ inline void CConfig::SetMu_PolyCoeffND(su2double val_coeff, unsigned short val_i
 
 inline void CConfig::SetKt_PolyCoeffND(su2double val_coeff, unsigned short val_index) { KtPolyCoefficientsND[val_index] = val_coeff; }
 
-inline unsigned short CConfig::GetKind_GridMovement(unsigned short val_iZone) { return Kind_GridMovement[val_iZone]; }
+inline unsigned short CConfig::GetKind_GridMovement() { return Kind_GridMovement; }
```
economon: Nice.. the multi config structure makes things like the grid motion specification much simpler 👍

Common/include/option_structure.hpp
```diff
@@ -2059,11 +2069,13 @@ static const map<string, ENUM_VERIFICATION_SOLUTIONS> Verification_Solution_Map
 
 class COptionBase {
 private:
+  vector<string> value;
 public:
   COptionBase() {};
   virtual  ~COptionBase() = 0;
   //  virtual string SetValue(string) {SU2MPI::PrintAndFinalize("shouldn't be here"); return "";};
```
economon: can remove this outdated commented line

Common/src/config_structure.cpp
```diff
@@ -2562,16 +2663,138 @@ bool CConfig::SetRunTime_Parsing(char case_filename[MAX_STRING_SIZE]) {
...
+      
+    }
+    
+    /*--- Check if subconfig files exist --- */
+    
+    if (nConfig_Files != 0){
+      for (unsigned short iConfig = 0; iConfig < nConfig_Files; iConfig++){
+        ifstream f(Config_Filenames[iConfig].c_str());
+        if (!f.good()){
+          SU2_MPI::Error("Config file " + Config_Filenames[iConfig] + " defined in CONFIG_FILES does not exist", CURRENT_FUNCTION);
```
economon: thanks for the error checks 👍

Common/src/geometry_structure.cpp
```diff
@@ -13760,205 +13759,6 @@ void CPhysicalGeometry::SetMaxLength(CConfig* config) {
 
 }
 
-void CPhysicalGeometry::MatchInterface(CConfig *config) {
```
economon: thanks for cleaning out marker_interface everywhere (including the old MPI)

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -107,8 +110,36 @@ void CIteration::SetGrid_Movement(CGeometry ****geometry_container,
...
+
+      break;
+
+ 	/*--- Already initialized in the static mesh movement routine at driver level. ---*/ 
+    case STEADY_TRANSLATION: case ROTATING_FRAME:
+      break;
+
+  }
+  
+  if (config_container[val_iZone]->GetSurface_Movement(DEFORMING)){
```
economon: One thing I have been meaning to change is to remove the calls to SetVolume_Deformation() in between each of the updates of the surface positions.. in the case that multiple sequential surface movements are present, they should be additive (they use an increment to the position) and we should be able to call SetVolume_Deformation() just once at the end of the sequence. Can you please give it a look and adjust if you agree?

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -2765,6 +2449,7 @@ void CEulerSolver::SetNondimensionalization(CConfig *config, unsigned short iMes
   Time_Ref          = Length_Ref/Velocity_Ref;                                     config->SetTime_Ref(Time_Ref);
   Omega_Ref         = Velocity_Ref/Length_Ref;                                     config->SetOmega_Ref(Omega_Ref);
   Force_Ref         = config->GetDensity_Ref()*Velocity_Ref*Velocity_Ref*Length_Ref*Length_Ref; config->SetForce_Ref(Force_Ref);
+  Heat_Flux_Ref     = Density_Ref*Velocity_Ref*Velocity_Ref*Velocity_Ref;           config->SetHeat_Flux_Ref(Heat_Flux_Ref);
```
economon: thx, this has been missing 👍 . we should also use it in BC_Heat_Flux() to non-dim the input heat flux at the wall I think to be consistent (see incompressible version of that BC)

Common/include/config_structure.inl
```diff
@@ -937,71 +937,51 @@ inline void CConfig::SetMu_PolyCoeffND(su2double val_coeff, unsigned short val_i
...
-inline void CConfig::SetMotion_Origin_X(unsigned short val_iZone, su2double val_origin) { Motion_Origin_X[val_iZone] = val_origin; }
-
-inline void CConfig::SetMotion_Origin_Y(unsigned short val_iZone, su2double val_origin) { Motion_Origin_Y[val_iZone] = val_origin; }
-
-inline void CConfig::SetMotion_Origin_Z(unsigned short val_iZone, su2double val_origin) { Motion_Origin_Z[val_iZone] = val_origin; }
-
-inline su2double CConfig::GetTranslation_Rate_X(unsigned short val_iZone) { return  Translation_Rate_X[val_iZone]; }
-
-inline su2double CConfig::GetTranslation_Rate_Y(unsigned short val_iZone) { return  Translation_Rate_Y[val_iZone]; }
+inline su2double* CConfig::GetMarkerMotion_Origin(unsigned short iMarkerMoving){return &MarkerMotion_Origin[3*iMarkerMoving];}
```
pcarruscag: Can you return const pointer instead?

Common/src/fem_geometry_structure.cpp
```diff
@@ -6929,4 +6863,66 @@ void CMeshFEM_DG::InitStaticMeshMovement(CConfig              *config,
...
+    for(unsigned short i=0; i<boundaries.size(); ++i) {
+      if( !boundaries[i].periodicBoundary ) {
+        
+        /* Check if for this boundary a motion has been specified. */
+        if (config->GetMarker_All_Moving(i) == YES) {
+          
+          /* Determine the prescribed translation velocity, rotation rate
+               and rotation center. */
+          const su2double Center[] = {config->GetMotion_Origin()[0],
+                                      config->GetMotion_Origin()[1],
```
pcarruscag: If this is the general way in which these methods will be called, why not pass iDim as an argument and void returning pointers?

talbring: Thanks for pointing that out. Thats something I wanted to change, thanks for reminding me :)

SU2_CFD/src/iteration_structure.cpp
talbring: Maybe we can deal with that in the new mesh solver implementation @rsanfer ?

economon: Agreed. Let's wait on this change. I will make an issue instead.

stephansmit:
Was playing around with the turbomachinery cases with the new multi-zone config. I like the new changes, makes it much clearer!

Just a remark, don't know if this is on purpose, but for the configuration parameter TURBOMACHINERY_KIND, you still need to supply them in the general configuration file for both zones (so in case of a two zone problem: TURBOMACHINERY_KIND = CENTRIFUGAL CENTRIFUGAL). Wouldnt it make more sense to have this per zone specified in the respective configuration file?

PR #718, opened 2019-06-21
Limiter related fixes

pcarruscag:
## Proposed Changes
I stumbled upon three things, one is a fix, the other two I have questions.
@talbring and someone familiar with the SA and SST solvers please have a look at my comments below.

## PR Checklist


SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -5819,9 +5818,20 @@ void CEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config) {
...
   unsigned short iVar, iDim;
   su2double **Gradient_i, **Gradient_j, *Coord_i, *Coord_j,
-  *Primitive, *Primitive_i, *Primitive_j, *LocalMinPrimitive, *LocalMaxPrimitive,
-  *GlobalMinPrimitive, *GlobalMaxPrimitive,
+  *Primitive, *Primitive_i, *Primitive_j,
+  *LocalMinPrimitive = NULL, *LocalMaxPrimitive = NULL,
+  *GlobalMinPrimitive = NULL, *GlobalMaxPrimitive = NULL,
   dave, LimK, eps2, eps1, dm, dp, du, y, limiter;
+
+#ifdef CODI_REVERSE_TYPE
```
pcarruscag: @talbring I think freezing the limiters for the discrete adjoint should mean not record the computation but still compute their values, otherwise the residuals shoot up. Do you agree?

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -6011,16 +6022,22 @@ void CEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config) {
...
+        AD::SetPreaccIn(Coord_i, nDim);
+        AD::SetPreaccIn(Coord_j, nDim);
+        AD::SetPreaccIn(node[iPoint]->GetSolution_Max(iVar));
+        AD::SetPreaccIn(node[iPoint]->GetSolution_Min(iVar));
+        AD::SetPreaccIn(node[jPoint]->GetSolution_Max(iVar));
+        AD::SetPreaccIn(node[jPoint]->GetSolution_Min(iVar));
         
         if (config->GetKind_SlopeLimit_Flow() == VENKATAKRISHNAN_WANG) {
+          AD::SetPreaccIn(GlobalMaxPrimitive[iVar]);
+          AD::SetPreaccIn(GlobalMinPrimitive[iVar]);
```
pcarruscag: This is the fix, these dependencies where missing for the Venkatakrishnan-Wang limiter.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -5965,33 +5975,34 @@ void CEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config) {
...
-    
-    Primitive = node[0]->GetPrimitive();
-    for (iVar = 0; iVar < nPrimVarGrad; iVar++) {
-      LocalMinPrimitive[iVar] = Primitive[iVar];
-      LocalMaxPrimitive[iVar] = Primitive[iVar];
-    }
-    
-    for (iPoint = 0; iPoint < geometry->GetnPoint(); iPoint++) {
+    if (config->GetKind_SlopeLimit_Flow() == VENKATAKRISHNAN_WANG) {
+
```
pcarruscag: Minor fix, global min and max are only needed for this version of the limiter.

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -1176,8 +1173,6 @@ void CTurbSASolver::Preprocessing(CGeometry *geometry, CSolver **solver_containe
 
   if (limiter_turb) SetSolution_Limiter(geometry, config);
 
-  if (limiter_flow) solver_container[FLOW_SOL]->SetPrimitive_Limiter(geometry, config);
-
```
pcarruscag: Person familiar with the SA and SST solvers, do we need this call to `SetPrimitive_Limiter`? There is already one in the preprocessing method of the Euler/NS solvers.

SU2_CFD/src/solver_direct_mean.cpp
talbring: Thanks for adding these.

SU2_CFD/src/solver_direct_mean.cpp
talbring: Yep, that is actually the better way of doing it. Thanks for fixingt that.

SU2_CFD/src/solver_direct_turbulent.cpp
talbring: I can remember that we had that discussion already a long time ago. Cannot remember what the conclusion was though. In my opinion it should not matter, whether we use the updated values or not ... and apparently it does not since the regression tests pass.

pcarruscag: I am not sure they would be updated, since the gradients are not. Let's give it a couple days before merging just in case. Thanks for the review.

economon: Yep, we have had this discussion before. I am also fine with removing them (in the typical case with upwind for mean flow and turb, everything will converge to the same solution anyway), and it will give a little performance boost.

I recall that the sticking point was whether any inconsistencies between the numerical methods in the mean flow and turb models would result in some problems, e.g., mixing JST for the mean flow with a second-order upwind method for the turb scalars (which uses the primitive limiter on the mean flow state too in CTurb:UpwindResid()). We should clarify this finally, even if it just means adding some error check.

pcarruscag: Oh... I see what you mean, that means the turbulence models are being solved for a velocity field that is not divergence-free. I think this warrants further discussion, is that old discussion in an issue? (all I could find was #422). If not I will open one.

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -3778,10 +3777,21 @@ void CIncEulerSolver::SetPrimitive_Limiter(CGeometry *geometry, CConfig *config)
...
   unsigned short iVar, iDim;
   su2double **Gradient_i, **Gradient_j, *Coord_i, *Coord_j,
-  *Primitive, *Primitive_i, *Primitive_j, *LocalMinPrimitive, *LocalMaxPrimitive,
-  *GlobalMinPrimitive, *GlobalMaxPrimitive,
+  *Primitive, *Primitive_i, *Primitive_j,
+  *LocalMinPrimitive = NULL, *LocalMaxPrimitive = NULL,
+  *GlobalMinPrimitive = NULL, *GlobalMaxPrimitive = NULL,
   dave, LimK, eps2, eps1, dm, dp, du, y, limiter;
   
+#ifdef CODI_REVERSE_TYPE
```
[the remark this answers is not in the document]
pcarruscag: No problem, thank you for the review. Would it make sense to move this method to CSolver? That is where the SolutionLimiter one is implemented.

SU2_CFD/src/solver_direct_turbulent.cpp
economon: I think I found the old discussion on this in #403

SU2_CFD/src/solver_direct_mean_inc.cpp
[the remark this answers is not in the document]
economon: Not pressing, but we could consider this as part of a strategy to reuse more code btwn the comp. and density-based incomp. solver classes. Could be interesting to have an intermediate base class where the repeated routines live and then specialized routines down in the Euler/NS classes specific to each. But, this can wait for some other large refactoring to happen elsewhere first

SU2_CFD/src/solver_direct_turbulent.cpp
pcarruscag: Thank you, I'll have a look.

PR #723, opened 2019-06-27
Revert "Fix for Q-criterion"

economon:
Actually, before merging the Q-criterion changes, it would be good to share with the community the following: 

#704 was a fix for the original quick and dirty implementation of Q in the flow output classes (I think there was an extra sqrt()), but in fixing that issue, we came across differing implementations in various codes. The version that was just accepted assumes a factor of 2.0 in the definition of the strain and vorticity tensors, which seems to be more common. Some implementations carry the full second variant of the velocity gradient tensor, which has some additional terms related to the divergence of velocity, that will vanish for incompressible flows (resulting in the Haller definition that was merged).

If folks have any additional comments or concerns we can adjust things on this PR. Otherwise, we can close this PR without reverting anything.


economon:
Alright, since there are no comments, let's go ahead and close this one. The comments above will be in the GitHub record for later.

PR #724, opened 2019-06-28
New Output Structure

talbring:
## Proposed Changes

This PR introduces a completely rewritten output structure. If you have ever looked into the old output_structure.cpp file, you probably noticed that it was quite messy. To add new output to screen or history people just copied what was there and modified it to suit there own needs. This resulted in a lot of if/switch statements and was hard to read ... one reason for this was that we have not really used an object oriented structure for the output.

Despite rearranging everything into separate classes, here is a non-exhaustive list of new features and changes:

- Fully customizable screen, history and volume output
- New multizone screen output
- New flexible way so set convergence criteria (any residual and/or coefficient can be used as a monitor)
- Consistent naming of iteration variables (TIME_ITER, OUTER_ITER, INNER_ITER)
- Consistent file naming options (RESTART_FILENAME, SOLUTION_FILENAME)
- On-the-fly writing of all visualization files
- New Singlezone and Multizone drivers are now used by default.

Note that this is a very big change. And i apologize for any inconvenience and troubles that this might cause. I can only maintain features covered by the regression tests. So please test it with your cases and let me know if anything is missing (especially in the output). 

I will create documentation for all the new features soon. If you want to test it already, please let me know, I need some feedback. 

User Documentation on the changes:
https://su2code.github.io/docs/Solver-Setup
https://su2code.github.io/docs/Custom-Output
https://su2code.github.io/docs/Guide-to-v7/

## Related Work
#715 is already included in this PR and should be merged before.

## PR Checklist


SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1345 @@
...
+//   return false;
+// }
+  return true;
+}
+
+void COutput::SetCommonHistoryFields(CConfig *config){
+  
+  /// BEGIN_GROUP: ITERATION, DESCRIPTION: Iteration identifier.
+  /// DESCRIPTION: The time iteration index.
+  AddHistoryOutput("TIME_ITER",     "Time_Iter",  FORMAT_INTEGER, "ITER");
```
economon: I had a suggestion to improve the config option tracking/documentation (and keep it automatically up to date), and I think we could try it here too for the new output options that are registered. 

Is it easy for you to add the string description as an input directly into the prototype for AddHistoryOutput() as well as add a print_description() method that would write it to the console (or write all registered options in one loop)? That way, if a user wanted to get the list of all available outputs for a particular solver, we could have a "help" option in the config file / at the command line that could first print all of the group names with their descriptions, followed by all available fields within a chosen group with their string descriptions. Should be easier to maintain and expose, since the descriptions will be required directly in the C++ whenever adding the option. 

We could consider adding a similar self-describing approach for all config options if this works.

SU2_CFD/src/output/filewriter/CFVMDataSorter.cpp
```diff
@@ -0,0 +1,1076 @@
+#include "../../../include/output/filewriter/CFVMDataSorter.hpp"
+
+
+CFVMDataSorter::CFVMDataSorter(CConfig *config, CGeometry *geometry, unsigned short nFields, std::vector<std::vector<su2double> >& Local_Data) : CParallelDataSorter(config, nFields){
```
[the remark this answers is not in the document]
talbring: Thats something we have to discuss actually. I don't know what is better ... anyway, I will make sure that everything is deleted properly.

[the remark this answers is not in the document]
economon: I think it is not too critical at first actually.. if you keep the deallocation within the separate smaller methods, then we can always move them out of the destructor easily later.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1424 @@
...
+    }
+  } else {
+    nPoint = geometry->GetnPoint();
+  }
+  
+  /*--- Now that we know the number of fields, create the local data array to temporarily store the volume output 
+   * before writing it to file ---*/
+   
+  Local_Data.resize(nPoint, std::vector<su2double>(GlobalField_Counter, 0.0));
+
```
pcarruscag: This is not a very good way to store the data, have you considered flattening this into a single dimension or at least transposing it?

SU2_CFD/src/output/CAdjElasticityOutput.cpp
```diff
@@ -0,0 +1,177 @@
...
+  
+}
+
+void CAdjElasticityOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint){
+
+  CVariable* Node_Struc = solver[FEA_SOL]->node[iPoint];
+  CPoint*    Node_Geo  = geometry->node[iPoint];
+
+  SetVolumeOutputValue("COORD-X", iPoint,  Node_Geo->GetCoord(0));
+  SetVolumeOutputValue("COORD-Y", iPoint,  Node_Geo->GetCoord(1));
```
pcarruscag: You are hashing into the map a bunch of times per node but the access sequence is always the same, you could cache the offset (column?) during preprocessing and not go through the map here.

SU2_CFD/src/output/filewriter/CFVMDataSorter.cpp
```diff
@@ -0,0 +1,693 @@
...
+      int source = ii;
+      int tag    = ii + 1;
+      SU2_MPI::Irecv(&(connRecv[ll]), count, MPI_DOUBLE, source, tag,
+                     MPI_COMM_WORLD, &(recv_req[iMessage]));
+      iMessage++;
+    }
+  }
+  
+  /*--- Launch the non-blocking sends of the connectivity. ---*/
+
```
pcarruscag: Is this comment correct? Is it not data you are sending?

SU2_CFD/include/output/filewriter/CFEMDataSorter.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include "CParallelDataSorter.hpp"
+
+class CFEMDataSorter : public CParallelDataSorter{
+private:
+  
+  std::vector<std::vector<su2double> >* Local_Data;
```
pcarruscag: Please make this const* or const&

SU2_CFD/src/output/COutput.cpp
pcarruscag: After looking at the parallelization part I understand why it cannot be transposed.
I am going to write a re-usable 2D container for CVariable, we could eventually use that.

SU2_CFD/src/output/CAdjElasticityOutput.cpp
talbring: Thanks for mentioning this! That was indeed not very efficient. I added caching of the offset. That speeds up the whole process a lot. I implemented it in a way so that the order the SetVolumeOutputValue calls can be still different from the order of the AddVolumeOutput calls.

pcarruscag: Looks good @talbring, should do the trick provided the function is not called twice for the same variable.
Will the list of output variables be exposed to the config? (I remember from yours and @rsanfer 's presentation that the screen output will be) Is that why the variables are being referenced via strings? Have you considered a pair of enum/map like we have for other options? That way you could create a vector of enum instead of a vector of strings and use that as a "map" which would be efficient without needing to rely on access order, etc.

talbring: What do you mean exactly by "exposing it to the config" ? The rational behind having it as strings is that it is independent of the config structure. So just adding output doesn't require to recompile the whole code. 

I added a further check to make sure that the SetVolumeOutputValue() was not called twice for the same field. The good thing is that by adding these checks I already found some bugs :)

pcarruscag: I mean having a config option that allows the user to pick which fields they want in the output file.

talbring: Oh thats what you mean! You can specify it in the config file. That was also an important point of the implementation. 

There are the options
```
VOLUME_OUTPUT
SCREEN_OUTPUT
HISTORY_OUTPUT
```
to specify the fields you want to have. The config structure just has a string list for these options and no enum/map combination.

A more detailed description follows soon.

pcarruscag: Ok I'll hold further comments until then, but if the concern with the enum/map pair is compilation dependencies those can be defined at the level of the COutput children classes.

talbring:
I wrote a small documentation on how to customize the output: https://su2code.github.io/docs/Custom-Output/

More will follow soon.

SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -0,0 +1,660 @@
...
+  AddVolumeOutput("LIMITER_MOMENTUM-X", "Limiter_Momentum_x", "LIMITER");
+  AddVolumeOutput("LIMITER_MOMENTUM-Y", "Limiter_Momentum_y", "LIMITER");
+  if (nDim == 3)
+    AddVolumeOutput("LIMITER_MOMENTUM-Z", "Limiter_Momentum_z", "LIMITER");
+  AddVolumeOutput("LIMITER_ENERGY", "Limiter_Energy", "LIMITER");
+  
+  switch(config->GetKind_Turb_Model()){
+  case SST:
+    AddVolumeOutput("LIMITER_TKE", "Limiter_TKE", "RESIDUAL");
+    AddVolumeOutput("LIMITER_DISSIPATION", "Limiter_Omega", "RESIDUAL");
```
clarkpede: The group for the turbulence variable limiters should be `LIMITER`.

clarkpede:
First off, I love the changes. A cleanup of COuput and greater output flexibility were much needed.

So let me see if I understand this code.  Say that I, as a developer, want to add a new variable to the volume output.  For example, I want the production of turbulent kinetic energy from the SST model during a compressible simulation.  As a user, I can always add new outputs by adding new field names to the config file.  But production isn't one of the (currently) available config-file options.  So I would have to code it.

First, I would obviously make sure that the production is properly stored in the turbulence variable class, with an accessor like `GetTurbProduction()`

Second, in `CFlowCompOutput::LoadVolumeData`, I would add to the appropriate place:

```cpp
SetVolumeOutputValue("PRODUCTION", iPoint, Node_Turb->GetTurbProduction());
```

Third, I would add `PRODUCTION` as a cfg input by adding the following line to `CFlowCompOutput::SetVolumeOutputFields`:

```cpp
AddVolumeOutput("PRODUCTION", "Turb_Production", "PRIMITVE")
```

I could also split off production into its own output group by changing `PRIMITVE` to any other name.  This name does not have to be registered anywhere else. In other words, I can change `PRIMITVE` to `SST_NUMERICS` and the only other place I need to put `SST_NUMERICS` is my cfg file.

Finally, I could add either of the two following lines to my cfg file:
```
VOLUME_OUTPUT= (COORDINATES, SOLUTION, PRODUCTION) 
VOLUME_OUTPUT= (COORDINATES, SOLUTION, SST_NUMERICS)
``` 

Is this example correct? If not, what am I misunderstanding? What am I missing?

talbring:
@clarkpede You explained everything correctly, I have nothing to add! Sorry for not explaining it yet, I still have to write a small developers documentation for that. But it seems to be more or less self explanatory :D

talbring:
A first draft of the Solver Setup documentation can be found here: https://su2code.github.io/docs/Solver-Setup 

The idea is that the options described there should be used by all solvers (which is also the goal of this PR) and that we have a short description on the different solvers in separate documents.

pcarruscag:
@talbring, ok if I start working on a branch merging this one and #753?

talbring:
@pcarruscag yeah of course. I guess that makes sense.

talbring:
I have created already a small document giving infos on how to prepare for version 7.0.

https://su2code.github.io/docs/Guide-to-v7/

talbring:
I am more than happy to get any reviews! However, I don't expect that anyone will go through all files and in order to properly test it, we have to slightly push it. 

We will give it a couple of days though.

economon:
I agree with @talbring. This one is a burden on both sides (to prepare and review), and it is easy for us to keep putting it off. I would recommend that everyone takes a look at this in the next day or two - at least take a look at the changes that will affect your own work.

Once the tests pass, we can be confident that we are preserving what is in the regressions, but we will need a short period of conflict fixing and testing for those with active branches. The majority of the PR is for output features, which should not affect correctness of the solvers, but there are some important changes to the drivers and options that folks should be aware of. Please give these things a pass when you have a minute.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+  /*--- Check if screen output should be written --- */
+  
+  if (!PrintOutput(curTimeIter, ScreenWrt_Freq_Time)&& 
+      !(curTimeIter == config->GetnTime_Iter() - 1)){
+    
+    return false;
+    
+  }
+  
+  if (convergence) {return true;}
```
pcarruscag: this check is not needed

SU2_CFD/src/output/filewriter/CCSVFileWriter.cpp
```diff
@@ -0,0 +1,150 @@
+#include "../../../include/output/filewriter/CCSVFileWriter.hpp"
+#include "../../../include/output/filewriter/CParallelDataSorter.hpp"
+
+CCSVFileWriter::CCSVFileWriter(vector<string> fields, unsigned short nDim, 
+                               string fileName, CParallelDataSorter *dataSorter) : 
+  CFileWriter(fields, fileName, dataSorter, ".csv", nDim){}
```
pcarruscag: use std::move on fields and fileName and inside CFileWriter

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
```diff
@@ -0,0 +1,412 @@
...
+  if (nParallel_Hexa > 0 && Conn_Hexa_Par != NULL) delete [] Conn_Hexa_Par;
+  if (nParallel_Pris > 0 && Conn_Pris_Par != NULL) delete [] Conn_Pris_Par;
+  if (nParallel_Pyra > 0 && Conn_Pyra_Par != NULL) delete [] Conn_Pyra_Par;
+  
+  if (connSend != NULL) delete [] connSend;
+  
+  
+  /*--- Deallocate memory for solution data ---*/
+  
+//  if (Parallel_Data != NULL){
```
pcarruscag: is this needed?

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
```diff
@@ -0,0 +1,412 @@
...
+   sending. ---*/
+  
+  connSend = NULL;
+  connSend = new su2double[VARS_PER_POINT*nPoint_Send[size]]();
+  
+  
+  /*--- Allocate the data buffer to hold the sorted data.
+   * It should be big enough to also temporarily hold the data received using the AD type ---*/
+  
+  dataBuffer = reinterpret_cast<passivedouble*>(new su2double[VARS_PER_POINT*nPoint_Recv[size]]());
```
pcarruscag: this is messed up

SU2_CFD/src/output/filewriter/CParaviewBinaryFileWriter.cpp
```diff
@@ -0,0 +1,1121 @@
...
+  char str_buf[MAX_STRING_LENGTH], fname[100];
+  
+  const int NCOORDS = 3;
+
+  strcpy(fname, fileName.c_str());
+  
+  /*--- Check for big endian. We have to swap bytes otherwise. ---*/
+  
+  bool BigEndian;
+  union {int i; char c[4];} val;
```
pcarruscag: this is not standard c++

pcarruscag:
This is obviously very good work and should make it much simpler to add new outputs etc.

My main concern when first reviewing was performance (something no one seems to care about) but from the limited testing I did while merging this and CVariable there seems to be no big impact, I did not time anything though, if you have numbers to share please do.
Nevertheless if you can get away with using unordered_map instead of map it would be better.

Possibly part of the reason everyone puts of reviewing this is that it is a major refactoring and yet there are no accompanying notes about the architecture the implementation choices etc. I absolutely guarantee you that is not how professional software is developed.

economon:
I know that it can be frustrating sometimes operating within the constraints of an open-source project such as ours. However, I can assure you that folks do indeed care about performance, and sometimes I have the impression that we stress over relatively small performance issues (remember to keep the total pie chart of where the major work of the solver resides in mind). 

The option WRT_PERFORMANCE= YES is available to get timings for runs broken down by preprocessing, compute, and output phases, and I think we should focus our performance concerns first on issues within the compute phase, unless a major bottleneck appears in the other two phases that completely prohibits us from running larger cases (we have been clearing many of those out lately).

We do not have the resources of a professional software company, but what we do have is a great community of folks who are putting in lots of effort on a volunteer basis. @pcarruscag: your reviews have been very helpful for improving contributions - thank you for that effort. Let's keep supporting each other, but let's also make sure we stay positive and foster a welcoming environment to encourage more participation.

pcarruscag:
@economon I find not getting comments, or engaging in discussion and then being left hanging for weeks for a reply with my work suspended, more discouraging than getting objective comments.
So please, if those replies are not coming because they are difficult to word, just give it to me straight.

Also do note I used the word "notes" and not documentation, i.e. scribbles are good enough.
It is not very easy to look through 400+ changed files and piece back together the intended design.

economon:
I understand your frustration, and the feedback is helpful. Encouraging more participation in Issues and PRs is very important for us. We are still learning and improving our processes. A good metric for us to increase in the project is the total number of *different* people submitting/participating in PRs/Issues (not the total number of comments from just a handful). The best way to scale is to have the work evenly distributed among many folks, rather than just a handful processing the PRs. This will likely take some time & training, but I expect we can accomplish it while remaining positive. Open to good ideas there on how to better achieve it. 

As for the other topic - I think that our recent move toward draft PRs may help with describing design decisions. This will also take some time to be adopted, and may or may not reach the depth necessary in the descriptions, but it is a good first step so that we can see things progress in real time. Using the project boards to list tasks and post comments is also helpful (but takes more effort). Open to ideas there too.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,770 @@
...
+    int    offset; 
+    /*! \brief The group this field belongs to. */
+    string outputGroup;
+    /*! \brief String containing the description of the field */
+    string description;
+    /*! \brief Default constructor. */
+    VolumeOutputField () {}
+    /*! \brief Constructor to initialize all members. */    
+    VolumeOutputField(string fieldname, int offset, string volumeoutputgroup, string description):
+      fieldName(fieldname), offset(offset), outputGroup(volumeoutputgroup), description(description){}
```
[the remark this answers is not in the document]
talbring: Done

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,770 @@
...
+   */
+  void Postprocess_HistoryFields(CConfig *config);
+  
+  /*!
+   * \brief Check whether we should print output.
+   * \param[in] iIter - Current iteration.
+   * \param[in] iFreq - Frequency of output printing.
+   */
+  inline bool PrintOutput(unsigned long iIter, unsigned long iFreq) {
+    if (iFreq == 0.0){
```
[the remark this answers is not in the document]
talbring: What do you mean ?

SU2_CFD/include/output/filewriter/CFileWriter.hpp
```diff
@@ -0,0 +1,142 @@
...
+   * \brief The MPI size, aka the number of processors.
+   */
+  int size;
+  
+  /*!
+   * \brief The file extension to be attached to the filename.
+   */
+  std::string file_ext;
+  
+  su2double StartTime, /*!< \brief Start time for time measurement  */
```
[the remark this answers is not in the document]
talbring: Hm I am not a big fan of mixing it as long as it is not important for performance.

SU2_CFD/include/output/filewriter/CParallelDataSorter.hpp
```diff
@@ -0,0 +1,242 @@
...
+   * \input iPoint - the point ID.
+   * \return the value of the data field at a point.
+   */
+  passivedouble GetData(unsigned short iField, unsigned long iPoint) {return dataBuffer[iPoint*GlobalField_Counter + iField];}
+  
+  /*!
+   * \brief Get the pointer to the sorted linear partitioned data.
+   * \return Pointer to the sorted data.
+   */
+  passivedouble *GetData() {return dataBuffer;}
```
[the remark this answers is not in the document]
talbring: Done

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -5509,23 +5541,23 @@ CDiscAdjFSIDriver::CDiscAdjFSIDriver(char* confFile,
...
-         break;
-       case DISC_ADJ_FEM:
-         direct_iteration[iZone] = new CFEAIteration(config_container[iZone]);
-         nVar_Struct = solver_container[iZone][INST_0][MESH_0][ADJFEA_SOL]->GetnVar();
-         structure_criteria    = config_container[iZone]->GetMinLogResidual_BGS_S();
-         structure_criteria_rel = config_container[iZone]->GetOrderMagResidual_BGS_S();
-         break;
-    }
-  }
+//  for (iZone = 0; iZone < nZone; iZone++){
```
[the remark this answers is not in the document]
talbring: For primal FSI problems it uses the new Multizone driver implementation now. Check the test cases. The output has changed, and consequently also the regression test values. But I checked it very carefully and made that they give identical results.

SU2_CFD/src/output/CAdjElasticityOutput.cpp
```diff
@@ -0,0 +1,189 @@
...
+  /*--- Set the default convergence field --- */
+
+  if (convField.size() == 0 ) convField = "ADJOINT_DISP_X";
+  
+}
+
+CAdjElasticityOutput::~CAdjElasticityOutput(void) {
+
+  if (rank == MASTER_NODE){
+    histFile.close();
```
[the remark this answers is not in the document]
talbring: Yep, removed it.

SU2_CFD/src/output/CAdjFlowCompOutput.cpp
```diff
@@ -0,0 +1,469 @@
...
+CAdjFlowCompOutput::CAdjFlowCompOutput(CConfig *config, unsigned short nDim) : COutput(config, nDim, false) {
+   
+  turb_model = config->GetKind_Turb_Model();
+  
+  cont_adj = config->GetContinuous_Adjoint();
+  
+  /*--- Set the default history fields if nothing is set in the config file ---*/
+  
+  if (nRequestedHistoryFields == 0){
+    requestedHistoryFields.push_back("ITER");
```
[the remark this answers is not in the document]
talbring: Thanks for the hint, changed it!

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+          (*fileWritingTable) << "SU2 restart" << fileName + CSU2BinaryFileWriter::fileExt;   
+      }
+      
+      fileWriter = new CSU2BinaryFileWriter(volumeFieldNames, nDim, fileName, volumeDataSorter);
+      
+      break;
+      
+    case MESH:
+      
+      if (fileName == "")
```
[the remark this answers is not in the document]
talbring: Done.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+        if (Field.screenFormat == FORMAT_INTEGER) width = std::max((int)Field.fieldName.size()+2, 10);  
+        else{ width = std::max((int)Field.fieldName.size()+2, 18);}
+        historyFileTable->AddColumn("\"" + Field.fieldName + "\"", width);
+      }
+    }  
+  }
+  
+  for (iField_Output = 0; iField_Output < historyOutputPerSurface_List.size(); iField_Output++){
+    for (iMarker = 0; iMarker < historyOutputPerSurface_Map[historyOutputPerSurface_List[iField_Output]].size(); iMarker++){  
+      const HistoryOutputField &Field = historyOutputPerSurface_Map[historyOutputPerSurface_List[iField_Output]][iMarker];
```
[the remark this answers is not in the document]
talbring: You are right, I always wanted to add references. Thanks for the reminder.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+    }
+    
+  } else {
+    
+    for (iPoint = 0; iPoint < geometry->GetnPointDomain(); iPoint++) {
+      
+      /*--- Check for halos & write only if requested ---*/
+      /*--- Load the volume data into the Local_Data() array. --- */
+      
+      buildFieldIndexCache = !fieldIndexCache.size() ? true : false;
```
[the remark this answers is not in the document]
talbring: Done

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+    curGetFieldIndex = 0;
+    fieldGetIndexCache.clear();
+    
+    for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
+      for (iVertex = 0; iVertex < geometry->GetnVertex(iMarker); iVertex++){
+        iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
+  
+        if(geometry->node[iPoint]->GetDomain()){
+          
+          buildFieldIndexCache = !fieldIndexCache.size() ? true : false;
```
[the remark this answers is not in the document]
talbring: Done

SU2_CFD/src/output/COutput.cpp
talbring: The rest of the function should not be executed if convergence is true

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1857 @@
...
+}
+
+
+
+
+
+void COutput::Postprocess_HistoryData(CConfig *config){
+   
+  map<string, su2double> Average;
+  map<string, int> Count;
```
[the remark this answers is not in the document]
talbring: Done

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
talbring: I want to avoid allocating additional memory, so I reuse the same memory using two pointers (one for su2double datatype and one for passivedouble). I changed it a little bit so that it is more obvious and that there will be now misleading new/delete.

talbring:
@pcarruscag Thanks for the review! I made modifications to most of the things you were mentioning. For the others I left some comments.

talbring:
Here are some numbers regarding the performance. I used an unsteady 3D case with 83425 points. I took the extreme case of calling the output routines every time-step and only 5 inner iterations. For the first run I enabled the writing of restart and tecplot file:
```
-------------------------- Performance Summary --------------------------
Simulation totals:
Cores:                   8            | DOFs/point:         6           
Points/core:             10428.1      | Ghost points/core:  2045.25     
Wall-clock time (hrs):   0.153299     | Core-hrs:           1.22639     

Preprocessing phase:
Preproc. Time (s):       2.69291      | Preproc. Time (%):  0.487956    

Compute phase:
Compute Time (s):        385.71       | Compute Time (%):   69.8909     
Iteration count:         560          | Avg. s/iter:        0.688769    
Core-s/iter/Mpoints:     55.2191      | Mpoints/s:          0.144877    

Output phase:
Output Time (s):         163.472      | Output Time (%):    29.6211     
Output count:            112          | Avg. s/output:      1.45957     
-------------------------------------------------------------------------
```
Note that the iteration count is the total number of all inner iterations. 30% of the time is spend in the output routine. That may sound like a lot, but lets take a look at the second case, where I still call the output routines, but disabled the writing of any files.

```
-------------------------- Performance Summary --------------------------
Simulation totals:
                   Cores:           8 |          DOFs/point:           6
             Points/core:     10428.1 |   Ghost points/core:     2045.25
   Wall-clock time (hrs):    0.112822 |            Core-hrs:    0.902578

Preprocessing phase:
       Preproc. Time (s):     3.84826 |   Preproc. Time (%):    0.947474

Compute phase:
        Compute Time (s):     399.446 |    Compute Time (%):      98.347
         Iteration count:         560 |         Avg. s/iter:    0.713297
     Core-s/iter/Mpoints:     57.1855 |           Mpoints/s:    0.139896

Output phase:
         Output Time (s):     2.86557 |     Output Time (%):    0.705527
            Output count:         112 |       Avg. s/output:   0.0255854
 Restart Aggr. BW (MB/s):           0 |           MB/s/core:           0
-------------------------------------------------------------------------
```  
You can see that now the time spend in the output routines (including gathering all data, partitioning and sorting) just takes 0.7% of the overall time. Its not nothing, but roughly one order of magnitude smaller than one iteration of the solver. 

I also compared it to the current develop using callgrind on a similar case and for both versions the number of instructions in the output routines is roughly the same (~1.5% of the total number instructions).

talbring:
And a comment on memory: buffers for the output are now allocated once at the beginning and the overall peak memory consumption has been reduced compared to develop.

SU2_CFD/include/output/COutput.hpp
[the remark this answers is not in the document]
pcarruscag: iFreq is an integer and it is being compared with a double literal

SU2_CFD/include/output/filewriter/CFileWriter.hpp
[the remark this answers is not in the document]
pcarruscag: I was actually looking at it from the mixing angle since functions returning times are usually double or integer. No strong feelings about it though.

SU2_CFD/src/drivers/CDriver.cpp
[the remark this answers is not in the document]
pcarruscag: Ok, should dead code be removed then? Not just for this driver that is about to be made redundant but for other areas that you may have commented out (I do not know if there are any).

I ask this because commented code is not possible to maintain as it does not trigger warnings or errors. For example with the work of CVariable I am sure I missed some areas that will probably not compile if commented in...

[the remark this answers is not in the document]
jayantmukho: I am chiming in unannounced, but I am a big proponent of removing commented out code.

SU2_CFD/include/output/CAdjFlowIncOutput.hpp
```diff
@@ -0,0 +1,123 @@
...
+ *  \brief Output class for incompressible flow discrete adjoint problems.
+ *  \author R. Sanchez, T. Albring.
+ *  \date June 5, 2018.
+ */
+class CAdjFlowIncOutput final: public COutput {
+private:
+  
+  bool cont_adj; /*!< \brief Boolean indicating whether we run a cont. adjoint problem */ 
+
+  unsigned short turb_model; /*!< \brief Boolean indicating whether have a turbulence model*/
```
pcarruscag: this boolean is not a bool, maybe the comment is out of date?

SU2_CFD/include/output/CFlowCompOutput.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+/*! \class CFlowCompOutput
+ *  \brief Output class for compressible flow problems.
+ *  \author R. Sanchez, T. Albring.
+ *  \date May 30, 2018.
+ */
+class CFlowCompOutput final: public CFlowOutput {
+private:
+  
+  unsigned short turb_model; //!< Kind of turbulence model
```
pcarruscag: If this, and similar variables are assigned Enum values it would be better to use the appropriate type.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,777 @@
...
+   *  The stored information is printed to the history file and to screen. 
+   * Each individual instance represents a single field (i.e. column) in the history file or on screen.   
+   */
+  struct HistoryOutputField {
+    /*! \brief The name of the field, i.e. the name that is printed in the screen or file header.*/
+    string              fieldName = ""; 
+    /*! \brief The value of the field. */
+    su2double           value = 0.0;
+    /*! \brief The format that is used to print this value to screen. */
+    unsigned short      screenFormat = FORMAT_FIXED;
```
pcarruscag: if possible the enum type should be used here to.

c++ people also recommend using scoped enums (syntax `enum class`) as that guards against unintentional conversions from other types to enum, and makes intentional conversions clear (as an explicit cast is needed)

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,777 @@
...
+      outputGroup(std::move(volumeOutputGroup_)), description(std::move(description_)){}
+  };
+  
+  /*! \brief Associative map to access data stored in the volume output fields by a string identifier. */  
+  std::map<string, VolumeOutputField >          volumeOutput_Map;
+  /*! \brief Vector that contains the keys of the ::volumeOutput_Map in the order of their insertion. */  
+  std::vector<string>                           volumeOutput_List;
+  
+  /*! \brief Vector to cache the positions of the field in the data array */
+  std::vector<short>                            fieldIndexCache;
```
pcarruscag: if I got this right these variables are associated with `offset`, maybe offset should also be a `short`?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1895 @@
...
+void COutput::SetHistoryFile_Header(CConfig *config) { 
+
+  unsigned short iField_Output = 0, 
+      iReqField = 0,
+      iMarker = 0;
+  stringstream out;
+  int width = 20;
+  
+  for (iField_Output = 0; iField_Output < historyOutput_List.size(); iField_Output++){
+    const string &fieldIdentifier = historyOutput_List[iField_Output];
```
pcarruscag: thanks this is indeed easier to follow

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
```diff
@@ -0,0 +1,410 @@
...
+  
+  /*--- Allocate memory to hold the connectivity that we are
+   sending. ---*/
+  
+  connSend = NULL;
+  connSend = new su2double[VARS_PER_POINT*nPoint_Send[size]]();
+  
+  /*--- Allocate the data buffer to hold the sorted data. ---*/
+  assert(sizeof(su2double) >= sizeof (double));
+  dataBuffer = new char[VARS_PER_POINT*nPoint_Recv[size]*sizeof(su2double)];
```
pcarruscag: thanks, this does look like a more standard way to re-use memory, you can possibly use `max(sizeof(su2double),sizeof(passivedouble))` to be on the safe side since the assert will not be compiled in the release version. Or since those values are compile time constants a compilation error can also be triggered using `#ERROR`.
Lastly, the alignment requirement of active and passive doubles is probably different (?) but that should be ok since most compilers align at 16B boundaries, you can always handle that explicitly if you want to be extra safe.

talbring:
Just as a hint: To print a list of all possible output fields, use the `-d` flag, i.e. `SU2_CFD -d config.cfg`.

SU2_CFD/include/output/CFlowCompOutput.hpp
talbring: That would require also to change the corresponding config routine, so I'll leave it like that for the moment. But in general you are right, we should keep that in mind.

SU2_CFD/include/output/COutput.hpp
talbring: Didn't know about the `enum class`. Seems like an appropriate way of doing it. Will be changed in the next commit. I am always learning new stuff! Nice

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
talbring: Using the `max` is probably a good idea.

Alignment should be alright for most of the cases. It probably will only be a problem if someone happens to use a very strange data type. I think we can safely assume that it will be always a combination of `float, double` or `int`.

SU2_CFD/include/output/CDriverOutput.hpp
```diff
@@ -0,0 +1,114 @@
...
+  
+  string bgs_res_name; //!< Block-Gauss seidel residual name
+  bool write_zone;     //!< Boolean indicating whether the individual zones write to screen
+  
+public:
+
+  /*!
+   * \brief Constructor of the class.
+   */
+  CDriverOutput(CConfig *driver_config, CConfig** config, unsigned short nDim);
```
economon: I think using 'Driver' again here might be confusing for folks, since we already have driver classes with different responsibilities (unless you meant to draw comparison). Can we go with something different (manager, multizone, etc) to avoid confusion?

SU2_CFD/include/output/CFlowCompFEMOutput.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+/*! \class CFlowCompFEMOutput
+ *  \brief Output class for the compressible FEM flow output.
+ *  \author R. Sanchez, T. Albring.
+ *  \date May 30, 2018.
+ */
+class CFlowCompFEMOutput final: public CFlowOutput {
+private:
+  
+  unsigned short nVar; //!< Number of variables
```
economon: Take one last pass at the comments to fill in missing \brief's for doxy

SU2_CFD/include/output/CFlowOutput.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "COutput.hpp"
+#include "../variables/CVariable.hpp"
+
+class CFlowOutput : public COutput{
```
economon: Nice 👍 . Like the new intermediate flow output class.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,774 @@
...
+  su2double *cauchySerie;        /*!< \brief Complete Cauchy serial. */
+  su2double oldFunc,             /*!< \brief Old value of the coefficient. */
+  newFunc;                       /*!< \brief Current value of the coefficient. */
+  bool convergence;              /*!< \brief To indicate if the solver  has converged or not. */
+  su2double initResidual;        /*!< \brief Initial value of the residual to evaluate the convergence level. */
+  string convField;              /*!< \brief Name of the field to be monitored for convergence */
+  
+  /*----------------------------- Adaptive CFL ----------------------------*/     
+  
+  su2double rhoResNew,    /*!< New value of the residual for adaptive CFL routine */
```
economon: Adaptive CFL in the output can be removed. It is replaced with a new version located in the solver in #790

SU2_CFD/include/output/COutput.hpp
```diff
@@ -0,0 +1,774 @@
...
+   * \param[in] config - Definition of the particular problem.
+   */
+  void Postprocess_HistoryFields(CConfig *config);
+  
+  /*!
+   * \brief Check whether we should print output.
+   * \param[in] iIter - Current iteration.
+   * \param[in] iFreq - Frequency of output printing.
+   */
+  inline bool PrintOutput(unsigned long iIter, unsigned long iFreq) {
```
economon: Is it possible to write the initial condition with the new output structure (before first iteration)? This is something I have been meaning to add, since it is very useful when setting up and debugging custom problems

SU2_CFD/include/output/filewriter/CFVMDataSorter.hpp
```diff
@@ -0,0 +1,106 @@
...
+   */
+  void SortConnectivity(CConfig *config, CGeometry *geometry, bool val_sort) override;
+  
+  /*!
+   * \brief Get the global index of a point.
+   * \input iPoint - the point ID.
+   * \return Global index of a specific point.
+   */
+  unsigned long GetGlobalIndex(unsigned long iPoint) override { 
+    return linearPartitioner->GetFirstIndexOnRank(rank) + iPoint;
```
economon: glad you found the linear partitioner class!

Common/src/config_structure.cpp
```diff
@@ -1490,43 +1497,16 @@ void CConfig::SetConfig_Options() {
...
-  addUnsignedShortOption("CAUCHY_ELEMS", Cauchy_Elems, 100);
-  /*!\brief CAUCHY_EPS\n DESCRIPTION: Epsilon to control the series convergence \n DEFAULT: 1e-10 \ingroup Config*/
-  addDoubleOption("CAUCHY_EPS", Cauchy_Eps, 1E-10);
-  /*!\brief CAUCHY_FUNC_FLOW
-   *  \n DESCRIPTION: Flow functional for the Cauchy criteria \n OPTIONS: see \link Objective_Map \endlink \n DEFAULT: DRAG_COEFFICIENT \ingroup Config*/
-  addEnumOption("CAUCHY_FUNC_FLOW", Cauchy_Func_Flow, Objective_Map, DRAG_COEFFICIENT);
-  /*!\brief CAUCHY_FUNC_ADJFLOW\n DESCRIPTION: Adjoint functional for the Cauchy criteria.\n OPTIONS: See \link Sens_Map \endlink. \n DEFAULT: SENS_GEOMETRY \ingroup Config*/
-  addEnumOption("CAUCHY_FUNC_ADJFLOW", Cauchy_Func_AdjFlow, Sens_Map, SENS_GEOMETRY);
+  /*!\brief CONV_RESIDUAL_MINVAL\n DESCRIPTION: Min value of the residual (log10 of the residual)\n DEFAULT: -14.0 \ingroup Config*/
+  addDoubleOption("CONV_RESIDUAL_MINVAL", MinLogResidual, -14.0);
```
jayantmukho: This is wicked late in the PR process but the RESIDUAL_REDUCTION option is gone? It seems like the only residual convergence criteria is the CONV_RESIDUAL_MINVAL?

This might pose a significant problem since the actual value of the residual varies wildly between cases, solvers, non-dimensionalizations etc. The value of the residuals also vary between DIRECT and ADJOINT simulations. If running shape optimizations or anything that requires DIRECT and ADJOINT calculations in series, the same value of CONV_RESIDUAL_MINVAL would mean different levels of convergence.

Would it be terribly difficult to reinstate the RESIDUAL_REDUCTION option? I can help if needed.

Another option would be to normalize the residual value such that the initial residual for any case is 0 or 1.

talbring: Using a relative residual is still possible, have a look here: https://su2code.github.io/docs/Guide-to-v7/

Unfortunately its not possible to choose different values for the adjoint and direct runs. We have to think about how we want to do that. There are also other options that may need different values for the adjoint (like the number of iterations). However, this PR does not remove any functionality regarding that.

Common/src/geometry_structure.cpp
```diff
@@ -15381,7 +15456,7 @@ void CPhysicalGeometry::Compute_Nacelle(CConfig *config, bool original_surface,
     
     /*--- Write an output file---*/
     
-    if (config->GetOutput_FileFormat() == PARAVIEW) {
+    if (config->GetTabular_FileFormat() == CSV) {
```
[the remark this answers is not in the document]
talbring: Ops thats a typo. Should be TAB_CSV in that case. Gonna change that. CSV is for volume output files, while TAB_CSV is for tabular structures like the history files and so on.

SU2_CFD/include/output/CDriverOutput.hpp
talbring: Good point. I will rename it to ´CMultizoneOutput´.

SU2_CFD/include/output/COutput.hpp
talbring: Let me check whether there are any regression tests that use adaptive CFL. If not I am gonna remove it.

SU2_CFD/include/output/COutput.hpp
talbring: This probably needs an extra call of the output routines, because right now they are always called at the end of the iteration. Maybe we can do it when you set the number of iterations to 0 ?

Common/src/config_structure.cpp
talbring: What I am gonna do for the moment is to make the `CONV_FIELD` option to accept a list of fields. If the field is not defined in the current solver it will simply be ignored (thats how I do it also when defining the fields for the history and volume output), The convergence criteria then applies to all available fields separately. This way you can at least define separate fields for adjoint and direct runs using the same config file.

SU2_CFD/include/output/COutput.hpp
talbring: There a couple of cases that use adaptive CFL. So lets remove it in the other PR then.

SU2_CFD/include/output/COutput.hpp
talbring: Actually it already works like that. Just set the `INNER_ITER`/`ITER` to 0.

Common/src/config_structure.cpp
talbring: With the last commit I added the possibility of providing a list of fields for `CONV_FIELD`. They can be a mix of residuals and coefficients, e.g. `CONV_FIELD= RMS_DENSITY, DRAG`. If the field is a residual, `CONV_RESIDUAL_MINVAL` is applied. Otherwise, if it is a coefficient, the Cauchy options.

Right now it is *not* possible to choose different values for the convergence settings for each field. That will be too complicated in my opinion. However, to make the behavior of the Cauchy criterion a little bit more independent of the coefficient, I changed it to relative values. That means it is essentially now the relative difference between to consecutive iterations, averaged over the last N iterations. N can be set using the `CAUCHY_ELEMS` options. This should be much more intuitive. 

As mentioned in the comment above, if a field is not defined for the current solver, it will simply be ignored.

jayantmukho: Oh that's perfect. I especially like the ability to specify multiple CONV_FIELD options so that CAUCHY and RESIDUAL convergences can be applied at the same time. Thanks for that. 
I agree that having separate convergence criteria/iterations for DIRECT and ADJOINT calculations would be nice but complicated. A problem for another time
Sorry for not reading that guide fully before posting here. I didn't realize that the relative residual is a new field that could be used. On that note, is there an easy way to see the options available for CONV_FIELD (which I guess is the same as the list of output fields?) I understand that they change depending on the solver being used, but it'd be nice to see the options that would be needed to preserve behavior from the previous versions.

talbring: > Just as a hint: To print a list of all possible output fields, use the `-d` flag, i.e. `SU2_CFD -d config.cfg`.

Every coefficient and residual listed there can be used for `CONV_FIELD`.

jayantmukho: Alright, I am just not a good reader then 🤦‍♂ 

Any chance that can be included in the custom output guide?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -0,0 +1,1980 @@
...
+  /// END_GROUP
+  
+  /// BEGIN_GROUP: TIME_DOMAIN, DESCRIPTION: Time integration information
+  /// Description: The current time
+  AddHistoryOutput("CUR_TIME", "Cur_Time", ScreenOutputFormat::SCIENTIFIC, "TIME_DOMAIN", "Current physical time (s)");
+  /// Description: The current time step
+  AddHistoryOutput("TIME_STEP", "Time_Step", ScreenOutputFormat::SCIENTIFIC, "TIME_DOMAIN", "Current time step (s)");
+ 
+  /// DESCRIPTION: Currently used wall-clock time.
+  AddHistoryOutput("PHYS_TIME",   "Time(sec)", ScreenOutputFormat::SCIENTIFIC, "PHYS_TIME", "Average wall-clock time");
```
economon: Did 'WALL_TIME' change to 'PHYS_TIME'?

SU2_CFD/src/output/filewriter/CSU2MeshFileWriter.cpp
```diff
@@ -0,0 +1,270 @@
...
+    SU2_MPI::Allreduce(&nElem, &offset, 1, MPI_UNSIGNED_LONG, MPI_SUM, MPI_COMM_WORLD);    
+    SU2_MPI::Barrier(MPI_COMM_WORLD);
+#endif
+  }
+  
+  
+  /*--- Write the node coordinates ---*/
+  if (rank == MASTER_NODE){
+    output_file << "NPOIN= " << dataSorter->GetnPointsGlobal();
+//    if (geometry->GetGlobal_nPointDomain() != nGlobal_Poin_Par)
```
economon: You can remove this.. we no longer will have halo nodes in the mesh files

SU2_CFD/obj/Makefile.am
```diff
@@ -85,13 +85,36 @@ libSU2Core_sources = ../src/definition_structure.cpp \
...
+  ../src/output/CFlowCompOutput.cpp \
+  ../src/output/CFlowCompFEMOutput.cpp \
+  ../src/output/CFlowIncOutput.cpp \
+  ../src/output/CHeatOutput.cpp \
+  ../src/output/CBaselineOutput.cpp \
+  ../src/output/CAdjElasticityOutput.cpp \
+  ../src/output/CAdjHeatOutput.cpp \
+  ../src/output/CAdjFlowCompOutput.cpp \
+  ../src/output/CAdjFlowIncOutput.cpp \
+  ../src/output/CDriverOutput.cpp \
```
economon: Don't forget to change this to CMultizoneOutput.cpp

SU2_CFD/src/drivers/CDiscAdjSinglezoneDriver.cpp
```diff
@@ -301,17 +308,17 @@ void CDiscAdjSinglezoneDriver::SetObjFunction(){
 
     solver[FLOW_SOL]->SetTotal_ComboObj(0.0);
 
-    if (config->GetnMarker_Analyze() != 0)
-      output->SpecialOutput_AnalyzeSurface(solver[FLOW_SOL], geometry, config, false);
+//    if (config->GetnMarker_Analyze() != 0)
```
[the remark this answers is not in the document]
talbring: Some of the Special outputs have been removed unfortunately. If people need them, we can consider adding them again at some point.

However the analyze surface stuff is still there. Although you can output the values on all individual surfaces (by adding a 'SURFACE_' to the field and/or group name name), you can only optimize regarding the total sum over all surfaces in 'MARKER_ANALYZE' at the moment. The same is true for the aerodynamic coefficients and 'MARKER_MONITORING'. As a side remark, maybe we should merge both MARKER options into one ...

SU2_CFD/src/output/COutput.cpp
talbring: Don't know what the best name would be. Maybe WALL_TIME is more appropriate.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -0,0 +1,2895 @@
...
+  
+}
+
+void CFlowOutput::WriteMetaData(CConfig *config, CGeometry *geometry){
+    
+  ofstream meta_file;
+  
+  string filename = "flow";
+  
+  filename = config->GetFilename(filename, "meta", curTimeIter);
```
jayantmukho: Here the meta data is written to `flowmeta`, but when the solver tries to read the meta data, the file it tries to read is `flow.meta`. I like the name `flow.meta` better, but it doesn't matter much as long as it is consistent

talbring: Oops thats a typo. Thought I'd solved that already. Thanks!

jayantmukho:
There are a couple of small tweaks that need to be made to the python scripts to make sure the changes I pushed for multipoint optimizations work with this new output structure. There are also a few changes that need to be made to make sure that at the end of the fixed CL mode, the code outputs the flow state before the finite-differencing takes place. 

It might be easiest if I make those changes and push them. Would that be a hassle? Should I just comment on the code and let you make the changes?

Common/src/config_structure.cpp
```diff
@@ -2514,6 +2467,34 @@ void CConfig::SetConfig_Options() {
...
   /* DESCRIPTION: Multipoint mesh filenames, if using different meshes for each point */
   addPythonOption("MULTIPOINT_MESH_FILENAME");
+
+  /*--- options that are used for the output ---*/
+  /*!\par CONFIG_CATEGORY:Output Options\ingroup Config*/
+
+  /* DESCRIPTION: Type of screen output */
+  addStringListOption("SCREEN_OUTPUT", nScreenOutput, ScreenOutput);
+  /* DESCRIPTION: Type of output printed to the history file */
+  addStringListOption("HISTORY_OUTPUT", nHistoryOutput, HistoryOutput);
```
jayantmukho: Would it be possible to define some defaults for the history option that would preserve previous behavior? For example, to include AERO_COEFF, RMS_RES if running a fluid simulation.

I understand that we want to keep the default as simple as possible, but the python scripts (file IO ones) read the history file to get functions like LIFT, DRAG etc. A lot of them assume that AERO_COEFF/COMBO is being written out to the history file. 

Another (less invasive) way to do it would be to set those options if running from the python scripts. But this might be confusing if running the same config file from the python script results in different outputs, than if running without them.

talbring:
> There are a couple of small tweaks that need to be made to the python scripts to make sure the changes I pushed for multipoint optimizations work with this new output structure. There are also a few changes that need to be made to make sure that at the end of the fixed CL mode, the code outputs the flow state before the finite-differencing takes place.
> 
> It might be easiest if I make those changes and push them. Would that be a hassle? Should I just comment on the code and let you make the changes?

Seems like you merged this one already in #780, then modify it there I would say.

Common/src/config_structure.cpp
talbring: I am not a big fan of adding it as defaults just to preserve previous behavior. I would go for just adding it to the output fields. In the future we could think of automatically parsing the available fields, but I don't want to do that since we first have to decide on a general strategy for the whole python framework.

talbring: Ok that's how we can do it at the moment (gonna push it soon): I wrote a small parser that goes through all output files and looks for `AddHistoryOutput`, reads the options and writes them into a python file where this information is assigned to a dictionary.  This dictionary is then used in the python framework for mapping the header and function names and to add missing fields to the `HISTORY_OUTPUT` config option. 

Note however this is a *semi-automatic* procedure: the parser has to be called at the moment every time there is a change in the output fields. But its much more flexible as what we have currently ... And the parser could be used in the  future to fully automate the procedure.

jayantmukho: Thank you! Yeah, the more I use this new output structure, the more I like it. So it is definitely a significant improvement. Just coming across these questions as I merge this with #780

talbring: If something else is missing, lets solve that in a separate PR I'd say. I am going to do a last pass to solve the conflicts/reg. tests tmrw.

economon: I like that idea @talbring. I will also wait for this one to merge before I pull it into #790 and clean things up there (like removing the old CFL adaption routines) and fix the regressions.

talbring:
From my side everything is ready for now.

pcarruscag:
@talbring you seem to have a few trailing blanks in the new files, after the pending PR's are merged can you please clean them up.

jayantmukho:
The .travis.yml file is currently cloning the feature_input_output branch of the TestCases and the Tutorials folders. Should those branches be merged with develop in their respective repos? And consequently, should the .travis.yml file be changed to point to the develop branches?

I ask because I am going to have to do something similar for #780 and I am not sure what the process for doing that would be.

pcarruscag:
That is the process I followed so far

Jayant Mukhopadhaya <notifications@github.com> escreveu no dia quarta,
2/10/2019 à(s) 16:20:

TestCases/parallel_regression_AD.py
```diff
@@ -264,16 +264,16 @@ def main():
...
-    discadj_fsi           = TestCase('discadj_fsi')
-    discadj_fsi.cfg_dir   = "disc_adj_fsi"
-    discadj_fsi.cfg_file  = "configAD_fsi.cfg" 
-    discadj_fsi.test_iter = 3000
-    discadj_fsi.test_vals = [0.958848,-0.157183,0.658415,1.302076] #last 4 columns
-    discadj_fsi.su2_exec  = "parallel_computation.py -f"
-    discadj_fsi.timeout   = 1600
-    discadj_fsi.tol       = 0.00001
-    test_list.append(discadj_fsi)
+#    # Structural model
```
pcarruscag: @talbring I'll start by saying I hope this was an accident, which is possible given the size of the pull request...

I found out yesterday the FSI discrete adjoint capabilities of the code no longer work, needless to say I am not very happy. I am sure this work could have been done without disrupting the functionalities of the code for people that actively use it, develop it, maintain it, and actively interact with the community.
I could have helped keeping this working while an alternative is not available, I clearly showed interest and availability to help, so, if this was not done accidentally, it is unfair to say the least.

There is no need to revert anything, but for what is worth I will disapprove any release related pull requests until the functionally is put back in the code. I will try to help on #774 put I would have preferred to do so without having to rush.

talbring:
@pcarruscag Yes, completely forgot about this. I can probably bring the functionality back in without using or waiting for #774 to make the regression test pass. I'll talk to @rsanfer next week to split the config files into the new multi-zone format.

Issue #716, opened 2019-06-20
Optimize data structures for performance without losing readability/flexibility.


pcarruscag:
Having spent some time on implementation details of linear algebra topics, and a bit of time on small algorithmic improvements to a few numerics classes, I would now like to focus on another performance bottleneck we have. One that always seemed quite daunting to tackle (to me) due to the amount of changes it would require, but recently I had a silly idea to reduce the invasiveness of the "procedure".

Discontiguous data structures are absolute performance killers for low order FVM.
*(skip this paragraph if this is obvious)*
What data structures? All the small ones the solver needs to "visit" to compute residuals.
Why? Mostly because they destroy locality. The useful data stored by consecutive nodes is very far apart. Modern CPU's try to hide the latency of accessing main memory (RAM) by caching frequently used areas of that memory, this effort is very easily undermined by accessing memory at random or predictably but at very large strides.
Let me give you some numbers, for CNSVariable in 2D the average distance between the primitives at consecutive nodes is 200 doubles or 1600 bytes. "What?! is the compiler dumb?" No, virtual methods take up quite a bit of space `sizeof(CNSVariable)` is 424 bytes, before it allocates any data.
And that can only get worse as more functionality is added to the code.

This is one of those "well known" things at least in a qualitative way, but let us try and quantify the cost using an approximation of our typical use case scenario.
*Spoiler alert:* It is up to one order of magnitude, if this does not surprise you feel free to go straight to "Proposed solution".

We will have a variable class that stores some data with some overhead (the 200 doubles):
```
class VariableBase
{
  protected:
    double* primary;
  public:
    VariableBase(int nVar, int overhead);
    virtual ~VariableBase();
    inline double* GetPrimary() {return primary;}
    inline virtual double* GetSecondary() {return nullptr;}
};

VariableBase::VariableBase(int nVar, int overhead)
{
    primary = new double [nVar+overhead];
    for(int i=0; i<nVar; ++i) primary[i] = 1.0;
}
```
All this does is give access to some primary variables held by the base class and allow derived classes to store their own secondary variables, e.g.:
```
class VariableDerived : public VariableBase
{
  private:
    double* secondary;
  public:
    VariableDerived(int nVar, int overhead);
    ~VariableDerived();
    inline double* GetSecondary() {return secondary;}
};
```
Then lets say we have dummy numerics classes.
```
class NumericsBase
{
  public:
    virtual ~NumericsBase();
[...]

Issue #583, opened 2018-09-20
Restructuring the files on the main repo


rsanfer:
Another important topic from the meeting was the need to restructure the files on the main repository, to reduce their sizes and improve both readability and compilation times.

Of course, any file movement will affect the developers with ongoing branches (see #582) so one suggestion is to restructure one file at a time, maybe once every month or so, to reduce the burden to everybody.

I have a number of files in my radar that are priorities:
- driver_structure.cpp (I already have a branch ongoing on this, as some of you know)
- iteration_structure.cpp
- config_structure.cpp
- geometry_structure.cpp
because of their length and the mixture of class structures. There is also a need to improve the includes throughout the code. 

I would like to collect your thoughts.

pcarruscag:
In terms of file organisation I can think of two strategies, one would be grouping them according to parent class, numerics, solver, etc. this would maybe make navigation easier for people who work on the code daily but it does little more than what the naming convention already does. The other would be to group families of classes and that, I think, would help people who are less familiar with the code, so for example numerics, solvers, variables, etc. used for fluid simulations would be separate from the ones used for other physics.

The main re-compilation trigger is probably the config, every file includes it and not all need it. Not much can be done here because most classes do use the config in some way but we could try to minimise how much the dependency spreads by:
- Reviewing what cpp files actually need to include config_structure.hpp
- Including from the cpp and using a forward declaration on the corresponding hpp

Other sensitive "triggers" are the header files of classes with many children (solver, numerics, and variable) splitting the header files for these (maybe by family for small classes to avoid an explosion of hpp files) would allow working on a child class without re-compiling all others so often. The clients of CSolver would only include the header for the abstract class so those too would be shielded from changes in children classes.

economon:
Nice post @pcarruscag : I think we can gain a lot by first breaking up the files, especially the headers, and then being much better about include statements. In particular, we can look at forward declarations, using more guards, and just minimizing the number of includes (I expect some are redundant).

As far as breaking up the files, I think geometry is a good place to start. I can take a shot at this first soon. 

More ideas out there? Anyone have some bandwidth to experiment a little with the headers?

economon:
Bumping this thread...

[...]

PR #725, opened 2019-06-29
Restructure CVariable files

pcarruscag:
## Proposed Changes
Breakup CVariable classes into individual hpp and cpp files, with inlines defined in hpp.
Revise the includes of client classes (CSolver and derived).

## Related Work
Addresses part of #583 
Preparation for #716

## PR Checklist


SU2_DEF/obj/Makefile.am
```diff
@@ -48,28 +48,9 @@ ___bin_SU2_DEF_SOURCES = \
...
-  ../../Common/src/toolboxes/MMS/libSU2_a-CMMSNSTwoHalfSpheresSolution.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CMMSNSUnitQuadSolution.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CMMSNSUnitQuadSolutionWallBC.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CNSUnitQuadSolution.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CRinglebSolution.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CTGVSolution.o \
-  ../../Common/src/toolboxes/MMS/libSU2_a-CUserDefinedSolution.o
+___bin_SU2_DEF_LDADD = \
+  ../../SU2_CFD/obj/libSU2Core.a \
+  ../../Common/lib/libSU2.a
```
pcarruscag: Is there any reason to link against individual object files instead of the static library?

talbring: Actually there is no reason :+1: Thanks!

economon: agreed.. I think that's just legacy behavior from before the libSU2Core.a was added around the time of the in-memory python wrapper

SU2_CFD/include/variables/CAdjEulerVariable.hpp
```diff
@@ -0,0 +1,150 @@
+/*!
+ * \file CAdjEulerVariable.hpp
```
economon: Looks like you are already doing it, but this is a nice opportunity to fix any indentation / white space / line ending issues when creating these new files. Can you please double check?

pcarruscag: I think indentation was already ok, I striped all trailing white spaces now.
There are also many (many...) comments that were copied and do not match the function, that is not a huge deal since these methods explain themselves. I fixed some but I will not fix all of them...

economon: Certainly not expected that you fix them all.. let's get some help on that in a later sweep once merged.

Thanks for the clean up! Can you also verify that all line endings are linux style and not windows format?

pcarruscag: Apparently everything was LF-ended already, at least the IDE I use did not pick up anything.

pcarruscag:
Thanks @talbring. You were right about the includes, even for these classes with lots of inlines everything seems to work fine if only the "client" solver includes the classes it uses.

Issue #721, opened 2019-06-26
Possible correctness issue in how turbulence equations are discretized


pcarruscag:
While working on #718 it was brought to my attention by @economon that CTurbSolver makes its own interpolation / reconstruction of bulk flow variables when computing the convective fluxes. This is mentioned in #403 and #422 but my concern is different.

This reconstruction is not consistent with the one used by the flow solver (except in some very special circumstances) which means the convective fluxes considered for the turbulence variables do not respect the discretized continuity equation.

Now, have I missed something and this is being compensated by some source term?
If I have not missed anything, I think the solution is relatively straight forward, one simply stores the convective fluxes computed in CEulerSolver::Centered/Upwind_Residual to later use in CTurbSolver.

I would be happy to implement this but do let me know if I got something wrong folks.

clarkpede:
As far as I know, there is no compensation in the source terms.  The source terms are implemented pretty closely to how they appear in published (analytical) equations.

pcarruscag:
So I tested this on 4 mesh levels for a NACA0006 at 2.0 degrees AoA, at low (0.6) and high-ish (0.8) Mach number (Roe scheme).
These are the results for low Mach:
![image](https://user-images.githubusercontent.com/38071223/61968547-3b08c680-afd0-11e9-8aae-9705a9441a00.png)
Very small differences between recomputing a mass flux based on primitives ("Reconstructed") or storing the flux computed during discretization of convection ("Consistent").
However, the convergence rate for the latter approach is much worse:
![image](https://user-images.githubusercontent.com/38071223/61968712-99ce4000-afd0-11e9-9c31-dafd7e26e3fb.png)
Which makes sense because we are going from a Gauss-Seidel coupling of flow and turbulence to a half GS, half Jacobi (since the turbulence source terms were still computed with current velocity gradients).
After seeing this I only ran one mesh level (second to finest) at high Mach number and again differences were very small and convergence worse.
Some memory would indeed be saved in the discrete adjoint through the reduction of the number of pre-accumulation input variables, but only 30MB out of almost 9GB for a 2D case without MG.

In summary the current approach seems to strike a good balance between accuracy, cost, and simplicity.

Issue #711, opened 2019-06-13
Linear solver / Jacobian improvements for robust convergence


pcarruscag:
Just to keep track of things, this is what is currently being done in #700, #691, and #705.
If anyone wants to join efforts (there are possibly similar efforts going on in personal branches) either by coding / testing / benchmarking / providing suggestions / etc. please do.

pcarruscag:
#691 introduced Jacobian improvements to some schemes, sadly I could not do anything for the bread and butter of upwind schemes (Roe).
Further improvements to linear solvers and preconditioners will come as part of #789.

economon:
One last Q on this: is there any reason not to make the adjustment to the centered scheme Jacobians the default behavior?

pcarruscag:
No not really, it is a fairly safe thing to change but it only makes sense to do it if the CFL is also increased. I did not make it default at the time because many cases would change over 2 lines of code...
Maybe it can be done now as part of #790? I can help you with the testcase updates.

PR #726, opened 2019-07-01
Consistent mass fluxes between bulk flow and turbulence solvers

pcarruscag:
## Proposed Changes
This is mainly to address #721 but it may have good side effects in other areas (see comments below).
Every turbulent testcase will fail and so far I only tested SA and SST on a simple airfoil case.

## Related Work
Addresses #721 
Should help a bit with #594
May help with #711

## PR Checklist

### DIFF
--- a/SU2_CFD/include/numerics_structure.hpp
+++ b/SU2_CFD/include/numerics_structure.hpp
@@ -70,6 +70,7 @@ protected:
   su2double *Enthalpy_formation;
   su2double Prandtl_Lam;        /*!< \brief Laminar Prandtl's number. */
   su2double Prandtl_Turb;    /*!< \brief Turbulent Prandtl's number. */
+  su2double MassFlux;      /*!< \brief Mass flux across edge. */
   
 public:
   
@@ -1484,9 +1485,14 @@ public:
    * \param[in] d: array that will hold the ordered eigenvalues
    * \param[in] e: supplemental data structure
    * \param[in] n: order of matrix V
-
    */
   static void tql2(su2double **V, su2double *d, su2double *e, unsigned short n);
+  
+  /*!
+   * \brief SetMassFlux
+   * \param[in] val_MassFlux: Mass flux across the edge
+   */
+  inline void SetMassFlux(const su2double val_MassFlux) {MassFlux = val_MassFlux;}
   
 };
 
@@ -2475,7 +2481,7 @@ private:
   /*!
    * \brief Adds any extra variables to AD
    */
-  void ExtraADPreaccIn();
+  inline void ExtraADPreaccIn() {}
 
   /*!
    * \brief SST specific steps in the ComputeResidual method
--- a/SU2_CFD/src/numerics_direct_turbulent.cpp
+++ b/SU2_CFD/src/numerics_direct_turbulent.cpp
@@ -65,38 +65,14 @@ void CUpwScalar::ComputeResidual(su2double *val_residual,
                                        CConfig *config) {
 
   AD::StartPreacc();
-  AD::SetPreaccIn(Normal, nDim);
   AD::SetPreaccIn(TurbVar_i, nVar);  AD::SetPreaccIn(TurbVar_j, nVar);
-  if (grid_movement) {
-    AD::SetPreaccIn(GridVel_i, nDim); AD::SetPreaccIn(GridVel_j, nDim);
-  }
 
   ExtraADPreaccIn();
 
-  Density_i = V_i[nDim+2];
-  Density_j = V_j[nDim+2];
-
-  q_ij = 0.0;
-  if (grid_movement) {
-    for (iDim = 0; iDim < nDim; iDim++) {
-      Velocity_i[iDim] = V_i[iDim+1] - GridVel_i[iDim];
-      Velocity_j[iDim] = V_j[iDim+1] - GridVel_j[iDim];
-      q_ij += 0.5*(Velocity_i[iDim]+Velocity_j[iDim])*Normal[iDim];
-    }
-  }
-  else {
-    for (iDim = 0; iDim < nDim; iDim++) {
-      Velocity_i[iDim] = V_i[iDim+1];
-      Velocity_j[iDim] = V_j[iDim+1];
-      q_ij += 0.5*(Velocity_i[iDim]+Velocity_j[iDim])*Normal[iDim];
-    }
-  }
-
-  a0 = 0.5*(q_ij+fabs(q_ij));
-  a1 = 0.5*(q_ij-fabs(q_ij));
+  a0 = max(0.0, MassFlux);
+  a1 = min(0.0, MassFlux);
 
   FinishResidualCalc(val_residual, val_Jacobian_i, val_Jacobian_j, config);
-
 
   AD::SetPreaccOut(val_residual, nVar);
   AD::EndPreacc();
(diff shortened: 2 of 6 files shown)


SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -65,39 +65,15 @@ void CUpwScalar::ComputeResidual(su2double *val_residual,
...
-      Velocity_i[iDim] = V_i[iDim+1];
-      Velocity_j[iDim] = V_j[iDim+1];
-      q_ij += 0.5*(Velocity_i[iDim]+Velocity_j[iDim])*Normal[iDim];
-    }
-  }
-
-  a0 = 0.5*(q_ij+fabs(q_ij));
-  a1 = 0.5*(q_ij-fabs(q_ij));
+  a0 = max(0.0, MassFlux);
+  a1 = min(0.0, MassFlux);
```
pcarruscag: Here it is possible to set `a0 = -a1` on the assumption that at convergence the summation of fluxes is 0, with that diagonal dominance is guaranteed (at least for the convective part) which improves the condition number of the assembled Jacobian matrix.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -65,39 +65,15 @@ void CUpwScalar::ComputeResidual(su2double *val_residual,
                                        CConfig *config) {
 
   AD::StartPreacc();
-  AD::SetPreaccIn(Normal, nDim);
   AD::SetPreaccIn(TurbVar_i, nVar);  AD::SetPreaccIn(TurbVar_j, nVar);
-  if (grid_movement) {
-    AD::SetPreaccIn(GridVel_i, nDim); AD::SetPreaccIn(GridVel_j, nDim);
-  }
```
pcarruscag: By not computing the mass flux again we greatly reduce the number of input variables for preaccumulation.

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -113,16 +91,18 @@ CUpwSca_TurbSA::~CUpwSca_TurbSA(void) {
...
 void CUpwSca_TurbSA::ExtraADPreaccIn() {
-  AD::SetPreaccIn(V_i, nDim+1); AD::SetPreaccIn(V_j, nDim+1);
+  AD::SetPreaccIn(V_i[nDim+2]); AD::SetPreaccIn(V_j[nDim+2]);
 }
 
 void CUpwSca_TurbSA::FinishResidualCalc(su2double *val_residual, su2double **val_Jacobian_i, su2double **val_Jacobian_j, CConfig *config) {
   
-  val_residual[0] = a0*TurbVar_i[0]+a1*TurbVar_j[0];
+  su2double OneOnAveRho = 2.0/(V_i[nDim+2]+V_j[nDim+2]);
+
```
pcarruscag: SA is funny and so density is still needed here...

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1272,10 +1245,10 @@ void CSourcePieceWise_TurbSST::ComputeResidual(su2double *val_residual, su2doubl
...
    /*--- Implicit part ---*/
 
-   val_Jacobian_i[0][0] = -beta_star*TurbVar_i[1]*Volume;
-   val_Jacobian_i[0][1] = -beta_star*TurbVar_i[0]*Volume;
+   val_Jacobian_i[0][0] = -beta_star*Density_i*TurbVar_i[1]*Volume;
+   val_Jacobian_i[0][1] = -beta_star*Density_i*TurbVar_i[0]*Volume;
    val_Jacobian_i[1][0] = 0.0;
-   val_Jacobian_i[1][1] = -2.0*beta_blended*TurbVar_i[1]*Volume;
+   val_Jacobian_i[1][1] = -2.0*beta_blended*Density_i*TurbVar_i[1]*Volume;
   }
```
jayantmukho: Damn, I've been staring at these lines of code for a while and I never realized that Density was missing from the Jacobian computation. Good catch!

jayantmukho:
I can run this on some OneraM6 meshes to confirm that it works in 3D as well.

SU2_CFD/src/numerics_direct_turbulent.cpp
pcarruscag: It was not missing, before the Jacobians were w.r.t. rho*k and rho*w, that is why in the diffusion part they had to be divided by density, I changed it to k and w (only for the SST model) since the mass flux already contains rho. I am not 100% sure this is correct though... In fact CTurbSolver updates the solution via `node[iPoint]->AddConservativeSolution`. So a better catch by you actually!

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1087,27 +1067,20 @@ CUpwSca_TurbSST::CUpwSca_TurbSST(unsigned short val_nDim,
...
-  
-}
-
 void CUpwSca_TurbSST::FinishResidualCalc(su2double *val_residual,
                                                su2double **val_Jacobian_i,
                                                su2double **val_Jacobian_j,
                                                CConfig *config) {
   
-  val_residual[0] = a0*Density_i*TurbVar_i[0]+a1*Density_j*TurbVar_j[0];
-  val_residual[1] = a0*Density_i*TurbVar_i[1]+a1*Density_j*TurbVar_j[1];
```
clarkpede: What's your rationale behind making these density changes?

I understand the problems with reconstruction and why the value of density is questionable.  But it seems to me like removing density is incorrect.  There's two reasons I think that:

1. The SST variables are updated by the `CVariable::AddConservativeSolution` call.  In that functions, the delta in the variables is assumed to be weighted by density, e.g. rho*k.  Based on that you can conclude that the equations are equations for the density-weighted variables, and the residuals should also be density weighted.
2. In the source term for the SST variables, (`CSourcePieceWise_TurbSST`) there are terms like (`2/3*rho*k*diverg*Volume`).  These residuals are density weighted. To be consistent, all the residuals should be density weighted.

For a converged, steady state solution, I think these changes will only change the conditioning.  Similarly, I expect no impact for a low-Mach test case.  Have you tested these changes on a case like the RAE 2822?

pcarruscag:
Thanks @jayantmukho that would be great. If the meshes you have show grid convergence the results should converge to the same values, both ways of computing the fluxes are second order, but hopefully by using consistent fluxes the error constant is smaller.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -2475,7 +2481,7 @@ class CUpwSca_TurbSST : public CUpwScalar {
   /*!
    * \brief Adds any extra variables to AD
    */
-  void ExtraADPreaccIn();
+  inline void ExtraADPreaccIn() {}
```
clarkpede: The "inline" here is redundant.  Functions defined inside a class definition are automatically "inline."

pcarruscag: In common practice yes but apparently that is not part of the [standard](https://stackoverflow.com/questions/9192077/is-inline-implicit-in-c-member-functions-defined-in-class-definition) (which I have not read) and so some compilers might complain about multiple definition.

clarkpede: Ah ok.  So no harm done in leaving it in.

SU2_CFD/src/numerics_direct_turbulent.cpp
pcarruscag: (I've been looking at the code a bit more to get an answer).
The rationale was: I did not want to divide the mass flux by an average density which would then break that "summation = 0" property we can use to improve conditioning a bit. Following that logic I linearized the other terms w.r.t. k and w instead of rho k and rho w, I did not change the source residual only its Jacobian. But I forgot to see how the solution was being updated so yeah in the end it is wrong.
Seeing how SA models need a volumetric flux anyway I will probably revert those changes to SST.

pcarruscag:
@clarkpede you are right this seems to make no difference to drag and lift regardless of Mach number and coarseness of the grid. But, it is absolutely terrible for convergence in the asymptotic region and therefore to the discrete adjoint, and so it's a garbage idea even if it saves a bit of memory/computation.
I'll post my results in #721 but in the meantime let no one else waste time with this, I owe you and @jayantmukho a beer after this...

clarkpede:
> @clarkpede you are right this seems to make no difference to drag and lift regardless of Mach number and coarseness of the grid. But, it is absolutely terrible for convergence in the asymptotic region and therefore to the discrete adjoint, and so it's a garbage idea even if it saves a bit of memory/computation.
> I'll post my results in #721 but in the meantime let no one else waste time with this, I owe you and @jayantmukho a beer after this...

I wouldn't say it's a garbage idea.  You correctly identified a shortcoming in the code (inconsistent mass fluxes).  That is a problem that **should** be fixed, and you came closer than anyone else has to fixing it.  But it turns out it's a trickier problem than you or I expected.

PR #728, opened 2019-07-13
New CGNS Mesh Reader

### LINKED ISSUE #727: Improvements to existing CGNS reader
The current CGNS mesh reader (serial, ADF format) fails when reading large grids in parallel (>100 mio cells) due to memory exhaustion issues.

Updates to the underlying CGNS library are needed (to v3.4.0), and some additional improvements to the CGNS reader in SU2 are being added. 

This is joint work with @MicK7 and a PR will be opened before long.

economon:
## Proposed Changes
 
This is joint work with @MicK7.

The CGNS mesh reader (ADF format, serial) has been rewritten to improve performance, and the underlying CGNS library has been updated to v3.4.0, along with some other modifications for mixed element grids.

This new version is capable of running meshes > 100 million cells. I was able to run an X-Fine mesh for the high-lift CRM, which contains 210 million grid nodes and 400 million cells. 

If folks have resources to try, please use the attached config file and meshes here: 

https://hiliftpw-ftp.larc.nasa.gov/HiLiftPW3/HL-CRM_Grids/Committee_Grids/B3-HLCRM_UnstrHexPrismPyrTet_PW/FullGap/

[turb_hl_crm.txt](https://github.com/su2code/SU2/files/3388290/turb_hl_crm.txt)

## Related Work

Resolves #727 

## PR Checklist


MicK7:
The externals/ directory changes are intentional. They correct some issues with CGNS library implementation. Those fixes are now being integrated in the develop branch of CGNS library and should be available in the next fix release around the end of the year.

Common/src/geometry_structure.cpp
```diff
@@ -9386,1429 +9383,1580 @@ void CPhysicalGeometry::Read_CGNS_Format_Parallel(CConfig *config, string val_me
...
+         IDs indexed starting from zero. ---*/
         
-        if (datatype != RealDouble) {
-          SU2_MPI::Error("CGNS coordinates are not double precision.", CURRENT_FUNCTION);
+        elemGlobalID[iElem] = elemB[rank] + iElem - 1 - elemOffset[s];
+        
+        switch (elmt_type) {
+          case NODE:
+            currentElem   = "Vertex";
+            elemTypes[iElem] = 1;
```
pcarruscag: I think I saw this kind of information already as I was scrolling down, is there a case to encapsulate this in some `getName`, `getCode` type methods?

Common/src/geometry_structure.cpp
```diff
@@ -9386,1429 +9383,1580 @@ void CPhysicalGeometry::Read_CGNS_Format_Parallel(CConfig *config, string val_me
...
-          int *nElem_Recv = new int[size+1]; nElem_Recv[0] = 0;
-          int *nElem_Flag = new int[size];
-          
-          for (int ii=0; ii < size; ii++) {
-            nElem_Send[ii] = 0;
-            nElem_Recv[ii] = 0;
-            nElem_Flag[ii]= -1;
-          }
-          nElem_Send[size] = 0; nElem_Recv[size] = 0;
+          bound[iMarke
```
pcarruscag: We really need a 2D container class to standardize this type of stuff. Can you mark this as a ToDo or Fixme for when we do have such a container.

Common/src/geometry_structure.cpp
```diff
@@ -8090,10 +8176,14 @@ void CPhysicalGeometry::Read_SU2_Format_Parallel(CConfig *config, string val_mes
...
   map<unsigned long, bool>::const_iterator MI;
+
 #ifdef HAVE_MPI
 #ifdef HAVE_PARMETIS
-  /*--- Initialize a vector for the adjacency information (ParMETIS). ---*/
-  vector< vector<unsigned long> > adj_nodes(nPoint, vector<unsigned long>(0));
+  /*--- Resize the vector for the adjacency information (ParMETIS). ---*/
+  vector< vector<unsigned long> >::iterator it2;
+  adj_nodes.resize(nPoint);
+  for (it2 = adj_nodes.begin(); it2 != adj_nodes.end(); it2++)
```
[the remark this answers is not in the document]
MicK7: Maybe it would be possible to do a adj_nodes.clear() before resizing and then use the second argument of resize since only new object in resize are initialized with the second argument.

Common/src/geometry_structure.cpp
```diff
@@ -6804,6 +6780,152 @@ void CPhysicalGeometry::CompleteCommsAll(int nSends,
...
+  
+  beg_node   = new unsigned long[size];
+  end_node   = new unsigned long[size];
+  nPoint_Lin = new unsigned long[size];
+  nPoint_Cum = new unsigned long[size+1];
+  
+  unsigned long total_points = 0;
+  for (int ii = 0; ii < size; ii++) {
+    nPoint_Lin[ii] = val_npoint_global/size;
+    total_points  += nPoint_Lin[ii];
```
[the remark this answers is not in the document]
MicK7: Nice line. May be we would be clearer by writing it like this:

```
unsigned long quotient = val_npoint_global/size;
unsigned long remainder = val_npoint_global%size;
for (int ii = 0; ii < size; ii++) {
  nPoint_Lin[ii] = quotient + int(ii < remainder);
}
```

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
pcarruscag: Ah that answers it, the vector may already contain something.
Yeah clear and resize or std::fill sound good.

MicK7:
@pcarruscag @economon 
I tried to do some modifications according to the review.
The adjency_vector filling is now splitted in two pass. I am not very confident in the new memory usage so if someone can check that it behaves well it would be nice.

Common/src/geometry_structure.cpp
```diff
@@ -6804,6 +6780,152 @@ void CPhysicalGeometry::CompleteCommsAll(int nSends,
...
+  for (iPoint = 0; iPoint < nPoint; iPoint++) {
+    
+    /*--- For each point, sort the adjacency in ascending order
+     so that we can remove duplicates and complete the size for
+     unique set of adjacent nodes for that point. ---*/
+    
+    sort(adj_nodes[iPoint].begin(), adj_nodes[iPoint].end());
+    it = unique(adj_nodes[iPoint].begin(), adj_nodes[iPoint].end());
+    local_size = it - adj_nodes[iPoint].begin();
+    adj_nodes[iPoint].resize(local_size);
```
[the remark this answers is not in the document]
MicK7: Now it is needed for the two passes strategy

Common/src/geometry_structure.cpp
pcarruscag: Sorry @MicK7 I think git is showing a preview that does not match the code where I put the TODO comment, it is supposed to be line 10626 `passivedouble **gridCoords = new passivedouble*[val_ncoords];`

Common/src/geometry_structure.cpp
```diff
@@ -6792,7 +6792,7 @@ void CPhysicalGeometry::PrepareOffsets(unsigned long val_npoint_global) {
   nPoint_Cum = new unsigned long[size+1];
   
   unsigned long quotient = val_npoint_global/size;
-  unsigned long remainder = val_npoint_global%size;
+  int remainder = int(val_npoint_global%size);
   for (int ii = 0; ii < size; ii++) {
     nPoint_Lin[ii] = quotient + int(ii < remainder);
```
pcarruscag: Cast the comparison to `unsigned long` instead of changing the type of remainder, I wrote `int` because I'm lazy.

Common/src/geometry_structure.cpp
MicK7: Ok I will move it.

Do you think about a light container like this :

```
template<typename T>
struct arr2d {
size_t dim1;
size_t dim2;
T*  arr_;

arr2d(dim1, dim2){
  this->arr_ = new T[dim1*dim2];
  this->dim1 = dim1;
  this->dim2 = dim2;
}

~arr2d() { del [] this->arr_; }

inline T* value( idx1, idx2){ return &(this->arr_[idx1*dim2+idx2]);}

const T& value( idx1, idx2) const { return this->arr_[idx1*dim2+idx2];}

}
```

I think it could be useful in other area if someone wants to experiment SOA structures.

Common/src/geometry_structure.cpp
MicK7: The comparison raised a warning because ii is an int and remainder was an unsigned long.
Now it does not complain but I understand that int(ii<remainder) should be changed to
(unsigned long)(ii<remainder). Correct ?

Common/src/geometry_structure.cpp
```diff
@@ -6804,6 +6780,152 @@ void CPhysicalGeometry::CompleteCommsAll(int nSends,
...
 }
 
+void CPhysicalGeometry::PrepareOffsets(unsigned long val_npoint_global) {
+  
+  /*--- Compute the number of points that will be on each processor.
+   This is a linear partitioning with the addition of a simple load
+   balancing for any remainder points. ---*/
+  
+  beg_node   = new unsigned long[size];
+  end_node   = new unsigned long[size];
```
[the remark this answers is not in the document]
economon: The arrays in question are only allocated once per geometry container / mesh read, but it is always good to have more checks. 

I also want to run this through valgrind again anyway, so I will give it a look.

Common/src/geometry_structure.cpp
pcarruscag: Ah yes... Never mind that then.

Common/src/geometry_structure.cpp
pcarruscag: Yep something of the sort, I am working on one as part of #716 .
For the exact purpose you mention, and maybe in the future support some basic matrix algebra operations that currently we have scattered throughout the code.

MicK7: On second thought if points coordinates were store in a contiguous array we would not even need this buffer as the CGNS library would do the mapping for us.

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
pcarruscag: I tested it with one of the cases in the repo and it seems fine. It is just since the method is public, and anyone can call it from anywhere in the code, it would be good to make sure at least one of those fields is NULL before allocating.

Common/src/geometry_structure.cpp
pcarruscag: Depending on whether we store by rows or columns :)

Common/src/geometry_structure.cpp
```diff
@@ -6806,6 +6782,153 @@ void CPhysicalGeometry::CompleteCommsAll(int nSends,
...
+      adjacency_vector.push_back(adj_nodes[iPoint][jPoint]);
+    }
+    
+    /*--- Increment the starting index for the next point (CSR). ---*/
+    
+    xadj[iPoint+1] = xadj[iPoint] + local_size;
+    
+    /*--- Free vector memory as we go. ---*/
+    
+    vector<unsigned long>().swap(adj_nodes[iPoint]);
```
pcarruscag: Thanks for putting up with me @MicK7, regarding your comment about memory usage, I tested with one of the small cases in the repo and all seems fine.
I would say however that this swap does not do much for the peak memory usage, since everything is allocated at this stage and it is still up to the OS whether you see memory usage dropping by doing this or not. The swap to free the entire vector should be enough.

Common/src/geometry_structure.cpp
```diff
@@ -6806,6 +6782,153 @@ void CPhysicalGeometry::CompleteCommsAll(int nSends,
...
+  vector< vector<unsigned long> >().swap(adj_nodes);
+
+  /*--- Now that we know the size, create the final adjacency array. This
+   is the array that we will feed to ParMETIS for partitioning. ---*/
+  
+  adjacency_size = xadj[nPoint_Lin[rank]];
+  adjacency      = new idx_t[adjacency_size];
+
+  for (iPoint = 0; iPoint < adjacency_size; iPoint++)
+    adjacency[iPoint] = (idx_t)adjacency_vector[iPoint];
```
pcarruscag: In looking at the code more closely, is this an exact copy of the adjacency vector? In the case I tested `adjacency_size` was always equal to the size of the `adjacency_vector`. If this is the case maybe `adjacency` can be filled directly?

Common/src/geometry_structure.cpp
MicK7: This is not a problem. For instance, if points are stored in an interleaved fashion, cgns 3.4 can read directly to memory by using a stride argument (this is how it is done in ParaView/VTK to read CGNS data).

Common/src/geometry_structure.cpp
MicK7: Good question.
adjacency use ` idx_t` type while adjacency_vector use `unsigned long` .  I am not sure if filling directly will not raise issue somewhere else in the code
@economon what is your opinion on this ?

Common/src/geometry_structure.cpp
MicK7: Ok thank you for testing

Common/src/geometry_structure.cpp
economon: I know it is a bit redundant, but I kept them separate bc the idx_t type size can be changed in the ParMETIS header separately from SU2 (we have actually done this recently to make it larger). It's basically for flexibility in case we want to change again.

economon:
It would indeed be good to have better coverage in the tests. I have one or two small cases I will add. @MicK7 : do you have a small mixed-element section case (maybe just a small square and cube)? I only seem to have meshes with sections of a single element type.

Common/src/geometry_structure.cpp
economon: A comment here.. this routine is only executed once in the pre-processing, so raw performance was not as critical for me as making sure that the largest mesh possible can be processed by minimizing peak memory. So, what I was trying to do there, since we need to prepare the adjacency, was to do it one point at a time with an attempt to clear that point's memory immediately so that our peak memory might stay roughly constant during this operation (rather than going to 2x as we shuffle the adjacency).

MicK7: With the two passes, the code try to do a reserve for the adjacency_vector  allowing around 2x the memory if it possible, thus when doing a push_back on the adjacency_vector there is no new allocation in the background. If the reserve fails, it should continue working and grow the adjacency_vector as usual.

Common/src/geometry_structure.cpp
pcarruscag: But since adjacency vector is a temporary what difference would it make to go directly from `adj_nodes` to `adjacency` casting to `idx_t` as you go?

Common/src/geometry_structure.cpp
pcarruscag: I got the idea @economon but std vector grows exponentially when it runs out of capacity and for a while both old and new buffers need to exist to move elements, so even if on the last grow the capacity becomes just right the peak working set might be larger than reserving just enough memory at the beginning... maybe, idk.

Common/src/config_structure.cpp
```diff
@@ -392,7 +420,9 @@ unsigned short CConfig::GetnDim(string val_mesh_filename, unsigned short val_for
 
       /*--- Check whether the supplied file is truly a CGNS file. ---*/
       if ( cg_is_cgns(val_mesh_filename.c_str(), &file_type) != CG_OK ) {
-        SU2_MPI::Error(val_mesh_filename + string(" was not found or is not a CGNS file."),
+        SU2_MPI::Error(val_mesh_filename +
+                       string(" was not found or is not a properly formatted CGNS file.\n") +
+                       string("Note that SU2 expects unstructured CGNS files in ADF data format."),
```
talbring: Thats a good idea to add that hint :)

Common/src/geometry_structure.cpp
```diff
@@ -9387,1429 +9384,1580 @@ void CPhysicalGeometry::Read_CGNS_Format_Parallel(CConfig *config, string val_me
...
+  if (rank == MASTER_NODE) {
+    cout << "Database " << iBase << ", " << basename << ": " << nzones;
+    cout << " zone(s), cell dimension of " << cell_dim << ", physical ";
+    cout << "dimension of " << phys_dim << "." << endl;
+  }
+  
+  /*--- Check if there is more than one zone. Throw an
+   error if the number of zones does not match the expected number. ---*/
+  
+  if (nzones != val_nZone) {
```
talbring: Just to clarify: it is possible to use CGNS for multizone problems right? Is it also possible to have separate mesh files for every zone ?

MicK7: Right now all zones are assumed to be in the same file (and it does not impact performance).
To be able to specify zones with separate mesh file would be interesting. For this purpose, it would be nice to improve the interface of CGNS reader. Indeed, we should be able to select the CGNS base (correponding to solver type fluid, structure, thermal ...) and the zone in a CGNS file that we want to read. We can also rely on CGNS Family to select groups of CGNS zone to load. I fully agree with your idea to put the mesh reader in a separate file.

economon: As @MicK7 mentions, if all zones are in one file, it should work, and I can also confirm that separate files with a single zone each should also work for MZ problems. We should consider improving the reader to take more advantage of the CGNS standard (there are a lot of features we don't use), and when we do, we should try to match the layout that the major mesh packages use when exporting meshes for our kind of multizone / multiphysics problems.

I would like to divide up the geometry files soon in a second pass, at which point, I can move the mesh readers into their own classes potentially (like the output writers).

economon:
Still testing this at scale and expect a couple small changes still, just fyi

MicK7:
Here comes a smaller test case converted from an existing SU2 test case while introducing mixed elements.

MicK7:
file:///data/home/mickael/github/TestCases/sliding_interface/rotating_cylinders/rotating_cylinders.cgns

file:///data/home/mickael/github/TestCases/sliding_interface/rotating_cylinders/rot_cylinders_WA.cfg

file:///data/home/mickael/github/TestCases/sliding_interface/rotating_cylinders/zone_1.cfg

file:///data/home/mickael/github/TestCases/sliding_interface/rotating_cylinders/zone_2.cfg

file:///data/home/mickael/github/TestCases/sliding_interface/rotating_cylinders/zone_3.cfg

MicK7:
Here is a smaller [testcase](https://drive.google.com/open?id=1FK8ijEEh8vbtKEeoqX7Qd0h_yj3U6m7a). It is an existing SU2 test case converted while adding MIXED elements.

Common/src/geometry_structure.cpp
```diff
@@ -10962,6 +10797,157 @@ void CPhysicalGeometry::PrepareCGNSAdjacency(CConfig *config) {
...
+  
+  /* For a specified section, check the element type and return
+   several pieces of information: the string name for the element,
+   the associated VTK type number, and whether this section
+   contains interior or boundary elements. */
+  
+  int nbndry, parent_flag;
+  cgsize_t startE, endE, sizeNeeded;
+  ElementType_t elemType;
+  vector<char> sectionName(CGNS_STRING_SIZE);
```
pcarruscag: This can probably be a "normal" array?

Common/src/geometry_structure.cpp
```diff
@@ -10962,6 +10797,157 @@ void CPhysicalGeometry::PrepareCGNSAdjacency(CConfig *config) {
   
 }
 
+#ifdef HAVE_CGNS
+string CPhysicalGeometry::GetCGNSElementType(CConfig       *config,
```
pcarruscag: Thanks for centralising this.

Common/include/geometry_structure.hpp
```diff
@@ -1912,6 +1912,35 @@ class CPhysicalGeometry : public CGeometry {
...
+#endif
+  
+  /*!
+   * \brief Routine to check whether a CGNS section is an interior or boundary section.
+   * \param[in] config        - definition of the particular problem.
+   * \param[in] val_fn        - CGNS file identifier.
+   * \param[in] val_base      - current CGNS database index.
+   * \param[in] val_zone      - current CGNS zone index.
+   * \param[in] val_section   - current CGNS section index.
+   * \param[in] val_interior  - boolean flag for whether this is an interior or boundary section to be returned.
```
pcarruscag: I think you mean [out] here, same for val_vtk_type in GetCGNSElementType.
Also I think passing these outputs by reference would be safer and more expressive, looking at just the function signature someone could think this is an array.

economon: Yep, will address this. I'm still not perfectly happy with the entire implementation, so a little more refactoring is on the way..

economon:
I have started dividing the geometry classes. First, I created a new set of classes for the FVM mesh readers. There is a base interface that should never need to be modified, and the new CGNS reader is the first implementation. I will move the SU2 ASCII to this framework soon. The new class will make it easy to add new readers or even just implement analytic meshes like squares and cubes directly as child classes.

I will break off a few more small classes in this PR, since they will be easy to handle, but I will likely wait for after this PR to completely divide CPhysicalGeometry just to be safe.

talbring:
@economon Thanks for separating the mesh reader! I think having the possibility to implement analytic meshes can be really beneficial for future testing.

economon:
That's the idea! We should be aiming for many small classes with as few inputs as possible, so that we can easily unit test them. They should have a single responsibility. Many of our legacy classes are doing too many things and can be broken down into smaller classes, especially geometry, for instance. It will take us time to get there, but we're going in the right direction.

economon:
I have added built-in support for 2D rectangular (quad) and 3D box (hexa) meshes, so no mesh file is required. Works in serial and parallel - all partitioning is taken care of for you. For example, if you wanted to run a 65^3 Taylor Green Vortex problem on a triply-periodic box, you could use the following new config options:

```
MESH_FORMAT= BOX
MESH_BOX_SIZE= 65 65 65
MESH_BOX_LENGTH = 6.283185307179586 6.283185307179586 6.283185307179586
MESH_BOX_OFFSET = 0.0 0.0 0.0
```

The marker names are "x_minus" "x_plus" "y_minus" "y_plus" "z_minus" and "z_plus" by default.

economon:
Happy to report that with this branch it is now possible to run a calculation with over 1 billion grid points. I ran freestream through a 1029^3 hexa CGNS grid of a cube as a test.

MicK7:
@economon changes to CRectangularMeshReaderFVM.cpp done .

How about creating a CSU2MeshReaderFVM.cpp to lighten the geometry_structure.cpp file ?

economon:
@Mick7: yep, I’ll look at that next. You may have noticed that the other routines for Loading and preparing adjacency are now general for any mesh reader, so all we need is to move the reader for the ASCII format into its own class. 

@pcarruscag: there is at least one simple stretching function I have in some old code I can put in. Other elements would also be nice. It’s easy to cut the quads into tris (I have the same implementation for this in a python script) and hexas into tets. Might wait for a compelling need to add these features though, but I have no doubt we’ll add them

talbring:
Just a quick question: would it be possible to support arbitrary polyhedral elements using NGON ? I guess for that we would have to change a lot of other things, right?

MicK7:
I agree with @economon  it would be nice to have an option to load a polyhedral mesh as a dual grid defining the control volumes and generate the primal grid from it.

MicK7:
vertex_centered_polygon
As mentioned by @economon supporting polyhedral meshes compatible with existing elements requires some general algorithm to divide the poly and generate the dual mesh. It seems easy in 2D  as seen on the picture but it can be a bit tedious in 3D.

economon:
The native SU2 ASCII mesh reader has been moved to a new class. If everything passes, I would consider this PR complete other than small clean up (like removing Read_SU2_Format_Parallel()).

We can save the next pass at dividing the geometry classes, especially CPhysicalGeometry, for after the release.

economon:
> Here is a smaller [testcase](https://drive.google.com/open?id=1FK8ijEEh8vbtKEeoqX7Qd0h_yj3U6m7a). It is an existing SU2 test case converted while adding MIXED elements.

@MicK7 : does this test work for you? We could consider replacing the existing SU2 mesh with your CGNS mesh for that regression test (that way we keep overhead down by not adding a new test, but get more CGNS coverage).

pcarruscag:
@economon can we merge this one soon? It would make reviewing the output changes easier (which already contain this branch).
You and @MicK7 could still update testcases before the release.

MicK7:
@economon the rotating_cylinder case is running well with CGNS mesh input.
However when it reaches the File output part, I get  an error :
*** Error in `/opt/SU2/SU2_CFD': munmap_chunk(): invalid pointer: 0x000000000b547cd0 ***
======= Backtrace: =========
#0  0x00007ffff66841f7 in raise () from /lib64/libc.so.6
#1  0x00007ffff66858e8 in abort () from /lib64/libc.so.6
#2  0x00007ffff66c3f47 in __libc_message () from /lib64/libc.so.6
#3  0x00007ffff66c9b54 in malloc_printerr () from /lib64/libc.so.6
#4  0x000000000045c95e in COutput::DeallocateConnectivity_Parallel(CConfig*, CGeometry*, bool) ()
#5  0x000000000045c41f in COutput::SetResult_Files_Parallel(CSolver*****, CGeometry****, CConfig**, unsigned long, unsigned short) ()
#6  0x0000000002191a5f in CMultizoneDriver::Output(unsigned long) ()
#7  0x00000000021918b1 in CMultizoneDriver::StartSolver() ()
#8  0x0000000001620fe1 in main ()

I haven't investigated much but I don't have the problem with the SU2 mesh.

economon:
This has the green light to be merged for me. Taking off the WIP tag.

@MicK7 : I have added an error message if a user tries to use a multizone CGNS file for now. It is still possible to run multizone problems by using a separate CGNS single zone file for each zone. Reading multizone CGNS would be a new capability anyway, so I think we should take it out of the critical path and see if we can get it working separately together in the next couple of weeks. That way, we can proceed with merging this and other PRs in the backlog.

MicK7:
@pcarruscag  you can merge this branch if it is ok for you

pcarruscag:
It is more than ok, it is great! I might just start using CGNS because of it.

MicK7:
I solved the issue with the multizone cgns testcase. Even if Family of BC are not read (it is a feature that can be implemented later), the testcase works perfectly fine. It means that multizone CGNS reading works great but the issue is with the output of multizone case in WriteParaView_... .


Indeed when writing multizone, the ParaView writing function did not behave well with the fact that the second zone of the case use MIXED elements and the last zone only use one type of elements and did not reset all the pointer to the previous zone when writing the last zone. SU2 ended up trying to free element that did not belong to the last zone but to the previous one thus a crash.
Since the output structure is under rewriting, this testcase may be a nice add-on to this branch to ensure that the new output structure work correctly.

monika1387:
@economon Is this multizone issue with .cgns files are resolved or still under progress? Because I am trying from last 1 week and having issues in running with. cgns file so just wondering is something wrong in code? 
Also as of now I am trying with FEM_LES (cgns) to FEM_LES (cgns) which didnt work, is it possible to run FEM_LES (cgns) with Unsteady RANS or NS (.su2) pair? It will work? My intent is to extend my LES simulation domain for far field analysis.  Appreciate your advise!

economon:
Hi @monika1387 : at the moment, we throw errors if folks try to use a CGNS mesh that is multi-zone. SU2 currently accepts only single zone, unstructured CGNS files. It is possible to run a multi-zone calculation if you supply two separate single zone CGNS files though, I believe.

Is there a particular error you are seeing that you can share? It is very common for vendors to implement CGNS differently, so depending on where you get the file, we may have compatibility issues. If so, feel free to open a new issue with the details.

monika1387:
Thank you @economon for comment! Yes I am trying with 2 different cgns mesh in multi zone with FEM_LES solver. I think I will open the issue and you can look into it once. Also attached file here the code for the reference if you can tell me quickly is this error is due to some bug in SU2 or just my cfg specs.






[FEL_LES_ERROR.docx](https://github.com/su2code/SU2/files/5396910/FEL_LES_ERROR.docx)

vdweide:
Hi @monika1387. Even if the reading of the CGNS file works with multiple zones, the FEM_LES solver does not support this at the moment. It will require significant changes in the data structures to allow for this, although I am also interested in it. However, this is going to be a long term effort, also because we are currently changing the FEM solver to allow for more efficient vectorization.

monika1387:
@vdweide this makes sense now. Thank you for the confirmation! I will work with single zone or using SU2 atleast as of now.

PR #734, opened 2019-07-19
Dry Run Mode

talbring:
## Proposed Changes
Added a dry run mode (can be enabled by adding -d as flag to the SU2_CFD binary). If this mode is enabled, all of the expensive geometrical preprocessing steps are skipped and no solver is started, but all other steps during the driver construction are executed. This is great to check whether the setup is sane and the physical settings are correct without having to wait until the mesh is read and the partitioning is done.

For the implementation I created a dummy geometry class with zero elements and points and a dummy driver which does essentially nothing. Initially I thought it would be more work, but apparently it almost worked out of the box.

I also added a table output during the solver preprocessing for the DG solver.

Furthermore I moved parts of the Input_Preprocessing() routine dealing with the geometry to the Geometrical_Preprocessing() routine where it should belong.

The changes in SU2_CFD.cpp look more complicated then they are. Just hide white space changes.

## PR Checklist


talbring:
Yep! Thats exactly what I had in mind. I will experiment a little bit with command line parsers. Maybe we can have the modes as arguments.

Common/src/linear_algebra/CSysVector.cpp
```diff
@@ -52,7 +52,7 @@ CSysVector<ScalarType>::CSysVector(const unsigned long & size, const ScalarType
   nVar = 1;
 
   /*--- Check for invalid size, then allocate memory and initialize values ---*/
-  if ( (nElm <= 0) || (nElm >= UINT_MAX) ) {
+  if ( (nElm < 0) || (nElm >= UINT_MAX) ) {
```
pcarruscag: nElm is not going to be less than 0, please change UINT_MAX to ULONG_MAX to be consistent with the other constructors

SU2_CFD/src/driver_dummy.cpp
```diff
@@ -0,0 +1,22 @@
+#include "../include/driver_structure.hpp"
+
+
+CDummyDriver::CDummyDriver(char* confFile,
```
pcarruscag: Should this new file be named CDummyDriver.cpp and put in a "drivers" folder?

SU2_CFD/src/SU2_CFD.cpp
```diff
@@ -52,6 +52,17 @@ int main(int argc, char *argv[]) {
   unsigned short nZone;
   char config_file_name[MAX_STRING_SIZE];
   bool fsi, turbo, zone_specific;
+  bool dry_run;
+  std::string filename = "default";
+  
+  /*--- Command line parsing ---*/
+  
+  CLI::App app{"SU2 v6.2.0 \"Falcon\", The Open-Source CFD Code"};
```
talbring: I added a commandline parser (https://github.com/CLIUtils/CLI11). For the time being the only options are `-d` for the dry run mode and `-h` to show a help message. Let me know what you think!

talbring:
@pcarruscag When you have a minute, can you have a look at the airfoil_rbf_fsi case ? The results seem to change between travis runs (here it passes: https://travis-ci.org/su2code/SU2/jobs/563635347, and here not https://travis-ci.org/su2code/SU2/jobs/563934509) although there was no change in the code. I checked with valgrind and only found and fixed these https://github.com/su2code/SU2/pull/734/commits/3176da1cdaebe146059c5beb07b1418f9935774a, still the case fails.

talbring:
The case is now also failing in #745. Tested it locally and found that the result is compiler dependent. GCC5.4 and GCC9 give different results.

pcarruscag:
Sorry I missed your previous message @talbring.
I imagine this is because of c++11 (which GCC9 might consider by default?)
I will test this and come back to you.

talbring:
No problem. Yep its the c++11 flag. When I enable it for gcc 5.4 I get the same residuals as on travis. Still, they are different from the ones I get with gcc9 ... It definitely has nothing to do with the RBF interpolation, because it also shows up with nearest neighbor. But when running the structural solver alone everything is fine. So I guess it is still something in the interpolation structure or transfer structure that causes that.

pcarruscag:
Fine by me to update reference residuals, since:

- I turned the compiler warnings up to 11 and apart from a lot of name shadowing and "float == float" there is no standard-breaking code in those units.
- I've been compiling the code with the c++11 flag for a long time now and never had any issues.
- That is as low as the residuals get and they oscillate more between consecutive iterations than the change you are seeing. 
- The change is smaller than what one gets by using LAPACK instead of the naive matrix multiplication and factorization methods.

Common/include/geometry_structure.hpp
```diff
@@ -3023,4 +3028,23 @@ class CMultiGridQueue {
 
 };
 
+/*!
+ * \brief The CDummyGeometry class
```
economon: Add some doxy comments when you have a moment

talbring:
Anyone else has a comment ? Like @economon mentioned its not very intrusive and should probably not affect a lot of people.

rsanfer:
LGTM, @talbring. This should not affect a lot of developments and is indeed a very useful feature. We can go ahead and mrege it in.

Issue #730, opened 2019-07-17
Redundant SetVolume_Deformation() calls for DEFORMING mode


economon:
We can remove the calls to SetVolume_Deformation() in between each of the updates of the surface positions for DEFORMING mode in iteration_structure.cpp.. in the case that multiple sequential surface movements are present, they should be additive (they use an increment to the position) and we should be able to call SetVolume_Deformation() just once at the end of the sequence instead.

economon:
WIP

PR #736, opened 2019-07-21
Restructuring driver files

talbring:
## Proposed Changes
Following up on #583 this is a restructuring of the fiels related to the driver. In addition I rearranged preprocessing routines in the driver constructor. I added arguments to the calls to make it more clear what the dependencies are and what is allocated. Furthermore I merged all loops over zones and instances into a single loop. This should make it much more readable. 

## Related Work

## PR Checklist


SU2_CFD/src/iteration_structure.cpp
```diff
@@ -139,80 +139,80 @@ void CIteration::SetGrid_Movement(CGeometry ****geometry_container,
...
-                                         config_container[val_iZone], ExtIter, val_iZone);
+      surface_movement->Surface_Rotating(geometry[MESH_0],
+                                         config, ExtIter, val_iZone);
       /*--- Deform the volume grid around the new boundary locations ---*/
 
       if (rank == MASTER_NODE)
         cout << " Deforming the volume grid." << endl;
-      grid_movement[val_iZone][val_iInst]->SetVolume_Deformation(geometry_container[val_iZone][val_iInst][MESH_0],
-                                           config_container[val_iZone], true);
+      grid_movement->SetVolume_Deformation(geometry[MESH_0],
```
economon: the previous 3 calls to SetVolume_Deformation() are the ones in question in #730 (this last one of the four should stay), can delete now if you like

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -51,34 +51,34 @@ CIteration::CIteration(CConfig *config) {
...
+          CSurfaceMovement *surface_movement,
+          CVolumetricMovement *grid_movement,
+          CSolver ***solver,
+          CConfig *config,
           unsigned long IntIter,
           unsigned long ExtIter)   {
 
   unsigned short iDim;
-  unsigned short Kind_Grid_Movement = config_container[val_iZone]->GetKind_GridMovement();
+  unsigned short Kind_Grid_Movement = config->GetKind_GridMovement();
```
economon: As you have done here, I'm a fan of shortening the container names throughout the code to be consistent, i.e., removing '_container' from config, solver, geometry, etc. These mostly appear at the highest levels of the code hierarchy. Not pressing - we could do it slowly in passes if folks agree.

talbring: I removed the _container from all local variables. The member variables of the driver are still called container to distinguish them from the locals.

SU2_CFD/src/iteration_structure.cpp
talbring: You can go ahead and do that if you want. Do you think it works by just removing them ?

talbring:
@economon This PR should definitely go in before #734.

talbring:
And another note: I deliberately left the old drivers in the file CDriver.cpp as they will be removed soon.

talbring:
Anyone else has a comment ? Otherwise we gonna merge soon.

SU2_CFD/src/iteration_structure.cpp
economon: You're right - let's not introduce any new issues here by accident. I can look at the grid movement change once all of the refactoring is in place.

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -130,73 +130,83 @@ class CDriver {
...
+  /*!
+   * \brief Init_Containers
+   */
+  void SetContainers_Null();
+  
   /*!
    * \brief Read in the config and mesh files.
    */
-  void Input_Preprocessing(SU2_Comm MPICommunicator);
+  void Input_Preprocessing(CConfig **&config, CConfig *&driver_config);
```
pcarruscag: Reference to pointer gives me the creeps, but it does read better.
Please make the preprocessing methods private.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -64,310 +62,98 @@ CDriver::CDriver(char* confFile,
...
-    numerics_container[iZone] = new CNumerics****[nInst[iZone]];
     for (iInst = 0; iInst < nInst[iZone]; iInst++){
-      numerics_container[iZone][iInst] = NULL;
-
-      numerics_container[iZone][iInst] = new CNumerics***[config_container[iZone]->GetnMGLevels()+1];
-
-      Numerics_Preprocessing(numerics_container[iZone], solver_container[iZone],
-          geometry_container[iZone], config_container[iZone], iInst);
+      
+      config_container[iZone]->SetiInst(iInst);
```
pcarruscag: Is it possible to avoid this way of passing information via weird side effects?

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -51,34 +51,34 @@ CIteration::CIteration(CConfig *config) {
...
+  bool adjoint = config->GetContinuous_Adjoint();
+  bool harmonic_balance = (config->GetUnsteady_Simulation() == HARMONIC_BALANCE);
+  bool discrete_adjoint = config->GetDiscrete_Adjoint();
 
   /*--- Only write to screen if this option is enabled ---*/
-  bool Screen_Output = config_container[val_iZone]->GetDeform_Output();
+  bool Screen_Output = config->GetDeform_Output();
+  
+  unsigned short val_iZone = config->GetiZone();
+  unsigned short val_iInst = config->GetiInst();
```
pcarruscag: Can you really please not do this? It is not enough to call a method one needs to remember to set things in config first?
And in this case it is not even the immediate caller of the function that sets the values... Please convince me otherwise but honestly this seems like a terrible idea.
(I known there is precedent with the Runtime stuff...)

talbring: I don't like it either ... But now that I looked at it again, we can just pass the iInst using the ExtIter parameter (its copied as ExtIter anyway)

SU2_CFD/src/drivers/CDriver.cpp
talbring: I'll have to check, probably it can be avoided.

SU2_CFD/include/drivers/CDriver.hpp
talbring: Yeah I am also not a very big fan. If we find a better way of indicating what will be allocated, I am open for that too.

SU2_CFD/src/drivers/CDriver.cpp
talbring: Unfortunately it cannot be avoided for now. Otherwise we have to change the whole Harmonic Balance structure and I don't want to do that at the moment.

SU2_CFD/src/iteration_structure.cpp
talbring: I removed it in that routine. This way also the "if (harmonic_balance)" was not necessary anymore.

pcarruscag: Thanks

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Fair enough but this stuff should be cleaned up before it becomes standard practice to use the config as a messenger.

SU2_CFD/include/drivers/CDriver.hpp
pcarruscag: Still the methods can be private or protected to avoid accidents right?

talbring:
Don't know what is going on, but the PR **is merged** but still open ... So I am gonna close it manually ...

Issue #735, opened 2019-07-19
Possible bug in develop compared to master branch.


EduardoMolina:
**Describe the bug**

Hi SU2 developers,

I will consider this as a possible bug until I hear the opinion of the SU2 community.

As a validation test case for turbulence models, I am running the NASA Hump test case (https://turbmodels.larc.nasa.gov/nasahump_val.html). However, the results obtained from the develop and master branch are quite different, specially the pressure distribution over the hump. The same difference also appears when other numerical scheme is used (SLAU2 instead of Roe). On the other hand, the skin friction seems not be affected. Please see attached figures.

Also, I attached the plot of the residual over iterations, showing that all simulations converged nicely.

Firstly, I thought that this was related to pull request #670 but I ran the same case with an older version of the dev branch and the results were identical. This suggest that the difference started from pull requests that were merged before #670.

I also attached the config and mesh files if someone wants to reproduce.

Thanks to Danah Park (PhD student) for pointed me out.

I appreciate your time and feedback.

Best,

Eduardo

![rans_cf_compare-branch_roe](https://user-images.githubusercontent.com/9790985/61551958-7113dc80-aa0b-11e9-8074-7d754d36a99c.png)
![rans_cf_compare-branch_slau2](https://user-images.githubusercontent.com/9790985/61551959-7113dc80-aa0b-11e9-83c2-918d89d26f07.png)
![rans_cp_compare-branch_roe](https://user-images.githubusercontent.com/9790985/61551960-71ac7300-aa0b-11e9-9f1a-e1be993377cb.png)
![rans_cp_compare-branch_slau2](https://user-images.githubusercontent.com/9790985/61551962-71ac7300-aa0b-11e9-80c6-2ee9390e61f9.png)
![rans_rho0Res_compare-branch](https://user-images.githubusercontent.com/9790985/61551963-71ac7300-aa0b-11e9-962a-7ea04a98ecd4.png)

**To Reproduce**
*Please provide a config file and mesh for a minimal example that recreates the problem.*
[hump_ddes_full.su2.gz](https://github.com/su2code/SU2/files/3412065/hump_ddes_full.su2.gz)
[turb_SA_hump.cfg.txt](https://github.com/su2code/SU2/files/3412070/turb_SA_hump.cfg.txt)

**Desktop (please complete the following information):**
 - OS: [Ubuntu 16.04]
 - C++ compiler and version: [g++ (GCC) 5.4.0]
[...]

PR #740, opened 2019-07-23
Allow BC_Sym_Plane to be used as Slip wall for viscous flow

TobiKattmann:
Hi all,

## Proposed Changes
BC_Sym_Plane was written with only straight lines or planes as symmetry boundaries in mind. Therefore a constant unit normal was used. If symmetry as a slip wall in viscous flows is used the boundary in MARKER_SYM can have some curvature which results in a non-constant unit normal. This is fixed in this PR. Additionally viscous terms are removed from Euler wall and MARKER_EULER can not be used in viscous flows.

Affected Regression tests:
a lot ... see a list in the Conversation below

## Related Work
This PR is meant to resolve issue #735 opened by @EduardoMolina  which was introduced by PR #657 . More details are in the issue

## PR Checklist


EduardoMolina:
Hi @TobiKattmann 
Is this PR ready for review?

Eduardo

TobiKattmann:
Hey @EduardoMolina ,
Yes, I guess it would be good if you can test if this actually fixes your validation Testcase. Maybe also adding the case as regression/validation case could be an idea as there is currently none to my knowledge. 
MfG, Tobi

talbring:
Maybe instead of adding a new one, we can modify an existing one to check that feature. Or we add a new one and remove an existing one. We have to somehow reduce the number of tests ...

TobiKattmann:
For clarification: you mean removing a Testcase from the active regression tests to get faster turnaround from travis, the Testcase-files can stay probably.
The person who put it in place most likely had some feature which (s)he wanted to test. The path can give the purpose of the case away, but not necessarily as it is purely the files. If you (or someone else) have a Testcase in mind which can go out or can be modified I would be thankful. Choosing myself is rather arbitrary.

Just as an additional idea: I removed README.md from the gitignore list of the Testcases in a branch and my plan is to shortly explain what the cases are intended to test specifically [Link](https://github.com/su2code/SU2/tree/feature_periodic_streamwise/TestCases/incomp_navierstokes/streamwise_periodic) (Note: The Readme there is not done at all, but it shows the idea kinda)

Common/src/config_structure.cpp
```diff
@@ -4371,6 +4371,11 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
                    string("is no longer necessary. Please use the original mesh file (prior to SU2_MSH)\n") +
                    string("with the same MARKER_PERIODIC definition in the configuration file.") , CURRENT_FUNCTION);
   }
+
+  if(Viscous && nMarker_Euler != 0) {
+    SU2_MPI::Error(string("Euler wall boundary conditions using MARKER_EULER cannot be used with viscous flows.\n") +
+                   string("To simulate slip walls for viscous flow use MARKER_SYM."), CURRENT_FUNCTION);
```
pcarruscag: This is not user-friendly. There has got to be a way to call BC_Sym when Euler wall is required for viscous flows.
Moreover, doing so will allow you to compute only one tangential vector when dealing with a "true" symmetry plane.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -11431,57 +11415,62 @@ void CEulerSolver::BC_Sym_Plane(CGeometry      *geometry,
...
   for (iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
     
+    /*---------------------------------------------------------------------------------------------*/
+    /*--- Preprocessing: On a symmetry-plane, the Unit-Normal is constant. Therefore a constant ---*/
+    /*---                Unit-Tangential to that Unit-Normal can be prescribed. The computation ---*/
+    /*---                of these vectors is done outside the loop (over all Marker-vertices).  ---*/
+    /*---                The "Normal" in SU2 is an Area-Normal and is most likely not constant  ---*/
+    /*---                on the symmetry-plane.                                                 ---*/
+    /*--- Edit July 2019: In order to use this BC_Sym_Plane method for slip walls in viscous    ---*/
+    /*---                 flow the unit-normal & tangent computation is moved into the loop over---*/
```
pcarruscag: I would just edit your initial comment, mention the issue number if you want, but the historical context should not be required to understand what any code currently does.

pcarruscag:
@TobiKattmann is the incompressible solver the only one that needs the special implementation?
Why don't you augment the BC_Sym function to take in a parameter telling it whether the normal is constant or not (default true) and then implement BC_EulerWall of the incompressible solver as a call to BC_Sym specifying varying normal?
I think this would save you the preprocessing of measuring flatness and keep the implementation of BC_Sym equal for the comp and incomp solvers, so that we may one day put it in common and remove duplication.

TobiKattmann:
Hi @pcarruscag ,
compressible needs this as well. There is some NICF (`Chi_b`, `Kappa_b`,...) stuff in the Euler_Wall implementation which I need some help with to get it right (that's why it is not done yet ). That is also the only compressible-exclusive part that could be a reason to not have one single code for inc+comp. But of course one could do it anyway with an `if(compressible)` conditional to avoid code duplication :)

Towards input boolean: I thought about that as well, but with that preprocessing step it is impossible to get it wrong imo. I personally agree with BC_Sym  always  being flat and Euler allowing to be curvy, but as seen in issue #735 this opinion is not undisputed. By paying the price of checking straightness once, one eliminates the use of Marker_Sym (assumes const normal after your proposal) on curvy slip walls (happened in #735 ), or the possibly correct use of Marker_Euler on flat boundaries which would be a lot more expensive than checking once, right.

SU2_CFD/src/solver_direct_mean.cpp
TobiKattmann: Yep, Going to change that, once the implementation itself is approved

pcarruscag:
I see... well every commercial code I have used to date complained about curvy symmetries, we would need to ask Euclid but I think symmetries are flat too.
At any rate, please no `if(compressible)` :). And on the issue of the compressible implementation couldn't `CNSSolver::BC_Euler_Wall` be implemented as:
```
{
    ...
    CEulerSolver::BC_Euler_Wall(...);
    "viscous pixie dust"
}
```

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -5308,57 +5239,64 @@ void CIncEulerSolver::BC_Sym_Plane(CGeometry      *geometry,
...
+      if (config->GetViscous()) {
+        switch( nDim ) {
+          case 2: {
+            Tangential[0] = -UnitNormal[1];
+            Tangential[1] =  UnitNormal[0];
+            break;
+          }
+          case 3: {
+            /*--- Find the largest entry index of the UnitNormal, and create Tangential vector based on that. ---*/
+            unsigned short Largest, Arbitrary, Zero;
```
pcarruscag: I think the method in the top voted answer should allow you to do this with less logic.
https://math.stackexchange.com/questions/137362/how-to-find-perpendicular-vector-to-another-vector

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -256,14 +256,23 @@ void CTurbSolver::Viscous_Residual(CGeometry *geometry, CSolver **solver_contain
   
 }
 
-void CTurbSolver::BC_Sym_Plane(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics, CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
+void CTurbSolver::BC_Sym_Plane(CGeometry      *geometry, 
+                               CSolver        **solver_container,
```
pcarruscag: @talbring should we start using the short container names everywhere now?

Common/include/geometry_structure.hpp
```diff
@@ -320,6 +320,7 @@ class CGeometry {
 	CVertex*** vertex;		/*!< \brief Boundary Vertex vector (dual grid information). */
   CTurboVertex**** turbovertex; /*!< \brief Boundary Vertex vector ordered for turbomachinery calculation(dual grid information). */
   unsigned long *nVertex;	/*!< \brief Number of vertex for each marker. */
+  bool *bound_is_straight;	/*!< \brief Bool if boundary-marker is straight(2D)/plane(3D) for each marker. */
```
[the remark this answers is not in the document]
TobiKattmann: yep, done

Common/src/geometry_structure.cpp
```diff
@@ -2792,6 +2794,90 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
+        searching ---*/
+      bound_is_straight[iMarker] = true;
+      RefUnitNormal_defined = false; 
+      iVertex = 0;
+
+      while(bound_is_straight[iMarker] == true && iVertex < nVertex[iMarker]) {
+        
+        iPoint = vertex[iMarker][iVertex]->GetNode();
+        
+        if (node[iPoint]->GetDomain()) {
```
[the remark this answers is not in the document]
TobiKattmann: Doesn't matter from my point of view. All points on parallel Send_recv marker are checked 'twice' then but this is alright. Therefore removed as suggested

Common/src/geometry_structure.cpp
```diff
@@ -2792,6 +2794,90 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
+          
+          for (iDim = 0; iDim < nDim; iDim++)
+            UnitNormal[iDim] = -Normal[iDim]/Area;
+          
+          /*--- Check if unit normal is within tolerance of the Reference unit normal. 
+                Reference unit normal = first unit normal found. ---*/
+          if(RefUnitNormal_defined) {
+            for (iDim = 0; iDim < nDim; iDim++) {
+              if( abs(RefUnitNormal[iDim] - UnitNormal[iDim]) > epsilon )
+                bound_is_straight[iMarker] = false; break;
```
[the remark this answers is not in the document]
TobiKattmann: Yes! thanks

Common/src/geometry_structure.cpp
```diff
@@ -2792,6 +2794,90 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
+          Area = sqrt (Area);
+          
+          for (iDim = 0; iDim < nDim; iDim++)
+            UnitNormal[iDim] = -Normal[iDim]/Area;
+          
+          /*--- Check if unit normal is within tolerance of the Reference unit normal. 
+                Reference unit normal = first unit normal found. ---*/
+          if(RefUnitNormal_defined) {
+            for (iDim = 0; iDim < nDim; iDim++) {
+              if( abs(RefUnitNormal[iDim] - UnitNormal[iDim]) > epsilon )
```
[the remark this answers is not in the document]
TobiKattmann: I think the advantage via *unit* normal is that it is independent of any 'real world problem' size. Therefore the chosen epsilon can stay easily constant.  Otherwise I guess some scaling of epsilon with problem dimension would be advisable.

Common/src/geometry_structure.cpp
```diff
@@ -2792,6 +2794,90 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
 
+
+void CGeometry::ComputeSurf_Straightness(CConfig *config, bool print_on_screen) {
+
+  bool RefUnitNormal_defined;
+  unsigned short iMarker,
+                 iDim;
+  unsigned long iVertex,
+                iPoint;
+  su2double epsilon = 1.0e-6,
```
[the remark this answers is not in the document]
TobiKattmann: Done.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -4986,12 +4995,17 @@ class CEulerSolver : public CSolver {
...
    * \param[in] solver_container - Container vector with all the solutions.
-   * \param[in] numerics - Description of the numerical method.
+   * \param[in] conv_numerics - Description of the numerical method.
+   * \param[in] visc_numerics - Description of the numerical method.
    * \param[in] config - Definition of the particular problem.
    * \param[in] val_marker - Surface marker where the boundary condition is applied.
    */
-  void BC_Euler_Wall(CGeometry *geometry, CSolver **solver_container, CNumerics *numerics, CConfig *config,
-                     unsigned short val_marker);
+   void BC_Euler_Wall(CGeometry      *geometry,
```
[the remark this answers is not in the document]
TobiKattmann: done

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: Well, 1. It has to be a unit normal, therefore the division by sqrt(...) and 2. this Largest, Arbitratry, Zero stuff (which I understand bothers you) is there to avoid having the unit vector build upon numerical zeros which I found to give wrong results. Example: normal is n=( 1e-13, 1e-13, 1)  then I could just take t=(n2, -n1, 0) (which is in fact the way v_1 from that top voted answer is build). Then t is no unit vector so I divide each entry by the norm of t which ends up with inaccurate tangentials. That problem occurred in my tests for #657 . I hope that makes it clearer why it is like that. Search the largest entry of the vector and do that switch (like here first and second  value t=(n2, -n1, 0)  with that largest value

Or did I misinterpret your suggestion

TobiKattmann:
> Question: What happens in moving mesh cases if the surface starts out flat and then gets deformed?
I now explicitly stated that: `config->GetKind_GridMovement() != RIGID_MOTION` ->then recompute normals.

Common/src/geometry_structure.cpp
```diff
@@ -2792,6 +2794,90 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
+            for (iDim = 0; iDim < nDim; iDim++) RefUnitNormal[iDim] = UnitNormal[iDim];   
+            RefUnitNormal_defined = true;
+          }
+
+        }//if domain
+      iVertex++;
+      } //while iVertex
+
+    /*--- Print results on screen. ---*/
+    if(rank == MASTER_NODE && print_on_screen) {
```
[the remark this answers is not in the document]
TobiKattmann: Yeah this one was a bit more involved because of the good'ol local vs global marker stuff. 

In this previous version each process had its own bound_is_straight array and accessed that in the BC_sym routine -> therefore the computation was correct i think. It would have happened that on the Master-process the value is `true` and on another `false` for the same global marker (e.g. picture the 2D wedge cases split in the middle), but as only the master-process prints its value to screen ...the info on screen would be wrong.

I changed it now: there is an `Allreduce` in the end such that each process has the same globally correct info. I acknowledge that it is prob not the most elegant implementation, so if there is a hint on how to do it better...

SU2_CFD/src/solver_direct_mean_inc.cpp
pcarruscag: Yeah I got the purpose, but he does a linear combination of two trivially perpendicular vectors.
Which, if the coefficients of the combination are arbitrary, can still give you a small vector.
So one has to choose the coefficients that maximize the magnitude of the potentially zero element (that's one "if").
Worst case scenario this keeps the largest and smallest element in the tangent vector, but if one adds a third trivially perpendicular vector to the result that is avoided:
```
n = ai + bj +ck
if |b| > |c|
  t = bi + (c-a)j - bk
else
  t = ci - cj + (b-a)k
```
Which can actually be written without the conditional. But anyway maybe this is not even expensive at all, I just puzzles.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -11431,57 +11263,77 @@ void CEulerSolver::BC_Sym_Plane(CGeometry      *geometry,
...
     }
   }
-  
+
   /*--- Loop over all the vertices on this boundary marker. ---*/
   for (iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
     
+    if (iVertex == 0 || 
+        geometry->bound_is_straight[val_marker_Global] != true ||
+        config->GetKind_GridMovement() != RIGID_MOTION) {
```
pcarruscag: Is the default grid movement RIGID_MOTION or NONE? I would bake this check into bound_is_straight during the straightness preprocessing, which lets you return early from that function without doing any computation / MPI and reduces the number of places you need to maintain should (when) the grid motions changes.

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
pcarruscag: MPI wise looks ok, idk if there is a better way in SU2 to get a iMarker to iMarkerGlobal correspondence than comparing strings though.
Maybe let each rank decide if its patch should be straight or not and communicate only the messages you want to print to master and don't do the reduction.
At least convert back from global to local in the preprocessing to avoid comparing strings inside the actual boundary condition.

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -5309,56 +5240,77 @@ void CIncEulerSolver::BC_Sym_Plane(CGeometry      *geometry,
...
+        config->GetKind_GridMovement() != RIGID_MOTION) {
+
+      /*---------------------------------------------------------------------------------------------*/
+      /*--- Preprocessing: On a symmetry-plane, the Unit-Normal is constant. Therefore a constant ---*/
+      /*---                Unit-Tangential to that Unit-Normal can be prescribed. The computation ---*/
+      /*---                of these vectors is done outside the loop (over all Marker-vertices).  ---*/
+      /*---                The "Normal" in SU2 is an Area-Normal and is most likely not constant  ---*/
+      /*---                on the symmetry-plane.                                                 ---*/
+      /*--- Edit July 2019: In order to use this BC_Sym_Plane method for slip walls in viscous    ---*/
+      /*---                 flow the unit-normal & tangent computation is moved into the loop over---*/
```
economon: Can you please remove the old comment in favor of the new? No need to keep the history, since git takes care of this for us.. might confuse folks more than it helps

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -5280,6 +5198,19 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
 
+
+void CIncEulerSolver::BC_Euler_Wall(CGeometry      *geometry,
+                                    CSolver        **solver_container,
+                                    CNumerics      *conv_numerics,
+                                    CNumerics      *visc_numerics,
+                                    CConfig        *config,
+                                    unsigned short val_marker) {
+
+  BC_Sym_Plane(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
```
economon: 👍 nice and easy to maintain

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -7820,182 +7820,14 @@ void CEulerSolver::Evaluate_ObjFunc(CConfig *config) {
...
-  delete [] NormalArea;
-  delete [] Velocity_b;
-  delete [] Velocity_i;
-  for (iVar = 0; iVar < nVar; iVar++) {
-    delete [] Jacobian_b[iVar];
-    delete [] DubDu[iVar];
-  }
-  delete [] Jacobian_b;
-  delete [] DubDu;
+  BC_Sym_Plane(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
```
economon: So, I understand that this is the only change needing some additional attention to make sure that the non-ideal cases are handled for inviscid flows or slip walls in viscous problems, correct? @AlbertoGuardone: if you or anyone in your group has a moment, can you please take a look?

AlbertoGuardone: Ciao @economon, ciao @TobiKattmann, we looked into it and it sounds good. 
Implementation is correct: it uses a ghost node with an inverted velocity which does not changes the thermodynamic state. The Jacobian is OK as well. Green light from a NICFD perspective.
Thanks!

SU2_CFD/src/solver_direct_mean.cpp
TobiKattmann: Done

Common/src/geometry_structure.cpp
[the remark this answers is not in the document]
TobiKattmann: Reverted implementation back to local `vector<bool> bound_is_straight` .  But I kept a global version in the `geometry->ComputeSurf_Straightness` to print the result on screen.

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: Ok, got it. And I agree that it is the more elegant solution so I took it.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -7233,12 +7252,17 @@ class CIncEulerSolver : public CSolver {
...
```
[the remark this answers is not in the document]
TobiKattmann: Went over (hopefully) everything and set correct indentation. Good'ol tabs vs spaces problems in IDE :neutral_face:

SU2_CFD/src/solver_direct_mean.cpp
TobiKattmann: Good catch again. As suggested i moved the check into the preprocessing routine.

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: Done

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: :blush:

SU2_CFD/src/solver_direct_mean.cpp
TobiKattmann: Thanks Alberto

TobiKattmann:
As the change of the Euler Wall BC affects quite a few regression tests I list them here. Also for me to keep the overview. Tiny delta = delta_vals all below 0.1, medium delta = one delta_val larger than 0.1, big delta = one delta_val larger than 1.

## serial_regression
- [x] euler/channel, tiny delta
- [x] euler/naca0012, tiny delta

## parallel_regression
- [x] euler/channel, tiny delta
- [x] euler/naca0012, tiny delta

TobiKattmann:
# Failing Regression tests (82 of'em)
The Euler Wall BC (comp. & incomp. affected) is used in many Testcases. All failing tests are listed. 
Box checkked -> values adapted to newer ones. 
tiny delta = all delta_vals below 0.1; 
medium delta = one delta val larger than 0.1 but all below 1.0; 
big delta = one delta_val larger than 1.0.

I use this post for me to keep track of what I already fixed on what is left to do.

## serial_regression
- [x] channel, tiny delta
- [x] naca0012, tiny delta
- [ ] wedge, medium delta
- [ ] oneram6, big delta, restart
- [x] fixedcl_naca0012, tiny delta
- [x] polar_naca0012, tiny delta
- [x] bluntbody, medium delta
- [x] inc_euler_naca0012, tiny delta
- [ ] inc_nozzle, medium delta
- [ ] contadj_naca0012, big delta, cont.Adj.
- [ ] contadj_oneram6, big delta, cont.Adj.
- [ ] contadj_wedge, big delta, cont.Adj.
- [ ] contadj_fixedcl_naca0012, big delta (no computed vals), cont.Adj.
- [x] harmonic_balance, tiny delta
- [x] sine_gust, tiny delta
- [x] aeroelastic, tiny delta
- [x] edge_VW, tiny delta, NICFD
- [x] edge_PPR, tiny delta, NICFD
- [x] uniform_flow, tiny delta
- [x] channel_2D, tiny delta
- [x] channel_3D, tiny delta
- [x] pipe, tiny delta
- [x] rotating_cylinders, medium delta
- [ ] supersonic_vortex_shedding, big delta, (check boundStraigthness)
- [x] fsi2d, tiny delta
- [ ] airfoil_fsi_rbf, big delta (no computed vals)
- [ ] shape_opt_euler_py, big delta
- [ ] opt_multiobj_py, medium delta
- [ ] opt_multiobjcombo_py, medium delta
- [ ] opt_multiobj1surf_py, medium delta
- [x] opt_2surf1obj_py, tiny delta
- [x] pywrapper_naca0012, tiny delta
- [x] pywrapper_aeroelastic, tiny delta
- [x] pywrapper_fsi2d, tiny delta

## parallel_regression
- [x] channel, tiny delta
- [x] naca0012, tiny delta
- [ ] wedge, medium delta
- [ ] oneram6, big delta, restart
- [x] fixedcl_naca0012, tiny delta
- [x] polar_naca0012, tiny delta
- [x] bluntbody, medium delta
- [x] inc_euler_naca0012, tiny delta
- [ ] inc_nozzle, medium delta
- [ ] contadj_naca0012, big delta, cont.Adj.
- [ ] contadj_oneram6, big delta, cont.Adj.
- [ ] contadj_wedge, big delta, cont.Adj.
- [ ] contadj_fixedcl_naca0012, big delta (no computed vals), cont.Adj.
- [x] harmonic_balance, tiny delta
- [x] sine_gust, tiny delta
- [x] aeroelastic, tiny delta
- [x] edge_VW, tiny delta, NICFD
- [x] edge_PPR, tiny delta, NICFD
- [x] uniform_flow, tiny delta
- [x] channel_2D, tiny delta
- [x] channel_3D, tiny delta
- [x] pipe, tiny delta
- [x] rotating_cylinders, medium delta
- [ ] supersonic_vortex_shedding, big delta, (check boundStraigthness)
- [x] fsi2d, tiny delta (todo OUTPUT_FORMAT error in SU2_SOL)
- [ ] airfoil_fsi_rbf, big delta (no computed vals)
- [ ] pywrapper_naca0012, big delta
- [x] pywrapper_aeroelastic, tiny delta
- [x] pywrapper_fsi2d, tiny delta
- [x] inviscid_bump_tutorial, medium delta
- [x] inviscid_wedge_tutorial, tiny delta
- [x] inviscid_onera_tutorial, tiny delta
- [x] design_inv_naca0012, tiny delta
- [x] design_multiobj, big delta

## serial_regression_AD
- [x] discadj_naca0012, tiny delta
- [x] discadj_cylinder3D, medium delta
- [x] discadj_arina2k, tiny delta
- [x] discadj_incomp_NACA0012, tiny delta
- [ ] discadj_euler_py, filediff
- [ ] discadj_multiple_ffd_py, filediff
- [ ] directdiff_euler_py, filediff
- [ ] directdiff_multiple_ffd_py, filediff
- [ ] discadj_multi_py, filediff

## parallel_regression_AD
- [x] discadj_naca0012, tiny delta
- [x] discadj_cylinder3D, medium delta
- [x] discadj_arina2k, tiny delta
- [x] discadj_incomp_NACA0012, tiny delta
- [ ] discadj_fsi_airfoil, filediff

pcarruscag:
@TobiKattmann please put at least a couple of the failing cases under the microscope, see if the convergence rate is the same starting from scratch, if the converged results are the same with wildly different number of cores and so on.
I know it is a pain but we have been tricked by small changes in the past, and for things related with parallelization 2 core regressions can easily give false positives.
I will run some independent tests myself and look into airfoil_fsi_rbf and discadj_fsi_airfoil for you.

TobiKattmann:
@pcarruscag Of course! I already looked at `channel ` and `oneram6` case more detailed also with up to 28 cores. Computed from freestream conditions to convergence I found that the difference in convergence is only a handful of iterations. I will post some residual comparisons for one or two cases. I also want to look a bit more detailed at supersonic wedge and at least one incompressible case. 
And all the cases with big deltas  and the filediff cases need special treatment (and a more in depth look) anyway...

TobiKattmann:
euler oneram6 14cores
sloweuler_oneram6

euler channel 14cores
sloweuler_channel
Hi @pcarruscag (and everyone else of course :) ), Here some selected convergence rate comparisons with of this PR with develop. I also compared results all of those case side by side and the differences (I used contour line positions for most of it) are pretty subtle. Most of the times I double checked that I didn't load the same flow.vtk. At shocks it was a bit more visible especially close the wall but one has to really zoom in for it. Concerning lift / drag coefficient which are written on screen the difference was below 1% throughout the cases I looked at. Some cases were tested with different amount of cores with no mentionable deviance in Residuals.

inc ns bend 2cores
incns_bend

inc euler nozzle 4cores
inceuler_nozzle

euler wedge 14cores
euler_wedge

TobiKattmann:
I do have some issues with the continuous adjoint Testcases. Taking for example the euler_wedge. In the primal, results/convergence are really close. The continuous adjoint case (contadj_wedge) however produces pretty different results especially along the Euler walls (will upload an image later). The cont.adj. case converges though. So I guess I am doing missing something which is specific to the continuous adjoint? 
Maybe someone with experience in that field can help me and take a look. @hlkline maybe?

TobiKattmann:
Alright everyone, this PR is now ready to get merged from my point of view. If there are further concerns regarding particular Testcases please let me know. @pcarruscag did you check airfoil_fsi_rbf and discadj_fsi_airfoil? If yes, your (dis-)approval for this PR would be appreciated (of course from everyone else as well). 

I would like to get this integrated before #790 and I would recommend @economon not to start changing reg-test values before this is merged.

The code still fixes the initial Issue #735

economon:
LGTM. Thanks for the fix and updating all regressions, @TobiKattmann 

Final question: in the end, the Euler and symmetry BCs are identical implementations, so do we have a practical guideline for when to use one or the other (or some error check), or will we just carry both and allow them to be used interchangeably?

TobiKattmann:
@economon Right now, Euler & Sym can be used interchangeably in both (incomp. & comp.) solvers. One could safely remove one but keeping both is reasonable I guess as both BC are expected by users and maybe the implementation deviates in the future.
Both will give the same results, independent of the 'curvyness' of the boundary. 

Bonus info: Comp & Incomp implementations are identical, so if a higher-level FlowSolver is implemented, EulerWall and SymBC could move up there.

talbring:
Thanks @TobiKattmann for that fix and going through the hassle of changing all the reg. test values.

pcarruscag:
@TobiKattmann,
In the Airfoil_2D case, why are the surfaces detected as straight? Also the messages are a bit strange:
Boundary marker leading_edge is a single straight.
Boundary marker 0.0 is a single straight.
Boundary marker pressure_side is a single straight.
Boundary marker 0.0 is a single straight.
Boundary marker suction_side is a single straight.
Boundary marker 0.0 is a single straight.

TobiKattmann:
Hi @pcarruscag ,
You probably mean `airfoil_fsi_rbf` from serial_regression (?)
I get the written on screen:

Boundary marker leading_edge is NOT a single straight.
Boundary marker 0.0 is a single straight.
Boundary marker pressure_side is NOT a single straight.
Boundary marker 0.0 is a single straight.
Boundary marker suction_side is NOT a single straight.
Boundary marker 0.0 is a single straight.

The diff of my local branch with (origin/)develop is empty, so can you double check for me if you are using the latest develop or fix_SlipWall branch?

For the 0.0 markers: In the configFlow.cfg you have 
`MARKER_EULER= ( leading_edge, 0.0, pressure_side, 0.0, suction_side, 0.0)`
I dont understand why there are zeros as nothing has to be prescribed on the Euler Wall. So the 0.0 is interpreted as a marker tag. Removing the 0.0 does not change anything for me in Residuals as well. In the mesh there are also no boundaries defined which are called `0.0` :) 

Does that help or am I on the wrong track?

TobiKattmann:
Maybe also an explanation why I specifically state  `single straight`:
Take a case with two symmetry planes on either side of a channel -> it could be reasonable to put both in the same Marker in the su2 mesh -> both planes are straight for themselves but as I simply loop over all nodes in a marker I then have 2 different unit normals for the same marker -> thats why the specific `single` is used

pcarruscag:
I mean the case in disc_adj_fsi/airfoil_2d for which you updated a filediff regression.

You are probably right on the 0's, those were RANS cases...

TobiKattmann:
For the fluid zone0 the boundaries are detected as NOT straight and for the solid zone1 the fluid markers are still in the Global marker list but not in the local list... thats why they are written as straight on screen (explanation in next paragraph)

In parallel cases this case (process knows the global marker but does not own any nodes of it i.e. it is also not in the local marker list) the process has to assume that the boundary is straight for the mpi communication otherwise we would never have straight boundary-straightness predictions in high process-count simulations. Here the situation is the same... except for it is a serial case

I have to admit I am not really sure how to best tackle that. Can you split the cfg files like in the Airfoil_RBF case (not sure if this is an old vs new driver thing)? I could exclude that structral solver from the surf_straightness check in CDriver::Geometrical_Preprocessing but I am not sure if that possible and if, what boolean to use.

pcarruscag:
Is the solver doing the right thing then? That is all I care about.

That case cannot be split for now, but since that driver will be retired in the near future I would say you don't need to worry (so long as the solver is doing the right thing).

TobiKattmann:
As each zone has its own geometry_container in which the bound_is_straight data is stored, the false data for the solid zone should not interfere with the fluid zone. And as the fluid zone contains the correct straightness info everything should be fine (except for the screen output of course).

PR #744, opened 2019-07-29
Small fix affecting MPI+MG+RANS

pcarruscag:
## Proposed Changes
A small typo was causing the eddy viscosity to not be communicated properly.
Thanks @economon for pointing me in the right direction.

## PR Checklist

### DIFF
--- a/TestCases/parallel_regression.py
+++ b/TestCases/parallel_regression.py
@@ -291,7 +291,7 @@ def main():
     turb_naca0012_sst_restart_mg.cfg_file  = "turb_NACA0012_sst_multigrid_restart.cfg"
     turb_naca0012_sst_restart_mg.test_iter = 20
     turb_naca0012_sst_restart_mg.ntest_vals = 5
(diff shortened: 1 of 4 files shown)


TestCases/parallel_regression.py
```diff
@@ -291,7 +291,7 @@ def main():
     turb_naca0012_sst_restart_mg.cfg_file  = "turb_NACA0012_sst_multigrid_restart.cfg"
     turb_naca0012_sst_restart_mg.test_iter = 20
     turb_naca0012_sst_restart_mg.ntest_vals = 5
-    turb_naca0012_sst_restart_mg.test_vals = [-6.437367, -4.558626, 1.231779, -0.007820, 0.081480] #last 5 columns
+    turb_naca0012_sst_restart_mg.test_vals = [-6.437400, -4.558626, 1.231779, -0.007820, 0.081480] #last 5 columns
```
pcarruscag: This restores the exact value before #652 so we should be ok now.

Issue #685, opened 2019-05-22
Meson build system


talbring:
Hi all,

I experimented a little bit with a different build system to replace automake/autotools. Although it works most of the time fine, it is cumbersome to work with. Especially when modifying the .am files and the configure.ac due to the complicated syntax. Furthermore it is relatively slow.

One alternative would be cmake, however, the syntax is also not very comfortable and it has too many features which we actually don't need. 

So I came across meson and ninja. It is lightweight and very easy to use (setting it up for SU2 took half a day, without any prior knowledge). Compilation time is faster because it builds all object files in parallel. Another advantage is that it is platform independent. Creating packages and binaries and also crafting releases is much simpler with this system (which we can exploit to automate the release process). It is also possible to embed our regression test framework into that process. 

In case you want to try it out, checkout the branch **meson_build**. Everything needed is added as submodule (meson and ninja), so make sure to do a `git submodule update` to clone them. To configure and build use the following commands:

`externals/meson/meson.py build`
`./ninja -C build`

Please test it and give feedback. I have not tried it on windows or mac, so if you are using any of these OS, give it a try!

Note: automake and meson can co-exist together without any interference. So the old way of building SU2 still works in that branch.

talbring:
To see a major advantage, just compare the `meson.build` and the `configure.ac` file ...

davetaflin:
Tim,

I can share some initial impressions. I’m trying to build the Mesa 3D graphics library, which recently switched to Meson and Ninja. They are pretty impressive. It looks like Meson uses both pkg-config and (more recently) CMake under the hood. For specific packages that don’t play nicely with pkg-config, it will look for other similar utilities, such as llvm-config for the LLVM library.

[...]

PR #745, opened 2019-07-29
Introducing Meson Build System

talbring:
## Proposed Changes
As proposed in #685, this introduces the [meson](https://mesonbuild.com/index.html) and ninja build system to eventually replace autotools and make. A [lot of bigger projects](https://mesonbuild.com/Users.html) already use that system (e.g. GNOME, systemd  ...) or are transitioning. Compared to autotools and make, building the code should be faster, especially if you build with more cores, as all objects files are build truly in parallel (with make only objects of the same target are build in parallel). Another advantage is that meson has multiplatform support for linux, mac and windows. 

### Why not cmake ?
Meson files are much easier to read and in general it is much better to maintain. Furthermore it has a great documentation. Finally, meson has no dependencies apart from python and we can easily ship it (as a submodule) since it needs no installation/compilation. 

### What is ninja ?
ninja is a replacement for make and is used to execute the compiler and to resolve dependencies. Meson uses ninja as the default backend. So instead of using configure + make, we use meson + ninja. As with meson, we can ship it as a submodule. It needs a compilation, but again it has no dependencies apart from python.

### How to use ?
The correct meson and ninja versions are added as git submodules. So in case you don't have meson and ninja installed (there should be packages available for all major distros), compile ninja using:

```
cd externals/ninja
./configure.py --bootstrap
cd ..
export PATH=$PWD/externals/ninja:$PWD/externals/meson:$PATH
```
and run meson using
```
meson.py build --prefix=$PWD -Denable-pywrapper=true -Denable-autodiff=true
```
to ocnfigure SU2 with the python wrapper and AD support enabled. MPI support will be enabled by default if a proper mpi installation can be found. 
Finally compile the code with
```
ninja -C build
```

- "build" in the commands  above is a folder where the configuration will be stored. You can configure multiple directories with different configurations (like for a debug build etc)
- ninja uses all available cores in your system by default. Still you can set the number of jobs using `-j N`

### Additional notes
-  Meson requires python 3, so implicitly we get this requirement now. But I think it is less of a problem than some of you probably might think. A lot of problems we have come from the fact the we support python 2 and python 3. As I was setting up travis (getting there in a moment) I noticed that is easily possible to compile the code with python 3 (for the python wrapper) and run the python scripts with python 2 and vice versa, which eventually leads to errors and crashes. If we require(!) version 3 (which should be installed on all machines now ...) we can avoid such problems from the beginning. Furthermore, python 2 is deprecated anyway ...

- I already changed travis to use the new build system. This as one major advantage: we can now properly use ccache and the caching feature of travis (that means only files that have changed will be compiled if a cache is available). This should speed up the runs by a lot.  

- automake and meson can coexist without any problems. Maybe we can keep this compatibility for version 7 and remove it later if we all agree. Also the preconfigure.py should not be necessary anymore then.

- (unit) tests can be easily added and integrated into the build process.

- I will add a proper documentation soon.

## Related Work

#685 

## PR Checklist


TobiKattmann:
Quick question towards: 

> building the code should be faster

Compiling + Linking: I guess it (the speedup) concerns mostly the compilation stage, from what I understand.  I have the feeling upon minor code modifications most of the building process is spend linking. Am I somehow correct with my thoughts?

talbring:
That heavily depends on the filesystem as the linker copies and moves huge amounts of data. On my local machine linking is pretty fast, in the same order of compiling one file. However, on our cluster it is rather slow because its a (parallel) network filesystem. Anyway, in contrast to make, ninja builds ALL objects files in parallel without waiting for a specific static library to be linked. Linking is done as soon as all required object files are ready in parallel with the compilation of other object files. So yeah, for minor code modifications linking can still be the limiting factor, but there is no way around that.

economon:
Very nice job, Tim. I can report that it works fine for me on Mac and a linux cluster.

One suggestion to make our transition easier: if you agree, can you please add a 'bootstrap_meson' file in the root that will set up the build system like our current 'bootstrap' script? I am thinking just something like:
```
git submodule update --init --recursive
cd externals/ninja
./configure.py --bootstrap
cd ..
export PATH=$PWD/externals/ninja:$PWD/externals/meson:$PATH
```
since I found that I usually need to get meson and ninja on my systems. This should also be familiar to folks given our current workflow.

talbring:
I included the bootstrapping in the meson script. A draft of the documentation can be found here:
https://github.com/su2code/su2code.github.io/blob/master/_docs/Build-SU2-From-Source.md

meson.build
```diff
@@ -0,0 +1,82 @@
+project('SU2', 'c', 'cpp', 
+        version: '6.2.0', 
+	license: 'LGPL2',
+	default_options: ['buildtype=release',
```
pcarruscag: What is the default optimization level, and should the warning level not be set consistently with the Wall Wextra recommendation?

meson.build
```diff
@@ -0,0 +1,82 @@
...
+	license: 'LGPL2',
+	default_options: ['buildtype=release', 
+                          'warning_level=0', 
+                          'cpp_std=c++11'])
+
+
+pymod = import('python')
+python = pymod.find_installation()
+
+su2_cpp_args = ['-Wno-unused-parameter', '-Wno-empty-body', '-Wno-format-security']
```
pcarruscag: What do you thing of adding -Wno-deprecated-declarations to this list? (We get tons of that in the AD build)

pcarruscag:
The documentation looks great! (Answered a couple of my questions)
Can you give an example of linking to external libraries? e.g. BLAS
For example I currently need:
```
export CPPFLAGS="-DHAVE_LAPACK"
export LDFLAGS="-L$OPENBLAS/haswell_omp/lib"
export LIBS="-lopenblas -lgfortran -fopenmp"
```

meson.build
talbring: > What is the default optimization level, and should the warning level not be set consistently with the Wall Wextra recommendation?

I added a section on advanced configuration at the link posted above. Hopefully that answers the questions.

meson.build
talbring: Thats a good idea. Let's do that.

talbring:
@pcarruscag I added 

> The documentation looks great! (Answered a couple of my questions)
> Can you give an example of linking to external libraries? e.g. BLAS
> For example I currently need:
> 
> ```
> export CPPFLAGS="-DHAVE_LAPACK"
> export LDFLAGS="-L$OPENBLAS/haswell_omp/lib"
> export LIBS="-lopenblas -lgfortran -fopenmp"
> ```

Well you have two options. First, if the libraries can be found with `pkg-config`, the easiest is to just use 
```
openblas = dependency('openblas')  
su2_deps += openblas
```
Otherwise you'll have to declare the dependency with
```
openblas = declare_dependency(...)
su2_deps += openblas
```
where the arguments for that routine are explained [here](https://mesonbuild.com/Reference-manual.html#declare_dependency)

talbring:
Just found a nice presentation on meson given by the project leader highlighting some of the features: https://www.youtube.com/watch?v=SCZLnopmYBM

talbring:
@pcarruscag Did adding the libraries work for you ?

pcarruscag:
I will try it right now (sorry I got absorbed by 753). If I make it work I volunteer to update the docs with the procedure.

talbring:
@pcarruscag Thanks for the hint. I am using `git -C` to check whether the current dir was cloned using git. Apparently that doesn't work with older versions. I changed the way of checking that. Can you try it again ?

pcarruscag:
@talbring I get this now:
Selection_090
Whether I build from scratch or on top of what I was trying before, all executables are created though.
(sorry it is a picture, the web terminal does not let me copy and > does capture the python stack trace )

talbring:
Thanks for testing :D Now it should work.

talbring:
Why do you think it is confusing ? 

It is either:
```
blas_dep = declare_dependency(include_directories: '/home/talbring/Projects/OpenBLAS-0.3.6/include',
                              link_args:'-L/home/talbring/Projects/OpenBLAS-0.3.6/lib/ -lopenblas')

```

or 

```
blas_dep = dependency('openblas')
```

if `pkgconfig` can find the `*.pc` file.

pcarruscag:
Because of "link_with, libraries to link against"

talbring:
I would always suggest adding the corresponding pkg config path to the `PKG_CONFIG_PATH` variable and use the `dependency` method. This way you get the correct version, additional dependencies, the include path and necessary flags.

talbring:
I am merging this in now. Please have a look at the documentation https://su2code.github.io/docs/Build-SU2-From-Source

Let me know if you have any problems with adding new files and/or external dependencies.

stephansmit:
This works awesome. Except one thing.
I'm trying to build SU2 using Meson inside a Singularity container. However, the configuring of ninja doesn't work there. 
It looks for `python` but if not installed, this is not there, only python3 command is there. It works on my desktop because I have python2 installed there. However, from the documentation it seems I only need python3, in reality it uses both.

Last part of the output of the Singularity build:
[meson_output.txt](https://github.com/su2code/SU2/files/3502142/meson_output.txt)

talbring:
@stephansmit For building ninja `python` has to be available. If its not, on ubuntu (which I assume you are using for the container) you can either use
`ln -s /usr/bin/python3 /usr/bin/python`
or 
`update-alternatives --install /usr/bin/python python /usr/bin/python3 10`
to set this up.

stephansmit:
@talbring Thanks! Works like a charm now! Thanks for implementing this, I this building method is a lot simpler than the previous one.

talbring:
@stephansmit Thanks, I am glad to head that it works for you!

PR #751, opened 2019-08-05
Fix driver/config restructure bug

pcarruscag:
## Proposed Changes
@rsanfer , @talbring please keep the testcases up to date with the new options you are introducing, avoidable bugs are starting to pass...

## PR Checklist

### DIFF
--- a/SU2_CFD/src/drivers/CDriver.cpp
+++ b/SU2_CFD/src/drivers/CDriver.cpp
@@ -37,6 +37,7 @@
 
 #include "../../include/drivers/CDriver.hpp"
 #include "../../include/definition_structure.hpp"
+#include <cassert>
 
 #ifdef VTUNEPROF
 #include <ittnotify.h>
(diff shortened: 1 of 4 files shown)


talbring:
@pcarruscag I would not call that an avoidable bug considering the fact that the regression test passed.

pcarruscag:
It passed because the config options were not updated and so no mesh deformation was specified for any of the FSI cases.

talbring:
If this has no influence on the test values then we should change these cases I would say.

economon:
Let’s try to keep it positive here.. @talbring has an interesting point that the tests should show some sensitivity to major changes like the grid deformation not being specified. Can they be made more sensitive to this?

pcarruscag:
I agree, but some do [pick up](https://github.com/su2code/SU2/commits/develop/TestCases/disc_adj_fsi/Airfoil_2d/config.cfg) those changes, just not all of them.

Anyway, you'll notice I put an assert for the condition that was causing the problem instead of an "mpi error". That is inline with the philosophy that errors/exceptions are to deal with user errors and asserts to help debugging (they are ignored with -DNDEBUG). I can off-course change it.

PR #752, opened 2019-08-05
Topology optimization improvements

pcarruscag:
## Proposed Changes
Couple efficiency/functionality improvements to the topology optimization implementation as detailed in TestCases/fea_topology/config.cfg

Edit: This requires c++11 so it will have to go after the PR that enables it.

## PR Checklist


TestCases/fea_topology/quick_start/settings_refnode.cfg
```diff
@@ -3,6 +3,9 @@ PHYSICAL_PROBLEM= ELASTICITY
 MATH_PROBLEM= DISCRETE_ADJOINT
 DYNAMIC_ANALYSIS= NO
 %
+SINGLEZONE_DRIVER=YES
+ITER= 1
```
pcarruscag: @rsanfer , @talbring can you guys please explain the current maximum iteration stopping criteria in an issue?

talbring: With this we already went a little bit ahead of ourselves. We thought we would make progress a little bit faster. I am currently preparing a documentation for controlling the solver, where I will explain all of this. For the singlezone adjoint solver it is simply EXT_ITER -> ITER. Everything else is still the same.

Common/src/geometry_structure.cpp
```diff
@@ -3337,11 +3336,11 @@ void CGeometry::FilterValuesAtElementCG(const vector<su2double> filter_radius,
         /*--- morphology kernels (image processing) ---*/
         case DILATE_MORPH_FILTER: case ERODE_MORPH_FILTER:
             
-          for (vector<long>::iterator it=neighbours.begin(); it!=neighbours.end(); ++it)
+          for (auto idx : neighbours)
```
talbring: We have to be careful not to overuse the `auto` keyword. It does not always help with readability.

pcarruscag:
Thanks for the reviews guys. At the moment the optimization framework does not support the topology variables, there is a script in SU2_PY for this feature that runs a testcase with hardcoded parameters...
It is not an easy thing to run a tutorial for :/

PR #753, opened 2019-08-05
Contiguous CVariable

pcarruscag:
## Proposed Changes
This is the beginning of the work proposed in #716 .
Comments will follow as work progresses.
I would appreciate if we could keep the restructuring of files that use CVariable heavily for when this is done.

## Related Work
#716

## PR Checklist


Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -0,0 +1,370 @@
+#pragma once
```
pcarruscag: This is the container class I am planning to use to replace the data members of CVariable.
Where CVariable has a scalar it will have a column vector, and where it has an array it will have a row-major matrix (so it can still return the solution as a pointer to the beginning of a row), where it has a matrix we'll see... this supports nesting but at the moment that is only efficient if the inner type has a static size.

Following some of the comments in #716, it supports aligned memory allocation (c++11 required) and compile-time sizes (no dynamic allocation), which may eventually be used to create an "easily-vectorizable" type (e.g. vector of 4 doubles) to explore vectorization of the numerics.

I am hoping to eventually build some basic dense linear algebra capabilities to collect some of the functionality we have throughout the code. I am not a meta-programming expert though, so the interface for that might not be the most user friendly.

I will add the SU2 header to this file and fix the formatting soonish.

MicK7: Looks very good. It would also be nice to store Nodes of CEdge class in such a container.

pcarruscag: Thanks @MicK7, CEdge does not have a lot of overhead since it does not have virtual members, but it does have some... I'll get there at some point.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -164,7 +164,7 @@ class CSolver {
   CSysVector<su2double> OutputVariables;    /*!< \brief vector to store the extra variables to be written. */
   string* OutputHeadingNames; /*< \brief vector of strings to store the headings for the exra variables */
   
-  CVariable** node;  /*!< \brief Vector which the define the variables for each problem. */
+  CVariable* node;  /*!< \brief Vector which the define the variables for each problem. */
```
pcarruscag: First step done, I updated the interface of CVariable so that "iPoint" is passed as the first argument to its methods, which means we drop one "*" here. node is still polymorphic but is no longer an array.

SU2_CFD/include/solver_structure.inl
```diff
@@ -2438,16 +2435,15 @@ inline void CTurbSolver::SetInlet_TurbVar(unsigned short val_marker, unsigned lo
...
 }
 
 inline su2double CTurbSASolver::GetNuTilde_Inf(void) { return nu_tilde_Inf; }
 
 inline void CTurbSSTSolver::SetFreeStream_Solution(CConfig *config){
   for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
-    node[iPoint]->SetSolution(0, kine_Inf);
-    node[iPoint]->SetSolution(1, omega_Inf);
+    node->SetSolution(iPoint, 0, kine_Inf);
+    node->SetSolution(iPoint, 1, omega_Inf);
```
pcarruscag: Which means that in a lot of places we have this kind of change.
In practice some methods of CVariable (e.g. moving the solution between time levels) could now operate on all points which could simplify some areas of the code and improve performance, but for now I think it is better to make the changes as simple as possible to minimize conflicts with other developments.

I tested this on a simple 2D case with 130k cells (Roe + SST) and it was about 30% faster. Not world shattering, I know, but hey it only took 3 days of search and replace xD

SU2_CFD/include/solver_structure.hpp
```diff
@@ -4684,6 +4682,8 @@ class CEulerSolver : public CSolver {
 
   CFluidModel  *FluidModel;  /*!< \brief fluid model used in the solver */
 
+  CEulerVariable* snode;  /*!< \brief The highest level in the variable hierarchy this solver can safely use. */
```
pcarruscag: As I mentioned in #716 the solver methods that access primitives (gradient and limiter computation) suffer from the accessors to those variables being virtual (therefore not inline-able if called on pointer to base class). To fix that, in some solvers I have defined a `snode` variable of the type that solver can safely assume will be instantiated.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -799,22 +796,22 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config, unsigned short
 
   /*--- Initialize the solution to the far-field state everywhere. ---*/
 
-  for (iPoint = 0; iPoint < nPoint; iPoint++)
-    node[iPoint] = new CEulerVariable(Density_Inf, Velocity_Inf, Energy_Inf, nDim, nVar, config);
+  node = new CEulerVariable(Density_Inf, Velocity_Inf, Energy_Inf, nPoint, nDim, nVar, config);
+  snode = static_cast<CEulerVariable*>(node);
```
pcarruscag: This "special" node is just a cast of the normal node, since this is done in only one place I think there is little risk of initializing it incorrectly.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -2897,15 +2894,15 @@ void CEulerSolver::SetInitialCondition(CGeometry **geometry, CSolver ***solver_c
           
           if (Distance < Radius) {
             
-            solver_container[iMesh][FLOW_SOL]->node[iPoint]->SetSolution(0, Density_CylND);
+            solver_container[iMesh][FLOW_SOL]->node->SetSolution(iPoint, 0, Density_CylND);
```
pcarruscag: `snode` is only used inside the solver (i.e. on uses equivalent to `this->snode`) if we are manipulating the FLOW_SOL position of the solver_container we still need to access `node` (well technically we could static cast but that would be ugly).

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -225,22 +225,26 @@ CFEASolver::CFEASolver(CGeometry *geometry, CConfig *config) : CSolver() {
...
+      if (iVertex >= 0) {
+        snode->Set_isVertex(iPoint,true);
+        break;
+      }
     }
-    if (isVertex) node[iPoint] = new CFEABoundVariable(SolRest, nDim, nVar, config);
-    else          node[iPoint] = new CFEAVariable(SolRest, nDim, nVar, config);
-  }
-  
+  snode->AllocateBoundaryVariables(config);
```
pcarruscag: @rsanfer this should address your concerns about memory wastage, storage for surface forces is allocated after the "isVertex" booleans are set. If the code is compiled without NDEBUG checks are performed to make sure the data is not being accessed before a call to Allocate, or after new calls to Set_isVertex.
Would this strategy work for the new mesh solver?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
-   */
-  inline void Set_OldSolution_Adj(void) {
-    for (unsigned short iVar = 0; iVar < nVar; iVar++)
-      Solution_Adj_Old[iVar] = Solution[iVar];
-  }
+//  /*!
+//   * \brief Set old discrete adjoint variables to the current value of the adjoint variables.
+//   * \param[in] iPoint - Point index.
+//   */
+//  inline void Set_OldSolution_Adj(Idx_t iPoint) {
```
pcarruscag: here

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
-  /*!
-   * \brief Set to zero a particular solution.
-   */
-  inline void SetSolutionZero(unsigned short val_var) {Solution[val_var] = 0.0;}
+//  /*!
+//   * \brief Specify a vector to set the velocity components of the old solution.
+//   * \param[in] iPoint - Point index.
+//   * \param[in] val_vector - Pointer to the vector.
+//   */
+//  inline void SetVelSolutionOldVector(Idx_t iPoint, const su2double *val_vector) {
```
pcarruscag: here

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
-   * \brief Get the value of the preconditioner Beta.
-   * \return Value of the low Mach preconditioner variable Beta
-   */
-  inline virtual su2double GetPreconditioner_Beta() {return 0; }
+//  /*!
+//   * \brief Get the value of the preconditioner Beta.
+//   * \param[in] iPoint - Point index.
+//   * \return Value of the low Mach preconditioner variable Beta
+//   */
+//  inline virtual su2double GetPreconditioner_Beta(Idx_t iPoint) const { return 0.0; }
```
pcarruscag: These methods for Precond_Beta are only used for the Roe-Turkel scheme, but we only set the value and never get it back.
Also the value is set for iPoint and jPoint based on a result from numerics (which means each value is overwritten multiple times in the edge loop).

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
    * \brief A virtual member.
+   * \param[in] iPoint - Point index.
    * \param[in] val_vector - Direction of projection.
    * \return Value of the projected velocity.
    */
-  inline virtual su2double GetProjVel(su2double *val_vector) {return 0; }
+  inline virtual su2double GetProjVel(Idx_t iPoint, const su2double *val_vector) const { return 0.0; }
 
   /*!
    * \brief A virtual membe
```
pcarruscag: here

pcarruscag: The gradients and limiters of secondaries are not used anywhere (but data is allocated for them). Is this legacy?

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -37,155 +37,73 @@
...
 
-  if (Rmatrix != NULL) {
-    for (iDim = 0; iDim < nDim; iDim++)
-      delete [] Rmatrix[iDim];
-    delete [] Rmatrix;
-  }
+void CVariable::Set_Solution() { Solution = Solution_Old; }
 
-}
+void CVariable::Set_Solution_time_n() { Solution_time_n = Solution; }
```
pcarruscag: Some methods that were used to operate on all nodes are now in the cpp and their use no longer requires writing a for loop.

.travis.yml
```diff
@@ -23,7 +23,8 @@ python:
 
 env:
     global:
-        CXXFLAGS="-O3 -Wall -Wextra -Wno-unused-parameter -Wno-empty-body -Wno-format-security" 
+        CXXFLAGS="-O3 -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-empty-body -Wno-format-security"
+        CPPFLAGS="-DNDEBUG"
```
pcarruscag: This flag is required to get the best performance out of the container class as it disables the in-range checks when accessing the data. I will update the build system accordingly.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
+
 /*!
  * \class CVariable
  * \brief Main class for defining the variables.
  * \author F. Palacios
  */
 class CVariable {
 protected:
+  using Idx_t = unsigned long;
+  template<class T> using TVec_t = C2DContainer<Idx_t, T, StorageType::ColumnMajor, 64, DynamicSize, 1>;
```
talbring: Maybe we find a more descriptive name for all these types? We can very well just write it out.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -381,14 +381,14 @@ bool CDriver::ComputeVertexForces(unsigned short iMarker, unsigned short iVertex
     Area = sqrt(AreaSquare);
 
     /*--- Get the values of pressure and viscosity ---*/
-    Pn = solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->node[iPoint]->GetPressure();
+    Pn = solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->node->GetPressure(iPoint);
     if (viscous_flow) {
       for(iDim=0; iDim<nDim; iDim++) {
         for(jDim=0; jDim<nDim; jDim++) {
-          Grad_Vel[iDim][jDim] = solver_container[ZONE_0][INST_0][FinestMesh][FLOW_SOL]->node[iPoint]->GetGradient_Primitive(iDim+1, jDim);
+          Grad_Vel[iDim][jDim] = solver_container[ZONE_0][INST_0][FinestMesh][FLOW_SOL]->node->GetGradient_Primitive(iPoint, iDim+1, jDim);
```
talbring: If it stays like that we could just remove the `node` ?

SU2_CFD/include/variables/CTurbSSTVariable.hpp
```diff
@@ -46,58 +46,54 @@
...
-                   su2double *constants, CConfig *config);
-
+  CTurbSSTVariable(su2double kine, su2double omega, su2double mut, Idx_t npoint,
+                   Idx_t ndim, Idx_t nvar, const su2double* constants, CConfig *config);
+                   
   /*!
    * \brief Destructor of the class.
    */
-  ~CTurbSSTVariable(void);
+  ~CTurbSSTVariable() = default;
```
talbring: Are all of these C++11 keywords necessary for functioning and/or performance ?

pcarruscag: `final` might be as the compiler will then know which virtual methods will no longer be redefined.
`override` was to guarantee that I was indeed overriding the virtual methods of base (easy mistake to make when changing function signatures). `default` means {} in that case. I think those keywords are good documentation for someone reading the code.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: I am open to suggestions, the short names were mostly to avoid carpal tunnel syndrome, it's easy to replace now.

SU2_CFD/src/python_wrapper_structure.cpp
pcarruscag: The interface to CSolver is already pretty big, if we concatenate it with the one for CVariable it is possibly going to get a bit confusing.

SU2_CFD/include/solver_structure.inl
economon: 👍 nice job. This is roughly what we saw for the optimizations performed for the C&F paper. We can check w a profiler to see how much you have improved the cache efficiency.. at some point we know we will be memory bandwidth-bound in the FVM solver, but I think we still have some more room to keep improving things until we hit that point.

Btw, the methods you mention that could operate on all points within CVariable could be good candidates to prototype some SIMD vectorization, since the implementation could be hidden inside CVariable

pcarruscag: Thanks, I played around with perf a bit already. I am trying to put together a representative case to get some numbers and help decide where to strike next.
Most of the methods I moved are array copies, clang and icc vectorize those, gcc is a bit more paranoid, no luck without pragmas.

economon: I have a few ideas if you are looking for some more targets...

You have already made memory access more efficient, so another option to improve performance is to increase the arithmetic intensity per access. For the FVM solver, our algorithms leave us with low flop/byte by nature, but there are still a few things we can try. With relatively minor adjustments, I think we can fuse some kernels to give us more work per edge/point loop. Any edge/point loops that can be processed concurrently could be merged, again as long as it doesn't hurt our flexibility/readability too badly. Goal would be to minimize the total number of these loops per iteration of the solver while stacking up as much compute as possible inside them.

For example, the Upwind_Residual and Viscous_Residual routines could be fused such that one edge loop takes care of both calls to the inviscid and viscous ComputeResidual() routines, since all of the data for both should be readily available. This could be done for other edge/point loops during the preprocessing (limiters, updating primitives, gradients), as long as we are careful with correctness (the order and MPI are important) and don't limit our choice of algorithms.

Also: it would be good to verify that the compiler is able to correctly unroll any loops for nDim and nVar down in the compute intensive kernels.

pcarruscag: I am glad you bring up fusing Upwind and Viscous residual as that is definitely where the next "easy" performance gain is hiding, and we would need that to make good use of SIMD, it's all about the flop/byte like you say.
It may be relatively simple to do this and another benefit would be lower discrete adjoint memory footprint (half the flux Jacobians to pre-accumulate).

I never thought about fusing gradients and limiters though, I guess that would require looping by cells instead of by edges. One problem I see is that sometimes we want gradients and limiters, but to compute turbulence source terms we only want gradients (and there we are already a bit inefficient as we compute the gradient of all primitives and just the velocities would do), in the interest of readability/maintenance I do not know if we want to end up two similar routines.

Loop unrolling can go terribly wrong (i.e. code bloat that can increase instruction cache misses) for small sizes if the compiler does not have some solid information about loop size for [example](https://gcc.godbolt.org/z/eM2I13) compilers do very different things with those two loops, the second one is unrolled nicely and icc even partly vectorizes, for the generic loop 90% of the instructions would be cold all the time.

SU2_CFD/src/solver_direct_elasticity.cpp
rsanfer: I've also changed the CFEABoundVariable and split it into 2 classes depending on whether there are Fluid loads or not. We can take a look. I plan to open the PR for the mesh solver hopefully today or at latest on Monday, there we can discuss how we could combine the developments.

SU2_CFD/src/python_wrapper_structure.cpp
talbring: Yep I'll agree actually ... lets rename that to "nodes" then ?

SU2_CFD/include/variables/CTurbSSTVariable.hpp
talbring: Thanks for the explanation! We should use these keywords whenever we can.

SU2_CFD/include/variables/CVariable.hpp
talbring: Can we leave the `unsigned long` ? And maybe we call the others `Array_2D<>`/`Array_1D<>` or something like that and move a typedef (with the most common template args) somewhere accessible by other parts of the code ?

SU2_CFD/include/variables/CVariable.hpp
talbring: I would say we remove whatever is not used at the moment.

pcarruscag: I share that opinion. @economon what do you think?

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: Yup that is fine by me, I had the Eigen-style common typedefs in mind for when I eventually add some linear algebra but I guess  it does not matter much if I add them now.

SU2_CFD/src/python_wrapper_structure.cpp
pcarruscag: It is the least disruptive option, what should I do with the new `snode` I added? `snodes` ? `Snodes` to make clear it is not a typo?

talbring: Or maybe it would be better to use a function called `GetNodes` to access the nodes from the outside. Internally we use then always the `snode` which we rename to be `nodes`. That is maybe less confusing ? The current `node` can then be something descriptive to make clear its not intended to be used internally.

pcarruscag: @talbring if we add a `GetNodes` function the best thing is to keep only a private "`snode`" in each solver which that function upcasts to CVariable, no room for doubt that way.

@economon being the solution to a PDE is what made me think of `fields` I am partial to short names otherwise it will mess with the code formatting too much.
IDK if something as descriptive as eulerVariables is needed, it is going to look weird in the NS solver and the IDE gives you that information already as it is the name of the variable type.

talbring: Then let's go with function and a cast I'd say.

talbring: And what if we call it `nodeArray` or `nodeData` and the class e.g. `CEulerData`?

AlbertoGuardone: I agree with @talbring we need to use GetNodes to access it. As for the naming, fields or DOF  might be too cryptic I think. Data is also confusing (parameter or solution?). What about SolutionNode or SolNode or SolNodeFlow? We can then have SolNodeEul and SolNodeNS or if you prefer SolNodeAdv (for Advection) and SolNodeVis (Viscous) in the pertinent solvers. To be clear, SolNodeVis should contain all variables, including density and velocity.

pcarruscag: We are bikeshedding, so here is what I can reasonably do:
- Rename `node` to `vars` or `nodes` so I don't have to reformat 5k lines of code by hand, both make clear it's multiple things, but `vars` is consistent with the class names, to which I can append an "s" (CVariables and so on).
- Put a pure virtual access method in all solvers to avoid ambiguity between `node` and `snode` within each solver class, the name for that is `GetVars` or `GetNodes`.

Take it or leave it :)

talbring: Once we are already at it, lets do it right. I can also help.

Anything with sol(ution) in it is confusing because it contains more than just that. Still thinking `data` would be the best, as that's what it is. 

Accessing it would be then:
`solver->GetNodeData()->GetSolution(iPoint, iVar)`

SU2_CFD/include/variables/CVariable.hpp
oleburghardt: E.g. all these routines have parallel changes in #774

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2469 +47,2615 @@
...
+    }
+
+    su2double& operator() (Idx_t i, Idx_t j, Idx_t k) { return storage(i*M*N + j*N + k); }
+    const su2double& operator() (Idx_t i, Idx_t j, Idx_t k) const { return storage(i*M*N + j*N + k); }
+
+    su2double** operator[] (Idx_t i) { return interface[i]; }
+  };
+
+  Mat_t Solution;       /*!< \brief Solution of the problem. */
+  Mat_t Solution_Old;   /*!< \brief Old solution of the problem R-K. */
```
oleburghardt: Some data structures have been added (mainly computational graph references/indices) on the variable level in #774 with some subsequent changes for the routines building on top of them. The changes here should go first so that one can apapt them easier.

SU2_CFD/include/variables/CVariable.hpp
oleburghardt: I strongly support @economon 's suggestion!

SU2_CFD/src/solver_direct_heat.cpp
```diff
@@ -235,11 +234,10 @@ CHeatSolverFVM::CHeatSolverFVM(CGeometry *geometry, CConfig *config, unsigned sh
     }
   }
 
-    for (iPoint = 0; iPoint < nPoint; iPoint++)
-      if (flow)
-        node[iPoint] = new CHeatFVMVariable(config->GetTemperature_FreeStreamND(), nDim, nVar, config);
-      else
-        node[iPoint] = new CHeatFVMVariable(Temperature_Solid_Freestream_ND, nDim, nVar, config);
+  if (flow)
+    node = new CHeatFVMVariable(config->GetTemperature_FreeStreamND(), nPoint, nDim, nVar, config);
```
oleburghardt: So all adaptions have been made for the heat solver already :+1:

pcarruscag: For all solvers and features that compile by default and are under regression.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: I prefer if we do it the other way, merge every PR with significant changes to CVariable, solver etc. ahead of this one, I'll fix the conflicts and port what needs porting.
I have some scripts to make my live easier, don't worry.

oleburghardt: Ok, that's important to know!

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: I can create some useful typedefs and change Idx_t to unsigned long (there goes my beautiful formatting...), I can call those typedefs su2vector and su2matrix, I think array1D and array2D are not very good names (relatively long and only one char difference...).

I am keeping the local typedefs in CVariable though (which I can spell out for clarity), for two reasons:
- After scribbling around with SIMD ideias and based on Edwin's suggestions in #716, the storage type will need to change.
- I was told by one of your (Tim's and Ole's) colleagues that to exploit spacial locality with the AD type it may be better to have a "AD vector type" instead of a vector of AD types (more SoA AoS stuff).

pcarruscag:
@rsanfer , I deleted CFEAFSIBoundVariable and kept its data in CFEABoundVariable because:
- It would require two vertex maps to be build, one for vertices another for interfaces.
- It would be less efficient memory wise, the vertex map is size O(N) the number of vertices is O(N^(2/3)) in the worst case, which means eventually having two maps is worse.
- Most FSI cases have a large interface and small fixed boundaries, allocating FSI tractions for the latter is not so bad.

I also removed CDiscAdjMeshVariable, as it was empty, and kept only its "Bound" counterpart. With this vertex allocation / mapping strategy we always need to instantiate the "Bound" version of the class so if the intermediates are empty they can be omitted.

Please review the changes and if you have some more thorough tests run them or send them over to me.

SU2_CFD/src/integration_structure.cpp
```diff
@@ -703,27 +695,23 @@ void CIntegration::SetFEM_StructuralSolver(CGeometry *geometry, CSolver **solver
...
-  for (iPoint = 0; iPoint < geometry->GetnPoint(); iPoint++) {
-    
-    solver_container[FEA_SOL]->node[iPoint]->Set_Solution_time_n();
-    solver_container[FEA_SOL]->node[iPoint]->SetSolution_Vel_time_n();
-    solver_container[FEA_SOL]->node[iPoint]->SetSolution_Accel_time_n();
-    
-  }
-  
+
+  solver_container[FEA_SOL]->node->Set_Solution_time_n();
```
pcarruscag: @rsanfer I saw the name of this method `Set_Solution_time_n` changed slightly can you check I didn't break anything.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2608 +47,2816 @@
...
+    }
+
+    su2double& operator() (Idx_t i, Idx_t j, Idx_t k) { return storage(i*M*N + j*N + k); }
+    const su2double& operator() (Idx_t i, Idx_t j, Idx_t k) const { return storage(i*M*N + j*N + k); }
+
+    su2double** operator[] (Idx_t i) { return interface[i]; }
+  };
+
+  /*--- Auxilary class for derived classes that only allocate data for boundary points. ---*/
+  class CVertexMap {
```
pcarruscag: I'll move this to its own file.

SU2_CFD/src/solvers/CDiscAdjMeshSolver.cpp
```diff
@@ -180,9 +171,8 @@ void CDiscAdjMeshSolver::RegisterVariables(CGeometry *geometry, CConfig *config,
 
   /*--- Register reference mesh coordinates ---*/
 
-  for (iPoint = 0; iPoint < nPoint; iPoint++){
-    direct_solver->node[iPoint]->Register_BoundDisp(input);
-  }
+  for (iPoint = 0; iPoint < nPoint; iPoint++)
+    direct_solver->node->Register_BoundDisp(iPoint,input);
```
rsanfer: The other failing case disc_adj_flow/mesh_disp_sens relies on the boundary displacement being correctly recorded. I wonder what's the reason for it to fail...

SU2_CFD/src/transfer_physics.cpp
```diff
@@ -83,7 +83,7 @@ void CTransfer_FlowTraction::GetPhysical_Constants(CSolver *flow_solution, CSolv
   /*--- We have to clear the traction before applying it, because we are "adding" to node and not "setting" ---*/
 
   for (unsigned long iPoint = 0; iPoint < struct_geometry->GetnPoint(); iPoint++) 
-    struct_solution->node[iPoint]->Clear_FlowTraction();
+    struct_solution->node->Clear_FlowTraction(iPoint);
```
rsanfer: There are not so many changes here, so I will work on the names of the split files in #769 so we can merge it in soonish, and I then can help updating this.

SU2_CFD/src/variables/CFEAVariable.cpp
```diff
@@ -37,141 +37,68 @@
...
-  Stress                = NULL;    // Nodal stress (for output purposes)
-  Residual_Ext_Body     = NULL;    // Residual component due to body forces
+  bool nonlinear_analysis = (config->GetGeometricConditions() == LARGE_DEFORMATIONS);
+  bool body_forces        = config->GetDeadLoad();
+  bool incremental_load   = config->GetIncrementalLoad();
+  bool prestretch_fem     = config->GetPrestretch();  // Structure is prestretched
+  bool discrete_adjoint   = config->GetDiscrete_Adjoint();
+  bool refgeom            = config->GetRefGeom(); // Reference geometry needs to be stored
+  bool dynamic_analysis   = (config->GetDynamic_Analysis() == DYNAMIC);
+  bool fsi_analysis       = config->GetFSI_Simulation();
```
rsanfer: Here the fsi check should also be updated.

SU2_CFD/src/integration_structure.cpp
pcarruscag: So remove the underscore from the time_n and time_n1 methods?

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -3100,9 +3080,9 @@ void CFEASolver::ImplicitNewmark_Iteration(CGeometry *geometry, CSolver **solver
   bool linear_analysis = (config->GetGeometricConditions() == SMALL_DEFORMATIONS);  // Linear analysis.
   bool nonlinear_analysis = (config->GetGeometricConditions() == LARGE_DEFORMATIONS);  // Nonlinear analysis.
   bool newton_raphson = (config->GetKind_SpaceIteScheme_FEA() == NEWTON_RAPHSON);    // Newton-Raphson method
-  
+  bool fsi = config->GetFSI_Simulation();
```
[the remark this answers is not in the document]
pcarruscag: Thanks for the hint on the testcase, I will move that logic inside the class, and remove the explicit fsi checking.

SU2_CFD/include/variables/CDiscAdjMeshVariable.hpp
```diff
@@ -1,67 +0,0 @@
-/*!
```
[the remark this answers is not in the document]
pcarruscag: Ups, I was not aware of that, yeah it was just to simplify since it was empty...

SU2_CFD/include/variables/CDiscAdjVariable.hpp
```diff
@@ -45,206 +45,214 @@
...
+  inline su2double Get_BGSSolution_k(Idx_t iPoint, Idx_t iDim) const override { return Solution_BGS_k(iPoint,iDim);}
 
-  /*!
-   * \brief Set the value of the adjoint geometry solution in the previous BGS subiteration.
-   */
-  inline void Set_BGSSolution_Geometry(void) {
-    for (unsigned short iDim = 0; iDim < nDim; iDim++)
-      Solution_Geometry_BGS_k[iDim] = Solution_Geometry[iDim];
-  }
+//  /*!
```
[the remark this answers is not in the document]
pcarruscag: Part of the cleanup of unused methods that I mentioned above, It is also commented out in CVariable so at present that method (and others) is not used.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: Please don't xD that would make my optimizations more intricate as I would need separate meshes for the direct problem and then merge them for the adjoint... Let's do it all at once, one PR -20k lines it'll be more impressive that way.

SU2_CFD/include/variables/CFEAVariable.hpp
```diff
@@ -49,461 +49,487 @@
...
    * \brief Get the body forces.
    */
-  inline su2double Get_BodyForces_Res(unsigned short iVar) {return Residual_Ext_Body[iVar];}
+  inline su2double Get_BodyForces_Res(Idx_t iPoint, Idx_t iVar) const final { return Residual_Ext_Body(iPoint,iVar); }
+
+  /*!
+   * \brief Set the value of the old solution.
+   * \param[in] val_solution_time_n - Pointer to the residual vector.
+   */
+  inline void SetSolution_time_n(Idx_t iPoint, const su2double *val_solution_time_n) final {
```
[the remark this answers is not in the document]
pcarruscag: Can you go over those? If possible we should keep the non virtual ones, I think at the moment only CDiscAdjFEAVariable overrides the virtual version of the method, and the implementation is the same as base.

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -0,0 +1,505 @@
...
+  {
+      for(size_t i=0; i<size(); ++i) m_data[i] = value;
+  }
+};
+
+
+/*!
+ * \brief Useful typedefs with default template parameters
+ */
+template<class T> using su2vector = C2DContainer<unsigned long, T, StorageType::ColumnMajor, 64, DynamicSize, 1>;
```
pcarruscag: Useful typedefs here, two templated on type, and active/passive versions.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2298 +47,2327 @@
...
+
 /*!
  * \class CVariable
  * \brief Main class for defining the variables.
  * \author F. Palacios
  */
 class CVariable {
 protected:
+  using Idx_t = unsigned long;
+  using VectorType = C2DContainer<Idx_t, su2double, StorageType::ColumnMajor, 64, DynamicSize, 1>;
```
pcarruscag: As requested I spelled out these names, I think they are uncontroversial now.
I'll do a replace all on Idx_t and the end, until then this helps me find new methods added to CVariable in other PR's.

SU2_CFD/include/variables/CVariable.hpp
AlbertoGuardone: I am fine with removing secondaries, they are not needed within the NICFD functionality.

pcarruscag:
I want to be done with this PR folks... This is really tiring to maintain so if you don't want it please just tell me...
Can we agree on:
Instead of CSolver having the `node` field it will have a pure virtual function `CVariable* GetNodes()`, derived solvers need to implement this method and have a `nodes` field of the most appropriate type (e.g. CEulerVariable for CEulerSolver).

I know some of you do not like the name `nodes` but there is something to be said about consistency (that has always been the name) and I do not think changing a generic name for another generic name justifies breaking the formatting everywhere and having needlessly long calls to get some data.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -1717,10 +1714,10 @@ void CEulerSolver::Set_MPI_ActDisk(CSolver **solver_container, CGeometry *geomet
...
             
             for (iVar = 0; iVar < nPrimVar; iVar++) {
-              Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+iVar] = node[iPoint]->GetPrimitive(iVar);
+              Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+iVar] = snode->GetPrimitive(iPoint,iVar);
             }
             if (rans) {
-              Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+nPrimVar] = solver_container[TURB_SOL]->node[iPoint]->GetSolution(0);
+              Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+nPrimVar] = solver_container[TURB_SOL]->node->GetSolution(iPoint,0);
               Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+(nPrimVar+1)] = 0.0;
             }
```
jayantmukho: Just so I understand how this node - snode relationship works:

snode is used only by the solver that it is instantiated in. It points to the node variable in that class, but is statically cast to be specific to the solver? And this is done to remove the need for virtual functions?

The fact that the node and snode pointers are the same means that if accessing the node of a different class, you can still just use the node variable?

Just trying to wrap my head around this

SU2_CFD/src/solver_direct_mean.cpp
jayantmukho: oh, i guess that is the answer to the question I asked above. nvm.

SU2_CFD/src/solver_adjoint_mean.cpp
```diff
@@ -6581,16 +6572,16 @@ void CAdjNSSolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_cont
           force_stress += Normal[iDim]*Tau[iDim][jDim]*d[jDim];
 
       /*--- \partial \mu_dyn \partial T ---*/
-      //        mu_dyn = solver_container[FLOW_SOL]->node[iPoint]->GetLaminarViscosity();
-      //        Temp = solver_container[FLOW_SOL]->node[iPoint]->GetTemperature();
+      //        mu_dyn = solver_container[FLOW_SOL]->node->GetLaminarViscosity(iPoint);
```
jayantmukho: ded?

SU2_CFD/src/solver_adjoint_turbulent.cpp
```diff
@@ -515,13 +503,13 @@ void CAdjTurbSolver::Source_Residual(CGeometry *geometry, CSolver **solver_conta
 //    jPoint = geometry->edge[iEdge]->GetNode(1);
```
jayantmukho: ded?

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -1633,16 +1613,16 @@ void CFEASolver::Compute_NodalStress(CGeometry *geometry, CNumerics **numerics,
       indexNode[iNode] = geometry->elem[iElem]->GetNode(iNode);
       //      for (iDim = 0; iDim < nDim; iDim++) {
       //        val_Coord = geometry->node[indexNode[iNode]]->GetCoord(iDim);
-      //        val_Sol = node[indexNode[iNode]]->GetSolution(iDim) + val_Coord;
+      //        val_Sol = node->GetSolution(indexNode[iNode],iDim) + val_Coord;
       //        element_container[FEA_TERM][EL_KIND]->SetRef_Coord(val_Coord, iNode, iDim);
       //        element_container[FEA_TERM][EL_KIND]->SetCurr_Coord(val_Sol, iNode, iDim);
       //      }
```
jayantmukho: more ded?

SU2_CFD/src/variables/CDiscAdjVariable.cpp
```diff
@@ -37,129 +37,44 @@
...
-  Geometry_CrossTerm_Derivative_Flow = NULL;
-
-}
-
-CDiscAdjVariable::CDiscAdjVariable(su2double* val_solution, unsigned short val_ndim, unsigned short val_nvar,
-                                   CConfig *config) : CVariable(val_ndim, val_nvar, config) {
-
-  bool dual_time = (config->GetTime_Marching() == DT_STEPPING_1ST)
-      || (config->GetTime_Marching() == DT_STEPPING_2ND);
+CDiscAdjVariable::CDiscAdjVariable(const su2double* sol, Idx_t npoint, Idx_t ndim, Idx_t nvar, CConfig *config)
```
jayantmukho: so cleeeaaannn

Common/src/config_structure.cpp
```diff
@@ -2903,12 +2903,12 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
     SU2_MPI::Error("A turbulence model must be specified with KIND_TURB_MODEL if SOLVER= INC_RANS", CURRENT_FUNCTION);
   }
 
-#ifndef HAVE_TECIO
-  if (Output_FileFormat == TECPLOT_BINARY) {
-    cout << "Tecplot binary file requested but SU2 was built without TecIO support." << "\n";
-    Output_FileFormat = TECPLOT;
-  }
-#endif
+//#ifndef HAVE_TECIO
```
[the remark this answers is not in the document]
pcarruscag: Thank you for the review @jayantmukho.
After I merged the new output this was breaking compilation without tecio (which is how I usually build the code) Output_FileFormat has been commented out elsewhere so it is no longer defined... I had to do the same

SU2_CFD/include/variables/CAdjEulerVariable.hpp
```diff
@@ -47,104 +47,101 @@
...
    * \brief Destructor of the class.
    */
-  virtual ~CAdjEulerVariable(void);
+  virtual ~CAdjEulerVariable() = default;
 
   /*!
    * \brief Set all the primitive variables for compressible flows.
    */
-  bool SetPrimVar(su2double SharpEdge_Distance, bool check, CConfig *config);
+  bool SetPrimVar(Idx_t iPoint, su2double SharpEdge_Distance, bool check, CConfig *config) final;
```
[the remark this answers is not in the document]
pcarruscag: Most of them actually... but they are so many xD

SU2_CFD/include/variables/CDiscAdjFEAVariable.hpp
```diff
@@ -39,239 +39,255 @@
...
-  CDiscAdjFEAVariable(su2double *val_solution, su2double *val_solution_accel, su2double *val_solution_vel, unsigned short val_ndim, unsigned short val_nvar, CConfig *config);
+  ~CDiscAdjFEAVariable() = default;
 
   /*!
    * \brief Set the sensitivity at the node
    * \param[in] iDim - spacial component
    * \param[in] val - value of the Sensitivity
    */
-  inline void SetSensitivity(unsigned short iDim, su2double val) {Sensitivity[iDim] = val;}
+  inline void SetSensitivity(Idx_t iPoint, Idx_t iDim, su2double val) final { Sensitivity(iPoint,iDim) = val; }
```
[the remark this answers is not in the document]
pcarruscag: Actually this was done in the PR where I split the variables over different files, at some point there was an agreement we would stop having .inl files.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -110,21 +110,21 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
 
     dTdn = (Twall - Tnormal)/dist;
   }
   else if (flow || heat_equation) {
-    Twall   = donor_solution->node[Point_Donor]->GetSolution(0)*Temperature_Ref;
-    Tnormal = donor_solution->node[PointNormal]->GetSolution(0)*Temperature_Ref;
+    Twall   = donor_solution->node->GetSolution(Point_Donor,0)*Temperature_Ref;
+    Tnormal = donor_solution->node->GetSolution(PointNormal,0)*Temperature_Ref;
 
     //    for (iDim = 0; iDim < nDim; iDim++) {
```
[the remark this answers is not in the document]
pcarruscag: I don't know... as you can imagine a lot of the refactoring was done programatically, that is why I mentioned on another PR we should not leave commented code laying around without justification

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2660,21 +2662,15 @@ void CDiscAdjFEAIteration::Preprocess(COutput *output,
 
     /*--- Push solution back to correct array ---*/
 
-    for(iPoint=0; iPoint<geometry[val_iZone][val_iInst][MESH_0]->GetnPoint();iPoint++){
-      solver[val_iZone][val_iInst][MESH_0][FEA_SOL]->node[iPoint]->Set_Solution_time_n();
-    }
+    solver[val_iZone][val_iInst][MESH_0][FEA_SOL]->node->Set_Solution_time_n();
```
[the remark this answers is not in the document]
pcarruscag: Thanks!

SU2_CFD/src/solver_direct_mean.cpp
pcarruscag: You can access node but not snode from the outside.
The main purpose was not to remove the need for virtual functions, it was to allow `final` virtual functions to be inlined, without derived type information the compiler needs to insert all the dynamic dispatch mechanisms which are relatively expensive when all you are doing is getting a double. Close to the bottom of #716 this is explained in more detail.
As an added bonus you can indeed add data members to a derived CVariable and not to the base (i.e. no virtual) provided you only access through snode.
As clever as the strategy may look it is not ideal design, ideally derived solvers would have a node of the right type for internal use, and for access from the outside they would cast to base class and return a CVariable pointer (less confusing than having two variables that allow you to do the "same").

SU2_CFD/src/variables/CDiscAdjVariable.cpp
pcarruscag: Thanks, I was also surprised when I started cleaning up constructors at how much could be removed, and no risk of memory leaks now...

SU2_CFD/include/variables/CDiscAdjFEAVariable.hpp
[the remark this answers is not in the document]
jayantmukho: Whoops, missed that memo. Will keep that in mind going forward.

SU2_CFD/src/solver_direct_mean.cpp
jayantmukho: Makes sense. Would that change be in this PR or after? And that's what is referred to in your comment:

> I want to be done with this PR folks... This is really tiring to maintain so if you don't want it please just tell me...
> Can we agree on:
> Instead of CSolver having the `node` field it will have a pure virtual function `CVariable* GetNodes()`, derived solvers need to implement this method and have a `nodes` field of the most appropriate type (e.g. CEulerVariable for CEulerSolver).
> 
> I know some of you do not like the name `nodes` but there is something to be said about consistency (that has always been the name) and I do not think changing a generic name for another generic name justifies breaking the formatting everywhere and having needlessly long calls to get some data.

For what its worth, I think I agree that the name `nodes` is the best bet.

.travis.yml
```diff
@@ -72,7 +72,7 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone --depth=1 -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone --depth=1 -b feature_contiguous_cvariable https://github.com/su2code/TestCases.git ./TestData
```
economon: Don't forget to switch this back to develop

Common/src/config_structure.cpp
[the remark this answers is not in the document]
economon: I have put in a small fix for this particular instance of this issue in #790

SU2_CFD/include/variables/CDiscAdjFEABoundVariable.hpp
```diff
@@ -46,69 +47,93 @@
...
+  MatrixType SourceTerm_DispAdjoint;   /*!< \brief Source term applied into the displacement
+                                                   adjoint coming from external solvers. */
 
-public:
-
-  /*!
-   * \brief Constructor of the class.
-   */
-  CDiscAdjFEABoundVariable(void);
+  CVertexMap<unsigned> VertexMap;  /*!< \brief Object that controls accesses to the variables of this class. */
```
economon: Is it possible we can use this mapping object in other places? For example, isolating all boundary nodes quickly when preparing output.

SU2_CFD/include/variables/CDiscAdjVariable.hpp
```diff
@@ -45,192 +45,186 @@
...
-   * \param[out] val_solution - geometrical adjoint solution in the previous BGS subiteration.
-   */
-  inline su2double Get_BGSSolution_Geometry(unsigned short iDim) {return Solution_Geometry_BGS_k[iDim];}
+  inline su2double Get_BGSSolution(Idx_t iPoint, Idx_t iDim) const override { return Solution_BGS(iPoint,iDim);}
 
   /*!
    * \brief Set the contribution of crossed terms into the derivative.
    */
-  inline void SetCross_Term_Derivative(unsigned short iVar, su2double der) {Cross_Term_Derivative[iVar] = der; }
+  inline void SetCross_Term_Derivative(Idx_t iPoint, Idx_t iVar, su2double der) override {
```
economon: I am not sure about other systems, but on mac/llvm, the override keyword is resulting in many warnings due to inconsistent/missing override statements on some functions... do you see these when you build this up?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2318 +47,2340 @@
...
 
   /*!
    * \brief A virtual member.
+   * \param[in] iPoint - Point index.
    * \param[in] val_vector - Direction of projection.
    * \param[in] val_species - Index of the desired species.
    * \return Value of the projected velocity.
    */
-  inline virtual su2double GetProjVel(su2double *val_vector, unsigned short val_species) {return 0; }
+  inline virtual su2double GetProjVel(Idx_t iPoint, su2double *val_vector, Idx_t val_
```
economon: Lots of 'val_*' have been removed from the input variable names. Traditionally, we used that to help identify inputs vs local data. But, fine for me if we all agree to that policy everywhere and start being consistent.

SU2_CFD/src/output/CFlowCompFEMOutput.cpp
```diff
@@ -285,35 +285,6 @@ void CFlowCompFEMOutput::LoadHistoryData(CConfig *config, CGeometry *geometry, C
...
 
-su2double CFlowCompFEMOutput::GetQ_Criterion(CConfig *config, CGeometry *geometry, CVariable* node_flow){
-  
-  unsigned short iDim, jDim;
-  su2double Grad_Vel[3][3] = {{0.0, 0.0, 0.0},{0.0, 0.0, 0.0},{0.0, 0.0, 0.0}};
-  su2double Omega[3][3]    = {{0.0, 0.0, 0.0},{0.0, 0.0, 0.0},{0.0, 0.0, 0.0}};
-  su2double Strain[3][3]   = {{0.0, 0.0, 0.0},{0.0, 0.0, 0.0},{0.0, 0.0, 0.0}};
-  for (iDim = 0; iDim < nDim; iDim++) {
-    for (unsigned short jDim = 0 ; jDim < nDim; jDim++) {
-      Grad_Vel[iDim][jDim] = node_flow->GetGradient_Primitive(iDim+1, jDim);
```
economon: Thanks for moving this up to the parent.. we don't have the CVariable class inside the DG solver, so we needed to adjust it anyway

economon:
> I want to be done with this PR folks... This is really tiring to maintain so if you don't want it please just tell me...
> Can we agree on:
> Instead of CSolver having the `node` field it will have a pure virtual function `CVariable* GetNodes()`, derived solvers need to implement this method and have a `nodes` field of the most appropriate type (e.g. CEulerVariable for CEulerSolver).
> 
> I know some of you do not like the name `nodes` but there is something to be said about consistency (that has always been the name) and I do not think changing a generic name for another generic name justifies breaking the formatting everywhere and having needlessly long calls to get some data.

In the interest of keeping the PRs moving, I am ok with this. It will also be natural for folks in the short term to learn the new system, since everyone is already accustomed to using the solver->node* construct. We can always go for a targeted refactoring later separate from the changes in this PR if we would like.

SU2_CFD/include/variables/CDiscAdjFEABoundVariable.hpp
pcarruscag: Hummm probably not as the mapping is in the opposite direction ( ivertex = f(ipoint) )... this is more of a workaround for the structural solvers not having "vertices"

SU2_CFD/include/variables/CDiscAdjVariable.hpp
pcarruscag: Nothing on gcc... I'll try to replicate, if I can I'll fix it.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: I tried to do that for all iPoint, iVar, iDim indices I don't recall changing it on many "value" arguments. I definitely did not make it less consistent :)

SU2_CFD/src/output/CFlowCompFEMOutput.cpp
pcarruscag: No prob, I think I kept all the comments/sources you had in your original PR

SU2_CFD/include/variables/CDiscAdjVariable.hpp
pcarruscag: With clang 3.8 I get no warning from CVariable files with -Wall -Wextra -Wno-unused-parameter.
I do get override warnings from other files due to the absence of the keyword (-Winconsistent-missing-override) clang is probably turning on this best-practices-based warning due to the c++11 flag.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: By the way, what is this "species" thing? Do you know if these overloads are still used anywhere? I never noticed them anywhere else in the code.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -156,8 +166,29 @@ class CSolver {
...
+  inline void SetBaseClassPointerToNodes() { base_nodes = GetBaseClassPointerToNodes(); }
+
+private:
+
+  /*--- Private to prevent use by derived solvers, each solver MUST have its own "nodes" member of the
+   most derived type possible, e.g. CEulerVariable has nodes of CEulerVariable* and not CVariable*.
+   This variable is to avoid two virtual functions calls per call i.e. CSolver::GetNodes() returns
+   directly instead of calling GetBaseClassPointerToNodes() or doing something equivalent. ---*/
+  CVariable* base_nodes;  /*!< \brief Pointer to CVariable to allow polymorphic access to solver nodes. */
+
```
pcarruscag: Ok, three votes in favour, so it is done.
It is all explained here in the comments, this was the cleanest way I could think of avoiding confusion by having two variables accessible in derived solvers that could be used seemingly interchangeably.
If a new CSolver is added people will invariably end up reading these comments about how to setup the system and runtime checks are performed for debug builds inside the GetNodes() method.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -11033,290 +11116,6 @@ class CAdjTurbSolver : public CSolver {
   
 };
 
-/*! \class CPoissonSolver
```
pcarruscag: I think these solvers were removed a while ago, I removed their declaration from the header.

SU2_CFD/include/variables/CVariable.hpp
economon: The species-related calls are left over from the old TNE2 solver (non-equilibrium flows). I think they should be kept for now, as the old TNE2 solver is currently being resurrected by @WallyMaier, @CatarinaGarbacz, @MarcoFossati and they may be needed. That can be cleaned once the new version of that solver is integrated with develop.

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -47,268 +47,281 @@
...
+    if (Primitive(iPoint,nDim+1) > 0.0) return false;
     else return true;
   }
 
   /*!
    * \brief Set the value of the beta coeffient for incompressible flows.
+   * \param[in] iPoint - Point index.
    */
-  inline void SetBetaInc2(su2double val_betainc2) {Primitive[nDim+3] = val_betainc2; }
+  inline void SetBetaInc2(Idx_t iPoint, su2double val_betainc2) final { Primitive(iPoint,nDim+3) = val_betainc2; }
```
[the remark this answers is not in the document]
economon: Yep needs to be looked at closely to find an option with good performance, but as it currently exists, it is very error-prone. Just food for thought

SU2_CFD/include/variables/CDiscAdjVariable.hpp
economon: Guess we have some cleanup to do then. 

Also, still plan to search and replace 'Idx_t'?

pcarruscag: Yup very last thing before merging as it makes it easier to spot new functions from other PR's
How do you propose handling #777 and #790?

pcarruscag:
@oleburghardt I merged #774 into this already.

SU2_CFD/include/variables/CDiscAdjVariable.hpp
economon: #790 is more or less complete in terms of content. Still need to update a couple of items in the output. #777 will need a bit of work now that #724 is merged and after this PR, but that is ok. In fact, I am fine if we do not rush just to get that into the release.

So, I would say we go ahead and merge this and #774 very soon. Then I can update #790 (only the new memory for the reconstruction gradient is affected), and we can get that in afterwards. I'll keep working on #777, but let's take it out of the critical path for the release (it is all new functionality anyway).

pcarruscag: Ok, sounds good, regardless of merging order I can help. I still want to try #790 on the benchmark case I used for this PR, from your results so far we'll gain more from that than from the memory layout.

MicK7:
@pcarruscag do you think this topic can be merged as it is or is an additional effort needed ? Merging early may help with new PR

pcarruscag:
@MicK7 apart from reverting travis and substituting the Idx_t typedef for unsigned long I think I addressed everything, #774 needs to go in first though as I merged it already.
@economon also said he had more comments coming.
A couple more approvals would also be good given the number of people commenting.

economon:
I'm good - I submitted the second short set of comments just a few hours after the first (6 days ago). Just the Idx_t item and travis, and I gladly approve.

Additional reviews from others always encouraged.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -58,2355 +59,2373 @@ using namespace std;
...
+  inline virtual void AddSolution_New(unsigned long iPoint, unsigned long iVar, su2double solution) {}
 
   /*!
-   * \brief A virtual member.
+   * \brief Add a value to the External vector.
+   * \param[in] iPoint - Point index.
+   * \param[in] val_sol - vector that has to be added component-wise
    */
-  inline virtual void Add_ExternalOld(const su2double* val_sol) {}
+  inline void Add_External(unsigned long iPoint, const su2double* val_sol) {
```
pcarruscag: I moved these method here as for now they do not need to be virtual

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -37,185 +37,134 @@
...
+
+void CVariable::SetAuxVarGradientZero() { Grad_AuxVar.setConstant(0.0); }
+
+void CVariable::SetGradientZero() { Gradient.storage.setConstant(0.0); }
+
+void CVariable::SetRmatrixZero() { Rmatrix.storage.setConstant(0.0); }
+
+void CVariable::SetUnd_LaplZero() { Undivided_Laplacian.setConstant(0.0); }
+
+void CVariable::SetExternalZero() { External.setConstant(0.0); }
```
pcarruscag: These simple methods operating on all points were moved from CSolver to here

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -37,185 +37,134 @@
...
+  for (unsigned long iPoint = 0; iPoint < nPoint; ++iPoint)
+    for(unsigned long iVar=0; iVar<nVar; ++iVar)
+      AD::RegisterInput(Solution_time_n1(iPoint,iVar));
+}
 
-  if (Rmatrix != NULL) {
-    for (iDim = 0; iDim < nDim; iDim++)
-      delete [] Rmatrix[iDim];
-    delete [] Rmatrix;
+void CVariable::RegisterSolution_intIndexBased(bool input) {
```
pcarruscag: And so were these ones for the new index-based AD

MicK7:
@economon @pcarruscag can we change the  `vector<vector<passivedouble> > localPointCoordinates` and  `vector<vector<unsigned long> > surfaceElementConnectivity` of CMeshReaderFVM class to use the new structure of this branch ?

economon:
> @economon @pcarruscag can we change the `vector<vector<passivedouble> > localPointCoordinates` and `vector<vector<unsigned long> > surfaceElementConnectivity` of CMeshReaderFVM class to use the new structure of this branch ?

@MicK7 : I think we can separate that from this PR. The memory used in the mesh readers is just temporary for the preprocessing.. I think we can have more speed improvement by switching our heavily used data-structures within CPhysicalGeometry to the new containers. But, this might be some effort that is worth a new PR. Just my thoughts

pcarruscag:
@MicK7 I did have a look as I remembered asking you to leave a "ToDo" in the code, I did not change anything because this container class does not implement push_back (I don't think it should actually since for 2D that would be strange).
I agree with @economon regarding CPhysicalGeometry, and in fact this is already an item in the work plan for #789.

SU2_CFD/include/variables/CDiscAdjFEABoundVariable.hpp
```diff
@@ -46,69 +47,93 @@
...
+   */
+  void AllocateBoundaryVariables(CConfig *config);
 
   /*!
    * \brief Set the FSI force sensitivity at the node
    * \param[in] iDim - spacial component
    * \param[in] val - value of the Sensitivity
    */
-  inline void SetFlowTractionSensitivity(unsigned short iDim, su2double val) { FlowTraction_Sens[iDim] = val; }
+  inline void SetFlowTractionSensitivity(unsigned long iPoint, unsigned long iDim, su2double val) override {
```
talbring: Is there any particular reason why all of the arguments that were previously `unsigned short` are now `unsigned long` ?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -58,2355 +59,2373 @@ using namespace std;
...
 class CVariable {
 protected:
+  using VectorType = C2DContainer<unsigned long, su2double, StorageType::ColumnMajor, 64, DynamicSize, 1>;
+  using MatrixType = C2DContainer<unsigned long, su2double, StorageType::RowMajor,    64, DynamicSize, DynamicSize>;
+
+  /*--- This contrived container is used to store matrices in a contiguous manner but still present the
+   "su2double**" interface to the outside world, it will be replaced by something more efficient. ---*/
+  struct VectorOfMatrix {
+    su2activevector storage;
+    su2matrix<su2double*> interface;
```
talbring: Whats the difference between `MatrixType` and `su2matrix` ?

SU2_CFD/include/variables/CDiscAdjFEABoundVariable.hpp
pcarruscag: unsigned long is the index type of the container for both row and column indexes, since those indexes are used to compute the linear index on each access, I did not think it was prudent to have different types. I thought it was better to make that consistent across the interface.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -58,2355 +59,2373 @@ using namespace std;
...
+    const su2double& operator() (unsigned long i, unsigned long j, unsigned long k) const { return storage(i*M*N + j*N + k); }
+
+    su2double** operator[] (unsigned long i) { return interface[i]; }
+  };
+
+  MatrixType Solution;       /*!< \brief Solution of the problem. */
+  MatrixType Solution_Old;   /*!< \brief Old solution of the problem R-K. */
+
+  MatrixType External;       /*!< \brief External (outer) contribution in discrete adjoint multizone problems. */
+  MatrixType External_Old;   /*!< \brief Old external (outer) contribution in discrete adjoint multizone problems. */
```
oleburghardt: This whole section looks very tidy and clean. Nice

pcarruscag: Thanks, sadly I broke the cht adjoint... residuals during recording are the same but I get NaN's on the first iteration, I may have to redo everything as I have no idea what to look for. Any clues?

oleburghardt: I'll have a look right now

oleburghardt: So I ran a single zone case (fluid) with the multizone discrete adjoint driver and all looks good. Which tells us that it's unlikely that something is wrong with your adaptions of e.g. `CVariable::SetAdjIndices`.
Just removed the routine that throws that error to see that extracting adjoints in the heat zones seem to fail. I'll keep testing different settings..

pcarruscag: Thank you for chasing this! I pushed that fix, if it passes I think we should merge as travis is currently failing in develop (#801) because I updated one testcase.
I fixed 3 uninitialized warnings, maybe those are the ones you are referring to?

SU2_CFD/src/variables/CDiscAdjVariable.cpp
```diff
@@ -37,143 +37,49 @@
...
-  }
-  
   if (config->GetMultizone_Problem()) {
-
-    Solution_BGS            = new su2double[nVar];
-    for (iVar = 0; iVar < nVar; iVar++) {
-      Solution_BGS[iVar]          = 0.0;
-    }
-
+    Solution_BGS.resize(nPoint,nVar) = su2double(0.0);
```
oleburghardt: > I fixed 3 uninitialized warnings, maybe those are the ones you are referring to?

Yes, and this one has been the only critical one ;)

pcarruscag: I also needed Solution_Old

SU2_CFD/src/integration_structure.cpp
```diff
@@ -311,8 +311,9 @@ void CIntegration::Adjoint_Setup(CGeometry ****geometry, CSolver *****solver_con
       if (iMGLevel != config[iZone]->GetnMGLevels()) {
         SetRestricted_Solution(RUNTIME_FLOW_SYS, solver_container[iZone][INST_0][iMGLevel][FLOW_SOL], solver_container[iZone][INST_0][iMGLevel+1][FLOW_SOL],
                                geometry[iZone][INST_0][iMGLevel], geometry[iZone][INST_0][iMGLevel+1], config[iZone]);
-        SetRestricted_Gradient(RUNTIME_FLOW_SYS, solver_container[iZone][INST_0][iMGLevel][FLOW_SOL], solver_container[iZone][INST_0][iMGLevel+1][FLOW_SOL],
-                               geometry[iZone][INST_0][iMGLevel], geometry[iZone][INST_0][iMGLevel+1], config[iZone]);
+//        ToDo: The flow solvers do not use the conservative variable gradients
```
pcarruscag: here

SU2_CFD/src/solver_direct_transition.cpp
```diff
@@ -335,8 +335,9 @@ void CTransLMSolver::Viscous_Residual(CGeometry *geometry, CSolver **solver_cont
     numerics->SetTransVar(nodes->GetSolution(iPoint), nodes->GetSolution(jPoint));
     numerics->SetTransVarGradient(nodes->GetGradient(iPoint), nodes->GetGradient(jPoint));
     
-    numerics->SetConsVarGradient(solver_container[FLOW_SOL]->GetNodes()->GetGradient(iPoint),
-                                 solver_container[FLOW_SOL]->GetNodes()->GetGradient(jPoint));
+//    ToDo: The flow solvers do not use the conservative variable gradients
```
pcarruscag: here

SU2_CFD/src/variables/CDiscAdjVariable.cpp
oleburghardt: Ok.. it should have had an influence on the residual computation only (it's fixed in the various fixes branch), as it's not used for the solution algorithm. Anyway, might suffice for a crash :p

SU2_CFD/src/integration_structure.cpp
economon: Looks like this is outdated. If anything, it should be a restriction of the primitive gradient to make sure everything is well set in multigrid before the adjoint iteration.

I think it is worth trying to remove and see if the continuous adjoint cases fail in the regressions. If so, we can look into it further, since it might need to be adjusted.

SU2_CFD/src/solver_direct_transition.cpp
economon: This one is probably a good question for @vdweide, since he has been updating the transition model in a separate branch. It looks like the viscous numerics is mostly commented out in the develop branch. I expect we should modify this to use the primitive gradients.

vdweide: No need to worry about the transition model at the moment. Since the data structures have changed so much lately, it is pretty much impossible to carry out a merge between the transition model branch and the develop branch. The implementation must be redone. @economon, we agreed that we would wait with that until PR 777 is merged in. I assume this is still the case.

SU2_CFD/src/integration_structure.cpp
pcarruscag: On the coarse grids the primitive gradients are never computed, they are only used for viscous residuals but they are zero at that point.
The continuous adjoint cases all pass without calling `SetRestricted_Gradient`

oleburghardt:
@pcarruscag Any other changes you want to add here? Otherwise I'll merge it.

pcarruscag:
Let's merge it then so the FSI case starts passing, @economon let me know if you want me to remove any of those unused areas and I will push that to #798.

SU2_CFD/src/solver_direct_transition.cpp
economon: Sounds good - thanks for confirming @vdweide. Agreed we'll work on this after #777

economon:
@pcarruscag : I think we are good to remove those two old gradient sections in #798. I will be working to merge this PR into #790 (and finish up there) next

economon:
Now that I am updating #790 to work with the new structure from this PR (#753), I am seeing that the aligned_alloc() function is not well defined on mac, so xcode/llvm builds are not working on mac at the moment. There are some issue related to C vs C++ with that (depends on the standard you choose for C or C++, 11 or 17, apparently). Might be some ways around it from reading comments by others on the issue: https://github.com/marian-nmt/marian-dev/issues/227.

@pcarruscag : can you give this a quick look when you have some time? Thanks!

PR #756, opened 2019-08-09
Renaming PHYSICAL_PROBLEM to SOLVER

talbring:
## Proposed Changes
While I was writing some documentation on how to set up a basic problem, I noticed that the `PHYSICAL_PROBLEM` had lost its meaning over the the last couple of years and it is hard to explain why it is called like that. It now refers actually to a numerical method that solves a particular problem and not to the physical problem. So I renamed it to `SOLVER`. 

Additionally, I removed the `REGIME_TYPE` option and added the incompressible solvers as dedicated values (`INC_EULER`, `INC_RANS`, `INC_NAVIER_STOKES`) for the new `SOLVER` option. The method `GetRegime_Type()` is also removed and it is now explicitly checked for the kind of solvers. This way we remove the implicit assumption that every feature also works with the incompressible solvers. Developers have to deliberately add a check now ...

It is a lot of code copy at the moment, but once the restructuring is finished we should be able to remove a lot of it.

## Related Work

## PR Checklist


Common/src/config_structure.cpp
```diff
@@ -4141,7 +4086,7 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
 
   /*--- Checks for incompressible flow problems. ---*/
 
-  if ((Kind_Solver == EULER) && (Kind_Regime == INCOMPRESSIBLE)) {
+  if (Kind_Solver == INC_EULER) {
```
pcarruscag: This looks nice, more expressive, and will allow cleaner code to be written.

Common/src/config_structure.cpp
```diff
@@ -5120,18 +5151,23 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
...
-      if (Ref_NonDim == DIMENSIONAL) { cout << "Dimensional simulation." << endl; }
-      else if (Ref_NonDim == FREESTREAM_PRESS_EQ_ONE) { cout << "Non-Dimensional simulation (P=1.0, Rho=1.0, T=1.0 at the farfield)." << endl; }
-      else if (Ref_NonDim == FREESTREAM_VEL_EQ_MACH) { cout << "Non-Dimensional simulation (V=Mach, Rho=1.0, T=1.0 at the farfield)." << endl; }
-      else if (Ref_NonDim == FREESTREAM_VEL_EQ_ONE) { cout << "Non-Dimensional simulation (V=1.0, Rho=1.0, T=1.0 at the farfield)." << endl; }
-    } else if (Kind_Regime == INCOMPRESSIBLE) {
-      if (Ref_Inc_NonDim == DIMENSIONAL) { cout << "Dimensional simulation." << endl; }
-      else if (Ref_Inc_NonDim == INITIAL_VALUES) { cout << "Non-Dimensional simulation using intialization values." << endl; }
-      else if (Ref_Inc_NonDim == REFERENCE_VALUES) { cout << "Non-Dimensional simulation using user-specified reference values." << endl; }
-    }
+      if (Kind_Solver == EULER         || Kind_Solver == FEM_EULER         || Kind_Solver == DISC_ADJ_EULER ||
```
pcarruscag: But this is the complete opposite, and it is all over the place so it is going to be a pain to maintain or even cleanup in the future.

talbring:
@pcarruscag In general I agree with you. It is kind of messy. I reintroduced the `Kind_Regime` again. I don't want to deal with finding a better solution for that problem in this PR.

economon:
I also like the idea of renaming to "SOLVER" but I would also say to avoid as much churn as possible in the conditionals throughout the code... looks like a wash when reading through the PR changes (almost as many +'s as -'s). Unless the changes are going to make something much more flexible or clear, I would say just keep Kind_Regime and set it in config postprocessing.

talbring:
Anyone else has comments ?

Common/src/fem_geometry_structure.cpp
```diff
@@ -340,7 +340,11 @@ CMeshFEM::CMeshFEM(CGeometry *geometry, CConfig *config) {
   /*--- and must be generalized later. As this number is also needed   ---*/
   /*--- in the solver, it is worthwhile to create a separate function  ---*/
   /*--- for this purpose.                                              ---*/
-  const bool compressible = (config->GetKind_Regime() == COMPRESSIBLE);
+  unsigned short Kind_Solver = config->GetKind_Solver();
+  const bool compressible = (Kind_Solver == FEM_EULER) ||
+                            (Kind_Solver == FEM_NAVIER_STOKES) ||
```
pcarruscag: My comment is still valid I suppose.

talbring: Which one? I reverted most of the GetRegimeType calls. I deliberately left it in some places to make clear for what solvers this part of the code is intended for. For the future we have to find a better way of getting properties for a certain kind of solver.

Common/src/toolboxes/MMS/CInviscidVortexSolution.cpp
```diff
@@ -77,12 +77,14 @@ CInviscidVortexSolution::CInviscidVortexSolution(unsigned short val_nDim,
     SU2_MPI::Error("Unsteady mode must be selected for the inviscid vortex",
                    CURRENT_FUNCTION);
 
-  if(config->GetKind_Regime() != COMPRESSIBLE)
+  if(Kind_Solver != EULER && Kind_Solver != NAVIER_STOKES && Kind_Solver != RANS &&
```
pcarruscag: This one is not solver specific

Common/src/toolboxes/MMS/CMMSNSUnitQuadSolutionWallBC.cpp
```diff
@@ -95,12 +95,14 @@ CMMSNSUnitQuadSolutionWallBC::CMMSNSUnitQuadSolutionWallBC(unsigned short val_nD
     SU2_MPI::Error("Steady mode must be selected for the MMS NS Unit Quad case with wall BCs.",
                    CURRENT_FUNCTION);
 
-  if(config->GetKind_Regime() != COMPRESSIBLE)
+  if(Kind_Solver != EULER && Kind_Solver != NAVIER_STOKES && Kind_Solver != RANS &&
```
pcarruscag: Any of these in MMS classes seem to be

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1027,32 +1029,38 @@ void CDriver::Solver_Preprocessing(CConfig* config, CGeometry** geometry, CSolve
   fem_dg_flow      = false;  fem_dg_shock_persson = false;
   e_spalart_allmaras = false; comp_spalart_allmaras = false; e_comp_spalart_allmaras = false;
   
-  bool compressible   = (config->GetKind_Regime() == COMPRESSIBLE);
-  bool incompressible = (config->GetKind_Regime() == INCOMPRESSIBLE);
+  bool compressible   = false;
```
pcarruscag: This you can keep the same without making the switch statement any more complicated.

talbring: Its even more complicated to understand when I keep it. Because then I'll have to add an if-statements later to distinguish between compressible/incompressible when actually allocating. This way you see immediately see what will be allocated (i.e. what boolean is set to true).

But anyway. All that this shows is that we have to do something about it at some point and neither approach adds or removes much readability. @rsanfer and myself have thought already about moving some the allocations to the iteration structure, as the only that class essentially has to know what is in the solver container. Likewise, the solver class is the only class that has to know what is in the numerics class and so on.

Common/src/toolboxes/MMS/CInviscidVortexSolution.cpp
talbring: Yeah it is. It only works and is intended for the compressible flow solver. What happens if the kind solver is not a flow solver ? I still think that this way is much more obvious for which solvers this part of the code is intended (and tested) for.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: And ideally own the actual container they instantiate.

Common/src/toolboxes/MMS/CInviscidVortexSolution.cpp
pcarruscag: Ok, we really need some more expressive and encapsulated way to express these things then.

config_template.cfg
```diff
@@ -34,9 +34,6 @@ KIND_VERIFICATION_SOLUTION= NO_VERIFICATION_SOLUTION
 % Mathematical problem (DIRECT, CONTINUOUS_ADJOINT, DISCRETE_ADJOINT)
 MATH_PROBLEM= DIRECT
 %
-% Regime type (COMPRESSIBLE, INCOMPRESSIBLE)
-REGIME_TYPE= COMPRESSIBLE
```
TobiKattmann: Hi, I am too late to the party, but I just noticed that `PHYSICAL_PROBLEM` a few lines above wasn't adapted in the config_template. So maybe that can be sneaked in one the next PR's.

Issue #713, opened 2019-06-14
Bug on multi-zone file output


rsanfer:
**Describe the bug**
When running a multi-zone case, SU2 at the current develop throws a Segmentation fault after writing out the Paraview ASCII for the second zone. Debugging with cgdb points to an error in free() when the following deallocation is called:
```
COutput::SetResult_Files_Parallel()
  COutput::DeallocateConnectivity_Parallel()
    if (Conn_Tria_Par != NULL) delete [] Conn_Tria_Par;
```
Running under valgrind, the code goes past the error point, but points to the same instruction
```
==29692== Invalid free() / delete / delete[] / realloc()
==29692==    at 0x483958B: operator delete[](void*) (vg_replace_malloc.c:641)
==29692==    by 0x991A10: COutput::DeallocateConnectivity_Parallel(CConfig*, CGeometry*, bool) (output_structure.cpp:18466)
==29692==    by 0x9CCFF6: COutput::SetResult_Files_Parallel(CSolver*****, CGeometry****, CConfig**, unsigned long, unsigned short) (output_structure.cpp:12608)
==29692==    by 0x905C0C: CMultizoneDriver::Output(unsigned long) (driver_direct_multizone.cpp:595)
==29692==    by 0x905E23: CMultizoneDriver::StartSolver() (driver_direct_multizone.cpp:188)
==29692==    by 0x364A19: main (SU2_CFD.cpp:174)
==29692==  Address 0x9b1e500 is 0 bytes inside a block of size 48,192 free'd
==29692==    at 0x483958B: operator delete[](void*) (vg_replace_malloc.c:641)
==29692==    by 0x991A10: COutput::DeallocateConnectivity_Parallel(CConfig*, CGeometry*, bool) (output_structure.cpp:18466)
==29692==    by 0x9CCFF6: COutput::SetResult_Files_Parallel(CSolver*****, CGeometry****, CConfig**, unsigned long, unsigned short) (output_structure.cpp:12608)
==29692==    by 0x905C0C: CMultizoneDriver::Output(unsigned long) (driver_direct_multizone.cpp:595)
==29692==    by 0x905E23: CMultizoneDriver::StartSolver() (driver_direct_multizone.cpp:188)
==29692==    by 0x364A19: main (SU2_CFD.cpp:174)
==29692==  Block was alloc'd at
==29692==    at 0x483850F: operator new[](unsigned long) (vg_replace_malloc.c:423)
[...]

PR #760, opened 2019-08-19
Elastic mesh solver and AD-wrapped code

rsanfer:
## Proposed Changes
This is the initial PR for a long overdue refactorization of the mesh deformation routines. The mesh deformation acquires here a full "solver" status, and uses the FEA Elasticity classes as its basis. The pseudo-linear elastic solver now always refers to the initial, undeformed geometrical discretization, therefore ensuring that the problem remains traceable to the original mesh, and that when volumes are computed to deform the mesh, the initial quality of the elements is observed. 

With this framework, the wall distances only need to be computed once for distance-based stiffness, and the stiffness matrix is now symmetric after applying BCs so conjugate gradient can be used to compute the mesh deformation. The adjoint sensitivities can now be propagated down to the surface mesh points without the need to use DOT, although further work is needed to propagate them to the FFD points or other DV of choice.

I tried to keep config options simple: the new mesh solver is enabled using
` DEFORM_MESH = YES `
and the relevant markers which will move and introduce deformations into the domain are set using
`MARKER_DEFORM_MESH= ( wall )`
Hopefully, previous mesh deformation structures can be deprecated and removed soon, but we will need to adapt the rigid/imposed boundary deformation of other kinds of grid movement. I kindly ask for your help in there, as I'm not very familiar with some of the legacy code for this purpose.

Additionally, in this PR I have done some refactoring so the AD-based adjoint routines are accessible from the python-wrapped version of the code. The SWIG wrapper now compiles the AD binaries and interfaces directly to sensitivity information, which should make it much easier to integrate SU2 within any python-based optimization framework. 

## Related Work
- Some work will be needed to adapt the new contiguous variable structure by @pcarruscag in #753
- I have added a couple of regression tests, but I will add tutorials once #724 is merged. As of now, a bug reported in #713 prevents me from adding any meaningful tutorial - but that should be fixed in #724 according to @talbring 
- I already started adding the mesh solver, numerics and transfer functions in separate files as discussed in #583 and already started in #725. For clarity and ease of merging, only new classes are split so far, but I will probably start splitting other structures soon in PRs to come.

## PR Checklist


SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2122,6 +2116,13 @@ class CVariable {
    */
   inline su2double GetSolution_time_n(unsigned short val_var) {return Solution_time_n[val_var]; }
 
+  /*!
+   * \brief Get the solution at time n-1.
```
pcarruscag: If the use case for these new functions is that only the solver that instantiates the derived CVariable uses them, I suggest not declaring them in base CVariable all together and using the "special node" mechanism I am proposing in #753, it would keep the CVariable header a lot cleaner.

SU2_CFD/include/variables/CMeshVariable.hpp
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "CVariable.hpp"
+
+class CMeshVariable : public CVariable {
```
pcarruscag: I think for brand new classes / methods we should try to make them const correct (getters are const and pointer/reference arguments should be also const if they are not modified).

SU2_CFD/include/variables/CMeshElement.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include "../../../Common/include/mpi_structure.hpp"
+
+#include <cmath>
+#include <iostream>
+#include <cstdlib>
+
+#include "../../../Common/include/config_structure.hpp"
+
+class CMeshElement {
```
pcarruscag: If one day this class needs to start keeping dynamic data I would say now is the time to implement a contiguous storage approach.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -1537,7 +1543,7 @@ class CUpwCUSP_Flow : public CNumerics {
  */
 class CUpwRoeBase_Flow : public CNumerics {
 protected:
-  bool implicit, grid_movement, roe_low_dissipation;
+  bool implicit, dynamic_grid, roe_low_dissipation;
```
pcarruscag: From the perspective of making live easier for people who are developing these classes, is the variable name change important?

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -707,6 +732,111 @@ class CDriver {
...
+   */
+  void SetLoads(unsigned short iMarker, unsigned short iVertex, passivedouble LoadX, passivedouble LoadY, passivedouble LoadZ);
+
+  /*!
+   * \brief Return the displacements from the FEA solver.
+   * \param[in] iMarker - Marker identifier.
+   * \param[in] iVertex - Vertex identifier.
+   * \return Vector of displacements.
+   */
+  vector<passivedouble> GetDisplacements(unsigned short iMarker, unsigned short iVertex);
```
pcarruscag: For methods of CDriver I think the names of these functions should be more descriptive.
Also they are not implemented yet but I am a bit concerned about returning vector by value, depending on the use case it may either be perfectly fine or not very efficient, can you elaborate on how this is going to be used?

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -508,95 +521,95 @@ class CDriver {
...
+  void SetVertexCoordZ(unsigned short iMarker, unsigned short iVertex, passivedouble newPosZ);
 
   /*!
    * \brief Set the VarCoord of a vertex on a specified marker.
    * \param[in] iMarker - Marker identifier.
    * \param[in] iVertex - Vertex identifier.
    * \return Norm of the VarCoord.
    */
-  su2double SetVertexVarCoord(unsigned short iMarker, unsigned short iVertex);
+  passivedouble SetVertexVarCoord(unsigned short iMarker, unsigned short iVertex);
```
pcarruscag: I like seeing passivedouble here and there :) no need to pay for what is not being used.

SU2_CFD/include/numerics_structure.hpp
rsanfer: Given that I changed the method to `GetDynamic_Grid()`, I think that renaming the variable makes things more consistent.

pcarruscag: Fair enough, I only mentioned it because I think there are 4 PR's at the moment with changes to this type of booleans, let's just make sure things are kept consistent after all is merged.

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -7323,6 +7545,42 @@ CIncNSSolver::CIncNSSolver(CGeometry *geometry, CConfig *config, unsigned short
     }
   }
 
+  VertexTraction = NULL;
```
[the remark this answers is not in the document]
rsanfer: Yep, just moved the initialization to CSolver in d3bc9b5ba. Instead of adding a function to the inline, given that we are moving already the inlines to the declarations as in #725, I declared the function directly in solver_structure.hpp, which soon should become CSolver.hpp

SU2_CFD/src/solver_direct_elasticity.cpp
```diff
@@ -228,17 +228,23 @@ CFEASolver::CFEASolver(CGeometry *geometry, CConfig *config) : CSolver() {
...
-      if (iVertex != -1){isVertex = true; break;}
+      if (iVertex != -1){
+        isVertex = true;
+        if (config->GetMarker_All_Fluid_Load(iMarker) == YES) {isInterface = true;}
+        break;
+      }
     }
-    if (isVertex) node[iPoint] = new CFEABoundVariable(SolRest, nDim, nVar, config);
-    else          node[iPoint] = new CFEAVariable(SolRest, nDim, nVar, config);
+    if (isVertex && isInterface)       node[iPoint] = new CFEAFSIBoundVariable(SolRest, nDim, nVar, config);
```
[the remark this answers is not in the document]
rsanfer: That's great to hear. We can talk and see how we get it adapted, but I can definitely use some support there. All FSI testcases implicitly use this derived class - I guess if it works for CFEABoundVariable it should also do for the extended FSI variable. Same applies to CMeshVariable.

SU2_CFD/include/variables/CMeshElement.hpp
rsanfer: Yep I think in the long run #753 is the way to go. In principle this class is not intended to store dynamic data, but you never know. I still need to go over #753 in more detail to understand completely the logic of the implementation.

SU2_CFD/include/numerics_structure.hpp
rsanfer: That comment was meant to go with the rest of the review :D Yeah, I agree, let's keep it consistent.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -4310,6 +4333,96 @@ class CSolver {
...
+
+    unsigned long iVertex;
+    unsigned short iMarker, iDim;
+    unsigned long nMarker = config->GetnMarker_All();
+
+    VertexTraction = new su2double** [nMarker];
+    for (iMarker = 0; iMarker < nMarker; iMarker++) {
+      VertexTraction[iMarker] = new su2double* [geometry->nVertex[iMarker]];
+      for (iVertex = 0; iVertex < geometry->nVertex[iMarker]; iVertex++) {
+        VertexTraction[iMarker][iVertex] = new su2double [nDim];
```
talbring: You can use `VertexTraction[iMarker][iVertex] = new su2double [nDim]();` to initialize it with zero.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1244,6 +1255,22 @@ void CDriver::Solver_Preprocessing(CConfig* config, CGeometry** geometry, CSolve
   
 }
 
+void CDriver::MeshSolver_Preprocessing(CSolver ***solver, CGeometry **geometry, CConfig *config) {
```
talbring: Hm can we move that to the solver preprocessing routine ?

SU2_CFD/src/numerics/CFEAMeshElasticity.cpp
```diff
@@ -0,0 +1,122 @@
...
+    for (iVar = 0; iVar < 8; iVar++) {
+      GradNi_Ref_Mat[iVar]   = new su2double[nDim];
+      GradNi_Curr_Mat[iVar]   = new su2double[nDim];
+    }
+  }
+
+}
+
+CFEAMeshElasticity::~CFEAMeshElasticity(void) {
+
```
talbring: Make sure that all of the things allocated in the constructor is deleted properly in the destructor

SU2_CFD/src/solvers/CDiscAdjMeshSolver.cpp
```diff
@@ -0,0 +1,315 @@
...
+      /*--- If sharp edge, set the sensitivity to 0 on that region ---*/
+      if (config->GetSens_Remove_Sharp()) {
+        eps = config->GetVenkat_LimiterCoeff()*config->GetRefElemLength();
+        if ( geometry->node[iPoint]->GetSharpEdge_Distance() < config->GetAdjSharp_LimiterCoeff()*eps )
+          Sensitivity = 0.0;
+      }
+
+      /*--- Store the sensitivities ---*/
+      if (!time_stepping) {
+        solver[ADJFLOW_SOL]->node[iPoint]->SetSensitivity(iDim, Sensitivity);
```
talbring: Why do you still store it in the flow adjoint solver ?

.travis.yml
```diff
@@ -72,7 +72,7 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone -b feature_elastic_mesh_solver https://github.com/su2code/TestCases.git ./TestData
```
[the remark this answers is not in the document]
rsanfer: I just opened a PR to TestCases, will replace this once it's merged.

SU2_CFD/include/solver_structure.hpp
rsanfer: Thanks! Just updated it.

SU2_CFD/src/drivers/CDriver.cpp
rsanfer: Yep, I just did. I'm sure I had a reason when I implemented the separate routine, but I agree it doesn't make sense as it stands now.

SU2_CFD/src/numerics/CFEAMeshElasticity.cpp
rsanfer: I believe that everything is already being deleted in the destructor of the base class CFEAElasticity::~CFEAElasticity() (see numerics_direct_elasticity.cpp)

SU2_CFD/src/solvers/CDiscAdjMeshSolver.cpp
rsanfer: This avoids changes to the output structure. We can update it after #724, but I thought that for now it was best to avoid conflicts.

Common/src/gauss_structure.cpp
```diff
@@ -75,6 +75,23 @@ CGaussVariable::~CGaussVariable(void) {
...
 
+CProperty::CProperty(void) {
+
+  iMat_Prop = 0;
+
+}
+
+CProperty::CProperty(unsigned long valMat_Prop) {
+
+  iMat_Prop = valMat_Prop;
```
pcarruscag: You can use initializer lists for this kind of stuff, it's regarded as the preferred way of doing it.

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -707,6 +723,112 @@ class CDriver {
...
+   */
+  vector<passivedouble> GetFEA_Velocity(unsigned short iMarker, unsigned short iVertex);
+
+  /*!
+   * \brief Return the velocities from the FEA Solver.
+   * \param[in] iMarker - Marker identifier.
+   * \param[in] iVertex - Vertex identifier.
+   * \return Vector of velocities at time n.
+   */
+  vector<passivedouble> GetFEA_Velocity_n(unsigned short iMarker, unsigned short iVertex);
```
pcarruscag: It would be good to have an architecture document about how this will work with python.

SU2_CFD/include/numerics_structure.hpp
```diff
@@ -4175,7 +4186,7 @@ class CFEAElasticity : public CNumerics {
 
   void Set_YoungModulus(unsigned short i_DV, su2double val_Young);
 
-  void SetElement_Properties(CElement *element_container, CConfig *config);
+  virtual void SetElement_Properties(CElement *element_container, CConfig *config);
```
pcarruscag: Why has this become virtual? Or was it already?

SU2_CFD/include/variables/CDiscAdjMeshBoundVariable.hpp
```diff
@@ -0,0 +1,117 @@
...
+  /*!
+   * \brief Destructor of the class.
+   */
+  ~CDiscAdjMeshBoundVariable(void);
+
+  /*!
+   * \brief Get the value of the displacement imposed at the boundary.
+   * \return Value of the boundary displacement.
+   */
+  inline su2double* GetBoundDisp_Direct() { return Bound_Disp_Direct; }
```
pcarruscag: Missing a const in the ctor argument and ideally this would return const* too.

SU2_CFD/include/variables/CMeshVariable.hpp
```diff
@@ -0,0 +1,138 @@
...
+protected:
+
+  unsigned short nDim;
+
+  su2double WallDistance;   /*!< \brief Store the wall distance in reference coordinates. */
+
+  su2double *Mesh_Coord;           /*!< \brief Store the reference coordinates of the mesh. */
```
pcarruscag: Isn't this equivalent to "Solution"? if it is I think you should use solution instead, saves you all the polymorphism and adding specific get/set methods.

SU2_CFD/src/numerics/CFEAMeshElasticity.cpp
```diff
@@ -0,0 +1,122 @@
...
+  KAux_ab = new su2double* [nDim];
+  for (iVar = 0; iVar < nDim; iVar++) {
+    KAux_ab[iVar] = new su2double[nDim];
+  }
+
+  if (nDim == 2) {
+    Ba_Mat = new su2double* [3];
+    Bb_Mat = new su2double* [3];
+    D_Mat  = new su2double* [3];
+    Ni_Vec  = new su2double [4];            /*--- As of now, 4 is the maximum number of nodes for 2D problems ---*/
```
pcarruscag: For maintenance and readability you should maybe put these hardcoded sizes as const/constexpr variables at the beginning of the ctor and explain all the "current max num of something" all in one place.

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -1038,7 +1052,25 @@ CEulerSolver::~CEulerSolver(void) {
     }
     delete [] SlidingStateNodes;
   }
-  
+
```
pcarruscag: You can move this to the dtor of CSolver too

SU2_CFD/include/drivers/CDriver.hpp
rsanfer: Yep, I'll add some documentation soon, but I'd rather do it with a tutorial. I think I should be able to do it by the time we do the release.

SU2_CFD/include/numerics_structure.hpp
rsanfer: Nice catch. I think I changed it at some point to overload the SetElement_Properties, but then I didn't need it anymore and I forgot to change it back. Solved.

SU2_CFD/include/variables/CMeshVariable.hpp
rsanfer: Nope, the Mesh_Coord variable stores the original coordinates of the mesh. The "Solution" vector is the displacement, so that the current coordinate of any mesh point is Mesh_Coord + Solution. At some point I had implemented it as "Displacement", but I changed it back for the exact same reason you mention here.

SU2_CFD/src/solver_direct_mean.cpp
rsanfer: I just did, although that meant I needed to move the nMarker and nVertex declarations also there (no big deal). That's sorted as well then.

PR #765, opened 2019-08-22
Feature sst sust

vdweide:
## Proposed Changes
This small PR introduces the SST turbulence model with sustaining terms, such that the model is free-stream preserving. Especially for external flows, where the farfield boundary is far away from the body, the turbulent intensity decays from the prescribed free stream values. This is not too much an issue when the model is run in fully turbulent mode, but it becomes important when it is combined with a transition model. It is the intention to add the transition model later on.
 

## Related Work
N/A.

## PR Checklist


talbring:
@vdweide All looks good here! Thanks! Would it maybe make sense always have this additional term included?

SU2_CFD/src/numerics_direct_turbulent.cpp
```diff
@@ -1256,10 +1262,27 @@ void CSourcePieceWise_TurbSST::ComputeResidual(su2double *val_residual, su2doubl
...
+
+   /*--- Sustaining terms, if desired. Note that if the production terms are
+         larger equal than the sustaining terms, the original formulation is
+         obtained again. This is in contrast to the version in literature
+         where the sustaining terms are simply added. This latter approach could
+         lead to problems for very big values of the free-stream turbulence
+         intensity. ---*/
+
+   if ( sustaining_terms ) {
+     const su2double sust_k = beta_star*Density_i*kAmb*omegaAmb;
```
pcarruscag: These changes are so slight I missed them while scrolling to get to the meat of the feature.

Why not simply add a config option for USE_SST_SUSTAINING_TERMS instead of through a new turb solver type? (which forces you to change lots of booleans in the code)
Or maybe use the turb solver option but in the config post processing set a different boolean to indicate sustaining_terms and reset KindTurbSolver to SST.

jayantmukho:
The PR looks good but more to @talbring 's point, do we want to conserve a vanilla SST implementation for validation purposes? 

I thought about this when I was trying to incorporate some improvements to the SST model that were published in a [2003 paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.460.2814&rep=rep1&type=pdf) by Menter et al. These are small changes in the coefficients but I wasn't sure if I should make the changes to the base SST model or make a new one.

economon:
Since most of the magic of turbulence models can be found in the source, you could adopt what we have done for the SA variants and make a new source term only in CNumerics for each new variant (or maybe have it inherit somehow from the existing one and add just an additional term without duplicating). That way, the bools can be left as SST everywhere and only one line is needed to instantiate the correct source term.

The checks on SA and SST throughout are mostly because they have 1 and 2 equations, respectively, so all of their variants can still be lumped under checks for just SA and SST type, I think.

vdweide:
Guys,

Thanks for your feedback.

@pcarruscag, the reason why I made a different enum rather than a boolean USE_SST_SUSTAINING_TERMS is that all different SA versions also have a different enum. So I thought this was more consistent. But if there is a strong preference for an additional boolean, I'm fine with that as well. What we can do is to keep the enum and set the boolean USE_SST_SUSTAINING_TERMS internally and overwrite SST_SUST to SST.

@talbring, @jayantmukho, I am in favor of keeping the original version of SST. Although the difference between the models is rather small, basically the addition of one term, the difference in results can be quite significant, especially for relatively low Reynolds numbers and large value of the turbulent intensity. 

@economon, you are right that a lot of the checks for SST are actually more general checks for a two equation model. So I think that most, if not all, checks for SST can be replaced be a check for the number of turbulent equations.  That is more general as well, in case we want to add additional turbulence models in the future, assuming that an equation is present for the turbulent kinetic energy if the number of turbulence equations is two or bigger.

talbring:
Some of the things are essentially similar to the Kind_Solver/Kind_Regime discussion in #756. We have to find a proper solution for that at some point.

PR #766, opened 2019-08-22
Feature dynamic mesh discrete adjoint

cvencro:
## Proposed Changes

This PR, created together with @TobiKattmann , makes the necessary changes to obtain time-accurate discrete adjoint gradients for unsteady problems with grid movement. Main changes are for the handling of grid velocities of previous time steps.

Regression test case to follow.

## Related Work
depends on #767 unsteady incompressible + ALE. Then a tested discrete adjoint for unsteady incompressible flow can be merged with this. 
Edit @TobiKattmann: PR767 is merged so I remove the dependent label

## PR Checklist


Common/src/dual_grid_structure.cpp
```diff
@@ -107,6 +108,9 @@ CPoint::CPoint(unsigned short val_nDim, unsigned long val_globalindex, CConfig *
 
   if ( config->GetGrid_Movement() ) {
     GridVel  = new su2double[nDim];
+    GridVel_Old  = new su2double[nDim];
+    GridVel_n  = new su2double[nDim];
+    GridVel_n1  = new su2double[nDim];
```
talbring: Do we always need this arrays or just in case of the discrete adjoint ?

Common/src/config_structure.cpp
```diff
@@ -4041,10 +4041,6 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
 
       Restart_Flow = false;
 
-      if (GetGrid_Movement()) {
-        SU2_MPI::Error("Dynamic mesh movement currently not supported for the discrete adjoint solver.", CURRENT_FUNCTION);
```
talbring: I guess it only works for rigid movement at the moment then, am I correct ?

Common/include/dual_grid_structure.inl
```diff
@@ -176,6 +182,36 @@ inline void CPoint::SetGridVel(su2double *val_gridvel) {
 		GridVel[iDim] = val_gridvel[iDim];
 }
 
+inline void CPoint::SetGridVel_Old(su2double *val_gridvel) { 
+	for (unsigned short iDim = 0; iDim < nDim; iDim++)
```
pcarruscag: Please change these tabs to spaces.

Common/src/dual_grid_structure.cpp
TobiKattmann: Wrapped the allocation it in a Discrete Adjoint conditional.

Common/src/config_structure.cpp
TobiKattmann: Correct. Now checks` Kind_GridMovement != RIGID_MOTION` and throws an error.

Common/include/dual_grid_structure.inl
TobiKattmann: Yep, went over all 'Files changed' and fixed all indentation mismatches I saw. 
Btw: In e.g. dual_grid_structure.hpp, most indentations are done using tabs and there are a lot of trailing whitespaces. Is it useful to change it in the whole file? Would be work of a few minutes but the latest git blame would be kindof misleading. I also noticed (which is probably) @WallyMaier in a branch `cleanup` just doing that cosmetic stuff. This is only a style(-guide) question.

talbring: Since we are currently restructuring the files anyway, we should do that clean up when creating/moving new files. This way we avoid duplicate conflicts.

Common/src/config_structure.cpp
cvencro: Rigid motion cases have been used for the verification at the moment but GCL is included to handle changing volumes of a mesh in the primal solvers. Therefore the changes in this request should capture both rigid motion and deforming mesh problems for discrete adjoints as the grid velocities are now captured correctly. Please advise if I'm missing to account for something else here? If anyone can point me to a good single zone mesh deformation test case that already exists for the primal solver, I can investigate the discrete adjoint explicitly for a deforming mesh problem as well.

Common/include/dual_grid_structure.inl
economon: Yep, please wait on the larger white space cleanup until we divide dual_grid_structure.hpp into files for each subclass.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2127,6 +2127,7 @@ void CDiscAdjFluidIteration::Preprocess(COutput *output,
   unsigned short iMesh;
   int Direct_Iter;
   bool heat = config[val_iZone]->GetWeakly_Coupled_Heat();
+  bool grid_movement_bool = config[val_iZone]->GetGrid_Movement();
```
economon: Can you please switch 'grid_movement_bool' to 'grid_movement' to keep with the convention throughout the rest of the code? Eventually, we could use something more expressive for the name like 'gridIsMoving' (find and replace everywhere), but for now, consistency is enough

TobiKattmann: Yep, but I would actually prefer `bool gridIsMoving` as `grid_movement `is used as a larger container for `CVolumetricMovement ***grid_movement`.

Common/src/dual_grid_structure.cpp
TobiKattmann: To give some background why: In the reverse sweep (say, with dual time stepping 2nd order) always the n-2 timestep is loaded from disk into memory. In that process, Coordinates and GridVels are stored. So compared to the corresponding primal timestep the adjoint had data from 2 timesteps before (...thinking in primal time) if these additional containers and corresponding data pushing wouldn't be there.
I thought it is more convenient than recomputing as all the data is already there in the restart files (and it was faster to code). I might also be beneficial for cases where GridVel is not a fixed function like in rigid motion but depends on the problem itself -> then it can be simply read from the restart file.

Common/include/dual_grid_structure.hpp
```diff
@@ -685,6 +688,24 @@ class CPoint : public CDualGrid {
 	 * \return Grid velocity at the point.
 	 */	
 	su2double *GetGridVel(void);
+
```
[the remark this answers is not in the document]
TobiKattmann: In the file mostly tabs are used for indentation, and github seems to display 4 spaces per tab. The added code follows https://su2code.github.io/docs/Style-Guide/ using 2 spaces as indentation. 
I can of course change it to be consistent with the file using tabs if that is wanted. Above in pedros review that topic came already up.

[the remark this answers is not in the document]
rsanfer: Oh sorry, I didn't notice before. Let's leave it like that then.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2127,6 +2127,7 @@ void CDiscAdjFluidIteration::Preprocess(COutput *output,
   unsigned short iMesh;
   int Direct_Iter;
   bool heat = config[val_iZone]->GetWeakly_Coupled_Heat();
+  bool grid_IsMoving = config[val_iZone]->GetGrid_Movement();
```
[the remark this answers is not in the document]
TobiKattmann: @economon mentions that as well above :) but `CVolumetricMovement ***grid_movement` an input to  that function, so the compiler throws an error `declaration of ‘bool grid_movement’ shadows a parameter`

TobiKattmann:
> So, really, obtaining the GridVel vector would just be an additional FD computation both for rigid movement and dynamic grids, and a lot of memory savings are possible.

Hey @rsanfer, I think that is ok for this PR. After a short search I guess such a routine does not exist yet. Or can you point me to the right direction? 
Otherwise `CVolumetricMovement::ComputeGridVel_from_Coord(..)` would be a naming suggestion. What do think?

talbring:
The routine you are looking for is probably this one:

`void CPhysicalGeometry::SetGridVelocity(CConfig *config, unsigned long iter)` geometry_structure.cpp:14760 (cannot paste a link, the file is too big)

cvencro:
Hi @rsanfer and @TobiKattmann , I agree with the more generalised solution for rigid and deforming meshes too and this is excellent timing of #760  to combine the new approach. 
I have FSI cases I can use, but thought it would be a good stepping stone to use fluid-only case to begin with to avoid the coupling terms for this verification. Using dynamic surface movement definitions for the single-zone pitching airfoil test which has previously been used for rigid motion, the primal solutions and gradients from finite differences for both are very similar as expected. Therefore this should be a good reference to check the discrete adjoints for deforming mesh when the adaptations for grid velocities are made using #760.

TobiKattmann:
@cvencro (& @rsanfer), did you check whether the GridVels computed via FD (in the adjoint) match with the ones from the primal? That's a question only for rigid motion (pitching & plunging) as here the GridVel is evaluated analytically, i.e. there could be a slight inconsistency between primal and adjoint regarding the GridVels.
For deforming cases the GridVels a probably evaluated the very same way in primal using `void CPhysicalGeometry::SetGridVelocity`, so no inconsistency here.

cvencro:
I haven't explicitly checked if they match but for the rigid motion cases the grid velocity is calculated from the transformation matrices and then imposed to update the grid. Therefore if the grid velocities are back calculated from the grid using FD then I would expect them to be close but not exact.

For a pitching airfoil case with FFD design variables, the average difference in gradients for an objective function of L/D was within 1% between discrete adjoint and finite differences. This is very similar for both imposed grid velocities in the adjoint preprocessing and with the FD calculation for the grid velocities. So I think the results are pretty reliable for the rigid motion cases using the updated grid velocities calculation. 

The deforming grid on the other hand is a different story. I still haven't been able to complete an adjoint run without diverging. More exploring to be done there...

Common/include/dual_grid_structure.inl
WallyMaier: Hi all, I did start to create a clean up branch, referencing the the style guide.  I can continue with it (and skip the dual grid).  But I think with new version coming in, we should clean up files that are going to be restructured.

SU2_CFD/src/drivers/CDiscAdjSinglezoneDriver.cpp
```diff
@@ -152,18 +152,6 @@ void CDiscAdjSinglezoneDriver::Run() {
     if(!config->GetTime_Domain() && (MainVariables == FLOW_CONS_VARS))
       config->SetExtIter(Adjoint_Iter);
 
-    /*--- Secondary sensitivities must be computed with a certain frequency. ---*/
-    /*--- It is also done at the beginning so all memory gets allocated.     ---*/
-    if ((Adjoint_Iter % config->GetWrt_Sol_Freq() == 0) && (SecondaryVariables != NONE)){
```
TobiKattmann: This `SecondaryRecording` is now removed entirely, as it introduces a bug for the unsteady adjoint. The Volume Mesh Sensitivity setting is a `+=` in each physical timestep for unsteady cases and `SecondaryRecording` is also called in `Postprocess() `right after the `Run()`. I.e. if it is called multiple times in one physical timestep the sensitivity adds up. With this code modification `SecondaryRecording` gets only called once per timestep in Postprocess.

For steady cases the implications are: The memory for the secondary recording gets allocated after the 'flow adjoint iteration' ... so it could be really annoying if the simulation crashes due to lack of memory after waiting for convergence. More importantly: You cannot use an intermediate restart file for e.g. DOT_AD as it is does not hold any/correct sensitivities. So it might be better to only disable this conditional for unsteady computations as I see the advantage of having this for steady computations. What do you think @rsanfer

TobiKattmann:
A little gradient validation using the added Testcase (comp euler, 150 timesteps, pitching airfoil, FFD points as DV, Efficiency obj function). 

gradient

SU2_CFD/src/drivers/CDiscAdjSinglezoneDriver.cpp
rsanfer: I honestly prefer it that way as it is much faster for steady-state computations in moving meshes (a lot less unnecessary recordings). In fact that was my first implementation, but then I added this SecondaryRecording after discussion with @talbring, because of the drawbacks you mention. No 100% what's the best approach...

talbring: My only concern is the memory thing, because usually the recording with coordinates as inputs takes more memory than the one with the solution variables. I am not so worried about the intermediate restart files, since you can just do a restart to get the correct sensitivity values. 

Lets do it without the SecondaryRecording. If it really leads to problems, we can always change it back.

talbring:
@TobiKattmann The validation looks perfect. Happy that it works now.

talbring:
Remember to revert the Travis changes.

SU2_CFD/src/drivers/CDiscAdjSinglezoneDriver.cpp
rsanfer: Ok, sounds good to me.

TobiKattmann:
Alright, the su2code/Testcases repo is already merged, .travis.yml is reverted. If there are no further request than this can be merged from my standpoint. 
Big thanks to @cvencro for collaboration :)

talbring:
Thanks. Merging it in. Just as a reminder: without a Tutorial or Documentation, there will be no advertising :/

PR #767, opened 2019-08-22
Feature incompressible ale

cvencro:
## Proposed Changes

This request enables unsteady incompressible flow simulation with grid movement. Changes are made as part of the collaboration with @economon  and @TobiKattmann  to expand the capabilities of the incompressible solver. Arbitrary Lagrangian-Eulerian formulation of the Low-Mach Navier-Stokes equations have been implemented which is suitable for both rigid grid motion and mesh deformation problems.

Regression test case to follow.

## Related Work

#766 

## PR Checklist

- [X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X ] I have added a test case that demonstrates my contribution, if necessary.


SU2_CFD/src/numerics_direct_mean_inc.cpp
```diff
@@ -849,6 +981,79 @@ void CSourceIncBodyForce::ComputeResidual(su2double *val_residual, CConfig *conf
...
+
+  unsigned short iDim, iVar, jVar; 
+  su2double Omega[3] = {0,0,0}, Momentum[3] = {0,0,0}, Velocity_i[3] = {0,0,0}; 
+
+  bool implicit     = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT); 
+
+  /*--- Retrieve the angular velocity vector from config. ---*/ 
+
+  for (iDim = 0; iDim < 3; iDim++){
+    Omega[iDim] = config->GetRotation_Rate(iDim)/config->GetOmega_Ref();
```
talbring: You can put that in the constructor of the class

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -2096,10 +2096,14 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
     
     for (iPoint = 0; iPoint < nPointDomain; iPoint++) {
       
-      /*--- Load the conservative variables ---*/
+      /*--- Load the primitive variables ---*/
       
-      numerics->SetConservative(node[iPoint]->GetSolution(),
-                                node[iPoint]->GetSolution());
+      numerics->SetPrimitive(node[iPoint]->GetPrimitive(), NULL);
```
talbring: Why is the second argument here NULL ...

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -2096,10 +2096,14 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+
+      /*--- Set incompressible density ---*/
+
+      numerics->SetDensity(node[iPoint]->GetDensity(),
```
talbring: ... and here not ?

SU2_CFD/src/numerics_direct_mean_inc.cpp
```diff
@@ -334,7 +384,36 @@ void CCentJSTInc_Flow::ComputeResidual(su2double *val_residual, su2double **val_
...
+  /*--- Jacobian contributions due to grid motion ---*/
+  if (grid_movement) {
+
+    /*--- Recompute conservative variables ---*/
+
+    U_i[0] = DensityInc_i; U_j[0] = DensityInc_j;
+    for (iDim = 0; iDim < nDim; iDim++) {
+      U_i[iDim+1] = DensityInc_i*Velocity_i[iDim]; U_j[iDim+1] = DensityInc_j*Velocity_j[iDim];
+    }
+    U_i[nDim+1] = DensityInc_i*Enthalpy_i; U_j[nDim+1] = DensityInc_j*Enthalpy_j;
```
pcarruscag: I think you have all of the mean quantities computed, is it necessary to recompute here?

This might be a compressible vs incompressible difference but, in compressible flows only Energy gets affected by grid motion, is it completely equivalent to use Enthalpy here?

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -6128,19 +6172,54 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
...
-          if (config->GetUnsteady_Simulation() == DT_STEPPING_2ND)
-            Jacobian_i[iVar][iVar] = (3.0*Volume_nP1)/(2.0*TimeStep);
+        
+        /*--- Calculating the inverse of the preconditioning matrix
+         that multiplies the time derivative during time integration. ---*/
+
+        /*--- For implicit calculations, we multiply the preconditioner
+        by the cell volume over the time step and add to the Jac diagonal. ---*/
+
+        Jacobian_i[0][0] = 1.0/BetaInc2;
```
economon: To be safe and avoid duplication, is it possible to call the SetPreconditioner() routine here to get the matrix? That way, if the matrix changes in the future, we won't introduce bugs here.

SU2_CFD/src/numerics_direct_mean_inc.cpp
```diff
@@ -500,7 +593,35 @@ void CCentLaxInc_Flow::ComputeResidual(su2double *val_residual, su2double **val_
       }
     }
   }
-  
+
+  /*--- Jacobian contributions due to grid motion ---*/
```
economon: Can you please double-check some of these comments? They are a little misleading now, since it says only Jacobian contributions but is also makes residual changes. It might be one of my old comments that just didn't get updated during changes..

SU2_CFD/src/numerics_direct_mean_inc.cpp
TobiKattmann: Yep, done.

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: As the source term integration uses cellwise (node i) constant data, the data for the 'node j' is not necessary for source terms. For density of 'node j' I now changed it to `0.0` to express clearly that that it is not necessary. To answer the question: I guess there was no bigger thought behind that... just couldn't think of a scalar alternative to the `NULL` pointer at that moment (still can't tbh) or just copied from the` if (axisymmetric)` part below where it is done as well :)

economon: You can also pass iPoint in for both arguments if you want something more clear, but yes, for the sources we don’t use jPoint

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -3737,7 +3738,7 @@ void CDriver::StartSolver(){
 
     /*--- Perform a dynamic mesh update if required. ---*/
 
-      if (!fem_solver) {
+      if (!fem_solver && !(config_container[ZONE_0]->GetGrid_Movement() && config_container[ZONE_0]->GetDiscrete_Adjoint())) {
```
rsanfer: The StartSolver() in CDriver should be deprecated soon. Has this feature also been tested with the single-zone driver, using `SINGLEZONE_DRIVER = YES`? This should become the default once #724 is merged.

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -6751,7 +6829,7 @@ void CIncEulerSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConf
   }
   
   /*--- Update the old geometry (coordinates n and n-1) in dual time-stepping strategy ---*/
-  if (dual_time && grid_movement)
+  if (dual_time && grid_movement && (config->GetKind_GridMovement() != RIGID_MOTION))
```
rsanfer: This just shows what we were discussing a couple of days ago, @talbring, that we should somehow distinguish between rigid grid motion and deformable grids. In rigid grid motion the Grid Velocity is an independent variable of the problem, while in deformable meshes is just derived from Solution_time_n and Solution_time_n1 in second order.

talbring:
@cvencro, @TobiKattmann You probably have test case ready to go. Would you mind writing a small tutorial? Because at the moment we have no unsteady cases there ...

cvencro:
Thank you all for the review and comments. I've added a test case with this pull request for a pitching airfoil. The config is in the main repository and the mesh is added in the TestCases repository with the same branch name. Can both pleased be merged in together? Also about the tutorial request, I'll have a go at writing one in a feature branch of the website repository.

SU2_CFD/src/drivers/CDriver.cpp
cvencro: I've now tested this with SINGLEZONE_DRIVER = YES and NO, the primal result is identical. The added test case has the setting SINGLEZONE_DRIVER= YES.

SU2_CFD/src/numerics_direct_mean_inc.cpp
cvencro: One of the main changes to the governing equations for the incompressible solver is the simplified energy equation. The final term in the convective flux for the low Mach equations used in the incompressible solver is rho*Cp*T*v.  As part of the change to account for grid motion, the grid velocity is removed from all the velocity terms in the incompressible convective flux, similar to the compressible flow, which I think should still hold true with enthalpy.

SU2_CFD/src/solver_direct_mean_inc.cpp
cvencro: Yes, I've updated to get the preconditioner matrix directly using SetPreconditioner() for the Jacobian contributions of both static and dynamic grid problems

SU2_CFD/src/numerics_direct_mean_inc.cpp
cvencro: The comments have been updated now for FDS, JST and LAX to make it clear when the residual and Jacobian contributions due to grid motion are handled

TobiKattmann:
Hi @cvencro , I think it would be good to add that Testcase to the regression tests to 'guard' the code. 
I actually wanted to do that real quick but it guess that with `SINGLEZONE_DRIVER` one  cannot access `TIME_ITER`, only `EXT_ITER` with the` .test_iter` variable of the python Testcase class. Not sure if that is addressed in #724 , maybe @rsanfer or @talbring can give a quick info as this is probably affecting all unsteady regression tests. 

Concerning the Testcase repo: You can simply open a PR for the develop branch of the Testcase repo, and refer to the corresponding PR here (and vice versa). As far as I know the Testcase repo is not protected, i.e. you can merge without any check. And as it is just one added mesh and no other change you could do it right now from my point of view.

Another point: What about the rotating_frame for incompressible flow? If it is untested I would like an error catch in place that this feature can't be used out of the box.

talbring:
@TobiKattmann Just test whether it works and gives the same results. The regression test should be added still with the old driver. I will adapt it accordingly.

TobiKattmann:
@cvencro & @talbring, Ok Testcase added using the old driver for serial and parallel regression. Residuals are identical to new driver. In the .cfg the new driver options are commented out such that switching is a matter of (un)commenting lines.

I put up a PR for the Testcase here https://github.com/su2code/TestCases/pull/36 and if the regression tests pass I would merge the Testcase PR and revert the .travis.yml

cvencro:
@TobiKattmann , thanks for the regression test. The unsteady grid motion simulations work with the updates to the ALE formulation. The rotating frame is solved as a steady problem with an added source term, an initial implementation of which has been implemented for the incompressible solver but that is still a work in progress and probably outside the scope of this PR as it stands, so I've added an error catch.

TobiKattmann:
Ok, I now believe that all issues/comments are addressed. 
@rsanfer , the SINGLEZONE DRIVER check is the only remaining requested change currently. @cvencro and I both confirmed that the residuals are identical to the old driver. 
So if you mark that as resolved and there are no further comments, I think this can be merged :)

rsanfer:
Yep, I think this is good to go. Thanks, @cvencro and @TobiKattmann!!

PR #771, opened 2019-08-25
Feature PaStiX wrapper (external direct sparse solver)

pcarruscag:
## Scope
Elasticity problems can be quite demanding linear solver wise as your typical case will have very few Dirichlet boundaries.
Common iterative methods can only go so far, and pseudo time strategies to improve conditioning do not work as well as for fluid problems due to the elliptic nature of the equations.
Scalability is also an issue since most preconditioners get weaker as the problem gets partitioned.
Direct sparse solvers handle these problems just fine, so life is good again, until you run out of RAM that is (which only ever happens in 3D). Time complexity is about O(N^2) and storage O(N^1.5).

## Proposed Changes
Wrapper class for Inria's direct sparse solver [PaStiX](http://pastix.gforge.inria.fr/files/README-txt.html).
Why this solver? Because it is open source, has active development in advanced features like GPU support and compression techniques to reduce memory footprint, and from my experience it is really robust.

## Implementation
As this feature benefits mostly the FSI people (@rsanfer, @cvencro, and me) I tried to keep it as standalone as possible.
The core implementation is in separate files and nothing needs to change in CSysMatrix for this to work.

## Usage
Available factorizations:
- LDLT
- LU
- ILU(k)

The ILU is different from the one we have in that it is computed for the entire matrix and not just for the local matrices (i.e. it considers halos), it is not very fast though.
The three factorizations can be used as preconditioners via options PASTIX_ILU, PASTIX_LU, and PASTIX_LDLT. And the first two as solvers via options PASTIX_LU and PASTIX_LDLT (in which case you should use a preconditioner type that does not cause CSysMatrix to allocate extra memory (LU_SGS).
The reason LDLT and LU can also be used as preconditioners is that it is possible to set a factorization frequency (via PASTIX_FACTORIZATION_FREQUENCY) so that the factorizations are only computed every N iterations or only on the first call (frequency set to 0), this is very useful for non linear elasticity.
The fill-in for the ILU is set via PASTIX_FILL_LEVEL, I did not reuse the option we have again to keep CSysMatrix from allocating any more than it needs to.
By default the solvers run silently, but more verbosity can be set via PASTIX_VERBOSITY_LEVEL (0 : silent, 1 : only during major operations, 2 : 1 + on every solve).

## Compilation
_TODO_
This needs linking with another library (Scotch) and with BLAS, I will explain this in detail.
Question: For v7 are we moving exclusively to meson?

## PR Checklist


talbring:
In general I don't have anything against using external libraries as soon as they are not essential for building the code. Also it seems that the changes are really non-intrusive. 

> Question: For v7 are we moving exclusively to meson?

The plan is not to move exclusively yet, so that the base version of the code can be still compiled with automake. But there is also no point in trying to support both at all cost, so if this feature can only be enabled when using meson, then that's totally fine.

pcarruscag:
Thanks for the comments guys.
I think for preconditioners CSysMatrix already gives default hooks to attach external solvers to, there is a build method and a compute method, and that was the logic I followed for this.
All the operations I do inside the wrapper are solver specific, the only thing that could eventually be abstracted from there is the conversion between our matrix format and the one PaStiX requires (which is a mere re-ordering so the column indices are ordered according to global index in linear partitioning, unfortunately this is enough to warrant a full copy).
My reason not to have gone with PETSc is [this table](https://www.mcs.anl.gov/petsc/documentation/linearsolvertable.html) that shows very little compatibility with the SU2 block sparse matrix format.

Common/src/config_structure.cpp
```diff
@@ -2445,6 +2445,15 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Permuting eigenvectors for UQ analysis */
   addBoolOption("UQ_PERMUTE", uq_permute, false);
 
+  /* DESCRIPTION: Number of calls to 'Build' that trigger re-factorization (0 means only once). */
```
rsanfer: We need a strategy for the config_template.cpp. Adding these there might be confusing...

talbring:
@pcarruscag Do you want to add the dependencies to meson in this PR?

Common/src/config_structure.cpp
pcarruscag: I prefixed all related options with PASTIX... What I can do is explain those options in an example cfg but not add them to the template config.

pcarruscag:
@talbring, I want to write up how to get/compile/link everything but I was not planing to change meson.py as the options can change quite a bit.
What is the best place to add a tutorial? Directly to the website?

talbring:
Ah I just mean adding it to the meson.build file as an option and to provide the paths to headers/libraries. I am not in favor of doing it via environment variables ...

pcarruscag:
Ok I'll see what I can do.

pcarruscag:
@talbring how do you want the "interface" to work for the user?
Do they modify meson.build?
Do they define environment variables for the root folder of each dependency? (I guess not)
Do we force the installation directory? (not a good idea since some components can be installed with apt-get)
Do we force them to link only statically / dynamically?
Also are you aware that linker flags for BLAS can be compiler dependent, especially for MKL?

talbring:
The standard way should use pkg-config. I see that pastix and BLAS both provide the required files. Should be quite easy actually to add them.

pcarruscag:
What about clusters?

talbring:
This approach also works for user installed libraries, by setting `PKG_CONFIG_PATH`

pcarruscag:
MKL does not work like that... https://software.intel.com/en-us/articles/intel-mkl-link-line-advisor
The flags change...

talbring:
Provide me with the commands you compile it and I will see how we can add that

pcarruscag:
@vdweide , you use mkl, what is your typical build recipe?
Let's take the opportunity and add options -Dhave-mkl and -Dhave-lapack to the build process.
Then we build this on top, i.e. to use -Dhave-pastix one of those is required. I can then restrict the build instructions for the solver to ensure it is compatible with what we need for the other BLAS features of the code.

vdweide:
@pcarruscag, I have not built the MKL executable with meson yet. What I typically did when using configure is to use environment variables for the compiler and linker flags.

vdweide:
Yes, compiler flags -DHAVE_MKL, linker flags -mkl=sequential

vdweide:
Yes, compiler flags -DHAVE_MKL, linker flags -mkl=sequential

TestCases/pastix_support/readme.txt
```diff
@@ -0,0 +1,40 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
pcarruscag: @rsanfer, tutorial, so far the recipe has only been tested on Ubuntu 18.04 (gcc 7.4) OpenMPI 3.1.4, MKL 2017 or OpenBLAS 0.3.2

meson.build
```diff
@@ -66,6 +66,52 @@ if get_option('enable-tecio')
   subdir('externals/tecio')
 endif
 
+# PaStiX
```
pcarruscag: @talbring, this is the best I came up with after one day of tweaking, this version of the solver does not have a pkg-config thingy and scotch (the other dependency) needs a custom build so apt-get is not possible (I think...).
If it is not up to snuff I'll drop the PR (too much time on this already).

talbring: Thanks of taking that effort. It is not to bother you or something like. The goal is always to make it as easy as possible for other people to use it. Don't get me wrong, but just having things integrated for the sake of it ... that's not something we want for the project in my opinion. And I know that it sometimes takes time, but it is worth it.

Having that said, it looks good to me!

TestCases/pastix_support/readme.txt
talbring: Eventually we should put that to the docs on the website.

meson.build
pcarruscag: Thanks, I understand that but it is not just for the sake of integrating, structural problems do need something like this, most FEA solvers have similar solutions.

I have yet to make this work with an Intel compiler and Intel mpi, I am adding WIP to the name.

TestCases/pastix_support/readme.txt
```diff
@@ -65,6 +65,6 @@
 % - Ubuntu 18.04, gcc 7.4, ompi 3.1.4, mkl 2017, openblas 0.2.20 and 0.3.2.dev
```
pcarruscag: Well on the systems I have access to I get mpi runtime errors when I use icc (17 or 19) I don't think the bug is on my side...
I did test this on a variety of gcc / mpi / mkl / blas combinations so I think it is good to go.

PR #773, opened 2019-08-26
Streamwise periodicity for incompressible flow

TobiKattmann:
Hi, 

this PR adds streamwise periodic flow  (incl energy eq) for the incompressible solver. For a more detailed information please see the [website additions](https://github.com/su2code/su2code.github.io/pull/53) including a tutorial and theory section.

~half of the lines is .cfg's as those CHT setups require a bit more real estate. So the actual code is not too long.

Here a squash merge could be done as it merged develop quite a few times and has a bunch of 'small fix' commits. But to my knowledge that is not really necessary as.

I am ready to pass on the red lantern to someone else 🐌 

## Related Work
This work builds directly on #652 of @economon

## PR Checklist


pcarruscag:
I've put a "WIP" in the tittle just to make it easier to create the change log for releases (there is the manual step where we need  to remove stuff that it not part of the release).

TobiKattmann:
I guess one can ignore the CodeFactor complaints here as it just gives a vague complex code without any further line information error for CIncEulerSolver.cpp (which I could understand to some degree) and CSolver.hpp (which I dont  understand at all).

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+
+          /*--- Get the squared norm of the current point. ---*/
+          auto norm = GeometryToolbox::SquaredNorm(nDim, nodes->GetCoord(iPoint));
+
+          /*--- Check if new unique reference node is found. ---*/
+          if (norm < min_norm || iVertex == 0) {
+            min_norm = norm;
+            for (unsigned short iDim = 0; iDim < nDim; iDim++)
+              Buffer_Send_RefNode[iDim] = nodes->GetCoord(iPoint,iDim);
+          }
```
pcarruscag, line 7507: I would initialize norm to a large value to avoid the iVertex logic, and make Buffer_Send a const pointer, i.e. without copying the data.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+  /*--- reference node is not relevant at all. This is probably true for most streamwise    ---*/
+  /*--- periodic cases. Other cases where it is relevant could look like this (--->( or     ---*/
+  /*--- \--->\ . The chosen metric is the minimal distance to the origin.                   ---*/
+  /*-------------------------------------------------------------------------------------------*/
+
+  /*--- Initialize/Allocate variables. ---*/
+  su2double min_norm = 0.0;
+
+  vector<su2double> Buffer_Send_RefNode(nDim, 1e300),
+                    Buffer_Recv_RefNode(static_cast<size_t>(size)*nDim);
```
pcarruscag, line 7479: This can be an su2activematrix (it has a .data() member like std::vector (this lets you use operator (i,j) instead of i*n+j)

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+
+  for (int iRank = 0; iRank < size; iRank++) {
+
+    /*--- Get the norm of the current Point. ---*/
+    auto norm = GeometryToolbox::SquaredNorm(nDim, &Buffer_Recv_RefNode[static_cast<size_t>(iRank)*nDim]);
+
+    /*--- Check if new unique reference node is found. ---*/
+    if (norm < min_norm || iRank == 0) {
+      min_norm = norm;
+      for (unsigned short iDim = 0; iDim < nDim; iDim++)
```
pcarruscag, line 7533: I would also communicate the minimum distance together with the coordinates (size * (ndim+1)) and then you don't need to look for the minimum again by redoing similar calculations.

SU2_CFD/include/numerics/flow/flow_sources.hpp
```diff
@@ -323,21 +341,83 @@ class CSourceWindGust final : public CSourceBase_Flow {
...
  */
-class CSourceRadiation : public CSourceBase_Flow {
+class CSourceIncStreamwise_Periodic final : public CSourceBase_Flow {
 private:
-  bool implicit;
 
+  bool turbulent; /*!< \brief Turbulence model used. */
+  bool energy;    /*!< \brief Energy equation on. */
+  bool streamwisePeriodic_temperature; /*!< \brief Periodicity in energy equation */
+  su2double Streamwise_Coord_Vector[MAXNDIM] = {0.0}; /*!< \brief Translation vector between streamwise periodic surfaces. */
```
pcarruscag, line 355: It is hard to choose one style of formatting given the variety in SU2 xD

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
   delete [] Buffer_Recv_GlobalIndex;
   delete [] Buffer_Recv_Vertex;
   delete [] Buffer_Recv_Marker;
+}
+
+void CPhysicalGeometry::FindUniqueNode_PeriodicBound(CConfig *config) {
```
pcarruscag, line 7461: const CConfig possible?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2551,6 +2551,34 @@ class CVariable {
...
+   * \param[in] val_pressure - pressure value.
+   */
+  inline virtual void SetStreamwise_Periodic_RecoveredPressure(unsigned long iPoint,su2double val_pressure) { }
+
+  /*!
+   * \brief A virtual member: Get the recovered pressure for streamwise periodic flow.
+   * \param[in] iPoint - Point index.
+   * \return Recovered/Physical pressure for streamwise periodic flow.
+   */
+  inline virtual su2double GetStreamwise_Periodic_RecoveredPressure(unsigned long iPoint) const { return 0.0; }
```
pcarruscag, line 2566: Do you need them to be virtual? Or are they used only in the Inc Solver?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -103,7 +104,8 @@ CIncEulerSolver::CIncEulerSolver(CGeometry *geometry, CConfig *config, unsigned
 
   nDim = geometry->GetnDim();
 
-  nVar = nDim+2; nPrimVar = nDim+9; nPrimVarGrad = nDim+4;
+  /*--- Make sure to align the sizes with the constructor of CIncEulerVariable. ---*/
+  nVar = nDim+2; nPrimVar = nDim+9; nPrimVarGrad = nDim+6;
```
pcarruscag, line 108: "This little maneuver is going to cost us 51 years"
Similarly to what is done for Axisymmetric you could use the AuxVar to compute the extra gradient you need and that is not normally used.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1490,6 +1495,80 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+      auto residual = numerics->ComputeResidual(config);
+      LinSysRes.AddBlock(iPoint, residual);
+
+      /*--- Add the implicit Jacobian contribution ---*/
+      if (implicit) Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
+
+    }// for iPoint
+
+    if(!streamwise_periodic_temperature && energy) {
+      CNumerics* second_numerics = numerics_container[SOURCE_SECOND_TERM];
```
pcarruscag, line 1531: You need the good'ol (good new?) OpenMP logic to use one numerics per thread, there should be an example on the top of this function.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1490,6 +1495,80 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+                                                     Streamwise_Periodic_InletTemperature);
+
+      /*--- This bit acts as a boundary condition rather than a source term. But logically it fits better here. ---*/
+      for (auto iMarker = 0ul; iMarker < config->GetnMarker_All(); iMarker++) {
+
+        /*--- Only "inlet"/donor periodic marker ---*/
+        if (config->GetMarker_All_KindBC(iMarker) == PERIODIC_BOUNDARY &&
+            config->GetMarker_All_PerBound(iMarker) == 1) {
+
+          for (auto iVertex = 0ul; iVertex < geometry->nVertex[iMarker]; iVertex++) {
```
pcarruscag, line 1542: Then on top of this loop SU2_OMP_FOR_STAT(OMP_MIN_SIZE) (or something)

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -113,6 +114,8 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
   Delta_Time.resize(nPoint) = su2double(0.0);
   Lambda.resize(nPoint) = su2double(0.0);
   Sensor.resize(nPoint) = su2double(0.0);
+  Streamwise_Periodic_RecoveredPressure.resize(nPoint)    = su2double(0.0);
+  Streamwise_Periodic_RecoveredTemperature.resize(nPoint) = su2double(0.0);
```
pcarruscag, line 119: if (streamwise_periodic) ?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -97,6 +98,43 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
 
   ComputeVorticityAndStrainMag<1>(*config, iMesh);
 
+  /*--- Compute recovered pressure and temperature for streamwise periodic flow ---*/
+  if (config->GetKind_Streamwise_Periodic() != NONE) {
+
+    /*--- Define and initialize helping variables ---*/
+    su2double dot_product, Pressure_Recovered, Temperature_Recovered;
```
pcarruscag, line 105: local to the loop please

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -97,6 +98,43 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
...
+    const su2double delta_p  = config->GetStreamwise_Periodic_PressureDrop() / config->GetPressure_Ref();
+
+    /*--- Reference node on inlet periodic marker to compute relative distance along periodic translation vector. ---*/
+    const su2double* ReferenceNode = geometry->GetStreamwise_Periodic_RefNode();
+
+    /*--- Compute square of the distance between the 2 periodic surfaces. ---*/
+    const su2double norm2_translation = GeometryToolbox::SquaredNorm(nDim, config->GetPeriodic_Translation(0));
+
+    /*--- Compute recoverd pressure and temperature for all points ---*/
+    for (auto iPoint = 0ul; iPoint < nPoint; iPoint++) {
```
pcarruscag, line 116: SU2_OMP_FOR_STAT(omp_chunk_size) on top

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -219,6 +263,22 @@ void CIncNSSolver::BC_Wall_Generic(const CGeometry *geometry, const CConfig *con
       Compute the residual due to the prescribed heat flux. ---*/
 
       LinSysRes(iPoint, nDim+1) -= Wall_HeatFlux*Area;
+
+      /*--- With streamwise periodic flow and heatflux walls an additional term is introduced in the boundary formulation ---*/
+        if (streamwise_periodic && streamwise_periodic_temperature) {
+
```
pcarruscag, line 269: misaligned?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -97,6 +98,43 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
...
+      /*--- InnerIter > 0 as otherwise MassFlow in the denominator would be zero ---*/
+      if (energy && InnerIter > 0) {
+        Temperature_Recovered  = nodes->GetSolution(iPoint, nDim+1);
+        Temperature_Recovered += Streamwise_Periodic_IntegratedHeatFlow / (Streamwise_Periodic_MassFlow * nodes->GetSpecificHeatCp(iPoint) * norm2_translation) * dot_product;
+        nodes->SetStreamwise_Periodic_RecoveredTemperature(iPoint, Temperature_Recovered);
+      }
+    }
+
+    /*--- Compute the integrated Heatflux Q into the domain, and massflow over periodic markers ---*/
+    GetStreamwise_Periodic_Properties(geometry, config, iMesh);
```
pcarruscag, line 136: SU2_OMP_MASTER
...
SU2_OMP_BARRIER

For the easy way out

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -97,6 +98,43 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
 
   ComputeVorticityAndStrainMag<1>(*config, iMesh);
 
+  /*--- Compute recovered pressure and temperature for streamwise periodic flow ---*/
+  if (config->GetKind_Streamwise_Periodic() != NONE) {
+
```
pcarruscag, line 103: Move this block to a private function.
We should have code that looks like a nice tree, big classes use medium that use small, small functions use smaller functions.

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -479,7 +482,7 @@ CNumerics::ResidualType<> CSourceGravity::ComputeResidual(const CConfig* config)
     residual[iVar] = 0.0;
 
   /*--- Evaluate the source term  ---*/
-  residual[nDim] = Volume * U_i[0] * STANDARD_GRAVITY;
+  residual[nDim] = Volume * U_i[0] * STANDARD_GRAVITY / Force_Ref;
```
pcarruscag, line 485: Looks like a bug fix :+1:

Common/include/CConfig.hpp
```diff
@@ -5729,6 +5743,42 @@ class CConfig {
...
+   * \brief Get the value of the pressure delta from which body force vector is computed.
+   * \return Delta Pressure for body force computation.
+   */
+  su2double GetStreamwise_Periodic_PressureDrop(void) const { return Streamwise_Periodic_PressureDrop; }
+  
+  /*!
+   * \brief Set the value of the pressure delta from which body force vector is computed.
+   * \param[in] delta_p - pressure difference between in- and outlet.
+   */
+  void SetStreamwise_Periodic_PressureDrop(su2double delta_p) { Streamwise_Periodic_PressureDrop = delta_p; }
```
pcarruscag, line 5774: Please please please don't use the config for this, keep it in the solver, pass it to numerics "explicitly" (with a set something) instead of mixed with the rest of the world in CConfig.

pcarruscag (commented):
Alright I think I covered everything.
World class documented code :+1: 
Loads of comments on this second pass, but they are just small re-arrangements to encapsulate the feature a bit better in terms of code, and in terms of side effects when not using it.

pcarruscag:
No need to squash, the testcases repo is actually a bit messed up ATM because of that (oooops)

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+
+  /*-------------------------------------------------------------------------------------------*/
+  /*--- Find reference node on the 'inlet' streamwise periodic marker for the computation   ---*/
+  /*--- of recovered pressure/temperature, such that this found node is independent of the  ---*/
+  /*--- number of ranks. This does not affect the 'correctness' of the solution as the      ---*/
+  /*--- absolute value is arbitrary anyway, but it assures that the solution does not change---*/
+  /*--- with a higher number of ranks. If the periodic markers are a line\plane and the     ---*/
+  /*--- streamwise coordiante vector is perpendicular to that |--->|, the choice of the     ---*/
```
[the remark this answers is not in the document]
TobiKattmann: I actually have a spell checker in vscode which didn't complain about that (and it found quite a bit elsewhere)

(also interesting, I had a pending comment myself which prevented me from comemnting anything here 🤔 )

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1253,7 +1253,7 @@ void COutput::PrepareHistoryFile(CConfig *config){
   historyFileTable->SetAlign(PrintingToolbox::CTablePrinter::CENTER);
   historyFileTable->SetPrintHeaderTopLine(false);
   historyFileTable->SetPrintHeaderBottomLine(false);
-  historyFileTable->SetPrecision(10);
+  historyFileTable->SetPrecision(config->OptionIsSet("OUTPUT_PRECISION") ? config->GetOutput_Precision() : 10);
```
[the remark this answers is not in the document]
TobiKattmann: In SU2_DOT the precision is not set and the default precision is 6... I guess in DOT it is more reasonable to extend to 10 (rather cut to 6 in history) so I will change the logic to have that over in DOT and then one could enforce the 10 over there and see what it all breaks (but not in this PR ;) )

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2782,6 +2861,154 @@ void CIncEulerSolver::GetOutlet_Properties(CGeometry *geometry, CConfig *config,
...
+
+          /*--- A = dot_prod(n_A*n_A), with n_A beeing the area-normal. ---*/
+
+          const auto AreaNormal = geometry->vertex[iMarker][iVertex]->GetNormal();
+
+          auto FaceArea = GeometryToolbox::Norm(nDim, AreaNormal);
+
+          // One could add a CVariable method to return a pointer to the first Vel element to directly plug into GeomToolbox
+          su2double Velocity[MAXNDIM] = {0.0};
+          for (auto iDim = 0; iDim < nDim; iDim++) { Velocity[iDim] = nodes->GetVelocity(iPoint, iDim); }
```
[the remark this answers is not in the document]
TobiKattmann: hmm. I now replaced auto with unsigned short

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2782,6 +2861,154 @@ void CIncEulerSolver::GetOutlet_Properties(CGeometry *geometry, CConfig *config,
...
+
+          auto iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
+
+          if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+          const auto AreaNormal = geometry->vertex[iMarker][iVertex]->GetNormal();
+
+          auto FaceArea = GeometryToolbox::Norm(nDim, AreaNormal);
+
+          HeatFlow_Local += FaceArea * (-1.0) * config->GetWall_HeatFlux(Marker_StringTag)/config->GetHeat_Flux_Ref();;
```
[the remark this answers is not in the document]
TobiKattmann: it is now 
```c++
if (config->GetMarker_All_KindBC(iMarker) == HEAT_FLUX) {

        /*--- Identify the boundary by string name adn retrive heatflux from config ---*/
        const auto Marker_StringTag = config->GetMarker_All_TagBound(iMarker);
        const auto Wall_HeatFlux = config->GetWall_HeatFlux(Marker_StringTag)

        for (auto iVertex = 0ul; iVertex < geometry->nVertex[iMarker]; iVertex++) {

          auto iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -97,6 +98,43 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
...
+      for (unsigned short iDim = 0; iDim < nDim; iDim++)
+        dot_product += fabs( (geometry->nodes->GetCoord(iPoint,iDim) - ReferenceNode[iDim]) * config->GetPeriodic_Translation(0)[iDim]);
+
+      /*--- Second, substract/add correction from reduced pressure/temperature to get recoverd pressure/temperature ---*/
+      Pressure_Recovered = nodes->GetSolution(iPoint, 0) - delta_p / norm2_translation * dot_product;
+      nodes->SetStreamwise_Periodic_RecoveredPressure(iPoint, Pressure_Recovered);
+
+      /*--- InnerIter > 0 as otherwise MassFlow in the denominator would be zero ---*/
+      if (energy && InnerIter > 0) {
+        Temperature_Recovered  = nodes->GetSolution(iPoint, nDim+1);
```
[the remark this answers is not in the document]
TobiKattmann: Yes that is good point. Always good to avoid these implicit accesses. There are some other places where I did that prob -> so I'll sweep through and change those as well if possible

SU2_CFD/src/solvers/CIncEulerSolver.cpp
TobiKattmann: Yep 👍

SU2_CFD/src/solvers/CIncEulerSolver.cpp
TobiKattmann: Yes. following the other loops:
```c++
for (auto iVertex = 0ul; iVertex < geometry->nVertex[iMarker]; iVertex++) {
```

SU2_CFD/src/variables/CIncEulerVariable.cpp
TobiKattmann: uff yeah for sure .. two nPoint allocations are not really great. it is now 
```c++
  if (config->GetKind_Streamwise_Periodic() != NONE) {
    Streamwise_Periodic_RecoveredPressure.resize(nPoint) = su2double(0.0);
    if (config->GetStreamwise_Periodic_Temperature())
      Streamwise_Periodic_RecoveredTemperature.resize(nPoint) = su2double(0.0);
  }
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: 👍 Pressure_Recovered is also now const, the other two not

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: Done (yeah i should have looked more carefully at those openMP things 😞 )

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: ⬅️ Go back where you belong, code!

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: Yes good catch. (I used config as a place to store the refnode before 🤫(shame) and when i changed that to Cphysicalgeometry i didn't think to make it const)

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: Done. 
SU2_OMP_BARRIER = #pragma omp barrier i.e. a synchronization point
SU2_OMP_MASTER = #pragma omp master i.e.  the next structured block is only executed by the master thread
Otherwise all threads would call and execute that function, right?

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: Yep made it a private function 
```c++
if (config->GetKind_Streamwise_Periodic() != NONE)
    Compute_Streamwise_Periodic_Recovered_Values(config, geometry, iMesh);
```

Common/include/CConfig.hpp
```diff
@@ -994,6 +995,13 @@ class CConfig {
   array<su2double, N_POLY_COEFFS> mu_polycoeffs{{0.0}};  /*!< \brief Array for viscosity polynomial coefficients. */
   array<su2double, N_POLY_COEFFS> kt_polycoeffs{{0.0}};  /*!< \brief Array for thermal conductivity polynomial coefficients. */
   bool Body_Force;                      /*!< \brief Flag to know if a body force is included in the formulation. */
+
+  unsigned short Kind_Streamwise_Periodic;          /*!< \brief Kind of Streamwise periodic flow (pressure drop or massflow) */
+  bool Streamwise_Periodic_Temperature;             /*!< \brief Use real periodicity for Energy equation or otherwise outlet source term. */
+  su2double Streamwise_Periodic_PressureDrop,       /*!< \brief Value of prescribed pressure drop [Pa] which results in an artificial body force vector. */
+            Streamwise_Periodic_TargetMassFlow,     /*!< \brief Value of prescribed massflow [kg/s] which results in an delta p and therefore an artificial body force vector. */
+            Streamwise_Periodic_OutletHeat;         /*!< /brief Heatflux boundary [W/m^2] imposed at streamwise periodic outlet. */
+
```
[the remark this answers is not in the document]
TobiKattmann: Now they all have their own su2double buddy :) and someone in the distant future introducing clang format is a little more happy
```c++
su2double Streamwise_Periodic_PressureDrop;       /*!< \brief Value of prescribed pressure drop [Pa] which results in an artificial body force vector. */
su2double Streamwise_Periodic_TargetMassFlow;     /*!< \brief Value of prescribed massflow [kg/s] which results in an delta p and therefore an artificial body force vector. */
su2double Streamwise_Periodic_OutletHeat;         /*!< /brief Heatflux boundary [W/m^2] imposed at streamwise periodic outlet. */
```

SU2_CFD/include/variables/CVariable.hpp
TobiKattmann: They are used in CFlowIncoutput 
```c++
  // Streamwise Periodicity
  if(streamwisePeriodic) {
    SetVolumeOutputValue("RECOVERED_PRESSURE", iPoint, Node_Flow->GetStreamwise_Periodic_RecoveredPressure(iPoint));
    if (heat && streamwisePeriodic_temperature)
      SetVolumeOutputValue("RECOVERED_TEMPERATURE", iPoint, Node_Flow->GetStreamwise_Periodic_RecoveredTemperature(iPoint));
  }
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
pcarruscag: omp_chunk_size is for nPoint loops, for ones over vertices should be OMP_MIN_SIZE which is a smaller value.
I think there is a "nVertex" in CSolver, getting it from geometry is not needed.

SU2_CFD/src/variables/CIncEulerVariable.cpp
pcarruscag: It would not make a whole in SU2's guts (as my grandmother would say) but if it can be avoided with a couple lines...

SU2_CFD/src/solvers/CIncNSSolver.cpp
pcarruscag: Yup, barrier makes shared variables consistent across threads, in GetStreamwise_Periodic_Properties you write to shared variables, after the barrier all threads "see" the new values.

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: I have to admit I do not understand  the `make Buffer_Send a const pointer, i.e. without copying the data.` part. I now changed the code to communicate the dist as well as you suggested, so Buffer_Send  has size nDim+1.
But for now assume Buffer_Send  just contains the coordinates then I could:
```c++
//outside
const su2double *Buffer_Send_RefNode;
...
//in the loop
Buffer_Send_RefNode = nodes->GetCoord(iPoint);
```
because I dont want the data Buffer_Send_RefNode  points to to change. That is ok, but if i wanna include the distance into that Buffer_Send_RefNode I need to have dedicated memory for that. Maybe my brain is in a const state unable to learn or you had two different ideas which contradict each other to some degree :)

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,92 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+  /*--- reference node is not relevant at all. This is probably true for most streamwise    ---*/
+  /*--- periodic cases. Other cases where it is relevant could look like this (--->( or     ---*/
+  /*--- \--->\ . The chosen metric is the minimal distance to the origin.                   ---*/
+  /*-------------------------------------------------------------------------------------------*/
+
+  /*--- Initialize/Allocate variables. ---*/
+  su2double min_norm = 0.0;
+
+  vector<su2double> Buffer_Send_RefNode(nDim, 1e300),
```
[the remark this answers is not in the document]
TobiKattmann: yep using numeric_limits::max(), is far more elegant

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: that also make the static cast unnecessary now without the int multiplication at construction 👍

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: I would initialize *min_* norm to a large value to avoid the iVertex logic ... completely right, that looks a bit nicer now

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: Yep 👍

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: They were indeed 2+ ideias, but not mutually exclusive, you can determine "who" is the closest point inside the loop by keeping either the corresponding "iPoint" or directly a const ptr to the closest coordinates, then you copy the coordinates into the send buffer only once outside of the loop.

Keeping "iPointMin" is actually the "right" way to do it in the sense that you don't rely on the xyz coordinates being together in memory (which the pointer implies), and with that you separate the algorithm (find min) from the data structure (how the coordinates are organized in memory).
This separation is the main reason I was talking about that "matrix view" for gradients, and it is why (according to the authors) the algorithms in the STL (the template library, not the format) take iterators instead of pointers.

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -385,4 +400,22 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, INCOMPRESSI
    */
   inline bool GetHasHybridParallel() const final { return true; }
 
+  /*!
+   * \brief Get the massflow of the streamwise periodic donor/outlet boundary.
+   * \return The streamwise periodic donor/outlet massflow.
+   */
+  su2double GetStreamwise_Periodic_MassFlow() const { return Streamwise_Periodic_MassFlow; }
```
pcarruscag, line 407: `override` on these three

pcarruscag: Sorry, `final` since they are in the EulerSolver

TobiKattmann: I am favouring (to cite you) `make a type for streamwise periodic things [...]these 3 go everywhere together ` and extending that to 4 with the one that is plugged through the conftig rn. So this method will probably disappear anyway, but i'll keep that in mind for the replacement 👍

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4411,6 +4411,24 @@ class CSolver {
...
+   * \brief Get the net sum of the heatflow into the domain.
+   * \return The net sum of the heatflow into the domain.
+   */
+  virtual su2double GetStreamwise_Periodic_IntegratedHeatFlow() const { return 0.0; }
+
+  /*!
+   * \brief Get the value of the area avg periodic inlet Temperature.
+   * \return Temperature value.
+   */
+  virtual su2double GetStreamwise_Periodic_InletTemperature(void) const { return 0.0; }
```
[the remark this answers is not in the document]
TobiKattmann: For gathering those 3+1(from the config which I'll add): do you have a preference of struct, tuple, or sth else? and why a solution might be preferred (if you have the time to give a rough idea)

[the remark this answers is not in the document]
pcarruscag: A struct defined in option_structure close to the streamwise enum, so that you can access things by name.
With tuples or arrays you would need to know the position that corresponds with what, and that is less readable IMO.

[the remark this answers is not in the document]
pcarruscag: But in the past I have also used "out params" to "return" multiple things from these methods...
Now I think using small custom types gives a more modular feel to things.

Common/include/CConfig.hpp
TobiKattmann: removed the Set function and made it a solver var 👍 And I actually agree ... the config should not be the place to make solver/or other things variables available to lots of places just because it is convenient

SU2_CFD/include/solvers/CSolver.hpp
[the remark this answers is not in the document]
TobiKattmann: is there a way to return s.th in a virtual function which is  like nullptr for pointer return types or 0.0 for su2doubles. 
```c++
virtual mystructtype() const { return  ?? }
```
I briefly searched but didn't find a good way. And having that return already a a member of CSolver is not exactly what I want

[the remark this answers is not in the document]
pcarruscag: Not in C++11 I don't think (https://en.cppreference.com/w/cpp/utility/optional).
Would returning a default constructed struct not work? The definition needs to be available to CSolver anyway so that you can access it in the output.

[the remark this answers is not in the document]
TobiKattmann: I guess one can
```c++
virtual mystructtype Getmystructtype () const {mystructtype foo; return  foo; }
```
I dont get any warnings on my machine with warnlevel=3 but I am not sure whether this is illegal... and I dont want to be woken up by the Cpp-Police banging on my door tomorrow early in the morning... 'Cause thats what they tend to do
On the other hand return zeros can create nasty silent bugs as well so I will take my chances here 😎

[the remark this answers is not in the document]
TobiKattmann: Ok typed that while you already answered, so i didn't see your answer right away. Is that what you refer to as `default constructed struct` ?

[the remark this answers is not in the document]
pcarruscag: return mystructtype();

SU2_CFD/src/solvers/CIncEulerSolver.cpp
TobiKattmann: :) Yep changed that to use AuxVar instead and reverted nPrimVarGrad back to original

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: I chose the "right" way ;) ... at least that is what i hope

TestCases/incomp_navierstokes/streamwise_periodic/README.md
```diff
@@ -0,0 +1,29 @@
+# Streamwise Periodicity testcases
+
+This folder contains the additional Testcases for streamwise periodic flow.
+A Tutorial can be found on the SU2 website.
+For all Testcases a gmsh .geo file is provided which allows to recreate/modify the mesh.
```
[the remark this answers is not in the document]
TobiKattmann: 😊 Thanks. I think I did myself a favour there. Especially now when addressing your comments I would build && runMyTestcases (which i made a separate file for) such that I directly knew if I got it right and because of the short time intervals it was clear what and where I broke s.th.. Feels a lot better to work like that when one has a soft&cozy safety net. I guess the next evolution step is TDD

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7456,6 +7456,99 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
...
+  /*--- absolute value is arbitrary anyway, but it assures that the solution does not change---*/
+  /*--- with a higher number of ranks. If the periodic markers are a line\plane and the     ---*/
+  /*--- streamwise coordinate vector is perpendicular to that |--->|, the choice of the     ---*/
+  /*--- reference node is not relevant at all. This is probably true for most streamwise    ---*/
+  /*--- periodic cases. Other cases where it is relevant could look like this (--->( or     ---*/
+  /*--- \--->\ . The chosen metric is the minimal distance to the origin.                   ---*/
+  /*-------------------------------------------------------------------------------------------*/
+
+  /*--- Initialize/Allocate variables. ---*/
+  su2double min_norm = numeric_limits<su2double>::max();
```
pcarruscag, line 7476: Ah! no way, they've defined numeric_limits for the AD types, those guys rock.

SU2_CFD/include/numerics/flow/flow_sources.hpp
TobiKattmann: Yeah, and I think I changed that back and forth a couple of times ... I can change that to a more consistent style if that aligns better w/ sth else

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -118,6 +119,17 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, INCOMPRESSI
...
   void Explicit_Iteration(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iRKStep);
 
+  /*!
+   * \brief Compute necessary quantities (massflow, integrated heatflux, avg density)
+   *        for streamwise periodic cases. Also sets new delta P for prescribed massflow.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] iMesh - current mesh level for the multigrid.
+   */
+  void GetStreamwise_Periodic_Properties(const CGeometry *geometry,
```
pcarruscag, line 129: Why is this one in Euler and the other in NS?

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -90,7 +91,7 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
 
   Primitive.resize(nPoint,nPrimVar) = su2double(0.0);
 
-  /*--- Incompressible flow, gradients primitive variables nDim+4, (P, vx, vy, vz, T, rho, beta) ---*/
+  /*--- Incompressible flow, gradients primitive variables nDim+6, (P, vx, vy, vz, T, rho, beta, lamMu, EddyMu) ---*/
```
pcarruscag, line 94: insert emoji with monocle

pcarruscag (approved):

TobiKattmann:
Thanks a lot on the comments so far @pcarruscag . Helps quite a bit 👍 
I think I addressed all your comments by now (using AuxVarGrad, a bunch of openMP instructions, overhauling the FindUnique_RefNode func, making a struct for the streamwise periodic values and using that in Numerics and the solver, not using config as a solver var container, and some other things )
Feel encouraged to comment/request-changes on more things.

SU2_CFD/src/variables/CIncEulerVariable.cpp
TobiKattmann: 🧐 there you go

SU2_CFD/include/numerics/flow/flow_sources.hpp
pcarruscag: Nah... I was just messing, if there is a clear style in the file I'm working on I try to follow it, otherwise... FREEEEDOMMMM

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: I hope it is too...

SU2_CFD/include/solvers/CIncEulerSolver.hpp
TobiKattmann: hmm, yeah, moved it to be a NS function

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: \o/

pcarruscag (approved):
I would not have written / documented / tested better myself.
I'll aprove it again for good measure.

SU2_CFD/src/variables/CIncEulerVariable.cpp
pcarruscag: Where the heck do you get them from?

TobiKattmann: `:monocle_face`

TobiKattmann: most of the time I just put the `:` and type want I want to have (crazy, right :) ) those github emoji cheatsheets are often incomplete

pcarruscag: :facepalm: here I was just trying to scroll down after typing :

pcarruscag:
All done? Is CodeFactor not letting you click the magic button?

TobiKattmann:
There we go :) 
The website stuff is no merged as well, so that can go live with the next master release.
Thanks again.

PR #774, opened 2019-08-26
Discrete adjoint multizone driver

oleburghardt:
## Proposed Changes
This PR introduces the discrete adjoint counterpart for the multizone driver. It also comes with some heat solver-related additions for the new input/output structure as CHT (conjugate heat transfer) cases are the primary application so far.

The new driver adapts many of the well-known routines for registering variables, recording, initializing and extracting adjoints. The main difference is the handling of multiple zones. 

The adjoint (fixed point-) iteration approach in `CDiscAdjMultizoneDriver` is to initialize them per-zone in an outer loop and to extract adjoints with respect to _dependent variables from all zones_, thereby inherently capturing the cross dependencies.
Multiple successive evaluations of adjoints with respect to the same zone (equivalent to inner iterations in the multizone driver) are possible.
 
In order not to have to re-record a tape in the (outer) zone-loop as it can be quite costly, this branch includes some new functionalities in the underlying AD-structure which make it possible to record a "full" tape that holds all derivative information and that can be accessed flexibly for different input and output variables across zones.
This also overcomes the restriction that adjoints have to be evaluated in the same order as their variables had been registered.
  
This is the first version of the driver and my intention is to continue contributing to it in terms of both performance and coverage of different kinds of physical couplings in SU2.

## PR Checklist
A regression test for a heated cylinder array with conjugate heat transfer is added to `TestCases/coupled_cht/disc_adj_incomp_2d`.

The very same case is used in a tutorial showcasing the new driver and CHT sensitivity analysis, see [Heated cylinders test case](https://github.com/su2code/su2code.github.io/pull/14).

I'll check carefully for compiler warnings during test phase.


pcarruscag:
I think you mean to merge this into develop right? ;)

oleburghardt:
> I think you mean to merge this into develop right? ;)

:D Thanks.

pcarruscag:
@oleburghardt can you create a branch with the open PR's you have already merged in so we can compare this branch against its true base? That you make the review easier.

oleburghardt:
> @oleburghardt can you create a branch with the open PR's you have already merged in so we can compare this branch against its true base? That would make the review easier.

It's true base is feature_input_output from #724, so one can quickly get an overview [here](https://github.com/su2code/SU2/compare/feature_input_output...sc_develop). It's the only one and I am constantly merging updates from there.

Common/src/config_structure.cpp
```diff
@@ -2256,26 +2206,37 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Determines if the multizone problem is solved for time-domain. */
   addBoolOption("TIME_DOMAIN", Time_Domain, false);
   /* DESCRIPTION: Number of outer iterations in the multizone problem. */
-  addUnsignedLongOption("OUTER_ITER", Outer_Iter, 1);
+  addUnsignedLongOption("OUTER_ITER", nOuterIter, 1);
   /* DESCRIPTION: Number of inner iterations in each multizone block. */
-  addUnsignedLongOption("INNER_ITER", Inner_Iter, 1);
+  addUnsignedLongOption("INNER_ITER", nInnerIter, 1);
+  /* DESCRIPTION: Number of adjoint inner iterations in each multizone block. */
+  addUnsignedLongOption("INNER_ITER_ADJOINT", nInnerIter_Adjoint, 1);
```
pcarruscag: Can you re-use the "normal" INNER_ITER?

Common/src/dual_grid_structure.cpp
```diff
@@ -424,6 +446,27 @@ void CPoint::SetBoundary(unsigned short val_nmarker) {
...
+    if(input) {
+      AD::Set_AdjIndex(Input_AdjIndices[iDim], Coord[iDim]);
+    }
+    else {
+      AD::Set_AdjIndex(Output_AdjIndices[iDim], Coord[iDim]);
+    }
+  }
+}
+
+void CPoint::SetAdjointSolution(su2double *adj_sol) {
```
pcarruscag: make the argument const

Common/src/dual_grid_structure.cpp
```diff
@@ -424,6 +446,27 @@ void CPoint::SetBoundary(unsigned short val_nmarker) {
...
+  for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+    AD::SetDerivative(Output_AdjIndices[iDim], SU2_TYPE::GetValue(adj_sol[iDim]));
+  }
+}
+
+void CPoint::GetAdjointSolution(su2double *adj_sol, unsigned short iDim) {
```
pcarruscag: please return by value, this looks weird (and what happens if people pass NULL?)

Common/include/linear_algebra/CSysSolve.hpp
```diff
@@ -259,7 +259,7 @@ class CSysSolve {
    */
   unsigned long BCGSTAB_LinSolver(const VectorType & b, VectorType & x, ProductType & mat_vec,
                                   PrecondType & precond, ScalarType tol, unsigned long m,
-                                  ScalarType *residual, bool monitoring, CConfig *config);
+                                  ScalarType *residual, bool monitoring, CConfig *config, bool TapeActive);
```
pcarruscag: Please take this off the function signature, AD:: is also available inside the method

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -241,7 +242,7 @@ unsigned long CSysSolve<ScalarType>::CG_LinSolver(const CSysVector<ScalarType> &
 
     norm_r = r.norm();
     norm0  = b.norm();
-    if ( (norm_r < tol*norm0) || (norm_r < eps) ) {
+    if ( ((norm_r < tol*norm0) || (norm_r < eps)) && !TapeActive ) {
```
pcarruscag: Are you recording the linear solver iterations to have to do this? :S

Common/src/config_structure.cpp
oleburghardt: Yes. The advantage would be that it's more elegant in the code. This disadvantage being that one cannot elegantly change `DIRECT` to `DISCRETE_ADJOINT` to compute adjoints but that one has to go though all config files and adapt `INNER_ITER`. 
To me it's a draw. No problem if you still would like to have it removed.

Common/src/linear_algebra/CSysSolve.cpp
oleburghardt: No, by no means. If we recording though, we don't want to return at this point in case the primal solution is fully converged, but we still want the solution update being included.
It's a fix that's not multizone-specific.

Common/src/config_structure.cpp
pcarruscag: That is what I would like to avoid, having to change more than one option when switching from DIRECT to ADJOINT, I think others would agree since there has been so much work in making Outer_Iter, Time_Iter, etc. consistent between single and multizone.

Common/src/linear_algebra/CSysSolve.cpp
pcarruscag: It is in CSysSolve::Solve.
The legacy grid movement calls the individual methods directly but that does not matter because it does the adjoint analytically (symmetric matrix).

Common/src/config_structure.cpp
oleburghardt: Ok. I'll change that.
But let's just think what can be done to avoid touching all config files (think, e.g., of a case with not two but quite a few zones) whenever ones switches from `DIRECT` to `DISCRETE_ADJOINT`.

pcarruscag: Old up, maybe I got something wrong as it feels we are saying the same thing with opposite meanings.
For a direct multizone problem INNER_ITER is how many (e.g.) fluid iterations are performed per coupling iteration, right?
What is the meaning of INNER_ITER_ADJOINT?

Common/src/linear_algebra/CSysSolve.cpp
oleburghardt: Alright.

Common/src/config_structure.cpp
oleburghardt: Right. Same for the adjoint, that is how many discrete adjoint (flow) solution iterations are performed before moving on to the next zone (i.e. the diagonal part, given the coupling contributions from a previous iteration).
They don't have to be necessarily the same, I'm lacking some experience here. (CHT problems are very stable in that regard.)

pcarruscag: Ok that's what I thought, so the testcases that would be problematic to change are your own existing ones?
I am thinking that from the user's perspective it would be annoying to have two sets of options in each file for the "same" thing (we have it for the linear solver though...)
We can wait for other opinions.

oleburghardt: My test cases are not problematic in the sense that I'm playing around with a bunch of different options anyway. Let's just go with `INNER_ITER` as we have at least one preference. Changes can be made quickly.

rsanfer: I agree with @pcarruscag here, given that we have INNER_ITER, I don't see a reason for duplicating the option for the adjoint. Specially since the direct iterations to be recorded are always 1 on the adjoint solution.

Common/src/dual_grid_structure.cpp
oleburghardt: Ok.

Common/src/dual_grid_structure.cpp
oleburghardt: No idea how that happened. Thanks

SU2_CFD/include/solver_structure.hpp
```diff
@@ -509,7 +528,51 @@ class CSolver {
...
+   * \brief Add the current Solution vector to External.
+   * \param[in] geometry - The geometrical definition of the problem.
+   */
+  void AddExternal_Solution(CGeometry *geometry);
+
+  /*!
+   * \brief Add the current Solution vector to External_Old.
+   * \param[in] geometry - The geometrical definition of the problem.
+   */
+  void AddExternalOld_Solution(CGeometry *geometry);
```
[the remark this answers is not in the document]
oleburghardt: That's the one (`Set_OldExternal`) that does not fit this structure. I wanted to have the same naming as for `Solution`.
Concerning the underscores in all the other routines: I wanted to give them at least some sense (the variable preceding the underscore being the one that is affected, the variable following the underscore being the input value). They are often used quite randomly or there's a different rule that I haven't discovered yet :)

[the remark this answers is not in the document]
pcarruscag: I've not noticed a pattern yet either. That is why I suggested stating everything (operation Add/Set, source, and destination) in the name. I get your convention but since like you say other methods do not seem to follow one, the best option might be to spell everything.
It may avoid future situation where someone comes along and creates a redundant method (that seems to have happened in CVariable).

SU2_CFD/include/variables/CDiscAdjVariable.hpp
```diff
@@ -158,6 +157,49 @@ class CDiscAdjVariable : public CVariable {
...
+    for (unsigned short iVar = 0; iVar < nVar; iVar++) {
+      External[iVar] = 0.0;
+    }
+  }
+
+  /*!
+   * \brief Add a value to the External vector.
+   * \param[in] val_sol - vector that has to be added component-wise
+   */
+  inline void Add_External(su2double* val_sol) {
```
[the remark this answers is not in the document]
oleburghardt: Ok. And sure, feel free to add some that I might overlook.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,1012 @@
...
+    bool output_files = false;
+    
+    /*--- Determine whether a solution needs to be written
+   after the current iteration ---*/
+    
+    if (
+        
+        /*--- General if statements to print output statements ---*/
+        
+        (TimeIter+1 >= config_container[iZone]->GetnTime_Iter()) || (StopCalc) ||
```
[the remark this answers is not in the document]
oleburghardt: Yes, it's a copy of `CMultizoneDriver::Output` with a single line changed. Though I just saw that this overload is not needed anymore. I'll remove it.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,1012 @@
...
+     *    since it does not change the computational effort. ---*/
+
+
+    /*--- If we want to set up zone-specific tapes later on,
+     *    we just record the objective function section here.
+     *    If not, the whole tape of a coupled run will be created. ---*/
+
+    if (retape) {
+      SetRecording(NONE, FULL_TAPE, ZONE_0);
+      SetRecording(FLOW_CONS_VARS, OBJECTIVE_FUNCTION_TAPE, ZONE_0);
```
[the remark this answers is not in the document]
oleburghardt: Yes. One could think of changing from `FLOW_CONS_VARS` to `CONS_VARS`.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,1012 @@
...
+
+      case DISC_ADJ_EULER: case DISC_ADJ_NAVIER_STOKES: case DISC_ADJ_RANS:
+      case DISC_ADJ_INC_EULER: case DISC_ADJ_INC_NAVIER_STOKES: case DISC_ADJ_INC_RANS:
+
+        switch(config_container[iZone]->GetKind_ObjFunc()) {
+
+          // Aerodynamic coefficients
+
+          case DRAG_COEFFICIENT:
+            ObjFunc+=direct_output[iZone]->GetHistoryFieldValue("DRAG")*Weight_ObjFunc;
```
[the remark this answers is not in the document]
oleburghardt: I didn't consider the weighted objective as this will in fact be moved to the new output classes (you are referring to the `Total_ComboObj` variable that is part of the solvers, right?) some day.
This driver-level routine just sums up contributions from different zones.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,1012 @@
...
+      }
+    }
+
+    /*--- For the adjoint iteration we need the derivatives of the iteration function with
+     *    respect to the state (and possibly the mesh coordinate) variables.
+     *    Since these derivatives do not change in the steady state case we only have to record
+     *    if the current recording is different from them.
+     *
+     *    To set the tape appropriately, the following recording methods are provided:
+     *    (1) NONE: All information from a previous recording is removed.
```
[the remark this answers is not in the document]
oleburghardt: One could add it to `ENUM_RECORDING` in the config header (?). I could do it just partially though. Could you then add those for `FEA_DISP_VARS`, ... ? I think they just determine whether what is initialized in `SetAdjoint_Output` and what is extracted in `ExtractAdjoint_Solution` but I'm not 100% sure about them.

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -0,0 +1,158 @@
...
+#include "CMultizoneDriver.hpp"
+
+class CDiscAdjMultizoneDriver : public CMultizoneDriver {
+protected:
+
+  bool retape;                        /*!< \brief Boolean whether derivative information for all zones is kept in memory.*/
+  unsigned short RecordingState;      /*!< \brief The kind of recording the tape currently holds.*/
+  su2double ObjFunc;                  /*!< \brief The value of the objective function.*/
+  int ObjFunc_Index;                  /*!< \brief Index of the value of the objective function.*/
+  unsigned short* direct_nInst;       /*!< \brief Total number of instances in the direct problem (per zone). */
```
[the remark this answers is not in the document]
oleburghardt: Why that?

SU2_CFD/include/solver_structure.hpp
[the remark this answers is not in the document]
oleburghardt: I made some names more clear and kept some for which there is a de facto convention.

Common/include/config_structure.hpp
```diff
@@ -8575,6 +8410,24 @@ class CConfig {
...
+  * \brief Get the indicator whether we want to use full (coupled) tapes.
+  * \return the full tape indicator.
+  */
+  bool GetFull_Tape(void);
+
+  /*!
+   * \brief Set the indicator whether we are using the multiphysics discrete adjoint driver.
+   * \return the multiphysics discrete adjoint indicator.
+   */
+  void SetMultiphysicsDiscrete_Adjoint(bool multi_disc_adj);
```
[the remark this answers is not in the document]
oleburghardt: Thanks, it's fixed.

pcarruscag:
Thanks for making the changes @oleburghardt , it is a bit easier to review and comment on this PR now that the output is merged, I will have another look by the end of the week and update my review.

oleburghardt:
> Thanks for making the changes @oleburghardt , it is a bit easier to review and comment on this PR now that the output is merged, I will have another look by the end of the week and update my review.

Yeah, merging with develop was such a relief a couple of minutes ago :D Thanks for your effort! No hurry.

Common/include/ad_structure.hpp
```diff
@@ -209,6 +240,17 @@ namespace AD{
    */
   void EndExtFunc();
 
+  /*!
+   * \brief Evaluates and saves gradient data from a variable.
+   * \param[in] data - where the information will be extracted from.
+   * \param[in] index - where obtained gradient data will be stored.
```
pcarruscag: It seems index is an output

Common/src/config_structure.cpp
```diff
@@ -5140,16 +5147,16 @@ void CConfig::SetMarkers(unsigned short val_software) {
...
-  for (iMarker_DV = 0; iMarker_DV < nMarker_DV; iMarker_DV++) {
-    for (iMarker_CfgFile = 0; iMarker_CfgFile < nMarker_CfgFile; iMarker_CfgFile++) {
-      if (Marker_CfgFile_TagBound[iMarker_CfgFile] == Marker_DV[iMarker_DV])
-        markerCount++;
-    }
-  }
-  if ((nMarker_DV > 0) && (markerCount != nMarker_DV)) {
-    SU2_MPI::Error("DV_MARKER contains marker names that do not exist in the lists of BCs in the config file.", CURRENT_FUNCTION);
-  }
+//  unsigned short markerCount = 0;
```
pcarruscag: If there is a strong reason to leave this code commented out please document it in the code, otherwise delete it.

Common/src/config_structure.cpp
```diff
@@ -6237,7 +6244,8 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
 
     cout << endl <<"------------------ Convergence Criteria  ( Zone "  << iZone << " ) ---------------------" << endl;
 
-    
+    cout << "Maximum number of solver subiterations: " << nInnerIter <<"."<< endl;
```
pcarruscag: maybe a merge issue here, it looks like this line is going to be duplicated?

Common/src/geometry_structure.cpp
```diff
@@ -2695,8 +2705,17 @@ void CGeometry::RegisterOutput_Coordinates(CConfig *config){
   unsigned long iPoint;
 
   for (iPoint = 0; iPoint < nPoint; iPoint++){
-    for (iDim = 0; iDim < nDim; iDim++){
-      AD::RegisterOutput(node[iPoint]->GetCoord()[iDim]);
+    if(config->GetMultiphysicsDiscrete_Adjoint()) {
+      for (iDim = 0; iDim < nDim; iDim++) {
+//        AD::RegisterOutput(node[iPoint]->GetCoord()[iDim]);
```
pcarruscag: again if possible don't leave code commented out

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -0,0 +1,183 @@
...
+#include "CMultizoneDriver.hpp"
+
+class CDiscAdjMultizoneDriver : public CMultizoneDriver {
+
+protected:
+
+  /*!
+   * \brief Kinds of recordings (three different ones).
+   */
+  enum ENUM_TAPE {
```
pcarruscag: please consider using `enum class` if you don't need to rely on implicit conversion to integer type

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -0,0 +1,183 @@
...
+    REGISTERED = 1,               /*!< \brief Solver variables are registered on the tape. */
+    DEPENDENCIES = 2,             /*!< \brief Derived values (e.g. gradients) are set. */
+    OBJECTIVE_FUNCTION = 3,       /*!< \brief Objective function is set. */
+    TRANSFER = 4,                 /*!< \brief Solution data is transferred between coupled solvers of
+                                              different physical zones (giving cross contributions). */
+    ITERATION_READY = 5           /*!< \brief Until here, all derivative information is gathered so
+                                              that it can be connected to a solver update evaluation. */
+  };
+
+  unsigned short RecordingState;  /*!< \brief The kind of recording that the tape currently holds. */
```
pcarruscag: can you use the right enum type instead of unsigned short here? I think it improves readability.

SU2_CFD/include/output/CHeatOutput.hpp
```diff
@@ -78,11 +84,18 @@ class CHeatOutput final: public COutput {
...
    */
-  void LoadVolumeData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint) override; 
-  
+  void LoadVolumeData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint) override;   
+
   /*!
-   * \brief Set the available history output fields
-   * \param[in] config - Definition of the particular problem.
+   * \brief LoadSurfaceData
+   * \param config
```
pcarruscag: thanks for keeping the parameters documented, I know in this case it is obvious but the param type [in] is missing

SU2_CFD/include/solver_structure.hpp
```diff
@@ -514,7 +528,51 @@ class CSolver {
...
+   * \brief Add the current Solution vector to External.
+   * \param[in] geometry - The geometrical definition of the problem.
+   */
+  void Add_Solution_To_External(CGeometry *geometry);
+
+  /*!
+   * \brief Add the current Solution vector to External_Old.
+   * \param[in] geometry - The geometrical definition of the problem.
+   */
+  void Add_Solution_To_ExternalOld(CGeometry *geometry);
```
pcarruscag: thanks for the name changes I think they are clearer

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -92,6 +94,11 @@ class CVariable {
                                                        note that this variable cannnot be static, it is possible to
                                                        have different number of nVar in the same problem. */
   su2double *Solution_Adj_Old;    /*!< \brief Solution of the problem in the previous AD-BGS iteration. */
+  
+  int *Input_AdjIndices,          /*!< \brief Indices of Solution variables in the adjoint vector. */
```
pcarruscag: Just for curiosity, we usually use unsigned long for large indices, is int the native type of CoDi? if so how large of a tape does that support per mpi rank?

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
pcarruscag: I will probably need some help to add the FEA stuff, we'll see

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
pcarruscag: I think that would make sense, otherwise it is a bit confusing.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,874 @@
...
+
+    for(iZone = 0; iZone < nZone; iZone++) {
+
+      AD::Push_TapePosition();
+
+      if (ZONE_SPECIFIC_TAPE && (iZone == record_zone)) {
+        DirectIteration(iZone, kind_recording);
+      }
+      else {
+        DirectIteration(iZone, kind_recording);
```
pcarruscag: It looks like the "if" and the "else" do the same thing?

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
pcarruscag: I see... I know it is not an easy problem to solve, I thought about it once and gave up. But since this will be moved to output classes would it make sense to start moving this logic there?

SU2_CFD/src/output_adjoint_heat.cpp
```diff
@@ -0,0 +1,206 @@
...
+  turb_model = config->GetKind_Turb_Model();
+  
+  heat = config->GetEnergy_Equation();
+  
+  weakly_coupled_heat = config->GetWeakly_Coupled_Heat();
+  
+  /*--- Set the default history fields if nothing is set in the config file ---*/
+  
+  if (nRequestedHistoryFields == 0){
+    RequestedHistoryFields.push_back("ITER");
```
pcarruscag: please use emplace_back here to be consistent with the rest of the output

SU2_CFD/src/solver_adjoint_discrete.cpp
```diff
@@ -1041,39 +1094,39 @@ void CDiscAdjSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfi
...
-          residual = node[iPoint]->Get_BGSSolution(iVar) - node[iPoint]->Get_BGSSolution_k(iVar);
+//  for (iPoint = 0; iPoint < nPointDomain; iPoint++){
+//      for (iVar = 0; iVar < nVar; iVar++){
+//          residual = node[iPoint]->Get_BGSSolution(iVar) - node[iPoint]->Get_BGSSolution_k(iVar);
 
-          AddRes_BGS(iVar,residual*residual);
-          AddRes_Max_BGS(iVar,fabs(residual),geometry->node[iPoint]->GetGlobalIndex(),geometry->node[iPoint]->GetCoord());
-      }
-  }
+//          AddRes_BGS(iVar,residual*residual);
```
pcarruscag: same comment about commented out code

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2303,9 +2303,9 @@ void CDiscAdjFluidIteration::InitializeAdjoint(CSolver *****solver, CGeometry **
   bool heat = config[iZone]->GetWeakly_Coupled_Heat();
   bool interface_boundary = (config[iZone]->GetnMarker_Fluid_Load() > 0);
 
-  /*--- Initialize the adjoint of the objective function (typically with 1.0) ---*/
+//  /*--- Initialize the adjoint of the objective function (typically with 1.0) ---*/
```
WallyMaier: why did this get commented?

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2303,9 +2303,9 @@ void CDiscAdjFluidIteration::InitializeAdjoint(CSolver *****solver, CGeometry **
...
   
-  solver[iZone][iInst][MESH_0][ADJFLOW_SOL]->SetAdj_ObjFunc(geometry[iZone][iInst][MESH_0], config[iZone]);
+//  solver[iZone][iInst][MESH_0][ADJFLOW_SOL]->SetAdj_ObjFunc(geometry[iZone][iInst][MESH_0], config[iZone]);
```
WallyMaier: Is this no longer used?

Common/include/ad_structure.hpp
oleburghardt: It's a reference to where we store the gradient information.. depends on the way you look at it I guess (?)

Common/src/config_structure.cpp
oleburghardt: That was probably something I commented out in this branch because it was dependent on others. Just checked that it can be undone. Thanks for spotting.

Common/src/geometry_structure.cpp
oleburghardt: Ok. One would have to add those lines to set mesh adjoints. I thought this will be needed for FSI integration at some point but things seem to work out a little different now.
Should I add a comment and we will then uncomment or delete it in the course of integration of FSI into this driver?

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
oleburghardt: Sure :)

SU2_CFD/include/output/CHeatOutput.hpp
oleburghardt: Added.

SU2_CFD/include/variables/CVariable.hpp
oleburghardt: I think it is but I'll make sure.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
oleburghardt: Sure. I guess I'll make up a small plan of everything that needs to be done (especially as there are now even two new discrete adjoint solvers, one has to do all the adaptions for those as well) and then we go through step by step. I wanted to send a mail regarding the theory anyway soon.

Thanks for finding the logic error. I'm sure I had fixed that somewhere already .. maybe on the HPC repository clone. Some stuff in between gets lost sometimes :p

SU2_CFD/src/solver_adjoint_discrete.cpp
oleburghardt: Oh! That commented out code is an issue.
I commented it out because of its design flaw to compute solutions where it should only compute residuals (which caused some problems). So I used the base class implementation this way.
I'll revert it and add `if` which I don't like at all but I think there's no way around.

SU2_CFD/src/output_adjoint_heat.cpp
oleburghardt: `SU2_CFD/src/output_adjoint_heat.cpp` for some reason was still traced by git, it's old and removed now.

SU2_CFD/src/solver_adjoint_discrete.cpp
pcarruscag: But if the code works with this commented out, and it has a design flaw, I would say remove it entirely.

Common/src/geometry_structure.cpp
pcarruscag: Yup leave a comment then

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
oleburghardt: Thanks for the hint!
I changed it for the one where I don't want this conversion, it's `enum class Kind_Tape` now. For the other one i need it.. so should one try `enum class Tape_Positions : int { ... } ` then?

Common/include/ad_structure.hpp
pcarruscag: the value of index gets set as a result of calling this function that's why it seems to be an output

SU2_CFD/include/solver_structure.hpp
```diff
@@ -12785,7 +12843,7 @@ class CDiscAdjSolver : public CSolver {
    */
   void LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig *config, int val_iter, bool val_update_geo);
   
-  void ComputeResidual_Multizone(CGeometry *geometry, CConfig *config);
+//  void ComputeResidual_Multizone(CGeometry *geometry, CConfig *config);
```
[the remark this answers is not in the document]
oleburghardt: It was due to a problem I had during the implementation of the upper level of the driver. I've completely forgotten about this (no test case to protect it) but it's solved by an extra `if` in that routine now.

SU2_CFD/src/iteration_structure.cpp
oleburghardt: Just because it's in an iteration class whereas it should rather be done on the driver level. Though that's the case anyway (might have been just done redundantly in the past) and I'm almost sure it can be removed. So it's commented out for a while if that makes sense.. O:) Can you confirm, @talbring ?

SU2_CFD/src/iteration_structure.cpp
oleburghardt: The functionality is covered on the driver level by e.g. `CDiscAdjSinglezoneDriver::SetAdj_ObjFunction()`. I think `CDiscAdjSolver::SetAdj_ObjFunc` is just used by `CDiscAdjFSIDriver`. I'll delete the commented lines in a distinct commit to avoid confusion about it.

oleburghardt:
@pcarruscag @WallyMaier 
I think I've adressed most of what you spotted in the latest commits. Thanks for taking so much care - you already helped me a lot getting everything ready to enter the main version, especially concerning the just-for-development adaptions that I've forgotten about.

@TobiKattmann 
I sneaked in another commit for primal CHT that _might_ help with your issue. I'm still running tests that might confirm my guesswork. (Though you can already try it by setting `CHT_ROBIN= NO` which will then just directly apply heat fluxes.) I'm coming back to this once I'm sure about it.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,876 @@
...
+      }
+
+      /*--- Evaluate the objective function gradient w.rt. to solution contributions from iZone.
+       *    (We always initialize from Solution_BGS_k and extract to Solution.) ---*/
+
+      AD::ClearAdjoints();
+
+      SetAdj_ObjFunction();
+
+      AD::ComputeAdjoint(3,0);
```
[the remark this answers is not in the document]
oleburghardt: I even added that enum already but then didn't use it there, thanks for the hint!

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,876 @@
...
+      iteration_container[iZone][INST_0]->RegisterOutput(solver_container, geometry_container,
+                                                         config_container, output_container[iZone], iZone, INST_0);
+
+      AD::Push_TapePosition();
+    }
+  }
+
+  if (rank == MASTER_NODE && kind_recording == FLOW_CONS_VARS) {
+
+    //  AD::PrintStatistics();
```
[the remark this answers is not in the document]
oleburghardt: Yes, done.

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+  /*!
+   * \brief Run an discrete adjoint update of all solvers within multiple zones.
+   */
+  void Run();
+
+//  /*!
+//   * \brief Check the convergence at the outer level.
+//   */
+//  bool OuterConvergence(unsigned long OuterIter);
```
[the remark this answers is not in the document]
oleburghardt: No, it's removed.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,876 @@
...
+                                                      surface_movement, grid_movement, FFDBox, jZone, INST_0);
+
+          /*--- Add the cross derivatives from iZone<-jZone dependencies to the External vector. ---*/
+
+          Add_Solution_To_External(jZone);
+
+        }
+      }
+
+      // TODO: update the off-diagonals here
```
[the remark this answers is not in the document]
oleburghardt: No, not important, just left as a note that the framework allows for updating off-diagonals there (already within the loop over zones - resembling the idea of Gauss-Seidel-like methods to treat diagonals and off-diagonals independently). I'll make the comment more precise.

SU2_CFD/src/SU2_CFD.cpp
```diff
@@ -118,17 +118,30 @@ int main(int argc, char *argv[]) {
       if (nZone > 1 ) {
         SU2_MPI::Error("The required solver doesn't support multizone simulations", CURRENT_FUNCTION);
       }
-      if (config->GetDiscrete_Adjoint())
+      if (config->GetDiscrete_Adjoint()) {
+        if (config->GetMultiphysicsDiscrete_Adjoint())
```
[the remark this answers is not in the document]
oleburghardt: > I am guessing these special cases will be removed once all of the drivers have settled into place?

They definitely could be removed, even in this PR (the `if (config->GetMultiphysicsDiscrete_Adjoint())` in fact just determines which AD-approach is used - right now that's only the case for `CDiscAdjMultizoneDriver`. 
However it feels to me that this would mean to many changes at once on such a basic level.
It would also require some performance tests, as the current one for single zone problems should be faster as there is just one `std::vector` instead of a large amount of dereferencings down to `CVariable`. It's very likely that this improves a lot with #753

> Which leads to the larger question: do you expect that all multiphysics problems can now be adapted to this driver?

Yes that's the overall idea behind coming up with this generic framework instead of a discrete adjoint CHT driver. Though a lot of work still has to be done to cover all with one and the same driver.
Fluid-fluid interface problems should be an easy round, FSI will be a joint work with @pcarruscag 

> Lots of potential

:+1:

SU2_CFD/src/solver_direct_heat.cpp
```diff
@@ -1238,11 +1264,10 @@ void CHeatSolverFVM::Heat_Fluxes(CGeometry *geometry, CSolver **solver_container
             thermal_conductivity = config->GetThermalDiffusivity_Solid()*rho_cp_solid;
           }
 
-          HeatFlux[iMarker][iVertex] = thermal_conductivity*dTdn*config->GetTemperature_Ref()*Area;
+          HeatFlux[iMarker][iVertex] = thermal_conductivity*dTdn; // TDE *config->GetHeat_Flux_Ref();
```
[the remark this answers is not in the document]
oleburghardt: I'll check. Do we agree that heat fluxes should be passed between zones in units of W/m² ? (It would make sense I think as different zones could eventually use different reference values?)

oleburghardt:
Thanks @economon for the latest round of remarks.

I think there are just few commits to be made here (though maybe one or two for primal CHT stuff that somehow fit in). Let me know if you like some other things to be changed!

Common/src/config_structure.cpp
```diff
@@ -874,6 +874,10 @@ void CConfig::SetConfig_Options() {
   addEnumOption("MULTIZONE_SOLVER", Kind_MZSolver, Multizone_Map, MZ_BLOCK_GAUSS_SEIDEL);
   /*!\brief MATH_PROBLEM  \n DESCRIPTION: Mathematical problem \n  Options: DIRECT, ADJOINT \ingroup Config*/
   addMathProblemOption("MATH_PROBLEM", ContinuousAdjoint, false, DiscreteAdjoint, false, Restart_Flow, false);
+  /*!\brief MULTIPHYSICS_DISCRETE_ADJOINT  \n DESCRIPTION: Boolean whether multiphysics adjoint shall be used \n Options: NO, YES \ingroup Config */
+  addBoolOption("MULTIPHYSICS_DISCRETE_ADJOINT", MultiphysicsDiscreteAdjoint, false);
```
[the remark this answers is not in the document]
oleburghardt: No, one doesn't, it would have been just a developer's option to use the discrete adjoint multiphysics driver for single zone cases. I removed it completely probably exactly when you wrote that comment as now a multiphysics setting with only one specified solver will also do (see latest commits above).

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,880 @@
...
+  bool            checkSensitivity  = false;
+  unsigned short  jZone             = 0,
+                  wrt_sol_freq      = config_container[ZONE_0]->GetVolume_Wrt_Freq();
+  unsigned long   iIter             = 0,
+                  nIter             = 0,
+                  iOuter_Iter       = 0,
+                  iInner_Iter       = 0;
+
+  if (nZone == 1) {
+    nIter = driver_config->GetnIter();
```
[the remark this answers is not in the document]
oleburghardt: So since yesterday, the discrete adjoint multizone driver will be launched whenever we have a driver/master config file with `SOLVER= MULTIPHYSICS` and `MATH_PROBLEM = DISCRETE_ADJOINT`, regardless of the number of zones. Which totally makes sense.

Also the marked lines above are removed as they are also not needed anymore. `CDiscAdjMultizoneDriver` will run the specified number of outer iteration in the outer loop (...) :)

I run quite some tests to be sure that both (single zone driver and multizone driver for nZone=1) give same adjoints and sensitivities.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,880 @@
...
+  AD::Push_TapePosition();
+
+  if (tape_type != Kind_Tape::OBJECTIVE_FUNCTION_TAPE) {
+
+    /*--- We do the communication here to not derive wrt updated boundary data. ---*/
+
+    for(iZone = 0; iZone < nZone; iZone++) {
+
+      /*--- In principle, the mesh does not need to be updated ---*/
+      UpdateMesh = 0;
```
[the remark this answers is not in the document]
oleburghardt: Please see recent commits, I think you are pointing to a line that has already been updated.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,880 @@
...
+
+    direct_output[iZone]->SetHistory_Output(geometry_container[iZone][INST_0][MESH_0],
+                                            solver_container[iZone][INST_0][MESH_0], config_container[iZone]);
+  }
+
+  /*--- Extract objective function values. ---*/
+
+  for (iZone = 0; iZone < nZone; iZone++){
+
+    nMarker_Analyze = config_container[iZone]->GetnMarker_Monitoring();
```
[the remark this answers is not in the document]
oleburghardt: Thanks!

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,880 @@
...
+            ObjFunc+=direct_output[iZone]->GetHistoryFieldValue("SECONDARY_OVER_UNIFORMITY")*Weight_ObjFunc;
+            break;
+          case TOTAL_AVG_TEMPERATURE:
+            ObjFunc+=direct_output[iZone]->GetHistoryFieldValue("AVG_TOTALTEMP")*Weight_ObjFunc;
+            break;
+          case SURFACE_TOTAL_PRESSURE:
+            ObjFunc+=direct_output[iZone]->GetHistoryFieldValue("AVG_TOTALPRESS")*Weight_ObjFunc;
+            break;
+
+          // Not yet covered by new output structure. Be careful these use MARKER_MONITORING.
```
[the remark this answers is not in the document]
oleburghardt: Why commenting them out?! Those can be used.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -0,0 +1,880 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "../include/drivers/CDiscAdjMultizoneDriver.hpp"
+
+CDiscAdjMultizoneDriver::CDiscAdjMultizoneDriver(char* confFile,
+                                      unsigned short val_nZone,
+                                      SU2_Comm MPICommunicator) : CMultizoneDriver(confFile,
```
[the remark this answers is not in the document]
oleburghardt: Ok

SU2_CFD/include/output/CHeatOutput.hpp
```diff
@@ -78,11 +84,18 @@ class CHeatOutput final: public COutput {
...
+   * \brief LoadSurfaceData
    * \param[in] config - Definition of the particular problem.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver - The container holding all solution data.
+   * \param[in] iPoint - Index of the point.
+   * \param[in] iMarker - Index of the surface marker.
+   * \param[in] iVertex - Index of the vertex on the marker.
    */
-  void SetHistoryOutputFields(CConfig *config) override;
+  void LoadSurfaceData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint, unsigned short iMarker, unsigned long iVertex);
```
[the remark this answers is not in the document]
oleburghardt: :+1:

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+  /*!
+   * \brief Destructor of the class.
+   */
+  ~CDiscAdjMultizoneDriver(void);
+
+  /*!
+   * \brief [Overload] Launch the computation for discrete adjoint multizone problems.
+   */
+  void StartSolver();
```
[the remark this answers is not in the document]
oleburghardt: > Edit: I see you did at some places already going further in the changed files :)

Oh that wasn't me. @talbring and/or @rsanfer added the heat output classes.

Thanks for the explanation. The safety mechanism argument makes sense to me. I added those identifiers.

SU2_CFD/include/output/CMultizoneOutput.hpp
```diff
@@ -49,6 +49,7 @@
 #include <cmath>
 #include <time.h>
 #include <fstream>
+#include <string>
```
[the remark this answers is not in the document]
oleburghardt: Good question.. removed unless someone complains.. ;)

oleburghardt:
> Ok, I am ~halfway through and I have mostly nitpicking stuff. In general it is really tidy and I am looking forward to use everything you added. Rest will follow tomorrow.

Thanks a lot Tobias! I think I made all the changes that you suggested, it's just the ouput that you and Pedro do not like. I'll think about what can be done .. (though part of it might be due to the fact that the `switch`-thing is more apparent now, same thing can be found in the solvers as well).

> I already did the tutorial you uploaded but I'll redo everything as I couldn't reproduce the nice gradients you show in the tutorial. I'll report to you... maybe I somewhere wrong

Could be please try to run `SU2_DOT_AD` in serial? (If this doesn't help and if you've got the time, also `SU2_CFD_AD`?)
That **really must work** .. it's quite important to me as I invested so much time so make it available for all. Could you send me your adjoint solution files?

SU2_CFD/src/solver_direct_heat.cpp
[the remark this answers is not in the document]
oleburghardt: Just another question concerning the reference heat flux. E.g. regarding this line from the `CIncNSSolver::Friction_Forces` routine, 
`HeatFlux[iMarker][iVertex] = -thermal_conductivity*GradTemperature*RefHeatFlux;` -
do I understand correctly that the idea behind `RefHeatFlux` is to undo all changes of units and magnitudes that might have been applied to `thermal_conductivity` and `GradTemperature` so that we don't have to care about them separately anymore?

TobiKattmann:
> Could be please try to run `SU2_DOT_AD` in serial? (If this doesn't help and if you've got the time, also `SU2_CFD_AD`?)
> That **really must work** .. it's quite important to me as I invested so much time so make it available for all. Could you send me your adjoint solution files?
Yep, Single core for both did the trick and I now get the same gradient results. I will double check with 28 cores to see if it still produces differing results.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -104,95 +93,112 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
   }
-  else {
 
-    thermal_diffusivity     = donor_config->GetThermalDiffusivity_Solid();
-    heat_flux_density       = (thermal_diffusivity*dTdn)*rho_cp_solid;
-    conductivity_over_dist  = thermal_diffusivity*rho_cp_solid/dist;
+  /*--- Set the conjugate heat variables that are transferred by default ---*/
+
+  Donor_Variable[0] = Twall*donor_config->GetTemperature_Ref();
+  Donor_Variable[1] = heat_flux_density*donor_config->GetHeat_Flux_Ref();
```
oleburghardt: @economon Could you look over this and the `Heat_Fluxes` routine in the heat solver to check whether these usages of `GetHeat_Flux_Ref` meet your idea?

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -104,95 +93,112 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
   if (compressible_flow) {
 
-    iPoint = donor_geometry->vertex[Marker_Donor][Vertex_Donor]->GetNode();
+    su2double Gamma         = donor_config->GetGamma();
+    su2double Gas_Constant  = donor_config->GetGas_ConstantND();
+    su2double Cp            = (Gamma / (Gamma - 1.0)) * Gas_Constant;
+
+    Prandtl_Lam             = donor_config->GetPrandtl_Lam();
+    laminar_viscosity       = donor_solution->node[Point_Donor]->GetLaminarViscosity(); // TDE check for consistency
+    Cp                      = (Gamma / (Gamma - 1.0)) * Gas_Constant;
```
oleburghardt: @economon This is now completely analogue to the code in `CNSSolver::Friction_Forces`. Do you agree with it?

SU2_CFD/src/solver_direct_heat.cpp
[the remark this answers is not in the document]
economon: The idea on that line is to re-dimensionalize the value of the heat flux, since we always print it to the output in dimensional format. In the fluid solver, assuming we run in non-dim. mode, we have already applied non-dim. to all intermediate quantities in the code within the SetNondimensionalization() routine (here, thermal_conductivity and Temperature). We also compute the RefHeatFlux there, so that we can undo the non-dim. like we see in that routine.

I think we should do similarly anytime we want to output quantities like heat flux, and your proposal to always pass dimensional values across interfaces like CHT also sounds reasonable to me. We should also double-check what is done for FSI. It is important that we define a standard and follow it consistently.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
economon: Perhaps it is best to have the flow solvers compute the information that you need here just once and store it (maybe we can use what is computed in Friction_Forces() or something similar), and only have the interface class access it from the corresponding solver (compressible vs incompressible). It could be dangerous to have two implementations of the same thing - if one changes and the other doesn't, the functionality breaks. It would be a lot easier to maintain, and simpler, to define it just once. Also, these quantities can be different between compressible and incompressible.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
economon: See previous comment above..

SU2_CFD/src/solver_direct_heat.cpp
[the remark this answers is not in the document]
oleburghardt: Okay, thanks for the clarification.

Then the changes in my latest commits do what they should :)

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
oleburghardt: Yes it's just about the compressible case. 
For the incompressible one, I get a non-dimensionalized thermal conductivity from `donor_solution->node[iPoint]->GetThermalConductivity()` to then compute everything and finally use `donor_config->GetHeat_Flux_Ref()` to dimensionalize.
I am just not so much familiar with the compressible solver, but if it's sure that I could e.g. get the thermal conductivity always from `FluidModel` I could replace the computation (which is unsafe, as you pointed out) there by such a solver call.

SU2_CFD/src/SU2_CFD.cpp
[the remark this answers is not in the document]
oleburghardt: Just a follow-up note as what I wrote here

> as the current one for single zone problems should be faster as there is just one std::vector instead of a large amount of dereferencings down to CVariable.

is nonsense, as in both cases we first go to `CVariable` (and then we read indices from an external structure in the standard or from an own one in added one) so there shouldn't be any difference in performance.

@talbring just encouraged me to change it directly everywhere, I'll consider it.

pcarruscag:
@oleburghardt, looks like this is taking a turn to also address primal CHT topics, but all comments on the main dish have been addressed. Can we merge this to shorten the dependency chain between PR's while you continue working on ironing out minor kinks?

oleburghardt:
> @oleburghardt, looks like this is taking a turn to also address primal CHT topics, but all comments on the main dish have been addressed. Can we merge this to shorten the dependency chain between PR's while you continue working on ironing out minor kinks?

True, but that was about it. They just have fit in so well. As for the main part, we just have to keep in mind the `CDiscAdjMultizoneDriver::SetObjFunction` that two of you didn't like. But as I said, it has much to do with other changes that should be done before.

SU2_CFD/src/SU2_CFD.cpp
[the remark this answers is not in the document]
oleburghardt: However adaptions would then probably also have to be made to `CDiscAdjFEASolver` to not break the adjoint FSI cases. Maybe we should really postpone it to the FSI integration into the multiphysics driver, that would line up well.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -104,95 +93,112 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
 
-    heat_flux_density       = thermal_conductivity*dTdn;
-    conductivity_over_dist  = thermal_conductivity/dist;
+      thermal_conductivity    = thermal_conductivityND*donor_config->GetViscosity_Ref();
+      conductivity_over_dist  = thermal_conductivity/dist;
+    }
   }
   else if (incompressible_flow) {
 
-    iPoint = donor_geometry->vertex[Marker_Donor][Vertex_Donor]->GetNode();
```
pcarruscag: @oleburghardt just in the process of merging I think this line was removed by mistake? iPoint is used right after

oleburghardt: Yes, it's a mistake. If you are in the merge window right now, can you revert that change/add that line in your branch? That should be the quickest solution to fix it in develop. Thanks in advance..

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2443,7 +2439,7 @@ void CDiscAdjFluidIteration::SetDependencies(CSolver *****solver,
   }
 
   /*--- Compute coupling between flow and turbulent equations ---*/
-
+  solver[iZone][iInst][MESH_0][FLOW_SOL]->Preprocessing(geometry[iZone][iInst][MESH_0], solver[iZone][iInst][MESH_0], config[iZone], MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, true);
```
pcarruscag: @oleburghardt this call is duplicated, can one of them be removed?

oleburghardt: I'll check and eventually delete in in #798 Thanks!

PR #778, opened 2019-08-28
Mesh Quality Metrics

economon:
## Proposed Changes
 
This PR contains a new routine to compute 3 mesh quality metrics (similar to what you may find in commercial codes): orthogonality angle, cell aspect ratio, and cell volume ratio. It replaces an old routine that was not being used. Note that this is for the **dual control volumes of the finite volume solver** (not the primal elements), so the values are vertex-based.

The min & max of the 3 metrics are reported to the console during preprocessing, and if you would like the fields to be written to the viz. files, you can use 

```
WRT_MESH_QUALITY= YES
```

For now, it is just for sanity checks. In the future, we can use this as an optimization objective/constraint, for automatically improving grids (r-adaptation), or as part of advanced limiters/numerical methods.

## Related Work

## PR Checklist


Common/src/geometry_structure.cpp
```diff
@@ -15554,49 +15554,352 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+  /*--- Volume ratio is computed by looping over all volume elements and
+   computing the sub-element volume contributions. The ratio between the
+   largest and smallest sub-elements making up the dual CV is a
+   measure of the volume stretching ratio for the cell. Smaller
+   is better (closer to isotropic). ----*/
+  
+  unsigned long face_iPoint = 0, face_jPoint = 0;
+  unsigned short nEdgesFace = 1;
+  
+  su2double *Coord_Edge_CG     = new su2double[nDim];
```
pcarruscag: Will we ever have nDim = 4? In general we allocate a lot of small arrays on the heap which is not very efficient and seems unnecessary.

aeroamit:
This is indeed important. Having some reference quality metric,  while creating mesh (acceptable limits, specific to the solvers, which  evolves with usage).

economon:
Thanks for the quick review @pcarruscag. I have cleaned up a little bit. Also added a nicer table of the output.

l prefer that we always compute the min/max values, at least for the FVM CFD solvers, for a few reasons. The most important reason is that we should start to build up some intuition about the correlation between mesh quality metrics and accuracy + convergence in the solvers. This can be especially important during optimization when the mesh is deforming. I think the metrics are valuable to print every time for this reason, which might help users diagnose problems eventually.

Additionally, the cost is relatively small (on par with the other pre-processing routines, it is only executed once and it is parallelized, and the memory is released if we don't write).

EduardoMolina:
Hi @economon 

Thanks for this PR. It is an essential feature that was missing. 
Also, I would like to see in the future acceptable min/max values for the metrics that SU2 can handle.
LGTM and ready to merge after you update from develop.

Eduardo

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+    const unsigned long GlobalIndex_i = node[iPoint]->GetGlobalIndex();
+    const unsigned long GlobalIndex_j = node[iPoint]->GetGlobalIndex();
+    
+    const su2double *Normal = edge[iEdge]->GetNormal();
+    
+    /*--- Face area (norm of the normal vector) ---*/
+    
+    su2double Area = 0.0;
+    for (unsigned short iDim = 0; iDim < nDim; iDim++)
+      Area += Normal[iDim]*Normal[iDim];
```
pcarruscag: You can write pow(x,2) and all compilers will do this optimization for you.

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+    Area = sqrt(Area);
+    
+    /*--- Aspect ratio is the ratio between the largest and smallest
+     faces making up the boundary of the dual CV and is a
+     measure of the aspect ratio of the dual control volume. Smaller
+     is better (closer to isotropic). ----*/
+    
+    if (node[iPoint]->GetDomain()) {
+      if (Area < Area_Min[iPoint]) Area_Min[iPoint] = Area;
```
pcarruscag: std::min and std::max

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+    
+    /*--- The definition of orthogonality is an area-weighted average of
+     90 degrees minus the angle between the face area unit normal and
+     the vector between i & j. If the two are perfectly aligned, then
+     the orthogonality is the desired max of 90 degrees. If they are
+     not aligned, the orthogonality will reduce from there. Good values
+     are close to 90 degress, poor values are typically below 20 degress. ---*/
+    
+    if (node[iPoint]->GetDomain()) {
+      Orthogonality[iPoint] += Area*(90.0 - acos(dot_prod)*180.0/PI_NUMBER);
```
pcarruscag: Should this be computed such that it is always in range 0-90? (135 is as bad as 45)

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+        switch (nDim) {
+          case 2:
+            
+            /*--- Check if sub-elem volume is the min or max for iPoint. ---*/
+            
+            Volume_i = edge[iEdge]->GetVolume(Coord_FaceiPoint.data(),
+                                              Coord_Edge_CG.data(),
+                                              Coord_Elem_CG.data());
+            if (node[face_iPoint]->GetDomain()) {
+              if (Volume_i < SubVolume_Min[face_iPoint])
```
pcarruscag: more std::min std::max

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+      SurfaceArea[iPoint]   += Area;
+    }
+    if (node[jPoint]->GetDomain()) {
+      Orthogonality[jPoint] += Area*(90.0 - acos(dot_prod)*180.0/PI_NUMBER);
+      SurfaceArea[jPoint]   += Area;
+    }
+    
+    /*--- Error check for zero volume of the dual CVs. ---*/
+    
+    if (node[iPoint]->GetVolume() == 0.0) {
```
pcarruscag: Is it safer to check for <= ?

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+  /*--- Volume ratio is computed by looping over all volume elements and
+   computing the sub-element volume contributions. The ratio between the
+   largest and smallest sub-elements making up the dual CV is a
+   measure of the volume stretching ratio for the cell. Smaller
+   is better (closer to isotropic). ----*/
+  
+  unsigned long face_iPoint = 0, face_jPoint = 0;
+  unsigned short nEdgesFace = 1;
+  
+  vector<su2double> Coord_Edge_CG(nDim);
```
pcarruscag: Thanks for the better management.

Common/src/geometry_structure.cpp
```diff
@@ -15564,49 +15564,349 @@ void CPhysicalGeometry::SetColorGrid_Parallel(CConfig *config) {
...
+    
+    vector<su2double> Unit_Edge_Vector(nDim);
+    for (unsigned short iDim = 0; iDim < nDim; iDim++)
+      Unit_Edge_Vector[iDim] = Edge_Vector[iDim]/dist_ij;
+    
+    /*--- Compute the angle between the unit normal associated
+     with the edge and the unit vector pointing from iPoint to jPoint. ---*/
+    
+    su2double dot_prod = 0.0;
+    for (unsigned short iDim = 0; iDim < nDim; iDim++)
```
pcarruscag: Some of these loops for areas, edge vectors and dot prod can be fused.

SU2_CFD/include/SU2_CFD.hpp
```diff
@@ -39,7 +39,7 @@
 #pragma once
 
 #include "../../Common/include/mpi_structure.hpp"
-#include <CLI11.hpp>
+#include "../../externals/CLI11/CLI11.hpp"
```
pcarruscag: Same comment as in another PR, this should probably be `#include "CLI11.hpp"` and the path added to the include directories of your IDE

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -815,6 +813,14 @@ void CDriver::Geometrical_Preprocessing_FVM(CConfig *config, CGeometry **&geomet
   if (rank == MASTER_NODE) cout << "Checking for periodicity." << endl;
   geometry[MESH_0]->Check_Periodicity(config);
   
+  /*--- Compute mesh quality statistics on the fine grid. ---*/
+  
+  if (!fea) {
```
pcarruscag: Thanks, I agree we need a larger sweep to see what can be trimmed.

SU2_CFD/include/SU2_CFD.hpp
talbring: Yep, I would also suggest doing it like that.

SU2_CFD/src/drivers/CDriver.cpp
talbring: Maybe it would make sense to create a child class of CPhysicalGeometry or CGeometry for the FEA mesh (not now but in the future).

economon: Absolutely. Let's come back to that when we split up PhysicalGeometry. I have some ideas about splitting that one.. it could easily be 3 or more separate classes.

SU2_CFD/include/SU2_CFD.hpp
economon: Will revert. I also reverted this in the CGNS PR. Yep, Xcode is picky, but we can just add an extra search path inside the project file

Common/src/geometry_structure.cpp
EduardoMolina: Hi @pcarruscag and @economon ,

To be on the safe side, I think we can check against a very small number like 1e-20.

economon: Yep, I'll clean up a few more things like this on this PR soon

Common/src/geometry_structure.cpp
economon: Done

Common/src/geometry_structure.cpp
economon: For now, let's go with <= 0.0

Common/src/geometry_structure.cpp
economon: Rearranged some of the smaller loops for efficiency

Common/src/geometry_structure.cpp
economon: During creation of the dual CVs, we orient the sub-face area of a CV in the direction from iPoint to jPoint. So when we take the dot product with the position vector from iPoint to jPoint, the max value will be 90 deg by construction (giving us a 0.0 orthogonality value - worst). If they are perfectly aligned, the dot product is zero (giving us a 90.0 orthogonality angle - best).

PR #780, opened 2019-08-28
Enhancement fixed cl

jayantmukho:
## Proposed Changes
This is a simple enhancement of the fixed CL mode. Currently the Fixed CL mode only exits when the specified iteration limit is reached. With these changes, the mode is a little more robust and checks for the specified Residual or Cauchy convergence and additionally checks that the CL is converged to the target CL value to within CAUCHY_EPS.

I have currently preserved the Finite Differencing that occurs at the end of the Fixed CL mode. I am not sure why this is required and would like to get rid of it if possible. If there is a reason for the finite differencing to calculate dCL/dAlpha, please let me know.

## PR Checklist


SU2_CFD/src/integration_time.cpp
```diff
@@ -845,7 +867,29 @@ void CSingleGridIntegration::SingleGrid_Iteration(CGeometry ****geometry, CSolve
   /*--- Convergence strategy ---*/
   
   Convergence_Monitoring(geometry[iZone][iInst][FinestMesh], config[iZone], Iteration, monitor, FinestMesh);
-  
+
+    /* --- If fixed C_L mode, ensure that C_L is converged to target C_L --- */
+  if (config[iZone]->GetFixed_CL_Mode() && Convergence){
```
talbring: See comment above. Same here.

jayantmukho:
@talbring That was a good point about placement. I was able to move this implementation to a function called `FixedCL_Convergence()` in the euler solver class. It also allowed me to make it a little more robust. Since it still requires information about the convergence that is calculated in the integration container, the method has to be called `CFluidIteration::Iterate()`

I included the details of the implementation in the Constant CL tutorial on the website. That is in a pull request for the website. Consequently, that [pull request ](https://github.com/su2code/su2code.github.io/pull/16) needs to be approved first so that the regression test for this one can pass. This is because of a change in options in the Fixed CL mode.

SU2_CFD/src/integration_time.cpp
```diff
@@ -131,6 +131,28 @@ void CMultiGridIntegration::MultiGrid_Iteration(CGeometry ****geometry,
   
   Convergence_Monitoring(geometry[iZone][iInst][FinestMesh], config[iZone], Iteration, monitor, FinestMesh);
 
+  /* --- If fixed C_L mode, ensure that C_L is converged to target C_L --- */
+  if (config[iZone]->GetFixed_CL_Mode() && Convergence){
```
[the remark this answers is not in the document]
jayantmukho: This was a good point and it has been addressed, details in comment below.

SU2_CFD/src/integration_time.cpp
jayantmukho: This was a good point and it has been addressed, details in comment below.

SU2_CFD/src/solver_adjoint_mean.cpp.orig
```diff
@@ -0,0 +1,6745 @@
+/*!
```
pcarruscag: I think you committed this by mistake @jayantmukho

jayantmukho: yup, you're correct :man_facepalming:

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -7320,231 +7308,195 @@ void CEulerSolver::SetActDisk_BCThrust(CGeometry *geometry, CSolver **solver_con
...
-      Wrt_Con_Freq = SU2_TYPE::Int(su2double(config->GetIter_dCL_dAlpha())/10.0);
-      config->SetWrt_Con_Freq(Wrt_Con_Freq);
-      Total_CD_Prev = Total_CD;
-      Total_CL_Prev = Total_CL;
-      Total_CMx_Prev = Total_CMx;
-      Total_CMy_Prev = Total_CMy;
-      Total_CMz_Prev = Total_CMz;
-      AoA_inc = 0.001;
-      AoA_FD_Change = true;
+    Wrt_Con_Freq = SU2_TYPE::Int(su2double(Iter_dCL_dAlpha)/10.0);
```
pcarruscag: Can you please check if all these casts are required? It seems Iter_dCL_dAlpha is of integer type, so you can probably just divide by 10.

jayantmukho:
I am having some trouble in getting the Fixed CL mode to work properly with the new Output structure and SingleZoneDriver was wondering if anyone had some advice. 

The way the fixed Cl mode works is that it first converges the solution to the target CL that is set in the config options. After that, it increases the `AOA` by 0.001 and runs for `ITER_DCL_DALPHA` iterations to determine the gradient `dCL_dAlpha`. This is essentially a finite differencing step to calculate gradients such as dCD_dCL so that it can be added to the objective function if the discrete adjoint is run in the fixed_cl mode. Currently, both the fixed cl convergence and the finite differencing occurs within the one call of `CFluidIteration::Solve` 

This PR just changes how the fixed_CL mode converges to the `TARGET_CL` value and doesn't change anything about how the finite differencing step is performed. 

My issue is to do with how to output the correct solution (the state before the finite differencing step) at the end of the simulation. Both the fixed cl convergence and the finite differencing occur within the one call of `CFluidIteration::Solve`. But the final solution output is controlled by the `CSinglezoneDriver::Output`. But by the time this is called the solution is at the state after the finite difference step has been taken and the CL is no longer matched. 

The proper way to do it in this new output structure/SingleZoneDriver format would be to run the fixed CL mode to convergence using the SingleZoneDriver (including doing all the outputs from the `CSinglezoneDriver::Output` and then running the finite difference after that. But that wouldn't be possible without putting the if (fixed_cl) statement in either the SingleZoneDriver or in SU2_CFD, both of which are much higher level, and shouldn't have a problem specific if statement.

The other way to do it would be to force the output structure to write the solution files before the finite differencing step (within the CFluidIteration) and to prevent the writing of the files after the finite differencing step. I was able to do this for the history file and the screen output by adding some functions that can alter the Frequency of writing these files. But I have not found a way to prevent the volume files from being written.

pcarruscag:
What if you start with the "perturbed value" of AoA and then set it to the correct value for the FD step?

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -7320,231 +7308,195 @@ void CEulerSolver::SetActDisk_BCThrust(CGeometry *geometry, CSolver **solver_con
...
-      cout << "Previous AoA: " << Old_AoA << " deg";
-      cout << ", new AoA: " << config->GetAoA() << " deg." << endl;
+      if (CL_Converged){
+        cout << "CL converged to within "<< config->GetCauchy_Eps() <<", not changing AoA" << endl;
+      }
+      else {
+        cout.precision(4);
+        cout << "Previous AoA: " << AoA << " deg";
+        cout << ", new AoA: " << AoA + AoA_inc << " deg." << endl;
+      }
```
talbring: Can you add all these things as history fields to the output structure ?

talbring:
> What if you start with the "perturbed value" of AoA and then set it to the correct value for the FD step?

I actually just wanted to suggest the same without reading your comment :)

SU2_CFD/src/solver_direct_mean.cpp
jayantmukho: Yup! I was thinking of doing that as well. Angle of attack is already at output field, so the only thing left is the C_L convergence. I will add that. 

Did you also mean to add the print statements like `Previous AoA:`?

jayantmukho:
> What if you start with the "perturbed value" of AoA and then set it to the correct value for the FD step?

Mhmm, I am just thinking this through. At the end of the fixed_cl part, the flow is converged to the `TARGET_CL` and is at some angle of attack `alpha`. If I wanted to start from the perturbed value and set it to the final value of `alpha`, I would have to first, converge it to the perturbed value (`alpha + 0.001`), and then perform the finite difference step back to `alpha`.

This would take approximately twice the number of `ITER_DCL_DALPHA` to perform. Which isn't the end of the world, but I would like to avoid if possible

Is there a cleverer way of doing this that I am not thinking off? The problem is that we don't know the final `alpha` until the fixed_cl simulation converges.

SU2_CFD/src/solver_direct_mean.cpp
jayantmukho: Good call, not needed

pcarruscag: There were a few other places.

pcarruscag:
Probably not, it was a silly idea anyway. Maybe you could perturb the target value of CL instead and measure the variation of alpha for an imposed variation of CL but that is starting to sound quite strange.

I don't known how the process to find the alpha that gives you a given CL works, but could the derivative not be obtained from this process? i.e. you could avoid the FD step entirely.

jayantmukho:
> I don't known how the process to find the alpha that gives you a given CL works, but could the derivative not be obtained from this process? i.e. you could avoid the FD step entirely.

It is a simple proportional controller that will change the angle of attack after a number of iterations depending on the difference between the current CL and the target CL. 

So if the the option `ITER_DCL_DALPHA = 0` then it will do exactly what you suggest. It calculates the derivative based on the last update. The only problem with that is, there is no guarantee that the simulation at the previous update was converged, so the derivative might be incorrect. 

But you are correct, I could just do a really tiny update (order of delta AoA ~10^-6) at the end and calculate the finite difference using that value. Let me try that and see if the gradients it calculates are reasonable, or if the update is too small and it gets clouded by numerical errors.

jayantmukho:
Unfortunately, there is an issue with doing a tiny update at the end of the simulation. To illustrate the numbers, lets assume that CL is converged to the `TARGET_CL` to within 10^-6. 

In this case the finite-difference update is tiny: \alpha is changed by 10^-7 degrees (this is because `DCL_DALPHA is usually between 0.5 and 0.05 and change in alpha = `(TARGET_CL-CL)/DCL_DALPHA`) . The change in CL resulting from this finite-difference step is small (10^-8) enough that if the solution isn't extremely converged (in this case, change in CL over two consecutive iterations is < 10^-9), the gradients calculated will be incorrect. 

Anecdotally, the reduction in density residual in the simulation would have to be ~8 orders to get these gradients to be correct. This could add a large number of iterations and/or time to solution, especially for 3D RANS cases. Additionally, I am not sure what the best way to force this level of convergence would be. 

Which brings me back to my original problem of trying to output the solution before the finite difference step. Well, I can definitely output the solution at the end of the fixed Cl mode, but it just gets overwritten when the solver finally exits after the finite difference step. The easiest way around this would be if there is a way to disable the output of the volume files after the solver exits. Something that could be changed from the CFluidIteration ideally. 

I am sorry this feature is causing this issue. Unfortunately, even the way it is currently implemented in develop (without my changes, but after the new output structure) is plagued with the same problem.

talbring:
@jayantmukho You can modify the routine CFlowOutput::WriteVolume_Output() in order to disable the writing of the files.

SU2_CFD/src/solver_direct_mean.cpp
talbring: Yeah I mean all of the values. Maybe we can also remove that output then ?

jayantmukho:
> @jayantmukho You can modify the routine CFlowOutput::WriteVolume_Output() in order to disable the writing of the files.

Yes, I thought that would be the easiest place to make the change. But unfortunately, this gets ignored when the `CSingleZoneDriver` runs `SetResult_Files` and forces the output with the `force_writing` option. `StopCalc` passed into this function for `force_writing` and that is basically always true at this point in the case of steady RANS since the `CFluidIteration` has just finished running `Solve()`

An easy way to fix this would be to add `force_writing` to the input of `WriteVolume_Output`. I can preserve the behavior in the `COutput::WriteVolume_Output` but add some logic for the fixed CL mode in the `CFlowOutput::WriteVolume_Output`.

SU2_CFD/src/solver_direct_mean.cpp
jayantmukho: I can add Previous AoA to the output_structure. But I do think that having a little header when it actually changes the AoA is useful.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -2820,10 +2820,11 @@ bool CFlowOutput::WriteVolume_Output(CConfig *config, unsigned long Iter){
       return true;
     }
   } else {
-    return ((Iter > 0) && Iter % config->GetVolume_Wrt_Freq() == 0);
+    if (config->GetFixed_CL_Mode() && config->GetFinite_Difference_Mode()) return false;
+    return ((Iter > 0) && Iter % config->GetVolume_Wrt_Freq() == 0) || force_writing;
```
jayantmukho: This isn't the cleanest implementation but is the least intrusive as it is mostly contained in CFlowOutput. @talbring 

The other way to do this (which I have a slight preference towards) would be to add an option to disable volume output completely in COutput itself. I would prefer it because it would be a more general function that could possibly be used in other situations as well. But it would be a pretty powerful option that might cause conflicts with other options/solvers if not handled with care.

Unfortunately, the way it is now, there is one state variable (Finite_Difference_Mode) that is stored in the config @pcarruscag . But everything else is contained in the solver

SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -691,3 +720,48 @@ bool CFlowCompOutput::SetUpdate_Averages(CConfig *config){
...
+    SetFixedCLScreenOutput(config);
+  } 
+}
+
+void CFlowCompOutput::SetFixedCLScreenOutput(CConfig *config){
+  PrintingToolbox::CTablePrinter FixedCLSummary(&cout);    
+  
+  if (fabs(historyOutput_Map["CL_DRIVER_COMMAND"].value) > 1e-16){
+    FixedCLSummary.AddColumn("Fixed CL Mode", 40); 
+    FixedCLSummary.AddColumn("Value", 30);
```
jayantmukho: I moved all the output to the CFlowCompOutput. I had to make a few new functions but I tried to copy the way the WriteAdditionalFiles function works. I hope this is what you had in mind @talbring 

It prints everytime the angle of attack changes and looks like this: 
image

SU2_CFD/src/solver_direct_mean.cpp
```diff
@@ -7320,71 +7310,51 @@ void CEulerSolver::SetActDisk_BCThrust(CGeometry *geometry, CSolver **solver_con
 void CEulerSolver::SetFarfield_AoA(CGeometry *geometry, CSolver **solver_container,
```
jayantmukho: This function got cleaned up, and now it does only SetFarfieldAoA (As opposed to perform all the logic for the Fixed CL mode). Moved all that to the FixedCL_Convergence function. Which now that I say out loud, should be name FixedCL_Driver or something of that sort

.travis.yml
```diff
@@ -72,11 +72,11 @@ install:
 
 before_script:
     # Get the test cases
-    - git clone --depth=1 -b develop https://github.com/su2code/TestCases.git ./TestData
+    - git clone --depth=1 -b enhancement_fixed_cl https://github.com/su2code/TestCases.git ./TestData
```
talbring: Dont forget to change that

SU2_CFD/include/output/CFlowOutput.hpp
```diff
@@ -122,8 +122,9 @@ class CFlowOutput : public COutput{
    * \brief Determines if the the volume output should be written.
    * \param[in] config - Definition of the particular problem.
    * \param[in] Iter - Current iteration index.
+   * \param[in] force_writing - boolean that forces writing of volume output
    */
-  bool WriteVolume_Output(CConfig *config, unsigned long Iter) override;
+  bool WriteVolume_Output(CConfig *config, unsigned long Iter, bool force_writing) override;
```
talbring: I am fine with that change!

SU2_CFD/include/output/CFlowCompOutput.hpp
```diff
@@ -118,5 +118,21 @@ class CFlowCompOutput final: public CFlowOutput {
    */
   bool SetUpdate_Averages(CConfig *config) override;
 
+  /*!
+   * \brief Write any additional output defined for the current solver.
+   * \param[in] config - Definition of the particular problem per zone.
+   */
+  void SetAdditionalScreenOutput(CConfig *config) override;
```
talbring: Good idea to add a routine like that.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -4910,16 +4947,14 @@ class CEulerSolver : public CSolver {
   su2double *Secondary,    /*!< \brief Auxiliary nPrimVar vector. */
   *Secondary_i,        /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point i. */
   *Secondary_j;        /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point j. */
-  
-  su2double Cauchy_Value,  /*!< \brief Summed value of the convergence indicator. */
-  Cauchy_Func;      /*!< \brief Current value of the convergence indicator at one iteration. */
-  unsigned short Cauchy_Counter;  /*!< \brief Number of elements of the Cauchy serial. */
-  su2double *Cauchy_Serie;      /*!< \brief Complete Cauchy serial. */
```
talbring: Thanks for cleaning that up

SU2_CFD/src/output/CFlowOutput.cpp
talbring: I am ok with that!

SU2_CFD/src/output/CFlowCompOutput.cpp
talbring: Yep looks good

jayantmukho:
In the process of the final merge which involves pulling corresponding TestCase and Tutorial branches into their develops as well. (and changing the travis.yml file)

Issue #733, opened 2019-07-19
Shape Optimization 


jayantmukho:
Hey everyone,

I have been taking a deep dive into the shape optimization features in SU2 and have resurfaced with some observations, suggestions, and questions.

First, the bug:
Even though the config_template, and option structure have changed the numbering of the DEFINITION_DV (for example, FFD_CONTROL_POINT went from being 7 to 11), this has not been reflected in the SU2_PY/SU2/io/tools.py, or any of the tutorials. Therefore, there is a mismatch in what the C++ code uses, and what the python code uses. This is an easy problem to fix and is being fixed in the branch fix_set_ffd_script.

Next, I want to have a discussion about the shape optimization options and scalings. Currently there are the following options in the config template to control scaling:

 ```
% Optimization objective function with scaling factor, separated by semicolons.
% To include quadratic penalty function: use OPT_CONSTRAINT option syntax within the OPT_OBJECTIVE list.
% ex= Objective * Scale
OPT_OBJECTIVE= DRAG * 10
% Optimization constraint functions with pushing factors (affects its value, not the gradient  in the python scripts), separated by semicolons
% ex= (Objective = Value ) * Scale, use '>','<','='
OPT_CONSTRAINT= ( LIFT > 0.328188 ) * 0.001; ( MOMENT_Z > 0.034068 ) * 0.001; ( AIRFOIL_THICKNESS > 0.11 ) * 0.001
%
% Factor to reduce the norm of the gradient (affects the objective function and gradient in the python scripts)
% In general, a norm of the gradient ~1E-6 is desired.
OPT_GRADIENT_FACTOR= 1E-6
%
% Factor to relax or accelerate the optimizer convergence (affects the line search in SU2_DEF)
% In general, surface deformations of 0.01'' or 0.0001m are desirable
OPT_RELAX_FACTOR= 1E3
%
% Optimization bound (bounds the line search in SU2_DEF)
OPT_LINE_SEARCH_BOUND= 0.1
%
% Upper bound for each design variable (bound in the python optimizer)
OPT_BOUND_UPPER= 1E10
%
% Lower bound for each design variable (bound in the python optimizer)
OPT_BOUND_LOWER= -1E10
%
% Optimization design variables, separated by semicolons
DEFINITION_DV= ( 19, 1.0 | airfoil | AIRFOIL, 0, 0, 0, 1.0 ); ...
```

[...]

PR #781, opened 2019-08-28
Enhancement multipoint opt

jayantmukho:
## Proposed Changes
This is a little improvement to the multipoint optimization setup. Currently if you want to run multipoint optimization with restarts, the python scripts only keep track of one restart file (the first of the multipoint conditions). This is acceptable when the multipoint conditions are similar, but does not work when the conditions are very different. For example, if one point is supersonic and the other is subsonic, and the subsonic case is restarted from the supersonic solution, it immediately diverges. 

This PR allows the python scripts to keep track of restarts for each point in the multipoint set up. It is robust in the sense that you can start with 0 restart files or some subset of restart files. For example, if you are running a 3 point optimization but only have the restart for one of the points, this will handle that case by starting the other two points from scratch, but keeping track of the restart files after the first design iteration

There is a naming convention that is presumed. The restarts get 'point#' added to its base name. So for example, if the SOLUTION_FILENAME= solution_flow.dat then it will look for files named solution_flow_point0.dat, solution_flow_point1.dat and so on. For adjoints this is solution_adj_point0_cd.dat etc. This naming convention can be tweaked according to comments. 

## PR Checklist


erangit:
@jayantmukho :
I think this is a very important addition. A typical optimization case (at least in the industry) deals with very different setups (say aircraft at cruise, take-off, and landing: different flight regimes and different configurations).  That way it makes sense to allow also for different mesh files, each for every point.
Nimrod Cohen who works with me has been working on that line. I don't know how far he has gone.
I shall have him posted to this PR, he might be able to contribute.
Best,
Eran

SU2_PY/SU2/io/state.py
```diff
@@ -190,11 +191,18 @@ def pullnlink(self,config):
...
                 link.extend(value)
-            elif 'ADJOINT_' in key:
+            elif 'ADJOINT_' in key and  (not 'MULTIPOINT' in key):
                 # adjoint solution
                 value = expand_zones(value,config)
                 value = expand_time(value,config)
                 link.extend(value)
+            elif 'MULTIPOINT' in key:
+                # multipoint files
+		value = expand_zones(value,config)
```
economon: check indentation here

SU2_PY/SU2/eval/gradients.py
```diff
@@ -469,14 +473,24 @@ def multipoint( func_name, config, state=None, step=1e-2 ):
...
+    orig_marker_outlet = config['MARKER_OUTLET']
+    orig_marker_outlet = orig_marker_outlet.replace("(", "").replace(")", "").split(',')
+    new_marker_outlet = "(" + orig_marker_outlet[0] + "," + outlet_value_list[0] + ")"
+    config.MARKER_OUTLET = new_marker_outlet
+    config.SOLUTION_FLOW_FILENAME = solution_flow_list[0]
+    config.SOLUTION_ADJ_FILENAME = solution_adj_list[0]
+    if MULTIPOINT_ADJ_NAME in state.FILES and state.FILES[MULTIPOINT_ADJ_NAME][0]:
+        state.FILES[ADJ_NAME] = state.FILES[MULTIPOINT_ADJ_NAME][0]
+
+    #state.find_files(config)
```
economon: remove this if unneeded

SU2_PY/SU2/opt/project.py
```diff
@@ -368,8 +375,14 @@ def init_design(self,config,closest=None):
         # update local state filenames ( ??? why not in Design() )
         for key in design.files:
             name = design.files[key]
-            name = os.path.split(name)[-1]
-            design.files[key] = name
+            if isinstance(name,list):
+            	built_name = []
```
economon: double check indentation throughout

SU2_PY/SU2/eval/gradients.py
```diff
@@ -469,14 +473,24 @@ def multipoint( func_name, config, state=None, step=1e-2 ):
...
+    
     config.AOA = aoa_list[0]
     config.SIDESLIP_ANGLE = sideslip_list[0]
     config.MACH_NUMBER = mach_list[0]
     config.REYNOLDS_NUMBER = reynolds_list[0]
     config.FREESTREAM_TEMPERATURE = freestream_temp_list[0]
     config.FREESTREAM_PRESSURE = freestream_press_list[0]
-    config.TARGET_CL = target_cl_list[0]   
+    config.TARGET_CL = target_cl_list[0]
+    orig_marker_outlet = config['MARKER_OUTLET']
```
economon: It would be really great if we had the ability to easily customize the types of quantities that we use for the multipoint. As written in the framework, every time you want to use a new quantity (such as the outlet values that I had added before), you need to manually add some of these options to the python scripts. If you have some ideas for that, I am all ears.

SU2_PY/SU2/opt/project.py
jayantmukho: Yup, I made the mistake of editing some of these files in vim and using tabs. Fixed

SU2_PY/SU2/eval/gradients.py
jayantmukho: I agree. This is not the easiest way to go about it. I think combining this with an overall effort of consolidating the IO python scripts and the in-memory ones might be the way to go. 

I haven't really played with those python wrappers but it might be time to do that. I would imagine writing new python scripts for shape optimization and some of the other functionality that utilized the new python wrapper would probably lend it more flexibility.

Specifically we might be able to better address some of the scaling issues that we see in #733 and perform some smart normalizations. I also imagine that we could handle solution divergences intelligently to give the optimizer some useful information even though the CFD solver doesn't converge the case. This could make the optimization process a little more robust.

jayantmukho:
@erangit I got the multipoint optimization to work with multiple mesh files. For a well posed optimization, the surfaces that can be deformed (for example the wing) should be same, and the FFD boxes have to be identical.

Since the meshes are likely unstructured, you might not get the exact same discretization on the surface, which is fine. As long as the FFD box definition is the same, identical movements in the FFD control points should result in similar surface deformations. Just the marker definitions need to be consistent.

jayantmukho:
This PR is ready for merging. The Codacy failure is for a os.system() call that introduces a security vulnerability that I am not sure how to fix. There are a lot of os.system() calls in the code already, but since I introduced one more, it is failing. 

Any help in fixing the issue would be fantastic. If this is not a big deal, I can merge the PR

economon:
Seems that 'shell=True' is a security risk. Any way to easily fix that by using different options with the system call? We could do a search and replace for the others already in the Python framework.

jayantmukho:
shell = false doesn't seem to work in this particular case (symbolic linking) but it might in others. There seem more secure ways to run commands but they need case by case treatment. As in the solution for a `cp` command is different from a `ln -s` command. 

We can also replace the `os.system` calls with other python functions (for example `os.symlink` for symbolic linking). 

Either way, it wont be a simple search and replace. There seem to be about 25 `os.system` calls across the python scripts. Let me try and replace them.

talbring:
@jayantmukho Thanks for dealing with the python calls ... I think as soon as you deal with the two unused variables everything should be fine and we can merge that in!

economon:
Agreed, thanks for fixing those security issues. I know it is more work, but it is important. Extra fixes added to PRs like this help improve the code quickly.

jayantmukho:
Unfortunately, I wasn't able to get all of them. I fixed a majority of the calls but couldn't find workarounds all of them. There are 8 left in the following files: 

`SU2_PY/change_version_number.py`: There's 2 in here. One is a complicated command with `grep` and pipes and the other is a simple `rm -rf`, but it isn't a security risk since there is no user input into that string

`SU2_PY/SU2/util/pyCppTap.py`: This is in the diff_routine and uses `tapenade`. I have no idea what this does so I didn't touch this one

`SU2/opt/server.py`: There are 4  `scp` calls that I didn't know how to replace

`SU2_PY/compute_polar.py`: There's one left in here which is a complicated `cat` call that I couldn't figure a workaround for. 

If you have any suggestions for any of these, I can try and implement them. 

Otherwise, once these tests pass, its good to merge.

economon:
We can remove pyCppTap.py. This is a holdover from early automatic differentiation work that is not functional any longer.

Can you also please check whether the server.py is still necessary? That may only be needed for design space exploration routines (farming out jobs on a cluster) that are also not used regularly.

jayantmukho:
Seems like `server.py` is only used by `patient_designspace.py` which is a python script I hadn't ever noticed. There really isn't much in the way of documentation, use cases, or comments for either of these files which makes me okay with scrapping them for this release. I will remove these three files (including `pyCppTap.py`) and take them out of the Makefiles as well.

talbring:
@jayantmukho Thanks! Is it then ready from your side?

jayantmukho:
This is good from my end. I removed the files we had talked about. The only thing I am unsure off, is how this affects the IDE files that are shipped with the code. When I grepped for the filenames that I removed, they show up in the `SU2_IDE/Wing/SU2_Wing.wpr file `

I haven't used any of the IDE files before so I am not sure what, if anything, needs to be done. 

Just updating with develop and pushing

economon:
Wing was one of the preferred IDEs for the development of the python IO-based wrappers early on in the project. I do not expect that folks are using it any longer, and unless anyone is against it (please speak up), I think it can also be removed.

jayantmukho:
Removed wing, ready for merge

economon:
Thanks for the final push, @jayantmukho! Time to get this one in.

PR #769, opened 2019-08-23
Restructuring CTransfer files

rsanfer:
## Proposed Changes
This PR restructures the CTransfer parent and child classes following discussions in #583 and #725. Given that I already started it in #760, I just went ahead and moved all the remaining classes. It should be compared to `feature_elastic_mesh_solver`.

## Related Work
Depends on #760 and should be merged afterwards.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.

PR #782, opened 2019-08-28
Fix path in Makefile which causes autotools builds to fail.

TobiKattmann:
## Proposed Changes
Latest develop: in SU2_CFD/obj/Makefile.am line 113 it needs to be `../src/transfer/physics/CTransfer_BoundaryDisplacements.cpp` instead of `../src/transfer/CTransfer_BoundaryDisplacements.cpp`.  
As travis uses the meson build, the autotools build is not checked anymore.

## Related Work
Introduced by #760 and #769 will fix it but I guess that should be merged quickly as everyone who merges develop and uses autotools will get this error.

## PR Checklist


talbring:
Thanks. Priority merge...

PR #786, opened 2019-09-04
Merge Contiguous CVariable into New Output WIP

pcarruscag:
## Proposed Changes
Merge feature_input_output into feature_contiguous_cvariable_PR.
There are some issues to address, I will point them out tomorrow.

## Related Work
#724
#753 

## PR Checklist


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -921,3 +921,36 @@ void CFlowOutput::Set_CpInverseDesign(CSolver *solver, CGeometry *geometry, CCon
   SetHistoryOutputValue("CP_DIFF", PressDiff);
 
 }
+
+su2double CFlowOutput::GetQ_Criterion(su2double** VelocityGradient) const {
```
pcarruscag: I think the Q criterion was broken, j-i was being used before it was set.
I copy pasted the changes from @economon on #704 and I moved the method to the common parent class CFlowOutput

.travis.yml
```diff
@@ -18,11 +18,11 @@ compiler:
 notifications:
     email:
         recipients:
-            - su2code-dev@lists.stanford.edu
+            - pcarruscag@gmail.com
   
 branches:
     only:
-        - develop
+        - merge_cvariable_output
```
pcarruscag: How do I make travis run on this branch?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -47,2318 +47,2340 @@
...
 
   /*!
    * \brief A virtual member.
+   * \param[in] iPoint - Point index.
    * \param[in] val_vector - Direction of projection.
    * \param[in] val_species - Index of the desired species.
    * \return Value of the projected velocity.
    */
-  inline virtual su2double GetProjVel(su2double *val_vector, unsigned short val_species) {return 0; }
+  inline virtual su2double GetProjVel(Idx_t iPoint, su2double *val_vector, Idx_t val_
```
pcarruscag: @talbring, @rsanfer,
I saw this method was moved to CVariable, however the new mesh adjoint solver still needs to override it because the BGS_k field is allocated only for boundaries, what should we do about this?

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -473,25 +473,25 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
   if (nDim == 3)
     SetVolumeOutputValue("COORD-Z", iPoint, Node_Geo->GetCoord(2));
   
-  SetVolumeOutputValue("PRESSURE",    iPoint, Node_Flow->GetSolution(0));
-  SetVolumeOutputValue("VELOCITY-X", iPoint, Node_Flow->GetSolution(1));
-  SetVolumeOutputValue("VELOCITY-Y", iPoint, Node_Flow->GetSolution(2));
+  SetVolumeOutputValue("PRESSURE",   iPoint, Node_Flow->GetSolution(iPoint, 0));
```
pcarruscag: @talbring, do you think it would make sense for `SetVolumeOutputValue` to operate on an entire column of data? e.g. set the data for Solution(0) on nPoints instead of point by point.

.travis.yml
talbring: Set it to feature_input_output, I think that should. Otherwise we have to manually trigger travis for now.

SU2_CFD/src/output/CFlowIncOutput.cpp
talbring: If we can make that work without having a loop or something like that in the LoadVolumeData routine, why not. I want to have that routine as simple as possible.

SU2_CFD/src/output/CFlowOutput.cpp
talbring: Thanks! That was already on my agenda ..

SU2_CFD/include/solvers/CMeshSolver.hpp
```diff
@@ -154,12 +154,6 @@ class CMeshSolver : public CFEASolver {
    */
   void SetBoundaryDisplacements(CGeometry *geometry, CNumerics *numerics, CConfig *config);
 
-  /*!
-   * \brief Set the value of the max residual and BGS residual.
-   * \param[in] val_iterlinsolver - Number of linear iterations.
-   */
-  void ComputeResidual_Multizone(CGeometry *geometry, CConfig *config);
```
pcarruscag: @rsanfer I changed CMeshSolver to use "BGS_k" instead of "Solution_Old" since the former is now allocated by CVariable whenever the problem is multizone, that allows the base implementation of 'ComputeResidual_Multizone' to be used. Do you foresee any problems?

Issue #789, opened 2019-09-09
Hybrid OpenMP-MPI implementation Strategy + Vectorization (SIMD)


pcarruscag:
### Preamble

I am moving the discussion about SIMD that started in #716 here and adding hybrid parallelization.
The two topics go hand in hand since both (SPMD and SIMD) consist of processing multiple data (MD) elements simultaneous, either by a single program (SP) that is run by multiple threads (generally with shared view of memory), or by a single instruction (SI) run by a single core.
The reason SIMD came up in #716 is that, as I will demonstrate, vectorization needs to be supported by data structures. On the other hand SPMD needs to be supported by algorithms designed to avoid race conditions, two or more threads modifying the same memory location.

Instead of continuing #716 I think it is better to let that become documentation for #753.
I did not add **without loss of readability** to the title because it is long as is, that requirement is present nonetheless.
I open these issues in the hope that people participate (I am not a fast writer so this is actually a lot of work) and so far great comments and insights have come from those with experience in these topics (kudos to @economon and @vdweide). But please participate even if you never heard of these topics, your opinion about readability and "developability" of the code is important! I think the code-style should be accessible to people starting a PhD (after they read a bit about C++...).

My (ambitious) goal with this work is to lay down an architecture for performance, i.e. not just to improve the performance of a few key numerical schemes but to create mechanisms applicable to all existing and future ones. Moreover I want that to be possible with minimal changes to the way those bits of code are currently written.

pcarruscag:
### Intro to SIMD
The ALU of modern CPU are capable of processing multiple elements of built-in types simultaneously by applying one instruction (e.g. add) to a register of those elements. Registers are at the very top of the memory hierarchy, for any computation to be performed data needs to be in registers.
An AVX register is 256 bits wide, that means 4 lanes of doubles or 8 of floats, AVX-512 (available in Xeon-Phi and SkylakeX processors) doubles the size. By GPU standards these are rookie numbers.

**Why should we care about SIMD?**
Because it is the only way to use the whole silicon, by and large vector instructions have the same latency and throughput of their scalar versions, therefore speedups proportional to the number of SIMD lanes are possible in compute-bound code.
As we saw in #716 there is some of that in the numerics, do not expect 4x speed-ups though, low order unstructured FVM is known to be bandwidth-bound, vectorization helps a bit there too (instructions are also data that needs to travel to the CPU) (maybe for explicit schemes and 8 SIMD lanes, maybe).

**Relation with data structures**
[...]

PR #790, opened 2019-09-14
New Nonlinear Iteration Controller for Improved Fluid Solver Robustness & Convergence

economon:
## Proposed Changes
 
This PR greatly improves the robustness and convergence behavior of the flow solvers by implementing a more sophisticated controller for nonlinear iterations along with several other important modifications.

The changes are the following:
- New under-relaxation strategy. Variables such as density and energy are only allowed to change by +/- 20% with each nonlinear iteration by applying an under-relaxation parameter to the updates at each node, which helps with difficult transients, especially at calculation startup. Also applied to turbulence model updates (though threshold is larger).
- CFL adaption via an exponential progression with under-relaxation approach, i.e., changes to CFL are coupled to the under-relaxation parameter (small under-relaxation parameter = decrease, and large = increase). CFL is now locally defined for each CV for this process.
- Modified realizability checks to avoid negative density/pressure/temperature at the nodes and also during 2nd-order reconstruction at cell faces.
- CV face states for 2nd-order MUSCL are now optionally reconstructed with a separate gradient method (unweighted least squares recommended), which is very important for robustness. Gradients for viscous terms and sources are still computed via GG or WLS for accuracy reasons.
- SA model was being clipped unnecessarily. This clipping has been removed.

Here is my recommendation on how to run all of your 2nd-order cases going forward:

### Euler / Laminar N-S
Turn off the limiters, unless you truly need them for preserving solution monotonicity (e.g., for shocks/discontinuities). Use 
```
NUM_METHOD_GRAD_RECON= LEAST_SQUARES
```
 for robustness. Start with a CFL around 10 and converge the implicit system with LINEAR_SOLVER_ITER= 10 for each nonlinear iteration (FGMRES + ILU0 recommended). Activate the CFL adaption with

```
% Parameters of the adaptive CFL number (factor down, factor up, CFL min value,
%                                        CFL max value )
CFL_ADAPT= YES
CFL_ADAPT_PARAM= ( 0.1, 2.0, 10.0, 1e10)
```

It is important that the linear solve behaves well for each nonlinear iteration, so if this does not converge immediately, adjust the tolerance or number of linear iterations. Roe seems very stable, but some of the other upwind schemes or centered schemes may require improved Jacobian accuracy (to recover fast convergence) with:

```
USE_ACCURATE_FLUX_JACOBIANS= YES
CENTRAL_JACOBIAN_FIX_FACTOR= 4.0
```

### RANS
Due to the segregated approach for the turbulence models, you have to be more conservative with the CFL adaption. All else is the same as above, but I recommend

```
% Parameters of the adaptive CFL number (factor down, factor up, CFL min value,
%                                        CFL max value )
CFL_ADAPT= YES
CFL_ADAPT_PARAM= ( 0.1, 1.2, 10.0, 1e3)
```

Most inviscid/laminar cases that I have tested so far converge to machine precision within a couple of hundred iterations. RANS cases within a few thousand. Ultra fine RANS meshes can still slow things down (MG could help there), but otherwise, speedups of an order of magnitude (or more) are possible for most problems. Experiment yourselves.. sometimes you can get even more performance, sometimes less, case depending.

I could use help testing this one with your cases, since I think almost all regression tests will change. As a start, I have run all of the NASA TMR 2D bump-in-channel cases. Everything looks good with that (see figures). I have also checked that the inviscid and laminar MMS give 2nd-order accuracy properly.

bump_cd_gridconv
bump_cf_0p63_gridconv
bump_cf_0p75_gridconv
bump_cf_0p87_gridconv
bump_cl_gridconv
residual_convergence

## Related Work

Gradient method strategy in #789 

## PR Checklist


jayantmukho:
YES! I am excited to try this out. I can probably test it on some of the other TMR cases (airfoils, flatplates). Will post the results when I get those done. 

Side note, there was one issue that @bmunguia and I encountered when performing optimizations with adaptive CFL. Say the DIRECT simulation is run with adaptive CFL and is well converged (6 to 8 orders of residual reduction). When the discrete adjoint performs the one direct iteration to store the computational graph, it uses the initial CFL value, not the CFL that the adaptive CFL routine ended at. This results in the residuals being very high for that one iteration, which then affects the convergence of the discrete adjoint. 

I will try to run an adjoint on one of these cases as well to see if the problem persists. Perhaps could be overcome with a simple additional field for CFL in the restart meta-data

economon:
> Do I understand correctly that two gradients are always computed per iteration?
> Is there any downside to using the unweighted LS for viscous flux correction? Is the statement that this type of gradient is better for reconstruction based on your observations or is it one of those well known things?

Yes - the gradient for now is computed twice and stored separately for viscous flows with 2nd-order upwind. Could be combined into one loop eventually.

It is known that weighted LSQ / GG is more accurate (see Mavriplis, "Revisiting the Least-Squares Procedure for Gradient Reconstruction on Unstructured Meshes" for instance). However, unweighted LSQ is known to be more robust.. so a good compromise is to use it only for the reconstruction step (which is more susceptible to robustness issues than the viscous term) and then use WLSQ or GG for all other gradients in the viscous flux/sources for accuracy.


> Sometimes high CFL leads to limit-cycle oscillations of the residuals and the solution is to reduce it, is this something this controller can pick up?
> High CFL also makes the linear systems harder to solve and as Edwin pointed out _somewhere_ there is not much point going above reasonable values with weakly coupled turbulence. Do you think it would be reasonable to build in some feedback from the linear solver (e.g. it is taking too much time or did not converge -> drop the CFL)?

Yes, I would also like to couple it to the linear solver so that we can remove the need to tune that as well. Ideally the user will not need to adjust parameters. There are some things I am going to try for that..

economon:
@pcarruscag: the approach to use different gradient methods for the convective and viscous terms is relatively common across codes. If it is a computational cost concern, this can be addressed by fusing the kernels. 

However, the bigger issue at hand is the other comment that is made in the article about the choice between LSQ and WLSQ/GG for 2nd-order upwind reconstructions: while less accurate, the former is much more robust, while the latter two usually require limiters just to obtain a stable solution on stretched RANS-type meshes, even for flows where we do not expect shocks/discontinuities. The problem with always requiring a limiter is that they stall convergence due to chatter, which apart from the obvious problems, also causes issues for the adjoint. I think everyone has experienced this.

Hence the compromise to use both gradients as I note above, which is also stated on p. 9 of Anderson and Bonhaus "An Implicit Upwind Algorithm for Computing Turbulent Flows on Unstructured Grids." In that paper they mention an additional interesting point that, in their numerical tests, LSQ outperforms WLSQ/GG for reconstructing nonlinear data at interfaces on highly-stretched meshes. Note the WLSQ/GG gradients are used for the viscous terms, which is important for accuracy. This type of approach is still applied in FUN3D. 

If there are any accuracy concerns, we can also address those via our typical verification tools. I agree that we should continue along the path of looking at grid quality issues and how they impact the numerics (potentially adding some grid-based corrections / limiting), but I think the proposed approach in this PR will serve us well.

pcarruscag:
Regardless of fusing routines computing two is more expensive than computing one, and the method is full of drawbacks, so I do not think it should be a forceful default.
Can this be implemented as a USE_ROBUST_GRADIENT option that if set to NO uses whatever gradient method is chosen for both convection and diffusion, computing and allocating only once? I do not care what the default is I just want to be able to turn it off.

EduardoMolina:
Hi @economon ,

Thanks for the effort in implementing this PR that is focus in robustness and convergence. 

However, I agree with @pcarruscag that we need to give to the user the ability to choose the gradient method that he/she thinks is appropriate. Although, this PR sounds promising we didn't test in complex geometries, i.e. High Lift PW or Drag PW (please correct me if I am wrong). Moreover, it may well be that in some cases the use of the more robust gradient method is not necessary since the CFL number is small for accuracy reasons, i.e. hybrid RANS/LES simulations like DDES.

Thanks again,
Eduardo

economon:
The motivation of having it as the default was to make the code as user-friendly as possible (fewer knobs exposed in the config), but options are good of course.

I would propose then that we add an option for the reconstruction gradient, something like:
```
NUM_METHOD_GRAD_RECON= LEAST_SQUARES
```
to let users decide if they want a separate option for the reconstruction gradients. If it does not appear, then the default is to use the same method as defined by NUM_METHOD_GRAD without a second gradient computation (basically what we have now). The nice thing about that is we can even try out other combos such as WLS+GG for the two different gradients. I will throw an error if users try to use LSQ for the viscous/source gradients, to avoid accuracy issues.

What do you think?

pcarruscag:
I agree with having suitable defaults that make the code robust, but "expert" options should still be available.
Having separate controls also sounds good, I would not throw an error though, maybe just a warning.

If you detect the two gradient options to be the same the associated CVariable classes make the reconstruction gradient point to the primitive gradient instead of allocating, and the call to compute the reconstruction gradient is skipped. If this is what you have in mind I support 100%.

koodlyakshay:
Hi @economon ,

I tested a couple of cases - laminar flow over a cylinder and turbulent flat plate and the convergence is greatly enhanced. I have attached a plot of convergence history for the flat plate case. I didn't have much luck with the laminar backward facing step case though. The residuals tends to oscillate around -5.5. I was planning to run the NACA 0012 test case soon. Do you have any cases specifically that you wanted to test? I can run some of them soon.
Convergence


Cheers,
Akshay

economon:
I have added the new option NUM_METHOD_GRAD_RECON to specify a separate method for computing the reconstruction gradient. If that option is not present, then no additional memory is allocated and no extra gradient computation occurs.

I have also put in simple feedback from the linear solver residual and the nonlinear residual to the nonlinear controller. If the linear system converges less than a half an order of magnitude, then the CFL is lowered. A Cauchy-like criteria checks for stall in the nonlinear residuals and drops the CFL to the minimum floor to kick the solver out of a rut. Both of these use factors that are empirical from my tests. Will probably be improved with time and more testing, but they do seem to improve behavior.

@koodlyakshay : I had success with the inc. laminar backward facing step after adding extra iterations to the linear solve. For some cases, this is necessary to get a large speedup. I am seeing good speedup for most of the cases within our TestCases repo. If you have some tough cases not covered by the repo, please give those a try.

pcarruscag:
Thanks for making the changes @economon, implementation looks ok.

economon:
Adding some additional residual convergence plots sampled from our test cases. Still some work to do to integrate #724 and clean up regressions.

residual_convergence
residual_convergence
residual_convergence

economon:
I have run some test to verify that the changes here also work ok for unsteady flows with dual time-stepping (even the CFL adaption seems to work pretty well for pseudo time). I also added some checks to make sure CFL adaption is disabled for TIME_STEPPING mode.

@EduardoMolina @clarkpede @cvencro: if you have some time, can you please test this branch out with some of your unsteady cases to verify that everything is ok for you as well? Let me know if you see any issues

jayantmukho:
I think it would be super useful to change some (or all) of the TestCases to include the suggested settings that you have outlined in this PR. Most users wouldn't know how to get the best performance out of the code otherwise.

Common/src/config_structure.cpp
```diff
@@ -2904,9 +2905,11 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
   }
 
 #ifndef HAVE_TECIO
-  if (Output_FileFormat == TECPLOT_BINARY) {
-    cout << "Tecplot binary file requested but SU2 was built without TecIO support." << "\n";
-    Output_FileFormat = TECPLOT;
+  for (unsigned short iOutputFile = 0; iOutputFile < nVolumeOutputFiles; iOutputFile++) {
+    if (VolumeOutputFiles[iOutputFile] == TECPLOT_BINARY) {
```
TobiKattmann: This probably won't catch `SURFACE_TECPLOT_BINARY ` for `VolumeOutputFiles[2]` (which is also .szplt format and should be catched right?). But I am already about to push the same thing 
bf2f6ef  to various_fixes #798 together with an SU2_MPI::Error to hinder this implicit conversion to TECPLOT(_ASCII).

economon:
> * From the history output it seems that the maximum CFL value is not being reached, even for small CFL values the update seems to stop one step before max.

There was a small factor applied to keep it above and below the CFL min and max (1.001, 0.999), respectively. But I have removed that, since it is not necessary. Btw, I have added the ability to print the min/max time step, and min/max/avg CFL to the screen by adding MIN_DELTA_TIME, MAX_DELTA_TIME, MIN_CFL, MAX_CFL, AVG_CFL to the SCREEN_OUTPUT.

> * I still see some residual oscillations with high CFL values (100-1000), I think the issue is that despite the oscillations the linear solver still converges fine, and so the reduction never kicks in.

There are reduction checks for both the linear and nonlinear residuals, but they likely still need some tuning (the factors were just empirical from my tests). I am also finding in my tests that if I am having trouble with a particular RANS case, sometimes it is most effective to turn off the adaption and go for a fixed CFL of 250, 500, or even 1000, if stable (don't forget you may need NUM_METHOD_GRAD_RECON= LEAST_SQUARES for that to be stable).

Common/src/config_structure.cpp
economon: Aok, thanks, @TobiKattmann. I will not touch this and go with your fix.

SU2_CFD/src/variables/CAdjEulerVariable.cpp
```diff
@@ -37,9 +37,8 @@
 
 #include "../../include/variables/CAdjEulerVariable.hpp"
 
-
 CAdjEulerVariable::CAdjEulerVariable(su2double psirho, const su2double *phi, su2double psie, unsigned long npoint, unsigned long ndim,
-                                     unsigned long nvar, CConfig *config) : CVariable(npoint, ndim, nvar, config) {
+                                     unsigned long nvar, CConfig *config) : CVariable(npoint, ndim, nvar, config), Gradient_Reconstruction(Gradient) {
```
pcarruscag: In c++11 you can have the initializer in the hpp file `VectorOfMatrix& Gradient_Reconstruction = Gradient;`

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -77,8 +78,14 @@ CVariable::CVariable(unsigned long npoint, unsigned long ndim, unsigned long nva
...
 	}
-
+  
+  /* Under-relaxation parameter. */
+  UnderRelaxation.resize(nPoint) = 1.0;
+  LocalCFL.resize(nPoint) = 0.0;
+  
+  /* Non-physical point (first-order) initialization. */
   Non_Physical.resize(nPoint) = false;
+  Non_Physical_Counter.resize(nPoint) = 0;
```
pcarruscag: Can you allocate these only where needed? Some of the classes inheriting from CVariable that use this constructor do not need these fields (DiscAdj variables, FEA variables). Maybe we need more overloads...

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -556,6 +570,30 @@ class CVariable {
...
+   * \brief Set the value of the local CFL number for the current control volume (CV).
+   * \param[in] val_cfl - the input value of the local CFL number for this CV.
+   */
+  inline void SetLocalCFL(unsigned long iPoint, su2double val_cfl) { LocalCFL(iPoint) = val_cfl; }
+  
+  /*!
+   * \brief Get the value of the local CFL number for the current control volume (CV).
+   * \return Value of the local CFL number for this CV.
+   */
+  inline su2double GetLocalCFL(unsigned long iPoint) { return LocalCFL(iPoint); }
```
pcarruscag: Thanks for adopting the `const` and `final` keywords for the gradient stuff, these two "gets" can be `const` too.

SU2_CFD/src/variables/CAdjEulerVariable.cpp
economon: Done

SU2_CFD/src/variables/CVariable.cpp
economon: Done

SU2_CFD/include/variables/CVariable.hpp
economon: Done

aeroamit:
Will it be a good idea to provide the optional control of these under-relaxation parameters to the users in configuration (.cfg) file? (ANSYS-Fluent does have such controls). It may be useful sometimes.

SU2_CFD/include/variables/CAdjEulerVariable.hpp
```diff
@@ -53,7 +53,7 @@ class CAdjEulerVariable : public CVariable {
   MatrixType IntBoundary_Jump;   /*!< \brief Interior boundary jump vector. */
   MatrixType HB_Source;          /*!< \brief Harmonic balance source term. */
 
-  VectorOfMatrix& Gradient_Reconstruction = Gradient; /*!< \brief Reference to the gradient of the primitive variables for MUSCL reconstruction for the convective term */
+  VectorOfMatrix& Gradient_Reconstruction = Gradient_Aux; /*!< \brief Reference to the gradient of the primitive variables for MUSCL reconstruction for the convective term */
```
pcarruscag: > Looks like the choice of initializer does affect the solution, btw. Looks like the reconstruction gradient vector should use the auxiliary vector's initializer for correctness...

@economon I don't see how that can happen, you always set `Gradient_Reconstruction` in the body of the constructor, the initializer is just a compilation trick.
By correctness do you mean it crashes or it does not use the gradient it is set to?

economon: It doesn't crash, but the solution is very different between those two option (residuals and converged forces, e.g., drag is tripled). So, I am guessing it is using the wrong gradient container somehow or the there is some issue with storage of the components. A quick way to see it is to change the VectorOfMatrix& Gradient_Reconstruction = Gradient_Aux; in CEulerVariable to VectorOfMatrix& Gradient_Reconstruction = Gradient_Primitive; and run a RANS type case.

pcarruscag: Ok, I'll see if I can replicate, it is a bit of a contrived thing that VectorOfMatrix... If it is causing issues I'll try to find a better solution.

pcarruscag: Damn I know what is happening... I gave you wrong information, references can be a bit more pernicious than pointer in that they cannot be reseated. i.e. when in the body of the ctor you write
`Gradient_Reconstruction = "something";` it does not change what the reference refers too but instead assigns `"something"` to the variable the reference currently refers too...

pcarruscag: @economon the solution is to move the conditional to the initializer list
`Gradient_Reconstruction{config->GetReconstructionGradientRequired() ? Gradient_Aux : Gradient}`
which if you like can pull from branch feature_nlctrl_reffix.
The only other alternative I see is to use a pointer, but then all the other methods need to be changed...

economon: Aok, let me pull in your reffix branch and keep testing. I am still verifying SST cases (Euler, laminar N-S, and SA look good so far).

SU2_CFD/src/solver_direct_turbulent.cpp
```diff
@@ -3608,8 +3611,8 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
     
     kine  = nodes->GetSolution(iPoint,0);
     omega = nodes->GetSolution(iPoint,1);
-    zeta = min(1.0/omega, a1/(strMag*F2));
-    muT = min(max(rho*kine*zeta,0.0),1.0);
+    zeta  = min(1.0/omega, a1/(Omega*F2));
```
clarkpede: This is technically a different version of the SST model.  For reference, the old version was SST-2003 and the new version is SST on [NASA LaRC's TMR model page](https://turbmodels.larc.nasa.gov/sst.html).  The paper cited on SST-2003 doesn't provide justification for the differences between versions.

There are actually a number of changes between SST and SST-2003.  I reviewed the changes I could find (model constants, production limiter, and CD_kw), and everything in the code now seem to match SST (and not SST-2003).  So I think you're consistent.

Is there a particular reason you chose the SST version over the SST-2003 version?

economon: This one is more straightforward.. I think consistency is the most important item here. We advertise the model as the "standard" SST model, and so it should be. As you know, folks take the differences in the models seriously, at least in the aeronautics world, which is important for expectations and reproduction of results. Seems to me that we only had a mismatch in the one term.

I think it would be great to also have more variants of the SST model, like we do for SA, since I think the later variants have become more commonly used, at least in some commercial codes. As long as we implement and catalog them as they appear on the TMR, they would be nice additions.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
```diff
@@ -0,0 +1,482 @@
...
+    WriteMarkerProfileTemplate();
+  } else {
+    ReadMarkerProfile();
+  }
+  
+}
+
+CMarkerProfileReaderFVM::~CMarkerProfileReaderFVM(void) { }
+
+void CMarkerProfileReaderFVM::ReadMarkerProfile() {
```
TobiKattmann: Again out of curiosity: Why don't you use the same file-structure as the Surface-Output uses? The inlet-files without column description can be a bit cryptic .

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4940,30 +5062,14 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
-    /*--- Close the file and start the loading. ---*/
-
-    inlet_file.close();
-
     /*--- Read the profile data from an ASCII file. ---*/
 
-    Read_InletFile_ASCII(geometry[MESH_0], config, profile_filename);
-
+    CMarkerProfileReaderFVM profileReader(geometry[MESH_0], config, profile_filename, KIND_MARKER, nCol_InletFile);
+
```
TobiKattmann: trailing whitespace

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -1612,30 +1622,26 @@ void CIncEulerSolver::Postprocessing(CGeometry *geometry, CSolver **solver_conta
...
+
     /*--- Incompressible flow, primitive variables ---*/
     
     physical = nodes->SetPrimVar(iPoint,FluidModel);
 
-    /*--- Record any non-physical points. ---*/
-
-    if (!physical) { nodes->SetNon_Physical(iPoint,true); ErrorCounter++; }
+    /* Check for non-realizable states for reporting. */
+
```
TobiKattmann: trailing whitespace

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -3512,7 +3566,51 @@ void CIncEulerSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **sol
...
 }
 
-void CIncEulerSolver::SetPrimitive_Gradient_GG(CGeometry *geometry, CConfig *config) {
+void CIncEulerSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, CConfig *config) {
+  
+  /* Loop over the solution update given by relaxing the linear
+   system for this nonlinear iteration. */
+  
+  su2double localUnderRelaxation = 1.0;
+  const su2double allowableRatio = 0.2;
```
TobiKattmann: Btw these 20%, is it your experience or is it from a source?

SU2_CFD/src/solver_direct_mean_inc.cpp
```diff
@@ -7471,18 +7607,14 @@ unsigned long CIncNSSolver::SetPrimitive_Variables(CSolver **solver_container, C
...
+
     /*--- Incompressible flow, primitive variables --- */
 
     physical = static_cast<CIncNSVariable*>(nodes)->SetPrimVar(iPoint,eddy_visc, turb_ke, FluidModel);
     
-    /*--- Record any non-physical points. ---*/
-
-    if (!physical) { nodes->SetNon_Physical(iPoint,true); ErrorCounter++; }
+    /* Check for non-realizable states for reporting. */
+
```
TobiKattmann: trailing whitespace

config_template.cfg
```diff
@@ -753,6 +753,10 @@ MARKER_ANALYZE_AVERAGE = MASSFLUX
 % Numerical method for spatial gradients (GREEN_GAUSS, WEIGHTED_LEAST_SQUARES)
 NUM_METHOD_GRAD= GREEN_GAUSS
 %
+% Numerical method for spatial gradients used for upwind reconstruction
+% (NONE, GREEN_GAUSS, LEAST_SQUARES, WEIGHTED_LEAST_SQUARES)
+NUM_METHOD_GRAD_RECON= LEAST_SQUARES
+%
```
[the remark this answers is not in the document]
pcarruscag: That is material for the docs page IMO

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1939,6 +1977,28 @@ class CVariable {
...
+   */
+  inline virtual su2double GetGradient_Reconstruction(unsigned long iPoint, unsigned long val_var, unsigned long val_dim) const { return 0.0; }
+  
+  /*!
+   * \brief Set the value of the primitive gradient for MUSCL reconstruction.
+   * \param[in] val_var - Index of the variable.
+   * \param[in] val_dim - Index of the dimension.
+   * \param[in] val_value - Value of the gradient.
+   */
+  inline virtual void SetGradient_Reconstruction(unsigned long iPoint, unsigned long val_var, unsigned long val_dim, su2double val_value) {}
```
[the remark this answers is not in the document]
pcarruscag: Good point... maybe an assertion just for debug purposes, do you want to try doing that @TobiKattmann ? We should be conscious of compilation times though... Adding all those checks will not be free.

SU2_CFD/src/variables/CHeatFVMVariable.cpp
```diff
@@ -39,7 +39,7 @@
 
 
 CHeatFVMVariable::CHeatFVMVariable(su2double heat, unsigned long npoint, unsigned long ndim, unsigned long nvar, CConfig *config)
-  : CVariable(npoint, ndim, nvar, config) {
+  : CVariable(npoint, ndim, nvar, config), Gradient_Reconstruction{config->GetReconstructionGradientRequired() ? Gradient_Aux : Gradient} {
```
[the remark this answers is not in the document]
pcarruscag: You did not post the complete error message... But I think I got it: https://gcc.godbolt.org/z/pUhCEu
I think it is a compiler bug, it is recognizing the {...} as an initializer list (which it is not) can you try changing it to (...) instead and see if that compiles?
Also... maybe update your compiler xD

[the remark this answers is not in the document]
TobiKattmann: Yep, normal braces are doing the trick! Thanks a lot (and sorry for beeing so lazy. That Website directly went to my bookmarks). Does anything speak against changing to normal brace? I can also do that and push... also for the cosmetic stuff @economon 

> Also... maybe update your compiler xD

Yeah. I think that every other day, but installing that on our cluster is not that straight forward for people with limited skillz :(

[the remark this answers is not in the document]
pcarruscag: Initially I tried normal braces and had some issue which I could not replicate with that example. I'll have a closer look. (I figured it would be a cluster thing)

SU2_CFD/include/variables/CVariable.hpp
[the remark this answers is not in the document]
TobiKattmann: Ok, a possibility could be using pure virtual functions https://gcc.godbolt.org/z/Ht2SO0 
That would require each child class to provide an implementation for it though. That is probably reasonable for some but not all parent methods.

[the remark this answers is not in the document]
pcarruscag: That creates more problems than it solves. And maybe some defaults are being used.
Are the empty implementations causing you problems?
Less intrusive solution is to return NaN in methods returning su2double and assert(false) on empty methods... NaN's and nullptr's are fairly easy to trap at runtime.

config_template.cfg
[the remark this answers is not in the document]
economon: Jury is still out whether it will necessarily be less accurate.. it will be very case-dependent. 

Nevertheless, I will be adding some text to the docs about tuning the solver. Essentially the same content that I have written in the PR description above.

Common/src/config_structure.cpp
```diff
@@ -4555,6 +4551,80 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+      Kind_Gradient_Method_Recon     = Kind_Gradient_Method;
+    }
+    
+  }
+  
+  /* Simpler boolean to control allocation of least-squares memory. */
+  
+  if ((Kind_Gradient_Method_Recon == LEAST_SQUARES) ||
+      (Kind_Gradient_Method_Recon == WEIGHTED_LEAST_SQUARES) ||
+      (Kind_Gradient_Method       == LEAST_SQUARES) ||
```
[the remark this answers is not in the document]
economon: Will leave for now. I wrote it so that it would work correctly for LEAST_SQUARES in case somebody really wants to use it in the future. They would only need to remove that error.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -93,7 +93,11 @@ class CVariable {
   MatrixType External_Old;   /*!< \brief Old external (outer) contribution in discrete adjoint multizone problems. */
 
   su2vector<bool> Non_Physical;  /*!< \brief Non-physical points in the solution (force first order). */
-
+  su2vector<unsigned short> Non_Physical_Counter; /*!< \brief Number of consecutive iterations that a point has been treated first-order. After a specified number of successful reconstructions, the point can be returned to second-order. */
+  
+  VectorType UnderRelaxation;  /*!< \brief Value of the under-relxation parameter local to the control volume. */
+  VectorType LocalCFL;         /*!< \brief Value of the CFL number local to the control volume. */
```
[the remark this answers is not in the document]
economon: I did not try that.. I am not sure if much can be gained since the equations are coupled and (most often) limited by the convective spectral radius (u + c). However, could be interesting to look into in the future. It would also require separate under-relaxation parameters for each equation too in order to drive the independent CFLs

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -6500,8 +6396,12 @@ void COutputLegacy::SpecialOutput_ForcesBreakdown(CSolver *****solver, CGeometry
 
   unsigned short FinestMesh = config[val_iZone]->GetFinestMesh();
   unsigned short nDim = geometry[val_iZone][INST_0][FinestMesh]->GetnDim();
-  bool flow = ((config[val_iZone]->GetKind_Solver() == EULER) || (config[val_iZone]->GetKind_Solver() == NAVIER_STOKES) ||
-               (config[val_iZone]->GetKind_Solver() == RANS));
+  bool flow = ((config[val_iZone]->GetKind_Solver() == EULER) ||
```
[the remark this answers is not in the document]
economon: I'll let @talbring comment on his plans for that file. Personally, I would prefer we not carry it around too long, and rather, if we need something there, folks should implement it cleanly within the new structure.

SU2_CFD/src/variables/CHeatFVMVariable.cpp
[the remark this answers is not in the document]
economon: Any news on this issue? I think it is good that you are testing on older compilers, since unfortunately, many folks will be constrained by older systems, especially clusters. Portability and easy install are critically important for us.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
economon: As you have noticed, I have moved this into its own reader class, like the mesh readers, since I feel it fits more in line with input rather than output (the output is really just an extra feature to help with preparing your own file). 

Another issue is that the variables and in particular the markers that we need for the output are not readily available - we only process the markers for output that are found in MARKER_PLOTTING at the moment. We may change this in the near future, but for now, the cleanest way to implement the inlet profile files in my opinion was with the new class I have added. It still has a bug or two according to the regressions, but I am working on that. I left the format the same for now, but adding a variable header line could be interesting..

SU2_CFD/src/variables/CHeatFVMVariable.cpp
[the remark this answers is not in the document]
pcarruscag: sorted

SU2_CFD/src/solver_direct_mean_inc.cpp
economon: Both. There are a few papers using similar approaches that I can point you to, if interested. Folks seem to use something in the 20-30% range. For the turbulence variables (actually, just SA), I did not limit it as much, since the turbulence variables tend to change rapidly during the initial transients of a calculation

SU2_CFD/src/variables/CHeatFVMVariable.cpp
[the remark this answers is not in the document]
TobiKattmann: > sorted? 

Is it ok then to change to normal braces? I tried multiple versions of gcc & clang with your provided link and couldn't and didn't get any errors with normal braces. Or do have issues with normal braces

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
TobiKattmann: > but adding a variable header line could be interesting..

Yep that would be great at some point.

SU2_CFD/src/output/output_structure_legacy.cpp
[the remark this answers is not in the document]
TobiKattmann: > Personally, I would prefer we not carry it around too long

Yep, me too

SU2_CFD/src/solver_direct_mean_inc.cpp
TobiKattmann: You already provided the source from Mavriplis above. If you have some other feel free to post it

SU2_CFD/src/output/output_structure_legacy.cpp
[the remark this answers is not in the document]
talbring: Only the turbomachinery and HB driver still use the old output structure ... if we would find someone willing to integrate these features into the new drivers (and the new output) we could remove all of these legacy stuff.

SU2_CFD/include/variables/CVariable.hpp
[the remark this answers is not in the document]
TobiKattmann: Well, I had some cases where I silently called the empty parent class implementation when changing function argument lists ... together with inattentive programming I managed to shoot in my foot few times. Using `override` can already help sometimes.
Do you see any drawbacks of returning NaN/nullptr or assert(false)? If no, I would give it a try.

SU2_CFD/src/output/output_structure_legacy.cpp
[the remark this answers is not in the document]
TobiKattmann: *ducking down :)

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
TobiKattmann: If you do not plan to add variable header lines in this PR (which would be fine for me) then you can mark this as resolved.

SU2_CFD/include/variables/CVariable.hpp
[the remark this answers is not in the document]
pcarruscag: only drawback is if somewhere the code relies on the default implementation being safe... I guess we will find out if that is the case

economon:
@jayantmukho : I am finding that the clipping limits for the SST model are very important for the UQ cases. If you adjust them slightly, the UQ regression cases tend to diverge immediately. Don't think any immediate action is needed, just wanted to bring it to your attention.

jayantmukho:
> @jayantmukho : I am finding that the clipping limits for the SST model are very important for the UQ cases. If you adjust them slightly, the UQ regression cases tend to diverge immediately. Don't think any immediate action is needed, just wanted to bring it to your attention.

Mhmmm, that's a little odd. I wouldn't think that the UQ methodology would be affected by the clipping. I will look into this. Just to be clear, you are changing the lowerlimit and upperlimit in CTurbSSTSolver constructor? What are you changing them too? Just want to reproduce the issue

economon:
> > @jayantmukho : I am finding that the clipping limits for the SST model are very important for the UQ cases. If you adjust them slightly, the UQ regression cases tend to diverge immediately. Don't think any immediate action is needed, just wanted to bring it to your attention.
> 
> Mhmmm, that's a little odd. I wouldn't think that the UQ methodology would be affected by the clipping. I will look into this. Just to be clear, you are changing the lowerlimit and upperlimit in CTurbSSTSolver constructor? What are you changing them too? Just want to reproduce the issue

If you revert my last commit, you will see the issue. It diverges right away for me, so perhaps it is just something with the initial transient that is caught by the clipping

economon:
@salvovitale @talbring : I have run the turbomachinery cases to convergence, and everything seems to be working fine. There are some non-negligible differences in the solution due to the changes to the SST model, the this is to be expected. So, I have updated all of the regression values. 

However, due to the legacy output, the turbo cases do not currently write restart files, which means a couple of the cases should be updated with proper restarts (and SU2_SOL is complaining that it can't find the restarts).

Can you please give it a look when you have a moment?

talbring:
@economon Output should be fixed for now. Volume output uses the new structure. History output still the legacy one. I am working with @salvovitale to also move that to the new structure, but that won't happen before the release of v7.

economon:
> @economon Output should be fixed for now. Volume output uses the new structure. History output still the legacy one. I am working with @salvovitale to also move that to the new structure, but that won't happen before the release of v7.

Aok. Thanks for the fix and the update.

economon:
I have updated the sliding mesh cases now, but I have found two cases that had odd behavior. The rotating cylinders case seems to give different results for values such as lift and drag depending on the core count, even in the develop branch. Also, the supersonic vortex shedding problem reports that hundreds of points are non-physical at the end of the run, both with this PR and in develop.

Anyone have any insight on these? I am going to keep moving through the final cases that need updated for this PR.

pcarruscag:
I've noticed those cases to be very sensitive in other PR's too, one of them had "diverged" reference residuals for a while. I never checked if the results were consistent with different number of cores, but from memory I think the inner iterations were not enough to converge each timestep, so results may well change with other settings too.

talbring:
Yep, these cases are not converged at all in every time-step. So I guess that this is reason for the behavior.

economon:
Oddly enough, I changed the number of inner iterations to a high value so that it would converge each physical time step, and I still see non-deterministic output for the rotating cylinders case. 

Perhaps these cases need to be flagged for checking if some folks using the sliding mesh interface have some time.

jayantmukho:
Hey everyone! Just posting the results from one of the NASA TMR verification test cases I have been running: [the DSMA661](https://turbmodels.larc.nasa.gov/airfoilwakeverif500c.html).

Using Least Squares (LSQ) for the gradient reconstruction does affect the results and can be seen from this force convergence plot. Using the same gradient reconstruction (GG) gives more accurate results. The two SU2 runs were done using identical linear solver parameters. The only difference is that for the GG+LSQ case, I specified : `NUM_METHOD_GRAD_RECON= LEAST_SQUARES` and for the GG case, I had to reduce the max CFL down to ~20 for it to get stable. 

cfl_adap_force_convergence_dsma661_SA

But due to the much lower maximum CFL, the GG run takes 40-50x longer to reach the same level of convergence. I'm not sure what else can be done to make adaptive CFL better for the GG or WLSQ cases.

pcarruscag:
Cool results @jayantmukho, few questions (which maybe you have answered somewhere already)
- If someone wants to play around with the same case, are your configs available somewhere?
- The grid levels in the plots correspond to the meshes available through that link right?
- Did you run the other solvers yourself? If so how does the runtime compare?
- The non monotonic variation with LSQ+GG is a bit weird, did the case converge to the same level? (I had some cases that at very high CFL would stagnate and not converge properly)

economon:
Nice work, and thanks for sharing, @jayantmukho. I expect you'll add everything to the V&V repo and eventually the V&V tab once we understand what is happening?

The GG results look great, potentially even better than the previous set of results we had for the DSMA case. I am also surprised to see the behavior of the GG+LSQ.. the finest mesh seems to be especially errant. I have been running the NACA 0012 case, and I also see that the results with pure GG or WLSQ are slightly better there, but not this drastic.

I think we still need to dial in the LSQ and make sure we do not have any bugs. Although, the flat plate and bump-in-channel cases were run with LSQ and gave very good results. It could be that too much curvature in the grid, especially near walls (which is known to be a potential problem), is causing these issues for the LSQ accuracy, but I'm surprised it would be that significant. Still looking into this...

As for adaptive CFL, sometimes I find that just turning it off for some rans cases and using a fixed 250, 500, or 1000 works best. Does that work for you with these cases with GG or WLSQ, @jayantmukho ?

jayantmukho:
- The meshes are available here: https://github.com/su2code/VandV/tree/master/rans/dsma661
- The config files with the GG + LSQ results are here: [dsma661_configs.zip](https://github.com/su2code/SU2/files/3842634/dsma661_configs.zip)

- I did not run the other solvers myself. Those are results from the NASA TMR website. I don't have access to those solvers unfortunately. 

- Yeah the non-monotonic variation worried me as well. But all the simulations were converged to a density residual of -13. (all of them have over 6 orders of residual reduction). The residuals for the GG+LSQ results shown here:
res_SA
I tried a few re-runs of that finest mesh with a couple of different options. I reduced the maximum CFL to ~30 from 1000 and it still gave the same result. I reduced the number of cores I was running on from 80 to 20 and that had no effect either. As soon as I ran without the `NUM_METHOD_GRAD_RECON= LEAST_SQUARES` option, it got a C_L of about 0.159402 which would be more in line with the other solvers.

@economon I haven't tried a high fixed CFL. Let me check that behavior and report back.

jayantmukho:
@economon  The behavior without adaptive CFL is similar to its non-adaptive behavior. I cannot increase the CFL greater than a certain value and get convergence. For example, for the 2nd coarsest mesh (297 x 57 or 129) I cannot increase the CFL > 20. This is regardless of weather I use adaptive CFL or not. I cannot set the CFL higher than 20. 

But this limit increases slightly for the finer meshes. For example, for the finest mesh I can push the CFL up to 30 (adaptive or otherwise). Basically cannot get high CFLs for the GG reconstruction. I should also mention that all these runs are without slope limiters to get the most accurate solutions.

economon:
To prepare for v7, I have now updated most of the tutorials to take advantage of the new features in this PR. I have also completed a sweep through all regression tests using this branch and have verified that the tests are ok. So, from my point of view, this is ready to be merged, other than updating the final residuals for a few tests which are pending.

I will keep looking into the LSQ issues, so for now, we should be cautious with that option.

economon:
Ok, it's finally time to get this one merged! Thanks all for the comments and help.

I'll merge this one in now as-is. I will leave the Tutorials branch in the regressions script set to the 'release_v7.0.0' branch, since we are collecting all changes to the website there for the release. @talbring @rsanfer : can you please revert that back to develop in the Release v7 PR when you are ready (after releasing the web updates)?

PR #777, opened 2019-08-27
WIP General Framework for Scalar Transport Equations

economon:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
 
This PR adds the necessary base classes (solver, numerics, variable) for having an arbitrary number of scalar transport equations added to flow problems in a segregated manner (alongside a turbulence model, if necessary). Most of the numerics is shared with the scalar methods for the turb. solvers to eliminate duplication, and it should be easy now to add child classes for any new scalar transport models.

In this PR, only a passive scalar has been implemented, but we are also building on this framework to add a flamelet model for combustion as a child class. A passive scalar test case will be added.

It still needs more work, but it is important to have the skeleton merged now to lessen the burden of fixing conflicts with the ongoing restructuring in driver, variable, output, etc. 

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

N/A

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.

PR #798, opened 2019-10-10
Various fixes

talbring:
## Proposed Changes
This PR fixes a couple of bugs I found recently. I also adapted the config preprocessing output to reflect the new output and convergence criteria options. 

Furthermore it fixes the artifacts mentioned by @TobiKattmann in #774 regarding the Heatflux sensitivity.   

Feel free to add any other fixes that you might came across.

## PR Checklist

- [] I have added a test case that demonstrates my contribution, if necessary.


TobiKattmann:
@oleburghardt @talbring This PR fixes the multicore heat-flux-sensitivitiy issues. Tested with the tutorial made by ole. Good job

I added a commit that fixes the error-message if you want to have tecplot-binary output but compiled using the --disable-tecio flag. It caused a compile time error ... it is behind a preprocessor statement #ifndef HAVE_TECIO, so this thing is only seen when using the --disable-tecio flag.

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -90,18 +90,27 @@ class AccessorImpl
    * runtime internal value that depend on the number of rows/columns.
    * What values need setting depends on the specialization as not all have
    * members for e.g. number of rows and cols (static size optimization).
+   * \note Aligned dynamic allocation is disabled for compilation on MAC.
    */
-#define REAL_ALLOCATOR(EXTRA)                                           \
+#ifndef __APPLE__
```
pcarruscag: Addresses the problem reported by @economon on #753

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -265,6 +265,9 @@ void CSysMatrix<ScalarType>::Initialize(unsigned long nPoint, unsigned long nPoi
     ILU_matrix = new ScalarType [nnz_ilu*nVar*nEqn];
     for (iVar = 0; iVar < nnz_ilu*nVar*nEqn; iVar++) ILU_matrix[iVar] = 0.0;
 
+    invM = new ScalarType [nPointDomain*nVar*nEqn];
```
pcarruscag: Addresses segfault found by @economon when using ILU with fill-in greater than 0.
Maybe in #790 one of the testcases can be modified to use this feature?

Common/include/toolboxes/C2DContainer.hpp
economon: This now build on Mac, but I now see issues with the assert() calls on the first access of one of these containers at the start of the solver

init.py
```diff
@@ -112,11 +112,11 @@ def submodule_status(path, sha_commit):
...
-      # Check that the SHA tag stored in this file matches the one stored in the git index
-      cur_sha_commit = status[1:].split(' ')[0]
-      if (cur_sha_commit != sha_commit):
-        print('SHA-1 tag stored in index does not match SHA tag stored in this script.')
-        sys.exit(1)
+    # Check that the SHA tag stored in this file matches the one stored in the git index
+    cur_sha_commit = status[1:].split(' ')[0]
+    if (cur_sha_commit != sha_commit):
+      print('SHA-1 tag stored in index does not match SHA tag stored in this script.')
+      sys.exit(1)
```
economon: Has the index been updated to have these new SHA tags for meson and ninja? I am still debugging #790 upon merging #753 #777 and this #798, and now when I try to build with meson, I get this error. Could be that my local copy still has some issues due to improper merge, or could be that this isn't updated yet

economon:
I also hit a segfault in one of my cases at this line in COutput for parsing the history per surface:

https://github.com/su2code/SU2/blob/af52ffd252282de14813424bd2e94defe1f49e63/SU2_CFD/src/output/COutput.cpp#L1566

Looks like a check for directdiff()/AD around that block is missing (I ran into it with the primal incompressible solver).

Common/include/toolboxes/C2DContainer.hpp
pcarruscag: @economon I need more details, I do not have a Mac to go test this myself.

economon: Here's what I am seeing in Xcode (trace on left and assertion failure at bottom). Just let me know what else I can provide to help with this...

Screen Shot 2019-10-16 at 12 12 50 AM

pcarruscag: Humm, is this after you updated #790? (I think I recognise the function name) If so that probably means the data is not being allocated. Push your changes and I can have a look directly in #790.

init.py
talbring: Previously the tags in git and in the init.py script where not matching and the error was not thrown because the indentation was wrong ... try `git submodule update` or deleting `meson`, `ninja`, `codi`, `medi` in the `externals` folder.

talbring:
> I also hit a segfault in one of my cases at this line in COutput for parsing the history per surface:
> 
> https://github.com/su2code/SU2/blob/af52ffd252282de14813424bd2e94defe1f49e63/SU2_CFD/src/output/COutput.cpp#L1566
> 
> Looks like a check for directdiff()/AD around that block is missing (I ran into it with the primal incompressible solver).

I removed the corresponding code lines. I have to think about how I want to do that, but its nothing urgent at the moment.

Common/include/toolboxes/C2DContainer.hpp
pcarruscag: Do note that asserts are disabled when building with ninja unless `buildtype=debug`.
You may want to define NDEBUG for your optimized builds in Xcode as the cost of those checks is not negligible.

economon: Noted. I typically only build in serial debug mode in Xcode and keep a separate build at the command line in release mode w/ MPI, but I can add flags for this.

Btw, I also noticed that the new C2DContainer class has the assignment operator overloaded to perform a deep copy. Is it possible to do a shallow copy, i.e., just set the pointer of an existing initialized container to an existing one without allocating memory?

pcarruscag: With what is implemented it is not possible. If the objective is avoiding the double allocation of gradients, I would suggest using a VectorOfMatrix& that is set to primitive gradient when no allocation is needed, otherwise a different VectorOfMatrix object is resized and the the reference is set to that instead. This way there are no ownership problems (who frees the pointer and so on). References need an initializer but that can be either of the other containers.
If it's for a different purpose I can think of something.

economon: Aok, let me give that a try (yes it is for the gradients). #790 is looking good after merging everything, will be pushing the cleaned version shortly

init.py
economon: I continued to have this issue, so I simply reset the stored SHA in the index to the one that you have in init.py. Should be all good. Will be adjusted in #790

talbring: It seems like you changed the version of ninja and meson in your branch. Maybe just make sure to use the newest ones if they work without any problem.

economon:
I noticed that the output phase values for WRT_PERFORMANCE= YES are currently off due to the number of output phases not being correctly tracked (we call the SetResultsFiles in the Iteration class most of the time, and the one in the driver with the timing only at the end of the run).

@talbring, did you already have something in mind for this? Otherwise, I can think of a cleaner way to handle those performance benchmarks. I think it would be good to start printing that information by default too potentially

Common/include/toolboxes/C2DContainer.hpp
economon: Looks like the choice of initializer does affect the solution, btw. Looks like the reconstruction gradient vector should use the auxiliary vector's initializer for correctness...

talbring:
I tried to solve the problem where sometimes the regression tests stall ... it seems to be a problem of openmpi. Changed it to mpich, but then mpi4py has problems. I reverted it for now. We have to live with that problem for a while unfortunately.

Common/src/config_structure.cpp
```diff
@@ -1994,8 +1994,6 @@ void CConfig::SetConfig_Options() {
   addBoolOption("DEFORM_CONSOLE_OUTPUT", Deform_Output, false);
   /* DESCRIPTION: Number of nonlinear deformation iterations (surface deformation increments) */
   addUnsignedLongOption("DEFORM_NONLINEAR_ITER", GridDef_Nonlinear_Iter, 1);
-  /* DESCRIPTION: Number of smoothing iterations for FEA mesh deformation */
-  addUnsignedLongOption("DEFORM_LINEAR_ITER", GridDef_Linear_Iter, 1000);
```
pcarruscag: I got rid of this option in favor of the equivalent and more descriptive DEFORM_LINEAR_SOLVER_ITER

oleburghardt:
@pcarruscag 
I added a commit that improves the bad names for the new routines that we discussed #803. They should be more instructional now.
Morever, at some places (default) function arguments are used instead of different functions, which hides the changes a bit and makes it more elegant.

Common/src/interpolation_structure.cpp
```diff
@@ -688,19 +689,24 @@ void CNearestNeighbor::Set_TransferCoeff(CConfig **config) {
...
+            if (Buffer_Receive_GlobalPoint[Global_Point_Donor] != -1){
+              
+              Coord_j = &Buffer_Receive_Coord[ Global_Point_Donor*nDim];
+              
+              dist = PointsDistance(Coord_i, Coord_j);
+              
+              if (dist < mindist) {
+                mindist = dist; pProcessor = iProcessor; 
+                pGlobalPoint = Buffer_Receive_GlobalPoint[Global_Point_Donor];                
+              }
```
TobiKattmann: Thanks again @talbring @rsanfer (and everyone involved I might not know about) for finding & fixing this bug!

talbring:
Yep from my side we can merge this in

oleburghardt:
@TobiKattmann Good idea to eventually just open up another one.

pcarruscag:
And... we broke travis again

PR #803, opened 2019-10-15
Re-enable Discrete Adjoint FSI Test Cases

rsanfer:
## Proposed Changes

This PR is intended to put some functionality back that was involuntarily removed in #724. 

Test cases for discrete adjoint FSI are now re-enabled with split files. They [correctly pass](https://travis-ci.org/su2code/SU2/builds/597983817) the travis tests [when the branch is based in feature_input_output](https://github.com/su2code/SU2/compare/feature_input_output...654ba3dfc207e25c31da9c09860224ad4044a610), which was the original branch merged in #724. 

Unfortunately, given that during this time both  #753 and #774 were merged, the tests are failing now. Particularly, the disc_adj_fsi case runs into a segfault in `CDiscAdjSolver::ExtractAdjoint_CrossTerm_Geometry`, when trying to execute 
`geometry->node[iPoint]->GetAdjointCoord(Solution_Geometry);`. 

I believe this is related to the changes to the index based recording introduced in #774, as the code was running before into the same problem in `CDiscAdjSolver::ExtractAdjoint_CrossTerm`, and now goes beyond this point having changed the Get routine into `direct_solver->GetNodes()->GetAdjointSolution_intIndexBased(iPoint,Solution);`.

Currently, there is no `GetAdjointCoord_intIndexBased` or similar available, and its implementation requires some constructs for the indices. Therefore, I would like to ask for support from @oleburghardt and @pcarruscag to make this work with their changes.

## Related Work
#724 involuntarily removed the discrete adjoint FSI cases from the regression tests.
Branch [fix_discadjfsi](https://github.com/su2code/SU2/tree/fix_discadjfsi) successfully recovers the functionality removed.
#774 introduced relevant changes to the adjoint solver that are making.
#753 was merged and, although I tried to adapt to it, I can't guarantee it's working correctly.

## PR Checklist


pcarruscag:
Thank you for bringing this back. The testcase was passing on #753 alone, so that should not be the problem, I'll have a look.

oleburghardt:
@rsanfer You shouldn't get in touch with the new AD routines at all unless you're using `CDiscAdjMultizoneDriver`.
So in case `direct_solver->GetNodes()->GetAdjointSolution_intIndexBased(iPoint,Solution);` gets executed, that likely means that the if-statement above, `config->GetMultizone_Problem()`, evaluates to true.
Some weeks ago I added another boolean (Get_MultiphysicsDiscreteAdjoint or similar) but I was asked to revert that to the line above.

pcarruscag:
I'm getting to the same conclusion, the easier step might be to set
Multizone_Problem to false during recording steps...

Ole Burghardt <notifications@github.com> escreveu no dia terça, 15/10/2019
à(s) 12:42:

oleburghardt:
> I'm getting to the same conclusion, the easier step might be to set
Multizone_Problem to false during recording steps...

That would then break `CDiscAdjMultizoneDriver`, doesn't it?

I'd suggest a more precise boolean there (as introduced [here](https://github.com/su2code/SU2/pull/774/commits/e632133eded5edab7ff04f0979334c25c0f44ff8) as long as we have multiphysics cases that use own drivers.

pcarruscag:
It does not break the multizone adjoint but it does not fix the problem.

So now when we register the coordinates we do it index-based because the multizone_problem boolean is set.
What are the implications for when we extract the adjoints? Do we need get an index-based GetDerivative?

oleburghardt:
Sure we do (e.g. `CDiscAdjSolver::RegisterSolution` will behave incorrectly).

So actually both ways are based on saving indices somewhere. However the routines that I added (with the `_intIndexBased` extension) do save them internally together with "their" corresponding variable (in the same variable class) which is important for the multizone stuff.

> So now when we register the coordinates we do it index-based because the multizone_problem boolean is set.

We don't want the `_intIndexBased` routines for the FSI case. They work by re-recording new tapes for each set of variables (fluid solution/coordinates/displacement solution) with added specific routines within the solvers/iterators for cross dependencies (which goes along with saving indices externally, but in a preassigned order).

> What are the implications for when we extract the adjoints? Do we need an index-based GetDerivative?

Yes there are counterparts in `RegisterSolution`, `SetAdjoint_Output` and `ExtractAdjoint_Solution`. But nothing more. We can go for the internal indices in `CVariable` exclusively once we have integrated the FSI capabilities to the multiphysics driver.

rsanfer:
@oleburghardt I figured why `direct_solver->GetNodes()->GetAdjointSolution_intIndexBased(iPoint,Solution)` was being executed, but as the FSI problem is multi-zone adjoint, it is required that `config->GetMultizone_Problem() = true` even when we are not using the driver `CDiscAdjMultizoneDriver`. 

There is, of course, the possibility to set a specific boolean for these test cases in particular, but I think that would over-complicate the code. Else, it should be possible to extend the index based to the rest of the features required (geometry and structural solvers), which would be my preferred option.

oleburghardt:
Ok. I don't think [these](https://github.com/su2code/SU2/pull/774/commits/e632133eded5edab7ff04f0979334c25c0f44ff8) changes are over-complicated.

But I'm fine with changing the index saving procedure everywhere. There'd be just some work to do (we would have to adapt everything for the FEA elasticity solvers and all the variables they need as well)... so it's also a question of time, testing and so on.

rsanfer:
I think one option would add yet one more config option and make it more difficult for the user, and the other would be to extend new features to all existing capabilities of the code. The idea behind the single and multi-zone drivers was precisely generalization. In my opinion it would be a no brainer to go for the second, but I'll leave it up to the community, but it depends also a lot on what timing we are moving in.

pcarruscag:
Is it possible for the fluid side to work index-based and the structural side as before?
We are combining the cross terms manually and re-recording each step anyway...

oleburghardt:
> Is it possible for the fluid side to work index-based and the structural side as before?
> We are combining the cross terms manually and re-recording each step anyway...

(Please rather say internal-based, or variable-based - both approaches are index-based, they differ in the way they are stored :-))

In principle, yes. One can have both at the same time. I'd have to think about it, sounds a bit messy to me right now to get it all consistent.

One could also change `GetMultizone_Problem()` to `GetMultizone_Problem() && !GetFSI_Problem()` or similar at those places. Anyway, we know the reason for this problem, so I'll leave it up to you which way we go.

> but it depends also a lot on what timing we are moving in.

Yes.

pcarruscag:
@rsanfer , @oleburghardt , I think I fixed it, I'll do further validation tomorrow.

oleburghardt:
> The "int" in "intIndexBased" is for internal then? Because its type is also int, easy mistake to make xD.

Yes.. The name was the first one I gave that routine. It somehow made it through.. Now that I had to type it several times I'd love to have it changed. But anyway..

I'm a bit puzzled that it seems to be so easy but maybe it's just as simple as you said - new approach inside `CDiscAdjSolver` and old in `CDiscAdjFEASolver` (if I got that correctly?). That would come in handy for all further developments. Let's wait for the validation. I'll also do one with this branch for the CHT adjoints tomorrow, just to be sure.

rsanfer:
Thanks a lot for your support, @pcarruscag. 

I run the other test cases this morning and I noticed there is a very small difference in the sensitivities around the 10th significant figure (in the order of the updates you made on the discrete adjoint airfoil case). Given that the order of magnitude of this difference remains consistent even when extending the simulation, I updated the test values. I leave here the [reference to the previous state](https://github.com/su2code/SU2/commit/654ba3dfc207e25c31da9c09860224ad4044a610) for our records. 

If there are no comments against it in the next day or so, I will be merging in this PR next, as it just puts back some functionality that was removed.

pcarruscag:
Does anyone know why we get this random failures from travis?

> No output has been received in the last 20m0s, this potentially indicates a stalled build or something wrong with the build itself.

It does not seem to be related with a particular test case...

talbring:
> Does anyone know why we get this random failures from travis?
> 
> > No output has been received in the last 20m0s, this potentially indicates a stalled build or something wrong with the build itself.
> 
> It does not seem to be related with a particular test case...

Unfortunately that happens a lot lately. Haven't figured out what the reason is. I assume its something with MPI on the virtual machines in Travis. Hard to debug though, I can test a different MPI version maybe...

oleburghardt:
> I'll also do one with this branch for the CHT adjoints tomorrow, just to be sure.

Just finished it. Everything's ok.

PR #806, opened 2019-10-25
Regression Testing using Github Actions

talbring:
## Proposed Changes
This PR adds a regression testing pipeline based on the new Github actions features (https://github.com/su2code/SU2/actions). It uses docker containers for building and running the tests (https://hub.docker.com/r/su2code/build-su2 and https://hub.docker.com/r/su2code/test-su2, respectively, based on Ubuntu18.04/gcc7.4/OpenMPI2). This has several advantages over the current (Travis based) system:

- Using the containers you can run the building/testing environment locally on your machine. You can even download the compiled binaries produced on github and run them locally to further debug problems.
- It is faster and more flexible. Compiling and running all tests requires ~1h since a lot of things run in parallel now (e.g. building the base, reverse AD and forward AD binaries). There is potential to further speed things up by splitting the test case scripts into several files. All of that requires minimal work. Unfortunately there is no caching using ccache at the moment, but it is likely that github introduces this feature mid of november. 
- Additional tests like unit tests can make use of the same docker containers and can be easily set up by creating a new workflow file.
- We can easily add a deployment step to automatically create releases and/or nightly builds.

**Note:** the linked docker containers are **only** for development/testing purposes. Hence we cannot give any support in this regard. For production use it is recommended to use singularity containers. But: the docker containers can be used as basis layers for singularity. I will get back with more information on that topic.

 ~~I had to change regression test values of two cases since apparently they are compiler dependent.~~ I changed it to the same versions as we have on Travis.

There will be some documentation on how to use the containers locally soon.

## PR Checklist


talbring:
The pipeline is ready now. It is possible to have both systems enabled for some time if necessary. They don't interfere with each other at all.

TobiKattmann:
As it does not interfere with the rest of the code I guess nothing speaks against merging this.
Couple of things though:
- the badge does not work [here](https://github.com/su2code/SU2/tree/new_testing_system) the badge is not rendered  and the link is dead. I get to a `Not Found` page. 
- My only docker-knowledge is literally an 8min [youtube](https://www.youtube.com/watch?v=pGYAg7TMmp0) video I just watched so I might have gotten s.th. wrong, but: where are the Dockerfiles (probably small textfiles again) for build and test? The two images on hub.docker.com are already ... well images, so fairly large. The beauty of travis was also that the .travis.yml file contains everything necessary to build and run in a rather straightforward to understand manner -> and therefore one was able to quickly change things here and there. Are those Dockerfiles intended to go [here](https://github.com/su2code/build-action)? The Dockerfiles should go into some sort of version control right?

Speeding up "time-to-finished-regression-test" is really cool though. 👍 

🐳 Take these two whales as a sign of gratitude 🐋

talbring:
> * the badge does not work [here](https://github.com/su2code/SU2/tree/new_testing_system) the badge is not rendered  and the link is dead. I get to a `Not Found` page.

The badge is for the develop branch. But the tests never ran so far for that branch, thats why there is no badge yet.

> * My only docker-knowledge is literally an 8min [youtube](https://www.youtube.com/watch?v=pGYAg7TMmp0) video I just watched so I might have gotten s.th. wrong, but: where are the Dockerfiles (probably small textfiles again) for build and test? The two images on hub.docker.com are already ... well images, so fairly large. The beauty of travis was also that the .travis.yml file contains everything necessary to build and run in a rather straightforward to understand manner -> and therefore one was able to quickly change things here and there. Are those Dockerfiles intended to go [here](https://github.com/su2code/build-action)? The Dockerfiles should go into some sort of version control right?

I am gonna remove the build_action repo. I created that because I thought it would be needed. But its not anymore. Good point about the files. Actually forgot to add them to the repo :+1: The nice thing is the only thing executed in the container is a bash script that you can run also without docker, in case you need to.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: |
+          mkdir -p install/bin
+          cp -r ${{format('Base{0}', matrix.tag)}}/. install/bin/
+          cp -r ${{format('Reverse{0}', matrix.tag)}}/. install/bin/
+          cp -r ${{format('Forward{0}', matrix.tag)}}/. install/bin/
+          chmod a+x install/bin/*               
+      - name: Run Tests in Container
+        uses: docker://su2code/test-su2
+        with:
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```
TobiKattmann: Alright, here are -c Testcases and -t Tutorail case repos specified (as explained in runTests.sh) and the only place to be changed(?)

.github/workflows/regression.yml
```diff
@@ -0,0 +1,85 @@
+name: Regression Testing
+
+on: 
+  pull_request:
+    branches: 
+      - 'develop'
+      - 'master'
```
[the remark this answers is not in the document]
talbring: I just added the push option. You'll have to add your branch to that list.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,85 @@
...
+          - config_set: ReverseMPI
+            flags: '-Denable-autodiff=true -Denable-normal=false -Denable-pywrapper=true'
+          - config_set: ForwardMPI
+            flags: '-Denable-directdiff=true -Denable-normal=false'          
+          - config_set: BaseNoMPI
+            flags: '-Denable-pywrapper=true -Dwith-mpi=disabled'
+          - config_set: ReverseNoMPI
+            flags: '-Denable-autodiff=true -Denable-normal=false -Dwith-mpi=disabled -Denable-pywrapper=true'
+          - config_set: ForwardNoMPI
+            flags: '-Denable-directdiff=true -Denable-normal=false -Dwith-mpi=disabled'
```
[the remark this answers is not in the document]
talbring: Every job in that matrix (and every job in general on actions) runs in total isolation from each other. But every build job uploads its binaries/scripts at the end. The regression test jobs then download the binaries they need (labeled BaseMPI, ReverseMPI, ForwardMPI, BaseNoMPI, ReverseNoMPI, ForwardNoMPI)

TobiKattmann:
Not mission critical but in the raw logs each line has some lengthy timestamp. Is it possible to get rid of that?

`2019-10-29T14:12:02.9434263Z sim_vals (computed): -1.293870, 4.141024, -0.002288, 0.008868`

Sidenote: It is actually quite nice that build and run are in separate logs

talbring:
> * Not mission critical but in the raw logs each line has some lengthy timestamp. Is it possible to get rid of that?
> 
> `2019-10-29T14:12:02.9434263Z sim_vals (computed): -1.293870, 4.141024, -0.002288, 0.008868`

I don't think so. Maybe they'll add that in the future.

> * Can you say a few words about resources we get from github? #parallelcores/processes, time-per-month.

20 concurrent jobs. There is no limitation in the time-per-month for public repositories. Every job has 2 cores, 7GB of memory and 14GB of disk space.

Source:
https://help.github.com/en/github/automating-your-workflow-with-github-actions/about-github-actions#usage-limits
https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#supported-virtual-environments-and-hardware-resources



> * Sidenote: It is actually quite nice that build and run are in separate logs

talbring:
Anyone else has comments ? Like @TobiKattmann already said, if we find that there are some drawbacks with that approach we can easily switch back.

.github/workflows/regression.yml
```diff
@@ -0,0 +1,89 @@
...
+          - config_set: ForwardNoMPI
+            flags: '-Denable-directdiff=true -Denable-normal=false -Dwith-mpi=disabled'  
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dump matrix context
+        env:
+            MATRIX_CONTEXT: ${{ toJson(matrix) }}
+        run: echo "$MATRIX_CONTEXT"
+      - name: Build 
+        uses: docker://su2code/build-su2:20191029
```
economon: What is the significance of pulling from the docker with the date 20191029? Any issue there, or will it always be the latest (and just a naming thing)?

This appears again below.

talbring: Putting there a specific tag makes sure that we don't accidentally break the regression tests, because anything without a specific tag will be tagged as "latest" once you push it to the docker repository.

talbring: So it just refers to a specific "named" version of the image (i thought the date would make sense).

talbring:
Then I would say we give it a try.

PR #809, opened 2019-11-01
Multi-physics and FEA discrete adjoint tweaks

pcarruscag:
## Proposed Changes
Some tweaks to the new multi-physics discrete adjoint framework plus the beginning of changes required to have FSI adjoints within it. No legacy code has been removed or disabled, yet...

All explained below.

## PR Checklist


SU2_CFD/include/solver_structure.hpp
```diff
@@ -12742,27 +12742,40 @@ class CDiscAdjFEASolver : public CSolver {
...
-  su2double *Local_Sens_EField, *Global_Sens_EField; /*!< \brief Local and global sensitivity of the Electric Field. */
-  su2double *Total_Sens_EField;
-
-  bool fea_dv;                  /*!< \brief Determines if the design variable we study is a FEA parameter. */
-  unsigned short nDV;           /*!< \brief Number of design variables in the problem. */
-  su2double *DV_Val;            /*!< \brief Value of the design variables. */
-  su2double *Local_Sens_DV, *Global_Sens_DV;          /*!< \brief Local and global sensitivity of the Design Variable. */
-  su2double *Total_Sens_DV;
+            *Rho_DL_i;          /*!< \brief Values of the density (for volume loading). */
+  int       *AD_Idx_E_i,        /*!< \brief Derivative index of the Young's Modulus. */
```
pcarruscag: It seems that for FSI adjoints the derivatives w.r.t. structural variables need extraction via local indexes (don't ask me why).

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -233,13 +223,28 @@ void CDiscAdjMultizoneDriver::Run() {
...
       Add_Solution_To_ExternalOld(iZone);
 
+      /*--- If the contents of BGSSolution are valid we initialize the inner iterations from them by moving
+       *    them to Solution, otherwise Solution contains only the contributions from the OF.
+       *    Note: This is not 100% correct as on the first inner iteration the updated cross terms
+       *          (from the previous zone) are missed. ---*/
+
+      if (iOuterIter > 0 || driver_config->GetRestart())
+        Set_Solution_To_BGSSolution(iZone);
+
```
pcarruscag: This is a semi-important change to have continuity from the last inner iteration to the first inner iteration of the next outer iteration. The solution in iZone gets overwritten when cross terms are extracted for another zone, so we need to restore from the BGS solution.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -768,8 +821,17 @@ void CDiscAdjMultizoneDriver::ComputeAdjoints(unsigned short iZone) {
...
-  AD::ComputeAdjoint(TRANSFER, OBJECTIVE_FUNCTION);
+  /*--- Compute adjoints of transfer and mesh deformation routines,
+   *    only needed on the last inner iteration. ---*/
+
+  if (eval_transfer)
+    AD::ComputeAdjoint(TRANSFER, OBJECTIVE_FUNCTION);
+
+  /*--- Adjoints of dependencies, needed if derivatives of variables
+   *    are extracted (e.g. AoA, Mach, etc.) ---*/
+
```
pcarruscag: This is an efficiency tweak for cases with mesh deformation, for which we do not need to compute adjoints on every inner iteration.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -286,36 +295,31 @@ void CDiscAdjMultizoneDriver::Run() {
...
-    
     /*--- Set the multizone output. ---*/
 
     driver_output->SetMultizoneHistory_Output(output_container, config_container, driver_config,
                                               driver_config->GetTimeIter(), driver_config->GetOuterIter());
 
     /*--- Check for convergence. ---*/
 
-    StopCalc = driver_output->GetConvergence();
+    StopCalc = driver_output->GetConvergence() || (iOuterIter == nOuterIter-1);
```
pcarruscag: This slightly different stopping logic is to make sure files are outputted.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -286,36 +295,31 @@ void CDiscAdjMultizoneDriver::Run() {
...
-      
-      output_container[iZone]->SetHistory_Output(geometry_container[iZone][INST_0][MESH_0],solver_container[iZone][INST_0][MESH_0], config_container[iZone]);
-      
-    }
 
-    /*--- Now all coupling terms are summed up, set External_Old to External for next outer iteration. ---*/
+      /*--- Now all iZone coupling terms are summed up, set External_Old to External so the
+       *    next zone receives the contributions from iZone (BGS-type iteration). ---*/
+
+      Set_OldExternal();
```
pcarruscag: @oleburghardt the residuals for the cht case changed probably due to moving this call into the loop over zones, which as the effect of carrying the cross-dependencies to the next zone earlier. If you could make sure all is ok it would be great.

oleburghardt: > External still accumulates over all zones
Yes but it has to accumulate over all zones for all zones.

nZone=2 in fact is not the decisive thing but one would have to require that a set of solution variables has influence on at most one other zone. 
Which would be alright to assume for the types of couplings that we have but it's not as general (and hard to debug once we are there).

Imagine a case with 3 zones and each one has an influence on all the others. Then at iZone=1, we would update adjoints in zone 1 *and* overwrite External for zone 2 and 3 with cross dependencies given through solvers in zone 1.
 But External in zone 2 could also have had an contribution from a possible dependence of solvers in zone 3 on the solution in zone 2.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -441,47 +440,21 @@ void CDiscAdjMultizoneDriver::SetRecording(unsigned short kind_recording, Kind_T
...
-    for (jZone = 0; jZone < nZone; jZone++){
-      /*--- The target zone is iZone ---*/
-      if (jZone != iZone && interface_container[iZone][jZone] != NULL) {
-        DeformMesh = DeformMesh || Transfer_Data(jZone, iZone);
-      }
-    }
-    /*--- If a mesh update is required due to the transfer of data ---*/
-    if (DeformMesh) DynamicMeshUpdate(iZone, ExtIter);
-  }
+  HandleDataTransfer();
```
pcarruscag: @oleburghardt @rsanfer @cvencro 
FSI adjoints seem to be working with the new driver but with legacy deformation and transfer classes (at least the regression case works in serial, needs more testing).
If you want to try it out you need to change line 127 of SU2_CFD.cpp to instantiate the new driver, the "legacy" ADJ_FSI option still needs to be set.

@oleburghardt one key change is this one I highlight here, the transfer and deformation needs to be included also in the OF part of the tape as since the mesh deformation is not treated as a third zone some fluid objectives depend explicitly on structural variables through the transfer+deformation path.

SU2_CFD/src/solver_adjoint_discrete.cpp
```diff
@@ -479,7 +478,7 @@ void CDiscAdjSolver::ExtractAdjoint_Solution(CGeometry *geometry, CConfig *confi
 
   /*--- Set the old solution ---*/
 
-  nodes->Set_OldSolution();
+  if(!multizone) nodes->Set_OldSolution();
```
pcarruscag: This change is required for the zone-wise residuals to make sense, on entry Solution contains cross-terms and OF contributions but on extraction it does not, as the zone-wise residual is based on the before-after difference we were measuring the magnitude of the cross-terms and OF.
The dirty solution is to set OldSolution to the newly extracted one that does not contain those contributions yet.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -756,6 +756,7 @@ void CDiscAdjMultizoneDriver::SetObjFunction(unsigned short kind_recording) {
               ObjFunc += solver[FEA_SOL]->GetTotal_OFRefGeom()*Weight_ObjFunc;
               break;
             case TOPOL_COMPLIANCE:
+              static_cast<CFEASolver*>(solver[FEA_SOL])->Integrate_FSI_Loads(geometry, config);
               solver[FEA_SOL]->Compute_OFCompliance(geometry, solver, config);
               ObjFunc += solver[FEA_SOL]->GetTotal_OFCompliance()*Weight_ObjFunc;
```
pcarruscag: So FSI adjoints work, at least the derivatives wrt material properties are correct for reasonably complex cases.
One downside that should be addressed is that there is no easy way to introduce BGS relaxation, and that really limits the type of FSI cases we can solve.
The separate recording of the OF introduces the need to call some solver routines that are usually in the "compute path" before evaluating exotic functions that are computed in one zone but have indirect contributions from the other zones, this is probably not the norm but it needs to be kept in mind when adding new functions.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
oleburghardt: > Setting external_old = external does not reset the contents of external.

Yes we accumulate in `External` (for each combination of `iZone` and `jZone`) but cross contributions only enter the solution through `External_Old`.
So by the above example, if they are overwritten in iZone, they might be incorrect in iZone+1 if there is a third party.

> Unless I'm missing something, do we have an all-to-all multizone case to put this to the test?

Unfortunately not, we have a 4-zone case but only with at most one off-diagonal block per zone.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2588,16 +2586,14 @@ bool CDiscAdjFluidIteration::Monitor(COutput *output,
...
-  
-  return StopCalc;
+  output->SetHistory_Output(geometry[val_iZone][INST_0][MESH_0],
+                            solver[val_iZone][INST_0][MESH_0],
+                            config[val_iZone],
+                            config[val_iZone]->GetTimeIter(),
+                            config[val_iZone]->GetOuterIter(),
+                            config[val_iZone]->GetInnerIter());
+
+  return output->GetConvergence();
```
pcarruscag: @talbring I could use your help with something, in some CIteration methods there is a lot of hardcoded ZONE_0 even when `iZone` is passed as argument, this particular case was causing segfaults when ZONE_0 was not a fluid zone.
Maybe there are others but I don't want to change them blindly, is there any rule for when to use ZONE_0 / INST_0 instead of iZone / iInst ?

talbring: whenever there is `val_iZone` and/or `val_iInst` passed as an argument it should be used in that function ...

talbring: Of course we could also think of just passing the correct pointers. Maybe that would be even better, since the iteration classes should only deal with individual zones. Will be part of a future restructuring of the CIteration structure.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -297,8 +300,15 @@ void CDiscAdjMultizoneDriver::Run() {
       Set_BGSSolution(iZone);
 
       /*--- Now all iZone coupling terms are summed up, set External_Old to External so the
-       *    next zone receives the contributions from iZone (BGS-type iteration). ---*/
+       *    next zone receives the contributions from iZone (BGS-type iteration) alternatively,
+       *    do it outside the "iZone loop" for Block-Jacobi type iterations. ---*/
+
+      if (MZSolver == MZ_BLOCK_GAUSS_SEIDEL) {
```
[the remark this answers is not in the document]
pcarruscag: How do I check that bloody gradient?!

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
cvencro: @pcarruscag I've tried with the new driver instantiated for FSI adjoints, gradients from both serial and parallel runs are accurate.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
oleburghardt: run `SU2_DOT_AD`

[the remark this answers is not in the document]
pcarruscag: I get very similar values around 1.39... for all combinations I tried and the fluid adjoints are almost exactly the same. Do I need to run SU2_DOT_AD zone by zone or something weird like that?
Nevertheless I FINALLY understood your point, this `External` vs `External_Old` and which one gets set / overwritten is a bit confusing and it does take a bit of an odd case to make it happen...
Anyway, I'll fix that and the relaxation we need for FSI.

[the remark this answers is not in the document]
oleburghardt: > Do I need to run SU2_DOT_AD zone by zone or something weird like that?

No, just run SU2_DOT_AD with the master config (check that it's set to `DISCRETE_ADJOINT`). If you have problems check out my [tutorial](https://github.com/su2code/su2code.github.io/pulls).

> Nevertheless I FINALLY understood your point, this External vs External_Old and which one gets set / overwritten is a bit confusing and it does take a bit of an odd case to make it happen...

It's nothing but constructing a counterexample.

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -79,6 +79,10 @@ class CDiscAdjMultizoneDriver : public CMultizoneDriver {
   unsigned short* direct_nInst;   /*!< \brief Total number of instances in the direct problem. */
   unsigned short* nInnerIter;     /*!< \brief Number of inner iterations for each zone. */
 
+  /*!< \brief Individual cross-terms of the coupled problem, 5D array [iZone][jZone][iSol](iPoint,iVar).
+              The column sum, i.e. along all iZones for each jZone, gives the External (total cross-term)
+              for jZone, we need to store all terms to have BGS-type updates with relaxation. */
+  vector<vector<vector<su2passivematrix> > > Cross_Terms;
```
pcarruscag: @oleburghardt, needs cleanup but I'll go with your suggestion for a nZone*nZone structure.

pcarruscag: @oleburghardt I think I've got it working. Thanks for pointing to the tutorial, now I know why all my values were close even when the code was wrong (10000 iterations is 3 orders of magnitude more than I am used to with FSI cases).

Just to summarize what I did, each cross term is stored separately in that `Cross_Terms` 5D array.
When a cross term is extracted, it is updated with relaxation (which we need in FSI but the default is no relaxation so no impact on CHT):
`x = alpha * x_new + (1-alpha) * x_old`
The delta in that update (`x-x_old`) is added to `External`, which carries the sum of all cross terms and the objective function gradient.
The latter is now only computed once, outside the outer iteration loop, and it initializes `External` (`External_Old` is not longer required).
In the inner iteration loop `External` is added to `Solution` before computing the adjoints instead of after, this means when exiting the inner loop `Solution` does not contain cross terms or the OF gradient.
`Solution_BGS` and `Solution_BGS_k` also do not contain `External` therefore by restoring `Solution_BGS_k` to `Solution` before the inner iterations, and adding `External` to `Solution` as the first thing we do in the inner loop, we pickup exactly from where we stopped before (that is the intention anyway, the whole thing might be wrong again).
The cross terms are stored in `passivedouble` as that occupies half the space, we can think of streamlining things further by not using BGS_k, which is not needed since BGS == Solution.

I will compare the converge rate of this type of update with the one you had before, if it is worse I will put in the option to run as before (that may require the OF term to be carried separately).

SU2_CFD/src/drivers/CSinglezoneDriver.cpp
```diff
@@ -209,11 +209,13 @@ void CSinglezoneDriver::Output(unsigned long TimeIter) {
 #else
   StartTime = MPI_Wtime();
 #endif
-    
+
+  bool force_write = StopCalc || !config_container[ZONE_0]->GetTime_Domain();
+
```
pcarruscag: @talbring , is it ok if I do this? I was not getting any file output for discrete adjoint fea cases (I think it would happen for any unconverged case in general) Let me know if there is a better way.

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
pcarruscag: Conclusions:
- No negative impact in CHT convergence rate, after the 10k iterations the residuals are nearly the same and the gradients (running everything on 1 core) are -0.719645 for develop and -0.719302 for this branch.
- No negative impact in correctness, converging the CHT case to below 10^-8 the gradient with this branch is -0.717685 whether the fluid is evaluated first or third. 
- Within this framework FSI adjoints seem not to need relaxation, which is great!

SU2_CFD/src/drivers/CSinglezoneDriver.cpp
talbring: Hm but this routine should not be called if the case is not converged anyway. The output routines in in CIteration are responsible for intermediate writing.

pcarruscag: It is not being handled by single zone adjoint in the Run method and the SecondaryRecording method, the latter being required for output of sensitivities.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -701,8 +701,10 @@ void CMeshSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig *
       for (iDim = 0; iDim < nDim; iDim++){
         /*--- Update the coordinates of the mesh ---*/
         curr_coord = Restart_Data[index+iDim];
-        /// TODO: "Double" deformation if this is set here?
-//        geometry[MESH_0]->node[iPoint_Local]->SetCoord(iDim, curr_coord);
+        /// TODO: "Double deformation" in multizone adjoint if this is set here?
+        ///       In any case it should not be needed as deformation is called before other solvers
+        ///geometry[MESH_0]->node[iPoint_Local]->SetCoord(iDim, curr_coord);
+
```
pcarruscag: For some reason if the coordinates are set here the multizone adjoint does not work properly, it seems to use a grid with the displacements applied twice.
It does not seem to be a problem to leave this out since the mesh deformation is called before the fluid solvers, the grid velocities are computed based on displacement, and I am now initialising the solution of the linear system (i.e. it will converge in 0 iterations when restarting).

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -342,7 +342,7 @@ void CIteration::SetMesh_Deformation(CGeometry **geometry,
   /*--- Perform the elasticity mesh movement ---*/
   if (config->GetDeform_Mesh()) {
 
-    if(kind_recording != MESH_DEFORM){
+    if ((kind_recording != MESH_DEFORM) && !config->GetMultizone_Problem()) {
       /*--- In a primal run, AD::TapeActive returns a false ---*/
```
pcarruscag: Mesh deformation needs to be recorded in FSI problems also for the conservative variables as it establishes the connection between structural displacements and fluid variables.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2984,6 +2985,12 @@ void CDiscAdjFEAIteration::SetDependencies(CSolver *****solver, CGeometry ****ge
     break;
   }
 
+  /*--- FSI specific dependencies. ---*/
+  if(fsi) {
+    /*--- Set relation between solution and predicted displacements, which are the transferred ones. ---*/
+    dir_solver->PredictStruct_Displacement(nullptr, config[iZone], solver[iZone][iInst]);
+  }
+
```
pcarruscag: Since we transfer "predicted displacements", and not directly the solution, there is this extra dependency.
@cvencro this may need some attention for time domain if a higher order predictor is used.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -424,8 +431,18 @@ void CDiscAdjMultizoneDriver::SetRecording(unsigned short kind_recording, Kind_T
...
+
+      unsigned short type_recording = kind_recording;
+
+      if (Has_Deformation(iZone) && (kind_recording == MESH_COORDS)) {
+        type_recording = MESH_DEFORM;
+      }
+
       iteration_container[iZone][INST_0]->RegisterInput(solver_container, geometry_container,
-                                                        config_container, iZone, INST_0, kind_recording);
+                                                        config_container, iZone, INST_0, type_recording);
```
pcarruscag: To extract geometric sensitivities we need to do this kind of ugly `kind_recording` conversion for zones with mesh deformation...

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
oleburghardt: Thanks for the summary. Just looked through the code and saw that is should all be correct.
But just to be on the safe side I repeated all the tests (fluid zone at different places, you remember) and they all give the gradient above.

SU2_CFD/src/solver_structure.cpp
```diff
@@ -2982,21 +2982,34 @@ void CSolver::SetSolution_Gradient_LS(CGeometry *geometry, CConfig *config) {
...
+  for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++) {
+    for (unsigned short iVar = 0; iVar < nVar; iVar++) {
+      passivedouble
+      new_val = SU2_TYPE::GetValue(base_nodes->GetSolution(iPoint,iVar)),
+      old_val = cross_term(iPoint,iVar),
+      relax_val = alpha*new_val + (1.0-alpha)*old_val;
+      /*--- Set new value of cross term. ---*/
+      cross_term(iPoint,iVar) = relax_val;
+      /*--- How much we need to update the sum of cross-terms. ---*/
+      Solution[iVar] = relax_val-old_val;
```
oleburghardt: @pcarruscag 
Thanks for this implementation. Technically it's alright I think, so I made it an approval.

I'd like to take another look at it (e.g. this took a me few minutes, as we are adding the new minus the old, effectively making it the correct update - it's correct but maybe one could point that out a bit clearer here).
If there's nothing from my side by the end of the week though let's merge it in.

pcarruscag: Thanks for reviewing, I'll make those comments extra explanatory.
I am in the process of re-doing the verification @cvencro and I did last year for FSI shape derivatives, so far it looks like results are better as with the new mesh solver from @rsanfer we can get geometric derivatives wrt to initial coordinates.
I'll post those results here once I am done and then we are good to go on my side too.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -227,19 +246,27 @@ void CDiscAdjMultizoneDriver::Run() {
...
 
-        Add_External_To_Solution(iZone);
+        if (no_restart || (iInnerIter > 0)) {
+          Add_External_To_Solution(iZone);
+        }
+        else {
+          /*--- If we restarted, Solution already has all contribution,
+           *    we run only one inner iter to compute the cross terms. ---*/
+          eval_transfer = true;
+        }
```
pcarruscag: On those last commits I tried to enable restarts, it is a bit tricky because we keep solution and cross terms separate during the solution and only combine them when needed. I think the logic is correct, the first outer iteration is used just to extract cross-terms, alas, it does not work perfectly (better than nothing, just not perfect...)

PR #815, opened 2019-11-10
Inlet Spanwise Interpolation

### LINKED ISSUE #791: Looking for a native piecewise linear interpolation function in SU2
Hello everyone,

I am starting to work on adding inlet file interpolation capabilities in SU2. I currently have a spanwise PT and TT distribution (r vs PT or TT) for an inlet boundary for a Stator (periodic, turbomachinery case). I want to carry out a spanwise interpolation at the inlet vertices.

I can write a program to read an inlet file for a new format (r vs PT or TT). Then I want to 1D interpolate this data against the radii of all the vertices on the inlet face. The best way is to generate a **piecewise 1D interpolation function** that can return the value of the interpolated data at that radius. I have done this before in python using 'interp1D'. I have found the following library for C++ that does have this capability:

http://people.math.sc.edu/Burkardt/cpp_src/pwl_interp_1d/pwl_interp_1d.html

which will require _'The PWL_INTERP_1D test code needs the R8LIB and TEST_INTERP libraries._'

I hope these libraries will be integrated with SU2 without any issue. However, it would be better if I can use some of the built-in interpolation capabilities in SU2 like what I found in 'interpolation_structure.cpp'. I also look forward to extending this capability for 2D interpolation(that is, in r and theta).

It would be very helpful if someone who has worked with interpolation before can point to a function that is capable of generating a piecewise interpolation function that could work for my case or can pitch in any idea related to this.

Thanks!

@vdweide  @clarkpede @economon @juanjosealonso @sravya91 @bmunguia @jayantmukho

auzbaig:
I have added 1D interpolation capability in SU2. It will be helpful in running cases with inlet profiles with radial variation. Additionally, I have added two classes CAkimaInterpolation and CLinearInterpolation which can be used for other purposes and can be initialized very simply by using pointers to C1DInterpolation class.

The following is the Inlet file format that you can give now:

radius, TT, PT, unit_Vr, unit_Vtheta, turb variables... OR
radius, TT, PT, alpha, phi, turb variables..

*alpha is the angle taken normal to the direction of flow (swirl angle), phi is the flow angle taken from the direction of flow (more details in the velocity triangle pdf attached).

The solver can then interpolate the data two ways: 1D Linear Interpolation or 1D Akima Interpolation.

I added three new options in .cfg file:
INLET_INTERPOLATION_FUNCTION = None, Linear_1D, Akima_1D
INLET_INTERPOLATION_TYPE = vr_vtheta, alpha_phi
PRINT_INLET_INTERPOLATED_DATA = yes/no (for those curious what is the exact interpolated data)

Given the generality of the approach, I have taken this will work for 2D cases or cases with multiple inlets without any issue.

I am attaching a sample test case with detailed examination of results shown below. I am also attaching my equations that I used. Note that I define theta counterclockwise positive from x-axis.

[testcase.zip](https://github.com/su2code/SU2/files/4217420/testcase.zip)

I changed the inlet file angle to degrees rather than radians for alpha,phi option. All the graphs below have been plotted with radians. I have included both radians and degrees inlet file in the testcase. From now on, the inlet angles have to be in degrees! I have plotted graphs to verify the conversion was successful but I am not posting the graphs here as it is already a lot of data.

For further details on Vr Vtheta and alpha, phi definitions you can refer this document: 
[velocity_triangles.pdf](https://github.com/su2code/SU2/files/4219369/velocity_triangles.pdf)

These are the coordinate conversion equations I have used. The right ones are the ones used in SU2 to convert to Vx, Vy the left ones are used in python script for comparison along with the equations from velocity triangles (reference: [Link](https://ocw.mit.edu/courses/aeronautics-and-astronautics/16-07-dynamics-fall-2009/lecture-notes/MIT16_07F09_Lec05.pdf)).

equations for coordinate conversion

Following are the plots for the interpolated data for every quantity. Akima is definitely smoother when it comes to interpolating with lesser data points.

phi comparsion
alpha comparsion
Pt comparsion
Tt comparsion
Turb comparsion

I am also attaching my python plotting results in case someone wants to reproduce the graphs and/or look at the equations in detail.

[plotting_script.txt](https://github.com/su2code/SU2/files/4217388/plotting_script.txt)

Please note that the inlet file was created with random fluctuating values to ensure correct inlet interpolation is happening. Some values and patterns may look unrealistic.

Additionally, note that Akima can extrapolate values as well but will display a warning if it is doing so. Linear will only do interpolation and will exit if the inlet file does not have the specific radius data. However, it will show on the console which radius data is missing so the user can add it.

Also, currently, the flow should be in z-direction only. I don't know how to find the flow direction through any built-in method in SU2 during run. If you can tell me that I can update the code to support any flow direction. During interpolation, it will display that flow direction must be in z-direction to avoid any mistake by the user.

Thanks,
Aman

## PR Checklist


talbring:
@auzbaig Thanks for opening this draft PR! Looking forward to your developments. If you have any specific questions let us know.

SU2_CFD/include/solver_structure.hpp
```diff
@@ -3595,6 +3596,51 @@ class CSolver {
...
+   */
+  inline passivedouble Get_Wi(unsigned long iRow,unsigned short index);
+
+    /*!
+   * \brief Get the weighed slope of the interval being interpolated by Akima.
+   * \param[in] iRow - Current matching row in inlet file for interpolation to Vertex
+   * \param[in] index - Current interpolating column in inlet file
+   * \param[in] jMarker - Current Marker to retrive the number of Rows for that Marker
+   */
+  inline passivedouble Get_Ai_dash(unsigned long iRow, unsigned short index,unsigned long jMarker);
```
pcarruscag: These would probably go to an interpolation class (toolbox) instead.

SU2_CFD/include/solver_structure.inl
```diff
@@ -1055,6 +1055,31 @@ inline void CSolver::SetNuTilde_WF(CGeometry *geometry, CSolver **solver_contain
...
+    else if (iRow == 1) {return (Get_Pi(iRow-1,index)+Get_Pi(iRow,index))/2;}
+    else if (iRow == nRow-2) {return (Get_Pi(nRow-2,index)+Get_Pi(nRow-3,index))/2;}
+    else if (iRow == nRow-1) {return Get_Pi(nRow-2,index);}
+    else{
+    if((Get_Wi(iRow+1,index)+Get_Wi(iRow-1,index)) != 0)
+      return (Get_Wi(iRow+1,index)*Get_Pi(iRow-1,index) + Get_Wi(iRow-1,index)*Get_Pi(iRow,index))/(Get_Wi(iRow+1,index) + Get_Wi(iRow-1,index));
+    else
+      return ((Get_Pi(iRow-1,index)-Get_Pi(iRow,index))/2);
+    }
+}
```
pcarruscag: This method is a bit large to be inline

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
...
+                    {
+                      myfile<<Inlet_InterpolatedData[iVertex*(maxCol_InletFile+nDim)+iVar]<<"\t";
+                    }
+                  myfile<<endl;
+                }
+        myfile.close();
+}
+  else
+    cout<<"file cannot be opened"<<endl;
+}
```
pcarruscag: tabs / spaces issue here? Also please try to be consistent with the rest of the code regarding where you leave the {}'s (I know it is not easy to define what consistent means, but we usually have `...) {`)

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
...
+                        switch(config->GetKindInletInterpolationFunction()){
+                          case (ONED_LINEAR_SPANWISE):
+                          /*--- 1D interpolate data for every index (column) using equation of a line. ---*/
+                          slope=(Inlet_Data[index+maxCol_InletFile*(iRow+1)]-Inlet_Data[index+maxCol_InletFile*iRow])/(Inlet_Data[maxCol_InletFile*(iRow+1)]-Inlet_Data[maxCol_InletFile*iRow]);
+                      
+                          /*--- If interpolating turbulence variables, shift them one index ahead ---*/
+                          if (index > nDim+1)
+                            Inlet_Values[index+(nDim-1)+1]=slope*(Interp_Radius - Inlet_Data[maxCol_InletFile*iRow]) + Inlet_Data[index+maxCol_InletFile*iRow];
+                          else
+                            Inlet_Values[index+(nDim-1)]=slope*(Interp_Radius - Inlet_Data[maxCol_InletFile*iRow]) + Inlet_Data[index+maxCol_InletFile*iRow];
```
pcarruscag: In situations like these it is better to handle just the index in the if statement to avoid repeating the entire line of code.
Also (micro optimization tip) if this is just to handle an index bound consider using `min/max` or transforming the conditional into arithmetic to avoid branching, i.e.:
```
offset = index > nDim+1;
Inlet_Values[index+nDim-1+offset] = ...
```

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
...
+#ifdef HAVE_MPI
+    SU2_MPI::Allreduce(&local_failure, &global_failure, 1, MPI_UNSIGNED_SHORT,
+                       MPI_SUM, MPI_COMM_WORLD);
+#else
+    global_failure = local_failure;
+#endif
+
+    if (global_failure > 0) {
+      SU2_MPI::Error(string("Vertex radius outside inlet file radius range."), CURRENT_FUNCTION);
+    }
```
pcarruscag: I think you can just throw the error locally (without the reduction). And you do not need the #ifdef HAVE_MPI, you can write the code as if MPI is available as the functions are silently replaced for serial compilation.

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
...
+    cout<<"file cannot be opened"<<endl;
+}
+
+
+void CSolver::LoadSpanwiseInletProfile(CGeometry **geometry,
+                                CSolver ***solver,
+                                CConfig *config,
+                                int val_iter,
+                                unsigned short val_kind_solver,
+                                unsigned short val_kind_marker)
```
pcarruscag: This method is quite long, it is well documented but please consider breaking it up in more manageable functions.
Especially the part that interpolates to coarse levels sounds like something that can be reused elsewhere.

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
...
+  unsigned short KIND_SOLVER = val_kind_solver;
+  unsigned short KIND_MARKER = val_kind_marker;
+
+  /*--- Local variables ---*/
+
+  unsigned short iDim, iVar, iMesh, iMarker, jMarker;
+  unsigned long iPoint, iVertex, iChildren, Point_Fine, iRow, nVertex;
+  passivedouble ai,bi,ci,di,delta,dxi; //akima interpolation variables
+  passivedouble Theta, Parameter1, Parameter2, unit_r, unit_Theta, unit_m, unit_z, Alpha, Phi;
+  su2double Area_Children, Area_Parent, *Coord, dist, slope, Interp_Radius;
```
pcarruscag: When moving the interpolation to a class make all parameters su2double as we might want to differentiate through it (it can also be a template class).

SU2_CFD/src/solver_structure.cpp
```diff
@@ -4825,6 +4825,365 @@ void CSolver::Read_InletFile_ASCII(CGeometry *geometry, CConfig *config, string
   
 }
 
+void CSolver::PrintNewInletData (passivedouble *Inlet_InterpolatedData, unsigned long nVertex,string Marker_Tag)
```
pcarruscag: Inlet_InterpolatedData can be `const passivedouble*`

SU2_CFD/include/solver_structure.inl
auzbaig: I had double thoughts about it too but I planned to include it anyway as it was consistent with other interpolation functions around it. I can probably remove it from inline.

SU2_CFD/include/solver_structure.hpp
auzbaig: So you want me to define all these functions in the interpolation class (interpolation_structure.cpp)? Won't it be simpler to stay in the solver class given I am calling everything from there?

SU2_CFD/src/solver_structure.cpp
auzbaig: Yes, this skipped my sight before. I have to fix it.

SU2_CFD/src/solver_structure.cpp
auzbaig: I get your idea. Will look into it.

SU2_CFD/src/solver_structure.cpp
auzbaig: I don't really have any experience with coding for MPI. This part of the code I got from the existing "LoadInletProfile" subroutine and I kept it the way it was. For some reason, my code is failing the Reverse/Forward No/MPI regression tests (4 tests) although it runs fine with no MPI. Do you have an idea how can I fix this?

SU2_CFD/src/solver_structure.cpp
auzbaig: The existing subroutine was "LoadInletProfile" and I reused the same part of the code here to get down to coarser grids.

SU2_CFD/src/solver_structure.cpp
auzbaig: Ok, will do.

SU2_CFD/src/solver_structure.cpp
auzbaig: Right.

SU2_CFD/src/solver_structure.cpp
pcarruscag: Ok can you please check if there is an easy way to factor that multigrid part out of both "LoadInlet" methods, so that both call a function to do that job instead of having the code copied.

SU2_CFD/src/solver_structure.cpp
pcarruscag: From the compilation logs it seems like it is failing because of mixing passivedouble and su2double, try making everything su2double.

SU2_CFD/include/solver_structure.hpp
pcarruscag: No not in interpolation_structure, that is for interfacing between zones.
Somewhere in Common/toolboxes as a generic 1D interpolation framework.

For now those functions are only used in solver but we should try to make things as modular as possible.

auzbaig: Right. That would mean that I would have to make and #include a new .hpp file in solver_structure.hpp?

SU2_CFD/src/solver_structure.cpp
auzbaig: I wanted to look into that but also didn't want to change the existing code. I will see if that can be done.

SU2_CFD/include/solver_structure.hpp
pcarruscag: That's right.

SU2_CFD/src/solver_structure.cpp
auzbaig: You were right. That solved the problem.

Common/include/toolboxes/inlet_interpolation_functions.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include <iostream>
+#include <string>
+#include <cmath>
+#include <vector>
+#include "../config_structure.hpp"
+#include "../../SU2_CFD/include/CMarkerProfileReaderFVM.hpp"
+
```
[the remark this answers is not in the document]
auzbaig: Thanks for commenting on my changes.

I agree that the Interpolator should ideally work as you are saying but in this case I need to fetch the vertex coordinates for the Interpolator to be able to interpolate over those points, that's why I passed the geometry pointer into the class.

I wanted the class to be 'branched' from the main program so that I could read the inlet file in the Interpolator class and not worry about passing the profileReader object from Solver to the Interpolator class.

This class can still be used in other contexts. I may decide to read the profile in the Solver only (now that I think more about it), but even without it its input arguments are generic.

Common/src/toolboxes/inlet_interpolation_functions.cpp
```diff
@@ -0,0 +1,206 @@
...
+      Parameter1=Inlet_Values[nDim+2];
+      Parameter2=Inlet_Values[nDim+3];
+
+      CorrectForInterpolationType();
+
+      Inlet_Values[nDim+2] = unit_r*cos(Theta) - unit_Theta*sin(Theta); //for ix
+      Inlet_Values[nDim+3] = unit_r*sin(Theta) + unit_Theta*cos(Theta); //for iy
+      Inlet_Values[nDim+4] = sqrt(1-pow(unit_r,2)- pow(unit_Theta,2));  //for iz
+
+      SetInterpolatedData();
```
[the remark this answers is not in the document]
auzbaig: I intentionally did that so that I may reduce the number of arguments passing to and fro between functions. It does make more sense to have clear inputs and outputs and I will change my functions accordingly.

Common/include/toolboxes/inlet_interpolation_functions.hpp
```diff
@@ -0,0 +1,85 @@
...
+    su2double Interp_Radius, Theta, *Coord;
+
+    su2double slope, interpolated_value, Parameter1, Parameter2, unit_r, unit_Theta, unit_m, Alpha, Phi;
+    su2double dxi, ai, bi, ci ,di ,delta;
+    
+    void LinearInterpolation(), AkimaInterpolation(unsigned long iRow_Akima), CorrectForInterpolationType(), SetInterpolatedData(), PrintInterpolatedData();
+    void Interpolate();
+    void GetClosestPointFromFile();
+    void SetVertex();
```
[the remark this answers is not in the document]
auzbaig: Yes this is something I delayed to fix at the end. Will try to be more consistent in the development phase so its easier for others to review.

Common/include/toolboxes/inlet_interpolation_functions.hpp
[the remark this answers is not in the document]
pcarruscag: > I agree that the Interpolator should ideally work as you are saying but in this case I need to fetch the vertex coordinates for the Interpolator to be able to interpolate over those points, that's why I passed the geometry pointer into the class.

Somewhere in the solver:
```
...
reader = ProfileReader(args)
switch(interpType) {
    case LINEAR:
        interpolator = LinearInterpolator(reader.data())
    case ...
}
for(vertex : marker) {
    interpValues = interpolator(vertex.GetCoord())
    // set the values somewhere
}
...
```
Why is something like that not possible? The profile data contains the coordinates right? Does each interpolated value depend on all "output" locations?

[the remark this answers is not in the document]
auzbaig: This can work. For such an approach, is it possible to just have a set of functions in .cpp and .hpp files which interpolate somewhat like this: LinearInterpolator(reader.data( ) , vertex.GetCoord( )). I have tried this approach in previous commits but abandoned it as I felt that since this subroutine (LoadInletProfile(args)) was changed in the newer v7.0 version, its better to branch out completely with a class rather than accommodate any more changes that happen without any prior notice. Is it ok to have a set of functions in the toolboxes rather than a class?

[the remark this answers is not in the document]
pcarruscag: I suppose that would be ok for the implementation methods you have, but it is not ideal because:
- You cannot force a common interface for all methods;
- For some interpolation methods (splines, radial basis functions, etc.) some non-trival computation needs to be performed, and ideally stored for re-use every time the object is called (even in linear interpolation you may want to build a three to quickly find the closest points).

That said, I do not object to functions, the optimum is the enemy of the good :)

>  I have tried this approach in previous commits but abandoned it as I felt that since this subroutine (LoadInletProfile(args)) was changed in the newer v7.0

Protect it with a testcase... No feature is safe without a regression for it :smiling_imp:.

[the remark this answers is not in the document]
auzbaig: > That said, I do not object to functions, the optimum is the enemy of the good :)

I agree. Now that I have the class l will on forward and make it work.

> Protect it with a testcase... No feature is safe without a regression for it 😈.

I have a regression. But after the new version release I wasn't able to merge my code because so much changed in the subroutine. Although I was level with develop most of the time.

auzbaig:
Config, mesh and inlet file for anyone who wants to run the code:
https://mailuc-my.sharepoint.com/:f:/g/personal/baigan_mail_uc_edu/Ep03wpFrJW1Hvwa73U_Sp6QBaZ_dNyyynRcwWZFCIMF0lQ?e=t5UZb4

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4265,74 +4267,175 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+
+            if(config->GetKindInletInterpolationFunction() == ONED_LINEAR_SPANWISE)
+              Interpolation_Function = "Linear";
+            else if (config->GetKindInletInterpolationFunction() == ONED_AKIMASPLINE_SPANWISE)
+              Interpolation_Function = "Akima";
+              
+              /* --- Creating and initializing a new pointer to object for class Interpolation ---*/
+              CInletInterpolation **s = new CInletInterpolation*[nRows];
+              for (int i=0;i<=nColumns;i++)
+                s[i] = new CInletInterpolation[nRows];
```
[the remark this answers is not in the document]
auzbaig: I need to make the interpolation function for every column of the inlet file. So if there are 6 columns in the file I need 6 separate, say, Akima Interpolation functions. Every "s [ i ]" points to the respective column. That's why I need a 2D matrix.

Otherwise, I would need to create separate objects for every column which I would not be sure of as the number of columns may vary between files.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -55,6 +55,7 @@
 #include "../../../Common/include/graph_coloring_structure.hpp"
 #include "../../../Common/include/toolboxes/MMS/CVerificationSolution.hpp"
 #include "../variables/CVariable.hpp"
+#include "../../../Common/include/toolboxes/CInletInterpolation.hpp"
```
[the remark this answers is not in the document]
auzbaig: I am calling object 's' at line 4401 in CSolver.cpp file. This was initialized in the switch statement, lines 4299 to 4301. For it to be accessible at line 4401 I need it to be globally declared in the .hpp file.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -120,6 +121,7 @@ class CSolver {
   *jPoint_UndLapl;            /*!< \brief Auxiliary variable for the undivided Laplacians. */
   su2double **Smatrix,        /*!< \brief Auxiliary structure for computing gradients by least-squares */
   **Cvector;                  /*!< \brief Auxiliary structure for computing gradients by least-squares */
+  CInletInterpolation **s;          /*!< \brief Pointer to Inlet Interpolation class */
```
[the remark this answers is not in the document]
auzbaig: same reason as for #include "../../../Common/include/toolboxes/CInletInterpolation.hpp".

Common/src/toolboxes/CInletInterpolation.cpp
```diff
@@ -0,0 +1,152 @@
...
+s->b[n-1]=p[n-2]; s->b [n-2]=(p[n-2]+p[n-3])/2;
+
+for (int i =2; i<n-2; i ++){
+su2double w1=fabs(p[i+1]-p[i]) , w2=fabs(p[i-1]-p[i-2]);
+if (w1+w2==0) s->b[i] = (p[i-1]+p[i])/2 ;
+else s->b [i] = (w1*p[i-1]+w2*p[i])/(w1+w2) ;
+}
+for (int i =0; i<n-1; i ++){
+s->c [i]=(3*p [i]-2*s->b [ i ]-s->b[i+1])/h[i] ;
+s->d [i]=(s->b[i+1]+s->b [ i ]-2*p[i])/h[i]/h[i];
```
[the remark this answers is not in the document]
auzbaig: Ok I will make it more readable. But you don't have to worry about the algorithm, its tried and tested by me through several ways including comparison with comparable codes in python. Its works fine for all points including the ones close to boundaries. For the purpose of standardization  I will bring it as close as possible to SU2 standard.

Common/include/toolboxes/CInletInterpolation.hpp
```diff
@@ -0,0 +1,84 @@
...
+                                    unsigned short nDim, 
+                                    su2double *Coord, 
+                                    unsigned short nVar_Turb, 
+                                    string Interpolation_Type);
+
+void PrintInletInterpolatedData(vector<su2double> Inlet_Values, string Marker, unsigned long nVertex, unsigned short nDim);
+
+void free_memory (CInletInterpolation *s){
+   delete[] s; s=NULL; }
+};
```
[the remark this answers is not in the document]
auzbaig: `            if (s!= NULL) {
              for (iVar = 0; iVar < nColumns; iVar++) {
                if (s[iVar] != NULL)
 {
                  delete s[iVar];
                }
              }
              delete [] s;`

I am doing this in the .cpp file as well when the interpolation functions execution is complete (line 4427). Do I need to do that or leave everything on the destructor? (which I have to include yet).

Common/include/option_structure.hpp
```diff
@@ -335,7 +335,37 @@ static const MapType<string, ENUM_RADIALBASIS> RadialBasisFunction_Map = {
...
+  VR_VTHETA = 0,
+  ALPHA_PHI = 1,
+  BETA_PHI = 2,
+};
+
+static const map<string, ENUM_INLET_INTERPOLATIONTYPE> Inlet_SpanwiseInterpolationType_Map{
+MakePair("VR_VTHETA",VR_VTHETA)
+MakePair("ALPHA_PHI",ALPHA_PHI)
+MakePair("BETA_PHI",BETA_PHI)
+};
```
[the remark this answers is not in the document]
auzbaig: Yes. I did this in passing in my recent merge with develop. The options changes a bit and I didn't focus much on formatting.

Common/include/toolboxes/CInletInterpolation.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+class CInletInterpolation:public CAkimaSpline,public CLinearSpline{
+string Interpolation_Type, Interpolation_Function;
+public:
+void Interpolation_Set(CInletInterpolation *s, vector<su2double> &x, vector<su2double> &y, string Interpolation_Function){
+    if(Interpolation_Type == "Akima")
+        return s->akima_spline_set(s,x,y);
+    else if(Interpolation_Type == "Linear")
+        return s->Linear_set(s,x,y);
+}
```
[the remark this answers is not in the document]
auzbaig: If I turn the interpolation inheritance upside down then I would need to make separate objects for the interpolation class I want to do. For Akima I would need to call 'CAkimaSpline' and for Linear 'CLinearSpline'. Right now I call only CInletInterpolation and pass a string (or enum) to let it know what to do. Don't you think the latter way is simpler?

I could include a constructor in CInletInterpolation to let it know which class to initialize from using the enum option passed.

I do need to pass the pointer to the Interpolator to let it know which column to interpolate (which interpolation function to use). I have included the details for this in my other comments.
https://github.com/su2code/SU2/blob/c29a8f042c3af802e8bcb8e3228bb2b900fad46c/SU2_CFD/src/solvers/CSolver.cpp#L4314

I will rename CInletInterpolation to C1DInterpolation.

SU2_CFD/include/solvers/CSolver.hpp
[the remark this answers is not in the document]
auzbaig: https://github.com/su2code/SU2/blob/c29a8f042c3af802e8bcb8e3228bb2b900fad46c/SU2_CFD/src/solvers/CSolver.cpp#L4299

https://github.com/su2code/SU2/blob/c29a8f042c3af802e8bcb8e3228bb2b900fad46c/SU2_CFD/src/solvers/CSolver.cpp#L4401

.gitignore
```diff
@@ -80,4 +82,4 @@ Mercurial
 .hg*
 
 # Ignore build folder
-./build/
+build/
```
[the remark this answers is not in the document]
auzbaig: ok

Common/src/toolboxes/CInletInterpolation.cpp
```diff
@@ -0,0 +1,152 @@
...
+    Inlet_Values[nDim+5] = Inlet_Values[5];
+if (nVar_Turb == 2)
+    Inlet_Values[nDim+6] = Inlet_Values[6];
+
+/*--- Correct for Interpolation Type now ---*/
+if(Interpolation_Type == "VRVTHETA"){
+    unit_r = Inlet_Interpolated[nDim+2];
+    unit_Theta = Inlet_Interpolated[nDim+3];
+}
+else if(Interpolation_Type == "ALPHA_PHI"){
```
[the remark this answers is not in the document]
auzbaig: will work on that.

SU2_CFD/include/solvers/CSolver.hpp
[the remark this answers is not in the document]
pcarruscag: Which are both in function CSolver::LoadInletProfile, so the object only needs to exist in that scope

Common/include/toolboxes/CInletInterpolation.hpp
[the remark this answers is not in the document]
pcarruscag: No it is not simpler, that line you show there `s[iCol]->Interpolation_Set(s[iCol],...` looks like trouble from a mile away, you have a method of an object getting the object itself as an argument. There situations where something like this might be needed, but this is not one of them, if you turn the inheritance the right way around this is what your code should look like.
```c++
C1DInterpolation* interpolator = nullptr; // use descriptive names for your variables
switch(InterpolationType) {
  case LINEAR_INTEPOLATION:
    interpolator = new CLinearInterpolation(InletData,nRows,nCols,...);
    break;
  case AKIMA_INTERPOLATION:
    interpolator = new CAkimaInterpolation(InletData,nRows,nCols,...);
    break;
  default:
    SU2_MPI::Error("...
    break;
}

// Do stuff with the interpolator
interpolator->

// Delete the interpolator
delete interpolator;
```

Common/include/toolboxes/CInletInterpolation.hpp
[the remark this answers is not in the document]
pcarruscag: Destructors are for managing the members of the class, if inside the class you allocate something with operator `new` you need a custom destructor (basically) at the moment your classes do not seem to need that. If some function creates an interpolation object with new it is the responsibility of that function to call delete, i.e. do not define this kind of convenience function.
By the way that function does not do what you want, the `s=NULL` part has no effect on the variable you pass to the function.

SU2_CFD/src/solvers/CSolver.cpp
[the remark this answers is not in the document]
pcarruscag: I just had a look at the profile reader, for each marker it returns a vector of size nRows * nColums right? which means you do not need to interpolate each column independently, before you changed the code you were operating on the entire profile, keep that feature as it is the right way to do it IMO, there is not need to create extra copies of the data to separate things.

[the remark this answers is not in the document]
auzbaig: Previously I was calculating the interpolated value for every vertex at that instant the vertex is called. So I took that point and some points before and after it in the same column to calculate the interpolated value. Now I am using "Set" and "Eval" functions which need to have the different splines "Set" before it starts to loop over the vertexes when it uses "Eval".

An alternative is to pass the Inlet_Data (from profileReader) to C1DInterpolation class and then do the separation there. But I would need interpolation functions equal to the number of columns that I have.

SU2_CFD/include/solvers/CSolver.hpp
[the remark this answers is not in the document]
auzbaig: Yes, I believe I would have to declare "C1DInterpolation **s" before I enter the switch statement. I will do that.

Common/include/toolboxes/CInletInterpolation.hpp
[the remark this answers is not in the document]
auzbaig: This does look very clean. However, I would have to separate the columns in each of the constructors. The "set" and "eval" functions can only handle one set of piece-wise interpolation functions (only one column). In some way I need to make more instances of them to allow for more columns. 
If you don't like this way of calling:
`Inlet_Interpolated[iVar]=s[iVar]->Interpolation_Evaluate(s[iVar],Interp_Radius,Interpolation_Function);`
can I use the same approach but use this-> pointer to access the object data rather than passing it as an argument?

In either case I would need to do something like this (matrix of interpolators):
`             CInletInterpolation **s = new CInletInterpolation*[nRows];
              for (int i=0;i<=nColumns;i++)
                s[i] = new CInletInterpolation[nRows];`

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4265,74 +4268,148 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+          /*--- Define Inlet Values vectors before and after interpolation (if needed) ---*/
+          vector<su2double> Inlet_Values(nCol_InletFile);
+          vector<su2double> Inlet_Interpolated(nColumns);
 
-          /*--- Loop through the nodes on this marker. ---*/
+          unsigned long nRows = profileReader.GetNumberOfRowsInProfile(jMarker);
 
-          for (iVertex = 0; iVertex < geometry[MESH_0]->nVertex[iMarker]; iVertex++) {
+          /*--- Pointer to call Set and Evaluate functions. ---*/
+          vector<C1DInterpolation*> interpolator (nColumns);
```
auzbaig: Is this a better way to initialize multiple pointers from the same class?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4265,74 +4268,148 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
-            /*--- Find the distance to the closest point in our inlet profile data. ---*/
+          switch(config->GetKindInletInterpolationFunction()){
 
-            for (iRow = 0; iRow < profileReader.GetNumberOfRowsInProfile(jMarker); iRow++) {
+            case (NONE):
+            Interpolate = false;
+            break;
 
-              /*--- Get the coords for this data point. ---*/
+            case (ONED_AKIMASPLINE_SPANWISE):
```
auzbaig: I tried to bring it as close as possible to what you suggested. Now my constructor is calling the SetSpline functions.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4265,74 +4268,148 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+              /* --- Calculating the radius and angle of the vertex ---*/
+              Interp_Radius = sqrt(pow(Coord[0],2)+ pow(Coord[1],2));
+              Theta = atan2(Coord[1],Coord[0]);
+
+              /* --- Evaluating and saving the final spline data ---*/
+              for  (unsigned short iVar=0; iVar < nColumns; iVar++){
+              //Evaluate spline will get the respective value of the Data set (column) specified
+              //for that interpolator[iVar], cycling through all columns to get all the 
+              //data for that vertex
+                Inlet_Interpolated[iVar]=interpolator[iVar]->EvaluateSpline(Interp_Radius);
```
auzbaig: I call it like this now without passing the object in the argument but I use `this->` pointer to keep track of which interpolation class instance I am dealing with.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,103 @@
...
+    /*--- Constructor for general class usage ---*/
+    CAkimaInterpolation(vector<su2double> &X, vector<su2double> &Data){
+        SetSpline(X,Data);
+    }
+
+    /*--- Constructor for inlet interpolation with correction ---*/
+    CAkimaInterpolation(vector<su2double>& Inlet_Data, unsigned short nColumns, unsigned long nRows, signed short iCol){
+    
+    SetDataFromInletColumns(Inlet_Data, nColumns, nRows, iCol, this->X, this->Data);
+    }
```
auzbaig: This is where `SetDataFromInletColumns(...)` separates the column for the specific 'iCol' and then calls SetSpline(X, Data).

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    int i =0, j=this->n-1;
+
+    while (j-i>1){ 
+        int m=(i+j) / 2 ; 
+            if (Point_Interp>this->x[m]) {i=m;}
+            else {j=m;} 
+    }
+
+    su2double h=Point_Interp-this->x[i] ;
+    return this->y[i]+h*(this->b[i]+h*(this->c[i]+h*this->d[i])) ;
```
pcarruscag: You don't need the `this` pointer, methods of CAkimaInterpolation have access to member variables.
It creates a bit of confusion when used next to `>` or `-`.

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    this->y.resize(n);
+    this->dydx.resize(n-1);
+
+    for (int i=0; i<n ; i++){
+        this->x[i] = X[i];
+        this->y[i] = Data[i];
+    }
+
+    for (int i=0; i<n-1;i ++){h[i]=X[i+1]-X[i];}
+    for (int i=0; i<n-1;i++) {this->dydx[i]=(Data[i+1]-Data[i]) / h [i];}
```
pcarruscag: `h` does not need to be a vector, you can combine the 2 last loops into just one.

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    this->b[1] =(p[0]+ p[1])/2 ;
+    this->b[n-1]=p[n-2]; 
+    this->b [n-2]=(p[n-2]+p[n-3])/2;
+
+    for (int i =2; i<n-2; i ++){
+        su2double w1=fabs(p[i+1]-p[i]) , w2=fabs(p[i-1]-p[i-2]);
+
+        if (w1+w2==0) {
+            this->b[i] = (p[i-1]+p[i])/2 ;
+        }
```
pcarruscag: it is usually recommended not to compare floating point values with `==`, it is better to do use `<` with a small value

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+
+/*--- Correcting for Interpolation Type ---*/
+vector<su2double> C1DInterpolation::CorrectedInletValues(vector<su2double> &Inlet_Interpolated , 
+                                                        su2double Theta ,
+                                                        unsigned short nDim, 
+                                                        su2double *Coord, 
+                                                        unsigned short nVar_Turb,
+                                                        CConfig *config){
```
pcarruscag: This function does not use any of the class members right?
Move it out of C1DInterpolation, i.e. make it a free function and, I would not pass it the `config`, just the interpolation type.
This way you do not need to instantiate a "corrector" object, you just call the function.

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    
+    unsigned long index;
+    for (unsigned long iRow = 0; iRow < nRows; iRow++){
+    index = iRow*nColumns;
+        X[iRow]=Inlet_Data[index];
+        Data[iRow]=Inlet_Data[index+iCol];
+    }
+
+    SetSpline(X,Data);
+}
```
pcarruscag: Make this a method of the profileReader, for example `profileReader.GetDataForProfile(jMarker, iCol)`, then you can have just one constructor for the interpolation classes right?

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,202 @@
...
+    int n = X.size();
+    vector<su2double> h (n-1);
+    this->x.resize(n);
+    this->y.resize(n);
+    this->dydx.resize(n-1);
+
+    for (int i=0; i<n ; i++){
+        this->x[i] = X[i];
+        this->y[i] = Data[i];
+    }
```
pcarruscag: If you want to copy vectors you can just do `x = X`

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+class C1DInterpolation{
+protected:
+    bool Point_Match = false;
+    vector<su2double> Data;
+    vector<su2double> X;
+public:
+virtual void SetSpline(vector<su2double> &x, vector<su2double> &y){}
+virtual su2double EvaluateSpline(su2double Point_Interp){return 0;}
+virtual bool GetPointMatch(){return Point_Match;}
```
pcarruscag: the GetPointMatch method does not need to be virtual.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /*--- Constructor for inlet interpolation with correction ---*/
+    CAkimaInterpolation(vector<su2double>& Inlet_Data, unsigned short nColumns, unsigned long nRows, signed short iCol){
+    
+    SetDataFromInletColumns(Inlet_Data, nColumns, nRows, iCol, this->X, this->Data);
+    }
+
+    void SetSpline(vector<su2double> &x, vector<su2double> &y);
+    su2double EvaluateSpline(su2double Point_Interp);
```
pcarruscag: Add the keyword `override` to virtual methods of derived classes, e.g.
```
void SetSpline(vector<su2double> &x, vector<su2double> &y) override;
su2double EvaluateSpline(su2double Point_Interp) override;
```
This is to guarantee that the methods really override the base class methods.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,103 @@
...
+using namespace std;
+
+class C1DInterpolation{
+protected:
+    bool Point_Match = false;
+    vector<su2double> Data;
+    vector<su2double> X;
+public:
+virtual void SetSpline(vector<su2double> &x, vector<su2double> &y){}
+virtual su2double EvaluateSpline(su2double Point_Interp){return 0;}
```
pcarruscag: You need to declare the destructor of this class as virtual.
`virtual ~C1DInterpolation() = default;`

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: Yes I did that because you pointed out not to use the 'side effect' of updating the variables like this.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: I agree.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: that makes sense.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: Is it ok to have free functions like that? I have seen in SU2 functions are always encapsulated in some class. Also, if I don't pass the config then I would have to receive an unsigned short for the enum options. How do I neatly use switch statements with it?

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: Yes that's correct. However, I am first thinking of resolving the segfaults before I go on here.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: ok

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: Yes I realize that now.

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: done.

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: Do I need destructors for each class too? I have included this but it doesn't seem to resolve the segfault issue.

auzbaig: @pcarruscag the segfault issue is resolved. I assigned wrong size of vector to one vector.

auzbaig:
@pcarruscag Both Akima and Linear capabilities are working totally as expected. No seg faults or any other kind of errors. I have to carry out some regression tests and post processing now to confirm everything is working fine. I am so greatly thankful for your help.

I have converted this pull request from Draft to Open. Tests are failing with MPI issues, if you can help me with that it will be great. Although I am running on MPI on my CentOS system on 4 cores it works fine.

I would be adding comments with all the methods explaining their purpose plus leveling my branch with most recent develop. Thanks again!

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,179 @@
...
+su2double CLinearInterpolation::EvaluateSpline(su2double Point_Interp){
+    int size = x.size();
+
+    for (int i=0;i<size-1;i++){
+        if(Point_Interp>=x[i] && Point_Interp<=x[i+1]){
+            Point_Match = true;
+            return (Point_Interp-x[i])*dydx[i]+y[i];}
+    }
+    return 0;
+}
```
pcarruscag: You check "PointMatch" after evaluating the spline but I don't see here where that boolean is set to false in case of failure.

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,179 @@
...
+    myfile.precision(16);
+    myfile.open("Interpolated_Data_"+Marker+".dat",ios_base::out);
+
+    if (myfile.is_open()){  
+    for (unsigned long iVertex = 0; iVertex < nVertex; iVertex++) {
+                for  (unsigned short iVar=0; iVar < nColumns; iVar++){
+                    myfile<<Inlet_Data_Interpolated[iVertex*nColumns+iVar]<<"\t";
+                }
+                myfile<<endl;
+            }
```
pcarruscag: reminder to fix the indentation

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
```diff
@@ -155,6 +156,17 @@ class CMarkerProfileReaderFVM {
...
+    /*!
+   * \brief Get the number of columns of data in a profile if an interpolation method was used (multigrid).
+   * \param[in] val_iProfile - current profile index.
+   * \param[in] Interpolate - bool value for Interpolation
+   * \returns Number of columns of data in the (if) interpolated profile.
+   */
+  inline unsigned long GetNumberOfColumnsInProfile(int val_iProfile, bool Interpolate) {
+    if (Interpolate == true) {return numberOfColumnsInProfile[val_iProfile]+dimension;}
+    else {return numberOfColumnsInProfile[val_iProfile];}
+  }
```
pcarruscag: This class should have no need to know about interpolation aspects, we want everything to be as orthogonal as possible, if you need to correct the number of columns do it on the call site.

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
```diff
@@ -164,4 +176,17 @@ class CMarkerProfileReaderFVM {
...
+   * \param[in] iCol - the column whose data is required
+   * \returns the specific column data.
+   */
+  inline vector<su2double> &GetColumnForProfile(int val_iProfile, unsigned short iCol) {
+    ColumnData.resize(numberOfRowsInProfile[val_iProfile]);
+  for (unsigned long iRow = 0; iRow < numberOfRowsInProfile[val_iProfile]; iRow++)
+    ColumnData[iRow]=profileData[val_iProfile][iRow*numberOfColumnsInProfile[val_iProfile]+iCol];
+  return ColumnData;
+  }
+
```
pcarruscag: Here I think this should return by value, like this you allow the outside world to have a non-const reference to the class member data, which is not desired.
```
  inline vector<su2double> GetColumnForProfile(int val_iProfile, unsigned short iCol) const {
    auto nRow = numberOfRowsInProfile[val_iProfile];
    auto nCol = numberOfColumnsInProfile[val_iProfile];
    vector<su2double> ColumnData(nRow);
    for (unsigned long iRow = 0; iRow < nRow; iRow++)
      ColumnData[iRow]=profileData[val_iProfile][iRow*nCol+iCol];
    return ColumnData;
  }
```
(and delete ColumnData from the member variables) (returning by value in C++11 is not bad when the object is local to the function)

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: To initialize a vector of pointers no, not really, to instantiate members of a polymorphic family there is the "Factory" design pattern.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+* \param[in] nVar_Turb - the number of turbulence variables as defined by turbulence model
+* \param[in] *config - Definition of the particular problem.
+* \returns the corrected Inlet Interpolated Data.
+*/
+vector<su2double> CorrectedInletValues(vector<su2double> &Inlet_Interpolated, 
+                                    su2double Theta ,
+                                    unsigned short nDim, 
+                                    su2double *Coord, 
+                                    unsigned short nVar_Turb, 
+                                    CConfig *config);
```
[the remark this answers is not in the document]
pcarruscag: Ah yes thank you, this was in one of my older comments :+1:

Common/include/option_structure.hpp
```diff
@@ -335,7 +335,35 @@ static const MapType<string, ENUM_RADIALBASIS> RadialBasisFunction_Map = {
...
+ */
+enum ENUM_INLET_INTERPOLATIONTYPE {
+  VR_VTHETA = 0,
+  ALPHA_PHI = 1,
+  BETA_PHI = 2,
+};
+static const map<string, ENUM_INLET_INTERPOLATIONTYPE> Inlet_SpanwiseInterpolationType_Map = {
+  MakePair("VR_VTHETA",VR_VTHETA)
+  MakePair("ALPHA_PHI",ALPHA_PHI)
+  MakePair("BETA_PHI",BETA_PHI)
```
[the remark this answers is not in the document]
auzbaig: Yes. Please see the link I have set in the text of the pull request above. I can post it here again. 
[Link to files](https://mailuc-my.sharepoint.com/:f:/g/personal/baigan_mail_uc_edu/Ep03wpFrJW1Hvwa73U_Sp6QBaZ_dNyyynRcwWZFCIMF0lQ?e=daN2r1).
You can see the files for velocity triangles. There it has all the equations and terminologies I that I have used, I have also highlighted them. Including alpha, phi, Vr, VTheta etc.
The reason I use to update everything on my fork is that I am using one machine to edit and another machine to run my code through remote. I don't want regression tests to be kicking off every time I push something here.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: The default set in C1DInterpolation class protected variables: `bool Point_Match = false`

auzbaig: I had set the default value in the initialization of C1DInterpolation class protected variables. But I changed that now to keep it default as false for every call of the evaluation function. I will push the new commit soon.

Common/include/toolboxes/C1DInterpolation.hpp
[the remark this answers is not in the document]
auzbaig: I would want to do that too, but then I would be passing an unsigned short (enum), can I check that neatly with switch without #include<config>?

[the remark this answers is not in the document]
auzbaig: @talbring you need to define the radius at which you want those values to be calculated. For that, it makes sense to have a Vr_Vtheta as default (think like cylindrical coordinates). I can include an option for Vx_Vy for defining at a particular radius if you think that might be useful.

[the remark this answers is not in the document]
pcarruscag: Include option_structure.hpp instead of CConfig.hpp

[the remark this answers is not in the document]
pcarruscag: And do not use unsigned short, use the actual enum type

Common/include/option_structure.hpp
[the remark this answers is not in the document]
pcarruscag: Edit your first post and attach the files directly, give a brief description of the options and reference this pull request in your example testcase, otherwise no one will find the files.

Common/include/toolboxes/C1DInterpolation.hpp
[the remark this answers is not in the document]
auzbaig: @pcarruscag also, I am printing 'Inlet Interpolation being done using AKIMA function and type ALPHA_PHI for inlet' during execution in the main LoadInletProfile function and it seems to appear twice. How can I avoid that? I think it is being called from CDriver twice on lines 1409-1413.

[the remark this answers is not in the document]
pcarruscag: And if you run on 4 cores does it print 4 times? If so you need
```
if (rank == MASTER_NODE) {
  // your print statements
}
```
If it is printed 2 times because it is called for flow and turbulence variables I guess it is fine...

Common/include/option_structure.hpp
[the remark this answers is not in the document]
auzbaig: done

Common/include/toolboxes/C1DInterpolation.hpp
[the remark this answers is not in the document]
auzbaig: @pcarruscag please be a little more specific about how to pass the enum options. I am doing this but it is not working:

`              switch(config->GetKindInletInterpolationType()){
              case(VR_VTHETA):
                Inlet_Values = CorrectedInletValues(Inlet_Interpolated, Theta, nDim, Coord, 
nVar_Turb, VR_VTHETA);
              break;
              case(ALPHA_PHI):
                Inlet_Values = CorrectedInletValues(Inlet_Interpolated, Theta, nDim, Coord, nVar_Turb, ALPHA_PHI);
              break;
              }`

```
vector<su2double> CorrectedInletValues(.....enum ENUM_INLET_SPANWISEINTERPOLATION)
```

I also tried passing `config->GetKindInletInterpolationType()` but it didnt work

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+    * \brief Virtual destructor of the C1DInterpolation class.
+    */
+    virtual ~C1DInterpolation() = default;
+
+    /*!
+    * \brief virtual method for setting the cofficients for the respective spline spline.
+    * \param[in] X - the x values.
+    * \param[in] Data - the f(x) values.
+    */
+    virtual void SetSpline(vector<su2double> &x, vector<su2double> &y){}
```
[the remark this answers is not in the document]
auzbaig: done.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+    /*!
+    * \brief bool variable to make sure all vertex points fell in range of the inlet data.
+    * \returns the bool variable.
+    */
+    bool GetPointMatch(){return Point_Match;}
+};
+
+
+class CAkimaInterpolation: public C1DInterpolation{
```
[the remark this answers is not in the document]
auzbaig: done.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+    * \param[in] Point_Interp - the point where interpolation value is required.
+    * \returns the interpolated value.
+    */
+    virtual su2double EvaluateSpline(su2double Point_Interp){return 0;}
+
+    /*!
+    * \brief bool variable to make sure all vertex points fell in range of the inlet data.
+    * \returns the bool variable.
+    */
+    bool GetPointMatch(){return Point_Match;}
```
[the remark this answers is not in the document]
auzbaig: done.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: done

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
auzbaig: done

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
auzbaig: done

Common/include/toolboxes/C1DInterpolation.hpp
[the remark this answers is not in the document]
auzbaig: Its not dependent on the number of cores. When I run Euler it only prints once.

auzbaig:
@pcarruscag @talbring I have completed all the work I had to do for the interpolation to work. Merging this branch to develop is a go from my side. If there are any other edits let me know. I hope you understand the capability I have included. I have included all possible documentation I could, although I could nicely combine the process in a pdf file for someone's reference. But that can take time. I hope this capability can be useful to you and other people and not just me or my lab mates.

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <cmath>
+#include <vector>
+#include<fstream>
+#include "../datatype_structure.hpp"
+#include "../option_structure.hpp"
+
+using namespace std;
+
+class C1DInterpolation{
+protected:
```
economon: I suggest making the class members private where possible

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+    /*!
+    * \brief bool variable to make sure all vertex points fell in range of the inlet data.
+    * \returns the bool variable.
+    */
+    bool GetPointMatch() const {return Point_Match;}
+};
+
+class CAkimaInterpolation final: public C1DInterpolation{ 
+protected:
```
economon: private here again is possible

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+    /*!
+    * \brief For evaluating the value of Akima Spline.
+    * \param[in] Point_Interp - the point where interpolation value is required.
+    * \returns the interpolated value.
+    */
+    su2double EvaluateSpline(su2double Point_Interp) override;
+};
+
+class CLinearInterpolation final: public C1DInterpolation{
+    protected:
```
economon: and here

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+* \brief to correct for interpolation type.
+* \param[in] Inlet_Interpolated - the interpolated data after spline evaluation.
+* \param[in] Theta - the angle of the vertex (in xy plane).
+* \param[in] nDim - the dimensions of the case.
+* \param[in] Coord - the coordinates of the vertex.
+* \param[in] nVar_Turb - the number of turbulence variables as defined by turbulence model
+* \param[in] ENUM_INLET_INTERPOLATIONTYPE - enum of the interpolation type to be done
+* \returns the corrected Inlet Interpolated Data.
+*/
+vector<su2double> CorrectedInletValues(vector<su2double> &Inlet_Interpolated,
```
economon: vector<su2double> &Inlet_Interpolated  --> const vector<su2double> &Inlet_Interpolated, if this is purely an input

Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+/*!
+* \brief to print the Inlet Interpolated Data
+* \param[in] Inlet_Interpolated_Interpolated - the final vector for the interpolated data
+* \param[in] Marker - name of the inlet marker
+* \param[in] nVertex - total number of vertexes.
+* \param[in] nDim - the dimensions of the problem.
+* \param[in] nColumns - the number of columns in the final interpolated data
+*/                      
+void PrintInletInterpolatedData(vector<su2double>& Inlet_Data_Interpolated, string Marker, unsigned long nVertex, unsigned short nDim, unsigned short nColumns);
```
economon: vector<su2double>& Inlet_Data_Interpolated --> const vector<su2double>& Inlet_Data_Interpolated, if purely an input

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -0,0 +1,190 @@
...
+    for (int i=0;i<n-1;i++)
+        if(Point_Interp>=x[i] && Point_Interp<=x[i+1])
+            Point_Match = true;
+
+    if (Point_Match == false){
+        cout<<"WARNING: Extrapolating data for radius: "<<Point_Interp<<endl;
+        Point_Match = true;
+    }
+
+    int i =0, j=n-1;
```
economon: This is a little picky, but the lack of spacing around operators makes the code a little harder to read.. if you have the ability to auto-format in your IDE, I would recommend adding consistent spaces at least for =, <, etc. (up to you on +, -, etc.)

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4265,74 +4269,187 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+                    cout << endl;
+                    cout << "You can widen the tolerance for point matching by changing the value" << endl;
+                    cout << "of the option INLET_MATCHING_TOLERANCE in your *.cfg file." << endl;
+                    local_failure++;
+                    break;
+                }
+
+            }
+
+            else if(Interpolate == true){
```
economon: This routine is getting a bit long (and I did not help you by writing a fairly long version in the first place).. if you see obvious opportunities, break this out into some smaller helper functions. A good target to shoot for is about 40 lines of code per function (or no more than roughly one screen length, to help people like me who forget what they read once they scroll down 😉!)

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: This class in inherited by CAkimaInterpolation and CLinearInterpolation and these variables are needed there.

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: yes here it is possible

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: yes

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: yes that can be done.

Common/include/toolboxes/C1DInterpolation.hpp
auzbaig: yes will do.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: I would see if I can find an auto-format tool in VS code but I will fix this one up atleast.

SU2_CFD/src/solvers/CSolver.cpp
auzbaig: Are you talking about the whole LoadInletProfile subroutine or just the min_dist<tolerance error output part?

auzbaig: and thanks alot for reviewing my code @economon

Common/src/toolboxes/C1DInterpolation.cpp
economon: You can definitely do this in VS Code if you select a style, such as google, and use the auto-format feature (shift+option+f on Mac)

SU2_CFD/src/solvers/CSolver.cpp
economon: I think the entire thing could be broken down further, if you see some opportunities.

Common/src/toolboxes/C1DInterpolation.cpp
auzbaig: I used the C, C++ style on VS code to auto format (which is only one installed on VS code), and it changed many things drastically including the for(....) {} brackets making the first bracket go to another line. And included a lot of unwanted spacing. So I just fixed this one up manually.

SU2_CFD/src/solvers/CSolver.cpp
auzbaig: It can. But maybe we can leave it for sometime later. I don't want to mix up these two things.

auzbaig:
@pcarruscag Thanks for approving the pull request. Your advice has been immensely important for me to complete this. I hope this is merged timely to avoid any further changes with the develop branch. I will update the config template and add a test case to the repo soon.

PR #818, opened 2019-11-15
stylistic+comment changes

WallyMaier:
Hi all, 

This is a first pass at cleaning up some of the spacing and comments in the headers files of SU2. I also spent some time to make the code more readable, so this is purely a stylistic change.  I'm sure to have missed a bunch.  I know this will create alot of merge conflicts, so i figured I would get your opinions on this.     
My thoughts on this were to get it done now and never have to worry about it again. I only touched the header files as future refactoring will be happening. 

Edit: It seems my attempt to remove tabs has caused some problems.


pcarruscag:
I'm going to piggy bag on your PR and restructure some geometry files, maybe just the small primitives to start.

Make sure when you are doing tab replacements you compile with `-Wmisleading-indentation` I fixed a couple of those, not sure if introduced here or on the other cleaning PR.

pcarruscag:
I moved, cleaned and separated the finite elements (low order ones) into folders `/Common/include(src)/geometry/elements` the idea is to do the same (either in this or other PR) for the primal and dual grid families. Then eventually, geometry_structure.xxx can be separated and moved into `/Common/include(src)/geometry/`

TobiKattmann:
Hi @WallyMaier and @pcarruscag ,

I would like to do a test merge on my branch(es) before this gets merged. I just briefly scrolled through the PR but would like to take a closer look. (As close as it can get for +10.000, -10.000 but a lot of that is copied from A to B as already mentioned)
I will report tomorrow (tuesday) evening.

One thing ahead: I noticed there are quite some files under `Files changed` which have `No changes. ` ...it is also not a mode change (i.e. change in executable permission) ... so what does that mean? At least that is what github shows me in the `Files changed` tab ... in the `jump to file or symbol` list I get some +<some green number>, -<some red number>... so maybe it is just github messing up.

Ok, two things ahead: @pcarruscag I guess I would have preferred to have your contributions split into a separate PR like #821 from the beginning as probably the reviews of each PR would be a bit easier

pcarruscag:
@TobiKattmann you can select the commits for which you want to see changes, if that is not good enough I give you (or anyone else) license to revert my changes here, they'll be carried over to #821 so no problems there :)
I did not merge test your PR by the way, this cleanup you started is needed (and it will get done come hell or high water). "Squash and merge" and you are done.

talbring:
> I wonder why the 'chore' label was removed by @talbring as I would categorize this PR as such?

First of all: the topic should be a little bit more descriptive if it should be listed in the change log. Second: there was already a `no_changelog`. Having both makes no sense and actually I don't know what the release drafter is doing then :D

pcarruscag:
The file diffs are also not correct on other PRs, they do display properly
if you look at particular commits it seems.

PR #824, opened 2019-11-25
WIP Feature Hybrid Parallel and SIMD

pcarruscag:
## Proposed Changes
I will use this draft PR for the work proposed in #789, the idea is to open small (non draft) PR's along the way for major milestones.

## Related Work
#789

## PR Checklist


pcarruscag:
First work item in #789 is:
**CSysMatrix lookups and any required parallel pragmas put in place.**
That may require adding some stuff to CGeometry so I started by restructuring those files (in #825).

pcarruscag:
Next work item: **Hybrid parallel limiter and gradient calculation** This will not involve moving any CSolver files.
I may also do some testing on the element colorings on the structural solver.

pcarruscag:
The work item from #789 being worked on at the moment is **colour-based parallel residual loops (at this point we have hybrid parallel!).**
As was done for the FEA solver one numerics instance per thread is required since to use these classes one needs to write data onto them. Similarly we need thread-private copies of the small structures we then write to the residual and Jacobian, these are currently allocated by the solver.
Possible solutions:
- Allocate local temporaries in each edge loop;
- The solver allocates an array of these instead of just one for each thread to use;
- The ownership is transferred into the numerics (as currently done by the FEA classes).

I am going with the third option, the solver will call `ComputeResidual` as usual but instead of passing in one array and two small matrices, it will get references to the internally (to CNumerics) allocated ones.
**As part of this I will restructure the CNumerics family**, applying the same cleanups that we have been doing for other classes. No significant code changes are planned.

Common/include/option_structure.hpp
```diff
@@ -174,9 +174,11 @@ enum AVERAGE_TYPE {
   AVERAGE_AREA = 1,     /*!< \brief Area-weighted average. */
   AVERAGE_MASSFLUX = 2  /*!< \brief Mass-flux weighted average. */
 };
+#ifdef ENABLE_MAPS
 static const map<string, AVERAGE_TYPE> Average_Map = CCreateMap<string, AVERAGE_TYPE>
 ("AREA", AVERAGE_AREA)
 ("MASSFLUX", AVERAGE_MASSFLUX);
+#endif
```
pcarruscag: The build times were increasing significantly as we divide the files, which does not make sense since the files are smaller. After some digging... the problem is the static maps declared in option_structure.hpp it seems that object code is generated for them in every file that includes option_structure.hpp (i.e. nearly every file in the project).
This adds a constant cost to any file (even if empty), as apparently the trickery that allows this convenient way of declaring the map is not cheap... also we have 129 maps...
We only need the declaration of the maps in config_structure.cpp and SU2_DOT.cpp, so my proposed solution is to guard the definition of each map, if you need them in a particular cpp define ENABLE_MAPS before including config_structure.hpp, if for some reason they are needed in an hpp ENABLE_MAPS should be undefined at the end to keep it from spreading.
One could of course move the maps to a separate file, but having them close to the corresponding enum is very convenient.
I will try to find a cleaner or at least less intrusive way of doing this.

Common/src/toolboxes/MMS/CMMSNSTwoHalfSpheresSolution.cpp
```diff
@@ -391,6 +391,6 @@ void CMMSNSTwoHalfSpheresSolution::GetMMSSourceTerm(const su2double *val_coords,
   val_source[4] /= Velocity_Ref*Pressure_Ref;
 }
 
-bool CMMSNSTwoHalfSpheresSolution::IsManufacturedSolution(void) {
+bool CMMSNSTwoHalfSpheresSolution::IsManufacturedSolution(void) const {
```
pcarruscag: @economon, @vdweide, I put the const and virtual specifiers on the MMS classes and striped the trailing spaces

Common/include/option_structure.hpp
pcarruscag: Update, the #ifdefs in every map can be avoided by making CCreateMap an empty class instead.
However that is not as good for compilation time as not doing any compilation...

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -32,12 +32,58 @@
...
+#include "../../include/numerics/turbulent/turb_sources.hpp"
+#include "../../include/numerics/elasticity/CFEAElasticity.hpp"
 #include "../../include/numerics/elasticity/CFEALinearElasticity.hpp"
-#include "../../include/numerics/elasticity/CFEAMeshElasticity.hpp"
-#include "../../include/numerics/elasticity/CFEM_NeoHookean_Comp.hpp"
-#include "../../include/numerics/elasticity/CFEM_Knowles_NearInc.hpp"
-#include "../../include/numerics/elasticity/CFEM_IdealDE.hpp"
-#include "../../include/numerics/elasticity/CFEM_DielectricElastomer.hpp"
+#include "../../include/numerics/elasticity/CFEANonlinearElasticity.hpp"
+#include "../../include/numerics/elasticity/nonlinear_models.hpp"
```
pcarruscag: On the subject of restructuring the CNumerics files, I initially did it one-class-per-file with some sub folders, the problem with that is that we have some 85 of those classes... so 170 files which I think is maybe too many, more important is that some of those classes form a smaller hierarchies which are not clearly identifiable with so many files.
So I did it by "themes" instead, I think this division should inspire some re-usage when new classes are added, the average cpp size is under 1k lines, which I think is quite manageable.

Common/include/option_structure.hpp
talbring: That seems indeed like a "problem" we should address. Thanks for pointing that out. However, I am not a big fan of the #ifdef solution.

pcarruscag: It is not a "problem" it is a **PROBLEM**, I know you guys over there have smart compilation tools that distribute compilation over the network, but for users with a normal computer or students with outdated laptops...

talbring: The quotes were misleading. Don't get me wrong. I really think that this is a problem and we should do something about it.

pcarruscag: :+1: I do not like #ifdefs sprinkled all over the code either...
@talbring what about this solution:
```C++
#include <string>
#include <map>

//#define ENABLE_MAPS

using namespace std;

template<class T, class U>
struct EmptyMap {
    EmptyMap(initializer_list<pair<T,U> >) {}
};

#ifdef ENABLE_MAPS
    template<class T, class U>
    using MapType = map<T,U>;
    #define MakePair(a,b) {a,b},
#else
    template<class T, class U>
    using MapType = EmptyMap<T,U>;
    #define MakePair(a,b)
#endif


enum Options {
    OPTION_1,
    OPTION_2
};

static const MapType<string,int> myMap = {
    MakePair("Option_1", OPTION_1)
    MakePair("Option_2", OPTION_2)
};
```
It still involves the preprocessor and would require a small change of semantics, but I don't think there is a 0 cost solution with metaprogramming alone, the compiler still generates code for the initializer lists even is they are sunk into an empty map...

talbring: I just tested in on a small file. It really makes a big difference whether you include option_structure.hpp or not .. How big is the overhead of your proposed solution? If it really helps, we can go for that one.

talbring: Also, can you open a separate branch for this ?

pcarruscag: Compilation-wise it should be 0 as no code is generated, we'll know by testing.

> Also, can you open a separate branch for this ?

I was afraid you were going to ask xD. I can, it is the right way of doing it.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3325,99 +3300,149 @@ void CEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_conta
...
+    }
+    else {
+      /*--- Reconstruction ---*/
+
+      /*--- Static arrays of primitives and secondaries (thread safety). ---*/
 
-      Gradient_i = nodes->GetGradient_Reconstruction(iPoint);
-      Gradient_j = nodes->GetGradient_Reconstruction(jPoint);
+      su2double Primitive_i[MAXNVAR] = {0.0}, Primitive_j[MAXNVAR] = {0.0};
+      su2double Secondary_i[MAXNVAR] = {0.0}, Secondary_j[MAXNVAR] = {0.0};
```
pcarruscag: For efficiency we need to rely on static arrays in some areas, in debug builds there is an assert to guarantee we are not going to run out of space and MAXVAR is easy to change.

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -0,0 +1,759 @@
...
+
+  /*--- Allocate required structures ---*/
+  Diff_U = new su2double [nVar];
+  Diff_Lapl = new su2double [nVar];
+  ProjFlux = new su2double [nVar];
+  Jacobian_i = new su2double* [nVar];
+  Jacobian_j = new su2double* [nVar];
+  for (iVar = 0; iVar < nVar; iVar++) {
+    Jacobian_i[iVar] = new su2double [nVar];
+    Jacobian_j[iVar] = new su2double [nVar];
```
pcarruscag: Jacobians allocated internally...

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -0,0 +1,759 @@
...
+  DissipationTerm(ProjFlux, Jacobian_i, Jacobian_j);
+
+  if (preacc) {
+    AD::SetPreaccOut(ProjFlux, nVar);
+    AD::EndPreacc();
+  }
+
+  residual = ProjFlux;
+  jacobian_i = Jacobian_i;
+  jacobian_j = Jacobian_j;
```
pcarruscag: and returned as const pointers

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -0,0 +1,1284 @@
...
+   * \overload For numerics classes that store the residual/flux and Jacobians internally.
+   * \param[out] residual - Pointer to the total residual.
+   * \param[out] jacobian_i - Jacobian of the numerical method at node i (implicit computation).
+   * \param[out] jacobian_j - Jacobian of the numerical method at node j (implicit computation).
+   * \param[in] config - Definition of the particular problem.
+   */
+  inline virtual void ComputeResidual(const su2double*  &residual,
+                                      const su2double* const* &jacobian_i,
+                                      const su2double* const* &jacobian_j,
+                                      CConfig *config) { }
```
pcarruscag: There is a separate overload for numerics that "return" the values like that.

pcarruscag:
In the process of making the MG routines OpenMP-friendly, I'll also restructure and clean those files (CIntegration & co).

PR #830, opened 2019-12-02
Hybrid parallel (OpenMP) implementation for Linear Algebra classes

pcarruscag:
## Proposed Changes
The first work item from #824 is kind of in place code-wise (lots of testing required).
I will use this PR to document the implementation, while continuing the work on the main PR, and to keep the discussion "encapsulated".

## Related Work
#789, #824

## PR Checklist


Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -0,0 +1,489 @@
...
+ * and the pattern should be used as (row,icol), otherwise as (col,irow).
+ */
+template<typename Index_t>
+class CCompressedSparsePattern {
+  static_assert(std::is_integral<Index_t>::value,"");
+
+private:
+  su2vector<Index_t> m_outerPtr; /*!< \brief Start positions of the inner indices for each outer index. */
+  su2vector<Index_t> m_innerIdx; /*!< \brief Inner indices of the non zero entries. */
+  su2vector<Index_t> m_diagPtr;  /*!< \brief Position of the diagonal entry. */
```
pcarruscag: I created this light class to old generic sparse patterns, it is mostly an aggregate of containers.

Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -0,0 +1,489 @@
...
+ * \param[in] geometry - Definition of the grid.
+ * \param[in] type - Of connectivity.
+ * \param[in] fillLvl - Target degree of neighborhood (immediate neighbors always added).
+ * \return Compressed-Storage-Row sparse pattern.
+ */
+template<class Geometry_t, typename Index_t>
+CCompressedSparsePattern<Index_t> buildCSRPattern(Geometry_t& geometry,
+                                                  ConnectivityType type,
+                                                  Index_t fillLvl)
+{
```
pcarruscag: For kind of a change of pace I am building the sparse patterns in a stand-alone function instead of adding yet another method to CGeometry (the return by value is not an issue for that class in c++ 11).

Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -0,0 +1,489 @@
...
+ * \return Coloring in the same type of the input pattern.
+ */
+template<class T, typename Color_t = char, size_t MaxColors = 64, size_t MaxMB = 128>
+T colorSparsePattern(const T& pattern, size_t groupSize = 1,
+                     std::vector<Color_t>* indexColor = nullptr)
+{
+  static_assert(std::is_integral<Color_t>::value,"");
+  static_assert(std::numeric_limits<Color_t>::max() >= MaxColors,"");
+
+  using Index_t = typename T::IndexType;
```
pcarruscag: Greedy coloring of one of those sparse patterns, I did it like this to avoid having to implement one version for edges, one for elements, etc. I know we had some coloring stuff laying around but this only needs to operate within the rank, and it supports coloring groups instead of single entities of the sparse pattern (as shown in #789 that will be quite important for performance).

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -27,162 +27,160 @@
...
-      vneighs.resize(it - vneighs.begin());
-      nNeigh[iPoint] = vneighs.size();
-    }
+  const auto& csr = geometry->GetSparsePattern(type,0);
 
-  }
+  row_ptr = csr.outerPtr();
+  col_ind = csr.innerIdx();
+  dia_ptr = csr.diagPtr();
+  nnz = csr.getNumNonZeros();
```
pcarruscag: This is the pattern coming from the CGeometry associated with the matrix, CGeometry does lazy construction of those patterns so there is no extra work or complicate logic on the config settings to determine types of solver etc.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -650,129 +481,188 @@ void CSysMatrix<ScalarType>::CompleteComms(CSysVector<OtherType> & x,
...
-    for (iVar = 0; iVar < nVar; iVar++)
-      matrix[index*nVar*nVar+row*nVar+iVar] = 0.0; // Delete row values in the block
-    if (col_ind[index] == block_i)
-      matrix[index*nVar*nVar+row*nVar+row] = 1.0; // Set 1 to the diagonal element
-  }
-
+void CSysMatrix<ScalarType>::SetValZero() {
+  SU2_OMP_PAR_FOR_STAT(OMP_STAT_SIZE)
+  for (unsigned long index = 0; index < nnz*nVar*nEqn; index++)
+    matrix[index] = 0.0;
```
pcarruscag: Simple example of the use of one of those "macro encapsulated" pragmas.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -650,129 +481,188 @@ void CSysMatrix<ScalarType>::CompleteComms(CSysVector<OtherType> & x,
...
-      mat_begin = (index*nVar*nVar); // offset to beginning of matrix block[row_i][col_ind[indx]]
+  /*--- OpenMP parallelization ---*/
+  SU2_OMP_PAR_FOR_DYN(omp_chunk_size)
+  for (auto row_i = 0ul; row_i < nPointDomain; row_i++) {
+    auto prod_begin = row_i*nVar; // offset to beginning of block row_i
+    for(auto iVar = 0ul; iVar < nVar; iVar++)
+      prod[prod_begin+iVar] = 0.0;
+    for (auto index = row_ptr[row_i]; index < row_ptr[row_i+1]; index++) {
+      auto vec_begin = col_ind[index]*nVar; // offset to beginning of block col_ind[index]
+      auto mat_begin = index*nVar*nVar; // offset to beginning of matrix block[row_i][col_ind[indx]]
```
pcarruscag: For non-transposed matrix multiplication this is all we need to make it parallel, it is however not ideal as I will explain in a bit. Note all local variables, if those indexes were declared outside the loop we could have problems.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -815,136 +708,152 @@ void CSysMatrix<ScalarType>::ComputeJacobiPreconditioner(const CSysVector<Scalar
...
-  for (iPoint = 1; iPoint < (long)nPointDomain; iPoint++) {
+  SU2_OMP_PARALLEL_ON(omp_num_parts)
+  {
+  int thread = omp_get_thread_num();
+  const auto begin = omp_partitions[thread];
+  const auto end = omp_partitions[thread+1];
+
+  ScalarType weight[MAXNVAR*MAXNVAR], aux_block[MAXNVAR*MAXNVAR];
+
+  for (auto iPoint = begin+1; iPoint < end; iPoint++) {
```
pcarruscag: For preconditioners that are serial by nature we have a parallel section where each thread works on its own large chunk of the matrix.

SU2_CFD/src/solver_direct_heat.cpp
```diff
@@ -520,38 +520,35 @@ void CHeatSolverFVM::Centered_Residual(CGeometry *geometry, CSolver **solver_con
...
+    for (iEdge = 0; iEdge < geometry->GetnEdge(); iEdge++) {
 
-        /*--- Points in edge ---*/
-        iPoint = geometry->edge[iEdge]->GetNode(0);
-        jPoint = geometry->edge[iEdge]->GetNode(1);
-        numerics->SetNormal(geometry->edge[iEdge]->GetNormal());
+      /*--- Points in edge ---*/
+      iPoint = geometry->edge[iEdge]->GetNode(0);
+      jPoint = geometry->edge[iEdge]->GetNode(1);
+      numerics->SetNormal(geometry->edge[iEdge]->GetNormal());
```
pcarruscag: Few indentation issues fixed here and there too.

pcarruscag:
I have covered all operations used in non adjoint use, the non ideal part of the implementation I mentioned above is that the parallelization is "local", i.e. we get to the operation we want to make parallel and launch the threads there, for simple vector-vector operations the overhead may be significant.
Ideally we would have a parallel construct at a higher level, say CSysSolve::Solve, so that the threads are already in flight when we get to those small operations.
In principle it is not too hard to do that, but it needs to be done carefully especially when the execution gets to an MPI part of the code (which thread(s) communicate, etc.).
I will try to benchmark this to put numbers on the performance / simplicity trade-off.

pcarruscag:
Ok the "simple" version of "going parallel" whenever we get to a linear algebra operation did not make the cut.
On an older architecture there was a 10% slowdown of the linear solvers at ~10k nodes per core and about the same on a newer architecture but only at ~1k node per core.
Since hybrid parallel is supposed to be good for strong scaling, this was not good enough... With the new strategy it is ok (see "performance" below), hence this is ready for review.

### Overall Strategy
The strategy now is to start a parallel section in CSysSolve::Solve that covers building the preconditioner and solving the linear system.
Linear algebra routines called within this section have _worksharing_ constructs instead of _parallel_ ones, i.e. the work is distributed by however many threads arrive to that routine. This also makes the routines safe to call in serial.
The only "dangerous" things to do in parallel are to: manage memory for a shared object (multiple threads call `new` but there is only one shared pointer on which to call `delete`); writing to the same memory locations concurrently.
I tried to make the first issue debugable by asserting that the initialization routines of CSysMatrix and CSysVector are only called by the master thread.
For the second issue I made the associated classes as const-correct as possible, that should at least make someone think twice before changing a member variable of those classes. The risk is still there for input variables as an algorithm development aspect... For example `MatrixVectorProductTransposed` cannot be made thread-parallel as simply/naively as its normal counterpart.

### Communication Model
The MPI + Threads communication model is very simple, currently only the master thread calls MPI routines (including `Error`), this requires thread barriers before and after the communication to make sure the correct values are passed and _seen_ by all threads.
We can test other alternatives in the future but at the moment this does not seem to be a significant bottleneck.
_Worksharing_ constructs have implicit barriers at completion, for CSysVector routines I used `nowait` modifiers, it is safe to call those routines in sequence since the loop sizes, and static work scheduling specifications are identical.
However, routines that access a CSysVector in a different way, should have an explicit barrier before using the vector (or risk having undefined behaviour). You will see these barriers on entry to matrix-vector product, and every `ComputeXXXPreconditioner` (if you don't, let me know xD). I think those routines are large enough to amortise the cost of this.

### Performance
Disclaimer:
- We are talking about linear solvers only, **you will not see a global improvement yet**.
- The large global improvements from "hybridization" will come from the multigrid behaving better on less decomposed domains, and from the ability to independently tune the number of cores used in the linear preconditioners. For now the objective is "just" not to loose performance while gaining flexibility.
- The performance of MPI+threads with 1 thread per rank will be worse than just MPI (no free lunches).

With this [small case](https://github.com/su2code/SU2/files/3933059/small_case.zip) using 8 cores off a machine with two 2650v4 CPU, Intel MPI 2018 + GCC 8.2, the hybrid (2 ranks of 4 threads) approach is about 5% faster thank the MPI-only (8 ranks), I expect larger cases to have identical performance.

### How To
- Compile: Add -fopenmp to the compiler and linker arguments.
- Run: Set number of threads with env variable `OMP_NUM_THREADS` (eventually I will make that a command line parameter), for best performance set `OMP_WAIT_POLICY=ACTIVE` and beware of thread binding settings, use `mpirun --bind-to socket` or `mpirun --bind-to numa` never `core`.

Common/src/linear_algebra/CSysVector.cpp
```diff
@@ -26,472 +26,257 @@
...
-  return &vec_val[val_ipoint*nVar];
+  PARALLEL_FOR
+  for(auto i=0ul; i<nElm; i++) u_array[i] = vec_val[i];
 }
 
 template<class ScalarType>
-template<class T>
-void CSysVector<ScalarType>::PassiveCopy(const CSysVector<T>& other) {
+ScalarType CSysVector<ScalarType>::dot(const CSysVector<ScalarType> & u) const {
+#if !defined(CODI_FORWARD_TYPE) && !defined(CODI_REVERSE_TYPE)
```
pcarruscag: Readability wise, the dot product operation (now a member of CSysVector) is as bad as it can get as we need to perform a reduction over threads and ranks.

economon:
Following along here.. will you eventually move the parallel section in CSysSolve up to the full application level (in the main() maybe) when you move to the next steps? In the earlier work, we found that, as you have seen in the linear solver routines, spawning parallel sections kernel-wise carried a large overhead. We found that the best performance was given by spawning right at the start and carry the threads through the entire program, just like the MPI ranks. 

My only other comments, which it sounds like you are addressing, are to make the threads as transparent as possible to developers (shouldn't need to touch them unless they want to, like the MPI), and to make the compilation painless (disable/enable). Have you connected it to meson somehow yet?

economon:
I should also mention though that moving the threading to a single high-level parallel section is also very problematic for readability/development. Folks will have to be aware that the threads are active, and it can be very error-prone. This was one of the major detractors of implementing the OpenMP framework as we had it in the C&F paper into the develop branch, even though the performance was quite good (and also the interoperability of threading and AD at the time). Any clever suggestions/techniques for hiding the threading as much as possible are most welcome.

pcarruscag:
No meson option yet, it is a very small one though, the whole system works by detecting `-fopenmp`.
Your second comment is the main argument against moving the parallel section further up.
Allocation routines have the highest risk of making a mess, but even seemingly innocuous things like the small auxiliary arrays we allocate e.g. in CSolver and then use in derived classes are a problem.
I am almost done making the FEA solver completely hybrid parallel and I had to refactor most uses of those arrays. This is also why I took a more functional approach to the new limiter and gradient routines. The way we use CConfig is also not thread safe, we would need to make all "SetSomethings" atomic, which would be monumental.
Initially I would have a few parallel sections (it is not too difficult to then move them up if we think that it is the way to go) I want to use the FEA solver to get an idea for the relative performance, after seeing the effect of `OMP_WAIT_POLICY` I am optimistic.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -27,162 +27,167 @@
...
-      sort(vneighs.begin(), vneighs.end());
-      it = unique(vneighs.begin(), vneighs.end());
-      vneighs.resize(it - vneighs.begin());
-      nNeigh[iPoint] = vneighs.size();
-    }
+  /*--- Get sparse structure pointers from geometry,
+   *    the data is managed by CGeometry to allow re-use. ---*/
 
-  }
+  const auto& csr = geometry->GetSparsePattern(type,0);
```
talbring: I don't know what your plans are, but in the future we should then try to pass the parsity pattern to the constructor in order to remove the dependencies of the matrix class on the geometry class

pcarruscag: That would be nice, but we still need the geometry for MPI.

talbring: Actually I think we can move the point-to-point communication to its own structure. What do you think @economon ?

pcarruscag:
@talbring I think GitHub does not recognize the "approved" label as an approval (the bot put it back to un-reviewed).

pcarruscag:
I think this is stable enough to be merged, while testing #834 and #843 I think I went through most combinations of solvers/preconditioners and things seem to work fine, the only source of problems was the dot product (it was locking in some conditions) but all seems fine now.

talbring:
> @talbring I think GitHub does not recognize the "approved" label as an approval (the bot put it back to un-reviewed).

Hm yeah, the label will be removed if there has been a new commit ... unfortunately there is no way to change that ...

PR #834, opened 2019-12-10
Hybrid parallel Gradients and Limiters

pcarruscag:
## Proposed Changes
Compute Green-Gauss gradients and limiters via a point-loop (like we do for Least Squares), in the limit this allows one thread per point of parallelization. Parallel regions are started inside the functions and again only the master thread calls MPI routines.

Move these routines out of CSolver and make them general to reduce code duplication (compressible/incompressible, solution/primitives), and to improve test-ability (if key areas of the code depend less on large objects we can more easily treat them as units).

Introduce a way to create different limiters by defining small details like the limiter function (and its associated constants), or any geometry-based modification.

I don't seem to have "hard" failures in the regressions but I need to run tests, especially for periodic cases as I had to change the "communication model" for the min / max computed for limiters.
I will start documenting anyway.

## Related Work
#789 #824 
Depends on #830

## PR Checklist


SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -0,0 +1,189 @@
...
+template<class FieldType, class GradientType>
+void computeGradientsGreenGauss(CSolver* solver,
+                                MPI_QUANTITIES kindMpiComm,
+                                PERIODIC_QUANTITIES kindPeriodicComm,
+                                CGeometry& geometry,
+                                CConfig& config,
+                                const FieldType& field,
+                                size_t varBegin,
+                                size_t varEnd,
+                                GradientType& gradient)
```
pcarruscag: The signature for these functions is a bit long... But by making the solver+mpi optional we can more easily devise unit tests or use the function for some other purpose.
I have also made these function able to compute gradients for only part of the variables, the idea is to use that to recompute only the velocity gradients before computing the turbulence source terms.

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -0,0 +1,335 @@
...
+ */
+
+
+/*!
+ * \brief A traits class for limiters, see notes for "computeLimiters_impl()".
+ * \note There is no default implementation (the code will compile but not
+ *       link) specialization is mandatory.
+ */
+template<ENUM_LIMITER LimiterKind>
+struct CLimiterDetails
```
pcarruscag: The way I've come up with to create new limiters without copy pasting a similar implementation is to specialise this class template. I did not want to make this polymorphic.

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -0,0 +1,335 @@
...
+  {
+    su2double y = delta*(delta+proj) + eps2;
+    return (y + delta*proj) / (y + 2*proj*proj);
+  }
+
+  inline su2double raisedSine(su2double dist)
+  {
+    su2double factor = 0.5*(1.0+dist+sin(PI_NUMBER*dist)/PI_NUMBER);
+    return max(0.0, min(factor, 1.0));
+  }
```
pcarruscag: Initially I took this function from the sharp-edge limiter but I think it was incorrect (the way the distance was normalized and the bounds checking (replaced by min/max)  did not seem to be continuous).
Now for distances less than the reference the "geometric limiter" is 0, at 3 times the distance the value is 1.

SU2_CFD/include/limiters/computeLimiters.hpp
```diff
@@ -0,0 +1,102 @@
...
+    case BARTH_JESPERSEN:
+    {
+      INSTANTIATE(BARTH_JESPERSEN);
+      break;
+    }
+    case VENKATAKRISHNAN:
+    {
+      INSTANTIATE(VENKATAKRISHNAN);
+      break;
+    }
```
pcarruscag: After specialising that class it is just a matter of adding the new `case` to this `switch`.
I tried to have this in a .cpp with explicit instantiation to avoid compiling the code for CSolver, CEulerSolver and CIncEulerSolver, but I was getting undefined references for AD builds.

SU2_CFD/include/limiters/computeLimiters_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+ * \param[out] fieldMax - As above but maximum values.
+ * \param[out] limiter - Reconstruction limiter for the field.
+ *
+ * Template parameters:
+ * \param FieldType - Generic object with operator (iPoint,iVar)
+ * \param GradientType - Generic object with operator (iPoint,iVar,iDim)
+ * \param LimiterKind - Used to instantiate the right details class.
+ */
+template<class FieldType, class GradientType, ENUM_LIMITER LimiterKind>
+void computeLimiters_impl(CSolver* solver,
```
pcarruscag: And this is the general limiter implementation, the idea is that this can stay independent of the particular details of each method.
For each point we loop over its direct neighbours, as explained in #789 this is faster than looping by edges as the computation is more local (more flops, less bytes), and it is more parallelizable.

SU2_CFD/include/variables/CAdjEulerVariable.hpp
```diff
@@ -167,4 +167,10 @@ class CAdjEulerVariable : public CVariable {
    */
   inline su2double **GetGradient_Reconstruction(unsigned long iPoint) final { return Gradient_Reconstruction[iPoint]; }
 
+  /*!
+   * \brief Get the reconstruction gradient for variables at all points.
+   * \return Reference to reconstruction gradient.
+   */
+  inline VectorOfMatrix& GetGradient_Reconstruction(void) final { return Gradient_Reconstruction; }
+
```
pcarruscag: To allow the new routines to work on primitives / solution CVariable needs to give access to the entire container... Not because the routine depends on a specific type of container, but to bypass the names of the AddGradient_XX SubtractGradient_XX methods.
To eventually vectorize the gradient and limiter computation only the container will need to implement "vector-accessors".

Common/include/option_structure.hpp
```diff
@@ -1916,6 +1916,7 @@ static const map<string, ENUM_INPUT_REF> Input_Ref_Map = CCreateMap<string, ENUM
  * \brief Vertex-based quantities exchanged during periodic marker communications.
  */
 enum PERIODIC_QUANTITIES {
+  PERIODIC_NONE       = 99,  /*!< \brief No periodic communication required. */
   PERIODIC_VOLUME     =  1,  /*!< \brief Volume communication for summing total CV (periodic only). */
```
pcarruscag: @economon  the AuxVar did not have periodic gradient communications, so I had to add a do-nothing mode.

pcarruscag:
This just turned a month old :) If I could have some reviews please :)
I'll post test results by the end of the week for the failing regressions.

pcarruscag:
The residuals (and indeed the converged results) change for some cases due to these changes (by very small amounts). Looking at limiter fields there are some differences that justify this. Those differences are present mostly in smooth flow regions, why in this regions?
Because this is where the old edge loops had a lot of logic to avoid divisions by zero, that is not required for the point loop implementation. There was also a change to the epsilon that initializes the min/max calculation (it now comes from the traits of the datatype).
I used the transonic_stator case to check that:
- The computed limiters are not too different, especially near periodic boundaries (for which a change in how min/max neighbor calculation was required).
- The results do not depend on number of ranks / threads.

This is the flow field (for reference):
image

This shows the x-mom limiters near the trailing edge for the develop version:
image

And this for this branch:
image

You can tell a slight "darkening" close to the blue spots, and also before the small spot close to the wall, before the shock. Near the periodic boundaries there are no visible changes.
I ran the case with 1 rank 1 thread (1/1), 1/4 and 4/8 (32 total), the results are the same (the convergence rate is different due to the behavior of the preconditioner).
[files.zip](https://github.com/su2code/SU2/files/4056576/files.zip)

If you would like a specific test, or see some weird behavior in one of your cases let me know, but since only a few regressions changed I will only post comparisons of cases with substantial changes.

pcarruscag:
Of the two cases with larger residual changes:
 - contadj_euler_naca0012 - No idea why they changed, neither primal nor adjoint compute limiters... the primal residuals are unchanged, and the case converges to the same values (residuals and solution) so I simply updated the residuals
- transonic_stator_restart - As shown above the case is fine, so I updated the restart file, however I do not know how to change the testcases branch anymore :) but I guess once the corresponding PR is merged this will start passing.

I ran some other tests with the Venkatakrishnan-Wang limiter (which requires a global min/max) and is does not seem to be covered by the tests ATM (maybe I'll use that restart case to fix that), everything looks perfect, same results with different ranks/threads and so on, the results are tens of MB so I won't upload unless someone wants to double check.

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -0,0 +1,335 @@
...
+template<>
+struct CLimiterDetails<BARTH_JESPERSEN>
+{
+  su2double eps2;
+
+  /*!
+   * \brief Set a small epsilon to avoid divisions by 0.
+   */
+  template<class... Ts>
+  inline void preprocess(Ts&...) {eps2 = LimiterHelpers::epsilon();}
```
jayantmukho: Had to look up what the `...` was for. Didn't know about parameter packs, very cool. 
Also super clean implementation of limiters.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -4178,6 +4178,19 @@ bool CFluidDriver::Monitor(unsigned long ExtIter) {
 
 void CFluidDriver::Output(unsigned long InnerIter) {
 
+  for (iZone = 0; iZone < nZone; iZone++) {
+    const auto inst = config_container[iZone]->GetiInst();
```
jayantmukho: Is there a reason to use `auto` as opposed to a specific type?

Also were different instantiations just not being output before you added these lines?

SU2_CFD/src/solver_structure.cpp
```diff
@@ -3483,447 +3109,16 @@ void CSolver::SetAuxVar_Surface_Gradient(CGeometry *geometry, CConfig *config) {
...
-  /*--- Limiter MPI ---*/
-  
-  InitiateComms(geometry, config, SOLUTION_LIMITER);
-  CompleteComms(geometry, config, SOLUTION_LIMITER);
-
-#ifdef CODI_REVERSE_TYPE
-  if (TapeActive) AD::StartRecording();
-#endif
+  computeLimiters(kindLimiter, this, SOLUTION_LIMITER, PERIODIC_LIM_SOL_1, PERIODIC_LIM_SOL_2,
+                  *geometry, *config, 0, nVar, solution, gradient, solMin, solMax, limiter);
```
jayantmukho: very satisfying deletion of duplicated code over the solver files 😃

SU2_CFD/include/limiters/computeLimiters_impl.hpp
```diff
@@ -0,0 +1,260 @@
...
+                          FieldType& fieldMax,
+                          FieldType& limiter)
+{
+  constexpr size_t MAXNDIM = 3;
+  constexpr size_t MAXNVAR = 8;
+
+  if (varEnd > MAXNVAR)
+    SU2_MPI::Error("Number of variables is too large, increase MAXNVAR.", CURRENT_FUNCTION);
+
+  CLimiterDetails<LimiterKind> details;
```
jayantmukho: my only gripe is the vague name `details`

maybe `limiterDetails`

I'm really nitpicking now

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -486,6 +488,70 @@ class C2DContainer :
...
+/*!
+ * \class C2DDummyLastView
+ * \brief Helper class, adds dummy trailing dimension to a reference of a
+ *        vector object making it a dummy matrix.
+ * \note The constness of the object is derived from the template type, but
+ *       we allways keep a reference, never a copy of the associated vector.
+ */
+template<class T>
+struct C2DDummyLastView
+{
```
jayantmukho: I don't really follow what these two structs `C2DDummyLastView` and `C3DDummyMiddleView` are doing, could you elaborate on what they are used for?

SU2_CFD/include/limiters/CLimiterDetails.hpp
pcarruscag: Thank you for the review @jayantmukho, to be honest I'm no specialist either, I am only using them to make those methods forward compatible in case some of the specializations need to be passed extra parameters in the future.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Mostly because I find boring to write `unsigned blabla` all the time. It is a recommend practice in the book Effective Modern C++ (but for other reasons).

Neither zones nor instances were outputted before.

SU2_CFD/include/limiters/computeLimiters_impl.hpp
pcarruscag: Fair enough, I'll change that.

Common/include/toolboxes/C2DContainer.hpp
pcarruscag: Ah yes... The gradient function operates on any type that provides operator (iPoint, iVar, iDim) and the limiter function on any type with operator (iPoint, iVar).
Now the gradient of a single scalar (e.g. the "AuxVar") is stored in a 2D container which only gives (iPoint, iDim) so to make it compatible with the function we wrap it to give it a dummy middle dimension. The 2D view does something similar for 1D containers.

This could be explored a bit further to delete some more code, we compute gradients of grid velocities using least squares, those are stored in CGeometry which could be wrapped to again use the same function. This would require small changes elsewhere (the mpi for these is different too) and since those gradients are only used by the continuous adjoint I did not bother...

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -0,0 +1,189 @@
...
+#ifdef HAVE_OMP
+  constexpr size_t OMP_MAX_CHUNK = 512;
+
+  size_t chunkSize = computeStaticChunkSize(nPointDomain,
+                     omp_get_max_threads(), OMP_MAX_CHUNK);
+#endif
+
+  /*--- Start OpenMP parallel section. ---*/
+
+  SU2_OMP_PARALLEL
```
pcarruscag: This starts a parallel section, the code within {} that comes after is executed simultaneously by the threads.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    /*--- For each (non-halo) volume integrate over its faces (edges). ---*/
+
+    SU2_OMP_FOR_DYN(chunkSize)
+    for (size_t iPoint = 0; iPoint < nPointDomain; ++iPoint)
```
pcarruscag: This divides the loop over however many threads arrive here, without the worksharing construct the entire loop would be performed by each thread.
For reasons of spatial locality it is beneficial to divide the loop in chunks instead of individual iterations, in this case those chunks are assigned to each thread dynamically, this helps with load balancing but has some overhead. That is why you also see static work division for light loops (e.g. setting an array to 0).

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -0,0 +1,315 @@
...
+          AD::SetPreaccOut(gradient(iPoint, iVar, iDim));
+
+      AD::EndPreacc();
+    }
+
+    /*--- Correct the gradient values across any periodic boundaries. ---*/
+
+    if (solver != nullptr)
+    {
+      SU2_OMP_MASTER
```
pcarruscag: This is the OpenMP equivalent of `rank == MASTER_NODE`.

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -0,0 +1,315 @@
...
+      {
+        for (size_t iPeriodic = 1; iPeriodic <= config.GetnMarker_Periodic()/2; ++iPeriodic)
+        {
+          solver->InitiatePeriodicComms(&geometry, &config, iPeriodic, kindPeriodicComm);
+          solver->CompletePeriodicComms(&geometry, &config, iPeriodic, kindPeriodicComm);
+        }
+      }
+      SU2_OMP_BARRIER
```
pcarruscag: All threads must arrive here before execution continues, more important this gives all threads a coherent view of memory, which is fundamental here since only the master thread was involved in communicating halo values (without the barrier the other threads might not "see" the halo values correctly).
There are implicit barriers at the end of normal work sharing constructs, they can be removed by specifying the `nowait` clause, this is done in CSysVector methods (I think I explained why there, near the top of the cpp).

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
jayantmukho: what is the `chunkSize` dependent on? Is it hardware? number of variables?

Does simply adding `SU2_OMP_FOR_DYN(chunkSize)` before the for loop automatically run the loop in chunks? (assuming you add relevant includes and calls beforehand)

Does anything new need to be done in how the code is compiled or run to ensure that multiple threads are used?

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
jayantmukho: thanks for the additional explanation of the OpenMP routines

SU2_CFD/src/drivers/CDriver.cpp
jayantmukho: is this still a problem in the `SingleZoneDriver` and `MultiZoneDriver` as well? 

Its not really related to this PR, but might be worth addressing

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
pcarruscag: The size is usually based on amount of work per loop iteration, if the loop is heavy a small size can be used to improve load balancing, otherwise a relatively large size needs to be used to amortize the parallelization overheads. Currently most sizes are determined such that each thread gets on average the same number of chunks but each chunk is no larger than a maximum size, there is an utility function in omp_structure.hpp for this.

Yes, distributing the work over the threads that arrive at that location, if only one arrives the code still executes properly.

The build system was updated in #843 since that introduces the first solver that is completely hybrid parallel, there is no advantage in using the feature for fluid solvers yet.
Basically you add -Dwith-omp=true to meson, the number of threads per rank can be specified with the -t option when calling SU2_CFD, e.g. `mpirun -n 2 --bind-to numa SU2_CFD -t 8 config.cfg` if -t is omitted the number of threads is whatever is defined for the environment.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: To my knowledge only the HBDriver deals with multiple time instances... so it should not be a problem for anything else.

PR #835, opened 2019-12-10
Fix for Geometry Evaluation that allows mixed surface elements

jayantmukho:
## Proposed Changes
Small change in the way an airfoil section is created when the geometry is evaluated. Previously the slicing of the geometry would fail if there were quads on the surface. This PR fixes that and now geometries are correctly evaluated regardless of the type of element on the surface. 

It has been tested on a set of oneraM6 grids with different surface elements

All Triangles:
image

Mix of Quads and Triangles
image

Mostly Quads
image

Taking 25 cuts along the span for each of these meshes, the computed airfoil sections are identical. In the image below, you see the 25 slices of the wing taken from all three meshes. Each airfoil section has 3 lines overlaid lines (one from each mesh): 

image

## Related Work
NA

## PR Checklist


Common/src/geometry/CGeometry.cpp
```diff
@@ -2065,82 +2065,75 @@ void CGeometry::ComputeAirfoil_Section(su2double *Plane_P0, su2double *Plane_Nor
...
-             Xcoord_Index0.size() > Xcoord_Index1.size()! ---*/
-
-            if ((jPoint > iPoint) && (CrossProduct >= 0.0)
-                && ((AveXCoord > MinXCoord) && (AveXCoord < MaxXCoord))
-                && ((AveYCoord > MinYCoord) && (AveYCoord < MaxYCoord))
-                && ((AveZCoord > MinZCoord) && (AveZCoord < MaxZCoord))) {
+          if ((CrossProduct >= 0.0)
+              && ((AveXCoord > MinXCoord) && (AveXCoord < MaxXCoord))
+              && ((AveYCoord > MinYCoord) && (AveYCoord < MaxYCoord))
+              && ((AveZCoord > MinZCoord) && (AveZCoord < MaxZCoord))) {
```
jayantmukho: This looks messy, but the only real change in this PR is here. Instead of having an iNode and jNode loop that traverse the nodes of each element, there is one iFace loop that traverses the edges of the elements and sets iNode and jNode from that face. The rest of this routine is unchanged.

Previously for quads with iNode and jNode loops, it would try and find intersections along edges that did not exist (imagine a diagonal cut of a quad). This would then mess up the routine further down.

jayantmukho: The lines in question are above the code snippet linked

economon:
LGTM, and thanks for cleaning up some of my poorly-indented code in CGeometry :).

PR #840, opened 2019-12-12
Fix periodic boundary condition for nodes on axis of rotation

koodlyakshay:
## Proposed Changes

This is a very small fix for periodic boundary conditions. In the MatchPeriodic routine, there is a check to make sure we never map to the same point when finding periodic pairs. However, if a node lies on the axis of rotation it has to map on to itself. The checks in the MatchPeriodic routine does not allow for this possibility and this node would end up getting mapped to the nearest neighbor which will lead to wrong results. 

This PR introduces a small modification to the MatchPeriodic routine that will allow nodes on the axis to map onto themselves. 

## Related Work

As a test, consider the dummy geometry below. The domain is periodic about the Z-axis. I added a wall BC in the middle of the domain and the flow is supposed to separate at that point. This case is not very realistic but the main idea was to have periodic boundaries with points on the axis of rotation. The domain and the boundary conditions are similar to what we typically have when solving turbine blades. I wasn't sure if I can share the geometry, so I made this dummy test case. This case diverges with the develop branch. The image is from an intermediate result using the fix.
sample_test_case_fix

A quick way to verify if the nodes are matched properly is to check the velocity on the nodes that lie on the axis of rotation. Due to the periodicity constraint, the velocities must be aligned with the axis (here the Z-axis) which can be seen here as well.

vel

## PR Checklist

- [ X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [ X] My contribution is commented and consistent with SU2 style.

### DIFF
--- a/Common/src/geometry/CPhysicalGeometry.cpp
+++ b/Common/src/geometry/CPhysicalGeometry.cpp
@@ -7976,11 +7976,11 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
   su2double Theta, Phi, Psi, cosTheta, sinTheta, cosPhi, sinPhi, cosPsi, sinPsi;
   su2double rotCoord[3] = {0.0, 0.0, 0.0};
   
-  bool pointonAxis = false;
+  bool pointOnAxis = false;
   
-  bool chksamePoint = false;
+  bool chkSamePoint = false;
   
-  su2double dist_to_Axis = 0.0;
+  su2double distToAxis = 0.0;
 
   /*--- Tolerance for distance-based match to report warning. ---*/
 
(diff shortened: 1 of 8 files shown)


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7975,6 +7975,12 @@ void CPhysicalGeometry::MatchPeriodic(CConfig        *config,
   su2double rotMatrix[3][3] = {{1.0,0.0,0.0},{0.0,1.0,0.0},{0.0,0.0,1.0}};
   su2double Theta, Phi, Psi, cosTheta, sinTheta, cosPhi, sinPhi, cosPsi, sinPsi;
   su2double rotCoord[3] = {0.0, 0.0, 0.0};
+  
+  bool pointonAxis = false;
+  
+  bool chksamePoint = false;
+  
+  su2double dist_to_Axis = 0.0;
```
pcarruscag: Hi @koodlyakshay,
Please follow the most common naming convention (`pointOnAxis`, `distToAxis`).
Is this going to match points to themselves if they are on the axis?

koodlyakshay: Hi @pcarruscag,

Fixed the naming scheme. 
Also, yes the idea is to map the CV around a node on the axis on one periodic face with the corresponding CV around the same node on the other periodic face.

Issue #842, opened 2019-12-17
Error message for missing required inputs in .cfg file


themrdjj:
I just managed to run my first SU2 CFD simulation, however it took me a few days to get it to work. "SU2_CFD.exe" was always crashing right after "Begin solver". It took me quite a while to figure out that I was missing some required inputs in the ".cfg" file (for example CONV_NUM_METHOD_FLOW), especially because this isn't mentioned clearly on the website.

I would suggest the following solutions:

- An error message telling which required input is missing in the .cfg to help beginners with debugging.
- A simple list with all the required inputs, i.e. those which don't have a default value, for instance on the website.
- An example of a minimal working .cfg file, basically the opposite of [this one](https://github.com/su2code/SU2/blob/master/config_template.cfg).

(This is my first time opening a feature request on Github, I hope I did everything right)

pcarruscag:
Hi @themrdjj,
Thank you for the feedback, there was already a similar report in #796 (which incidentally went stale and was closed), I will try to do something about it, or if you would like to contribute code to the project I can point you to right place to add an error message.

Meanwhile my best advice is to not start a config from scratch until you know SU2 very well (and even then...) look for a test case that uses similar features and go from there.
The minimal config is the one in Quickstart, the template if more of a catalog :) I don't think we'll ever have a unified minimal config, SU2 does many things, some mutually exclusive.
Convective options are reasonably well documented here: https://su2code.github.io/docs_v7/Convective-Schemes/
The output messages before "Begin Solver" can be helpful (although in this particular case you would need to know what to expect).

Cheers,
Pedro

themrdjj:
Thanks for your reply! Could you please point me to the right place in the code? I will have a look and see if I can do something about it.

pcarruscag:
In CDriver::Numerics_Preprocessing(...), around line 2030, begins a (large) switch `switch (config->GetKind_ConvNumScheme_Flow()) {` which is where the numerics classes get instantiated, there is a check there for `NO_CONVECTIVE` which should trigger an error (´SU2_MPI::Error(...)´).
But it appears that the error is not triggered, I would start by checking what value CConfig::GetKind_ConvNumScheme_Flow returns when the option is not present in the config, possibly the default value (set somewhere in CConfig) is not `NO_CONVECTIVE`.
If you look for `CONV_NUM_METHOD_FLOW` in config_structure.cpp you will find the line that tries to read that option.

themrdjj:
[...]

PR #843, opened 2019-12-21
Hybrid parallel CFEASolver and CMeshSolver

pcarruscag:
## Proposed Changes
Another snapshot from #824.
All major routines of CFEASolver (and by association CMeshSolver) now use OpenMP directives.
Parallelization is at the level of each individual routine, i.e. threads are "started" locally.
This required avoiding the use of small auxiliary array member variables, as that is not thread safe (multiple threads cannot write to the same location) In doing that I ended up doing some cleanup and so I thought it would be better to move the files, strip spaces, etc., it should make the review easier actually.
I will document to the major changes below.

## Related Work
#824 #789 
Depends on #834 and #830

## PR Checklist


SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2638,492 +2632,125 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
...
-          switch (config->GetKind_Upwind_AdjFlow()) {
-            case ROE:
-              for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
-                delete numerics[val_iInst][iMGlevel][ADJFLOW_SOL][CONV_TERM];
-                delete numerics[val_iInst][iMGlevel][ADJFLOW_SOL][CONV_BOUND_TERM];
-              }
-              break;
-          }
+        if (!(properties_file.fail())) {
+          numerics[MESH_0][FEA_SOL][MAT_NHCOMP+offset]  = new CFEM_NeoHookean_Comp(nDim, nVar_FEM, config);
```
pcarruscag: The container can/should then be used with an "offset" of thread*MAX_TERMS.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2638,492 +2632,125 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
...
-        delete numerics[val_iInst][iMGlevel][ADJFLOW_SOL][SOURCE_SECOND_TERM];
-
-      }
-    }
-
-  }
+}
 
+void CDriver::Numerics_Postprocessing(CNumerics *****numerics, CSolver***, CGeometry**,
+                                      CConfig *config, unsigned short val_iInst) {
```
pcarruscag: I simplified (a lot) the numerics postprocessing to just delete everything.

meson.build
```diff
@@ -41,6 +41,13 @@ else
   endif
 endif
 
+# OpenMP
+omp = get_option('with-omp')
+
+if omp
+  omp_dep = dependency('openmp', language:'cpp')
+endif
+
```
pcarruscag: I updated meson to build with OpenMP support, add -Dwith-omp=true (default false) in the call to meson.py, the number of threads is still set via environment variable (see #830).

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -0,0 +1,3774 @@
...
+
+  for(auto iColor = 0ul; iColor < nColor; ++iColor) {
+    ElemColoring[iColor].size = coloring.getNumNonZeros(iColor);
+    ElemColoring[iColor].indices = coloring.innerIdx(iColor);
+  }
+
+  ColorGroupSize = 1; /// TODO: This needs to come from geometry or config
+
+  omp_chunk_size = computeStaticChunkSize(nPointDomain, omp_get_max_threads(), OMP_MAX_SIZE);
+#endif
```
pcarruscag: The element coloring is obtained from CGeometry as a shallow copy at construction.
I have yet to test the performance impact of the color group size, group size 1 gives maximum parallelization but worse spatial locality.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -0,0 +1,3774 @@
...
+    {
+      SU2_OMP_FOR_DYN(chunkSize)
+      for(auto k = 0ul; k < color.size; ++k) {
+
+        auto iElem = color.indices[k];
+#else
+    /*--- Natural coloring. ---*/
+    {
+      for (auto iElem = 0ul; iElem < nElement; iElem++) {
+#endif
```
pcarruscag: The coloring is then used like this and the body of the loop is almost unchanged apart from the use of local variables and picking one CElement and one CNumerics per thread.
When compiling without OpenMP support everything reverts to the natural way of looping over elements and no overhead or indirection is introduced by coloring.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -0,0 +1,3774 @@
...
+        int NUM_TERM = thread*MAX_TERMS + element_properties[iElem]->GetMat_Mod();
+
+        numerics[NUM_TERM]->Compute_Tangent_Matrix(element, config);
+
+        /*--- Update residual and stiffness matrix with contributions from the element. ---*/
+        for (iNode = 0; iNode < nNodes; iNode++) {
+
+          auto Ta = element->Get_Kt_a(iNode);
+          for (iVar = 0; iVar < nVar; iVar++)
+            LinSysRes[indexNode[iNode]*nVar+iVar] -= simp_penalty*Ta[iVar];
```
pcarruscag: In a number of places we now write directly to CSysVector / CSysMatrix instead of populating a temporary first.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -0,0 +1,3774 @@
...
+  const bool topology_mode = config->GetTopology_Optimization();
+  const su2double simp_exponent = config->GetSIMP_Exponent();
+
+  const unsigned short nStress = (nDim == 2) ? 3 : 6;
+
+  /*--- Reduction variable to compute the maximum von Misses stress,
+   *    each thread uses the first element of each row, the rest of
+   *    the row is padding to prevent false sharing. ---*/
+  su2activematrix auxMaxVonMisses(omp_get_max_threads(), 64/sizeof(su2double));
+  auxMaxVonMisses.setConstant(0.0);
```
pcarruscag: This is probably the trickiest bit in this PR, reduction operations in general are tricky, I can try to clean it up a bit by using the OpenMP reduction clause but this "manual" way is future compatible with the AD type (OpenMP reductions are only defined for builtin types, they can be declared for other types but that sounds like witchcraft to me).

SU2_CFD/src/SU2_CFD.cpp
```diff
@@ -39,24 +39,26 @@
...
+  std::string filename = "default.cfg";
+
   /*--- Command line parsing ---*/
-  
+
   CLI::App app{"SU2 v7.0.0 \"Blackbird\", The Open-Source CFD Code"};
-  app.add_flag("-d,--dryrun", dry_run, "Enable dry run mode.\n" 
+  app.add_flag("-d,--dryrun", dry_run, "Enable dry run mode.\n"
                                        "Only execute preprocessing steps using a dummy geometry.");
+  app.add_option("-t,--threads", num_threads, "Number of OpenMP threads per MPI rank.");
```
pcarruscag: The number of threads per rank can now be set via this command line option, it defaults to whatever is defined by the environment.

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: Update: The impact of not grouping elements for coloring (size 1) is not significant for the FEA solver, not a lot of data is accessed per element relative to the computation effort and linear algebra takes up 80%+ of the time anyway.

Common/src/config_structure.cpp
```diff
@@ -1442,6 +1442,8 @@ void CConfig::SetConfig_Options() {
   addUnsignedLongOption("LINEAR_SOLVER_RESTART_FREQUENCY", Linear_Solver_Restart_Frequency, 10);
   /* DESCRIPTION: Relaxation factor for iterative linear smoothers (SMOOTHER_ILU/JACOBI/LU-SGS/LINELET) */
   addDoubleOption("LINEAR_SOLVER_SMOOTHER_RELAXATION", Linear_Solver_Smoother_Relaxation, 1.0);
+  /* DESCRIPTION: Custom number of threads used for additive domain decomposition for ILU and LU_SGS (0 is "auto"). */
+  addUnsignedLongOption("LINEAR_SOLVER_PREC_THREADS", Linear_Solver_Prec_Threads, 0);
```
pcarruscag: With this PR the ILU and LU_SGS preconditioners can now run with a custom number of threads specified through the config file.
For structural cases (and mesh deformation) it gets harder and harder (more linear iterations) to solve the linear systems as more processors are throw at the problem (which is annoying in multizone problems), being able to independently control the number of preconditioner threads alleviates the problem.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -0,0 +1,3599 @@
...
+          simp_penalty = simp_minstiff+(1.0-simp_minstiff)*pow(density,simp_exponent);
+        }
+
+        /*--- Set the properties of the element ---*/
+        element->Set_ElProperties(element_properties[iElem]);
+
+        /*--- Compute the components of the jacobian and the stress term, one numerics per thread. ---*/
+        int NUM_TERM = thread*MAX_TERMS + element_properties[iElem]->GetMat_Mod();
+
+        numerics[NUM_TERM]->Compute_Tangent_Matrix(element, config);
```
rsanfer: What's the advantage of having one numerics term per thread? Also, I think I missed the point where the numerics container is extended beyond MAX_TERMS

SU2_CFD/src/solvers/CFEASolver.cpp
rsanfer: I like it, it's cleaner like this.

SU2_CFD/src/solvers/CFEASolver.cpp
rsanfer: Is this linked to the number of threads of the preconditioners, or totally independent?

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -189,94 +179,99 @@ CMeshSolver::CMeshSolver(CGeometry *geometry, CConfig *config) : CFEASolver(true
 
 }
 
-CMeshSolver::~CMeshSolver(void) {
```
rsanfer: Moved to .hpp?

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -540,30 +525,28 @@ void CMeshSolver::UpdateDualGrid(CGeometry *geometry, CConfig *config){
...
-    /*--- Coordinates of the current point at n+1, n, & n-1 time levels ---*/
+  SU2_OMP(parallel for schedule(static,omp_chunk_size))
+  for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
 
-    Disp_nM1 = nodes->GetSolution_time_n1(iPoint);
-    Disp_n   = nodes->GetSolution_time_n(iPoint);
-    Disp_nP1 = nodes->GetSolution(iPoint);
+    /*--- Coordinates of the current point at n+1, n, & n-1 time levels. ---*/
+
+    const su2double* Disp_nM1 = nodes->GetSolution_time_n1(iPoint);
```
rsanfer: Is this for efficiency reasons?

TestCases/parallel_regression_AD.py
```diff
@@ -261,8 +261,8 @@ def main():
     discadj_fea           = TestCase('discadj_fea')
     discadj_fea.cfg_dir   = "disc_adj_fea"
     discadj_fea.cfg_file  = "configAD_fem.cfg" 
-    discadj_fea.test_iter = 9
-    discadj_fea.test_vals = [-6.070230, -6.262517, -0.000364, -8.708700] #last 4 columns
+    discadj_fea.test_iter = 4
```
rsanfer: Why changing these test cases, are the results for iteration 9 still the same as before?

meson.build
rsanfer: Thanks! That should make it much easier to build.

Common/include/option_structure.hpp
```diff
@@ -2646,17 +2643,15 @@ class COptionConvect : public COptionBase {
...
+    SetDefault();
     return badValue(option_value, "convect", this->name);
 
   }
 
   void SetDefault() {
     this->centered = NO_CENTERED;
     this->upwind = NO_UPWIND;
-    this->space = SPACE_CENTERED;
+    this->space = NO_CONVECTIVE;
```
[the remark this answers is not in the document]
pcarruscag: Thank @koodlyakshay 
@rsanfer you asked about this below, it addresses #842

Common/src/geometry/CGeometry.cpp
```diff
@@ -3531,37 +3572,44 @@ void CGeometry::GetGlobalElementAdjacencyMatrix(vector<unsigned long> &neighbour
...
 
   /*--- Allocate ---*/
   unsigned long matrix_size = neighbour_start[Global_nElemDomain];
   neighbour_idx = new long [matrix_size];
-  /*--- Initialize ---*/
-  for(iElem=0; iElem<matrix_size; ++iElem) neighbour_idx[iElem] = -1;
-  /*--- Populate ---*/
-  for(iElem=0; iElem<nElem; ++iElem)
+
+  SU2_OMP_PARALLEL
```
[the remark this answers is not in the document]
pcarruscag: "Business as usual" there should be no runtime overhead, no coloring is applied for the loops, the SU2_OMP macros are disabled and the "omp_xxx" functions are constexpr so the compiler should optimize them away very easily.

Common/src/config_structure.cpp
pcarruscag: Yup it uses the same threads that are used everywhere else.
To answer another of your questions:
> When using the mesh solver in a CFD case (so, single zone), is it possible to use different levels of parallelism for the CFD problem and for the mesh deformation problem? And, if so, is it something worth doing?

Although one could change the number of threads used in all mesh deformation and fea routines very easily, at the moment the mesh deformation runs synchronously with the rest of the code, and so running the discretization parts with fewer threads would not be advantageous.
The area where it might be worth reducing the number of threads is those 2 preconditioners, if and only if:
- The linear solver is not able to reach adequate convergence (or is diverging) due to the preconditioner being too weak (too partitioned).
- The code is running on a very bandwidth-limited machine, in which case reducing threads a bit might give a speedup due to fewer linear iterations being required.

Other areas of the code will not suffer as much from having too many threads, since threads do not increase the number of halo points (communication costs) and allow for "automatic" load balancing.

Common/include/config_structure.hpp
```diff
@@ -569,16 +569,17 @@ class CConfig {
...
-  unsigned long Deform_Linear_Solver_Iter;        /*!< \brief Max iterations of the linear solver for the implicit formulation. */
-  unsigned long Linear_Solver_Iter_FSI_Struc;	  /*!< \brief Max iterations of the linear solver for FSI applications and structural solver. */
-  unsigned long Linear_Solver_Restart_Frequency;  /*!< \brief Restart frequency of the linear solver for the implicit formulation. */
-  unsigned short Linear_Solver_ILU_n;		      /*!< \brief ILU fill=in level. */
+  bool Inc_Inlet_UseNormal;        /*!< \brief Flag for whether to use the local normal as the flow direction for an incompressible pressure inlet. */
+  su2double Linear_Solver_Error;   /*!< \brief Min error of the linear solver for the implicit formulation. */
+  su2double Deform_Linear_Solver_Error;          /*!< \brief Min error of the linear solver for the implicit formulation. */
+  su2double Linear_Solver_Error_FSI_Struc;		   /*!< \brief Min error of the linear solver for the implicit formulation in the structural side for FSI problems . */
+  su2double Linear_Solver_Smoother_Relaxation;   /*!< \brief Relaxation factor for iterative linear smoothers. */
```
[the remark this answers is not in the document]
pcarruscag: This and option_structure get me every time... you think things are aligned... thanks I'll fix it

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -556,16 +578,8 @@ class CSysMatrix {
...
-    unsigned long iVar, index;
-
-    for (index = row_ptr[block_i]; index < row_ptr[block_i+1]; index++) {
-      if (col_ind[index] == block_i) { // Only elements on the diagonal
-        for (iVar = 0; iVar < nVar; iVar++)
-          matrix[index*nVar*nVar+iVar*nVar+iVar] += PassiveAssign<ScalarType,OtherType>(val_matrix);
-        break;
-      }
-    }
+    for (auto iVar = 0ul; iVar < nVar; iVar++)
```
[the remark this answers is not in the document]
pcarruscag: ul is for unsigned long (which I was getting tired to write xD) otherwise auto will deduce int, which is the default type for integer literals.

SU2_CFD/obj/Makefile.am
```diff
@@ -161,8 +161,7 @@ libSU2Core_sources = ../src/definition_structure.cpp \
   ../src/variables/CAdjEulerVariable.cpp \
   ../src/variables/CDiscAdjVariable.cpp \
   ../src/variables/CIncNSVariable.cpp \
-  ../src/variables/CEulerVariable.cpp \
-  ../src/limiters/computeLimiters.cpp
+  ../src/variables/CEulerVariable.cpp
```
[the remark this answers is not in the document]
pcarruscag: There isn't one I just forgot to delete that in the previous PR (there was a build issue... it's something in my agenda for when I visit you guys).

SU2_CFD/src/SU2_CFD.cpp
```diff
@@ -92,117 +94,107 @@ int main(int argc, char *argv[]) {
...
+
+    /*--- Turbomachinery problem. ---*/
+    driver = new CTurbomachineryDriver(config_file_name, nZone, MPICommunicator);
+
+  }
+  else {
+
+    /*--- Instantiate the class for external aerodynamics by default. ---*/
+    driver = new CFluidDriver(config_file_name, nZone, MPICommunicator);
+
```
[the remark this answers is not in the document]
pcarruscag: I guess the fluid driver is not needed anymore? Its just a base for the turbomachinery drivers right?

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -3495,7 +3495,7 @@ void COutputLegacy::MergeSolution(CConfig *config, CGeometry *geometry, CSolver
       
       /*--- Loop over this partition to collect the current variable ---*/
       
-      jPoint = 0; su2double *Stress;
+      jPoint = 0; const su2double *Stress;
```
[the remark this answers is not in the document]
pcarruscag: here here

TestCases/parallel_regression_AD.py
pcarruscag: This case is super sensitive to minor changes. I think that is because it is converged to machine precision, so as an experiment I'm converging it less...
The residuals at the 9th iteration were not the same (still in the -6 ballpark) but the derivatives were.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1986,19 +1988,11 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
...
-    for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
-      numerics[iMGlevel] = new CNumerics** [MAX_SOLS];
-      for (iSol = 0; iSol < MAX_SOLS; iSol++)
-        numerics[iMGlevel][iSol] = new CNumerics* [MAX_TERMS];
-    }
+
+  for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
+    numerics[iMGlevel] = new CNumerics** [MAX_SOLS];
+    for (iSol = 0; iSol < MAX_SOLS; iSol++)
+      numerics[iMGlevel][iSol] = new CNumerics* [MAX_TERMS*omp_get_max_threads()]();
```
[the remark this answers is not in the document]
pcarruscag: It looked like we only have 5 material models / terms:
```
const int FEA_TERM = 0;      /*!< \brief Position of the finite element analysis terms in the numerics container array. */
const int DE_TERM = 1;       /*!< \brief Position of the dielectric terms in the numerics container array. */

const int MAT_NHCOMP  = 2;   /*!< \brief Position of the Neo-Hookean compressible material model. */
const int MAT_IDEALDE = 3;   /*!< \brief Position of the Ideal-DE material model. */
const int MAT_KNOWLES = 4;   /*!< \brief Position of the Knowles material model. */
```
And so 6 max terms would be enough?

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: Totally independent.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -1266,45 +1266,52 @@ template<class ScalarType>
...
 
-  /*--- Both row and column associated with node i are eliminated (Block_ii = I and all else 0) to preserve eventual symmetry. ---*/
-  /*--- The vector is updated with the product of column i by the known (enforced) solution at node i. ---*/
+  /*--- Eliminate the row associated with node i (Block_ii = I and all other Block_ij = 0).
+   *    To preserve eventual symmetry, also attempt to eliminate the column, if the sparse pattern is not
+   *    symmetric the entire column may not be eliminated, the result (matrix and vector) is still correct.
+   *    The vector is updated with the product of column i by the known (enforced) solution at node i. ---*/
 
-  unsigned long iPoint, iVar, jVar, index, mat_begin;
+  for (auto index = row_ptr[node_i]; index < row_ptr[node_i+1]; ++index) {
```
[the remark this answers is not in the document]
pcarruscag: If for some reason we need / want to change to something else it is automatic, other than that it is faster to write and makes the lines shorter.

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: Deleted as there is nothing to explicitly delete in this class, the compiler will create a default.

pcarruscag:
Thank you for the thorough review @rsanfer! I'll reply to your main questions and some of the smaller ones here to centralize things.

> Just a request, if it's possible that you add one or two test cases so the implementation is safe onwards (and, of course, so I can play around with the new features a little bit ).

The testcases are the same, no changes there other than the one optional option introduced above. When the hybrid stuff covers most of the code I would add an entire build configuration e.g. BaseMPIOMP and corresponding testcase suite.

> * Should this just run "out of the box" with a working installation of OpenMP in any machine, or is there anything else _fancy_ needed?

I would leave it to the community to decide what the defaults should be, probably for a lot of new users that don't run on clusters just calling SU2_CFD and not having to worry about mpi would be nice (a lot of the issues on CFD online are mpi related).

> * Is the previous behaviour exactly kept, or are there any modifications in the basic, non OpenMP version of code? (Not that I mind, just curious).

Other than the algorithmic changes (but mathematically equivalent) introduced to limiters and gradients in #834, yes.

> What's the advantage of having one numerics term per thread?

It is a requirement, we need to write data into numerics before using them, multiple threads cannot write to the same location (i.e. the internal structures of CNumerics) therefore one per thread is required.

> ...Also, I think I missed the point where the numerics container is extended beyond MAX_TERMS

The allocation of space for one numerics per thread is done above in line 1995 of my 21 Dec 2019 comment: `...MAX_TERMS*omp_get_max_threads()...`.
The instantiation of one numerics per thread is then done by executing the rest of the preprocessing in parallel and instead of using `XYZ_TERM` using `XYZ_TERM+offset` where `offset = thread_id * MAX_TERMS`.
I think someone mentioned this (maybe Tim) that we could revisit the ownership relations of the numerics classes, i.e. allocate them as members of their respective solvers, which if we do, we can think of having a purpose built container that automates the per-thread creation and access.

> Why are they redefined each time inside the loop?
> Is this for efficiency reasons?

Referring to variables being declared inside loops. One stylist reason is that declaring everything at the top of a function is the C way of doing things, the C++ people whose books/blogs I've read and talks I've watched, recommend keeping namespaces (the inside of the loop being one) as clean as possible.
The only reason not to do this is if you explicitly want re-use, in the case of trivial types this does not improve efficiency, and in the context of OpenMP code it can create issues. Just like we need one numerics per thread, if we declare variables outside a parallel loop the default OpenMP behaviour is to consider them shared, and concurrent writes to shared locations = gdb and many bad words xD.

> Also, just an additional (hopefully constructive) comment: I find all of these developments great, and I honestly think that you are doing an amazing job on performance and overall code improvement. However, as a non-C++-master myself, I'm just a little concerned of whether some advanced programming may become an entrance barrier to new additions to the community.

As I wrote in the preamble of #789:
"But please participate even if you never heard of these topics, your opinion about readability and "developability" of the code is important! I think the code-style should be accessible to people starting a PhD (after they read a bit about C++...)."
I try to encapsulate and hide the tricky bits as much as possible to make the code as readable as possible, whether I am succeeding or not is for the community to decide, in all these PR's I've been pointing to the areas I think are trickier, if someone, anyone, feels they are absolutely incomprehensible please say something... either here, or trough slack, or by email (I think it shows in the commits) (I understand not everyone is keen on github exposure).

> I'm aware that you have been doing very well at documenting the code and the various PRs, but I'd say we should try to find an strategy to ease the learning curve on potential new developers (maybe some developer tutorials? a collection of the comments/discussions on the PRs moved to the wiki? a list of links/useful resources?).

I agree with documentation of broad design decisions, that is the intent of #789, and developer tutorials (how to implement a new X) once we are content with the restructurings, otherwise they will quickly go outdated... or actually...
We should probably first think about the answers to "how to implement a new X" and restructure/refactor as a function of that.
Based on previous efforts of maintaining wiki's updated while code is being developed, I much prefer this github style where you can clearly tell what version of the code the comments refer to. A collection of comments/discussions organized by topic and linked to a feature is somewhat what I had in mind when I opened a "big PR" (#824) with little branches such as this one, I can try to complete that with a list of links/useful resources, references as it were, good idea!

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: see big reply

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: see big reply

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -1323,31 +1323,25 @@ void CMultiGridGeometry::SetMultiGridWallTemperature(CGeometry *geometry, unsign
...
 
     /*--- Loop over all of the children for this coarse CV and compute
      a grid velocity based on the values in the child CVs (fine mesh). ---*/
-    for (iChild = 0; iChild < node[Point_Coarse]->GetnChildren_CV(); iChild++) {
-      Point_Fine    = node[Point_Coarse]->GetChildren_CV(iChild);
-      Area_Child    = fine_mesh->node[Point_Fine]->GetVolume();
-      Grid_Vel_Fine = fine_mesh->node[Point_Fine]->GetGridVel();
-      for (iDim = 0; iDim < nDim; iDim++)
+    for (unsigned short iChild = 0; iChild < node[Point_Coarse]->GetnChildren_CV(); iChild++) {
+      unsigned long Point_Fine       = node[Point_Coarse]->GetChildren_CV(iChild);
```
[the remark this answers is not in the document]
pcarruscag: see big reply

Issue #694, opened 2019-06-03
STL-writer does not handle quad surface elements


TobiKattmann:
**Describe the bug**
*A clear and concise description of what the bug is and what you expect the behavior to be instead. If applicable, add screenshots to help explain your problem.*

Hi all,

The STL-file writer does not write quad surface elements (So far so good :) ). So in case of a Hexa-Grid (i.e. quad surface elements) the .stl is pretty much empty, which is not really what we want.
I will attach an image later (firewall's at it again) down below which shows a mixed surface mesh (tris and quads). On the left will be the current version of SU2 where quads are simply not written and on the right a fixed version where each quad is separated into 2 tris.

![stl_quads_tris](https://user-images.githubusercontent.com/31306376/58827759-17667700-8644-11e9-9e6f-4643d2e7fdb7.JPG)

That was found (and fixed) by Fabian Woerenkaemper @ElGarno . As the output is currently overhauled, I shortly chatted with its adoptive father @talbring and I will adapt the fixed STL-writer to the new structure and it will (hopefully) go into develop with that in a somewhat near future.

If you desperately ;) need that fix right now, I can add a patch for the current develop down below.

Cheers, Tobi

**Desktop (please complete the following information):**
 - SU2 Version: latest develop, but this problem is probably also present in older versions.

TobiKattmann:
![stl_quads_tris](https://user-images.githubusercontent.com/31306376/58827730-ff8ef300-8643-11e9-9737-f9fb595e06ec.JPG)

PR #847, opened 2020-01-09
Add *.stl (STereoLithography) output format 

TobiKattmann:
## Proposed Changes
Hi all,

This PR adds the possibility to write ASCII STL files. Example usage:
`OUTPUT_FILES= (RESTART,PARAVIEW, STL, ...)`
All surfaces in `MARKER_PLOTTING` will be used, just like the surface files.

The major problem was to re-process the element connectivity for parallel cases from what the data available after the parallel data sorter does its thing. You'll notice that in the `CSTLFileWriter.cpp` : Part of the code was copied over from the TecIO stuff and modified to our needs. The code probably could be nicer and it could be better documented, but I already forgot some details and it most likely would still be a mess to understand. Big thanks to @talbring for making this code work.

I added a simple regression test by `diff`ing against a reference stl-file from in the Testcases repo.

I'll add that information [on the website here](https://su2code.github.io/docs_v7/Custom-Output/) once this PR is merged.

## Related Work
Resolves issue #694 (which was marked stale and was closed in the meantime)

## PR Checklist


Common/include/option_structure.hpp
```diff
@@ -1478,21 +1478,24 @@ static const map<string, ENUM_INPUT> Input_Map = CCreateMap<string, ENUM_INPUT>
...
+  TECPLOT_BINARY          = 2,  /*!< \brief Tecplot binary format for the solution output. */
+  SURFACE_TECPLOT         = 3,  /*!< \brief Tecplot format for the solution output. */
+  SURFACE_TECPLOT_BINARY  = 4,  /*!< \brief Tecplot binary format for the solution output. */
+  CSV                     = 5,  /*!< \brief Comma-separated values format for the solution output. */
+  SURFACE_CSV             = 6,  /*!< \brief Comma-separated values format for the solution output. */
+  PARAVIEW                = 7,  /*!< \brief Paraview ASCII format for the solution output. */
+  PARAVIEW_BINARY         = 8,  /*!< \brief Paraview binary format for the solution output. */
+  SURFACE_PARAVIEW        = 9,  /*!< \brief Paraview ASCII format for the solution output. */
+  SURFACE_PARAVIEW_BINARY = 10, /*!< \brief Paraview binary format for the solution output. */
+  MESH                    = 11, /*!< \brief SU2 mesh format. */
```
WallyMaier: Thanks for cleaning this up :)

SU2_CFD/include/output/filewriter/CParallelDataSorter.hpp
```diff
@@ -31,6 +31,7 @@
 #include "../../../../Common/include/option_structure.hpp"
 #include "../../../../Common/include/toolboxes/CLinearPartitioner.hpp"
 
+
```
WallyMaier: extra line?

SU2_CFD/include/output/filewriter/CSurfaceFVMDataSorter.hpp
```diff
@@ -67,13 +67,52 @@ class CSurfaceFVMDataSorter final: public CParallelDataSorter{
...
+   * \brief Get the ending global renumbered node ID of the linear partition owned by a specific processor.
+   * \param[in] rank - the processor rank.
+   * \return The ending global renumbered node ID.
+   */
+  unsigned long GetNodeEnd(unsigned short rank) override {
+    for (unsigned short iRank = 0; iRank < size-1; iRank++){
+      return nPoint_Recv[iRank+1]-1;
+    }
+    return nPoint_Recv[size];
+  }
```
pcarruscag: This is weird.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+  /* Sorted list of halo nodes for this MPI rank. */
+  sorted_halo_nodes.assign(halo_nodes.begin(), halo_nodes.end());
+
+  /*--- We effectively tack the halo nodes onto the end of the node list for this partition.
+    TecIO will later replace them with references to nodes in neighboring partitions. */
+  num_halo_nodes = sorted_halo_nodes.size();
+  vector<int64_t> halo_node_local_numbers(max((size_t)1, num_halo_nodes)); /* Min size 1 to avoid crashes when we access these vectors below. */
+  vector<int32_t> neighbor_partitions(max((size_t)1, num_halo_nodes));
+  vector<int64_t> neighbor_nodes(max((size_t)1, num_halo_nodes));
```
pcarruscag: I've seen, int, unsigned long, size_t, int32, int64, size_t, ... is this really needed? or can we make do with just a couple types (just because the static_casts everywhere are ugly).

SU2_CFD/include/output/filewriter/CSurfaceFVMDataSorter.hpp
talbring: Hmm indeed. That looks like a strange loop. Cannot remember why we did it like that. Gonna check that.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,328 @@
...
+                            nDim){}
+
+
+CSTLFileWriter::~CSTLFileWriter(){}
+
+
+void CSTLFileWriter::Write_Data(){
+
+  /*--- This Write_Data routine has 3 major parts:
+    1. Prerequisite info: The parallel data-sorter distributes nodes of the primal mesh onto the processes
```
economon: Looks like you've already divided this into 3 logical tasks.. when you do the clean up, can you please decompose this into 3 smaller methods if possible? For functions/methods, smaller is better, and they should do 1 thing.

SU2_CFD/include/output/filewriter/CParallelDataSorter.hpp
TobiKattmann: I somehow like 2 line spacing between some parts in a file so that it is better noticeable while scrolling e.g. in between function implementations in .cpp files. But I removed that newline now ... I guess its a matter of preference

Common/include/option_structure.hpp
TobiKattmann: 😊

SU2_CFD/include/output/filewriter/CSurfaceFVMDataSorter.hpp
TobiKattmann: I am also a bit clueless what happened there, but that function never got called anyway... which kindof makes it even worse. So I removed it and also two additional useless code-lines.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: This is copied TecIO code, so it was probably important for tecplot output. I removed all size_t, int32, int64 and replaced them by unsigned long, int, long. I got rid of a few casts, but not all.

If you want to take a second look, the code is now in a seperate ReprocessElementConnectivity routine

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: I now made two separate functions for the first 2 points: `ReprocessElementConnectivity ` and `GatherCoordData`. In the `Write_Data() `method now the two functions are called and then Write_Data still has the code in it which actually writes the data to file.

SU2_CFD/include/output/filewriter/CSurfaceFVMDataSorter.hpp
TobiKattmann: And I made the all "Get" functions I added `const` where possible

SU2_CFD/include/output/filewriter/CSTLFileWriter.hpp
```diff
@@ -0,0 +1,111 @@
...
+   */
+  void GatherCoordData();
+
+  /*!
+   * \brief Get the halo-node value of a global renumbered Point for a specific variable.
+   * \param[in] global_node_number - global renumbered Point ID
+   * \param[in] iVar - Variable number
+   * \return Value of the halo-node variable
+   */
+  double GetHaloNodeValue(unsigned long global_node_number, unsigned short iVar);
```
talbring: Make the return type  `passivedouble`

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,352 @@
...
+                nParallel_Quad = dataSorter->GetnElem(QUADRILATERAL);
+  unsigned short iVar;
+
+  num_nodes_to_receive.resize(size, 0);
+  values_to_receive_displacements.resize(size);
+  halo_nodes.clear();
+  sorted_halo_nodes.clear();
+
+  /* We output a single, partitioned zone where each rank outputs one partition. */
+  vector<int> partition_owners;
```
talbring: This variable is not used anymore

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,352 @@
...
+  /*--- We effectively tack the halo nodes onto the end of the node list for this partition.
+    TecIO will later replace them with references to nodes in neighboring partitions. */
+  num_halo_nodes = sorted_halo_nodes.size();
+  vector<long> neighbor_partitions(max(static_cast<unsigned long>(1), num_halo_nodes));
+  vector<long>      neighbor_nodes(max(static_cast<unsigned long>(1), num_halo_nodes));
+
+  for(unsigned long i = 0; i < num_halo_nodes; ++i) {
+    int owning_rank = dataSorter->FindProcessor(sorted_halo_nodes[i]);
+    unsigned long node_number = sorted_halo_nodes[i] - dataSorter->GetNodeBegin(owning_rank);
+    neighbor_partitions[i] = owning_rank; /* Partition numbers are 1-based, i.e. start with 1! */
```
talbring: Also remove that comment

TobiKattmann: Done

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: removed together with all other occurences

SU2_CFD/include/output/filewriter/CSTLFileWriter.hpp
TobiKattmann: Done in prototype and implementation

Common/include/option_structure.hpp
```diff
@@ -1506,7 +1509,9 @@ static const map<string, ENUM_OUTPUT> Output_Map = CCreateMap<string, ENUM_OUTPU
 ("SURFACE_PARAVIEW", SURFACE_PARAVIEW_BINARY)
 ("RESTART_ASCII", RESTART_ASCII)
 ("RESTART", RESTART_BINARY)
-("CGNS", CGNS);
+("CGNS", CGNS)
+("STL", STL)
+("STL_BINARY", STL_BINARY);
```
economon: Can you please add an error check for STL_BINARY if there isn't one already in case someone tries to use this?

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
economon: 👍

Common/include/option_structure.hpp
TobiKattmann: Done

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,346 @@
...
+          bufD_Send[index] = GetHaloNodeValue(global_node_number, iVar);
+        }
+
+        index++;
+      }//iVar
+    }//iPoint
+  }//iElem
+
+  /*--- Collective comms of the solution data and global IDs. ---*/
+  SU2_MPI::Gather(bufD_Send, static_cast<int>(MaxLocalTriaAll*3*3), MPI_DOUBLE,
```
economon: Use N_POINTS_TRIANGLE, etc., to be more clear (this is in a few places)

TobiKattmann: Do you mean replacing `MaxLocalTriaAll*3*3 ` by `N_POINTS_TRIANGLE` ? This is actually a coordinate (9 per triangle) count. I introduced `max_nLocalCoords = MaxLocalTriaAll*3*3` -> Does that fully address your point?

pcarruscag: I think Tom is referring to the number 3 being replaced by the constant defined in option_structure.
Of course you have two threes xD, one for points the other for dimensions (i.e. don't do  N_POINTS_TRIANGLE*N_POINTS_TRIANGLE).

economon: 👍

TobiKattmann: 🤦‍♂

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+  /*--- Allocate buffer for send/recv of the coordinate data. Only the master rank allocates buffers for the recv. ---*/
+  bufD_Send = new su2double[max_nLocalCoords](); /* Triangle has 3 Points with 3 coords each */
+  if (rank == MASTER_NODE)
+    bufD_Recv = new su2double[nProcessor*max_nLocalCoords];
+
+  /*--- Load send buffers with the local data on this rank. ---*/
+  /*--- Write triangle element coordinate data into send buffer---*/
+  unsigned long index = 0;
+  for (iElem = 0; iElem < nLocalTria; iElem++) {
+    for (iPoint = 0; iPoint < 3; iPoint++) {
```
pcarruscag: This is an N_POINTS_TRIANGLE case.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+                nProcessor = size,
+                iElem,
+                nLocalTria,
+                nLocalQuad,
+                nLocalTriaAll = 0,
+                global_node_number;
+
+  su2double *bufD_Send = NULL;
+
+  vector<unsigned short> Nodelist = {0,1,3, 1,2,3}; // for Quad2Tri, assumes clockwise or counterclockwise rotation
```
pcarruscag: This is static sized, you do not need to use `vector`

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+      for (iVar = 0; iVar < 3; iVar++){
+        if (dataSorter->FindProcessor(global_node_number) == rank) {
+          bufD_Send[index] = dataSorter->GetData(iVar, local_node_number);
+        } else {
+          bufD_Send[index] = GetHaloNodeValue(global_node_number, iVar);
+        }
+
+        index++;
+      }//iVar
+    }//iPoint
```
pcarruscag: I would make this loop a function and call it twice, once with TRIANGLE and a node list {0 1 2}, then with the QUADRILATERAL node list.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+    for (iProcessor = 0; iProcessor < nProcessor; iProcessor++) {
+      for (iElem = 0; iElem < Buffer_Recv_nTriaAll[iProcessor]; iElem++) { // loops over nLocalTriaAll
+
+        /*--- Write the coordinate data for each node in a triangle. ---*/
+        /*--- Every tested software recomputes the normal anyway,
+              such that this arbitray face normal does not matter.  ---*/
+        Surf_file << "facet normal " << 1 << " " << 2 << " " << 3 << endl;
+        Surf_file << "    outer loop" << endl; // 4 leading whitespaces
+
+        for(iPoint = 0; iPoint < 3; iPoint++) {
```
pcarruscag: This is a N_POINTS_TRIANGLE case

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
pcarruscag: You may also go over the node list with a range for (`for(auto node : Nodelist)`).
By the way there is a wee bit of a naming scheme mix going on, dataSorter Nodelist, local_node_number...

SU2_CFD/include/output/filewriter/CSurfaceFVMDataSorter.hpp
```diff
@@ -67,13 +67,40 @@ class CSurfaceFVMDataSorter final: public CParallelDataSorter{
 
   /*!
    * \brief Get the global index of a point.
-   * \input iPoint - the point ID.
+   * \param[in] iPoint - the local renumbered point ID.
    * \return Global index of a specific point.
    */
-  unsigned long GetGlobalIndex(unsigned long iPoint)  override{
+  unsigned long GetGlobalIndex(unsigned long iPoint) override{
```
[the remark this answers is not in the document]
TobiKattmann: This took me a bit because this routine `GetGlobalIndex `returns a value of a (class data) std::map<ul, ul> via operator[] ... which is not const [stackoverflow](https://stackoverflow.com/questions/1474894/why-isnt-the-operator-const-for-stl-maps) . Using the `.at(...)` function was the way out ([compilerExplorer code](https://gcc.godbolt.org/z/esFpGx))

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+     (sorted_halo_nodes) whose data we need to receive, and receiving
+     lists of nodes whose data we need to send. */
+  if (sorted_halo_nodes.empty()) sorted_halo_nodes.resize(1); /* Avoid crash. */
+  SU2_MPI::Alltoallv(&sorted_halo_nodes[0], &num_nodes_to_receive[0], &nodes_to_receive_displacements[0], MPI_UNSIGNED_LONG,
+                     &nodes_to_send[0],     &num_nodes_to_send[0],    &nodes_to_send_displacements[0],    MPI_UNSIGNED_LONG,
+                     MPI_COMM_WORLD);
+
+  /* Now actually send and receive the data */
+  data_to_send.resize(max(1, total_num_nodes_to_send * (int)fieldnames.size()));
+  halo_var_data.resize(max(static_cast<unsigned long>(1), fieldnames.size() * num_halo_nodes));
```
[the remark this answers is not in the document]
TobiKattmann: I used the `max<the type you want>(...)` option. Didn't know this style exists `1ul` :)

SU2_CFD/include/output/filewriter/CSTLFileWriter.hpp
```diff
@@ -0,0 +1,112 @@
...
+ * \brief Class for writing STL output files.
+ * \author T. Kattmann, T. Albring
+ * \version 7.0.0 "Blackbird"
+ */
+class CSTLFileWriter final : public CFileWriter{
+private:
+
+  /*--- Variables for reprocessing element connectivity. ---*/
+
+  std::set<unsigned long> halo_nodes; /*!< \brief Set of global renumbered node numbers which the current rank refers to, but does not own. Happens on process borders. Is that mutual? */
```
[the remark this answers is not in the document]
TobiKattmann: Replaced the set by a vector plus a sort a bit below which is equivalent in this case

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+#include "../../../include/output/filewriter/CSTLFileWriter.hpp"
+#include "../../../include/output/filewriter/CParallelDataSorter.hpp"
+#include <iomanip> /*--- Used for ofstream.precision(int). ---*/
+
+
+const string CSTLFileWriter::fileExt = ".stl";
+
+
+CSTLFileWriter::CSTLFileWriter(vector<string> fields,
```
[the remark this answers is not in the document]
TobiKattmann: Done

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+
+  /* Sorted list of halo nodes for this MPI rank. */
+  sorted_halo_nodes.assign(halo_nodes.begin(), halo_nodes.end());
+
+  /*--- We effectively tack the halo nodes onto the end of the node list for this partition.
+    TecIO will later replace them with references to nodes in neighboring partitions. */
+  num_halo_nodes = sorted_halo_nodes.size();
+  vector<long> neighbor_partitions(max(static_cast<unsigned long>(1), num_halo_nodes));
+  vector<long>      neighbor_nodes(max(static_cast<unsigned long>(1), num_halo_nodes));
```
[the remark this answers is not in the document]
TobiKattmann: Done

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: not necessary any more

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: this code bit also does not exist anymore

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: Made a seperate function `StoreCoordData(enum GEO_TYPE elemType,...)` and used the range-for-loop.

I adopted `thisVariableStyle` for everything except the copied TecIO code, which I prefer to keep close to that implementation for comparability.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: Already Done

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,332 @@
...
+      /*--- Quads split into two Triangles. Assumes either clock- or counterclockwise rotation ---*/
+      nodeList.insert(nodeList.end(), {0,1,3, 1,2,3} ); break;
+    default:
+      SU2_MPI::Error("Unsupported element type for the STL writer." , CURRENT_FUNCTION); break;
+  }
+
+  /*--- Write Triangle-Point-Coordinates subsequently in a local send buffer. ---*/
+  for (iElem = 0; iElem < nLocalElements; iElem++) {
+    for (auto iPoint : nodeList) {
+      globalNodeNumber = dataSorter->GetElem_Connectivity(elemType, iElem, nodeList[iPoint]) - 1;
```
pcarruscag: By doing iPoint : nodeList, iPoint is the contents of nodeList, i.e. instead of nodeList[iPoint] you just use iPoint.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,332 @@
...
+  /* The terminology gets a bit confusing here. We're sending the node numbers
+     (sorted_halo_nodes) whose data we need to receive, and receiving
+     lists of nodes whose data we need to send. */
+  if (sorted_halo_nodes.empty()) sorted_halo_nodes.resize(1); /* Avoid crash. */
+  SU2_MPI::Alltoallv(&sorted_halo_nodes[0], &num_nodes_to_receive[0], &nodes_to_receive_displacements[0], MPI_UNSIGNED_LONG,
+                     &nodes_to_send[0],     &num_nodes_to_send[0],    &nodes_to_send_displacements[0],    MPI_UNSIGNED_LONG,
+                     MPI_COMM_WORLD);
+
+  /* Now actually send and receive the data */
+  data_to_send.resize(max(1, total_num_nodes_to_send * (int)fieldnames.size()));
```
pcarruscag: You missed a max<> here.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+
+const string CSTLFileWriter::fileExt = ".stl";
+
+
+CSTLFileWriter::CSTLFileWriter(const vector<string>& fields,
+                               unsigned short nDim,
+                               string fileName,
+                               CParallelDataSorter *dataSorter) :
+                CFileWriter(std::move(fields),
```
pcarruscag: Don't move references.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
pcarruscag: This is probably why the test failed.

TobiKattmann: yep, fixed

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: I needed a `max<unsigned long>` or `1ul` in the other instances because the compiler threw an error because the  naked `1` seems to be interpreted as an <int>. Here this is `max<int>`, so the compiler does not need it, but I added that for consistency

pcarruscag: Because you have an old style cast to int in fieldnames.size(), what I was suggesting was to remove that cast in favor of promoting 1 to unsigned long (either with 1ul or the max<>) to be consistent with the other parts.

TobiKattmann: True, it is now a max<unsigned long> as well and the old style cast is removed of course.

SU2_CFD/src/output/filewriter/CSTLFileWriter.cpp
TobiKattmann: removed the std::move.

TobiKattmann:
Thanks a lot for the reviews and comments @WallyMaier @pcarruscag @economon @talbring !
I highly appreciate you taking the time of working through the code to make it better (and hopefully these improvements sticks to my brain 💡 )

If no comments are left unaddressed (I think so) I would be happy to get this merged 🐘

SU2_CFD/src/output/COutput.cpp
```diff
@@ -552,15 +550,34 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
                                                 curTimeIter, GetHistoryFieldValue("TIME_STEP"));
 
       break;
+      
+    case STL:
+
+     if (fileName.empty())
+       fileName = config->GetFilename(surfaceFilename, "", curTimeIter);
+
+       /*--- Load and sort the output data and connectivity. ---*/
```
pcarruscag: Looks like this is not aligned as intended.

config_template.cfg
```diff
@@ -1256,7 +1256,7 @@ FFD_DEGREE= (10, 10, 1)
 FFD_CONTINUITY= 2ND_DERIVATIVE
 %
 % Definition of the FFD planes to be frozen in the FFD (x,y,z).
-% Value from 0 FFD degree in that direction. Pick a value larger than degree if you don't want to fix any plane.
+% Value from 0 FFD degree in that direction. Pick a value larger than degree if you dont want to fix any plane.
```
pcarruscag: The apostrophe was correct.

talbring: Seems like he has been a little bit overeager :smile:

pcarruscag: Gotta save space for :house:

SU2_CFD/src/output/COutput.cpp
TobiKattmann: Yep, fixed that

config_template.cfg
TobiKattmann: got me 👮 ...  I just was annoyed that for my IDE (vscode) this tick `'`opens a block which is never closed.. so the coloring broke for the rest of the file

TobiKattmann:
Thanks @talbring for fixing the broken parallel output for the DG-FEM solver.

SU2_CFD/include/output/filewriter/CSurfaceFEMDataSorter.hpp
```diff
@@ -74,7 +75,34 @@ class CSurfaceFEMDataSorter final: public CParallelDataSorter{
...
   }
+  
+  /*!
+    * \brief Get the beginning global renumbered node ID of the linear partition owned by a specific processor.
+    * \param[in] rank - the processor rank.
+    * \return The beginning global renumbered node ID.
+    */
+   unsigned long GetNodeBegin(unsigned short rank) const override {
+     unsigned long offsetSurfaceDOFs = 0;
+     for(int i=0; i<rank; ++i) offsetSurfaceDOFs += nSurfaceDOFsRanks[i];
```
TobiKattmann: Make this unsigned short? Or does the compiler optimize this anyway?
btw. unsigned short is max= 65,535 ... that would be quite a lot of processors but probably not too far away from possible for some people out there (idk)?

pcarruscag: Unsigned shorts are nonsense... Unless the purpose is to reduce storage and/or vectorize calculations.

TobiKattmann: `grep -r "unsigned short" SU2_CFD/ Common/ | wc -l` ->10564
It is used all over the place in SU2. 
I found Stroustroup, Sutter & Carruth [here (~2min)](https://www.youtube.com/watch?v=Puio5dly9N8&t=2560) with their bits on the topic. I guess these are the reasons you mean as well

pcarruscag: They take longer to write too xD.
We should define su2int as either int or long and do a replace all :smiling_imp:

PR #848, opened 2020-01-10
Restructuring dual/primal grid structure

talbring:
## Proposed Changes
This is further step in restructuring the files. In this PR I splitted the primal_grid_structure and dual_grid_structure. Further I moved the mesh readers in to a separate directory.  

## Related Work

## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -29,10 +29,20 @@
...
-#include "../../include/CSU2ASCIIMeshReaderFVM.hpp"
-#include "../../include/CCGNSMeshReaderFVM.hpp"
-#include "../../include/CRectangularMeshReaderFVM.hpp"
-#include "../../include/CBoxMeshReaderFVM.hpp"
+#include "../../include/geometry/meshreader/CSU2ASCIIMeshReaderFVM.hpp"
+#include "../../include/geometry/meshreader/CCGNSMeshReaderFVM.hpp"
+#include "../../include/geometry/meshreader/CRectangularMeshReaderFVM.hpp"
+#include "../../include/geometry/meshreader/CBoxMeshReaderFVM.hpp"
+
+#include "../../include/geometry/primal_grid/CPrimalGrid.hpp"
```
economon: Instead of primal/dual, another option is to divide up the files in the subfolders would be between FVM and FEM. I think either would work. Also, there is some subset of files needed by the FEM-based structural solver that is not really reflected in the naming here (we also have the files in elements/ which are FEA primal elements, for instance), which could be confusing.

Common/src/grid_adaptation_structure.cpp
```diff
@@ -26,6 +26,15 @@
...
+#include "../include/geometry/primal_grid/CTriangle.hpp"
+#include "../include/geometry/primal_grid/CQuadrilateral.hpp"
+#include "../include/geometry/primal_grid/CTetrahedron.hpp"
+#include "../include/geometry/primal_grid/CHexahedron.hpp"
+#include "../include/geometry/primal_grid/CPyramid.hpp"
+#include "../include/geometry/primal_grid/CPrism.hpp"
+#include "../include/geometry/primal_grid/CVertexMPI.hpp"
 #include <math.h>
 
 CGridAdaptation::CGridAdaptation(CGeometry *geometry, CConfig *config) {
```
economon: Wondering when we should deprecate the grid adaptation code...

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: Those FEA elements are used more like numerics classes.

talbring: I also thought about dividing it up between FVM and FEM. However, I would say that at the moment they are not independent enough (e.g. CPrimalGridFEM is derived from CPrimalGrid). In order to change this, a restructuring of the CGeometry classes is necessary. Moving these files around should not lead to a lot of problems since they are rarely changed.

talbring: To phrase it differently: the structure proposed here reflects how it is implemented right now.

Common/include/geometry/dual_grid/CEdge.hpp
```diff
@@ -0,0 +1,178 @@
...
+  inline void GetNormal(su2double *val_normal) const override {
+    for (unsigned short iDim = 0; iDim < nDim; iDim++)
+      val_normal[iDim] = Normal[iDim];
+  }
+
+  /*!
+   * \brief Get the normal to a face of the control volume asociated with an edge.
+   * \return Dimensional normal vector, the modulus is the area of the face.
+   */
+  inline su2double *GetNormal(void) const override {  return Normal; }
```
pcarruscag: This is unusual, if the method is const it should not return a pointer that allows member data to be modified.

Common/include/geometry/primal_grid/CPrimalGridFEM.hpp
```diff
@@ -0,0 +1,264 @@
...
+  /*!
+  * \brief Function to make available the number of DOFs of the grid in the element.
+  * \return The number of DOFs of the grid in the element.
+  */
+  inline unsigned short GetNDOFsGrid(void) override { return nDOFsGrid; }
+
+  /*!
+  * \brief Function to make available the number of DOFs of the solution in the element.
+  * \return The number of DOFs of the solution in the element.
+  */
```
pcarruscag: The comments in this file seem to be missing one space.

PR #849, opened 2020-01-11
Restructuring of solver files

jayantmukho:
## Proposed Changes
This is a restructuring of the solver classes in the fashion of what has been done for variable classes, output classes etc. Changes include:

- Moving all solver .hpp and .cpp files into individual class specific files
- Moving all inline functions into the header files 
- Some bits of code cleanup

Things still left to be done: 

- Updating the IDE files
- Any other things that might come up in the review process

This has mostly just been copying and pasting things around. If there any thing else to address in the solver classes (menial or significant) that would be appropriate to include in this restructuring, let me know in the comments!

## Related Work
@pcarruscag is already working on the CFEASolver in #843. I have split the files here as well since it was already automated. But I can just merge with his branch in the likely case that his PR is ready for merging before this one. 

## PR Checklist


SU2_DOT/src/meson.build
```diff
@@ -33,27 +34,28 @@ endif
...
+				                                      'output/filewriter/CCSVFileWriter.cpp',
+				                                      'output/filewriter/CTecplotFileWriter.cpp',
+				                                      'output/filewriter/CTecplotBinaryFileWriter.cpp',
+				                                      'output/filewriter/CParaviewFileWriter.cpp',
+				                                      'output/filewriter/CParaviewBinaryFileWriter.cpp',
+				                                      'output/filewriter/CSU2FileWriter.cpp',
+				                                      'output/filewriter/CSU2BinaryFileWriter.cpp',
+				                                      'output/filewriter/CSU2MeshFileWriter.cpp',
+				                                      'variables/CBaselineVariable.cpp',
+				                                      'variables/CVariable.cpp'])
```
pcarruscag: indentation

SU2_CFD/include/SU2_CFD.hpp
```diff
@@ -38,7 +38,8 @@
 #include "drivers/CDiscAdjSinglezoneDriver.hpp"
 #include "drivers/CDiscAdjMultizoneDriver.hpp"
 #include "drivers/CDummyDriver.hpp"
-#include "solver_structure.hpp"
+// #include "solver_structure.hpp"
```
pcarruscag: don't forget to delete the old include

jayantmukho:
> Nice work @jayantmukho, I hope that script helped with moving the inlines.

Yes! It did. I was able to build off the script you gave me to enable moving multi-line inlines. Now it only fails when the function signatures don't match up (because of different variable names). I've attached it here:
[split_inlines.txt](https://github.com/su2code/SU2/files/4054552/split_inlines.txt)
[merge_inlines.txt](https://github.com/su2code/SU2/files/4054551/merge_inlines.txt)

> General question, did you update the `#include "solver_structure.hpp"` automatically or manually (fixing the compilation errors one by one)? I ask because this kind of file restructuring gives us a good opportunity to make sure we are including files only where needed, which helps with build times.

I replaced the solver_structure.hpp files with just CSolver.hpp first. Then quashed errors by specific solver includes when needed. But its a good point. I will go through it again to remove any unnecessary includes that might have crept in. 

 > Apart from the two comments below, I would add that it would be nice to put virtual (`override` or `final`) and `const` specifiers where appropriate, at some point, not necessarily now.

Ah yes, this is a good time to do this. Will do.

SU2_DOT/src/meson.build
jayantmukho: addressed

SU2_CFD/include/SU2_CFD.hpp
jayantmukho: addressed

talbring:
@jayantmukho Thanks for that restructuring. When you are finished use the two commands in the new folders from the script `externals/utils/replace-tabs-and-trim-trailing-whitespaces.sh` to replace tabs and to remove whitespaces.

 This is probably the part where most of the people will encounter conflicts and problems. I can only encourage everybody to merge with develop ASAP. It won't get any easier.

pcarruscag:
> This is probably the part where most of the people will encounter conflicts and problems. I can only encourage everybody to merge with develop ASAP. It won't get any easier.

And are we going to merge open PR's in chronological order? Asking for a friend.

talbring:
> > This is probably the part where most of the people will encounter conflicts and problems. I can only encourage everybody to merge with develop ASAP. It won't get any easier.
> 
> And are we going to merge open PR's in chronological order? Asking for a friend.

Hm as you know we don't really follow a chronological order for merging ...

pcarruscag:
I know, but maybe we could try doing it as a way to create incentive for
all PR to get reviewed. I.e. out of two PR declared ready by the authors
(comments addressed, etc.) we would give priority to the oldest. Draft and
PR with significant changes requested would not count and therefore not
block newer ones.
(This discussion is not related to this PR, it is just an idea...)

jayantmukho:
> Apart from the two comments below, I would add that it would be nice to put virtual (`override` or `final`) and `const` specifiers where appropriate, at some point, not necessarily now.

I am in the process of doing this. There are a large number of `virtual` functions in the solvers. 

I wanted to verify how I am thinking about using `final`, `override`, and `const`:

1) Any virtual class that is overridden by a child class should have the `override` specifier
2) If the function is not overriden further in another child class, should have the `final` specifier. 
3) Any functions that don't change any variables should be declared `const` (mostly getter functions)
4) Any function arguments that are passed by reference and should not be changed during the execution of the function, should be declared `const`

Is there any case I am missing? Is there a reason why I don't see any functions with the `override final` specification?

pcarruscag:
Override and final are not compatible, one or the other.
The class itself can also be declared final but that is not fundamental.
With const the philosophy is usually "const by default" not just the gets (this is hard to achieve with the size of our classes) what you wrote makes sense.
You can compile with -Wsuggest-override to facilitate the task.
I recommend doing that before changing the signatures of the functions, otherwise some functions may not override, there is a warning for that but I don't remember it...

SU2_CFD/include/solvers/CAdjTurbSolver.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+  su2double Gamma;              /*!< \brief Fluid's Gamma constant (ratio of specific heats). */
+  su2double Gamma_Minus_One;    /*!< \brief Fluids's Gamma - 1.0  . */
+
+  CAdjTurbVariable* nodes = nullptr;  /*!< \brief The highest level in the variable hierarchy this solver can safely use. */
+
+  /*!
+   * \brief Return nodes to allow CSolver::base_nodes to be set.
+   */
+  inline CVariable* GetBaseClassPointerToNodes() override { return nodes; }
```
pcarruscag: Hey @jayantmukho the specifiers are looking good, a quick comment for cases like these where "all" methods of a class seem final.
What I did with the CVariable family was to declare the class itself final e.g. `class CAdjTurbSolver final : public CSolver` and then the methods override as usual, for two reasons:
- From the optimization standpoint (final gives the compiler license to de-virtualize) the methods of a final class are final by definition.
- From the maintenance perspective if a derived class is added in the future we only need to change one line of code (that was a more important aspect for CVariable with its 100's of methods).

I would reserve final for methods in the middle of the hierarchy as I think that sends the stronger message "this method stops here, no need to navigate further". This being said I know you have been doing your own reading, so if this is advice from some C++ guru with many stripes, forget what I said xD.

jayantmukho: @pcarruscag you are my C++ guru. That makes sense. 

I am basically done with all the virtual specs. Going to make this change now.

pcarruscag: Ahah that genuinely made me laugh, thanks!

SU2_CFD/src/integration_structure.cpp
```diff
@@ -153,9 +153,9 @@ void CIntegration::Space_Integration(CGeometry *geometry,
       case DIELEC_BOUNDARY:
         solver_container[MainSolver]->BC_Dielec(geometry, solver_container, numerics[CONV_BOUND_TERM], config, iMarker);
         break;
-      case NEUMANN:
-        solver_container[MainSolver]->BC_Neumann(geometry, solver_container, numerics[CONV_BOUND_TERM], config, iMarker);
-        break;
+      // case NEUMANN: TODO
+      //   solver_container[MainSolver]->BC_Neumann(geometry, solver_container, numerics[CONV_BOUND_TERM], config, iMarker);
+      //   break;
```
jayantmukho: This is the only call to BC_Neumann

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -0,0 +1,4995 @@
...
+  /*!
+   * \brief Move the mesh in time
+   */
+  inline virtual void SetDualTime_Mesh(void){ }
+
+  /*!
+   * \brief Store the BGS solution in the previous subiteration in the corresponding vector.
+   */
+  void UpdateSolution_BGS(CGeometry *geometry, CConfig *config);
+
```
jayantmukho: Is there a reason why this is not declared virtual? There are instantiations of it in `CDiscAdjFEASolver` and in CDiscAdjSolver`

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -0,0 +1,4995 @@
...
+  /*! TODO
+   * \brief A virtual member. 
+   * \param[in] geometry - Geometrical definition of the problem. 
+   * \param[in] solver_container - Container vector with all the solutions. 
+   * \param[in] numerics - Description of the numerical method. 
+   * \param[in] config - Definition of the particular problem.  
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.  
+   */ 
+  // inline virtual void BC_Neumann(CGeometry *geometry,
+  //                                CSolver **solver_container,
```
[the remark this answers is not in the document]
pcarruscag: Maybe it is from the time when there was a Poisson solver?

SU2_CFD/include/solvers/CSolver.hpp
pcarruscag: Hmmm curious... I 🤦 myself for a minute thinking this was the explanation to something important, turns out it is not very relevant.
Please delete the versions of the method in those two DiscAdj solvers, they only affect the legacy adjoint FSI driver (namely the computation of multizone residuals) and that is on its way out so...
You may be able to catch more of these with warning `-Woverloaded-virtual`

SU2_CFD/include/solvers/CSolver.hpp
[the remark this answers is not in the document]
koodlyakshay: It is. The BC_Dirichlet was also from the Poisson solver, not sure if it is used elsewhere.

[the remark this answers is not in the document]
jayantmukho: @koodlyakshay you are right, the BC_Dirichlet is just an empty function in CEulerSolver. 

I'll remove both. Thanks for the input!

jayantmukho:
Thanks for taking care of the merge with develop @pcarruscag, I am going to make some simple  formatting changes to the `CFEASolver` and `CMeshSolver` so that they conform to the rest of the solver files

SU2_CFD/include/solvers/CFEASolver.hpp
```diff
@@ -329,8 +371,10 @@ class CFEASolver : public CSolver {
    * \param[in] config - Definition of the particular problem.
    * \param[in] val_marker - Surface marker where the boundary condition is applied.
    */
-  inline void BC_Normal_Displacement(CGeometry *geometry, CNumerics *numerics,
-                                     CConfig *config, unsigned short val_marker) final { }
+  inline void BC_Normal_Displacement(CGeometry *geometry,
+                                     CNumerics *numerics,
+                                     CConfig *config,
+                                     unsigned short val_marker) final { }
```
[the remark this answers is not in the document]
pcarruscag: I do not

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -57,11 +57,11 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
                || (donor_config->GetKind_Solver() == DISC_ADJ_NAVIER_STOKES)
                || (donor_config->GetKind_Solver() == DISC_ADJ_RANS));
-  bool incompressible_flow = ((donor_config->GetKind_Solver() == INC_NAVIER_STOKES)
+  bool incompressible_flow = (((donor_config->GetKind_Solver() == INC_NAVIER_STOKES)
                || (donor_config->GetKind_Solver() == INC_RANS)
                || (donor_config->GetKind_Solver() == DISC_ADJ_INC_NAVIER_STOKES)
-               || (donor_config->GetKind_Solver() == DISC_ADJ_INC_RANS)
-               && (donor_config->GetEnergy_Equation()));
+               || (donor_config->GetKind_Solver() == DISC_ADJ_INC_RANS))
+               && donor_config->GetEnergy_Equation());
```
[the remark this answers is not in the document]
pcarruscag: Don't suppress that warning until someone says something about it, I left a post merge comment in #839, the logic looks fishy.

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -30,7 +30,26 @@
...
+#include "../solvers/CAdjEulerSolver.hpp"
+#include "../solvers/CAdjNSSolver.hpp"
+#include "../solvers/CAdjTurbSolver.hpp"
+#include "../solvers/CHeatSolverFVM.hpp"
+#include "../solvers/CFEASolver.hpp"
+#include "../solvers/CTemplateSolver.hpp"
+#include "../solvers/CDiscAdjSolver.hpp"
+#include "../solvers/CDiscAdjFEASolver.hpp"
+#include "../solvers/CFEM_DG_EulerSolver.hpp"
+#include "../solvers/CFEM_DG_NSSolver.hpp"
```
pcarruscag: These should go to the cpp otherwise a change to any solver will trigger recompilation of all drivers.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -0,0 +1,4837 @@
...
+  virtual ~CSolver(void);
+
+  /*!
+   * \brief Allow outside access to the nodes of the solver, containing conservatives, primitives, etc.
+   * \return Nodes of the solver.
+   */
+  inline CVariable* GetNodes() const {
+    assert(base_nodes!=nullptr && "CSolver::base_nodes was not set properly, see brief for CSolver::SetBaseClassPointerToNodes()");
+    return base_nodes;
+  }
```
pcarruscag: Methods returning non-const pointers to class members should not be declared const IMO. While the method itself does not change the state of the object, it still provides a way to (changing the nodes in this case), this would allow a function taking a const CSolver* to change its via such methods which would be misleading.
So, until we can return const CVariable*, this should not be const.

SU2_CFD/include/drivers/CDriver.hpp
jayantmukho: Oooo, that's a good point. Will move these

SU2_CFD/include/solvers/CSolver.hpp
jayantmukho: Fair enough, I agree.

Issue #698, opened 2019-06-04
Add Unit Testing to SU2


clarkpede:
I propose adding a unit-testing framework and unit-tests to SU2. After chatting with @economon, I've decided to move the discussion here to get additional input.

## What is unit testing?

For those not familiar with unit testing, unit testing allows the testing of small bits of behavior, ideally using isolated bits of code.  It is not intended to replace validation testing or formal verification tests.  Instead, it serves a unique purpose.  Consider the three following use cases:

+ You're developing a new feature, and you want to test it to see if it works.  You could do a full simulation, but that takes a lot of time and computing power.  You want to check if your new behavior behaves as you suspect before you throw a lot of resources at it.
+ You submit a PR and discover that one of the regression tests has failed.  But...why? You know that something is broken, but its hard to track down what broke.  You want more granular test coverage that can demonstrate what broke.
+ You are fixing a very small bug.  You know that you should prove that your bug fix worked, but it doesn't seem logical to dedicate an entire validation case to one small bug fix.  You want to write a small test for a small fix.

In all of these cases, unit testing fills a unique role.  Unit testing increases time spent in development, but decreases the amount of time spent in bug-fixing and maintaining.

For more information, see this [relevant Stack Exchange question.](https://scicomp.stackexchange.com/questions/206/is-it-worthwhile-to-write-unit-tests-for-scientific-research-codes)

## What do I propose?

Our research group at UT Austin has implemented a unit testing framework on our branch, which we're happy with.  Some choices were arbitrary, and some choices were made based on our development environment.  Those choices may be different for other groups.  Here's what we have done:

The unit testing framework is compiled and run using autotools.  For more information on autotool's setup, see [their documentation](https://www.gnu.org/software/automake/manual/html_node/Tests.html). Since autotools is the build system for SU2, this involves minimal changes.

Using automake, the build process for building unit tests becomes:

```
    ./bootstrap
    ./configure
    make
    make check
```

We use [Boost's unit testing framework](https://www.boost.org/doc/libs/1_70_0/libs/test/doc/html/index.html).   This provides a convenient set of macros for instatiating tests, grouping tests into suites, and running checks.  This choice was based on what is available in our development setup.

We have integrated our unit tests into our Travis CI regression testing.  Every time we push commits or submit a pull request, the unit tests are run and checked.

## What is my vision for unit testing in SU2?

I am *not* proposing that we start trying to get 100%  code coverage with pre-existing code.  That would not provide a good return on investment.

[...]

PR #850, opened 2020-01-17
Add unit-testing framework to SU2

clarkpede:
## Proposed Changes

This is intended to be a small PR, but a big design shift.  As discussed in issue #698, unit tests can serve an important role in software development.  They complement, rather than replace, validation and verification tests.  They are most useful when you want to isolate a specific behavior. Unit-testing increases development time, but improves the quality of the code and reduces code-maintenance time \[[source](https://collaboration.csc.ncsu.edu/laurie/Papers/TDDpaperv8.pdf)\].

### Vision

In my mind, the most efficient way to proceed is to add unit tests as new behaviors are added and bugs are fixed.  With each new PR, we should ask "Should this have a unit test to verify this behavior?" Unit tests will provide a clear way to show that each PR does what the developer promises it will do.

This will require refactoring some of the code.  Some parts of the code, like the CConfig class, are not unit-testing friendly. Unit testing is hard when you have God-classes and "do-it-all" functions.  The hope is that refactoring these classes will lead to more modular and easier-to-maintain code.

### Decision to use Catch2

Catch2 was chosen for the unit testing library for the following reasons:

+ It can be included as a header-only library.
+ It has a very clean, easy-to-use syntax
+ It has widespread use, including the FEniCS library.

### Design choices

The following design choices are in place.  If you disagree with them, then please start a discussion here:

+ All tests are placed in a root-level directory named `UnitTests`.  Inside this directory is a structure just like the existing folder structure (e.g. `SU2_CFD/numerics`), except that there are not separate `src` and `include` folders.
+ Tests are grouped by class, and put in files such as `CNumerics_tests.cpp`.
+ A single test executable is compiled and run, as opposed to a separate test executable for each group of tests.  With the single test executable, you can always filter down to groups of tests or individual tests.
+ When `ninja test` or `meson test` is run, the test executable is only run once, sweeping through all the unit tests.  The result only shows a single failure or a single success.  If more detail is desired, then the test executable can be run manually.
+ The relevant SU2 code is compiled as a library (e.g. `libSU2core`) and then linked into both the main executable (`SU2_CFD`) and the test program (`test_driver`).

### How to Run Unit Tests

If you compile SU2 using `ninja -C builddir`, then the test executable will automatically be compiled. You then have three options for running the tests:

+ `meson test` (see documentation [here](https://mesonbuild.com/Unit-tests.html))
+ `ninja test`
+ Run the exectuable manually: `./UnitTests/test_driver`

When running the executable manually, there are many customization options.  Run `./UnitTests/test_driver --help` to see them all.

### How to add unit tests

I have already included a single unit-test example (`UnitTests/SU2_CFD/src/numerics/CNumerics_tests.cpp`).  Aside from the headaches with CConfig, this example should be clear.  If not, raise questions and I will answer them.  ~~I plan on adding a few more example tests, as well as a "How-To" page on the website.~~ Please see the PR on the website repo for more documentation.

To learn more, check out the [great tutorial](https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md) for Catch2.  You can also find more in-depth documentation there.

### I still don't understand unit tests.  Where can I learn more?

I highly recommend reading at least one of these two books:

+ "Working Effectively with Legacy Code," by Michael Feathers
+ "Modern C++ Programming with Test-Driven Development," by Jeff Langr

## Remaining work

+ ~~Add unit-testing to Travis CI~~
+ ~~Add support for unit testing in AD builds~~
+ ~~Add more examples of unit tests~~
+ ~~Add "How-To Add Unit Tests" documentation~~

## Related Work
This resolves issue #698.

## PR Checklist


talbring:
Thanks @clarkpede for setting this up! I will add some tests by myself and see how its gonna work out.

UnitTests/test_driver.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+  /*--- Startup MPI, if supported ---*/
+#ifdef HAVE_MPI
+  int  buffsize;
+  char *buffptr;
+#ifdef HAVE_OMP
+  int provided;
+  SU2_MPI::Init_thread(&argc, &argv, MPI_THREAD_FUNNELED, &provided);
+#else
+  SU2_MPI::Init(&argc, &argv);
```
economon: Looks like the driver sets up MPI.. it would be great if we can also use MPI in the tests. Is that working?

clarkpede: I haven't tested it myself, but that's the intention.

A lot of classes (like the solver classes) use MPI calls.  For example, some constructors set the MPI rank and size. So there's really no getting around MPI, even for non-MPI tests.

talbring:
I added unit testing to github actions. In order not to recompile the code, but still have the unit testing in a separate step, I had to add to install the test driver. To not have it installed by default, there is the flag `-Denable-tests` that has to be set to true for meson in order to enable the tests.  The unit testing step directly calls the test driver executable.

clarkpede:
I've added support for tests using algorithmic differentiation and direct differentiation.  In order to make sure the right libraries are linked, I separated out the test drivers.  Now there's three test drivers:

+ A forward test driver (`test_driver`)
+ A test driver for direct differentiation tests (`test_driver_DD`)
+ A test driver for AD tests (`test_driver_AD`)

economon:
@clarkpede: looks like you marked this ready to go a while back, still ready for review? Anything else you (or @talbring) want to change?

clarkpede:
@economon No, this looks good to me. It is merge-ready, from my perspective.

I chatted with @talbring, and in a future PR we would like to add a simple set of classes to use with unit tests.  For example, I've created a "one-point geometry" class for use in some of my tests.  But I think that we should keep the PRs as incremental as possible.

PSA: If anyone else wants to review this PR, they are welcome to.  It is no longer a WIP.

EduardoMolina:
Hi @clarkpede ,

Thanks for the contribution. The example of the central/upwind blending is very clear. Ready to merge.

Eduardo

clarkpede:
@economon I'm documenting my thoughts on mocking here, partially in response to you but also for any collaborators. The biggest hurdle with mocking frameworks is that they usually require versatile abstract base classes and depedency inversion.  SU2 is relatively good at that with solvers, variables, and numerics.  But classes like CConfig are completely incompatible with mocking.  CGeometry seems to be somewhere in the middle. Mocking can also be [overused](https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html).

I agree, mocking is a great tool, and has its legitimate uses. [Martin Fowler recommends mocking](https://www.martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs) in at least two use cases:
> One is if you're spending a lot of time debugging when tests fail because they aren't breaking cleanly and telling you where the problem is...The second area is if your objects don't contain enough behavior, mockist testing may encourage the development team to create more behavior rich objects.
But I believe we can get pretty far just with what we have, and start including mocking as needed.

PR #858, opened 2020-02-04
Fix error during dry-run and hinder overwrite of history files

### LINKED ISSUE #857: Dry run errors, i.e. SU2_CFD -d <configname>.cfg
**Describe the bug**
When running any dry run in the form
`SU2_CFD -d <configname>.cfg`
an integer divide by zero (`Floating point exception: integer divide by zero`) happens.

A backtrace pointed me towards `roundUpDiv (numerator=0, denominator=0)` in `omp_structure.hpp` +104 and further up `computeStaticChunkSize (totalWork=0, numThreads=1, maxChunkSize=4096)`. I guess one might check somewhere along the line whether there is even work to do, i.e.  wrap the `*.Initialize(nPoint,...)` in `C<Inc>EulerSolver` etc in `if (nPoint != 0)` conditionals (that got the Quickstart working again for me). But maybe there is a better more elegant solution without too much overhead.

I would like @pcarruscag to take a look as the omp stuff is involved :)

**To Reproduce**
Every dryrun I tried, using Quickstart should be enough to understand the problem.

 - OS: Red Hat Enterprise Linux 7.6
 - C++ compiler and version: GCC 4.8.5
 - MPI implementation and version: openMPI 3.1.2
 - SU2 Version: latest develop, but this probably affects 7.0.1 as well, i.e. hotfix time 🕙

TobiKattmann:
## Proposed Changes
Fixes #857, an error during dry-run. I know @pcarruscag added the fixing commit to his PR #824 already and mentioned it below the issue, i.e. all kudos regarding the first commit go to him. I only changed the omp_structure.hpp of ded9c08e90496634a38b4998d9fe7aa5b72cdcc3 as I guess that bit is enough for this PR. 
Additionally I hinder the dry-run to overwrite the history files, which is kind of annoying.

If anyone as an elegant solution to simply check for EXIT_SUCCES in the regression test one could add some dry_run regression tests.

## PR Checklist


clarkpede:
> If anyone as an elegant solution to simply check for EXIT_SUCCES in the regression test one could add some dry_run regression tests.

This can be added to the meson tests simply enough.  In other words, you could add it alongside the unit tests.  You can read more about meson tests [here](https://mesonbuild.com/Unit-tests.html). If you add `SU2_CFD` as a test executable with the dry run option as a command line argument, then meson will do the dry run and mark it as failing if it does not receive `EXIT_SUCCESS`.

Common/include/omp_structure.hpp
```diff
@@ -116,6 +116,7 @@ inline size_t computeStaticChunkSize(size_t totalWork,
                                      size_t numThreads,
                                      size_t maxChunkSize)
 {
+  if(!totalWork) return maxChunkSize;
```
pcarruscag: This fixes the issue if the code is compiled without OpenMP, which I guess for now is ok.
Do you still get the same information as before by not overwriting the history files? If so LGTM

talbring:
@clarkpede We could add use that feature probably. But we have to check that essentially for all available solvers, so maybe it makes more sense to put it to the regression tests.

Common/include/omp_structure.hpp
TobiKattmann: As the other stuff goes in with your PR, I think that should be Ok for most people. 
There was actually a small difference in the `--- Output Information ---` tab which is also shown in a normal run, but that is addressed now as well 👍

Common/src/geometry/CGeometry.cpp
```diff
@@ -56,6 +56,9 @@ CGeometry::CGeometry(void) {
   nPoint     = 0;
   nPointNode = 0;
   nElem      = 0;
+  Global_nPointDomain = 0;
+  Global_nPoint = 0;
+  nPointDomain = 0;
```
TobiKattmann: Do these have s.th. to do with this PR?

talbring: Yes, dry run was not working when MG was enabled because of that

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -3757,14 +3757,18 @@ bool CFluidDriver::Monitor(unsigned long ExtIter) {
     case DISC_ADJ_FEM_EULER: case DISC_ADJ_FEM_NS: case DISC_ADJ_FEM_RANS:
       StopCalc = integration_container[ZONE_0][INST_0][ADJFLOW_SOL]->GetConvergence(); break;
   }
+  
+  /*--- Set StopCalc to true if max. number of iterations has been reached ---*/
+  
+  StopCalc = (ExtIter == Max_Iter - 1);
 
   return StopCalc;
```
TobiKattmann: The upper `StopCalc` is obsolete now as it is overwritten with the code addition anyway.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -3757,14 +3757,18 @@ bool CFluidDriver::Monitor(unsigned long ExtIter) {
...
 
 }
 
 
 void CFluidDriver::Output(unsigned long InnerIter) {
-
+
```
TobiKattmann: ⬜️

TobiKattmann:
Concerning the Regression tests: @talbring added a check for every testcase whether it exits successfully. Like this, `SU2_CFD -d` passes succesfully when it exits with `EXIT_SUCCESS` (i.e. returns 0). That is the only check, the actual output is not checked.
For quite some reg-tests e.g. with parallel_computation.py the SU2_SOL part silently failed which then causes the particular reg-test to fail. "Silently" because the test_values from the successful SU2_CFD match and that was the only criteria until now.

Additionally we found a little bug where steady primal multizone restarts wouldn't work. That is fixed as well.

talbring:
Thanks @TobiKattmann for the summary :+1:

TobiKattmann:
✅  **TobiKattmann** approved these changes 1 day ago

I cant approve this PR because I opened it, but Tim wrote most parts and I have no more comments to these bits.... so all I can do is do a fake-approval in the hope someone falls for it.

PR #853, opened 2020-01-29
Faster compilation

pcarruscag:
## Proposed Changes
Addresses the compilation time issue (#851) in the way discussed with @talbring in #824.
Plus, I think it is time to bite the bullet and cleanup CConfig, no more .inl, tabs replaced (but the comments still look ok-ish), methods made const when possible.
This will probably cause you issues, but the changes to these files are usually very linear and so, easy to re-apply.
There is no more CCreateMap function, in C++11 braced initialization can be used instead e.g.:
`map<string, string> AverageGroupName = {{"BGS_RES", "bgs"},{"RMS_RES","rms"},{"MAX_RES", "max"}};`

## PR Checklist
- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X] My contribution is commented and consistent with SU2 style.
- [X] I have added a test case that demonstrates my contribution, if necessary.
- [X] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #861, opened 2020-02-06
Hybrid Parallel Compressible RANS Solvers

pcarruscag:
## Proposed Changes
Extend the MPI+Threads framework to CEulerSolver, CNSSolver and the CTurb***Solver(s).
I can finally start showing what this effort is about in pictures, this is how the relatively small benchmark case from #716 scales on Imperial's cluster:
image
**C** is for cores, **R** for MPI ranks, and **T** for OpenMP threads per rank.

It may not look like much close to the ideal scaling but on 192 cores there are only 2.7k cells per core, and at different level of "hybridisation" this is what you get:
image

As mentioned in #789 this behaviour is due to the worse quality / quantity of coarse grids that can be generated as the domain is decomposed further and further.

**EDIT**
This is **finished for now**, apart from the odd routine that I will document below (and all the boundary conditions) every method called in `SingleGrid_Iteration` and `MultiGrid_Iteration` (including prolongations, smoothings and whatnot) is hybrid parallel.
Threads are started in those functions which means that for a RANS case we enter parallel regions only twice per iteration, this should give minimal overhead, I will update the scalability results once they converge (statistically :) )

I will document the changes by "themes" below.

## Related Work
Another snapshot of #824

## PR Checklist


Common/include/toolboxes/MMS/CIncTGVSolution.hpp
```diff
@@ -92,5 +92,5 @@ class CIncTGVSolution: public CVerificationSolution {
    */
   void GetBCState(const su2double *val_coords,
                   const su2double val_t,
-                  su2double       *val_solution);
+                  su2double       *val_solution) const override;
```
pcarruscag: As mentioned in #824 I added the `const` and `override/final` specifiers to the MMS classes, `const` methods cannot alter the state of the class and so the "Get's" should be ok to call in parallel.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -58,6 +49,43 @@
...
+#include "../../include/numerics/continuous_adjoint/adj_convection.hpp"
+#include "../../include/numerics/continuous_adjoint/adj_diffusion.hpp"
+#include "../../include/numerics/continuous_adjoint/adj_sources.hpp"
+#include "../../include/numerics/turbulent/turb_convection.hpp"
+#include "../../include/numerics/turbulent/turb_diffusion.hpp"
+#include "../../include/numerics/turbulent/turb_sources.hpp"
+#include "../../include/numerics/elasticity/CFEAElasticity.hpp"
+#include "../../include/numerics/elasticity/CFEALinearElasticity.hpp"
+#include "../../include/numerics/elasticity/CFEANonlinearElasticity.hpp"
+#include "../../include/numerics/elasticity/nonlinear_models.hpp"
```
pcarruscag: I separated the numerics into these files, as mentioned in #824 
>On the subject of restructuring the CNumerics files, I initially did it one-class-per-file with some sub folders, the problem with that is that we have some 85 of those classes... so 170 files which I think is maybe too many, more important is that some of those classes form smaller hierarchies which are not clearly identifiable with so many files.
So I did it by "themes" instead, I think this division should inspire some re-usage when new classes are added, the average cpp size is under 1k lines, which I think is quite manageable.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -0,0 +1,1302 @@
...
+
+    ResidualType(const Vector_t& res, const Matrix_t& jac_i, const Matrix_t& jac_j) :
+      residual(res), jacobian_i(jac_i), jacobian_j(jac_j) { }
+
+    /*!
+     * \brief The object can be directly cast to the vector type, this
+     * allows discarding the Jacobians when they are not needed.
+     */
+    operator Vector_t() { return residual; }
+  };
```
pcarruscag: For thread safety we need one numerics instance per thread, likewise in the call to `CNumerics::ComputeResidual` we would need dedicated copies of the residual and its Jacobians, rather than allocating multiple of those in CSolver I am allocating them in CNumerics (since we already have one per thread) and then returning this lightweight struct that gives read-only access to those members of CNumerics.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -0,0 +1,1302 @@
...
+   */
+  inline virtual void ComputeResidual(su2double *val_residual, su2double **val_Jacobian_i,
+                                      su2double **val_Jacobian_j, CConfig* config) { }
+
+  /*!
+   * \overload For numerics classes that store the residual/flux and Jacobians internally.
+   * \param[in] config - Definition of the particular problem.
+   * \return A lightweight const-view (read-only) of the residual/flux and Jacobians.
+   */
+  inline virtual ResidualType<> ComputeResidual(const CConfig* config) { return ResidualType<>(nullptr,nullptr,nullptr); }
```
pcarruscag: This is the overload for that return type. You will also notice that following #853 I am making the use of CConfig read-only whenever possible in low level classes such as CNumerics.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -164,50 +201,75 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
...
 
-    LinSysRes.AddBlock(iPoint, Residual);
-    LinSysRes.SubtractBlock(jPoint, Residual);
+    LinSysRes.AddBlock(iPoint, residual);
+    LinSysRes.SubtractBlock(jPoint, residual);
 
     /*--- Implicit part ---*/
 
-    Jacobian.UpdateBlocks(iEdge, iPoint, jPoint, Jacobian_i, Jacobian_j);
+    Jacobian.UpdateBlocks(iEdge, iPoint, jPoint, residual.jacobian_i, residual.jacobian_j);
```
pcarruscag: This is how that overload and return type are used.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -0,0 +1,1105 @@
...
+  cb2_sigma = cb2/sigma;
+  cw1 = cb1/k2+(1.0+cb2)/sigma;
+
+  /*--- Setup the Jacobian pointer, we need to return su2double** but
+   *    we know the Jacobian is 1x1 so we use this "trick" to avoid
+   *    having to dynamically allocate. ---*/
+
+  Jacobian_i = &Jacobian_Buffer;
+
+}
```
pcarruscag: I have also done some cleanup, for example by inserting some intermediate classes to at least centralise the definition of hardcoded values.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -788,9 +413,38 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config, unsigned short
...
+    }
+  }
+  ColorGroupSize = geometry->GetEdgeColorGroupSize();
+
+  if ((MGLevel != MESH_0) && (omp_get_max_threads() > 1)) {
+    EdgeFluxes.Initialize(geometry->GetnEdge(), geometry->GetnEdge(), nVar, nullptr);
+  }
+
+  omp_chunk_size = computeStaticChunkSize(nPoint, omp_get_max_threads(), OMP_MAX_SIZE);
+#endif
```
pcarruscag: When in parallel, coarse grids initialise this CSysVector to store edge fluxes, the coloring they obtain from CGeometry is the natural color, i.e. a single color with all edges which is equivalent to `for(iEdge=0; iEdge < nEdge; ...)`

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3426,238 +3122,226 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
 
-    /*--- Set the final value of the Roe dissipation coefficient ---*/
-
-    if (kind_dissipation != NO_ROELOWDISS){
-      nodes->SetRoe_Dissipation(iPoint,numerics->GetDissipation());
-      nodes->SetRoe_Dissipation(jPoint,numerics->GetDissipation());
+      /*--- Set implicit computation ---*/
+      if (implicit)
+        Jacobian.UpdateBlocks(iEdge, iPoint, jPoint, residual.jacobian_i, residual.jacobian_j);
     }
```
pcarruscag: Then, when looping through the edges, we only write to locations that are unique to the edge, i.e. we do not write to `LinSysRes` nor to the diagonal of `Jacobian`, notice that the call to `Jacobian.UpdateBlocks` is different.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3426,238 +3122,226 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+    Viscous_Residual(iEdge, geometry, solver_container,
+                     numerics_container[VISC_TERM + omp_get_thread_num()*MAX_TERMS], config);
   }
+  } // end color loop
+
+  if (reducer_strategy) {
+    SumEdgeFluxes(geometry);
+    if (implicit)
+      Jacobian.SetDiagonalAsColumnSum();
+  }
```
pcarruscag: Having finished the edge loop we do a loop over points, in `SumEdgeFluxes` where we gather the computed fluxes for each cell.
Similarly we set each diagonal entry of the system matrix as the negative of the column sum (sum of blocks with the same column index).
This strategy is not as efficient as coloring because, as explained in #789, it involves an extra pass over the entire matrix with a non contiguous access pattern. However this strategy gives as much parallelism as can be hoped for, and the lower efficiency is not a big concern on coarse grids (which are smaller).

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: When in parallel, coarse grids initialise this CSysVector to store edge fluxes, the coloring they obtain from CGeometry is the natural color, i.e. a single color with all edges which is equivalent to `for(iEdge=0; iEdge < nEdge; ...)`

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Then, when looping through the edges, we only write to locations that are unique to the edge, i.e. we do not write to `LinSysRes` nor to the diagonal of `Jacobian`, notice that the call to `Jacobian.UpdateBlocks` is different.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Having finished the edge loop we do a loop over points, in SumEdgeFluxes where we gather the computed fluxes for each cell.
Similarly we set each diagonal entry of the system matrix as the negative of the column sum (sum of blocks with the same column index).
This strategy is not as efficient as coloring because, as explained in #789, it involves an extra pass over the entire matrix with a non contiguous access pattern. However this strategy gives as much parallelism as can be hoped for, and the lower efficiency is not a big concern on coarse grids (which are smaller).

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -38,304 +39,276 @@
...
+  vector<EdgeColor> EdgeColoring;   /*!< \brief Edge colors. */
+  unsigned long ColorGroupSize;     /*!< \brief Group size used for coloring, chunk size in edge loops must be a multiple of this. */
+
+  /*--- Edge fluxes, for OpenMP parallelization on coarse grids. As it is difficult to
+   * color them, we first store the fluxes and then compute the sum for each cell.
+   * This strategy is thread-safe but lower performance than writting to both end
+   * points of each edge, so we only use it when necessary, i.e. coarse grids and
+   * with more than one thread per MPI rank. ---*/
+
+  CSysVector<su2double> EdgeFluxes; /*!< \brief Flux across each edge. */
```
pcarruscag: This is also explained in the code.

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -38,304 +39,276 @@
...
+  AeroCoeffsArray InvCoeff;          /*!< \brief Inviscid pressure contributions for each boundary. */
+  AeroCoeffsArray SurfaceInvCoeff;   /*!< \brief Inviscid pressure contributions for each monitoring boundary. */
+  AeroCoeffs AllBoundInvCoeff;       /*!< \brief Total pressure contribution for all the boundaries. */
+
+  AeroCoeffsArray MntCoeff;          /*!< \brief Inviscid momentum contributions for each boundary. */
+  AeroCoeffsArray SurfaceMntCoeff;   /*!< \brief Inviscid momentum contributions for each monitoring boundary. */
+  AeroCoeffs AllBoundMntCoeff;       /*!< \brief Total momentum contribution for all the boundaries. */
+
+  AeroCoeffsArray SurfaceCoeff;      /*!< \brief Totals for each monitoring surface. */
+  AeroCoeffs TotalCoeff;             /*!< \brief Totals for all boundaries. */
```
pcarruscag: ...
defining auxiliary types to store aerodynamic coefficients that were "repeated" for inviscid, momentum, viscous, total, etc. contributions. The auxiliary types manage allocation / destruction and common operations like setting everything to zero.

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -38,304 +39,276 @@
...
-  Total_MaxHeat,  /*!< \brief Maximum heat flux on all boundaries. */
-  Total_AeroCD,   /*!< \brief Total aero drag coefficient for all the boundaries. */
-  Total_CEquivArea,   /*!< \brief Total Equivalent Area coefficient for all the boundaries. */
-  Total_CNearFieldOF, /*!< \brief Total Near-Field Pressure coefficient for all the boundaries. */
-  Total_CpDiff,       /*!< \brief Total Equivalent Area coefficient for all the boundaries. */
-  Total_HeatFluxDiff, /*!< \brief Total Equivalent Area coefficient for all the boundaries. */
-  Total_MassFlowRate; /*!< \brief Total Mass Flow Rate on monitored boundaries. */
+  *Inflow_MassFlow = nullptr,    /*!< \brief Mass flow rate for each boundary. */
+  *Exhaust_MassFlow = nullptr,   /*!< \brief Mass flow rate for each boundary. */
+  *Inflow_Pressure = nullptr,    /*!< \brief Fan face pressure for each boundary. */
```
pcarruscag: ...
using c++11 default initialisation instead of having to repeat this in all constructors

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -27,160 +27,69 @@
...
+  string description;
+  unsigned short nSecVar;
+  if (navier_stokes) {
+    description = "Navier-Stokes";
+    nSecVar = 8;
+  }
+  else {
+    description = "Euler";
+    nSecVar = 2;
+  }
```
pcarruscag: ...
re using the main CEulerSolver constructor also when instantiating the NS solver, which now only needs to allocate/init its own members.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -481,193 +251,73 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config, unsigned short
...
+      X[i] = new su2double* [N[i]];
+      for(unsigned long j = 0; j < N[i]; ++j)
+        X[i][j] = new su2double [P] ();
     }
-  }
+  };
+
+  /*--- Store the value of the characteristic primitive variables at the boundaries ---*/
+
+  Alloc3D(nMarker, nVertex, nPrimVar, CharacPrimVar);
```
pcarruscag: ...
using some lambdas to avoid repetition of allocation operations.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -4215,32 +3917,16 @@ void CEulerSolver::Pressure_Forces(CGeometry *geometry, CConfig *config) {
 
   /*-- Variables initialization ---*/
 
-  Total_CD = 0.0;           Total_CL = 0.0;    Total_CSF = 0.0;     Total_CEff = 0.0;
-  Total_CMx = 0.0;          Total_CMy = 0.0;   Total_CMz = 0.0;
-  Total_CoPx = 0.0;         Total_CoPy = 0.0;  Total_CoPz = 0.0;
-  Total_CFx = 0.0;          Total_CFy = 0.0;   Total_CFz = 0.0;
-  Total_CT = 0.0;           Total_CQ = 0.0;    Total_CMerit = 0.0;
+  TotalCoeff.setZero();
```
pcarruscag: (here is an example of how those auxiliary types help)

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -4425,143 +4107,114 @@ void CEulerSolver::Pressure_Forces(CGeometry *geometry, CConfig *config) {
...
+
+    su2double* buffer = new su2double [nMarkerMon];
+
+    auto Allreduce_inplace = [buffer](int size, su2double* x) {
+      SU2_MPI::Allreduce(x, buffer, size, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+      for(int i=0; i<size; ++i) x[i] = buffer[i];
+    };
+
+    Allreduce_inplace(nMarkerMon, SurfaceInvCoeff.CL);
+    Allreduce_inplace(nMarkerMon, SurfaceInvCoeff.CD);
```
pcarruscag: ...
using some more lambdas to simplify these MPI reductions.

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -38,304 +39,276 @@
...
    * \brief Return nodes to allow CSolver::base_nodes to be set.
    */
   inline CVariable* GetBaseClassPointerToNodes() final { return nodes; }
 
-public:
+  /*!
+   * \brief Generic implementation of explicit iterations (RK, Classic RK and EULER).
+   */
+  template<ENUM_TIME_INT IntegrationType>
+  void Explicit_Iteration(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iRKStep);
```
pcarruscag: ...
implementing all explicit iterations (which are very similar) via a function template.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -376,65 +167,40 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config, unsigned short
...
 
-  Vector   = new su2double[nDim]; for (iDim = 0; iDim < nDim; iDim++) Vector[iDim]   = 0.0;
-  Vector_i = new su2double[nDim]; for (iDim = 0; iDim < nDim; iDim++) Vector_i[iDim] = 0.0;
-  Vector_j = new su2double[nDim]; for (iDim = 0; iDim < nDim; iDim++) Vector_j[iDim] = 0.0;
-
-  /*--- Define some auxiliary vectors related to the primitive solution ---*/
-
-  Primitive   = new su2double[nPrimVar]; for (iVar = 0; iVar < nPrimVar; iVar++) Primitive[iVar]   = 0.0;
-  Primitive_i = new su2double[nPrimVar]; for (iVar = 0; iVar < nPrimVar; iVar++) Primitive_i[iVar] = 0.0;
-  Primitive_j = new su2double[nPrimVar]; for (iVar = 0; iVar < nPrimVar; iVar++) Primitive_j[iVar] = 0.0;
```
pcarruscag: ...
this kind of small auxiliary member variables, are giving way to local variables

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3325,99 +2950,170 @@ void CEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_conta
...
+  /*--- Start OpenMP parallel section. ---*/
+
+  SU2_OMP_PARALLEL_(reduction(+:counter_local))
+  {
+  /*--- Pick one numerics object per thread. ---*/
+  CNumerics* numerics = numerics_container[CONV_TERM + omp_get_thread_num()*MAX_TERMS];
+
+  /*--- Static arrays of MUSCL-reconstructed primitives and secondaries (thread safety). ---*/
+  su2double Primitive_i[MAXNVAR] = {0.0}, Primitive_j[MAXNVAR] = {0.0};
+  su2double Secondary_i[MAXNVAR] = {0.0}, Secondary_j[MAXNVAR] = {0.0};
```
pcarruscag: ...
which are statically allocated (per thread)

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3325,99 +2950,170 @@ void CEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_conta
...
 
-void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_container, CNumerics *numerics,
-                                   CConfig *config, unsigned short iMesh) {
+void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_container,
+                                   CNumerics **numerics_container, CConfig *config, unsigned short iMesh) {
 
-  su2double **Gradient_i, **Gradient_j, Project_Grad_i, Project_Grad_j, RoeVelocity[3] = {0.0,0.0,0.0}, R, sq_vel, RoeEnthalpy,
-  *V_i, *V_j, *S_i, *S_j, *Limiter_i = NULL, *Limiter_j = NULL, sqvel, Sensor_i, Sensor_j, Dissipation_i, Dissipation_j, *Coord_i, *Coord_j;
+  assert(nDim <= MAXNDIM && nPrimVar <= MAXNVAR && nSecondaryVar <= MAXNVAR &&
+         "Oops! The CEulerSolver static array sizes are not large enough.");
```
pcarruscag: ...
the static sizes are checked (I might move this to the constructor) and defined in only one place.

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -38,304 +39,276 @@
  */
 class CEulerSolver : public CSolver {
 protected:
+  enum : size_t {MAXNDIM = 3};    /*!< \brief Max number of space dimensions, used in some static arrays. */
+  enum : size_t {MAXNVAR = 12};   /*!< \brief Max number of variables, used in some static arrays. */
```
pcarruscag: ...
i.e. here.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3426,238 +3122,226 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
-      nodes->SetRoe_Dissipation(jPoint,numerics->GetDissipation());
+      /*--- Set implicit computation ---*/
+      if (implicit)
+        Jacobian.UpdateBlocks(iEdge, iPoint, jPoint, residual.jacobian_i, residual.jacobian_j);
     }
 
+    /*--- Viscous contribution. ---*/
+
+    Viscous_Residual(iEdge, geometry, solver_container,
+                     numerics_container[VISC_TERM + omp_get_thread_num()*MAX_TERMS], config);
```
pcarruscag: For a bit of extra performance and generality in the context of that scatter-to-gather transformation, I merged the convective and viscous loops, I've done this the polymorphic way with a `Viscous_Residual` method that acts on a single edge and only the NS solver implements.

I may yet do the same for the turbulence solvers since there it makes less sense to have separate convective and viscous loops.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3694,27 +3381,21 @@ void CEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_contain
...
-      numerics->ComputeResidual(Residual, Jacobian_i, config);
+      auto residual = numerics->ComputeResidual(config);
 
       /*--- Add the source residual to the total ---*/
-      LinSysRes.AddBlock(iPoint, Residual);
+      LinSysRes.AddBlock(iPoint, residual);
 
       /*--- Add the implicit Jacobian contribution ---*/
-      if (implicit) Jacobian.AddBlock(iPoint, iPoint, Jacobian_i);
+      if (implicit) Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
```
pcarruscag: You will see that in boundary conditions and Source_Residual loops this new method is used to write directly to the diagonal of the matrix, `AddBlock2Diag` uses the lookup array to access the diagonal instead of the exhaustive search in `AddBlock`, there is no loss of generality since that lookup array is always created.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -11037,31 +10554,37 @@ void CEulerSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_cont
...
+            }
           }
 
           /*--- Add Residuals and Jacobians ---*/
 
           LinSysRes.AddBlock(iPoint, Residual);
+
           if (implicit)
-            Jacobian.AddBlock(iPoint, iPoint, Jacobian_i);
+            Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
```
pcarruscag: There was an inconsistency here where a weighted average of the residual was being done without the corresponding averaging of the Jacobian, some regressions are failing due to this change.

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -37,9 +37,9 @@ COutputLegacy::COutputLegacy(CConfig *config) {
   size = SU2_MPI::GetSize();
 
   unsigned short iDim, iZone, iSpan, iMarker;
-  
+
   /*--- Initialize point and connectivity counters to zero. ---*/
```
pcarruscag: It seems I accidentally removed the trailing spaces in this file (output_structure_legacy.cpp)... If that inconveniences anyone let me know.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -58,6 +49,48 @@
...
+
+#include "../../include/integration/CSingleGridIntegration.hpp"
+#include "../../include/integration/CMultiGridIntegration.hpp"
+#include "../../include/integration/CStructuralIntegration.hpp"
+#include "../../include/integration/CFEM_DG_Integration.hpp"
```
pcarruscag: The integration has been restructured into these files, I also refactored the classes a bit (lot) moving methods from the parent to the children when appropriate and making things private/protected to reduce the public interface of CIntegration as much as possible.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -0,0 +1,763 @@
...
+    default:
+      direct = false;
+      break;
+  }
+
+  const unsigned short Solver_Position = config[iZone]->GetContainerPosition(RunTime_EqSystem);
+
+  /*--- Start an OpenMP parallel region covering the entire MG iteration, if the solver supports it. ---*/
+
+  SU2_OMP_PARALLEL_(if(solver_container[iZone][iInst][MESH_0][Solver_Position]->GetHasHybridParallel()))
```
pcarruscag: The code only starts threads if the current solver "declares" support for OpenMP, solvers not covered by the implementation (Incompressible Euler/NS, heat, and transition) are **not** broken by compiling the code with OpenMP support.

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -39,53 +40,54 @@ CIntegration::CIntegration(CConfig *config) {
...
-  
+
   if (dual_time)
     solver_container[MainSolver]->SetResidual_DualTime(geometry, solver_container, config, iRKStep, iMesh, RunTime_EqSystem);
-  
+
+  /// TODO: No boundary condition supports hybrid parallelism yet, master thread does all the work.
+
+  SU2_OMP_MASTER
+  {
```
pcarruscag: As mentioned, boundary conditions are all dealt with inside a large "master" region.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2987,28 +2512,59 @@ void CEulerSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
 
   /*--- Update the angle of attack at the far-field for fixed CL calculations (only direct problem). ---*/
 
-  if ((fixed_cl) && (!disc_adjoint) && (!cont_adjoint)) { SetFarfield_AoA(geometry, solver_container, config, iMesh, Output); }
+  if (fixed_cl && !disc_adjoint && !cont_adjoint) {
+    SU2_OMP_MASTER
+    SetFarfield_AoA(geometry, solver_container, config, iMesh, Output);
+    SU2_OMP_BARRIER
```
pcarruscag: Making those methods protected allows the decision to call them in parallel or serial to be centralized all in one function (Preprocessing in this case).
Similar, but less extensive, changes were made to the turbulence solvers.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -0,0 +1,763 @@
...
+
+  /// TODO: Need to check for possible race condition here (multiple coarse points setting the same fine).
+
+  SU2_OMP_FOR_STAT(roundUpDiv(geo_coarse->GetnPointDomain(), omp_get_num_threads()))
+  for (Point_Coarse = 0; Point_Coarse < geo_coarse->GetnPointDomain(); Point_Coarse++) {
+    for (iChildren = 0; iChildren < geo_coarse->node[Point_Coarse]->GetnChildren_CV(); iChildren++) {
+      Point_Fine = geo_coarse->node[Point_Coarse]->GetChildren_CV(iChildren);
+      sol_fine->LinSysRes.SetBlock(Point_Fine, sol_coarse->GetNodes()->GetSolution_Old(Point_Coarse));
+    }
+  }
```
pcarruscag: You might know the answer to this by heart, is there possibility for overlap in the multigrid agglomeration? i.e. one fine cell assigned to more than one coarse cell.
Empirically it seems the answer is no (as the code does not crash).

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2331,6 +2331,8 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
                              CSolver   ***solver_container,
                              CConfig   *config) {
 
+  /// TODO: Add OpenMP stuff to this method.
+
   /* Adapt the CFL number on all multigrid levels using an
    exponential progression with under-relaxation approach. */
```
pcarruscag: This is the only outstanding large loop used in a fluid iteration, but I could use some clarification. It seems that the final values of max/min/ave CFL end up being based only on the last visited grid (there is an outer loop over mesh levels).
Is this by design or were these values intended to be min/max/avg over all levels?

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -155,7 +155,7 @@ struct CLimiterDetails<VENKATAKRISHNAN>
 template<>
 struct CLimiterDetails<VENKATAKRISHNAN_WANG>
 {
-  su2activematrix fieldMin, fieldMax;
+  static su2activevector sharedMin, sharedMax;
   su2activevector eps2;
```
pcarruscag: For this limiter we need a min/max of the field over the entire domain, computing these reductions with OpenMP requires some shared variables that all threads can access. Since all threads call `computeLimiters`, the `CLimiterDetails` object(s) will be thread-private, the cleanest way I could think of to have shared state was to make the variables static.

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -39,53 +40,54 @@ CIntegration::CIntegration(CConfig *config) {
...
   bool dual_time = ((config->GetTime_Marching() == DT_STEPPING_1ST) ||
                     (config->GetTime_Marching() == DT_STEPPING_2ND));
 
   /*--- Compute inviscid residuals ---*/
-  
+
   switch (config->GetKind_ConvNumScheme()) {
     case SPACE_CENTERED:
-      solver_container[MainSolver]->Centered_Residual(geometry, solver_container, numerics[CONV_TERM], config, iMesh, iRKStep);
+      solver_container[MainSolver]->Centered_Residual(geometry, solver_container, numerics, config, iMesh, iRKStep);
```
pcarruscag: I had to remove the `[CONV_TERM]` bit because I need to pass an array of numerics such that each thread picks one... which gives a bit more freedom to those functions than what they should have IMO, but there we are...

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -861,22 +846,14 @@ void CFEASolver::Compute_StiffMatrix(CGeometry *geometry, CNumerics **numerics,
...
+      SU2_OMP_FOR_DYN(roundUpDiv(OMP_MIN_SIZE, ColorGroupSize)*ColorGroupSize)
       for(auto k = 0ul; k < color.size; ++k) {
 
         auto iElem = color.indices[k];
-#else
-    /*--- Natural coloring. ---*/
-    {
-      for (auto iElem = 0ul; iElem < nElement; iElem++) {
-#endif
+
```
pcarruscag: I've done some cleanup, to avoid this #ifdef regions in loops with grid coloring, for readability and so on.

Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -495,3 +518,77 @@ T colorSparsePattern(const T& pattern, size_t groupSize = 1,
...
+struct DummyGridColor
+{
+  static_assert(std::is_integral<T>::value,"");
+
+  T size;
+  struct {
+    inline T operator[] (T i) const {return i;}
+  }
+  indices;
+
```
pcarruscag: When compiling without OpenMP support the vector of colors is a static array with one color of this dummy type, which after optimization is perfectly equivalent to a plain for loop with an integer variable (see operator [] above).

Common/include/omp_structure.hpp
```diff
@@ -122,3 +126,49 @@ inline size_t computeStaticChunkSize(size_t totalWork,
...
+  SU2_OMP_CRITICAL
+  lhs += rhs;
+}
+template<class T,
+         typename std::enable_if<std::is_arithmetic<T>::value,bool>::type = 0>
+inline void atomicAdd(T rhs, T& lhs)
+{
+  SU2_OMP_ATOMIC
+  lhs += rhs;
+}
```
pcarruscag: Another small bit of probably unusual code for most.
This is SFINAE (substitution failure is not an error), the bottom function is active for arithmetic types (int, long, floats, doubles, etc.) for which atomic operations are defined by the OpenMP standard.
The top function will be used otherwise, using a critical region is slower but compatible with any type, which allows writing code that compiles with the AD types without #ifdefs in the code.
Forward mode AD + OpenMP is now enabled.

pcarruscag:
Thanks for the review @talbring 
Yeah too many changes... but like the Portuguese say "I already had my hand in the dough...".

I'm thinking of creating a page under developer docs for this OpenMP stuff, brief introduction (with links to external resources), some applied "theory" to SU2, explanation of the high level design, general do's and don'ts, some documentation of the mechanics that allow hiding #pragmas and #ifdefs, and finally links to all related PR's and issues which will serve to document the implementation. How does that sound?
It would also be good to record the virtual meeting and add it to this documentation page.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: @economon, you seem to be around more often these days, can you have a look at this question please.

economon: Yep, you are right.. this was just an oversight. Since the min/max/avg are just for reporting, it is probably best to report only the fine grid in the output. We can just add if (iMesh == MESH_0) { ... } around the relevant parts.

PR #866, opened 2020-02-09
restruct single-zone turbo problem to run without TurboDriver and Tur…

salvovitale:
## Proposed Changes
This PR is a preparatory step to reintegrate all the turbo-machinery features with the current develop.
In this first step, I have restructured the single-zone turbomachinery problem to run using both the SingleZoneDriver and the FluidIteration. 
The Next steps will be 

- [x] Adapt the turbomachinery output with the new Output structure. 
- [ ] Fix the Multizone steady turbomachinery problem to run with the MultizoneDriver.


SU2_CFD/include/iteration_structure.hpp
```diff
@@ -542,35 +542,36 @@ class CTurboIteration : public CFluidIteration {
...
-                   CGeometry ****geometry,
-                   CSolver *****solver,
-                   CNumerics ******numerics,
-                   CConfig **config,
-                   CSurfaceMovement **surface_movement,
-                   CVolumetricMovement ***grid_movement,
-                   CFreeFormDefBox*** FFDBox,
-                   unsigned short val_iZone,
-                   unsigned short val_iInst);
+  // void Preprocess(COutput *output,
```
arubino: Why these methods have been commented out?

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -291,18 +291,62 @@ class CEulerSolver : public CSolver {
              **ExtAverageKine,
              **ExtAverageOmega;
 
+  su2double  ***TurboVelocityIn,
```
arubino: Let's comment these variables (please, see comment style variables above)

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -133,20 +133,64 @@ CEulerSolver::CEulerSolver(void) : CSolver() {
...
-  DensityOut                    = NULL;
-  PressureOut                   = NULL;
-  TurboVelocityOut              = NULL;
-  KineIn                        = NULL;
-  OmegaIn                       = NULL;
-  NuIn                          = NULL;
-  KineOut                       = NULL;
-  OmegaOut                      = NULL;
-  NuOut                         = NULL;
+  /*--- Initialize quantities for turboperformace ---*/
```
arubino: I am wondering, I like you replaced `NULL` to `nullptr` but this looks quite inconsistent with general style of the code. I would open a separate issue to replace all `NULL` into `nullptr`

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -291,18 +291,62 @@ class CEulerSolver : public CSolver {
...
+             **TotalPressureIn,
+             **TotalPressureOut,
+             **TotalTemperatureOut,
+             **EnthalpyIn,
+             **TurbIntensityIn,
+             **Turb2LamViscRatioIn,
+             **TurbIntensityOut,
+             **Turb2LamViscRatioOut,
+             **NuFactorIn,
+             **NuFactorOut;
```
pcarruscag: Is there a more generic way to store these quantities? Maybe something like what I propose in #861 to cleanup the constructors / destructors of CEulerSolver and CNSSolver (the bit on the auxiliary type to store aerodynamic coefficients),

SU2_CFD/src/integration_time.cpp
```diff
@@ -691,6 +691,25 @@ void CMultiGridIntegration::NonDimensional_Parameters(CGeometry **geometry, CSol
...
+
+      /*--- Average quantities at the inflow and outflow boundaries ---*/ 
+  
+      solver_container[FinestMesh][FLOW_SOL]->TurboAverageProcess(solver_container[FinestMesh], geometry[FinestMesh],config,INFLOW);
+      solver_container[FinestMesh][FLOW_SOL]->TurboAverageProcess(solver_container[FinestMesh], geometry[FinestMesh], config, OUTFLOW);
+    
+      /*--- Gather Inflow and Outflow quantities on the Master Node to compute performance ---*/
+    
+      solver_container[FinestMesh][FLOW_SOL]->GatherInOutAverageValues(config, geometry[FinestMesh]);
+
```
arubino: I would remove empty line (but I see this in many other parts of the code)

SU2_CFD/src/integration_time.cpp
```diff
@@ -691,6 +691,25 @@ void CMultiGridIntegration::NonDimensional_Parameters(CGeometry **geometry, CSol
       solver_container[FinestMesh][FLOW_SOL]->Pressure_Forces(geometry[FinestMesh], config);
       solver_container[FinestMesh][FLOW_SOL]->Momentum_Forces(geometry[FinestMesh], config);
       solver_container[FinestMesh][FLOW_SOL]->Friction_Forces(geometry[FinestMesh], config);
+
+      /*--- Calculate the turbo performance ---*/
+      if (config->GetBoolTurbomachinery()){
+
+      /*--- Average quantities at the inflow and outflow boundaries ---*/ 
+
```
arubino: I would remove empty line (but I see this in many other parts of the code)

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -13408,3 +13897,246 @@ void CEulerSolver::GatherInOutAverageValues(CConfig *config, CGeometry *geometry
...
+
+  unsigned short nDim = geometry->GetnDim();
+  unsigned short iMarkerTP, iSpan, iDim, iStage, iBlade;
+  unsigned short nMarkerTP = config->GetnMarker_Turbomachinery();
+  su2double area, absVel2, soundSpeed, mach, tangVel, tangVel2, *relVel, relVel2;
+  su2double relPressureIn, relPressureOut, enthalpyOutIs, relVelOutIs2;
+  relVel = new su2double[nDim];
+  su2double muLam, kine, omega, nu;
+  bool turbulent = ((config->GetKind_Solver() == RANS) || (config->GetKind_Solver() == DISC_ADJ_RANS));
+  bool menter_sst       = (config->GetKind_Turb_Model() == SST);
```
pcarruscag: Beware that there is also SST_SUST now, might not apply to turbomachinery but... it exists.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -13408,3 +13897,246 @@ void CEulerSolver::GatherInOutAverageValues(CConfig *config, CGeometry *geometry
...
+
+        for (iDim = 0; iDim < nDim; iDim++){
+          absVel2   += TurboVelocityOut[iMarkerTP][iSpan][iDim]*TurboVelocityOut[iMarkerTP][iSpan][iDim];
+        }
+        TurboVelocityOut[iMarkerTP][iSpan][nDim] = sqrt(absVel2);
+
+        area   = geometry->GetSpanAreaOut(iMarkerTP, iSpan);
+
+        /*--- Compute all the Outflow quantities ---*/
+        FluidModel->SetTDState_Prho(PressureOut[iMarkerTP][iSpan], DensityOut[iMarkerTP][iSpan]);
```
arubino: Maybe let's reduce column length.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -13408,3 +13897,246 @@ void CEulerSolver::GatherInOutAverageValues(CConfig *config, CGeometry *geometry
     }
   }
 }
+
+void CEulerSolver::ComputeTurboPerformance(CConfig *config, CGeometry *geometry) {
```
pcarruscag: I find this kind of post processing we have (turbo or not) very long. Would it make sense to break this up a bit? e.g. inlet quantities, outlet quantities, differences thereof.
Similarly these functions often run over a number of markers, maybe you could have an array of some type that stores the quantities per marker and the post processing function would operate on one marker at a time? (it would at least shorten the long lines)

SU2_CFD/include/solvers/CEulerSolver.hpp
salvovitale: Yeah I agree, I ll create a new Struct or a Class as a container for these quantities.
Perhaps we can come up also with a new file structure where we can put this classes. Do you have any suggestion?

SU2_CFD/src/solvers/CEulerSolver.cpp
salvovitale: Honestly I did not know, thanks for pointing it out.

SU2_CFD/src/solvers/CEulerSolver.cpp
salvovitale: Have you done the same for Aerodynamic coefficients in your #861? But indeed the method can be improved and simplified. I ll definitely improved for the final PR to develop. For the timebeing I am just trying to have a common branch with @arubino so that he has something to work on for the new Turbo Output

SU2_CFD/include/solvers/CEulerSolver.hpp
pcarruscag: I think that @talbring wants to eventually move this post processing to the new output.
Until then maybe we could have a "tools" folder in "solvers", I think we have something similar in the output.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: I only cleaned-up the mpi regions a little.

SU2_CFD/include/solvers/CEulerSolver.hpp
salvovitale: @pcarruscag  I have created new set of classes under the folder structure "objectives/**". Now there is only a single object going around in the CEulerSolver, the CTurbomachineryPerformance.

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
+/*!
+ * \file CTurbomachineryPerformance.hpp
+ * \brief Headers of the Turbomachinery Performance class.
+ * \author S. Vitale
+ * \version 7.0.1 "Blackbird"
```
pcarruscag: 7.0.2

pcarruscag: 7.0.2

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  private:
+  su2double Density, Pressure, TangVelocity;
+  vector <su2double> Velocity;
+  public:
+  CTurbomachineryPrimitiveState();
+
+  CTurbomachineryPrimitiveState(su2double density, su2double pressure, su2double *velocity, unsigned short nDim,
+                                su2double tangVel);
+
+  ~CTurbomachineryPrimitiveState();
```
pcarruscag: If the default ctor and the dtor are empty I prefer not to declare/define them at all, if you want to use the `= default;` syntax.
Because user defined destructors disable the compiler-generated move operations.

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  su2double GetPressure() const { return Pressure; }
+
+  su2double GetEntropy() const { return Entropy; }
+
+  su2double GetEnthalpy() const { return Enthalpy; }
+
+  su2double GetTemperature() const { return Temperature; }
+
+  su2double GetTotalTemerature() const { return TotalTemperature; }
```
pcarruscag: typo

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+class CTurbomachineryBladePerformance {
+  protected:
+  CTurbomachineryState InletState;
+  CTurbomachineryState OutletState;
+  su2double KineticEnergyLoss, TotalPressureLoss, EntropyGen, PressureRatio, EulerianWork;
+  CFluidModel *FluidModel;
+
+  public:
+  CTurbomachineryBladePerformance(CFluidModel *const fluidModel, unsigned short nDim, su2double areaIn,
```
pcarruscag: This usage of `*const` (and in other methods too) is somewhat misleading, as it means the pointer is const but the object pointed to is not. Also, references should be preferred IMO as they live no room for passing nullptr.

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+};
+
+
+class CTurbomachineryBladePerformance {
+  protected:
+  CTurbomachineryState InletState;
+  CTurbomachineryState OutletState;
+  su2double KineticEnergyLoss, TotalPressureLoss, EntropyGen, PressureRatio, EulerianWork;
+  CFluidModel *FluidModel;
```
pcarruscag: Here especially, a reference would leave no doubt as to whether this class is responsible for destroying the fluid model.
And in this class you should probably delete the default constructor `= delete;` syntax since a default-constructed `CTurbomachineryBladePerformance` is probably no good?

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+class CTurbomachineryPerformance {
+  private:
+  vector <vector<shared_ptr < CTurbomachineryBladePerformance>>>
+  BladesPerformances;
+
+  static void ComputePerBlade(vector <shared_ptr<CTurbomachineryBladePerformance>> const bladePerformances,
+                              vector <CTurbomachineryCombinedPrimitiveStates> const bladePrimitives);
+
+  static void ComputePerSpan(shared_ptr <CTurbomachineryBladePerformance> const spanPerformances,
+                             const CTurbomachineryCombinedPrimitiveStates &spanPrimitives);
```
pcarruscag: Try to be consistent with where you place `const`, and if in the bottom function spanPrimitives is passed by reference I suppose a vector of them (bladePrimitives in top function) should also be passed by reference).

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  // vector<shared_ptr<CTurbomachineryStagePerformance>> StagePerformances;
+  // shared_ptr<CTurbomachineryStagePerformance> MachinePerformances;
+  public:
+  CTurbomachineryPerformance(CConfig *const config, CGeometry *const geometry, CFluidModel *const fluidModel);
```
pcarruscag: CConfig is mostly const-correct right now, so you should be able to use `const CConfig*`. The other things I would not make const to make eventual refactoring easier.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4357,6 +4374,12 @@ class CSolver {
                                          unsigned short valSpan,
                                          su2double valOmega) { }
 
+  /*!
+   * \brief Getter for TurbomachineryPerformance
+   * \return TurbomachineryPerformance container
+   */
+  inline virtual shared_ptr<CTurbomachineryPerformance> GetTurbomachineryPerformance() const {return 0;}
```
pcarruscag: nullptr

SU2_CFD/src/objectives/turbomachinery/CTurbomachineryPerformance.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+CTurbomachineryCombinedPrimitiveStates::CTurbomachineryCombinedPrimitiveStates(const CTurbomachineryPrimitiveState& inletPrimitiveState, 
+                                        const CTurbomachineryPrimitiveState&  outletPrimitiveState) : InletPrimitiveState(inletPrimitiveState), OutletPrimitiveState(outletPrimitiveState){}
+
+CTurbomachineryCombinedPrimitiveStates::~CTurbomachineryCombinedPrimitiveStates(){}
+
+CTurbomachineryState::CTurbomachineryState(){
+  Density = Pressure = Entropy = Enthalpy = Temperature = TotalTemperature = TotalPressure = TotalEnthalpy = 0.0;
+  AbsFlowAngle = FlowAngle = MassFlow = Rothalpy = TotalRelPressure = 0.0;
+  Area = Radius = 0.0;
```
pcarruscag: In c++11 this type of default value initialization can be done in the declaration of the class members directly, which saves a lot of repetition.

SU2_CFD/src/objectives/turbomachinery/CTurbomachineryPerformance.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+CTurbomachineryState::CTurbomachineryState(unsigned short nDim, su2double area, su2double radius) : CTurbomachineryState() {
+  generate_n(back_inserter(Velocity),nDim, [] {return 0.0;});
```
pcarruscag: Is this the same as std::fill ?

SU2_CFD/src/objectives/turbomachinery/CTurbomachineryPerformance.cpp
```diff
@@ -0,0 +1,177 @@
...
+  std::for_each(RelMach.begin(), RelMach.end(), [&](su2double &el) { el /= soundSpeed; });
+
+  /*--- Compute total relative TD quantities ---*/
+  Rothalpy = Enthalpy + 0.5 * relVel2 - 0.5 * tangVel2;
+  fluidModel->SetTDState_hs(Rothalpy, Entropy);
+  TotalRelPressure = fluidModel->GetPressure();
+
+}
+
+CTurbomachineryBladePerformance::CTurbomachineryBladePerformance(CFluidModel* const fluidModel, unsigned short nDim, su2double areaIn, su2double radiusIn, su2double areaOut, su2double radiusOut) : FluidModel(fluidModel){
```
pcarruscag: some lines are a bit long

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -2627,6 +2643,12 @@ class CEulerSolver : public CSolver {
     ExtAverageOmega[valMarker][valSpan] = valOmega;
   }
 
+  /*!
+   * \brief Getter for TurbomachineryPerformance
+   * \return TurbomachineryPerformance container
+   */
+  inline shared_ptr<CTurbomachineryPerformance> GetTurbomachineryPerformance() const {return TurbomachineryPerformance;}
```
pcarruscag: This method is not really const, it returns the shared pointer by value by that pointer allows the modification of the pointed-to object. Also if the purpose here if to provide read-only access a bare const pointer would be an acceptable return type IMO (saves the reference counting etc. and should let you use unique_ptr for `TurbomachineryPerformance` instead)

SU2_CFD/src/objectives/turbomachinery/CTurbomachineryPerformance.cpp
```diff
@@ -0,0 +1,177 @@
...
+CTurbomachineryPerformance::~CTurbomachineryPerformance() {}
+
+void CTurbomachineryPerformance::ComputeTurbomachineryPerformance(
+  vector <vector<CTurbomachineryCombinedPrimitiveStates>> const bladesPrimitives) {
+  boost::for_each(BladesPerformances, bladesPrimitives, boost::bind(&ComputePerBlade, _1, _2));
+}
+
+void CTurbomachineryPerformance::ComputePerBlade(
+  vector <shared_ptr<CTurbomachineryBladePerformance>> const bladePerformances,
+  vector <CTurbomachineryCombinedPrimitiveStates> const bladePrimitives) {
```
pcarruscag: possibly more example of pass by reference vs pass by value

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+//    public:
+//       CCompressorStagePerformance();
+//       ~CCompressorStagePerformance();
+//       void ComputePerformance(CFluidModel* fluidModel, su2double tangVel) override;
+
+// };
+
+class CTurbomachineryPerformance {
+  private:
+  vector <vector<shared_ptr < CTurbomachineryBladePerformance>>>
```
pcarruscag: Is shared_ptr absolutely necessary or would unique_ptr be enough? i.e. is there shared ownership of the various CTurbomachineryBladePerformance's ?

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+#pragma once
+
+#include <vector>
+#include <memory>
+#include <algorithm>
+#include <iterator>
+#include <numeric>
+#include <boost/bind.hpp>
+#include <boost/range/algorithm_ext/for_each.hpp>
```
economon: I'm also for avoiding boost if possible

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  su2double GetPressure() const { return Pressure; }
+
+  su2double GetTangVelocity() const { return TangVelocity; }
+
+  vector <su2double> GetVelocity() const { return Velocity; }
+
+};
+
+class CTurbomachineryCombinedPrimitiveStates {
```
economon: Like the expressive names, but can they be shortened slightly without losing meaning (this is a personal preference)?

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+
+  su2double GetTotalEnthalpy() const { return TotalEnthalpy; }
+
+  su2double GetAbsFlowAngle() const { return AbsFlowAngle; }
+
+  su2double GetFlowAngle() const { return FlowAngle; }
+
+  su2double GetMassFlow() const { return MassFlow; }
+
+  su2double GetRothalpy() const { return Rothalpy; }
```
economon: Don't forget doxy docs for all new methods

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  su2double GetRelMachValue() const {
+    return sqrt(inner_product(RelMach.begin(), RelMach.end(), RelMach.begin(), 0.0));
+  }
+};
+
+
+class CTurbomachineryBladePerformance {
+  protected:
```
economon: I would go for private data members wherever possible (I know a lot of legacy classes do it, but it is not the best practice)

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
economon: You could make it a const reference to fluid model and set it using an initializer list with the constructor

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  public:
+  CCompressorBladePerformance(CFluidModel *const fluidModel, unsigned short nDim, su2double areaIn, su2double radiusIn,
+                              su2double areaOut, su2double radiusOut);
+
+  ~CCompressorBladePerformance();
+
+  void ComputePerformance(const CTurbomachineryCombinedPrimitiveStates &primitives) override;
+
+};
+// class CTurbomachineryStagePerformance :  CTurbomachineryBladePerformance {
```
economon: Remove commented code wherever possible

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+  public:
+  CTurbomachineryPerformance(CConfig *const config, CGeometry *const geometry, CFluidModel *const fluidModel);
+
+  ~CTurbomachineryPerformance();
+
+  vector <vector<shared_ptr < CTurbomachineryBladePerformance>>>
+
+  GetBladesPerformances() const { return BladesPerformances; }
+
+  // vector<shared_ptr<CTurbomachineryStagePerformance>> GetStagePerformances() const { return StagePerformances; }
```
economon: More commented code

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -688,6 +688,25 @@ void CMultiGridIntegration::NonDimensional_Parameters(CGeometry **geometry, CSol
       solver_container[FinestMesh][FLOW_SOL]->Momentum_Forces(geometry[FinestMesh], config);
       solver_container[FinestMesh][FLOW_SOL]->Friction_Forces(geometry[FinestMesh], config);
 
+      /*--- Calculate the turbo performance ---*/
```
economon: This is just food for thought, but I have never really liked that we compute the output quantities in this location. Seems like we could have a better design to make it more general. Just a TODO for the future.

SU2_CFD/src/objectives/turbomachinery/CTurbomachineryPerformance.cpp
```diff
@@ -0,0 +1,177 @@
...
+  /*--- Compute performance ---*/
+  EntropyGen = (OutletState.GetEntropy() - InletState.GetEntropy()) / abs(InletState.GetEntropy() + 1);
+  EulerianWork = InletState.GetTotalEnthalpy() - OutletState.GetTotalEnthalpy();
+  TotalPressureLoss = (InletState.GetTotalRelPressure() - OutletState.GetTotalRelPressure()) /
+                      (InletState.GetTotalRelPressure() - OutletState.GetPressure());
+  KineticEnergyLoss = 2 * (OutletState.GetEnthalpy() - enthalpyOutIs) / relVelOutIs2;
+}
+
+CTurbineBladePerformance::~CTurbineBladePerformance() {}
+// CCompressorBladePerformance::CCompressorBladePerformance(unsigned short nDim, su2double area, su2double radius) : CTurbomachineryBladePerformance(nDim, area, radius){}
```
economon: more commented code

SU2_CFD/include/objectives/turbomachinery/CTurbomachineryPerformance.hpp
```diff
@@ -0,0 +1,254 @@
...
+                                         const CTurbomachineryPrimitiveState &outletPrimitiveState);
+
+  ~CTurbomachineryCombinedPrimitiveStates();
+
+  CTurbomachineryPrimitiveState GetInletPrimitiveState() const { return InletPrimitiveState; }
+
+  CTurbomachineryPrimitiveState GetOutletPrimitiveState() const { return OutletPrimitiveState; }
+};
+
+class CTurbomachineryState {
```
economon: We've moved to one class per file - break classes out into new files is possible

PR #877, opened 2020-02-14
Rough wall boundary condition for SA and SST turbulence models.

koodlyakshay:
## Proposed Changes
This PR is about the implementation of a rough wall boundary condition for SA and SST models. A given realistic roughness distribution must be first converted to an "equivalent sand grain roughness" height and specified as an input.  

For the SA turbulence model, the roughness boundary condition is based on:

Aupoix, B. and Spalart, P. R., "Extensions of the Spalart-Allmaras Turbulence Model to Account for Wall Roughness," International Journal of Heat and Fluid Flow, Vol. 24, 2003, pp. 454-462. 

Very briefly, the new model assumes there is a "virtual wall" located on top of the roughness elements and thus the distance to nearest wall is modified for all points based on the specified sand grain roughness. Additionally, the boundary condition at the wall is now not set to zero but a mixed condition (robin BC) is applied allowing for a non-zero eddy viscosity at the wall.  More information is also available [here](https://turbmodels.larc.nasa.gov/spalart.html#sarough)

For the SST model, the roughness boundary condition is based on:

Wilcox, D.C Turbulence Modeling for CFD., 1998 and 2006, DWC Industries

There are two suggestions (first in the 2nd edition and a corrected version in the 3rd edition). I have implemented the latest one and commented the old one. There have been numerous papers about some limitations of these models themselves however. The modification consists of changing the omega boundary condition at the wall to a different value based on local quantities. 

This PR introduces two extra config options - one to specify whether a wall is rough or not and another to specify the sand grain roughness height for each of them. 
 
    WALL_TYPE = SMOOTH/ROUGH
    WALL_ROUGHNESS = <array of doubles>

By default all walls are smooth. If neither option is present in the config file, smooth walls are assumed everywhere. However, if there are multiple walls some rough, some smooth, they need to be listed (similar to INLET TYPE) in the order they appear in the MARKER_HEATFLUX definition. I have attached some examples here. 

1. Flatplate
Here is the input file [turb_flatplate_v7_cfg.txt](https://github.com/su2code/SU2/files/4198440/turb_flatplate_v7_cfg.txt),

 The meshes can be found in TestCases/rans/flatplate or the NASA TMR database. Here are the results with both SA and SST
yplus_vs_uplus_sa_comparison

Here I am plotting the non-dimensionalised velocity u^+ against y^+. The clean results give fairly good agreement and we can see some over prediction in velocity shifts. This was also reported elsewhere

yplus_vs_uplus_sst_comparison

The same plot for SST. The results are not quite as good as SA in general, but I expected that considering some limitations reported in the literature (ex a min y^+ ~ 0.01 and so on).

Here is a comparison against experimental data by Blanchard (1977)
blanchard_compariosn

Once again, the SA roughness model does fairly well as opposed to the SST.

2. Airfoils

If only a part of the wall is rough, as is the case on airfoils with corrosion for ex, we can split the marker definition as follows:

    MARKER_HEATFLUX= (wall1, hf1, wall2, hf2 ...)
    WALL_TYPE= SMOOTH,ROUGH,...
    WALL_ROUGHNESS= 0.0, k1...

In case there is even a single rough wall, the WALL_TYPE option must specified for all the markers and WALL_ROUGHNESS set to 0.0 for smooth walls and the desired value for rough walls. 
[Sample input file for airfoil with smooth and rough walls](https://github.com/su2code/SU2/files/4198580/turb_naca62215_v7cfg.txt). The airfoil used here is the NACA 622215 and I can send the mesh and experimental data in case anyone is interested. 

## Additional notes

Modification of wall distance computation required some changes to the ADT routines and subsequently to some FEM stuff as well. To modify the wall distance with roughness, I needed to know which marker was the closest wall and add the corresponding roughness. However, I wasn't sure how to properly do this in MPI. My solution was to pass the marker roughness information to the ADT class itself. This meant I had to add a dummy argument whenever the corresponding ADT classes were called elsewhere. Just to clarify, the modification is only in CADTElem class constructor and ComputeWallDistance routines. I hope this will not be a problem. I made this as a draft PR in case there are some issues with this. Also, if there are any issues with how the config file is changed, I can update that as well. I haven't modified the config_template for this yet.

## PR Checklist


pcarruscag:
Let me know if you need any help merging with develop @koodlyakshay, I had a quick look and it seems to be mostly a matter of copy pasting the new numerics options you added into the split files introduced in #861.

Common/include/CConfig.hpp
```diff
@@ -4930,6 +4935,17 @@ class CConfig {
    */
   unsigned short GetKind_ActDisk(void) const { return Kind_ActDisk; }
 
+  /*!
+   * \brief Get the kind of wall.
+   * \return Kind of wall - smooth or rough.
+   */
+  unsigned short GetKindWall(string val_marker);
```
pcarruscag: const these new "Gets"

Common/src/CConfig.cpp
```diff
@@ -4652,6 +4658,53 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+  if ((nMarker_HeatFlux > 0) || (nMarker_Isothermal > 0)) {
+    /*--- If the config option is not declared, then assume smooth walls and set roughness to zero. ---*/
+    if (nWall_Types != nMarker_HeatFlux + nMarker_Isothermal) {
+      /*--- If this option is not used at all, assume all walls are smooth and set roughness height to zero. ---*/
+      if (nWall_Types == 0) {
+        nWall_Types = nMarker_HeatFlux + nMarker_Isothermal;
+        if (Roughness_Height != NULL) Roughness_Height = NULL;
+        Roughness_Height = new su2double [nWall_Types];
+        if (Kind_Wall != NULL) Kind_Wall = NULL;
+        Kind_Wall = new unsigned short [nWall_Types];
```
pcarruscag: This raises an eyebrow, if the thing is not null it was either not initialized properly or it was already allocated in which case it becomes a memory leak.

Common/src/CConfig.cpp
```diff
@@ -4652,6 +4658,53 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+      else
+        SU2_MPI::Error("Mismatch in number of wall markers. Specify type of wall for all markers (even if smooth).", CURRENT_FUNCTION);
+    } else {
+       if (nRoughWall != nWall_Types)
+         SU2_MPI::Error("Mismatch in number of roughness height definition and wall type definition.", CURRENT_FUNCTION);
+       else {
+         su2double *temp_rough;
+         unsigned short *temp_kindrough;
+         temp_rough = new su2double [nWall_Types];
+         temp_kindrough = new unsigned short [nWall_Types];
```
pcarruscag: memory leak, use std::vector please

Common/src/CConfig.cpp
```diff
@@ -8849,6 +8903,49 @@ su2double CConfig::GetWall_HeatFlux(string val_marker) {
   return Heat_Flux[iMarker_HeatFlux];
 }
 
+unsigned short CConfig::GetKindWall(string val_marker) {
+  unsigned short iMarker;
+  short flag = -1;
+  for (iMarker = 0; iMarker < nMarker_HeatFlux; iMarker++)
+    if (Marker_HeatFlux[iMarker] == val_marker) {
+    flag = iMarker;
```
pcarruscag: "Misleading" indentation (as some compilers call it).

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7050,7 +7068,7 @@ void CPhysicalGeometry::SetAvgTurboValue(CConfig *config, unsigned short val_iZo
     MyTotalGridVel         = new su2double[nDim];
 
     for (iDim = 0; iDim < nDim; iDim++) {
-      MyTotalTurboNormal[iDim]	= TotalTurboNormal[iDim];
+      MyTotalTurboNormal[iDim]  = TotalTurboNormal[iDim];
```
pcarruscag: Thanks for the cleanup :+1:

Common/src/CConfig.cpp
koodlyakshay: Thanks for the catch. Turns out I hadn't initialized one of the pointers (Kind_Wall) to NULL. The other one is initialized earlier. I put the if condition mostly as an insurance. Valgrind does not report any errors when I remove the if condition. Should I just remove it?

koodlyakshay:
Thanks for the comments @pcarruscag. I fixed them. I will work on merging with develop soon.

Common/src/CConfig.cpp
pcarruscag: Yes please. That method is called from the constructors so there is no risk of leaking (at the moment...). If the purpose of the check is to avoid a memory leak you need to do `delete [] Roughness_Height;`
The null-check is actually not required because a) operator delete does it; b) if the pointer is not null because it was not initialized you get an even worse bug (deleting someone else's memory).

In summary, use std::vector :)

Reasons not to use std::vector: it is usually implemented as 3 pointers (begin end and capacity) so if you have many small arrays that share a common size (the case of CConfig) it is a bit more efficient (albeit error prone) to use c-style arrays.

Common/src/CConfig.cpp
```diff
@@ -4657,6 +4664,52 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
 
+  /*--- Check that if the wall roughness array are compatible and set deafult values if needed. ---*/
+
+  if ((nMarker_HeatFlux > 0) || (nMarker_Isothermal > 0)) {
+    /*--- If the config option is not declared, then assume smooth walls and set roughness to zero. ---*/
+    if (nWall_Types != nMarker_HeatFlux + nMarker_Isothermal) {
+      /*--- If this option is not used at all, assume all walls are smooth and set roughness height to zero. ---*/
+      if (nWall_Types == 0) {
+        nWall_Types = nMarker_HeatFlux + nMarker_Isothermal;
+        Roughness_Height = new su2double [nWall_Types];
```
koodlyakshay: The pointers are initialized to NULL earlier in the SetPointersNULL routine now, so the check is removed.

Common/src/CConfig.cpp
```diff
@@ -4657,6 +4664,52 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+           Roughness_Height[iMarker] = temp_rough[iMarker];
+           Kind_Wall[iMarker] = temp_kindrough[iMarker];
+         }
+         for (iMarker = 0; iMarker < nMarker_Isothermal; iMarker++) {
+           Roughness_Height[nMarker_HeatFlux + iMarker] = temp_rough[nMarker_HeatFlux + iMarker];
+           Kind_Wall[nMarker_HeatFlux + iMarker] = temp_kindrough[nMarker_HeatFlux + iMarker];
+         }
+         /*--- Release memory allocated earlier. ---*/
+         vector<su2double> ().swap(temp_rough);
+         vector<unsigned short> ().swap(temp_kindrough);
```
koodlyakshay: Based on advice from @pcarruscag I have used std::vector here. Apparently the clear function does not release the memory, so I used swap() instead (https://stackoverflow.com/questions/10464992/c-delete-vector-objects-free-memory). This is also used elsewhere in the code.

pcarruscag: Please remove the swaps, objects are destroyed when they go out-of-scope, the use cases for explicitly deleting vectors this way is when they are class members, in which case their scope is the same of the object that owns them.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -324,7 +324,9 @@ void CTurbSASolver::Preprocessing(CGeometry *geometry, CSolver **solver_containe
 
 void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iMesh) {
 
-  const su2double cv1_3 = 7.1*7.1*7.1;
+//  su2double rho = 0.0, mu = 0.0, nu, *nu_hat, muT, Ji, Ji_3, fv1;
```
talbring: Remove these comments here

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -473,17 +507,52 @@ void CTurbSASolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_conta
 
       if (geometry->node[iPoint]->GetDomain()) {
 
-        /*--- Get the velocity vector ---*/
+        if (!rough_wall) {
```
talbring: Fix the indentation in the following block

config_template.cfg
```diff
@@ -770,7 +770,17 @@ GILES_EXTRA_RELAXFACTOR= ( 0.05, 0.05)
 %
 % YES Non reflectivity activated, NO the Giles BC behaves as a normal 1D characteristic-based BC
 SPATIAL_FOURIER= NO
+% ------------------------ WALL ROUGHNESS DEFINITION --------------------------%
+% Type of wall (SMOOTH or ROUGH). By default all walls are assumed to be smooth.
+% If multiple walls are present, list type for each of them (ex- SMOOTH, ROUGH,..) 
+% in the order they are listed in MARKER_HEATFLUX definition. If all walls are smooth
+% this option can be removed.
+%WALL_TYPE= SMOOTH
```
talbring: You have to list here all markers specified in MARKER_HEATFLUX **and** MARKER_ISOTHERMAL, right ?

koodlyakshay: Yes, you are right. Will fix this and the other comments soon. 
Thanks.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -5201,10 +5209,20 @@ void CPhysicalGeometry::ComputeWall_Distance(CConfig *config) {
...
       WallADT.DetermineNearestElement(node[iPoint]->GetCoord(), dist, markerID,
-                                   elemID, rankID);
+                                   elemID, rankID, localRoughness);
+
       node[iPoint]->SetWall_Distance(dist);
+
+      /*--- Use the markerID to find the corresponding wall roughness height. ---*/
+
+    Marker_Tag = config->GetMarker_All_TagBound(markerID);
+    roughness = config->GetWall_RoughnessHeight(Marker_Tag);
```
pcarruscag: Indentation here

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -556,6 +558,16 @@ class CNumerics {
...
 
+  /*!
+   * \brief Set the value of the roughness from the nearest wall.
+   * \param[in] val_dist_i - Value of of the roughness of the nearest wall from point i 
+   * \param[in] val_dist_j - Value of of the roughness of the nearest wall from point j 
+   */
+  void SetRoughness(su2double val_roughness_i, su2double val_roughness_j) {
+  roughness_i = val_roughness_i;
+  roughness_j = val_roughness_j;
+}
```
pcarruscag: more indentation

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -403,9 +412,22 @@ void CTurbSASolver::Source_Residual(CGeometry *geometry, CSolver **solver_contai
 
     if (config->GetKind_HybridRANSLES() == NO_HYBRIDRANSLES) {
 
+  /*--- For the SA model, wall roughness is accounted by modifying the computed wall distance
+       *                  d_new = d + 0.03 k_s
+       *    where k_s is the equivalent sand grain roughness height that is specified in cfg file.
+       *    For smooth walls, wall roughness is zero and computed wall distance remains the same. */
```
pcarruscag: looks a bit out of alignment too

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -473,17 +504,52 @@ void CTurbSASolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_conta
...
-        /*--- Includes 1 in the diagonal ---*/
+          RoughWallBC = nodes->GetSolution(iPoint,0)/(0.03*Roughness_Height);
 
-        Jacobian.DeleteValsRowi(iPoint);
+          Res_Wall[0] = coeff*RoughWallBC*Area;
+          LinSysRes.SubtractBlock(iPoint, Res_Wall);
+
+          Jacobian_i[0][0] = (laminar_viscosity*Area)/(0.03*Roughness_Height*sigma);
+          Jacobian_i[0][0] += 2.0*RoughWallBC*Area/sigma;
+          Jacobian.SubtractBlock(iPoint,iPoint,Jacobian_i);
```
pcarruscag: Replace by `Jacobian.SubtractBlock2Diag(iPoint,Jacobian_i);`

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -509,18 +584,54 @@ void CTurbSASolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_con
...
+          coeff = (nu_total/sigma);
+
+          RoughWallBC = nodes->GetSolution(iPoint,0)/(0.03*Roughness_Height);
+
+          Res_Wall[0] = coeff*RoughWallBC*Area;
+          LinSysRes.SubtractBlock(iPoint, Res_Wall);
+
+          Jacobian_i[0][0] = (laminar_viscosity*Area)/(0.03*Roughness_Height*sigma);
+          Jacobian_i[0][0] += 2.0*RoughWallBC*Area/sigma;
+          Jacobian.SubtractBlock(iPoint,iPoint,Jacobian_i);
```
pcarruscag: same here

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -420,31 +420,73 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
+      if (rough_wall) {
+        
+        /*--- Set wall values ---*/
+
+        density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
+        laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
+
+        WallShearStress = solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 0)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 0);
+        WallShearStress += solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 1)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 1);
+        if (nDim == 3) WallShearStress += solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 2)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 2);
```
pcarruscag: please write this as a loop and use the `pow` function

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -468,31 +508,73 @@ void CTurbSSTSolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_co
...
+      if (rough_wall) {
+
+        /*--- Set wall values ---*/
+
+        density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
+        laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
+
+        WallShearStress = solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 0)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 0);
+        WallShearStress += solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 1)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 1);
+        if (nDim == 3) WallShearStress += solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 2)*solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, 2);
```
pcarruscag: same here

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -269,6 +269,10 @@ void CTurbSolver::Viscous_Residual(CGeometry *geometry, CSolver **solver_contain
       numerics->SetF1blending(nodes->GetF1blending(iPoint),
                               nodes->GetF1blending(jPoint));
 
+    /*--- Roughness heights. ---*/
+    if (config->GetKind_Turb_Model() == SA)
+      numerics->SetRoughness(geometry->node[iPoint]->GetRoughnessHeight(),geometry->node[jPoint]->GetRoughnessHeight());
```
pcarruscag: break the line to be consistent with the rest of the function

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -509,18 +584,54 @@ void CTurbSASolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_con
...
-      /*--- Get the velocity vector ---*/
-      for (iVar = 0; iVar < nVar; iVar++)
-        Solution[iVar] = 0.0;
+      if (!rough_wall) {
 
-      nodes->SetSolution_Old(iPoint,Solution);
-      LinSysRes.SetBlock_Zero(iPoint);
+        /*--- Get the solution vector ---*/
+        for (iVar = 0; iVar < nVar; iVar++)
+          Solution[iVar] = 0.0;
```
pcarruscag: `Solution[0] = 0.0` instead, the same assumption (nVar==1) is made when setting the diagonal of the matrix to 1.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -120,15 +124,31 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSA::ComputeResidual(const CConfig
...
+    fv1 = Ji_3/(Ji_3+cv1_3);
+    
+    /*--- Using a modified relation so as to not change the Shat that depends on fv2. ---*/
+    fv2 = 1.0 - TurbVar_i[0]/(nu+TurbVar_i[0]*fv1);   // From NASA turb modeling resource and 2003 paper
+    
     ft2 = ct3*exp(-ct4*Ji_2);
     S = Omega;
     inv_k2_d2 = 1.0/(k2*dist_i_2);
-
+
```
pcarruscag: rogue space, you can strip all spaces to be sure

koodlyakshay: Fixed it with the replace tabs and spaces script from utils. 
Thanks.

pcarruscag:
@koodlyakshay I was looking at the ADT modifications that you mention.
Do I understand correctly that the roughness height does not influence the wall distance calculation itself? But that you simply need to know what is the roughness height associated with the closest wall point? If this is the case you can probably just use the markerId returned by the wall distance function?
As for mpi aspects, each rank sees the same ADT and I recall that we do have mechanisms to access global marker information.

koodlyakshay:
>Do I understand correctly that the roughness height does not influence the wall distance calculation itself?...

@pcarruscag , Yes. I can use the markerID to find the corresponding roughness as long as the field element and the nearest wall marker are on the same rank. I was not sure how to relate the markerID on rankID from a different rank to the global config list, hence the current implementation. I am open to fixing it with a different implementation if I can figure out how to actually do it. The few things I tried all ended up causing problems and this was one way to avoid them (also my inexperience with MPI did not help).

pcarruscag:
If I recall correctly @tobikattman was dealing with a similar issue
(relating local markers to global ones) on his symmetry boundary condition
pull request. I think it would be worth keeping the ADT as clean and
generic as possible.

koodlyakshay:
Okay, will try to fix it without modifying the ADT.  Will probably take a while though, so I will mark this as WIP.

Thank you.

koodlyakshay:
@talbring @pcarruscag I undid the changes in ADT. 

I hope this implementation (apart from a stray line in the .hpp file that I will fix) is more suitable for future plans?

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11591,9 +11590,86 @@ void CPhysicalGeometry::SetWallDistance(const CConfig *config, CADTElemClass *Wa
...
+
+  int *displs = new int [size];
+  int* recvCounts = new int [size];//, displs(size);
+  int sizeLocal = (int) nMarker_All; // number of local markers
+  
+  //Communicate size of local marker array and make an array large enough to hold all data
+  SU2_MPI::Allgather(&sizeLocal, 1, MPI_INT, recvCounts, 1,
+                       MPI_INT, MPI_COMM_WORLD);
+  
+  SU2_MPI::Barrier(MPI_COMM_WORLD);   // Not sure if needed
```
koodlyakshay, line 11642: This routine is perhaps not very efficient and I appreciate any comments to improve it.

pcarruscag:
Hi @koodlyakshay thanks for making the changes.
By default MPI barriers are "never" needed, the normal communication routines already do all synchronization required.

Efficiency is probably not fundamental for that routine but this solution feels too complicated somehow, can you attend tomorrow's developers meeting? (I'm getting the "there's gotta be a simpler way" feeling, and if we pick the brains of a few people we are certain to find it).

In the meantime, if you specify the roughness as a string+double list (exactly like MARKER_HEATFLUX), which is read with `addStringDoubleListOption` you could probably simplify the logic around heatflux and isothermal markers, which would make the setup more user friendly (having to stick with an order is bound to trip someone at some point).
Also the cht interface is "just" an isothermal boundary, any reason not to make this feature available for that?

koodlyakshay:
>Also the cht interface is "just" an isothermal boundary, any reason not to make this feature available for that?

No, I am not really familiar with the CHT part, but If it is same as an isothermal boundary, I can add this feature there as well. 

Are you suggesting I make an option say MARKER_ROUGH and read in roughness values only for the corresponding marker? I went with the current input method to make sure only wall boundaries were given roughness heights but I suppose that can be ensured in other ways too. I will experiment with a few things and see which works the best.

I will wait to see if there any suggestions in the meeting and then fix all the MPI stuff. 

Thanks.

pcarruscag:
Yes for MARKER_ROUGH, or MARKER_ROUGH_WALL to be clearer.

Regarding the MPI stuff, I had the following idea over lunch:
In CPhysicalGeometry::SetWallDistance we compute the closest distance, and in so doing we also get the mpi rank and markerID associated with the closest vertex.
So, before computing the wall distances you communicate the local marker ids and wall roughness's (via Allgather as you are doing now), with this info you can build a `unordered_map<pair<int,int>, double>` to map pairs of <rank,markerID> to the values of roughness.
Now when you loop over the points to compute the distances you can retrieve the roughness from this map instead of from config.
It's almost exactly the same as you have, but avoids using the config as a messenger between routines, and using a stl map should also make the code simpler.

koodlyakshay:
I have simplified the input method now. Users need to simply add a line
    
    MARKER_ROUGHWALL = (marker_name_1, k_1, marker_name_2, k_2, ..)

By default all walls are smooth and only the rough walls need to be listed and the order shouldn't matter.

pcarruscag:
MARKER_ROUGHWALL is fine too. Do you want some help with the MPI to avoid setting data into the config? (I don't want to create too much friction but that is something that should really be avoided).

koodlyakshay:
Hello Pedro,

No problem at all. I just got stuck on a few issues in the PB solver and other things so I wasn't able to focus on this PR in a while. But yeah, I am indeed stuck on how to update the map in MPI. What I have done so far is kept the Global roughness array in config class for the BC_* routines and created an unordered map in physical geometry class which will be used in the SetWallDistance routine instead of the config call. I haven't used maps in C++ at all, so it is taking a while. I will try a few more things and if I can't fix it I will ask on slack or in the meeting next time.

Cheers,
Akshay

pcarruscag:
Hey Akshay,
I found out the other day that this: `unordered_map<pair<int,int>, su2double>` does not work without extra tricks, and the alternative `map<pair<int,int>, su2double>` may not be very efficient, so if you have it out of the config in some matrix format it is probably better.
Cheers,
Pedro

koodlyakshay:
>I found out the other day that this: unordered_map<pair<int,int>, su2double> does not work without extra tricks,

Yes, I found some examples to do it. But I was not quite understanding what are the size of local maps and how to communicate them over MPI. 

>so if you have it out of the config in some matrix format it is probably better.

In that case, I can just move the current global arrays to physical geometry class and simplify some of the function calls.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11378,9 +11382,94 @@ void CPhysicalGeometry::SetWallDistance(const CConfig *config, CADTElemClass *Wa
...
   }
   // end SU2_OMP_PARALLEL
 }
+
+void CPhysicalGeometry::SetGlobalMarkerRoughness(CConfig *config) {
+
+unsigned short iMarker;
+#ifndef HAVE_MPI
+  
+  int *local_displ = new int [1];
```
pcarruscag, line 11405: This variable does not seem to be used.
Also, there is hardly ever a need for `new int [1]`, you can just have an int, and if you need to pass it by pointer to a function you just take its address (`int local_displ;  functionWantsPointer(&local_displ);`)

Common/include/geometry/CPhysicalGeometry.hpp
```diff
@@ -804,4 +807,23 @@ class CPhysicalGeometry final : public CGeometry {
...
+   * \param [in] total number of ranks
+   */
+  void SetGlobalMarkerArray(int *global_displ, int size) ;
+  
+    /*!
+   * \brief Set the roughness for each marker in global array.
+   * \param [in] global list of roughness for every marker.
+   * \param [in] total number of marker including send recieves.
+   */
+  void SetGlobalRoughnessArray(su2double *global_rough, int sizeGlobal) ;
```
pcarruscag, line 827: I think these are not used?

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -859,6 +859,10 @@ void CDriver::Geometrical_Preprocessing_FVM(CConfig *config, CGeometry **&geomet
     geometry[MESH_0]->ComputeMeshQualityStatistics(config);
   }
 
+  /*--- Store marker list and roughness in a global array ---*/
+
+  geometry[MESH_0]->SetGlobalMarkerRoughness(config);
+
```
pcarruscag, line 865: Is there an easy way to do this only if there are rough markers?

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11378,9 +11382,94 @@ void CPhysicalGeometry::SetWallDistance(const CConfig *config, CADTElemClass *Wa
...
       int            rankID;
       su2double      dist;
+      su2double      localRoughness;
+      int index;
+
       WallADT->DetermineNearestElement(nodes->GetCoord(iPoint), dist, markerID, elemID, rankID);
+
       nodes->SetWall_Distance(iPoint, min(dist,nodes->GetWall_Distance(iPoint)));
+      index = GlobalMarkerStorageDispl[rankID]+ markerID;
+      localRoughness = GlobalRoughness_Height[index];
```
pcarruscag, line 11392: You can use `auto` here instead of declaring the variables separately, again if possible skip this section if there are no rough markers.

Perhaps you could detect if roughness is needed in this function and call "SetGlobalMarkerRoughness" here before the wall distance loop, that way the method does not need to be virtual, it can be a private member function of CGeometry.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -314,8 +314,15 @@ void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_contain
 
     su2double nu  = mu/rho;
     su2double nu_hat = nodes->GetSolution(iPoint,0);
+    su2double roughness = geometry->nodes->GetRoughnessHeight(iPoint);
+    su2double dist = geometry->nodes->GetWall_Distance(iPoint);
+
+    dist += 0.03*roughness;
+
+    su2double Ji   = nu_hat/nu ;
+    if (roughness > 1.0e-10)
```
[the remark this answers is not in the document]
koodlyakshay: The 0.03 is, I used the 1e-10 just to check if the node is nearest to a rough wall or not. I changed it to EPS now.

Common/src/CConfig.cpp
```diff
@@ -8662,7 +8746,69 @@ su2double CConfig::GetWall_HeatFlux(string val_marker) const {
   return Heat_Flux[iMarker_HeatFlux];
 }
 
+unsigned short CConfig::GetKindWall(string val_marker) const {
+  unsigned short iMarker;
+  short flag = -1;
```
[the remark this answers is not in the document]
koodlyakshay: Indeed the two functions have the same logic. I made a new function that return a `pair<unsigned short, su2double>`  instead.

koodlyakshay:
@pcarruscag thanks for the comments. I have fixed most of them I think. 

In general, I think the PR is ready to be reviewed. I will think of a decent test case and add it soon.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11378,9 +11385,80 @@ void CPhysicalGeometry::SetWallDistance(const CConfig *config, CADTElemClass *Wa
...
+  GlobalMarkerStorageDispl[0] = 0;
+  for (int iRank = 1; iRank < size; iRank++) 
+    GlobalMarkerStorageDispl[iRank] = displs[iRank];
+  
+  //total size
+  int sizeGlobal = displs[size-1] + recvCounts[size-1];
+
+  /*--- Allocate local and global arrays to hold roughness. ---*/
+  su2double *localRough = new su2double [nMarker_All]; // local number of markers
+  su2double *globalRough = new su2double[sizeGlobal];  // all markers including send recieve
```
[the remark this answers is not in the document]
koodlyakshay: I get a compiler error when I use vectors in the no MPI version. Looks like vector.data() is not returning a pointer when the size is 1. I will use delete for now.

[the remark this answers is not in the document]
pcarruscag: That is not possible, the size of std::vector is assumed not to be known during compilation so you cannot get size related compiler errors :)
What you found is a bug in the MPI Wrapper, I'll push the fix directly to this branch, just a sec.

Common/include/mpi_structure.inl
```diff
@@ -612,7 +612,7 @@ inline void CBaseMPIWrapper::Scatter(void *sendbuf, int sendcnt, Datatype sendty
 }
 
 inline void CBaseMPIWrapper::Allgatherv(void *sendbuf, int sendcnt, Datatype sendtype,
-                                   void *recvbuf, int recvcnt, int *displs, Datatype recvtype, Comm comm){
+                                   void *recvbuf, int *recvcnt, int *displs, Datatype recvtype, Comm comm){
```
pcarruscag, line 615: This was the problem, the signature was incorrect, I've had some issues (weird segfaults) in the past with AD and allgatherv by the way... not because of this problem, and maybe it was my fault... But anyway just something to keep in the back of your mind if it causes you problems too.

koodlyakshay:
Thanks for the fix @pcarruscag.

koodlyakshay:
Can I merge this one?

Lucas-Lucatelli:
Hello everyone. I ran simulations with the file “turb_flatplate_v7_cfg.txt”, but the results obtained do not converge with those presented in Blanchard's Figure of numerical results and experiments. I also ran simulations with different values of speed and roughness, but the results did not converge either. Could you please help me? Thank you!

PR #881, opened 2020-02-17
Solver Factory class

talbring:
## Proposed Changes
In order to structure the construction of the solvers a little bit and to remove the dependency of the driver on the solver, I created a static solver factory class. There are routines to create individual instances as well as the whole container. It should be much easier now to see what position in the containers are actually allocated for the different solvers.

The `createSolver` routine is a generic one that simply calls `new`. The others have some additional checks or calls for the construction. The idea here was to remove any `if` statements in the `createSolverContainer` routine to ease readability. 

If everybody agrees, we could use a similar structure for other classes as well. Let me know what you think.

## Related Work

## PR Checklist


pcarruscag:
LGTM what other families to you have in mind? I think doing this for the numerics would be very nice but they should move to the solver.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -1034,13 +1034,13 @@ void CNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container, C
...
-  solver_container[FLOW_SOL]->GetNodes()->SetVorticity_StrainMag();
+  GetNodes()->SetVorticity_StrainMag();
 
   StrainMag_Max = 0.0; Omega_Max = 0.0;
   for (iPoint = 0; iPoint < nPoint; iPoint++) {
 
-    StrainMag = solver_container[FLOW_SOL]->GetNodes()->GetStrainMag(iPoint);
-    Vorticity = solver_container[FLOW_SOL]->GetNodes()->GetVorticity(iPoint);
+    StrainMag = GetNodes()->GetStrainMag(iPoint);
+    Vorticity = GetNodes()->GetVorticity(iPoint);
```
pcarruscag: Good catch here, but not even GetNodes() is needed, just nodes-> would do the trick.
Can you revert changes to this file and CEulerSolver? I'll put these fixes in #861 where these routines were modified.

talbring: I'll have to change it here, as `solver_container[FLOW_SOL]` is not allocated yet when I call it in the factory methods. I am gonna change it to `nodes` though. Git should not show a conflict if the changes are exactly the same.

talbring:
Since I am already at it, I am gonna add factory classes for numerics (we can use them even if we move the allocation to the solvers) and integration.

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -404,7 +404,7 @@ bool CMultizoneDriver::OuterConvergence(unsigned long OuterIter) {
     /*--- Account for all the solvers ---*/
 
     for (unsigned short iSol = 0; iSol < MAX_SOLS; iSol++){
-      if (solver_container[iZone][INST_0][MESH_0][iSol] != NULL){
+      if (solver_container[iZone][INST_0][MESH_0][iSol] != nullptr){
```
economon: I like the switch to nullptr

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -41,23 +41,8 @@
 
 #include "../../../Common/include/omp_structure.hpp"
 
-#include "../../include/solvers/CEulerSolver.hpp"
-#include "../../include/solvers/CIncEulerSolver.hpp"
-#include "../../include/solvers/CNSSolver.hpp"
-#include "../../include/solvers/CIncNSSolver.hpp"
-#include "../../include/solvers/CTurbSASolver.hpp"
```
economon: Getting all of the headers out of CDriver is great 👍

SU2_CFD/src/integration/CIntegrationFactory.cpp
```diff
@@ -0,0 +1,183 @@
...
+CIntegration** CIntegrationFactory::createIntegrationContainer(ENUM_MAIN_SOLVER kindMainSolver, CConfig *config){
+
+  CIntegration **integration = new CIntegration* [MAX_SOLS]();
+
+  switch (kindMainSolver) {
+    case TEMPLATE_SOLVER:
+      integration[FLOW_SOL] = createIntegration(SUB_SOLVER::TEMPLATE, config);
+      break;
+    case INC_EULER:
+      integration[FLOW_SOL] = createIntegration(SUB_SOLVER::INC_EULER, config);
```
pcarruscag: This looks nice @talbring :+1:, but, what if, you create an enum for the type of integration instead and add a method to CSolver that returns the required type of integration? This factory would then just be a loop over the non-null entries of the solver container with a switch for the type of integration.

SU2_CFD/include/integration/CIntegrationFactory.hpp
```diff
@@ -0,0 +1,59 @@
...
+  CIntegrationFactory();
+
+public:
+
+  /*!
+   * \brief Create the integration container based on the current main solver
+   * \param[in] kindSolver    - The kind of main solver
+   * \return                  - Pointer to the allocated integration container
+   */
+  static CIntegration** createIntegrationContainer(ENUM_MAIN_SOLVER kindSolver, CSolver **solver_container);
```
TobiKattmann: missing `solver_container`

SU2_CFD/src/integration/CIntegrationFactory.cpp
talbring: Yep good idea! I did something similar. Instead of putting it into the solver, the solver factory now creates some meta data for every solver. So far it only stores the type of the solver and the integration method, but the ultimate goal would be that it stores also the kind of numerics that should be allocated and so on. This way there is only one place to edit in case someone adds a new solver.

SU2_CFD/include/solvers/CSolverFactory.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+class CSolver;
+class CGeometry;
+class CConfig;
+
+class CSolverFactory {
+  
+private:
+
+  static map<CSolver*, SolverMetaData> allocatedSolvers;
```
pcarruscag: Cool! I like this even more, only one place to edit, can you make the key (CSolver*) const? Just to make sure this does not get misused.
Not so sure on the numerics idea though, isn't that going to make the solver factory too complex? (there are a lot more numerics options per solver than there are integration options).

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1229,7 +1229,7 @@ void CEulerSolver::Set_MPI_ActDisk(CSolver **solver_container, CGeometry *geomet
             for (iVar = 0; iVar < nPrimVar; iVar++) {
               Buffer_Send_PrimVar[(nPrimVar_)*(PointTotal_Counter+iPointTotal)+iVar] = nodes->GetPrimitive(iPoint,iVar);
             }
-            if (rans) {
+            if (rans && (solver_container[TURB_SOL] != nullptr)) {
```
pcarruscag: You can remove this null check, I put it in the definition of the boolean

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1363,160 +1402,30 @@ void CDriver::Solver_Postprocessing(CSolver ****solver, CGeometry **geometry,
...
   delete [] solver[val_iInst];
+
+  CSolverFactory::ClearSolverMeta();
   
 }
 
-void CDriver::Integration_Preprocessing(CConfig *config, CIntegration **&integration) {
-
-  unsigned short iSol;
+void CDriver::Integration_Preprocessing(CConfig *config, CSolver **solver, CIntegration **&integration) {
```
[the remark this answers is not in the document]
talbring: I will leave it like that for the moment. We can consider a clean-up once we have factory methods for the rest of the containers (especially numerics).

SU2_CFD/include/solvers/CSolverFactory.hpp
talbring: Yes that might mean that it gets more complex. But then everything the solver needs is determined at one place. I kind of like that idea.

SU2_CFD/include/integration/CIntegrationFactory.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+public:
+
+  /*!
+   * \brief Create the integration container based on the current main solver
+   * \param[in] kindSolver    - The kind of main solver
+   * \param[in] solver        - The solver container
+   * \return                  - Pointer to the allocated integration container
+   */
+  static CIntegration** createIntegrationContainer(ENUM_MAIN_SOLVER kindSolver, CSolver **solver_container);
```
pcarruscag: const const solver_container please.

Ok, why not just delegate the numerics creation to the solvers then? That's pretty much what it boils down to, and it would be one step in the direction of moving the numerics into the solvers.

SU2_CFD/include/output/COutputFactory.hpp
```diff
@@ -0,0 +1,67 @@
...
+  COutputFactory();
+
+public:
+
+  /*!
+   * \brief Create the Output class based on the current main solver
+   * \param[in] kindSolver       - The kind of main solver
+   * \return                     - Pointer to the allocated Output
+   */
+  static COutput* createOutput(ENUM_MAIN_SOLVER kindSolver, CConfig *config, int nDim);
```
TobiKattmann: missing `param[in]`

SU2_CFD/include/integration/CIntegrationFactory.hpp
talbring: I am gonna create a separate PR to deal with the numerics.

SU2_CFD/include/output/COutputFactory.hpp
```diff
@@ -0,0 +1,69 @@
...
+class COutputLegacy;
+
+class COutputFactory{
+
+private:
+
+  /*!
+   * \brief Private constructor to avoid creating instances of this class
+   */
+  COutputFactory();
```
pcarruscag: Thanks for the const stuff, everything looks very nice.
I think it is a good idea to have a separate numerics PR.

Final note, the preferred C++11 way to achieve this is to delete the default constructor instead of making it private.

talbring: Perfect, thanks! If that's the preferred way then we should do it :+1:

pcarruscag: At least according to Scott Meyers xD

TobiKattmann:
As all comments are addressed, I am merging this in. Thanks @talbring for the effort 💐

PR #900, opened 2020-03-04
Feature restart

ScSteffen:
## Proposed Changes
Enhanced the script discrete_adjoint.py such that in unsteady simulations
- one can execute an "only adjoint" run. This requires the solution files of a direct run (option -a True, default is -a False).
- it supports restart solutions. Configure the restart iteration in the config-file as one would in a restartet direct simulation. The script then runs the direct iteration until final time has been reached. Then the adjoint run is performed from final time up to the restart iteration.   
 

## Related Work
None

## PR Checklist


SU2_CFD/src/output/tools/CWindowingTools.cpp
```diff
@@ -1,8 +1,9 @@
 /*!
  * \file signal_processing_toolbox.cpp
  * \brief Signal processing tools
- * \author T. Albring
+ * \author S. Schotthöfer
  * \version 7.0.2 "Blackbird"
+>>>>>>> d1118950d691d9d01401e3d5c7460d23a654f588
```
economon: Looks like an old merge conflict marker here.. please remove

SU2_PY/discrete_adjoint.py
```diff
@@ -101,7 +106,7 @@ def discrete_adjoint( filename           ,
     konfig = copy.deepcopy(config)
 
     # Direct Solution
-    if compute:
+    if compute and not onlyadjoint:
```
economon: It seems like these two options should be incompatible. I see three exclusive modes: compute primal & adjoint & gradient, compute adjoint (with primal restart) & gradient, or compute gradient only (with primal and adjoint restarts.

Can you please add an error check or combine these into a single MODE option, perhaps?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1403,7 +1403,7 @@ void COutput::CheckHistoryOutput(){
   }
 
   if (rank == MASTER_NODE){
-    if(wndConvFields.size() == 0){
+    if(convFields.size() == 0){
```
pcarruscag: `convFields.empty()` always prefer text to numbers, makes the code more expressive.

TestCases/disc_adj_rans/naca0012/naca0012.cfg
```diff
@@ -0,0 +1,247 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: NACA0012 Laminar simulation (Re 5000)                      %
+% Author: Carlos Ilario                                                        %
+% Institution: Stanford University                                             %
+% Date: Mar 16, 2017                                                           %
+% File Version 5.0.0 "Raven"                                                   %
+%                                                                              %
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
pcarruscag: Please revise the description which seems incorrect, and try to highlight the features that this testcase exercises, both in its name and in the config file (by providing extra description or grouping the important stuff at the top). I also recommend trying to keep the config as clean as possible, if some options are not needed (e.g. RK coefficients) remove them as that can generate confusion for new users.

SU2_CFD/src/output/tools/CWindowingTools.cpp
ScSteffen: Done :)

SU2_CFD/src/output/COutput.cpp
ScSteffen: Okay, done :)

TestCases/disc_adj_rans/naca0012/naca0012.cfg
ScSteffen: Okay yes, I corrected the misleading description of the test case and removed unnecessary options.

SU2_PY/discrete_adjoint.py
ScSteffen: I've implemented your suggestions :)

SU2_PY/discrete_adjoint.py
```diff
@@ -42,27 +42,30 @@ def main():
...
                       help="number of PARTITIONS", metavar="PARTITIONS")
-    parser.add_option("-c", "--compute",    dest="compute",    default="True",
-                      help="COMPUTE direct and adjoint problem", metavar="COMPUTE")
     parser.add_option("-s", "--step",       dest="step",       default=1E-4,
                       help="DOT finite difference STEP", metavar="STEP")
     parser.add_option("-v", "--validate", dest="validate", default="False",
                       help="Validate the gradient using direct diff. mode", metavar="VALIDATION")
     parser.add_option("-z", "--zones", dest="nzones", default="1",
                       help="Number of Zones", metavar="ZONES")
+    parser.add_option("-m", "--mode", dest="mode", default="0",
```
economon: I would suggest using a string name instead of an int for the mode options, just to be more readable

talbring: Yep, that would be also my only remark! Otherwise looks good.

ScSteffen: Okay, I've updated it.

Issue #797, opened 2019-10-08
Pressure field near wall using SST 


emoralest6:
While doing some simulations with the version 6.2.0, I found a strange behaviour on the pressure field.
The test case is a naca0012 at AoA=0°, M=0.7, Re=9e6.
When running with SST turbulence model, the pressure field in the "boundary layer region" is not constant. Using SA turbulence model, this behaviour is not shown. It looks like the tubulent kinetic energy field affect the computation of pressure field, leading to a non zero gradient pressure at wall, as shown in the attached figures:

In the figures it can be seen a comparison between the in-house solver of the Italian Aerospace Research Center (CIRA), where SST model is implemented, and SU2 using both available turbulence models, SST and SA. The isolines represent the pressure field and the contour is the TKE field obtained with SU2 with SST. In particular, solid red line is CIRA solver, the solid black line is SU2-SST, and dashed black line is SU2-SA.

Near the LE where TKE is 0, the isolines are similar and 0 pressure gradient at the wall is obtained with both turbulence models. When TKE starts to grow, the pressure computed with SU2-SST shows a non constant behaviour and at the wall has non 0 gradient. Furthermore,  in the wake region the defect is more evident.

Leading edge detail:
![cp_tke_detail_Le](https://user-images.githubusercontent.com/36844965/66407922-53448b00-e9ee-11e9-877d-88ace3a222fb.png)

Detail at 30% of the chord length:
![cp_tke_detail_1](https://user-images.githubusercontent.com/36844965/66407921-53448b00-e9ee-11e9-8a22-87106066984b.png)

Wake region detail:
![cp_tke_detail_wake](https://user-images.githubusercontent.com/36844965/66407920-53448b00-e9ee-11e9-92ad-b770e0ac5743.png)

Has anyone found previously this behaviour? Could it depend on the set up  that I have used? 

I attach the grid and configurations files.
[grid_and_conf_files.zip](https://github.com/su2code/SU2/files/3703135/grid_and_conf_files.zip)

Thanks in advance.

economon:
Thanks for reporting this, @emoralest6. My first suggestion is that we go through the mean flow solver and double-check our inclusion of the -2/3*rho*k term from the stress tensor (and TKE correction for energy) to make sure we do not have a bug there (see https://turbmodels.larc.nasa.gov/sst.html). This is one of the primary differences between the SA and SST implementations and would affect pressure. Could be a good place to start.

emoralest6:
[...]

Issue #770, opened 2019-08-23
Cleanup bit flags


pcarruscag:
We have quite a few common booleans we use in many places of the code.
Some of them are very simple, like checking for compressible vs incompressible, or for FSI.
But when we want to express something more complicated like is it SA or SST, is the mesh moving or deforming, is this FEM or FVM or FEA, and so on, things are not so simple anymore and we end up with long conditionals repeated in many places.

Since that logic is all built up from data in config I propose that we create a custom type akin to a bit-field that we can get from config and query whatever flag we want.
We move all the logic of setting up / updating this object to the config so that for example this:
```
  unsigned short Kind_Solver = config->GetKind_Solver();
  bool flow = ((Kind_Solver == DISC_ADJ_EULER)          ||
               (Kind_Solver == DISC_ADJ_RANS)           ||
               (Kind_Solver == DISC_ADJ_NAVIER_STOKES)  ||
               (Kind_Solver == ADJ_EULER)               ||
               (Kind_Solver == ADJ_NAVIER_STOKES)       ||
               (Kind_Solver == ADJ_RANS));
```
Can become `bool flow = config->GetFlags().flow();`, or similar.

If we move this kind of logic to a separate object we can possibly make the config lighter, not so many methods etc.
This flags class can be absolutely tiny if the bits are packed (8 times tinier than storing bools).

pcarruscag:
WIP

pcarruscag:
wip

pcarruscag:
This is slowly getting done, I've moved a few KindSolver-based boolean to CConfig, e.g. `GetFluidProblem`.

PR #905, opened 2020-03-06
Fix for turbulent viscosity calculation in SST Solver

jayantmukho:
## Proposed Changes
In a previous PR the SST model implementation in the CTurbSSTSolver was corrected to use the vorticity magnitude to calculate the turbulent viscosity (mu_T). But this change was not copied in the relevant numerics class where mu_T is used to calculate production terms. This is probably going to break all the SST related regression tests, so I will ensure convergence of those cases and update regression values. 
 
## Related Work
A relevant discussion to have here is whether to update the SST turbulence model according to a 2003 paper by Menter (author of the SST model): [Ten years of industrial experience with the SST turbulence model](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.460.2814&rep=rep1&type=pdf)

It involves tiny changes to coefficients, production limits, and uses strain magnitude instead of the vorticity magnitude to calculate mu_T. The last point means reversing the change that this PR is making. But currently the model is just inconsistent and needs to be corrected. 

The changes are supposed to make the model more stable, which has been a cause for concern with the SST model before. Not sure if people have thoughts about altering the original turbulence model. 

## PR Checklist

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -89,5 +89,5 @@ jobs:
       - name: Run Tests in Container
         uses: docker://su2code/test-su2:20200303
         with:
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c fix_sst_term -s ${{matrix.testscript}}
   
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -89,5 +89,5 @@ jobs:
       - name: Run Tests in Container
         uses: docker://su2code/test-su2:20200303
         with:
-          args: -b ${{github.ref}} -t develop -c fix_sst_term -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
   
--- a/SU2_CFD/src/numerics/turbulent/turb_sources.cpp
+++ b/SU2_CFD/src/numerics/turbulent/turb_sources.cpp
@@ -796,6 +796,9 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSST::ComputeResidual(const CConfi
   unsigned short iDim;
   su2double alfa_blended, beta_blended;
   su2double diverg, pk, pw, zeta;
+  su2double VorticityMag = sqrt(Vorticity_i[0]*Vorticity_i[0] +
+                                Vorticity_i[1]*Vorticity_i[1] +
+                                Vorticity_i[2]*Vorticity_i[2]);
 
   if (incompressible) {
     AD::SetPreaccIn(V_i, nDim+6);
@@ -846,12 +849,12 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSST::ComputeResidual(const CConfi
    pk = min(pk,20.0*beta_star*Density_i*TurbVar_i[1]*TurbVar_i[0]);
    pk = max(pk,0.0);
 
-   zeta = max(TurbVar_i[1], StrainMag_i*F2_i/a1);
+   zeta = max(TurbVar_i[1], VorticityMag*F2_i/a1);
 
    /* if using UQ methodolgy, calculate production using perturbed Reynolds stress matrix */
 
    if (using_uq){
-    pw = PerturbedStrainMag * PerturbedStrainMag - 2.0/3.0*zeta*diverg;
+     pw = PerturbedStrainMag * PerturbedStrainMag - 2.0/3.0*zeta*diverg;
    }
    else {
      pw = StrainMag_i*StrainMag_i - 2.0/3.0*zeta*diverg;
--- a/TestCases/disc_adj_rans/naca0012/turb_NACA0012_sst.cfg
+++ b/TestCases/disc_adj_rans/naca0012/turb_NACA0012_sst.cfg
@@ -87,11 +87,11 @@ CFL_NUMBER= 10.0
 MAX_DELTA_TIME= 1E10
 %
 % Adaptive CFL number (NO, YES)
-CFL_ADAPT= NO
+CFL_ADAPT= YES
 %
 % Parameters of the adaptive CFL number (factor down, factor up, CFL min value,
 %                                        CFL max value )
-CFL_ADAPT_PARAM= ( 1.5, 0.5, 1.0, 100.0 )
+CFL_ADAPT_PARAM= ( 0.1, 1.2, 1.0, 100.0 )
 %
 % Number of total iterations
 ITER= 10000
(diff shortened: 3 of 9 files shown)


bmunguia:
@jayantmukho LGTM. I'm also in favor of moving to SST-2003 if it means a more robust SST solver.

This might also be a good time to continue to look into other issues with SST, e.g. some TKE terms mentioned in #797.

Another possible discrepancy I noticed is that in CTurbSSTVariable we have
```
CDkw(iPoint) = 0.0;
for (unsigned long iDim = 0; iDim < nDim; iDim++)
    CDkw(iPoint) += Gradient(iPoint,0,iDim)*Gradient(iPoint,1,iDim);
CDkw(iPoint) *= 2.0*val_density*sigma_om2/Solution(iPoint,1);
CDkw(iPoint) = max(CDkw(iPoint), pow(10.0, -20.0));
...
arg1 = min(arg2, 4.0*val_density*sigma_om2*Solution(iPoint,0) / (CDkw(iPoint)*val_dist*val_dist+EPS*EPS));
F1(iPoint) = tanh(pow(arg1, 4.0));
```
which is consistent with TMR, but when computing the cross diffusion residual we have
```
val_residual[1] += (1.0 - F1_i)*CDkw_i*Volume;
```
However, here TMR (as well as both SST-1993 and SST-2003) has
Screen Shot 2020-03-06 at 4 44 14 PM
So to be consistent with TMR, would it make more sense to have e.g.
```
CDkw(iPoint) = 0.0;
for (unsigned long iDim = 0; iDim < nDim; iDim++)
    CDkw(iPoint) += Gradient(iPoint,0,iDim)*Gradient(iPoint,1,iDim);
CDkw(iPoint) *= 2.0*val_density*sigma_om2/Solution(iPoint,1);
...
arg1 = min(arg2, 4.0*val_density*sigma_om2*Solution(iPoint,0) / (max(CDkw(iPoint),pow(10.0,-10.0))*val_dist*val_dist+EPS*EPS));
F1(iPoint) = tanh(pow(arg1, 4.0));
```
so we don't clip the cross diffusion residual? Not sure how much of a difference this makes (or other discrepenacies like the TKE issue).

pcarruscag:
Just a couple of non scientific comments.
- The release after this kind of change should get more than a minor version increase, so if the changes are localized it may be worth waiting and bundling them with other significant ones (e.g. removing some legacy features).
- If you guys add more SST variants let's please do it in a smarter way than `case SST: case SST_2003: ...` as we are still seeing bug fixes from the last time we changed that kind of enum... I'm happy to help implement something like #770.

jayantmukho:
Sorry took a while to get around to it but here are the results from the validation cases that were requested. Turns out, this change is minimal in its effect to the overall simulation (at least for the two cases I have run)


## Flatplate
For the flatplate case, there was no difference in the C_D calculation so the lines mesh convergence were on top of each other. But while I was doing that, I was playing around with the solver parameters and there was an interesting difference between simulations run with `WEIGHTED_LEAST_SQUARES` and those run with `GREEN_GAUSS`. I have included them here:

force_convergence_flatplate_SA
force_convergence_flatplate_SST

This shows interesting parallels between `WEIGHTED_LEAST_SQUARES` results and CFL3D results vs `GREEN_GAUSS` and FUN3D. Also interesting the SU2 with both gradient methods converges to a slightly different C_f value than CFL3D and FUN3D. Not really connected with this PR but interesting results that can be uploaded to the VandV section of the website. 

## RAE 2822

Luckily I have been working on the RAE2822 as a validation case so I already had a family of grids that worked for this case. Again seems like the change didn't make much of a difference. dev_SST refers to the current SST implementation in develop. SST refers to the results from this PR

force_convergence_rae2822_SST

Consequently, the C_P plots are also identical
cp_rae2822_SST

clarkpede:
Ok.  Unless there's significant differences in any of the other regression test cases, then this PR looks good to me.

economon:
Hmm.. you are right that the Cf results look a little different than the ones posted on the V&V page (for both SA and SST). Is there anything else different in the configuration for those cases from the set up for the cases on the V&V page? For the flat plate, I may have used unweighted LSQ for the posted V&V results.

jayantmukho:
@economon, from the config files in the V&V folder, it seems like you ran without adaptive CFL and without the unweighted LSQ.

The main difference between the V&V runs and these ones is that I don't use any slope limiter. In the V&V cases the `SLOPE_LIMITER_TURB= VENKATAKRISHNAN` option was used. 

There are some other differences: 
1. I use adaptive CFL
2. I use `REF_DIMENSIONALIZATION= FREESTREAM_PRESS_EQ_ONE`
3. Linear solver settings are slightly different. For the V&V runs the following are used: 
```
% Minimum error of the linear solver for implicit formulations
LINEAR_SOLVER_ERROR= 1E-15
%
% Max number of iterations of the linear solver for the implicit formulation
LINEAR_SOLVER_ITER= 25
```
For my runs, I used the settings you had suggested in #790 : 
```
% Minimum error of the linear solver for implicit formulations
LINEAR_SOLVER_ERROR= 1E-10
%
% Max number of iterations of the linear solver for the implicit formulation
LINEAR_SOLVER_ITER= 10
```
4. For the SST runs, I did not specify these options that you used in the V&V case: 
```
% Free-stream turbulence intensity for the SST model
< FREESTREAM_TURBULENCEINTENSITY = 0.00038729
< %
< % Free-strem turbulence to laminar viscosity ratio for the SST model
< FREESTREAM_TURB2LAMVISCRATIO   = 0.009
```
I have attached the config files I used. I am still working on confirming all the regression tests, everything seems to be reasonable thus far. 

[flatplate_configs.zip](https://github.com/su2code/SU2/files/4336003/flatplate_configs.zip)

jayantmukho:
Updated all the regression test values. Ran all of them to convergence and the difference in the integrated quantities (lift, drag, sens_press, entropy generation etc) were < 1%. 

While running the regression tests, I saw that some of them weren't up to date with the best practices from the recent PRs. I know they are for Regressions and not necessarily as example configs, but I think it'd be good to have the configs reflect best practices so that people can look to them for reference. 

If people are on board with that, I can update some of the rans configs to include things like ILU, adaptive CFL, the new convergence monitoring systems etc.  I will make sure that the tests still cover all the relevant config options. I will also make sure that if you were to run the test individually, it would run to convergence.

pcarruscag:
I think coverage should take priority over best practices.

.github/workflows/regression.yml
```diff
@@ -89,5 +89,5 @@ jobs:
       - name: Run Tests in Container
         uses: docker://su2code/test-su2:20200303
         with:
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c fix_sst_term -s ${{matrix.testscript}}
```
talbring: Remember to change that

Issue #894, opened 2020-02-28
SU2 v7 is slower on Ethernet cluster


drewkett:
**Describe the bug**
*A clear and concise description of what the bug is and what you expect the behavior to be instead. If applicable, add screenshots to help explain your problem.*

I've used SU2 off and on for a long time. I fairly recently set up a 4 machine cluster (16 cores each) to run CFD problems on. The cluster is connected together with 1G ethernet. On SU2 v6.2 and earlier, when I ran a problem on multiple machines the time per iteration would improve up until i hit a limit where presumably network overhead exceeded the benefit of adding another machine (The limit for my set up was something like 0.5 s/iteration. 

I just tried SU2 v7 on this same cluster and consistently the time per iteration gets worse when I add a second machine and double the process count.  I've tried this on smaller problems (~1Mill cells) but I've also tried it on larger ones (~27Million cells). On the larger problem, one machine with 16 processes ran at 34s/iteration and 2 machines with 32 processes also ran at 34s/iteration. For that problem size, I would expect a near 2x improvement in iteration time when doubling processes.

Other than this, the apples to apples comparison between SU2 v6 and v7 is otherwise very impressive where iteration times have come down even for higher CFL's.

I did compile SU2 from source. I tried to compile it the same way for both v6.2 and v7 though with the different build system its a little bit harder to be certain that its all exactly the same. Given that v7 performs fine going up to 16 processes on one machine, I don't think that its a compilation issue, but admittedly I don't know all the intricacies of mpi. 

Since its just multi machine that's an issue, I wonder if there's something about v7 that make it more bandwidth/latency dependent that would go away with a faster network. Or is this just a regression do to (compiler version, mpi version, a network misconfiguration...)?

Anyway I know this a bit rambling, but any help would be appreciated. I'm happy to test anything on my setup this if anyone has any ideas.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - C++ compiler and version: g++ 7.4.0
 - MPI implementation and version: OpenMPI 2.1.1
 - SU2 Version: 7.0.1

pcarruscag:
There was a substantial change going from 6.2 to 7.0 affecting the MPI (PR #652).
I have no problem running on Infiniband clusters.
I guess it would be worth updating to a newer MPI version.

drewkett:
Thanks for the response and pointing me to the PR

I updated gcc version 9 and open mpi to version 4.02. It slightly improved speed but the behavior is still the same. 

[...]

PR #908, opened 2020-03-11
Hybrid parallel coloring fallback strategies (better strong scaling and user friendliness)

pcarruscag:
## Proposed Changes
The hybrid parallel implementation uses grid coloring (edge and element) to avoid race conditions when building the residual vector and Jacobian matrix.
Depending on grid size, type, and ordering, it may be difficult to find a good coloring, i.e. providing locality and enough parallelism (all threads getting almost the same number of work chunks to work on per color).

This PR extends the use of the "reducer strategy" to fine grids (introduced and described in #861 for coarse grids, which before #894 I thought would be the only problematic scenarios) as a fallback strategy that is automatically selected when the coloring efficiency (highlighted below) drops below a threshold value. The reducer strategy can be forced by setting EDGE_COLORING_GROUP_SIZE to 0 (for some cases it seems to give better performance, contrary to the initial assessment from #789).
For element loops (CFEASolver and CMeshSolver, modified in #843) an automatic fallback was also introduced, but using "locks" (explained below) as a "reducer strategy" would use too much memory, and element loops are more compute intensive (and therefore less affected by the overhead of using these locks).

## Related Work
Part of #789, namely:
Improves scaling of the work from #861 #843 (should help with #894)
Fixes a bug from #830

## PR Checklist


Common/include/toolboxes/C1DInterpolation.hpp
```diff
@@ -84,11 +84,6 @@ class CAkimaInterpolation final: public C1DInterpolation{
       SetSpline(X,Data);
   }
 
-  /*!
-   * \brief Destructor of the CAkimaInterpolation class.
-   */
-  ~CAkimaInterpolation(){}
-
```
pcarruscag: This was removed for old compiler compatibility.

Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -592,3 +628,23 @@ struct DummyGridColor
...
+{
+  using Index_t = typename SparsePattern::IndexType;
+
+  /*--- Ideally compute time is proportional to total work over number of threads. ---*/
+  su2double ideal = coloring.getNumNonZeros() / su2double(numThreads);
+
+  /*--- In practice the total work is quantized first by colors and then by chunks. ---*/
+  Index_t real = 0;
+  for(Index_t color = 0; color < coloring.getOuterSize(); ++color)
+    real += chunkSize * roundUpDiv(roundUpDiv(coloring.getNumNonZeros(color), chunkSize), numThreads);
```
pcarruscag: The computation of coloring efficiency is described here, it is just a simple heuristic.

Common/src/CConfig.cpp
```diff
@@ -3677,96 +3677,96 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
       for (iMarker = 0; iMarker < nMarker_Moving; iMarker++){
         for (iDim = 0; iDim < 3; iDim++){
           MarkerMotion_Origin[3*iMarker+iDim] = 0.0;
-    }
-  }
+        }
+      }
```
pcarruscag: A few indentation fixes here and there.

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -658,6 +661,7 @@ unsigned long CSysSolve<ScalarType>::Smoother_LinSolver(const CSysVector<ScalarT
    product (A_x), for the latter two this is done only on the first call to the method. ---*/
 
   if (!smooth_ready) {
+    SU2_OMP_BARRIER
     SU2_OMP_MASTER
     {
       auto nVar = b.GetNVar();
```
pcarruscag: Bug fix relating to #830, sometimes the code would hang on the first iteration because of this:
```c++
if (!smooth_ready) { // if working variables have not been allocated we need to allocate...
  SU2_OMP_BARRIER // <- This is the fix.
  SU2_OMP_MASTER // ...only the master thread can do allocation as those variables are shared
  {
    // allocate variables
    ...
    smooth_ready = true; // set this flag so we don't allocate on the next call.
  }
  SU2_OMP_BARRIER // now we need a barrier so all threads "see" the new state of the variables.
  // but because we are "smart" the barrier is inside the "if" so that we don't hit it every time.
  // But if some threads arrive (very) late to the party they might see smooth_ready == true already
  // and they miss the barrier which results in a deadlock (all threads must arrive before execution
  // can continue). One solution is to make sure all threads are inside the "if" (via another barrier)
  // before changing the condition deciding whether that statement is executed.
}
```

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -919,7 +923,7 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
         /*--- Enforce a hard limit on total number of iterations ---*/
         unsigned long IterLimit = min(RestartIter, MaxIter-IterLinSol);
         IterLinSol += FGMRES_LinSolver(*LinSysRes_ptr, *LinSysSol_ptr, mat_vec, *precond, SolverTol, IterLimit, residual, ScreenOutput, config);
-        if ( residual < SolverTol*norm0 ) break;
+        if ( residual <= SolverTol*norm0 ) break;
```
pcarruscag: Small fix, restarted FGMRES was going into an infinite loop when the RHS was zero (happens in mesh deformation cases).

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -264,16 +264,18 @@ class CEulerSolver : public CSolver {
 
 #ifdef HAVE_OMP
   vector<GridColor<> > EdgeColoring;   /*!< \brief Edge colors. */
+  bool ReducerStrategy = false;        /*!< \brief If the reducer strategy is in use. */
 #else
   array<DummyGridColor<>,1> EdgeColoring;
+  /*--- Never use the reducer strategy if compiling for MPI-only. ---*/
+  static constexpr bool ReducerStrategy = false;
```
pcarruscag: The idea of `constexpr` here is to allow associated conditional statements to be optimized away by the compiler.

SU2_CFD/include/solvers/CTurbSolver.hpp
```diff
@@ -79,6 +84,28 @@ class CTurbSolver : public CSolver {
...
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void Viscous_Residual(unsigned long iEdge,
+                        CGeometry *geometry,
+                        CSolver **solver_container,
+                        CNumerics *numerics,
+                        CConfig *config);
```
pcarruscag: I fused the viscous and convective residual loops of the turbulent solvers in the same way I did for CEuler and CNSSolver. This makes the reducer strategy more cost-effective as we do not need to clear the Jacobian matrix and rhs vector during preprocessing (we overwrite instead).

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2586,52 +2586,50 @@ void CDriver::Interface_Preprocessing(CConfig **config, CSolver***** solver, CGe
         /*--- On the target side ---*/
         nMarkerTarget = config[targetZone]->GetnMarker_All();
 
-      for (iMarkerTarget = 0; iMarkerTarget < nMarkerTarget; iMarkerTarget++) {
+        for (iMarkerTarget = 0; iMarkerTarget < nMarkerTarget; iMarkerTarget++) {
```
pcarruscag: more indentation fixes

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2489,22 +2518,18 @@ void CEulerSolver::SetInitialCondition(CGeometry **geometry, CSolver ***solver_c
 
 }
 
-void CEulerSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iMesh,
-                                 unsigned short iRKStep, unsigned short RunTime_EqSystem, bool Output) {
+void CEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iMesh,
+                                       unsigned short iRKStep, unsigned short RunTime_EqSystem, bool Output) {
```
pcarruscag: I've done some cleanup by factoring out common (to both Euler and NS solvers) preprocessing steps.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2605,9 +2609,51 @@ void CEulerSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
...
-  /*--- Initialize the Jacobian matrices ---*/
+  /*--- Initialize the Jacobian matrix and residual, not needed for the reducer strategy
+   *    as we set blocks (including diagonal ones) and completely overwrite. ---*/
 
-  if (implicit && !disc_adjoint) Jacobian.SetValZero();
+  if(!ReducerStrategy && !Output) {
+    LinSysRes.SetValZero();
+    if (implicit && !disc_adjoint) Jacobian.SetValZero();
+    else {SU2_OMP_BARRIER} // because of "nowait" in LinSysRes
+  }
```
pcarruscag: The `Jacobian.SetValZero();` bit was previously not guarded by `!Output` which was a bit wasteful.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3632,142 +3664,82 @@ void CEulerSolver::SetMax_Eigenvalue(CGeometry *geometry, CConfig *config) {
 
 }
 
-void CEulerSolver::SetUndivided_Laplacian(CGeometry *geometry, CConfig *config) {
+void CEulerSolver::SetUndivided_Laplacian_And_Centered_Dissipation_Sensor(CGeometry *geometry, CConfig *config) {
```
pcarruscag: These routines were transformed into point loops (no need for grid coloring) and fused since they are always called together. (The code is actually much simpler especially w.r.t. the `boundary_i, boundary_j` conditionals)

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -829,7 +862,7 @@ void CFEASolver::Compute_StiffMatrix(CGeometry *geometry, CNumerics **numerics,
     for(auto color : ElemColoring) {
 
       /*--- Chunk size is at least OMP_MIN_SIZE and a multiple of the color group size. ---*/
-      SU2_OMP_FOR_DYN(roundUpDiv(OMP_MIN_SIZE, ColorGroupSize)*ColorGroupSize)
+      SU2_OMP_FOR_DYN(nextMultiple(OMP_MIN_SIZE, color.groupSize))
```
pcarruscag: In a number of places I did some replacing to make this operation more expressive.

config_template.cfg
```diff
@@ -1333,6 +1333,26 @@ UQ_URLX= 0.1
...
+%
+% Independent "threads per MPI rank" setting for LU-SGS and ILU preconditioners.
+% For problems where time is spend mostly in the solution of linear systems (e.g. elasticity,
+% very high CFL central schemes), AND, if the memory bandwidth of the machine is saturated
+% (4 or more cores per memory channel) better performance (via a reduction in linear iterations)
+% may be possible by using a smaller value than that defined by the system or in the call to
+% SU2_CFD (via the -t/--threads option).
+% The default (0) means "same number of threads as for all else".
+LINEAR_SOLVER_PREC_THREADS= 0
+%
```
pcarruscag: New options introduced in #853 and #861 added to config_template.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -410,23 +388,37 @@ void CMeshSolver::SetWallDistance(CGeometry *geometry, CConfig *config) {
...
 void CMeshSolver::SetMesh_Stiffness(CGeometry **geometry, CNumerics **numerics, CConfig *config){
 
-  unsigned long iElem;
+  /*--- Use the config option as an upper bound on elasticity modulus.
+   *    For RANS meshes the range of element volume or wall distance is
+   *    very large and leads to an ill-conditioned stiffness matrix.
+   *    Absolute values of elasticity modulus are not important for
+   *    mesh deformation, since linear elasticity is used and all
+   *    boundary conditions are essential (Dirichlet). ---*/
+  const su2double maxE = config->GetDeform_ElasticityMod();
```
pcarruscag: @rsanfer I introduced a small change to the mesh elasticity, the config option above is used as an upper bound for elasticity modulus when using wall-distance or volume based mesh stiffness (for the reason in the comment).
When using constant stiffness the option is ignored and E is 1 everywhere, this change made no difference to the existing testcases.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -879,6 +912,8 @@ void CFEASolver::Compute_StiffMatrix(CGeometry *geometry, CNumerics **numerics,
         /*--- Update residual and stiffness matrix with contributions from the element. ---*/
         for (iNode = 0; iNode < nNodes; iNode++) {
 
+          if (LockStrategy) omp_set_lock(&UpdateLocks[indexNode[iNode]]);
```
pcarruscag: In the loops over the points associated with an element, that follow the computation of element stiffness matrices (for example) we protected against concurrent accesses by setting (or acquiring) a lock for the point before writing to shared memory locations. Once done we unset the lock.

Common/include/omp_structure.hpp
```diff
@@ -80,6 +80,19 @@ inline void omp_set_num_threads(int) { }
...
+ */
+struct omp_lock_t {};
+struct DummyVectorOfLocks {
+  omp_lock_t l;
+  inline omp_lock_t& operator[](int) {return l;}
+};
+inline void omp_init_lock(omp_lock_t*){}
+inline void omp_set_lock(omp_lock_t*){}
+inline void omp_unset_lock(omp_lock_t*){}
+inline void omp_destroy_lock(omp_lock_t*){}
```
pcarruscag: As usual we define "do-nothing" types and functions when compiling without hybrid parallel support to make compilation compatible without having to throw #ifdefs everywhere.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -92,6 +96,13 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
   const bool muscl = config->GetMUSCL_Turb();
   const bool limiter = (config->GetKind_SlopeLimit_Turb() != NO_LIMITER);
 
+  /*--- Only reconstruct flow variables if MUSCL is on for flow (requires upwind) and turbulence. ---*/
+  const bool musclFlow = config->GetMUSCL_Flow() && muscl &&
+                        (config->GetKind_ConvNumScheme_Flow() == SPACE_UPWIND);
+  /*--- Only consider flow limiters for cell-based limiters, edge-based would need to be recomputed. ---*/
+  const bool limiterFlow = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) &&
+                           (config->GetKind_SlopeLimit_Flow() != VAN_ALBADA_EDGE);
```
pcarruscag: Since we are introducing turbulence-related fixes in #905, I wonder if it is a good time to revisit the reconstruction logic in the turbulent solver (which currently is not adequate for centered schemes), this only affects a couple of regressions.
@jayantmukho , @economon , @clarkpede what do you think?

SU2_CFD/src/solvers/CEulerSolver.cpp
talbring: :+1:

Common/src/linear_algebra/CSysSolve.cpp
jayantmukho: Oh! I think I did see this happen a couple of times when I briefly tested the multi-threading. I built SU2 with `-Dwith-omp=true`

It happened when I didn't specify a number of threads with the `-t` option. But it ran okay if I specified even `-t 1`. Does this fix that? Or do you have to specify a thread count if you build with OMP?

SU2_CFD/src/solvers/CTurbSolver.cpp
jayantmukho: Seems like you have revisited some of the logic here already? 

I haven't looked at the reconstructions before. Before these changes the MUSCL reconstruction for all (flow and turbulent) variables was only done when `MUSCL_TURB= YES` (when MUSCL reconstruction was requested for the turbulence solver as well). 

Here you have changed it such that the flow variables get reconstructed if `MUSCL_FLOW  = YES` but the turbulent variables are treated as before, i.e. they get reconstructed only when  `MUSCL_TURB= YES`. 

Are you suggesting changing the reconstruction itself? Or having an alternate reconstruction when centered schemes are used? Either way, I am for it, but don't know any better reconstruction schemes. Do you have any references I could check out? CFD wiki is best I've got so far :stuck_out_tongue:

config_template.cfg
jayantmukho: :+1:

SU2_CFD/src/solvers/CTurbSolver.cpp
jayantmukho: actually, I am surprised this didn't change more regression tests.

Common/src/linear_algebra/CSysSolve.cpp
pcarruscag: When you don't use option -t (--threads) SU2 picks up the number of threads from the environment, which often is the maximum. If you use MPI at the same time like this the system will probably be oversubscribed, I noticed that OpenMPI 3.1.4 on Ubuntu 16 deadlocks when this happens (oversubscribing with just threads does not seem to deadlock).
This fix was for a more subtle problem.

Another aspect of using MPI+threads is that you need to pay attention to the binding mode, some mpi will "--bind-to core" when the number of ranks is small, this then locks all threads spawn by each rank to the core. Always use "--bind-to socket/numa" as appropriate.

SU2_CFD/src/solvers/CTurbSolver.cpp
pcarruscag: Well this is all I think the logic needs. Central schemes never use MUSCL so it did not make sense to reconstruct the flow variables using gradients.
This only affects cases with central schemes + MUSCL turbulence (we seem to have two :) )

In theory what we need here is not a reconstruction, but a recomputation of the mass flux that is consistent with the main solver. I played with that once here #721 #726, by storing the mass flux from the convective residual loop. It made convergence worse since it un-staggers the solution (block Gauss-Seidel vs block Jacobi) so I did not pursue it further...

jayantmukho: Oh nevermind, I didn't see the `&& muscl` in the declaration of `musclFlow`. 

I am not sure why you would need a recomputation of the mass flux as opposed to flow variables. The reconstructions in the EulerSolver are identical to that in TurbSolver (except for things like the low mach corrections and non-physical points). I can see an argument made for copying all those corrections in the TurbSolver as well.

pcarruscag: If you want the mass fluxes to respect the discretized continuity equation then they should be consistent with what is computed by the flow solver. In any case this did not seem to make a huge difference... But it is the recommended approach especially for incompressible methods.

Issue #889, opened 2020-02-26
Turbomachinery adjoint objective functions


auzbaig:
What are the objective functions that are available with Turbomachinery for the adjoint solver? The test case "disk_adj_turbomachinery/transonic_stator_2D" does not work with the python script:

`$ shape_optimization.py -g DISCRETE_ADJOINT -f transonic_stator_2D.cfg`

and shows the following error:
```
  File "/home/baigan/SU2/bin/SU2/opt/project.py", line 129, in __init__
    group = historyOutFields[this_obj]['GROUP']
KeyError: 'ENTROPY_GENERATION'
```
If I change the objective function to DRAG then this error disappears but then it won't support the constraint for FLOW_ANGLE_OUT. Is ENTROPY_GENERATION supported for Turbomachinery or not?

I generated the FFD mesh using DV_KIND=FFD_SETTING and then changed to FFD_CONTROL_POINT_2D with the new mesh.

A discrete adjoint run of SU2_CFD_AD seems to converge nicely. But how can I manually go through the process of generating config files for SU2_DOT / SU2_DEF to get the next deformed shape and what is the sequence they must be run if the python script is not working?

auzbaig:
Hi , @economon  @talbring , @pcarruscag , have any of you tried running this case before? I also face similar problems for 3D cases. I don't know whether this is a bug or I am doing something wrong.

economon:
@auzbaig : sorry that this has been giving you issues. I think there are a couple of things happening here. First, there were important changes when all of the output was rewritten for v7, including a change in how the output fields are recognized by the python framework. In this commit (https://github.com/su2code/SU2/commit/39dbebc574a0886fa7c165d342dce8490947dbbe), a new history mapping capability was added, so that, rather than hard-coding the possible types of output/objective fields, the list can be updated automatically using updateHistoryMap.py, which parses the C++ files for available fields. You can see the available output fields for your problem by executing the dry run mode.

The second thing is that updateHistoryMap.py only parses a set list of files (you can find the list at the top of the file), and at the moment, the turbomachinery-specific outputs are not included. The turbomachinery features are being restructured currently in #866, which includes connecting it to the new output classes.

@talbring @salvovitale : please correct me if I am missing anything

auzbaig:
[...]

PR #913, opened 2020-03-19
Expression language for custom output fields

talbring:
## Proposed Changes
The is a first step in updating the output structure in order to make it a little bit more flexible and easier to maintain. The plan is to eventually have a consistent way of defining objective functions and/or define custom output quantities on individual surfaces.  

I introduced a new structure to store scalar quantities for the history output. This makes it much easier to filter and loop through a specific subset of fields. Next step is to have per surface values stored in the same structure. Furthermore, creating structure to compute problem-specific outputs like aerodynamic coefficients, time-averaged quantities, surface quantities and so on.

## PR Checklist


talbring:
I am currently testing an expression parser to define custom fields in the config file. For this I modified the handling of the curly brackets {}. They now declare a "literal" interpretation, which means everything inside is considered as one string and will not be split while strings are tokenized when reading the config file. 

When you declare the output fields you can then use expressions involving all available fields, e.g.

`SCREEN_OUTPUT= INNER_ITER, RMS_DENSITY, {sqrt(LIFT)/DRAG**2}`

At the moment only inline expression are possible. But I am considering also to use a file to define more complex functions. The goal is to make that framework more flexibel to that we can use it also to define custom integrals over markers, functions defined on the whole flow field, custom boundary conditions, inlet conditions and so on. 

Important to note is that I already differentiated the parser, which means everything that is defined can be used also for the adjoint solver.

SU2_CFD/src/output/CMultizoneOutput.cpp
```diff
@@ -87,55 +87,39 @@ CMultizoneOutput::~CMultizoneOutput() {}
...
-    for (iField = 0; iField < nField; iField++){
-
-      if (ZoneHistoryNames[iField] != "TIME_ITER" && ZoneHistoryNames[iField] != "OUTER_ITER"){
-
-        name   = ZoneHistoryNames[iField]+ "[" + PrintingToolbox::to_string(iZone) + "]";
-
-        SetHistoryOutputValue(name, ZoneHistoryFields[ZoneHistoryNames[iField]].value);
+    const COutFieldCollection& ZoneHistoryFields = output[iZone]->GetHistoryFieldsAll();
 
+    for (auto field : ZoneHistoryFields.GetReferencesAll()){
```
pcarruscag: I think auto type deduction looses "referenceness" i.e. you may want `auto& field` or `const auto& field` depending on the return type of GetReferencesAll

Common/include/toolboxes/CIndexedMap.hpp
```diff
@@ -0,0 +1,147 @@
...
+  void AddItem(const Key& key, const Item& item){
+    insertionVector.emplace_back(map.insert({key, item}).first);
+  }
+
+  /*!
+   * \brief Check if a specific key exists in the container
+   * \param[in] key - The key that should be checked
+   * \return <TRUE> then when the key could be found, otherwise <FALSE>
+   */
+  bool FindKey(const Key& key) const {
```
pcarruscag: Why not call it "CheckIsKey" or something like that? Find sounds like something more than a check

Common/include/toolboxes/CIndexedMap.hpp
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /*!
+   * \brief Get the item at a specific position
+   * \param[in] i - The position of the item
+   * \return The item at the specified position
+   */
+  const Item& GetItem(const int i) const {
+    return insertionVector[i]->second;
+  }
```
pcarruscag: The code will not compile for `Key = int`, and it will have very surprising behavior if e.g. `Key = unsigned long`, you may want to disambiguate the names e.g. GetItemByIndex, GetNthItem

talbring:
@pcarruscag Thanks. I already incorporated your comments!

I extended the expressions to the volume output fields. Similar to the `SCREEN_OUTPUT` you can now define custom volume field variables directly in the `VOLUME_OUTPUT` option, e.g,

`VOLUME_OUTPUT= COORDINATES, SOLUTION, {MOMENTUM_X/DENSITY}`

This required to remove all `-` in the field names for obvious reasons.

Note that at the moment it is required that all fields used in the expression are also added manually to the output. I will probably make that automatic soon.

talbring:
It is now possible to define more complex functions in a separate file called `functions.su2x`. The functions defined there are available in the inline expressions. You can either give the functions explicit arguments or you can use them without any arguments. All fields are available in the global scope. Which variables are available depends on the context in which the function is used in the code. 

functions.su2x:
```
function dragOverLift(){
  l = LIFT;
  d = DRAG;
  return d/l;
}

function speedOfSound(gamma){
  p = PRESSURE;
  d = DENSITY;
  return sqrt(gamma*p/d);
}
```

config file:
```
SCREEN_OUTPUT=(INNER_ITER, WALL_TIME, RMS_DENSITY, {dragOverLift()})
VOLUME_OUTPUT = (COORDINATES, SOLUTION, {speedOfSound(1.4)})
```

The language is much more flexible and allows also for if-statements, loops, maps and so on. For the output it does not make that much sense, but for initial conditions it might be useful later.

jayantmukho:
> The language is much more flexible and allows also for if-statements, loops, maps and so on. For the output it does not make that much sense, but for initial conditions it might be useful later.

This could be useful even for general outputs. One example that comes to mind is if you wanted to calculate the momentum/pressure loss due to a shock. Excited for these new features!

QuickStart/inv_NACA0012.cfg
```diff
@@ -261,7 +261,7 @@ CONV_CAUCHY_ELEMS= 100
 % Epsilon to control the series convergence
 CONV_CAUCHY_EPS= 1E-6
 %
-SCREEN_OUTPUT=(INNER_ITER, WALL_TIME, RMS_DENSITY, LIFT, DRAG, CAUCHY_SENS_PRESS, CAUCHY_DRAG RMS_ADJ_DENSITY RMS_ADJ_ENERGY)
+SCREEN_OUTPUT=(INNER_ITER, WALL_TIME, RMS_DENSITY, LIFT@airfoil, DRAG, CAUCHY_SENS_PRESS, CAUCHY_DRAG RMS_ADJ_DENSITY RMS_ADJ_ENERGY)
```
jayantmukho, line 264: That's a cool feature. Would it be possible to specify multiple markers? Something along the lines of `LIFT@(upper_surface, lower_surface)

economon:
@talbring : great progress! As previously discussed, I think this goes in the correct direction.. I would love to see completely customizable obj/constraints/parameters in a flexible way, for any problem a user can dream up.

Like @clarkpede says, sometimes the objectives can require some heavy lifting (and require many local solution variables), and this thing is quite difficult to implement currently (it would be great if things like #889 were made easier). Do you think the language parser can eventually be used for this, or do you think we should just stick to writing those implementations in the COutput classes for each problem type (flow, elasticity, heat, etc).?

pcarruscag:
On this issue of custom outputs we should probably follow the industry, i.e. provide basic building blocks such as area/volume integrals, mass flow averages, etc. of arbitrary functions (functions of the output fields) and then on top of that structure build some useful defaults (e.g. force / moment coefficients).

talbring:
Here is a small description on how the user defined functions work at the moment : https://gist.github.com/talbring/cfd26e25218a15b8788ce4e047f97ed6. Please let me know what you think.

economon:
@talbring : since you are already working on the output here and if you agree, can you please add the sensitivities as a required output for adjoint runs even if the volume fields are customized (like we do for coords/conservatives for the primal)? I have gotten several reports that this is tripping up some folks, e.g., see #964

talbring:
> @talbring : since you are already working on the output here and if you agree, can you please add the sensitivities as a required output for adjoint runs even if the volume fields are customized (like we do for coords/conservatives for the primal)? I have gotten several reports that this is tripping up some folks, e.g., see #964 

In fact I already deactivated it in this branch. It makes no sense to allow customizable output for the adjoint anyway and it was giving troubles when you want to use the same config for primal and adjoint.

talbring:
Some update here: I had to get rid of the CParse library ... It was too slow since it is based  solely on RTTI. I replaced it with [exprtk](https://github.com/ArashPartow/exprtk). This is a fast and widely used expression library. The only drawback is that it is a header-only file with ~40k(!) lines of code. Compilation takes quite long, so I am using pimpl approach to interface it with SU2 in order not to recompile it for every compilation target. Furthermore I added the exprtk code as an archive (similar to what we do with the tecplot IO library).

pcarruscag:
Don't forget #1025 @talbring since you are moving some code into modules.

erangit:
Tim,
Stale is taking over this PR. I think it is useful, not only for the ventilator's effort. Would you like to pack at least the part that is completed and merge to develop?
Thxs,
Eran

PR #833, opened 2019-12-08
Discrete adjoint for deforming meshes

cvencro:
## Proposed Changes
This PR introduces surface deformation routines in CMeshSolver and enables discrete adjoint calculations for deforming mesh problems. Gradient verification for a pitching airfoil using a deforming mesh is good compared against rigid motion for the same pitching airfoil:

image

## Related Work
Contains PR #919 which cleaned up legacy FSI and deformation methods 
Builds on work previously discussed in PR #766 
Includes fix for issue #860 

## PR Checklist

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #914, opened 2020-03-20
Multiphysics interpolation new features and improvements

pcarruscag:
## Proposed Changes
Improvements and general cleanup / restructuring of CInterpolator and co. consisting of:
#### New features:
 - Control of the RBF interpolation matrix sparsity via option `RADIAL_BASIS_FUNCTION_PRUNE_TOLERANCE` to drop relatively small coefficients.
 - Generalize nearest neighbor interpolation to KNN + inverse distance weighting, "k" determined by option `NUM_NEAREST_NEIGHBORS` (because "why not").
#### Enhancements:
 - The setup of CRadialBasisFunction was optimized (more use of BLAS) and it is now parallel (MPI and threads) over multiple interface patches.
 - The setup of CIsoparametric was optimized, now it only evaluates isoparameters for a few nearby candidate donor elements (instead of all possible ones).
 - The setup of CMirror was optimized both for memory (by communicating only information as needed) and speed (by reducing algorithmic complexity from quadratic to linearithmic) this item is in #833.
 - The setup of CIsoparametric, CNearestNeighbor, and CMirror are now hybrid parallel.
#### Maintenance:
 - Restructure (split) and cleanup the files.
 - Heavy refactoring of RBF interpolation.
 - Heavy refactoring of isoparametric interpolation, the method was incorrect for quadrilaterals (and the code too complex).
 - Heavy refactoring of mirror (conservative) interpolation, the algorithm is perhaps more complex but the code is cleaner and 10x faster (this is in #833).
 - Medium cleanup of CInterpolator (const correctness, unnecessary mpi ifdefs, etc.).
 - Light cleanup of CDriver/CTransfer (deleting copy-pasted code)
#### Other enhancements:
 - Initialize the linear system solutions in CMeshSolver to speedup 3D FSI simulations, but especially to reduce overhead in FSI discrete adjoint cases.
 - Too improve the quality of deformed grids in FSI cases with large "out-of-plane" displacements, a thicker layer of stiff cells can be imposed around the deforming boundaries. The thickness of the layer is defined by option `DEFORM_STIFF_LAYER_SIZE = [0]` when `DEFORM_STIFFNESS_TYPE = WALL_DISTANCE`. This allows refined regions around a wing tip to be dragged instead of sheared (which happens with the simple 1/x variation).
#### Other fixes:
 - CFEASolver displacement boundary conditions (the weird bug detailed below).
 - CFEASolver load integration for FSI simplified to fix mpi-related bug.
 - Option CFL_REDUCTION_TURB having no effect.
 - Restarts with moving wall cases work by default.

I know, there's way too much stuff in this PR, but it was all required to make 3D FSI more reliable, and hey, it works!
3D_FSI

## PR Checklist


Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -113,15 +113,14 @@ class AccessorImpl
                                                                         \
   AccessorImpl& operator= (AccessorImpl&& other) noexcept               \
   {                                                                     \
-    if(m_data!=nullptr) free(m_data);                                   \
+    MemoryAllocation::aligned_free<Scalar_t>(m_data);                   \
```
pcarruscag: Also fixes a potential memory leak for mac and windows.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2560,278 +2547,212 @@ void CDriver::Interface_Preprocessing(CConfig **config, CSolver***** solver, CGe
...
-        for (iMarkerTarget = 0; iMarkerTarget < nMarkerTarget; iMarkerTarget++) {
-
-            /*--- If the tag GetMarker_All_ZoneInterface(iMarker) equals the index we are looping at ---*/
-          if ( config[targetZone]->GetMarker_All_ZoneInterface(iMarkerTarget) == iMarkerInt ) {
-              /*--- We have identified the identifier for the interface marker ---*/
-              markTarget = iMarkerTarget;
-
-              break;
-          }
+        /* --- Check if zones are actually sharing the interface boundary, if not skip. ---*/
```
talbring: Maybe you can put these things that follow in a factory class ?

pcarruscag: I thought about it, but there's all kinds of complicated logic here with some classes requiring info from CSolver, it did not seem right to just copy paste the entire thing somewhere else.

TestCases/sliding_interface/bars_SST_2D/bars.cfg
```diff
@@ -252,7 +252,7 @@ SLOPE_LIMITER_TURB= NONE
 TIME_DISCRE_TURB= EULER_IMPLICIT
 %
 % Reduction factor of the CFL coefficient in the turbulence problem
-CFL_REDUCTION_TURB= 0.10
+CFL_REDUCTION_TURB= 1.0
```
pcarruscag: I changes all but 2 test cases to use CFL_REDUCTION_TURB=1 (the lower value previously had no effect and so the residuals changed).

TestCases/parallel_regression.py
```diff
@@ -834,7 +834,7 @@ def main():
     Jones_tc.cfg_dir   = "turbomachinery/APU_turbocharger"
     Jones_tc.cfg_file  = "Jones.cfg"
     Jones_tc.test_iter = 5
-    Jones_tc.test_vals = [-5.316335, 0.355081, 44.772280, 2.269966] #last 4 columns
+    Jones_tc.test_vals = [-5.290553, 0.372874, 44.765930, 2.270197] #last 4 columns
```
pcarruscag: For two other test cases I updated the residuals to have some coverage of the CFL_REDUCTION_TURB option.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -755,29 +763,68 @@ void CFEASolver::Set_ReferenceGeometry(CGeometry *geometry, CConfig *config) {
 
 }
 
+void CFEASolver::Set_VertexEliminationSchedule(CGeometry *geometry, const vector<unsigned short>& markers) {
```
pcarruscag: This is the method to detect problem nodes.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -2781,21 +2730,21 @@ void CFEASolver::GeneralizedAlpha_UpdateLoads(CGeometry *geometry, CSolver **sol
...
-    for (auto iPoint = nPointDomain; iPoint < nPoint; iPoint++)
-      LinSysRes.SetBlock_Zero(iPoint);
+  Jacobian.InitiateComms(LinSysSol, geometry, config, SOLUTION_MATRIX);
+  Jacobian.CompleteComms(LinSysSol, geometry, config, SOLUTION_MATRIX);
 
-    SU2_OMP(section)
-    for (auto iPoint = nPointDomain; iPoint < nPoint; iPoint++)
-      LinSysSol.SetBlock_Zero(iPoint);
+  for (auto iPoint : ExtraVerticesToEliminate) {
+    Jacobian.EnforceSolutionAtNode(iPoint, LinSysSol.GetBlock(iPoint), LinSysRes);
```
pcarruscag: And this is what we need to do before solving the system.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Done in #919

pcarruscag:
I'll merge this in with #833...

PR #916, opened 2020-03-23
fixing for support for SST model in C1Dinterpolation

auzbaig:
small change to fix support for SST model.

@pcarruscag I think you inadvertently added an else-if during your formatting corrections. I intentionally did not use an else-if as it will support both SA and SST models.

Thanks.

### DIFF
--- a/Common/src/toolboxes/C1DInterpolation.cpp
+++ b/Common/src/toolboxes/C1DInterpolation.cpp
@@ -148,7 +148,7 @@ vector<su2double> CorrectedInletValues(const vector<su2double> &Inlet_Interpolat
   /*---For turbulence variables columns---*/
   if (nVar_Turb == 1)
     Inlet_Values[nDim+5] = Inlet_Interpolated[5];
-  else if (nVar_Turb == 2)
+  if (nVar_Turb == 2)
     Inlet_Values[nDim+6] = Inlet_Interpolated[6];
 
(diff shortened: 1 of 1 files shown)


auzbaig:
@pcarruscag and thanks alot for doing the formatting corrections on my behalf.

Common/src/toolboxes/C1DInterpolation.cpp
```diff
@@ -148,7 +148,7 @@ vector<su2double> CorrectedInletValues(const vector<su2double> &Inlet_Interpolat
   /*---For turbulence variables columns---*/
   if (nVar_Turb == 1)
     Inlet_Values[nDim+5] = Inlet_Interpolated[5];
-  else if (nVar_Turb == 2)
+  if (nVar_Turb == 2)
```
pcarruscag: It was not by accident, nVar_Turb cannot be both 1 and 2, only one of those branches executes, if your intention is to execute both branches when nVar_Turb is 2, the code is wrong.

auzbaig: @pcarruscag Thanks for the correction. I have changed the code, correctly with the else-if.

>   else if (nVar_Turb == 2){
    Inlet_Values[nDim+5] = Inlet_Interpolated[5];
    Inlet_Values[nDim+6] = Inlet_Interpolated[6];
  }

PR #921, opened 2020-04-01
Fix bugs when restarting a shape optimization problem

clarkpede:
## Proposed Changes
I found two bugs that lead to crashes when trying to restart a shape optimization problem:

1. The python code still uses the deprecated `has_key` function for dicts.
2.  If the default config file is used in the python scripts, then the aerodynamic coefficients aren't saved anywhere even if `aerodynamics` is called.

You can reproduce the bugs using the following MWE:

#### Minimal working example

As per the documentation, you should be able to restart a previous shape optimization problem with a command like:
```
python shape_optimization.py -f turb_SA_RAE2822.cfg -r project.pkl
```
where the `-r` command specifies a pickle files with the saved project state from a previous run.

I checked the problem on the [RAE2822 constrained optimization tutorial](https://su2code.github.io/tutorials/Turbulent_2D_Constrained_RAE2822/). You can download the files using the instructions on that page.  In order to reproduce the bug, run the following chain of commands:

1. As per the tutorial, run `SU2_GEO turb_SA_RAE2822.cfg`
2. Modify the `turb_SA_RAE2822.cfg` file to specify that `OPT_ITERATIONS=1`
3. Run `python $SU2_RUN/shape_optimization.py  -f turb_SA_RAE2822.cfg`
4. Modify the `turb_SA_RAE2822.cfg` file to specify that `OPT_ITERATIONS=2`
5. Run `python $SU2_RUN/shape_optimization.py -f turb_SA_RAE2822.cfg -r project.pkl`

## Remaining work

This should be covered by a test case. The problem I used takes a bit of time to compute, even using just the restart.  Is there an cheaper shape optimization problem that could run quickly as a test case?

## Related Work
None.

## PR Checklist


jayantmukho:
Usually when I am testing and changing shape_optimization features, I use the `TestCases/optimization_euler/steady_naca0012` case. It's quick and you can make it faster by specifying a better convergence characteristic (currently it runs until density residual is < -13)

pcarruscag:
Even without regression I suggest we merge this for now for the next release.

clarkpede:
Yeah, sorry.  I still plan to add a regression test sometime in the future.  But I've been swamped this past month.

PR #923, opened 2020-04-01
[WIP] Shape optimization overhaul

jayantmukho:
## Proposed Changes

This is a draft PR to discuss and document a much needed overhaul of the shape optimization features in SU2. There have been significant improvements to the main code base (for example, python wrapper integration, the new output structure, new mesh solver etc.) which haven't been leveraged by the shape optimization structure of SU2. 

The main idea behind this overhaul is to move as much of the functionality of the current shape optimization framework to python. This creates a more user-friendly and robust interface for defining design variables and constraints. Its easy to make that claim, but let me elucidate it with an example:

> Suppose I am trying to optimize a wing for a given flight condition but I want more design freedom than altering the airfoil shapes along the wing using FFDs. I want to be able to control wing span, taper ratio, dihedral angle, twist, and any number of other geometrical representations of the wing. Currently, I would have to create a new Design Variable (DV) for most of those variables. If anyone has tried this, they know it is a tedious task with having to make alterations in various parts of the code.
>
> But all of these design variables can be represented by aggregations of FFD point movements. The span of the wing can be controlled by constraining the movement of FFD points in the y-direction to be proportional to their y coordinate. The current twist variable does a similar aggregation, at each twist location it rotates the corresponding 4 FFD points by the given angle. 
>
>These design variables could be represented by simple python functions that take in  a set of FFD points and aggregate their movement in the desired way. All this can be done without any incision into the C++ code, given a python interface that can provide the C++ code the final FFD point locations and letting it do the deformation. 

Moving functionality to python could also remove the need for some of the other SU2 modules that haven't been updated for a while. For example, SU2_GEO is a bloated wrapper c++ script that calls a couple of CGeometry/CSurfaceMovement functions. The calling of these functions could be easily ported to python and increase the functionality of the Python Wrapper interface. 

Some of the specific things I want to do in this PR:

1. Refactor SU2_GEO such that all its functionality can be performed in python (in progress)
2. Add a generalized thickness constraint (in progress)
3. Be able to do multiple types of geometry evaluation (for example  wing+fuselage)
4. Expose some of the mesh deformation functions to python to allow for robust design variable definition
5. Perform all of the shape optimization using the in-memory python scripts
6. Intelligent handling of solution divergences that doesn't break the optimization
7. Better way to define multipoint optimizations
8. Better multi-objective optimization definition (#913 might be helpful in this)
9. Simplify the multitude of shape optimization options (#733)

As you can probably tell, a lot of my examples and motivations revolve around aerospace applications. So while I try and make a general framework that is robust, I might not be thinking in a broad enough fashion. I would love to get input from the community for any functionality that they might want to see in this framework. The more robust this is, the more widely useful it will be.

I am also a newbie to the python wrapper world of SU2. If there are any glaring road blocks or things that have been tried before but failed, please let me know. 
   
## Related Work
#733 

## PR Checklist


clarkpede:
A question and a comment:

1.How does this relate to my new PR #921? Should I close that one, since the shape optimization is going to be overhauled anyways?
2. Both Beckett Zhou and I have performed shape optimization with objective functions computed as a post-processing step.  From my (admittedly biased) perspective, I would like to have this sort of capability built into SU2. You can reference one of Beckett's papers ([\[1\]](https://arc.aiaa.org/doi/abs/10.2514/6.2015-3355) or [\[2\]](https://arc.aiaa.org/doi/abs/10.2514/6.2017-3658)) to see the sort of framework I'm talking about.

talbring:
@jayantmukho Thanks for taking the lead on that. This is long overdue ... I already talked to some people about that and the conclusion was often that we should get rid of the current python framework and rely solely on the in-memory python wrapper functionalities. From what I can tell this is also what you would suggest.

I have some additional things to add here:

- We should keep in mind other things than "just" shape optimization problems.
- The surface parameterization should be more or less independent of all the other things, so that we can easily integrate other tools/methods. In a first step it might be a good idea to just put all the FFD stuff into a separate (c++) library with a python interface.
- As you already mentioned, I am currently working on a more general way of defining output fields and in particular also objective functions in #913. I will add more information to this PR soon. The new python implementation should, in theory, not need to know about the specific objective function, but rather just asks for the value and the (surface) gradient. 
- We have to expose more c++ classes/functions to python. Especially the config and the geometry. 

Maybe I will add more stuff/ideas, but thats it for now.

@clarkpede : We will probably keep most of the postprocessing stuff in c++, so don't worry about that at the moment.

jayantmukho:
> 1.How does this relate to my new PR #921? Should I close that one, since the shape optimization is going to be overhauled anyways?

@clarkpede this is going to take a while to implement (on the order of months). In the mean time, it would still be good to fix the issue you bring up in #921, if it is an easy fix. If you think the fix is going to be on the same order of time, then might not be worth investing the effort if it is going to get overhauled anyway. 

As for the post-processed objective functions, I am planning on fully incorporating #913 into this work so that any variables that can be defined using the expression language, can be used as an objective function. According to that PR, the post-processing has been auto-differentiated as well, so getting gradients should be automatic as well. 

@talbring I agree with all of your points. I especially want to extract the FFD functionality and expose CGeometry and CConfig functions to python. I haven't used swig before so I need to do some homework. And yes, this PR is called Shape Optimization overhaul but I am hoping the changes made here are general enough that they can be used in a plethora of other situations. If I end up being myopic in any of the refactoring, please point it out :)

clarkpede:
@jayantmukho The fix in PR in #921 is done.  Whereas the shape optimization restart did not work before, it does work now, at least on the two test cases I checked.  The only thing missing is a regression test, to make sure bugs like that don't happen again.  So I agree that it should be merged soon while the long-term refactoring occurs.

As for post-processing, the objective function calculation can be quite complex in the use-cases Beckett and I have encountered.  That's why it's run as a post-processing step, rather than updated continuously during runtime. But I agree that those types of changes can be limited to the C++ side of the code, with minimal changes to the python wrappers.  So it may not be relevant to the current refactoring.

PR #924, opened 2020-04-02
Fix wall distance computation for multizone problems

talbring:
## Proposed Changes
I changed the wall distance computation so that it correctly incorporates also viscous walls in other zones.

I am not 100% happy where it is implemented, but I have not found a better place for this computation. Furthermore this splits the nice indepedent preprocessing routines in the driver constructor. Nevertheless, I think it should be fine.

## PR Checklist


Common/src/geometry/CGeometry.cpp
```diff
@@ -4149,3 +4110,58 @@ void CGeometry::SetNaturalElementColoring()
...
+      CGeometry *geometry = geometry_container[iZone][iInst][MESH_0];
+      if (wallDistanceNeeded[iZone])
+        geometry->SetWallDistance(numeric_limits<su2double>::max());
+    }
+
+    /*--- Loop over all zones and compute the ADT based on the viscous walls in that zone ---*/
+    for (int iZone = 0; iZone < nZone; iZone++){
+      CGeometry *geometry = geometry_container[iZone][iInst][MESH_0];
+      unique_ptr<CADTElemClass> WallADT = geometry->ComputeViscousWallADT(config_container[iZone]);
+      if (WallADT && !WallADT->IsEmpty()){
```
pcarruscag, line 4147: Is it ok to do this? Don't you need two nested if's to avoid potentially dereferencing a nullptr?

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11800,3 +11672,125 @@ void CPhysicalGeometry::Compute_Nacelle(CConfig *config, bool original_surface,
...
+  std::unique_ptr<CADTElemClass> WallADT(new CADTElemClass(nDim, surfaceCoor, surfaceConn, VTK_TypeElem,
+                                                           markerIDs, elemIDs, true));
+
+  /* Release the memory of the vectors used to build the ADT. To make sure
+     that all the memory is deleted, the swap function is used. */
+  vector<unsigned short>().swap(markerIDs);
+  vector<unsigned short>().swap(VTK_TypeElem);
+  vector<unsigned long>().swap(elemIDs);
+  vector<unsigned long>().swap(surfaceConn);
+  vector<su2double>().swap(surfaceCoor);
```
pcarruscag, line 11766: Can you remove this please, these variables are local, this is not necessary.

Common/src/geometry/CGeometry.cpp
```diff
@@ -4149,3 +4110,58 @@ void CGeometry::SetNaturalElementColoring()
...
+      /*--- Check if a zone needs the wall distance and store a boolean ---*/
+
+      ENUM_MAIN_SOLVER kindSolver = static_cast<ENUM_MAIN_SOLVER>(config_container[iZone]->GetKind_Solver());
+      if (kindSolver == RANS ||
+          kindSolver == INC_RANS ||
+          kindSolver == DISC_ADJ_RANS ||
+          kindSolver == DISC_ADJ_INC_RANS ||
+          kindSolver == FEM_LES ||
+          kindSolver == FEM_RANS){
+        wallDistanceNeeded[iZone] = true;
```
pcarruscag, line 4132: I hope this logic only appears once :)

Common/src/fem_wall_distance.cpp
```diff
@@ -277,31 +310,30 @@ void CMeshFEM_DG::ComputeWall_Distance(CConfig *config) {
...
-          /*--- Not a viscous wall boundary, while viscous walls are present. 
+        else if ( !WallADT->IsEmpty() ) {
+          /*--- Not a viscous wall boundary, while viscous walls are present.
                 The distance must be computed. Loop over the integration points
                 and do so. ---*/
           for(unsigned short i=0; i<nInt; ++i) {
 
             const su2double *coor = surfElem[l].coorIntegrationPoints.data()
-                                  + i*nDim;
+                + i*nDim;
```
pcarruscag, line 326: This looks a bit awkward, all in one line or maybe even `&surfElem[l].coorIntegrationPoints[i*nDim]`

Common/include/fem_geometry_structure.hpp
```diff
@@ -1364,6 +1358,25 @@ class CMeshFEM_DG: public CMeshFEM {
...
+   * \brief Set wall distances a specific value
+   */
+  void SetWallDistance(su2double val) override;
+
+  /*!
+   * \brief Set the wall distance based on an previously constructed ADT
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] WallADT - The ADT to compute the wall distance
+   */
+  void SetWallDistance(CConfig *config, CADTElemClass* WallADT) override;
```
pcarruscag, line 1379: :+1: on the use of unique_ptr.
Can you make the use of CConfig `const` in these methods?

Common/src/geometry/CGeometry.cpp
talbring: Yep that is totally legal, even written in the standard. If the first operand of && is false then the second one is not evaluated. See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3092.pdf page 118/119

Common/src/geometry/CPhysicalGeometry.cpp
talbring: Huh yeah I copied it over from the old implementation. I was already wondering why it was there. But assumed that someone had a reason to do that.

Common/src/geometry/CGeometry.cpp
talbring: With this implementation I removed even tell of these statements in the driver ;) but we really need some common structure for all this information soon ...

Common/src/fem_wall_distance.cpp
talbring: Here again, just used the old implementation. Maybe would have been easier, if I just would have done an in-place modification of this routine. But I can change that anyway.

pcarruscag: Yup I know this is just moved around, but it caught my eye.

talbring:
> Looks good, but probably makes @koodlyakshay work in #877 a bit harder?
> Comments below.

I might make it harder. I thought about maybe making the adt structure more general so that it communicates a user defined struct containing some information.

pcarruscag:
Well it already communicates marker ID's and so on, it is probably less error prone for the user to map their information to those indices (type punning structs for global communication sounds a bit meh).

Issue #926, opened 2020-04-07
False flow rate output for axisymmetric cases in SetAnalyzeSurface(...) 


FlorianDm:
The flow rate output for axisymmetric cases from SetAnalyzeSurface(...) is wrong due to the "else" on line 196 in CFlowOutput.cpp.

When y[i] = 0 (the node lies on the axis) AxiFactor is just set to 1.0 instead of y[i] * 2 * pi (y-coordinate of the vertex of the primal grid) which may cause a significant error

For me I fixed it simply by setting " else AxiFactor = 0.5 * y[i-1] * 2 * pi " as the loop always ends on the axis. This works for my inlet and outlet (left to right) but this will fail if the loop starts at y=0 instead of ending there. Perhaps that could happen in a right to left domain. 

There is certainly a safer more general way I imagine by accessing the y-coordinate of the vertex of the dual grid 

AxiFactor = PI_NUMBER*geometry->node[geometry->vertex[iMarker][iVertex-1]->GetNode()]->GetCoord(1);

economon:
Another idea is to use the y-coordinate of the nearest normal neighbor to the point that falls on the axis with something like:

```
Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
AxiFactor = PI_NUMBER*geometry->node[Point_Normal]->GetCoord(1);
```

Note that this only works for the node that lies on the marker, all others should be computed as before.

If you have some time, could you please submit a pull request with your proposed changes for us to consider? Thanks!

FlorianDm:
Hi, thanks for the fast response. 

GetNormal_Neighbor() didn't return the index needed so I did it the other way. Perhaps someone can improve on that easily.

PR #927, opened 2020-04-07
Fix mass flow calculation near axis for axisymmetric problems

FlorianDm:
Fix for #926, AxiFactor for Vertex next to axis (y=0)


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -111,7 +111,7 @@ void CFlowOutput::AddAnalyzeSurfaceOutput(CConfig *config){
 void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfig *config, bool output){
 
   unsigned short iDim, iMarker, iMarker_Analyze;
-  unsigned long iVertex, iPoint;
+  unsigned long iVertex, iPoint, iVertex_Normal, iPoint_Normal;
```
pcarruscag, line 114: Don't declare more variables here, we do it everywhere but it is not the c++ way and easily leads to bugs.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -192,8 +192,11 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
           if (axisymmetric) {
             if (geometry->node[iPoint]->GetCoord(1) != 0.0)
               AxiFactor = 2.0*PI_NUMBER*geometry->node[iPoint]->GetCoord(1);
-            else
-              AxiFactor = 1.0;
+            else {
+              if (iVertex != 0) iVertex_Normal = iVertex-1; else iVertex_Normal = iVertex+1;
+              iPoint_Normal = geometry->vertex[iMarker][iVertex_Normal]->GetNode();
```
pcarruscag, line 197: Instead:
```c++
auto iVertex_Normal = (iVertex != 0) (iVertex-1) : (iVertex+1);
auto iPoint_Normal = geometry->vertex[iMarker][iVertex_Normal]->GetNode();
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -266,10 +269,10 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
 
           Surface_StreamVelocity2[iMarker]   += Vn2*Weight;
           Surface_TransvVelocity2[iMarker]   += Vtang2*Weight;
-
+          
         }
       }
-
+
```
pcarruscag, line 275: Strip all trailing white spaces.

FlorianDm:
Yes it depends on the order of the grid points but it will give the correct y-coordinate whenever the previous vertex in the loop is the one adjacent to the y=0 vertex on the marker or if the loop starts at y=0 it works if the next one in the loop which is adjacent on the marker. It seems to be the case for my examples but I don't think it always is. 

I don't like it but I can't find a Get which returns the index of the adjacent node on the marker when passing the index of the current vertex or node in the loop. 

` geometry->vertex[iMarker][iVertex]->GetNormal_Neighbor() ` does not return the right one 

` geometry->node[iPoint]->GetPoint(index) ` with some `index` should return the right one but the index where it is stored varies so I need to look into the logic behind that

SU2_CFD/src/output/CFlowOutput.cpp
FlorianDm: neat thanks!

pcarruscag:
I see, GetNormal_Neighbor gives you the closest point in the normal direction, that normal is 90 degrees from the direction you want to search, so that is a dead end.

You can loop over all possible `geometry->node[iPoint]->GetPoint(index)` and for each call `GetVertex(iMarker)` the one that does not return -1 is (should be) the point you want.

economon:
Thanks @FlorianDm for putting in the fix!

Although, it sound like we still need something more general.. I think the normal neighbor concept could still work, it's just that you need to be starting from the vertex list on the axis (not the inlet or the outlet) to find it.

Another simple thing to try is just setting AxiFactor = 0.0 if the y coord = 0.0, instead of 1.0. If this grid is stretched with small spacing near the wall, the contribution from the dual CV on the axis could be very small.

FlorianDm:
This is now completely general I believe. `geometry->node[iPoint]->GetPoint(index)` always returns the right point either for index 0 or 1 at least in a quad mesh (there are only two neighbours at the corner) and the wrong one equals the normal neighbour. I tried it on a triangular mesh and it worked as well but perhaps it won't always be correct.

The worst case is an error equal to the one of setting it to 0.0 and it's just as safe. Setting it to 0.0 is what I did at first but the axis is not a wall in general so it's usually the point with the largest cells and the highest velocities unless it's an Euler case. If there are very few nodes overall on the marker the contribution may not be that small.

I suppose looking for the vertices on the axis marker here would be more messy and prone to bugs (and much more work for me)

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -184,16 +184,20 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
...
               AxiFactor = 2.0*PI_NUMBER*geometry->node[iPoint]->GetCoord(1);
-            else
-              AxiFactor = 1.0;
+            else{
+              auto neighbor0 = geometry->node[iPoint]->GetPoint(0);
+              auto neighbor1 = geometry->node[iPoint]->GetPoint(1);
+              auto neighbor_normal = geometry->vertex[iMarker][iVertex]->GetNormal_Neighbor();
+              AxiFactor = PI_NUMBER*((neighbor0 != neighbor_normal)? geometry->node[neighbor0]->GetCoord(1) : geometry->node[neighbor1]->GetCoord(1));
+            }
           } else {
```
pcarruscag, line 201: Can you try the following please, I think this problem has a general and robust solution, and indeed significant mass flow may go through this cell (even if the radius is small).
```c++
else {
  AxiFactor = 0.0;
  for (unsigned short iNeigh = 0; iNeigh < geometry->node[iPoint]->GetnPoint(); ++iNeigh) {
    auto jPoint = geometry->node[iPoint]->GetPoint(iNeigh);
    if (geometry->node[jPoint]->GetVertex(iMarker) >= 0) {
      AxiFactor = PI_NUMBER * geometry->node[jPoint]->GetCoord(1);
      break;    
    }
  }
}
```

FlorianDm: Thanks! That works perfectly. I didn't get it when you suggested it at first

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -184,16 +184,26 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
 
       for (iVertex = 0; iVertex < geometry->nVertex[iMarker]; iVertex++) {
         iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
-
+
```
pcarruscag, line 187: Great! Just be sure to remove these spaces and we should be good to go.

FlorianDm:
Yes I get that now and I am doing it exactly like that. For this one I thought I leave it. I tried to do a rebase of my develop to remove older commits from this PR but then git was telling me it was not able to include all the following commits including merges etc so I aborted. 

From now on for any fix I would branch every time anew from my develop (which is a copy of su2code +this PR). I also do the same for my own development and then merge any fix branches I need into that. If later you don't merge those fixes because that part of the code is revamped anyway I will bin that branch and redo it on top of your new structure as that would probably be better anyway and just in case I ever want to propose those own developments to su2code. I will watch and adapt.

Hope that's fine for now but let me me if you want me to redo or rebase anything for this PR

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -266,12 +276,10 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
 
           Surface_StreamVelocity2[iMarker]   += Vn2*Weight;
           Surface_TransvVelocity2[iMarker]   += Vtang2*Weight;
-
+
```
pcarruscag, line 279: Oh no problem on this one, that was just general for the future.
Just take care of this second blank space and feel free to merge.

PR #932, opened 2020-04-11
Call SetSurface_... (again)

FlorianDm:
Those Set functions need to be called at least for the Get functions in CEulerSolver::Evaluate_ObjFunc unless they are supposed to be called somewhere else.

These calls exist in the legacy output and seem to have been removed because SetHistoryOutputPerSurfaceValue is replacing them for the history/screen output.

Only the last two commits are relevant. I branched off after fix #927 since it has been reviewed and the delta will disappear here once it has been merged.


FlorianDm:
I suppose when #913 and #923 are done this will be redundant (?)

PR #933, opened 2020-04-12
CFEASolver CMeshSolver maintenance

pcarruscag:
## Proposed Changes
Implements some features that will be necessary to eventually replace the elasticity-based mesh deformation. Namely symmetry boundary condition and the use of the DEFORM_LIMIT option to limit how much of the domain is deformed.

Reduces the discrete adjoint memory usage by disabling the recording of routines that are not relevant or differentiable.

Cleans up some redundant MPI comms in CFEASolver and moves all of them to the methods of that solver (i.e. no more explicit MPI comms in structural integration / iteration).

Makes the structural objective functions available as screen/history output under the name COMBO.

Improves/fixes the incremental load approach when used for multizone problems.

## Related Work
#919 #594

## PR Checklist


SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -331,12 +331,12 @@ void CIntegration::SetStructural_Solver(CGeometry *geometry, CSolver **solver_co
...
       break;
     case (NEWMARK_IMPLICIT):
-      if (fsi) solver_container[FEA_SOL]->ImplicitNewmark_Relaxation(geometry, solver_container, config);
+      if (fsi) solver_container[FEA_SOL]->ImplicitNewmark_Relaxation(geometry, config);
       break;
     case (GENERALIZED_ALPHA):
-      //if (fsi)  solver_container[FEA_SOL]->Update_StructSolution(geometry, solver_container, config);
-      solver_container[FEA_SOL]->GeneralizedAlpha_UpdateSolution(geometry, solver_container, config);
-      solver_container[FEA_SOL]->GeneralizedAlpha_UpdateLoads(geometry, solver_container, config);
+      //if (fsi)  solver_container[FEA_SOL]->Update_StructSolution(geometry, config);
```
economon, line 337: delete comment if not needed

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -1285,51 +1285,34 @@ void CFEAIteration::Iterate(COutput *output,
...
-    Criteria_UTOL = config[val_iZone]->GetIncLoad_Criteria(0);
-    Criteria_RTOL = config[val_iZone]->GetIncLoad_Criteria(1);
-    Criteria_ETOL = config[val_iZone]->GetIncLoad_Criteria(2);
+    /*--- Check user-defined criteria to either increment loads or continue with NR iterations. ---*/
 
-    Residual_UTOL = log10(feaSolver->LinSysSol.norm());
-    Residual_RTOL = log10(feaSolver->LinSysRes.norm());
-    Residual_ETOL = log10(feaSolver->LinSysSol.dot(feaSolver->LinSysRes));
+    bool meetCriteria = true;
+    for (int i = 0; i < 3; ++i)
```
economon, line 1312: can this 3 be an NDIM or other named var?

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -1464,17 +1412,16 @@ void CFEAIteration::Update(COutput *output,
 
     /*--- Verify convergence criteria (based on total time) ---*/
 
-    Physical_dt = config[val_iZone]->GetDelta_DynTime();
-    Physical_t  = (TimeIter+1)*Physical_dt;
-    if (Physical_t >=  config[val_iZone]->GetTotal_DynTime())
+    su2double Physical_dt = config[val_iZone]->GetDelta_DynTime();
```
economon, line 1415: const for these two?

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -2670,7 +2581,7 @@ void CDiscAdjFEAIteration::SetRecording(COutput *output,
 
     /*--- Clear indices of coupling variables ---*/
 
-    SetDependencies(solver, geometry, numerics, config, val_iZone, val_iInst, ALL_VARIABLES);
+    SetDependencies(solver, geometry, numerics, config, val_iZone, val_iInst, COMBINED);
```
economon, line 2584: Personal preference: COMBINED is a bit unclear for me, since I don't know what you're combining just by looking at it.. if it's not too long, I would suggest explicitly naming the combined vars

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -1791,7 +1798,70 @@ void CFEASolver::BC_Clamped_Post(CGeometry *geometry, CNumerics *numerics, CConf
...
-void CFEASolver::BC_DispDir(CGeometry *geometry, CNumerics *numerics, CConfig *config, unsigned short val_marker) {
+void CFEASolver::BC_Sym_Plane(CGeometry *geometry, CNumerics *numerics, const CConfig *config, unsigned short val_marker) {
+
+  if (geometry->GetnElem_Bound(val_marker) == 0) return;
+
+  const bool dynamic = config->GetTime_Domain();
+
+  /*--- Determine axis of symmetry based on the normal of the first element in the marker. ---*/
+
+  const su2double* nodeCoord[4] = {nullptr};
```
economon, line 1809: We have N_POINTS_QUADRILATERAL (and others) in case useful throughout the FEA solver. I know it can be a quad or triangle here, but you might look elsewhere to replace things where it can be more clear

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -2127,15 +2227,15 @@ void CFEASolver::BC_Damper(CGeometry *geometry, CNumerics *numerics, CConfig *co
...
+void CFEASolver::BC_Deforming(CGeometry *geometry, CNumerics *numerics, const CConfig *config, unsigned short val_marker){
 
   for (auto iVertex = 0ul; iVertex < geometry->nVertex[val_marker]; iVertex++) {
 
     /*--- Get node index ---*/
     auto iNode = geometry->vertex[val_marker][iVertex]->GetNode();
 
     /*--- Retrieve the boundary displacement ---*/
-    su2double Disp[3] = {0.0, 0.0, 0.0};
+    su2double Disp[3] = {0.0};
```
economon, line 2238: Use MAXNDIM instead of the 3's throughout?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1958,10 +1958,6 @@ void CSolver::InitiateComms(CGeometry *geometry,
       COUNT_PER_POINT  = nVar*3;
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
-    case SOLUTION_DISPONLY:
```
economon, line 1961: 👍 Thanks for removing if not needed. Any other stale ones not being used in the FEA solver?

SU2_CFD/src/integration/CIntegration.cpp
pcarruscag: I'm not the one who commented, so I don't really know if it is needed. The generalized alpha method is not fully implemented, at least in some areas it prints "NOT IMPLEMENTED" messages, I would rather leave it like this until the implementation is finished.

SU2_CFD/src/iteration_structure.cpp
pcarruscag: No, it is the number of convergence criteria for nonlinear FEM there are 3 even for 2D problems.

SU2_CFD/src/iteration_structure.cpp
pcarruscag: There are only two variables to combine now, solution and geometry, we also had the ALL_VARIABLES enum option, should I use that instead?

SU2_CFD/src/iteration_structure.cpp
pcarruscag: My criteria lately has been, if the variable is used right away no need for const, but I can change it.

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: Yeap, I've been delaying that out of laziness.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: Just that one I think.

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: I've changed it to MAXNNODES_2D to be consistent to what was already there for 3D elements.
Thanks for the review :+1:

SU2_CFD/src/integration/CIntegration.cpp
economon: Got it, sounds good. Can it just be uncommented then if it throws a proper error? Would be less confusing for the next person who comes across it

SU2_CFD/src/iteration_structure.cpp
economon: Sounds good. I think you get my gist.. just looking for places where enums/named vars can be used instead of literals for clarity

SU2_CFD/src/iteration_structure.cpp
economon: I would prefer that, or feel free to make a new one SOL_AND_GEO (unless the number of these options is going to become very large and this type of naming would become problematic)

SU2_CFD/src/solvers/CFEASolver.cpp
economon: 👍 You're welcome. Thanks for all of the contributions you've been making across SU2.

SU2_CFD/src/integration/CIntegration.cpp
pcarruscag: I ended up removing the method entirely, if that operation is needed it can be moved to `GeneralizedAlpha_UpdateSolution`

talbring:
Thanks @pcarruscag ! The symmetry BC was exactly what is needed to remove the old grid deformation and to maintain the same functionalty. I am gonna remove it in a separate PR. Furthermore, since we can include the grid deformation in the adjoint solver, in the end SU2_DOT will only have to deal with the surface deformation.

pcarruscag:
@talbring symmetry it is not the only thing. Can we decide as a group what that is going to look like and discuss the trade-offs? Specially how it will be used, right now the deformation settings are very messy, and the code has no information regarding how/if the different motions can be combined.

talbring:
Sure! We also have a lot of other things to discuss. Lets organize a more structured meeting. I will open a poll for the date on slack (@all: please join if you haven't so far). Furthermore I will prepare an agenda.

SU2_CFD/src/iteration_structure.cpp
economon: Prob best to just do it always... makes the intent clear for others than come after you and may move things around or reuse

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -461,29 +463,31 @@ void CMeshSolver::DeformMesh(CGeometry **geometry, CNumerics **numerics, CConfig
...
+
+  Compute_StiffMatrix(geometry[MESH_0], numerics, config);
+
+  if (ActiveTape) AD::StartRecording();
+
+  /*--- Clear residual (loses AD info), we do not want an incremental solution. ---*/
+  SU2_OMP_PARALLEL
+  {
+    LinSysRes.SetValZero();
+  }
```
pcarruscag, line 489: I am having a third second look at this bit.
Although the results did not change from this memory usage tweak, I think some dependencies might be lost here as this is equivalent to assuming the stiffness matrix does not depend on the reference (undeformed) mesh coordinates.
Which is also an approximation made in SU2_DOT.

PR #935, opened 2020-04-12
Call SetSurface_... 

FlorianDm:
Those Set functions need to be called at least for the Get functions in CEulerSolver::Evaluate_ObjFunc unless they are supposed to be called somewhere else.

These calls exist in the legacy output and seem to have been removed because SetHistoryOutputPerSurfaceValue is replacing them for the history/screen output.

Perhaps this will be redundant when #913 is done


talbring:
Hi @FlorianDm 

thanks for fixing this. We can merge this in for the moment. Note however that I'm reworking at the moment how all of the objective functions are defined in #913.

FlorianDm:
Hello @talbring 

Yess really looking forward to the result of your work!

economon:
Also, we're happy to give regular contributors direct access to the repository (easier to collaborate and participate in reviews, etc.). Just let me know if you would like to be added.

FlorianDm:
@economon it would be a privilege ;)

Issue #946, opened 2020-04-21
TOTAL_HEATFLUX keyword not working as objective function with python optimization


MiGaff:
The shape_optimization.py script fails when setting `OPT_OBJECTIVE= TOTAL_HEATFLUX `.
The name of the variables have apparently been changed, but this change is not reflected in files  _SU2_CFD/src/output/CFlowCompOutput.cpp_ and _SU2_CFD/src/output/CFlowIncOutput.cpp_

In those files, replacing all instances of `"HEATFLUX"` with `"TOTAL_HEATFLUX"` solves the problem. 
For reference, I also run the script SU2/SU2_PY/updateHistoryMap.py after recompiling, but I am not sure if this is needed.

**Desktop (please complete the following information):**
 - OS: Linux (Red Hat 7.6)
 - C++ compiler and version: g++ 8.3.0
 - MPI: mvapich 2.3.1
 - SU2 Version: v7.0.3

economon:
Thanks for posting @MiGaff! This issue is similar to #889, and it sounds like you have the fix, including the python update.

If you have a moment, could you please submit your changes via PR to the develop branch? It would be great to have your contribution.

MiGaff:
Thanks for your reply @economon,
I created a pull request. It's the first time I use git (other than for cloning / pulling) so I hope what I did is correct.

PR #947, opened 2020-04-22
Change HEATFLUX to TOTAL_HEATFLUX in output files so that optimizatio…

MiGaff:
Change HEATFLUX to TOTAL_HEATFLUX in output files so that optimization works correctly with  OPT_OBJECTIVE= TOTAL_HEATFLUX
 
Additionally, a small change to updateHistoryMap.py has been made (a "/" in a path was missing), and historyMap.py has been modified by running updateHistoryMap.py

This fixes [#946](https://github.com/su2code/SU2/issues/946)

## PR Checklist


SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -231,7 +231,7 @@ void CFlowCompOutput::SetHistoryOutputFields(CConfig *config){
 
   ///   /// BEGIN_GROUP: HEAT_COEFF, DESCRIPTION: Heat coefficients on all surfaces set with MARKER_MONITORING.
   /// DESCRIPTION: Total heatflux
-  AddHistoryOutput("HEATFLUX", "HF",      ScreenOutputFormat::SCIENTIFIC, "HEAT", "Total heatflux on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
+  AddHistoryOutput("TOTAL_HEATFLUX", "HF",      ScreenOutputFormat::SCIENTIFIC, "HEAT", "Total heatflux on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
```
pcarruscag, line 234: Is this change required for the fix? (thinking about backwards compatibility, documentation, etc.)

MiGaff: I think so. According to the AddHistoryOutput function, the first parameter is:

> name - Name for referencing it (in the config file and in the code).

If I understand correctly this parameter is called TOTAL_HEATFLUX everywhere in the code, but not in these 2 files.

SU2_PY/updateHistoryMap.py
```diff
@@ -95,7 +95,7 @@ def parse_output(files):
 
 
     outputFields.update(addedOutputFields)
-    f = open(os.path.join(su2_home) + 'SU2_PY/SU2/io/historyMap.py', 'w')
+    f = open(os.path.join(su2_home) + '/SU2_PY/SU2/io/historyMap.py', 'w')
```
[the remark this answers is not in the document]
MiGaff: Yes, that is definitely more elegant.

SU2_CFD/src/output/CFlowCompOutput.cpp
pcarruscag: CHT people @oleburghardt @TobiKattmann, can you give an opinion here, how does this influence existing tutorials, testcases, etc.?

oleburghardt: Hi, these changes shouldn't affect them as everything is done internally there (the value always comes from `solver->GetTotal_HeatFlux()`, if needed).

So from my side I'd be fine with changing the name. For tutorials and verification it's just about producing and checking the gradient, I've never checked for compatibility with the python framework.
Would be great if it can be integrated into the regression tests some day.. O:-)

Issue #942, opened 2020-04-18
What should we do with tricky legacy output options? (and other not very useful ones)


pcarruscag:
Setting WRT_BINARY_RESTART=NO is irrelevant, the output format is determined by OUTPUT_FILES. Is this option around just because of the legacy output? Should it have any effect in the new output?

One needs to set both VISUALIZE_SURFACE_DEF and a surface output format to get a file.
One needs to set both VISUALIZE_VOLUME_DEF and a volume output format to get a file.
Should we just remove these two and let OUTPUT_FILES "decide" what is outputted?

Can we get rid of DISCADJ_LIN_SOLVER and DISCADJ_LIN_PREC and use what is set for the direct solver? Or alternatively, in discrete adjoint problems, use these settings for the primal (recording phase)?
This is annoying because if you use FGMRES for the direct problem and happen to need more linear iterations for the discrete adjoint (say 10 -> 100) in which case you probably switch to BCGSTAB (due to memory), one needs to set/change two options.

If there are options that grind your gears or you think could be more user-friendly lets list them.

pcarruscag:
Another option mentioned by @talbring is MATH_PROBLEM for which the default could depend on the code being used, i.e. for SU2_CFD_AD and SU2_DOT_AD the default should be DISCRETE_ADJOINT

pcarruscag:
This is now all covered in #1126

PR #950, opened 2020-04-24
Small discrete adjoint tweaks and other fixes

pcarruscag:
## Proposed Changes
- Fixes the adjoint residual computation for multizone cases (which was wrong on first inner iterations due to cross term extraction).
- Fixes (hopefully for good) discadj FEA solutions when INCREMENTAL_LOAD=YES (left over from #933)
- Add relaxation to CDiscAdjSolver via option RELAXATION_FACTOR_ADJOINT, that replaces the old RELAXATION_FACTOR_ADJFLOW . I found this to be necessary for adjoint FSI problems, the adjoint of the interpolation matrix (and therefore the cross terms) are not smooth which makes convergence more difficult. In version 6 this was achievable via the old options RELAXATION_FACTOR_FLOW/TURB.
- Adds OpenMP directives to the CFL adaptation routine (a "todo" from #861), restricts the Min/Max/Avg CFL calculation to the fine grid (which will change the output but not the solution).

## PR Checklist


SU2_CFD/include/solvers/CFEASolver.hpp
```diff
@@ -55,8 +55,8 @@ class CFEASolver : public CSolver {
   su2double Conv_Check[3];          /*!< \brief Current values for convergence check: UTOL, RTOL, ETOL. */
   su2double FSI_Conv[2];            /*!< \brief Values to check the convergence of the FSI problem. */
 
-  unsigned long idxIncrement;       /*!< \brief Index of the current load increment */
-  su2double loadIncrement;          /*!< \brief Coefficient that determines the amount of load which is applied. */
+  unsigned long idxIncrement = 0;   /*!< \brief Index of the current load increment */
```
economon, line 58: just a matter of taste, but I like braced initialization for these, e.g., 

```unsigned long idxIncrement{0};```

File: SU2_CFD/src/solvers/CDiscAdjSolver.cpp [hunk 2]
```diff
@@ -476,9 +476,11 @@ void CDiscAdjSolver::SetAdj_ObjFunc(CGeometry *geometry, CConfig *config) {
...
 void CDiscAdjSolver::ExtractAdjoint_Solution(CGeometry *geometry, CConfig *config){
 
-  bool time_n1_needed = config->GetTime_Marching() == DT_STEPPING_2ND;
-  bool time_n_needed = (config->GetTime_Marching() == DT_STEPPING_1ST) || time_n1_needed;
-  bool multizone = config->GetMultizone_Problem();
+  const bool time_n1_needed = config->GetTime_Marching() == DT_STEPPING_2ND;
+  const bool time_n_needed = (config->GetTime_Marching() == DT_STEPPING_1ST) || time_n1_needed;
+  const bool multizone = config->GetMultizone_Problem();
+
+  const su2double relax = (config->GetInnerIter()==0)? 1.0 : config->GetRelaxation_Factor_AdjFlow();
```
economon, line 483: Perhaps we should remove the term "Flow" from this parameter if it will just be a general adjoint relaxation factor for any physical model.. 

I also assume we can assign this in each config separately for multiphysics calculations so that each solver does not need to apply the same value?

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -476,9 +476,11 @@ void CDiscAdjSolver::SetAdj_ObjFunc(CGeometry *geometry, CConfig *config) {
...
 
   unsigned short iVar;
```
economon, line 485: While you are at it.. can you please move these two variable declarations down into their respective loops?

I know it's not your original code, but we should all try to reduce the scope of variables as much as possible (safer that way) and clean up as we go.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2476,20 +2490,37 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
...
+        Min_CFL_Local = min(Min_CFL_Local,myCFLMin);
+        Max_CFL_Local = max(Max_CFL_Local,myCFLMax);
+        Avg_CFL_Local += myCFLSum;
+      }
+      SU2_OMP_BARRIER
+
+      SU2_OMP_MASTER
+      { /* MPI reduction. */
+        myCFLMin = Min_CFL_Local; myCFLMax = Max_CFL_Local; myCFLSum = Avg_CFL_Local;
+        SU2_MPI::Allreduce(&myCFLMin, &Min_CFL_Local, 1, MPI_DOUBLE, MPI_MIN, MPI_COMM_WORLD);
```
economon, line 2517: can you please add a ```if (config->GetComm_Level() == COMM_FULL) { ... }``` around these reductions? The idea is to be able to turn off all non-essential comms in the event that we want to debug or do some specific performance benchmarking

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
pcarruscag: I agree with that, but this goes back to my question during the last Dev meeting, do we want to maintain backwards compatibility in minor releases?

SU2_CFD/include/solvers/CFEASolver.hpp
pcarruscag: I like `=` :)

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
talbring: I would love to maintain backward compatability, but I fear it is not really possible. Unless we kind of remember before every major release what config options we wanted to change... maybe it is enough if we throw an error that options have been renamed/removed. We already do that for some in CConfig::SetConfig_Parsing (which btw can be cleaned up anyway).

However in that particular case we first have to decide whether we need a different value for each individual solver I assume ..

pcarruscag: It is in part for this purpose that I opened #942, so that we can list options we think could be better etc.

TestCases/parallel_regression_AD.py
```diff
@@ -58,7 +58,7 @@ def main():
     discadj_cylinder3D.cfg_dir   = "disc_adj_euler/cylinder3D"
     discadj_cylinder3D.cfg_file  = "inv_cylinder3D.cfg"
     discadj_cylinder3D.test_iter = 5
-    discadj_cylinder3D.test_vals = [-3.724803, -3.838647, 0.000000, 0.000000] #last 4 columns
+    discadj_cylinder3D.test_vals = [-3.755976, -3.858734, 0.000000, 0.000000] #last 4 columns
```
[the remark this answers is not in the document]
pcarruscag: Yes because the case had a value of 0.95 specified (even if it did not do anything because the option did not apply to the discrete adjoint solver) I updated the residuals to keep the "feature" under regression, as it can break if e.g. solution old is set before/after.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
pcarruscag: I vote one size fits all, otherwise we need to get into the whole `if KindSolver == ...`

economon: We could try to be much more systematic about it and do proper deprecation, i.e., find a way to support both a new option and the old one (use version numbers internally?), but mark the old one deprecated and remove in a future release (along with other similarly marked options). A little more effort, but it would achieve what we want

pcarruscag: I went with @talbring 's suggestion.

PR #959, opened 2020-04-28
Contiguous storage of CEdge(s)

pcarruscag:
## Proposed Changes
Make CEdge a class of arrays, instead of using it as an array of classes.
Ended up having to add some const-correctness to CNumerics too.
Forgot to turn off "auto blank space stripping" so a few files will show more changes than this actually introduces.

## Related Work
Needed for #789 (to eventually support the SIMD part).
Similar changes to #753.
Fixes a bug in the SST fluid-fluid interface.

## PR Checklist


Common/include/geometry/CGeometry.hpp
```diff
@@ -191,7 +191,7 @@ class CGeometry {
   CPrimalGrid** face;                    /*!< \brief Face vector (primal grid information). */
   CPrimalGrid*** bound;	                 /*!< \brief Boundary vector (primal grid information). */
   CPoint** node;                         /*!< \brief Node vector (dual grid information). */
-  CEdge** edge;                          /*!< \brief Edge vector (dual grid information). */
+  CEdge* edges;                          /*!< \brief Edge vector (dual grid information). */
```
pcarruscag, line 194: Significant change 1.

Common/include/toolboxes/graph_toolbox.hpp
```diff
@@ -403,8 +403,8 @@ CEdgeToNonZeroMap<Index_t> mapEdgesToSparsePattern(Geometry_t& geometry,
 
   for(Index_t iEdge = 0; iEdge < geometry.GetnEdge(); ++iEdge)
   {
-    Index_t iPoint = geometry.edge[iEdge]->GetNode(0);
-    Index_t jPoint = geometry.edge[iEdge]->GetNode(1);
+    Index_t iPoint = geometry.edges->GetNode(iEdge,0);
+    Index_t jPoint = geometry.edges->GetNode(iEdge,1);
```
pcarruscag, line 407: Significant change 2.

TestCases/parallel_regression.py
```diff
@@ -976,7 +976,7 @@ def main():
     bars_SST_2D.cfg_dir   = "sliding_interface/bars_SST_2D"
     bars_SST_2D.cfg_file  = "bars.cfg"
     bars_SST_2D.test_iter = 13
-    bars_SST_2D.test_vals = [13.000000, -0.619179, -1.564701] #last 4 columns
+    bars_SST_2D.test_vals = [13.000000, -0.619179, -1.564685] #last 4 columns
```
pcarruscag, line 979: This case is full of valgring warnings of the type "Uninitialized bla bla", I could not track the root cause as there are way too many of them. I'm updating the residuals for now.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -404,6 +404,121 @@ void CTurbSolver::BC_Periodic(CGeometry *geometry, CSolver **solver_container,
...
+      visc_numerics->SetPrimitive(PrimVar_i, PrimVar_j);
+
+      /*--- Turbulent variables and their gradients ---*/
+
+      visc_numerics->SetTurbVar(Solution_i, Solution_j);
+      visc_numerics->SetTurbVarGradient(nodes->GetGradient(iPoint), nodes->GetGradient(iPoint));
+
+      /*--- Menter's first blending function ---*/
+
+      if(sst) visc_numerics->SetF1blending(nodes->GetF1blending(iPoint), nodes->GetF1blending(iPoint));
```
pcarruscag, line 503: The F1 blending factor was missing in BC_Fluid_Interface of the SST solver (classic copy paste bug...), which is what was causing a bunch of valgrind warnings.
I moved the entire method to CTurbSolver, I rather have one `if(sst)` than 1 copy pasted method.

Common/include/geometry/dual_grid/CEdge.hpp
```diff
@@ -25,154 +24,187 @@
...
+   * \brief Set the normal vector of an edge.
+   * \param[in] iEdge - Edge index.
+   * \param[in] normal - Vector to initialize the normal vector.
    * \return Value of the normal vector.
    */
-  inline void SetNormal(const su2double *val_face_normal) override {
-    for (unsigned short iDim = 0; iDim < nDim; iDim++)
-      Normal[iDim]=val_face_normal[iDim];
+  template<class T>
+  void SetNormal(unsigned long iEdge, const T& normal) {
```
talbring, line 183: Why are they templated? To pass different array types? :+1:

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7206,80 +7206,56 @@ void CPhysicalGeometry::GatherInOutAverageValues(CConfig *config, bool allocate)
 
 
 void CPhysicalGeometry::SetCoord_CG(void) {
-  unsigned short nNode, iDim, iMarker, iNode;
+  unsigned short iMarker, iNode;
   unsigned long elem_poin, edge_poin, iElem, iEdge;
-  su2double **Coord;
+
+  /// TODO: What is a reasonable MAXNNODE?
```
talbring, line 7212: Is that a todo that you still want to resolve? Otherwise I would say that 64 is definitely enough.

Common/include/geometry/dual_grid/CEdge.hpp
pcarruscag: That was the idea yes, in case we need to pass `std::vector` or `std::array` or even just an object with `operator []`

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: Oh I forgot that, I think 64 is enough too but if anyone as a definitive answer based on the element types we handle... I don't know how crazy a dual volume can get in terms of how many edges connect to it. (@economon any idea?)

talbring: But its just based on the primal structure, isn't it? So the maximum would be 8 for a hexahedron.

pcarruscag: Duh! Your are absolutely right, thanks. I'll fix this, but don't hold the release back because of this PR, this has quite a few changes and we had a ton of small bug fixes, it would be a shame if this introduces some.

talbring: Alright. Then I'll go ahead :)

pcarruscag:
I'll merge this one to make the diff on my other PR's cleaner, but feel free to comment if you see something that does not look right.
I think these are fairly safe changes bug-wise, CEdge does not have a lot of methods and so I did this almost manually.

PR #961, opened 2020-04-29
Time coefficient in classical RK4 explicit

leminhman0312:
## Proposed Changes
For classical RK4 explict, arent the time coefficients suppose to go (1.0, 0.5, 0.5, 1.0)?  The functional step are correct as (1/6, 1/3, 1/3, 1/6).  I couldnt use the explicit RK4 on the tutorial to test this because it says SU2 diverges. 
 

## Related Work
Modified time coefficient in classical RK4

## PR Checklist


pcarruscag:
I think the coefficients are correct. Do you have a reference?

leminhman0312:
I juse use this to look up the formula for rk4 explicit.  Aren the time coefficients used to do: ti, ti+1/2, ti+1/2, ti+ h for k1, k2. k3 and k4?

economon:
Hi @leminhman0312 : we are using a low-storage implementation of RK4 that only requires two storage locations (old & new). I think that, if you multiply out what you see in CEulerSolver::Explicit_Iteration() for the RK4 option, you will see that is should match the standard formulas. Can you give it another look (always good to verify these things)?

pcarruscag:
On the first step (iRKstep=0) the residual (k1) was evaluated for the current solution (y) we then set the solution as y+0.5*k1 (via nodes->AddSolution) so that on the next call (iRKstep=1) k2 is computed.
We keep 3 states, SolutionOld (y), Solution (y+ci*ki), and SolutionNew (y(t+h)) which is built incrementally until it replaces Solution on the last step.

leminhman0312:
Thank you @pcarruscag  and @economon for clarifying.  I understood it better now.

PR #963, opened 2020-04-30
Multigrid and preprocessing efficiency tweaks

pcarruscag:
## Proposed Changes
This is a "don't pay for what you don't use" optimization.
I've disabled implicit time integration during the computation of the MG forcing term.

I've also refactored CMultiGridQueue, it had a bad scaling problem because to remove a point from the queue we were doing a linear search followed by vector::erase, both being O(N) operations.
I've put together a queue-type class that allows fast (amortized O(1)) "seek-and-destroy" of arbitrary items.

## PR Checklist


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2610,7 +2611,7 @@ void CEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_con
 
   if(!ReducerStrategy && !Output) {
     LinSysRes.SetValZero();
-    if (implicit && !disc_adjoint) Jacobian.SetValZero();
+    if (implicit) Jacobian.SetValZero();
```
pcarruscag, line 2614: The changes in residuals of the AD tests are partly due to this change, we were not clearing the matrix in AD mode, which means when multigrid is on we would add on top of the matrix at least once (in practice nPreSmooth + 1 + nPostSmooth) which I think explains why all the residuals I looked at changed for the better (since doubling the matrix halves the correction, kinda).

pcarruscag: This is the improvement in convergence rate for the first case that failed Arina2K:
image
The picture is similar for the other discadj cases with MG, at convergence the sensitivities and adjoints are "the same". I've updated the regressions.

Common/include/geometry/CPhysicalGeometry.hpp
```diff
@@ -38,16 +38,17 @@
...
+  long *Local_to_Global_Point;              /*!< \brief Local-global indexation for the points. */
+  unsigned short *Local_to_Global_Marker;   /*!< \brief Local to Global marker. */
+  unsigned short *Global_to_Local_Marker;   /*!< \brief Global to Local marker. */
+  unsigned long *adj_counter;               /*!< \brief Adjacency counter. */
+  unsigned long **adjacent_elem;            /*!< \brief Adjacency element list. */
+  su2activematrix Sensitivity;              /*!< \brief Matrix holding the sensitivities at each point. */
 
   vector<vector<unsigned long> > Neighbors;
-  map<unsigned long, unsigned long> Color_List;
+  unordered_map<unsigned long, unsigned long> Color_List;
```
pcarruscag, line 51: Moving down the list of geometry preprocessing bottlenecks, accessing std::map items is O(log(N)), std::unordered_map is O(1).
Why do I care about preprocessing you ask? Because it is not (as) easy to make hybrid parallel, and so it better be quick.

Common/src/adt_structure.cpp
```diff
@@ -849,14 +849,18 @@ void CADTElemClass::DetermineNearestElement(const su2double *coor,
...
+          ds = min(0.0, coor[2]-coorBBMin[2]) + max(0.0, coor[2]-coorBBMax[2]); posDist2 += ds*ds;}
+          /*--- The above is equivalent to this:
           for(unsigned short k=0; k<nDim; ++k) {
             su2double ds = 0.0;
             if(     coor[k] < coorBBMin[k]) ds = coor[k] - coorBBMin[k];
             else if(coor[k] > coorBBMax[k]) ds = coor[k] - coorBBMax[k];
-
             posDist2 += ds*ds;
-          }
+          }, but a bit faster. ---*/
```
pcarruscag, line 863: Now the single largest time consuming preprocessing step is wall distance computation, so here are some micro optimizations of a very hot function.

Common/include/geometry/CPhysicalGeometry.hpp
economon: Agreed - it is generally better performance with unordered_map. The only warning is to be careful when replacing any maps where we relied on the ordering.. prob not too common, but possible. We don't want to introduce any non-deterministic behavior

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2679,7 +2680,7 @@ void CEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_container,
                                 unsigned short iMesh, unsigned long Iteration) {
 
   const bool viscous       = config->GetViscous();
-  const bool implicit      = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+  const bool implicit      = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
```
economon, line 2683: What's the implication of switching away from the '_Flow()' time integration option? Shall we standardize on just one function for integration scheme? There's also the Turb() version. Whichever we choose we should make it very clear (this is an easy source for bugs).

pcarruscag: So far as I understand, the _Flow, _Turb versions are what is set in the config file and they are immutable (there is no "Set" method for them).
The mutable version (GetKind_TimeIntScheme) returns what is set by CConfig::SetGlobalParam, so maybe something like this was the intention all along?
The solver-specific versions are used in the constructors while other methods use the runtime version, that is what I changed in the compressible and incompressible solvers.
I also thought of passing a special value of iRKstep (akin to the NO_RK_STEP) (as I am not a big fan of relying on global state) to the pre-processing and residuals methods, and leave the boundary conditions un-touched. But this would not work for the discrete adjoint.
I don't think the chance for bugs is very high, SetGlobalParam is used only in Iteration::Iterate so unless the Integrations start calling each other there is little chance of mixing the settings.

Common/include/geometry/CPhysicalGeometry.hpp
pcarruscag: And since we adopted c++11, we might as well take advantage of it.
Yes I think one of those changes caused all the residuals changes I'm seeing now, we were iterating over map items somewhere.
As for determinism I think unordered_map is with integer keys, as the hash function for those is the identity.

SU2_CFD/src/solvers/CEulerSolver.cpp
economon: If you remove the _Flow() version can you also remove the _Turb() version over in the turb solvers with this PR, to avoid confusion? It will be simpler then if we always use only the single GetKind_TimeIntScheme() throughout. We might consider removing the _Flow() and _Turb() getters altogether and just set the correct IntScheme internally within the config class when switching btwn flow & turb, which would simplify further. In general, less options (w/ the same behavior) = less code = better.

In the future, we could apply a similar treatment for the ConvNumScheme, etc.

Common/src/adt_structure.cpp
```diff
@@ -336,14 +321,16 @@ CADTPointsOnlyClass::CADTPointsOnlyClass(unsigned short nDim,
 
   /*--- Reserve the memory for frontLeaves and frontLeavesNew,
         which are needed during the tree search. ---*/
-  frontLeaves.reserve(200);
-  frontLeavesNew.reserve(200);
+  for (auto& vec : FrontLeaves) vec.reserve(200);
```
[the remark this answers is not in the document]
pcarruscag: It is just a worstcase estimate I think (proportional to log2(N)), nothing breaks by changing or omitting that value, but push_back performance will be better in the first few calls to DetermineNearestNode and co. Maybe @vdweide can add to this.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: It's done for that family of options where possible, constructors and so on need the specific versions for now, but I strongly agree with fewer code = better code.

economon: 👍

Issue #909, opened 2020-03-12
shape_optimization restart for unsteady cases


Eduardo-Carvalho:
Dear SU2 Team Members,

I've been trying to optimize bluff bodies geometries with the shape_optimization script, but I wasn't able to restart any optimization with restart files from previous simulations. This causes too much computational time to be expended to reach a periodic condition of the flow (starting from freestream).

Is it possible to implement this kind of restart for unsteady flow optimization? I've looked for answers at the forúm, doing test simulations and at the tutorial's commentary section, but this feature seems to not exist.

During my tests, I found out that the script doesn't recognize restart files with dual-time step indexes. When the script doesn't find a restart file, it starts from freestream condition, setting the config file inside the first direct simulation folder with the command "RESTART_SOL=NO". This problem can be bypassed by making a copy of one restart file and renaming it without its index. Then the script runs until it reaches other index problems with the output files. This also can be solved by renaming files while the script goes, but it still crashes at the end of the first design cycle.

economon:
Thanks for the issue @Eduardo-Carvalho.

If anyone working in an area related to unsteady optimization has a moment to take a look at this (maybe @TobiKattmann or @ScSteffen?), I'd appreciate it.

ScSteffen:
I'll have a look, but it may take a while until I find some time to work on it :) .

ScSteffen:
Hi @Eduardo-Carvalho ,

your request is now merged into the develop branch and you can test/use it if you like. 
Handling is intuitive: Just set your restart iteration in the config file as you would do normally and activate the restart solution option. Furthermore, place your restart file (two in the case of 2nd order time integration) in the same directory as the config file of your test case.

Then you are set up and can run the scripts as normal. 
For more details, I refer to pull request #964. 

Best
Steffen

PR #964, opened 2020-04-30
Restart Features for some python scripts

ScSteffen:
## Proposed Changes
* Allows direct flow computations in shape_optimization.py to run from given restart solution for unsteady optimization. 
All you have to do is to copy your restart files e.g. `restart_flow_00008.dat` and `restart_flow_00009.dat` in the directory of your `.cfg` and `.su2` (mesh)  file and set the corresponding options:
`RESTART_SOL=YES`
`RESTART_ITER=10` (in this example). 
The `shape_optimization.py` script will use this direct restart file for each design. This means, the direct computation will be started from iteration 10 and will run up to the final iteration. The adjoint will run from the final iteration, but stop at the iteration of the restart file.

* `direct_differentiation.py` now also accepts restart files in the same manner.

* `SENSITIVITY` is now a default field in `VOLUME_OUTPUT`, if the computation mode is `DISCRETE_ADJOINT`.

## Related Work
Issue #909

## PR Checklist


ScSteffen:
Somehow the discrete_adjoint.py and shape_optimization.py scripts throw the error 


`Error in "virtual void CPhysicalGeometry::SetSensitivity(CConfig*)": `
`----------------------------------------`
`Sensitivity x not found in file. `
`------------------------------ Error Exit -------------------------------`
  
when running 'SU2_DOT'. 
Does anyone have a suggestions what the cause of this coud be?

Example is attached
[example.zip](https://github.com/su2code/SU2/files/4559034/example.zip)

economon:
Concerning this error (I am glad you brought it up): 

Error in "virtual void CPhysicalGeometry::SetSensitivity(CConfig*)": 
----------------------------------------
Sensitivity x not found in file. 
------------------------------ Error Exit -------------------------------

Did you solve this issue? I think that the problem is related to specifying VOLUME_OUTPUT fields different than the default (or at least I have had one other person report this to me). When you put in custom fields, the sensitivities are not also added by automatically. 

I think that we should enforce that the sensitivities are always written to the adjoint restart files, like we do with the coords and conservatives for the primal.

ScSteffen:
Hi @economon 

> Did you solve this issue? I think that the problem is related to specifying VOLUME_OUTPUT fields different than the default (or at least I have had one other person report this to me). When you put in custom fields, the sensitivities are not also added by automatically.

yes I noticed that adding sensitivity in the output solves the issue. 
I agree, it makes sense to put `SENSITIVITY` as adefault entry in the case of adjoints to avoid this problem.

ScSteffen:
I have a question regarding the failed test case:
`opt_multiobj1surf_py` using the config file `inv_wedge_ROE_multiobj_1surf`. 
In this config, is the option `RESTART_SOL=YES` neccessary? There is no `RESTART_ITER` specified. 
(Setting `RESTART_SOL=NO` fixes the failed test, however, I don't want to roam in other test cases without asking first ;) )

pcarruscag:
That option has a default value, set somewhere in CConfig.cpp.

ScSteffen:
> That option has a default value, set somewhere in CConfig.cpp.

I've added this in the last commit. :)

ScSteffen:
I'm wondering, if a test case for unsteady restart shape optimization would be computationally too expensive for the test system. 
Please let me know you opinion on that. 
Otherwise this request is complete.

Best
Steffen

pcarruscag:
If you add it to the tutorials test suite... and it runs in < 5min, or it covers a few of the commonly untested options, I'd say go for it.

Out of curiosity, do you use the windowing feature to discard a few initial time steps?

Common/src/CConfig.cpp
```diff
@@ -3156,6 +3156,28 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+      // Add sensitivity volume output for adjoint runs
+      string * vals = new string[nVolumeOutput+1];
+      for (unsigned long i  = 0; i < nVolumeOutput; i++) {
+        vals[i].assign(VolumeOutput[i]);
+      }
+      vals[nVolumeOutput] = string("SENSITIVITY");
+      nVolumeOutput++; //Array has one entry more now
+
+      delete [] VolumeOutput;
+      VolumeOutput = &vals[0];
```
pcarruscag, line 3177: Is this kind of "manual push_back" done somewhere else in config?

SU2_PY/SU2/eval/functions.py
```diff
@@ -222,7 +222,18 @@ def aerodynamics( config, state=None ):
...
     link.extend(name)
-    
+
+    # files: restarts
+    if config.get('TIME_DOMAIN', 'NO') == 'YES' and config.get('RESTART_SOL','NO') =='YES':
+        if  'RESTART_FILE_1' in files: # not the case for directdiff restart
+            name = files['RESTART_FILE_1']
+            name = su2io.expand_part(name, config)
+            link.extend(name)
+        if 'RESTART_FILE_2' in files:
```
pcarruscag, line 232: Will this work for 1st order time integration?

ScSteffen: Not right now. The main script `shape_optimization.py` exits if there are less than two restart options. The code snipped above would work though. 
Which config option is used to specify the order of time integration? Then, I can add this feature.

Common/src/CConfig.cpp
ScSteffen: I'm not quite sure, what you want to say by this. 
Is it bad practise? If yes, how can I improve?

ScSteffen:
> If you add it to the tutorials test suite... and it runs in < 5min, or it covers a few of the commonly untested options, I'd say go for it.
Ok!
> 
> Out of curiosity, do you use the windowing feature to discard a few initial time steps?
No not directly. The idea and motivation was, that I didn't want to wait until the transient phase of the flow field is computed. So I use a precomputed restart file, which serves as a better "inital guess" than freestream values. Then, so the idea, the transient phase is skipped for the first design, and it is at least shortened for other designs. This way, we safe CPU time.

Common/src/CConfig.cpp
pcarruscag: The bad practice was making `VolumeOutput` a raw array of strings instead of `vector<string>` but that is too ingrained now.
I just wanted to make sure this snipped of code was not being duplicated, we are bound to be doing something similar in CConfig...
But actually would it not be better to move the logic of always making sure the sensitivities are added to the output classes?
You can just check `requestedVolumeFields` and if "SENSITIVITY" is not there you do a simple `requestedVolumeFields.emplace_back("SENSITIVITY")`

ScSteffen: I tried this idea. However, in a script run via `discrete_adjoint.py`, where `SU2_DOT` is called after `SU2_CFD_AD`, we get again the above error message, since in the file `config_DOT_AD` sensitivity is still not an volumeoutput.

pcarruscag: Maybe we are not talking about the same thing.
If in the output classes you add a field it will be outputted... SU2_DOT just looks for the name of the variable in the file, it does not care about the options you have in the CConfig, to be clear in CAdjFlowOutput.cpp try adding after line 65:
```
if (find(requestedVolumeFields.begin(), requestedVolumeFields.end(), string("SENSITIVITY")) == requestedVolumeFields.end()) {
  requestedVolumeFields.emplace_back("SENSITIVITY");
}
```

ScSteffen: Okay, I think I got your idea now. 
Pushed it :)

SU2_CFD/src/output/CAdjElasticityOutput.cpp
```diff
@@ -62,6 +62,11 @@ CAdjElasticityOutput::CAdjElasticityOutput(CConfig *config, unsigned short nDim)
     nRequestedVolumeFields = requestedVolumeFields.size();
   }
 
+  if (find(requestedVolumeFields.begin(), requestedVolumeFields.end(), string("SENSITIVITY")) == requestedVolumeFields.end()) {
```
pcarruscag, line 65: :+1: I think this way of what happens in the output stays in the output is clearer, even if it requires duplicating a bit of code across the different adjoint output classes.

ScSteffen: Alright, thanks for your feedback on this :)

PR #969, opened 2020-05-04
Replace 0/NULL with nullptr

economon:
## Proposed Changes
 
Also removes checks for nullptr before delete. These checks are not necessary, since deleting nullptr is perfectly valid.

## Related Work

N/A

## PR Checklist


SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -501,34 +501,34 @@ void CDriver::Postprocessing() {
...
         delete output_container[iZone];
       }
     }
     delete [] output_container;
   }
 
-  if(driver_output != NULL){
+  
     delete driver_output;
-  }
```
pcarruscag, line 531: alignment

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -3404,8 +3404,8 @@ CHBDriver::~CHBDriver(void) {
   unsigned short kInst;
 
   /*--- delete dynamic memory for the Harmonic Balance operator ---*/
-  for (kInst = 0; kInst < nInstHB; kInst++) if (D[kInst] != NULL) delete [] D[kInst];
-  if (D[kInst] != NULL) delete [] D;
+  for (kInst = 0; kInst < nInstHB; kInst++) if (D[kInst] != nullptr) delete [] D[kInst];
+  if (D[kInst] != nullptr) delete [] D;
```
pcarruscag, line 3408: (not your code) but this does not look right.

SU2_CFD/src/fluid_model.cpp
```diff
@@ -47,14 +47,14 @@ CFluidModel::CFluidModel(void) {
...
+  LaminarViscosity = nullptr;
+  ThermalConductivity = nullptr;
 
 }
 
 CFluidModel::~CFluidModel(void) {
-  if (LaminarViscosity!= NULL) delete LaminarViscosity;
-  if (ThermalConductivity!= NULL) delete ThermalConductivity;
+  delete LaminarViscosity;
+  delete ThermalConductivity;
```
pcarruscag, line 57: Glad to see the null checks going away.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -189,10 +189,10 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
     Donor_Variable[3] = Tnormal*donor_config->GetTemperature_Ref();
   }
 
-  if (Edge_Vector != NULL) {
+  
 
     delete [] Edge_Vector;
-  }
```
pcarruscag, line 195: more alignment

SU2_CFD/src/numerics/elasticity/CFEANonlinearElasticity.cpp
```diff
@@ -197,22 +197,22 @@ CFEANonlinearElasticity::~CFEANonlinearElasticity(void) {
...
+      if (b_Mat_Iso[iVar] != nullptr) delete [] b_Mat_Iso[iVar];
     }
     delete [] b_Mat_Iso;
   }
 
-  if (EField_Ref_Unit  != NULL)   delete [] EField_Ref_Unit;
-  if (EField_Ref_Mod   != NULL)   delete [] EField_Ref_Mod;
-  if (EField_Curr_Unit != NULL)   delete [] EField_Curr_Unit;
+    delete [] EField_Ref_Unit;
+    delete [] EField_Ref_Mod;
```
pcarruscag, line 214: more alignment, I'm guessing you have a script for this that can be fine tuned?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -180,12 +180,12 @@ COutput::~COutput(void) {
...
 
-  if (volumeDataSorter != nullptr)
+  
     delete volumeDataSorter;
 
   volumeDataSorter = nullptr;
 
-  if (surfaceDataSorter != nullptr)
+  
     delete surfaceDataSorter;
```
pcarruscag, line 189: more alignment.

economon:
Fixed up the alignment issues, but I really think it is time that we move to an auto-formatter as a project (clang-format or another tool). It is not sustainable in a large project such as this to manually manage it.

We can do it as a pre-commit hook, which is a bit harsh as it forces everyone to install the formatter and could cause issues at commit time for folks, or we could also do it as a GitHub action perhaps. For instance, ninja can already do this for us as a one liner: install clang-format, drop a .clang-format file into the root SU2/ directory, configure a build, cd into build folder and ```$ ninja clang-format``` = entire project auto-formatted into a style of our choosing.

talbring:
I like the idea of doing the formatting with ninja! We just have to decide on a format. Maybe someone can prepare a base clang-format file and then we discuss?

economon:
> I like the idea of doing the formatting with ninja! We just have to decide on a format. Maybe someone can prepare a base clang-format file and then we discuss?

That's easy :). Try this in your .clang-format:

```
BasedOnStyle: Google
PointerAlignment: Left
DerivePointerAlignment: false
```

pcarruscag:
Thanks for fixing the alignment.
With those options the code actually looks ok in areas that were recently cleaned up.
But the need for the manual formatting we have in loads of places comes from having code that is far too verbose and repetitive e.g. 10 calls to MPI routines with tons of arguments, 10 operator [] in a single line to get to the CSolver we need, entire copy pasted sections that should be functions, etc..
I am not a fan of that kind of wholesale automatic change because it does not fix any of the structural problems we have, and it tends to make the fixing of those issues harder because the aforementioned problematic areas are usually less readable after such changes.
I like the optimization approach of making small gradual improvements, but unfortunately this kind of maintenance is not very glamorous and few of us are up for it...

economon:
I understand.. some of the lines are going to look very ugly after wrapping to 80 column widths, etc, but it will also highlight these trouble spots.

I see it as two orthogonal issues: (1) fix the formatting problems once and for all and have them automatically fixed in the future so we stop wasting time on it, and (2) we continue to fix the problem areas in the code structure as we go, which we will need to do anyway, regardless.

For me, I think the benefits of (1) outweigh the negative of seeing some ugly code.

economon:
I have placed a .clang-format file in the root of the repo for anyone to use with clang-format. I recommend folks play around with it and apply it as needed on the files you change in future PRs. Eventually, we will enforce its use through a pre-commit hook or GitHub action.

p.s. bonus for VS Code users, this .clang-format will be used by default when using the auto-format shortcut (shift+option+F on Mac). Might check whether this is the case for other IDEs.

PR #974, opened 2020-05-09
Linear solver history/screen output for discrete adjoint solvers

pcarruscag:
## Proposed Changes
Linear solver history/screen output for discrete adjoint solvers

## PR Checklist

### DIFF
--- a/SU2_CFD/src/output/CFlowCompFEMOutput.cpp
+++ b/SU2_CFD/src/output/CFlowCompFEMOutput.cpp
@@ -115,8 +115,9 @@ void CFlowCompFEMOutput::SetHistoryOutputFields(CConfig *config){
   AddHistoryOutput("MAX_ENERGY",     "max[RhoE]", ScreenOutputFormat::FIXED,   "MAX_RES", "Maximum residual of the energy.", HistoryFieldType::RESIDUAL);
   /// END_GROUP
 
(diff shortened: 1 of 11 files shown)


SU2_CFD/src/output/CFlowCompFEMOutput.cpp
```diff
@@ -115,8 +115,9 @@ void CFlowCompFEMOutput::SetHistoryOutputFields(CConfig *config){
   AddHistoryOutput("MAX_ENERGY",     "max[RhoE]", ScreenOutputFormat::FIXED,   "MAX_RES", "Maximum residual of the energy.", HistoryFieldType::RESIDUAL);
   /// END_GROUP
 
-  /// DESCRIPTION: Linear solver iterations
-  AddHistoryOutput("LINSOL_ITER", "Linear_Solver_Iterations", ScreenOutputFormat::INTEGER, "LINSOL_ITER", "Number of iterations of the linear solver.");
+  /// DESCRIPTION: Linear solver iterations and residual
```
economon, line 118: I think these can be removed from the DG solver output, as it is entirely explicit in the current implementation

pcarruscag: Thought so too, but maybe they were being recycled for something else, they're gone now.

PR #975, opened 2020-05-09
Hybrid parallel boundary conditions for turbulence and compressible flow solvers

pcarruscag:
## Proposed Changes
So far BC's were done by a single thread, this PR addresses that which required making all that code thread-safe. There is also some duplication-avoiding cleanup.
A batch of regression tests for hybrid parallel (compilation + execution) was added. Boundary conditions are perhaps the easiest places to introduce non-thread-safe code, due to the common pattern of writing into temporary arrays of CSolver (e.g. Solution_i), thus it becomes important to have some tests in place.

## Related Work
Continuation of #789, so far as compressible flow goes we now have "full coverage", the only major areas left out (that matter to scalability) are MPI and parts of the initial preprocessing.

## PR Checklist


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2727,80 +2722,58 @@ void CSolver::SetResidual_BGS(CGeometry *geometry, CConfig *config) {
...
+  SU2_OMP_FOR_STAT(roundUpDiv(nPointDomain,2*omp_get_max_threads()))
+  for (auto iPoint = 0ul; iPoint < nPointDomain; ++iPoint) {
 
-    /*--- Solution at each edge point ---*/
-
-    su2double *Solution_i = base_nodes->GetSolution(iPoint);
-    su2double *Solution_j = base_nodes->GetSolution(jPoint);
-
-    for (iVar = 0; iVar < nVar; iVar++)
-      Solution[iVar] = 0.5* (Solution_i[iVar] + Solution_j[iVar]);
```
pcarruscag, line 2748: This function (`CSolver::SetRotatingFrame_GCL`) was averaging the solution but then updating the iPoint / jPoint residuals with Solution_i / j instead, if anyone knows of a good reason for that let me know and I'll put it back, otherwise "it's fixed".

pcarruscag: Actually this apparent fix makes one of the turbomachinery cases fail, so I've put the apparent bug back.

Common/include/CConfig.hpp
```diff
@@ -8340,26 +8340,6 @@ class CConfig {
...
-  void SetSpline(vector<su2double> &x, vector<su2double> &y, unsigned long n, su2double yp1, su2double ypn, vector<su2double> &y2);
-
-  /*!
-   * \brief Given the arrays xa[1..n] and ya[1..n], which tabulate a function (with the xai’s in order),
-   and given the array y2a[1..n], which is the output from spline above, and given a value of
-   x, this routine returns a cubic-spline interpolated value y.
-   Numerical Recipes: The Art of Scientific Computing, Third Edition in C++.
-   * \return The interpolated value of for x.
-   */
-  su2double GetSpline(vector<su2double> &xa, vector<su2double> &ya, vector<su2double> &y2a, unsigned long n, su2double x);
```
pcarruscag, line 8361: This code was unused and duplicated in CGeometry.

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -76,21 +76,21 @@ void CIntegration::Space_Integration(CGeometry *geometry,
   if (dual_time)
     solver_container[MainSolver]->SetResidual_DualTime(geometry, solver_container, config, iRKStep, iMesh, RunTime_EqSystem);
 
-  /// TODO: No boundary condition supports hybrid parallelism yet, master thread does all the work.
+  /*--- Pick convective and viscous numerics objects for the current thread. ---*/
 
-  SU2_OMP_MASTER
-  {
+  CNumerics* conv_bound_numerics = numerics[CONV_BOUND_TERM + omp_get_thread_num()*MAX_TERMS];
+  CNumerics* visc_bound_numerics = numerics[VISC_BOUND_TERM + omp_get_thread_num()*MAX_TERMS];
```
pcarruscag, line 82: Usual change no 1: One numerics object per thread.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -169,14 +169,8 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config,
...
   /*--- Define some auxiliar vector related with the residual ---*/
 
-  Residual      = new su2double[nVar]();
   Residual_RMS  = new su2double[nVar]();
   Residual_Max  = new su2double[nVar]();
-  Residual_i    = new su2double[nVar]();
-  Residual_j    = new su2double[nVar]();
-  Res_Conv      = new su2double[nVar]();
-  Res_Visc      = new su2double[nVar]();
-  Res_Sour      = new su2double[nVar]();
```
pcarruscag, line 179: Usual change no 2: Avoid using the small arrays of CSolver.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -7305,17 +7267,23 @@ void CEulerSolver::BC_Riemann(CGeometry *geometry, CSolver **solver_container,
...
   {
     P_Tensor[iVar] = new su2double[nVar];
     invP_Tensor[iVar] = new su2double[nVar];
+    Jacobian_i[iVar] = new su2double[nVar];
   }
 
   /*--- Loop over all the vertices on this boundary marker ---*/
+
+  SU2_OMP_FOR_DYN(OMP_MIN_SIZE)
   for (iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
```
pcarruscag, line 7287: Usual change no 3: Add worksharing directives to for loops.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -941,1120 +940,648 @@ void CNSSolver::Evaluate_ObjFunc(CConfig *config) {
...
-            thetax = theta2 + UnitNormal[0]*UnitNormal[0]/3.0;
-            thetay = theta2 + UnitNormal[1]*UnitNormal[1]/3.0;
-
-            etaz   = UnitNormal[0]*UnitNormal[1]/3.0;
-
-            pix = GridVel[0]*thetax + GridVel[1]*etaz;
-            piy = GridVel[0]*etaz   + GridVel[1]*thetay;
+      AddDynamicGridResidualContribution(iPoint, Point_Normal, geometry, UnitNormal,
+                                         Area, geometry->nodes->GetGridVel(iPoint),
+                                         Jacobian_i, Res_Conv, Res_Visc);
```
pcarruscag, line 1158: Significant change no 1: Find duplicated code and move it to something that can be reused.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -941,1120 +940,648 @@ void CNSSolver::Evaluate_ObjFunc(CConfig *config) {
...
-      if (implicit) {
-        for (iVar = 1; iVar <= nDim; iVar++) {
-          total_index = iPoint*nVar+iVar;
-          Jacobian.DeleteValsRowi(total_index);
-        }
-      }
-    }
-  }
+  BC_Isothermal_Wall_Generic(geometry, solver_container, conv_numerics, nullptr, config, val_marker, true);
 }
```
pcarruscag, line 1395: Significant change no 2: The Isothermal and ConjugateHeat BC's are basically the same thing but the latter gets its wall temperature from a different place, so I moved the difference to a small private function, half the code, half the bugs.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -941,1120 +940,648 @@ void CNSSolver::Evaluate_ObjFunc(CConfig *config) {
...
 
-  su2double Vel[3] = {0.0, 0.0, 0.0}, VelNormal, VelTang[3], VelTangMod, VelInfMod, WallDist[3], WallDistMod;
-  su2double T_Normal, P_Normal;
-  su2double Density_Wall, T_Wall, P_Wall, Lam_Visc_Wall, Tau_Wall = 0.0, Tau_Wall_Old = 0.0;
-  su2double *Coord, *Coord_Normal;
-  su2double diff, Delta;
-  su2double U_Tau, U_Plus, Gam, Beta, Phi, Q, Y_Plus_White, Y_Plus;
-  su2double TauElem[3], TauNormal, TauTangent[3], WallShearStress;
-  su2double Gas_Constant = config->GetGas_ConstantND();
-  su2double Cp = (Gamma / Gamma_Minus_One) * Gas_Constant;
```
pcarruscag, line 1848: Usual change no 5: Define everything as it is used.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -471,99 +432,50 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
-      for (iVar = 0; iVar < nVar; iVar++) {
-        total_index = iPoint*nVar+iVar;
-        Jacobian.DeleteValsRowi(total_index);
-      }
-
-    }
-  }
+  BC_HeatFlux_Wall(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
 
 }
```
pcarruscag, line 437: ...

.github/workflows/regression.yml
```diff
@@ -85,12 +87,36 @@ jobs:
           cp -r ${{format('Base{0}', matrix.tag)}}/. install/bin/
           cp -r ${{format('Reverse{0}', matrix.tag)}}/. install/bin/
           cp -r ${{format('Forward{0}', matrix.tag)}}/. install/bin/
-          chmod a+x install/bin/*               
+          chmod a+x install/bin/*
+      - name: Run Tests in Container
+        uses: docker://su2code/test-su2:20200303
+        with:
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+  regression_tests_omp:
```
talbring, line 95: I am gonna simplify the script a little bit. I don't think its necessary to copy all that stuff. You are only doing it because the Reverse/Forward binaries are not available for the BaseOMP case, right? 

I am also surprised that OpenMP is installed in the docker container ...

pcarruscag: Yep.
I don't think one installs OpenMP, it comes with the compiler.

.github/workflows/regression.yml
```diff
@@ -68,50 +68,17 @@ jobs:
             tag: NoMPI
           - testscript: 'serial_regression_AD.py'
             tag: NoMPI
-    steps:
-      - name: Download Base
-        uses: actions/download-artifact@v1
-        with:
-          name: ${{format('Base{0}', matrix.tag)}}
```
pcarruscag: :+1: Thanks for that!

pcarruscag:
@talbring files have been moved

SU2_CFD/src/solvers/CSolver.cpp
economon: Can you make an issue for this please with a link to the spot in the code? I think this could be a bug too but need to look through it...

Issue #949, opened 2020-04-24
Failure to Restart on Large Meshes with Time-Averaged Data


GomerOfDoom:
[sd2_case1b_ddes_v7.cfg.txt](https://github.com/su2code/SU2/files/4531160/sd2_case1b_ddes_v7.cfg.txt)

Hello,

We are currently unable to restart SU2 in DDES mode from a restart file that includes time-averaged data on very large (~180 million cell) meshes.

Compiled in release mode, the code gives the error "FGMRES orthogonalization failed, linear solver diverged."

Compiled in debug mode, the code issues an assertion failure at line 1881 of $SU2_HOME/SU2_CFD/src/numerics_structure.cpp, which is a check in the CNumerics::SetRoe_Dissipation(...) method to make sure that variable 'Dissipation_j' is between zero and one.

This problem only appears when attempting to restart from solution files that include "TIME_AVERAGE" data on very large meshes.

Note the above behavior is occurring with commit 382e82f5e43fadb5d64e63289db8e125e37ffe95 of the "develop" branch. 

I have pulled the latest commits of develop (c093a62a2b) and master (d9c867d342), but get segfaults during Jacobian structure initialization when attempting to restart on multiple cores.

All help is appreciated.

-Paul

**To Reproduce**
Config file attached, but mesh file is quite large... 17.6 GB.

**Desktop (please complete the following information):**
 - Department of Defense Unclassified System: "Onyx"
 - System type: Cray XC40/50
 - OS: Variant of SuSe Linux 12.3 and/or Cray Linux Environment
 - Compiler: Intel 19.0.1.144
 - MPI: cray-mpich 7.6.3
 - SU2 v. 7.0.1, develop branch, commit 382e82f5e43fadb5d64e63289db8e125e37ffe95 (and segfaults with latest commits of develop (c093a62a2b) and master (d9c867d342) ).

economon:
Hey Paul! Sorry to hear this is an issue for you. ~180 mio cells is quite large... any chance you can recreate on something much smaller? If you roll back some commits, does it work again? If you can find the exact commit, that would be helpful.

On a smaller case, you might check whether the time averaged data is being written as you expect to the restart file by using an ASCII file and viewing it (add RESTART_ASCII to your OUTPUT_FILES). It is just a normal CSV. Might give you some clues. 

[...]

PR #980, opened 2020-05-11
Fix large mesh

GomerOfDoom:
## Proposed Changes
Changes have been made to accomodate very large meshes/data sets.
MPI calls were changed to accomodate values over the limit of INT_MAX, and some standard int variables were changed to unsigned long variables.

## Related Work
Addresses problems raised in issue #949 

## PR Checklist


SU2_CFD/src/output/filewriter/CParaviewXMLFileWriter.cpp
```diff
@@ -330,7 +330,8 @@ void CParaviewXMLFileWriter::WriteDataArray(void* data, VTKDatatype type, unsign
                                             unsigned long globalSize, unsigned long offset){
 
 
-  int totalByteSize, byteSize;
+  unsigned long totalByteSize;
+  int byteSize;
```
economon, line 334: while you're at it, I would just move the declaration of these two variables down to wherever they are used (rather than up here at the top)

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3930,16 +3930,21 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, CConfig *config, stri
...
 
   int *blocklen = new int[geometry->GetnPointDomain()];
-  int *displace = new int[geometry->GetnPointDomain()];
+  MPI_Aint *displace = new MPI_Aint[geometry->GetnPointDomain()];
   int counter = 0;
   for (iPoint_Global = 0; iPoint_Global < geometry->GetGlobal_nPointDomain(); iPoint_Global++ ) {
     if (geometry->GetGlobal_to_Local_Point(iPoint_Global) > -1) {
       blocklen[counter] = nFields;
-      displace[counter] = iPoint_Global*nFields;
+      //if (iPoint_Global*nFields > INT_MAX){
```
economon, line 3938: Go ahead and remove the commented warning here since it is no longer needed

PR #982, opened 2020-05-12
Modules Framework Proposal

talbring:
## Proposed Changes

This is a quick draft for a module framework. Note that this is a draft and this is just to get some ideas and comments whether something like that would be useful.

### Motivation

Right now we have a lot of routines scattered throughout the code that essentially functionally belong together as they perform the same overall task. This leads to a lot of if-conditions and special cases we have to consider and its hard to actually see what their overall task actually is. One example that I also used to demonstrate a possible solution is the Fixed CL computation. We have routines in CIteration and in CEulerSolver and they modify certain behavior of the baseline code. But its complicated to figure out which routines actually belong together, what they do and when they are called.

### Possible solution

To solve that problem I propose a modular structure that puts all routines that are responsible for a certain functional task together in the same class. At certain predefined places in the code they can modify things by overloading certain functions. These functions are called *Hooks*. What arguments these hooks provide can be different depending on where they are called.  As an example I did a quick implementation of such a framework and defined a Fixed CL module. 

Right now, I defined a global list of modules (with only one module currently). We can also define modules lists per class, e.g. for CDriver, CIteration, CSolver. But that would mean that we have to provide potentially more than one class per module, which kind of defeats the purpose again.

I will add some further comments below.

## PR Checklist

### DIFF
--- a/Common/include/toolboxes/CModule.hpp
+++ b/Common/include/toolboxes/CModule.hpp
@@ -1,0 +1,48 @@
+#include <tuple>
+#include <utility>
+#pragma once
+
+
+#define CREATE_ACTION(name, func_args...) \
+  static struct Action##name {} Action##name; \
+  template <typename type, typename... Targs> \
+  static void Action(type& Module, struct Action##name, Targs& ...args){ Module.name(args...); }
+
+
+template<class DerivedModule>
+class CModule {
+
+protected:
+
+  bool enabled;
+
+public:
+
+  CModule(bool enabled_ = true) :
+    enabled(enabled_){}
+
+  bool IsEnabled() { return enabled; }
+
+  template<std::size_t I = 0, typename TAction, typename... Tp, typename... Targs>
+  inline static typename std::enable_if<I == sizeof...(Tp), void>::type
+  for_each(std::tuple<Tp...> &, TAction, Targs&...) { }
+
+  template<std::size_t I = 0, typename TAction, typename... Tp, typename... Targs>
+  inline static typename std::enable_if<I < sizeof...(Tp), void>::type
+  for_each(std::tuple<Tp...>& t, TAction action, Targs&... args)
+  {
+    if (std::get<I>(t).IsEnabled()){
+      DerivedModule::Action(std::get<I>(t), action, args...);
+    }
+    for_each<I + 1, TAction, Tp...>(t, action, args...);
+  }
+
+};
+
+
+template<typename T, typename... Modules>
+class CModuleList : public std::tuple<Modules...>{
+public:
+  explicit CModuleList(T* t): std::tuple<Modules...>(Modules(t)...){};
+};
+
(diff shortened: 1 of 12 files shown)


Common/include/toolboxes/CModule.hpp
```diff
@@ -0,0 +1,48 @@
...
+public:
+
+  CModule(bool enabled_ = true) :
+    enabled(enabled_){}
+
+  bool IsEnabled() { return enabled; }
+
+  template<std::size_t I = 0, typename TAction, typename... Tp, typename... Targs>
+  inline static typename std::enable_if<I == sizeof...(Tp), void>::type
+  for_each(std::tuple<Tp...> &, TAction, Targs&...) { }
```
talbring, line 28: I opted for a compile-time construction of the iteration over the module hooks. But that is up or discussion. Can be very well also be done dynamically.

SU2_CFD/include/iteration/modules/CSolverModule.hpp
```diff
@@ -0,0 +1,19 @@
...
+class CConfig;
+class CSolver;
+class COutput;
+
+class CSolverModule : public CModule<CSolverModule>{
+
+public:
+  CSolverModule(bool enabled_ = true) : CModule(enabled_) {};
+
+  virtual void PreIterationHook(COutput *output, CConfig* config, CGeometry* geometry, CSolver** solver) {};
```
talbring, line 14: Its not directly necessary here to define the functions virtual, but I makes the interface more clear.

SU2_CFD/include/iteration/modules/CFixedCLModule.hpp
```diff
@@ -0,0 +1,36 @@
...
+  su2double dCL_dAlpha;
+  su2double AoA_inc;
+
+  bool Start_AoA_FD, End_AoA_FD;
+  unsigned long Iter_Update_AoA;
+
+
+public:
+
+  explicit CFixedCLModule(CConfig *config) : CSolverModule(config->GetFixed_CL_Mode()) {};
```
talbring, line 22: This is the only place where the `CConfig::GetFixed_CL_Mode()` is called.

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -664,13 +655,28 @@ void CFluidIteration::Solve(COutput *output,
 
     config[val_iZone]->SetInnerIter(Inner_Iter);
 
+    CSolverModule::for_each(*modules, CSolverModule::ActionPreIterationHook, output, config[val_iZone],
+                            geometry[val_iZone][val_iInst][MESH_0], solver[val_iZone][val_iInst][MESH_0]);
```
talbring, line 659: Thats the call that iterates (at compile time) through all modules and calls their respective hooks.

SU2_CFD/include/iteration/modules/SolverModules.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include "CFixedCLModule.hpp"
+
+using Modules    = CModuleList<CConfig, CFixedCLModule>;
```
talbring, line 5: Here we define what modules we actually have (the CConfig is just used to define the argument for the constructor of the modules, can be also moved to the base class of CModuleList.

Issue #978, opened 2020-05-11
MOMENTUM_DISTORTION definition


auzbaig:
Can anyone give an explanation or a reference of how the momentum distortion is defined in the CFlowOutput.cpp file?

It seems to be a ratio of the normal velocity over area-weighted normal velocity squared, however, looking at a mathematical equation will make it look more plausible.

The nomenclature in the file can be confusing sometimes.

Thanks.

auzbaig:
My general suggestion is that in the future there could be a document to show how all of the output quantities are defined.

economon:
The uniformity / distortion parameters are those defined by Banko et al. in Sec. 3.2 of https://www.sciencedirect.com/science/article/pii/S0142727X16301412

auzbaig:
Thanks @economon. That exactly answers my question!

economon:
👍 If you have a minute, perhaps it would help to embed the link to the paper in a comment near the definition for the future

auzbaig:
You mean here?
https://github.com/su2code/SU2/blob/20888f488752e4e8aaa5230e56c5417923c7c344/SU2_CFD/src/output/CFlowOutput.cpp#L513

auzbaig:
or this is actually better:
https://github.com/su2code/SU2/blob/20888f488752e4e8aaa5230e56c5417923c7c344/SU2_CFD/src/output/CFlowOutput.cpp#L453

economon:
They are first mentioned here (some of the factors are first computed there): https://github.com/su2code/SU2/blob/20888f488752e4e8aaa5230e56c5417923c7c344/SU2_CFD/src/output/CFlowOutput.cpp#L273

You could put it in the comments above those two sections where they are computed if you like

auzbaig:
Ok, I will try to do it.

PR #985, opened 2020-05-14
adding reference for distortion and uniformity parameters

auzbaig:
Following up from #978


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -448,7 +448,10 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
     /*--- Compute flow uniformity parameters separately (always area for now). ---*/
 
     Area = fabs(Surface_Area_Total[iMarker_Analyze]);
-
+    
+    /*--- The definitions for Distortion and Uniformity Parameters is taken as defined by Banko, Andrew J., et al. in section 3.2 of
```
pcarruscag, line 452: "are taken"

auzbaig: Oops, corrected that.

PR #989, opened 2020-05-16
Mixed precision linear algebra

pcarruscag:
## Proposed Changes
Compile-time option (`-Denable-mixedprec=true`) to have single precision linear solves while keeping all other calculations in double precision. Which means that among other things we get the ILU for the price of LU-SGS (memory wise).
This uses the same type-mixing logic used to have a passive matrix in reverse AD builds (thus this feature is also compatible with reverse AD, but not direct).
For the 3D RANS SST case with about 500k nodes I've been using for benchmarking and profiling this gives a 1.19 speedup:
image
This case is very light on the linear solvers (CFL of 20) so this result is actually really good.
There seems to be no compromise in accuracy, the residuals eventually go below -11 (8 orders reduction) in the same number of iterations, and the mesh metrics for this case are not stellar:
```
+--------------------------------------------------------------+
|           Mesh Quality Metric|        Minimum|        Maximum|
+--------------------------------------------------------------+
|    Orthogonality Angle (deg.)|        36.7556|        89.9375|
|     CV Face Area Aspect Ratio|        1.03036|        4477.19|
|           CV Sub-Volume Ratio|        1.00693|        97.7189|
+--------------------------------------------------------------+
```
For purely linear applications like mesh deformation (including SU2_DOT) or linear elasticity, for which the solution is not iterative, an impact on accuracy is expected.
For example deforming a grid of 830k nodes (with stretched cells and all that good stuff) with SU2_DEF using CG+ILU converges like this:
image
So it is still possible to converge the linear system to very tight tolerances, but after a certain tolerance the convergence rate becomes worse, that threshold will depend on the size of the problem (larger -> harder to solve) or more formally on the condition number of the matrix.
For this example SU2_DEF uses 25% less memory in single precision than in double (or double requires 33% more), and solves to 1e-10 in the same time (as it requires more iterations).

## Related Work
#789

## PR Checklist


Common/src/CMultiGridQueue.cpp
```diff
@@ -164,7 +166,8 @@ bool CMultiGridQueue::EmptyQueue(void) const {
 
   if (QueueCV.size() == 1) {
     for (auto iPoint : QueueCV[0])
-      if (RightCV[iPoint]) return false;
+      if ((iPoint != QueueType::ErasedValue) && RightCV[iPoint])
+        return false;
```
pcarruscag, line 170: This is not related to this PR, it's a fix for a problem that @economon found.

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -76,7 +76,7 @@ CNumerics::ResidualType<> CCentBase_Flow::ComputeResidual(const CConfig* config)
   if (preacc) {
     AD::SetPreaccIn(Normal, nDim);
     AD::SetPreaccIn(V_i, nDim+5); AD::SetPreaccIn(V_j, nDim+5);
-    AD::SetPreaccIn(Lambda_i);    AD::SetPreaccIn(Lambda_j);
+    AD::SetPreaccIn(Lambda_i, Lambda_j);
```
pcarruscag, line 79: ad_structure done as per #583 
I also tried to add some syntax sugar to be able to specify preaccumulation inputs and outputs like this.

Common/include/datatype_structure.hpp
```diff
@@ -31,92 +30,190 @@
...
+  FORCEINLINE void SetDerivative(su2double &, const passivedouble &) {}
+#endif
 
   /*!
    * \brief Casts the primitive value to int (uses GetValue, already implemented for each type).
    * \param[in] data - The non-primitive datatype.
    * \return - The primary value casted to int.
    */
-  int Int(const su2double& data);
+  FORCEINLINE int Int(const su2double& data) {return int(SU2_TYPE::GetValue(data));}
```
economon, line 186: prefer static_cast over C-style casts here..

Common/include/datatype_structure.hpp
```diff
@@ -31,92 +30,190 @@
...
   /*!
    * \brief Casts the primitive value to short (uses GetValue, already implemented for each type).
    * \param[in] data - The non-primitive datatype.
    * \return - The primary value casted to short.
    */
-  short Short(const su2double& data);
-}
-
-#include "datatype_structure.inl"
+  FORCEINLINE short Short(const su2double& data) {return short(SU2_TYPE::GetValue(data));}
```
economon, line 193: and here. Check for others if you have a min

Common/include/datatype_structure.hpp
```diff
@@ -31,92 +30,190 @@
...
+   * \param[in] args - Values to be printed to the string.
+   */
+  template<class... Ts>
+  FORCEINLINE void sprintf(char* str, const char* format, Ts&&... args) {
+    ::sprintf(str, format, SU2_TYPE::_printGetValue(args)...);
+  }
+  FORCEINLINE void sprintf(char* str, const char* literal) {
+    ::sprintf(str, "%s", literal);
+  }
+#define SPRINTF SU2_TYPE::sprintf
```
economon, line 217: 👍 nice clean up

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -630,7 +594,7 @@ class CSysMatrix {
    * \param[in] block_i - Diagonal index.
    * \param[in] val_block - Block to add to the diagonal of the matrix.
    */
-  template<class OtherType>
+  template<class OtherType, int Sign = 1>
```
economon, line 597: I think this default template arg might be better as an input arg to the function.. I could see this causing confusion. Same with keeping the name AddBlock2Diag and using for subtraction. I like the reuse you introduced here, but can you make it a little clearer?

Common/src/CMultiGridQueue.cpp
economon: 👍

Common/src/grid_movement_structure.cpp
```diff
@@ -127,16 +125,13 @@ void CVolumetricMovement::UpdateMultiGrid(CGeometry **geometry, CConfig *config)
...
+  unsigned long Tot_Iter = 0;
+  su2double MinVolume, MaxVolume;
 
   /*--- Retrieve number or iterations, tol, output, etc. from config ---*/
 
-  Smoothing_Iter = config->GetDeform_Linear_Solver_Iter();
-  Screen_Output  = config->GetDeform_Output();
-  NumError       = config->GetDeform_Linear_Solver_Error();
-  Nonlinear_Iter = config->GetGridDef_Nonlinear_Iter();
+  auto Screen_Output  = config->GetDeform_Output();
```
economon, line 133: This is a matter of taste, but in situations like this where it isn't clear what the type being returned is from the method name alone (or if it doesn't save you a ton of keystrokes), I would avoid auto and stick with bool to be more clear.

Common/include/ad_structure.hpp
```diff
@@ -145,99 +149,348 @@ namespace AD{
...
 
   /*!
    * \brief Pushes back the current tape position to the tape position's vector.
    */
-  void Push_TapePosition();
-}
+  inline void Push_TapePosition() {}
+
+#else
+  typedef codi::DataStore CheckpointHandler;
```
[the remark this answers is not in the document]
pcarruscag: Ah yes I did it for datatype_structure...

PR #1001, opened 2020-05-24
Move FEM files

pcarruscag:
## Proposed Changes
Move fem_* files in Common to Common/fem/. Move inline function to the hpp.
I've also deleted some unnecessary functions that compilers can generate automatically.
I did not split the files to minimize conflicts (this way the files just show as renamed) and also because I don't have time to figure out how this should be organized.

## Related Work
#583

## PR Checklist


pcarruscag:
Sounds like a good compromise, no point doing the same work twice.

PR #1008, opened 2020-05-29
Radial Equilibrium equation fix for Turbo solver

auzbaig:
It appears that was some mistake when the Radial Equilibrium equation was defined. From the equation below:
image

The Vtheta and R in Vt2/R (Vtheta^2/R above) need to be at the same span. 
In the 0->midspan SU2 is currently doing Vt2(i-1)/R(i) whereas it should be Vt2(i-1)/R(i-1). I have made the corrections in CEulerSolver.cpp.

Please cross-check my changes and let me know if these changes need to made somewhere else as well.

Thanks,
Aman


auzbaig:
yes I have. This is the only place where it is being defined and used (in CEulerSolver.cpp)

pcarruscag:
I see, is there a chance of dividing by zero here? I recall @koodlyakshay fixing a bug like that some time ago.
Maybe @arubino, @salvovitale could pitch in.

auzbaig:
The only way `Radius1 ` could be zero would be if the very first span lies on the axis of rotation which does not make sense for a Turbomachinery case. There is always some hub radius. Or unless you have a reference axis of rotation other than (0,0,z), where `Radius2` could also become zero at some point.

pcarruscag:
Having worked in the industry I guarantee you people run simulations with zero hub radius.

auzbaig:
In my runs I am seeing very minor difference in the resultant outlet profiles with either Radius1 or Radius2. However, for the sake of being accurate numerically we may want to correct this or include an if statement to use `Radius2` when `Radius1` is zero.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -12446,7 +12446,12 @@ void CEulerSolver::TurboAverageProcess(CSolver **solver, CGeometry *geometry, CC
                   Radius2    = geometry->GetTurboRadius(iMarker,iSpan);
                   Radius1    = geometry->GetTurboRadius(iMarker,iSpan-1);
                   Vt2        = AverageTurboVelocity[iMarker][iSpan - 1][1]*AverageTurboVelocity[iMarker][iSpan - 1][1];
-                  RadialEquilibriumPressure[iMarker][iSpan -1] =  RadialEquilibriumPressure[iMarker][iSpan] - AverageDensity[iMarker][iSpan -1]*Vt2/Radius1*(Radius2 - Radius1);
+                  if (Radius1 < 1E-15){
+                  RadialEquilibriumPressure[iMarker][iSpan -1] =  RadialEquilibriumPressure[iMarker][iSpan] - AverageDensity[iMarker][iSpan -1]*Vt2/Radius2*(Radius2 - Radius1);
+                  }
+                  else{
+                    RadialEquilibriumPressure[iMarker][iSpan -1] =  RadialEquilibriumPressure[iMarker][iSpan] - AverageDensity[iMarker][iSpan -1]*Vt2/Radius1*(Radius2 - Radius1);
+                  }
```
pcarruscag: c'mon... not like this...
Avoid if's inside inner loops, not good for performance, please don't repeat code, creates bugs, and don't use "magic numbers" (1e-15) there is an EPS defined in option_structure.hpp that can be used for this kind of generic check.
```
...
Radius1 = geometry->GetTurboRadius(iMarker,iSpan-1);
Radius1 = (Radius1 > EPS)? Radius1 : Radius2;
...
```
Sorry for asking for changes after approving.

auzbaig: lol I knew you suggested something like that's why I didn't merge it. I didn't think of this syntax so I went the crude way. I fixed it now.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -12446,7 +12446,8 @@ void CEulerSolver::TurboAverageProcess(CSolver **solver, CGeometry *geometry, CC
                   Radius2    = geometry->GetTurboRadius(iMarker,iSpan);
                   Radius1    = geometry->GetTurboRadius(iMarker,iSpan-1);
                   Vt2        = AverageTurboVelocity[iMarker][iSpan - 1][1]*AverageTurboVelocity[iMarker][iSpan - 1][1];
-                  RadialEquilibriumPressure[iMarker][iSpan -1] =  RadialEquilibriumPressure[iMarker][iSpan] - AverageDensity[iMarker][iSpan -1]*Vt2/Radius2*(Radius2 - Radius1);
+                  Radius1    = (Radius1 > EPS)? Radius1 : Radius2;
+                  RadialEquilibriumPressure[iMarker][iSpan -1] =  RadialEquilibriumPressure[iMarker][iSpan] - AverageDensity[iMarker][iSpan -1]*Vt2/Radius1*(Radius2 - Radius1);
```
pcarruscag, line 12450: Thanks :+1:

PR #1009, opened 2020-05-30
Hybrid parallel MPI

### LINKED ISSUE #1011: Wrong output from meson
**Describe the bug**
When I run the meson script for building it gives the wrong build command. This is both for the master and develop version.

**To Reproduce**
Simply type
./meson.py build

One of the last output lines is
Use './ninja -C false install' to compile and install SU2

while it should be
Use './ninja -C build install' to compile and install SU2

 - OS: Ubuntu 18.04
 - C++ compiler and version: GCC 8.4.0
 - MPI implementation and version: Sequential
 - SU2 Version: 7.0.5

pcarruscag:
## Proposed Changes
This makes the routines InitiateComms and CompleteComms (and periodic counterparts) safe to call in parallel, until now they had to be guarded inside SU2_OMP_MASTER sections.
The calls to MPI are still only made by the master thread (funneled communication) but the buffers are packed and unpacked by all threads.
I also made a slight change which seems to make communications more efficient, we were always communicating the entire buffer, which is sized for the maximum number of variables per point, because the data was packed like this:
`o o o o _ _ _ _ o o o o _ ...` (count = 4, maxCount = 8);
I changed it to:
`o o o o  o o o o ... _ _ _ _ ...`
Which allows only part of the buffer to be communicated.
(The maximum size `nPrimVarGrad*nDim*2` is actually quite large compared to the median `nVar`)

In the process I also had to make a few more CGeometry routines thread safe.

## Related Work
#789
Resolves #1011

## PR Checklist


Common/include/geometry/CGeometry.hpp
```diff
@@ -293,20 +293,24 @@ class CGeometry {
    * \param[in] geometry - Geometrical definition of the problem.
    * \param[in] config   - Definition of the particular problem.
    * \param[in] commType - Enumerated type for the quantity to be communicated.
-   * \param[in] val_reverse  - Boolean controlling forward or reverse communication between neighbors.
+   * \param[in] countPerPoint - Number of variables per point.
+   * \param[in] val_reverse - Boolean controlling forward or reverse communication between neighbors.
    */
-  void PostP2PRecvs(CGeometry *geometry, CConfig *config, unsigned short commType, bool val_reverse) const;
+  void PostP2PRecvs(CGeometry *geometry, CConfig *config, unsigned short commType,
+                    unsigned short countPerPoint, bool val_reverse) const;
```
pcarruscag, line 300: A consequence of this is that the current count-per-point needs to be explicitly passed to a few routines, whereas before we could always rely on the maximum value.

pcarruscag:
This is the sort of scalability (in terms of time to solution, not per iteration) we get now:
image
Which you can directly compare with #861

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -499,33 +498,37 @@ void CMeshSolver::DeformMesh(CGeometry **geometry, CNumerics **numerics, CConfig
...
   if (time_domain) ComputeGridVelocity(geometry[MESH_0], config);
 
   /*--- Update the multigrid structure. ---*/
   UpdateMultiGrid(geometry, config);
 
+  } // end parallel
+
+  /*--- Check for failed deformation (negative volumes). ---*/
+  /*--- This is not recorded as it does not influence the solution. ---*/
+  AD::StopRecording();
```
talbring, line 520: You can use AD_BEGIN_PASSIVE/AD_END_PASSIVE have passive regions. Although I am not a big fan of preprocessor macros and maybe we find another way soon, but at least it is consistent with the rest for now :)

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1343,24 +1130,7 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
-                                          rotMatrix[0][1]*base_nodes->GetLimiter_Primitive(iPoint,2) +
-                                          rotMatrix[0][2]*base_nodes->GetLimiter_Primitive(iPoint,3));
-                bufDSend[buf_offset+2] = (rotMatrix[1][0]*base_nodes->GetLimiter_Primitive(iPoint,1) +
-                                          rotMatrix[1][1]*base_nodes->GetLimiter_Primitive(iPoint,2) +
-                                          rotMatrix[1][2]*base_nodes->GetLimiter_Primitive(iPoint,3));
-                bufDSend[buf_offset+3] = (rotMatrix[2][0]*base_nodes->GetLimiter_Primitive(iPoint,1) +
-                                          rotMatrix[2][1]*base_nodes->GetLimiter_Primitive(iPoint,2) +
-                                          rotMatrix[2][2]*base_nodes->GetLimiter_Primitive(iPoint,3));
-              }
+              Rotate(zeros, &base_nodes->GetLimiter_Primitive(iPoint)[1], &bufDSend[buf_offset+1]);
```
talbring, line 1133: :+1:

Common/src/linear_algebra/CSysSolve_b.cpp
```diff
@@ -25,6 +25,7 @@
  * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "../../include/CConfig.hpp"
```
talbring, line 28: Why did you add that include?

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: Yeah I thought of making it a function of the AD namespace, maybe I'll do it now :)

Common/src/linear_algebra/CSysSolve_b.cpp
pcarruscag: I moved it from the hpp, maybe it is not needed at all, I'll check.

talbring: Ah I didn't see that :) yep, then its definitely needed.

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: Made a function and replaced both this kind of logic and the macro.

Common/src/linear_algebra/CSysSolve_b.cpp
pcarruscag: Actually only needed the datatype include not the full config.

Common/src/linear_algebra/CSysSolve_b.cpp
```diff
@@ -36,27 +36,27 @@ void CSysSolve_b<ScalarType>::Solve_b(const codi::RealReverse::Real* x, codi::Re
...
   d->getData(config);
 
-  CSysSolve<ScalarType>* solver = NULL;
+  CSysSolve<ScalarType>* solver = nullptr;
   d->getData(solver);
 
   /*--- Initialize the right-hand side with the gradient of the solution of the primal linear system ---*/
 
-  for (unsigned long i = 0; i < n; i ++) {
+  for (unsigned long i = 0; i < n; i++) {
```
economon, line 59: CSysVector has overloaded =, can we just use (*LinSysSol_b) = 0.0 without the loop? Not sure if that is implemented for a scalar though

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -365,6 +327,63 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
+  su2double r11, r12, r13, r22, r23_a, r23_b, r33, weight;
+  const su2double *center, *angles, *trans;
+  su2double rotMatrix2D[2][2] = {{1.0,0.0},{0.0,1.0}};
+  su2double rotMatrix3D[3][3] = {{1.0,0.0,0.0},{0.0,1.0,0.0},{0.0,0.0,1.0}};
+  su2double rotCoord_i[3] = {0.0}, rotCoord_j[3] = {0.0};
+  su2double translation[3] = {0.0}, distance[3] = {0.0};
+  const su2double zeros[3] = {0.0};
+  su2activematrix Cvector;
+
+  auto Rotate = [&](const su2double* origin, const su2double* direction, su2double* rotated) {
```
economon, line 375: I think it's best practice to avoid universal references in the lambda capture and be explicit instead

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -432,28 +450,22 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
-        rotMatrix[2][0] = -sinPhi;
-
-        rotMatrix[0][1] = sinTheta*sinPhi*cosPsi - cosTheta*sinPsi;
-        rotMatrix[1][1] = sinTheta*sinPhi*sinPsi + cosTheta*cosPsi;
-        rotMatrix[2][1] = sinTheta*cosPhi;
-
-        rotMatrix[0][2] = cosTheta*sinPhi*cosPsi + sinTheta*sinPsi;
-        rotMatrix[1][2] = cosTheta*sinPhi*sinPsi - sinTheta*cosPsi;
-        rotMatrix[2][2] = cosTheta*cosPhi;
+        if (nDim==2) {
```
economon, line 460: can the nDim checks be put inside the function efficiently? Feels like they could be collapsed with an extra arg or template param

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -818,29 +767,13 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
-                                     rotMatrix[0][2]*jacBlock[iVar][2]);
-                rotBlock[iVar][1] = (rotMatrix[1][0]*jacBlock[iVar][0] +
-                                     rotMatrix[1][1]*jacBlock[iVar][1] +
-                                     rotMatrix[1][2]*jacBlock[iVar][2]);
-                rotBlock[iVar][2] = (rotMatrix[2][0]*jacBlock[iVar][0] +
-                                     rotMatrix[2][1]*jacBlock[iVar][1] +
-                                     rotMatrix[2][2]*jacBlock[iVar][2]);
-              }
+            for (iVar = 0; iVar < nVar; iVar++) {
+              Rotate(zeros, jacBlock[iVar], rotBlock[iVar]);
```
economon, line 776: 👍 much safer with Rotate()

Common/include/linear_algebra/CSysSolve.hpp
```diff
@@ -1,7 +1,7 @@
 /*!
- * \file linear_solvers_structure.hpp
+ * \file CSysSolve.hpp
  * \brief Headers for the classes related to linear solvers (CG, FGMRES, etc)
- *        The subroutines and functions are in the <i>linear_solvers_structure.cpp</i> file.
+ *        The subroutines and functions are in the <i>CSysSolve.cpp</i> file.
  * \author J. Hicken, F. Palacios, T. Economon
```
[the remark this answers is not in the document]
pcarruscag: I think I've sneaked it into the class declaration recently xD thanks!

Common/src/geometry/CGeometry.cpp
```diff
@@ -1469,20 +1452,22 @@ void CGeometry::PostPeriodicSends(CGeometry *geometry,
...
-      default:
-        SU2_MPI::Error("Unrecognized data type for periodic MPI comms.",
-                       CURRENT_FUNCTION);
-        break;
-    }
-    iRecv++;
+
+  switch (commType) {
+    case COMM_TYPE_DOUBLE:
+      SU2_OMP_FOR_STAT(1024)
```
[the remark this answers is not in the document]
pcarruscag: yup I'll had a few to CGeometry I don't want to make them too global

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: It is... But it looks a bit clunky.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: The implementation of Rotate determines a template nDim (for unrolling) from the size of the rotation matrix that's why there is a 2D and 3D copy... In short I don't see an easy 0 cost alternative.

Common/include/CConfig.hpp
```diff
@@ -6339,17 +6339,17 @@ class CConfig {
   /*!
    * \brief Center of rotation for a rotational periodic boundary.
    */
-  su2double *GetPeriodicRotCenter(string val_marker);
+  const su2double *GetPeriodicRotCenter(string val_marker) const;
```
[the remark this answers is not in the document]
economon: Whoops.. I misread and didn't see the * (thought it was just su2double), keep as-is

Common/src/linear_algebra/CSysSolve_b.cpp
pcarruscag: Probably slightly better performance if the loop is fused like that.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: I cannot capture nDim without either creating a local copy or capturing "this", might a limitation of c++11 lambdas.

Common/src/geometry/CGeometry.cpp
```diff
@@ -47,97 +47,9 @@
...
-  Local_Point_PeriodicSend = nullptr;
-  Local_Point_PeriodicRecv = nullptr;
-
-  Local_Marker_PeriodicSend = nullptr;
-  Local_Marker_PeriodicRecv = nullptr;
+CGeometry::CGeometry(void) :
+  size(SU2_MPI::GetSize()),
+  rank(SU2_MPI::GetRank()) {
 
 }
```
pcarruscag, line 54: The requested spring cleaning is done, I think I'll stop here.

SU2_CFD/src/solvers/CSolver.cpp
TobiKattmann: Much safer indeed :) :baby_chick: 

Unfortunately the loop here needs to go to nPrimVarGrad. This is the origin of the error we talked about in the SU2 dev meeting yesterday.

TobiKattmann: ironicPrequelMemes
... in order to keep the good tradition of prequel memes :)

PR #1010, opened 2020-05-31
Divides and Moves the Fluid Property Classes

economon:
## Proposed Changes
 
Separates all classes related to the fluid properties and moves them to a new folder. Also cleans up and refactors some of the classes. For example, the viscosity and conductivity classes are now pure abstract classes (interfaces).

I have moved them to a folder named thermophysical/, which makes sense given the current contents (equations of state and transport properties), but once we add chemistry and more advanced models, I am not sure this will make sense. @CatarinaGarbacz, since you are going to be working this area, is there a name you would prefer?

p.s. in a future pass, there is room for improvement in the current CFluidModel class. I think it is best practice not to use the type of implementation inheritance found there. Instead, the EoS classes could be given a dedicated interface class and moved inside the CFluidModel class (just like the viscosity and conductivity classes). Then, the CFluidModel class can simply be a composition of those classes (with no data members) and serve back their data to the client.

## Related Work

#583 

## PR Checklist


SU2_CFD/include/thermophysical/CPolynomialConductivityRANS.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+  /*!
+   * \brief Set thermal conductivity.
+   */
+  void SetConductivity(su2double t, su2double rho, su2double mu_lam, su2double mu_turb, su2double cp) override {
+    /* Evaluate the new kt from the coefficients and temperature. */
+    kt_ = coeffs_[0];
+    for (int i = 1; i < N; ++i) {
+      kt_ += coeffs_[i] * pow(t, i);
+    }
```
pcarruscag, line 74: micro optimization you can compute the powers without the pow function

SU2_CFD/src/thermophysical/CIdealGas.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+CIdealGas::CIdealGas(su2double gamma, su2double R, bool CompEntropy) : CFluidModel() {
+  Gamma = gamma;
+  Gamma_Minus_One = Gamma - 1.0;
+  Gas_Constant = R;
+  Cp = Gamma / Gamma_Minus_One * Gas_Constant;
+  Cv = Cp - R;
+
+  ComputeEntropy = CompEntropy;
+}
```
pcarruscag, line 48: These ctors can be merged by giving CompEntropy a default value

SU2_CFD/src/thermophysical/CPengRobinson.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+  // rtbis algorithm NR
+
+  f = fx1;
+  fmid = fx2;
+  if (f * fmid >= 0.0) {
+    cout << "Root must be bracketed for bisection in rtbis" << endl;
+    SetTDState_rhoT(Density, Temperature);
+  }
+  rtb = f < 0.0 ? (static_cast<void>(dx = x2 - x1), x1) : (static_cast<void>(dx = x1 - x2), x2);
```
pcarruscag, line 258: funny looking code

SU2_CFD/src/thermophysical/CVanDerWaalsGas.cpp
```diff
@@ -0,0 +1,251 @@
...
+  for (int j = 1; j <= NTRY; j++) {
+    if (fx1 * fx2 > 0.0) {
+      if (fabs(fx1) < fabs(fx2)) {
+        x1 += FACTOR * (x1 - x2);
+        fx1 = log(x1 - b) - s / Gas_Constant +
+              log((h + 2 * a / x1) / Gas_Constant / (1 / Gamma_Minus_One + x1 / (x1 - b))) / Gamma_Minus_One;
+      } else {
+        x2 += FACTOR * (x2 - x1);
+        fx2 = log(x2 - b) - s / Gas_Constant +
+              log((h + 2 * a / x2) / Gas_Constant / (1 / Gamma_Minus_One + x2 / (x2 - b))) / Gamma_Minus_One;
```
pcarruscag, line 121: some of this repetition could be cleaned up with lambdas

Common/src/CConfig.cpp
```diff
@@ -1085,7 +1085,7 @@ void CConfig::SetConfig_Options() {
    Cp(T) = b0 + b1*T + b2*T^2 + b3*T^3 + b4*T^4. By default, all coeffs
    are set to zero and will be properly non-dim. in the solver. ---*/
 
-  nPolyCoeffs = 5;
+  nPolyCoeffs = N_POLY_COEFFS;
   default_cp_polycoeffs = new su2double[nPolyCoeffs]();
   default_mu_polycoeffs = new su2double[nPolyCoeffs]();
   default_kt_polycoeffs = new su2double[nPolyCoeffs]();
```
pcarruscag, line 1091: maybe these can be static?

SU2_CFD/include/thermophysical/CFluidModel.hpp
```diff
@@ -0,0 +1,300 @@
...
+
+  /*!
+   * \brief Set viscosity model.
+   */
+  void SetLaminarViscosityModel(CConfig* config);
+
+  /*!
+   * \brief Set thermal conductivity model.
+   */
+  void SetThermalConductivityModel(CConfig* config);
```
pcarruscag, line 220: Please make the config const if possible.

MarcoFossati:
I am fine with this request from a conceptual point of view. Once we complete the alignment with develop and the full integration with Mutation++ we might need to review the config structure according to the discussions we had sometime ago. I am attaching here a snapshot illustrating where we would like to head with it. We will need to review it especially considering the efficiency of the approach that uses Mutation++ but allow introducing a custom mechanism of reaction in CHEMKIN format.
config-01
config-02

CatarinaGarbacz:
@economon thanks for asking. once we add chemistry it would be something that gathers thermo-physico-chemical. I am not sure if there's a good not-too-long term that would have all of these together. 

another possibility is to call this folder something like mixture/, since there are properties that are all computed for a single point in the mesh and instant of time. given the local flow state. and inside mixture/ then we could specify, as you also suggest, transport/ chemistry/ thermodynamics/, etc... let me know what you think

economon:
This one is all set after a little more cleanup. I left some low-level changes to the EoS models for later.

@CatarinaGarbacz @MarcoFossati : we can use a simple directory name, as you suggest. Let's just go with fluid/ since we also treat liquids with the incompressible solver.

Bonus: adds a rotating frame calculation test and some extra error checks on CFL adapt params that should help avoid unreported problems with old configs.

PR #1014, opened 2020-06-02
NEMO solver - simulation of chemically-reactive and nonequilibrium flows

CatarinaGarbacz:
## Proposed Changes
Adding the new solver NEMO to the develop branch. The NEMO solver concerns the simulation of hypersonic flows, accounting for thermal nonequilibrium and finite-rate chemistry, following Park's 2Temperature model. The system of equations to solve is the conservation of each chemical species mass, momentum, total energy and vibrational energy. Both inviscid and viscous simulations are available, respectively, solver NEMO_EULER and NEMO_NAVIER_STOKES.

Most of the changes consisted in the addition of new files (listed below), and all the necessary changes in the high-level classes (drivers, integration, config).
 
new files concerning this solver:
output/CNEMOCompOutput.cpp
variables/CNEMOEulerVariable.cpp
variables/CNEMONSVariable.cpp
solvers/CNEMOEulerSolver.cpp
solvers/CNEMONSSolver.cpp
numerics/NEMO/convection/lax.cpp
numerics/NEMO/convection/roe.cpp
numerics/NEMO/convection/ausm.cpp
numerics/NEMO/convection/ausmplusup2.cpp
numerics/NEMO/convection/ausmpwplus.cpp
numerics/NEMO/convection/msw.cpp
numerics/NEMO/NEMO_diffusion.cpp
numerics/NEMO/NEMO_sources.cpp

Future/ongoing work is looking into making use of the CFluidModel class , introducing a new child class CNEMOGas to handle all mixture/thermodynamics/transport/chemistry related computations - follwing the generic SU2 structure.

## Related Work

## PR Checklist

- [not yet] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.


Common/src/CConfig.cpp
```diff
@@ -1521,6 +1542,18 @@ void CConfig::SetConfig_Options() {
...
+  addStringDoubleListOption("MARKER_ISOTHERMAL_NONCATALYTIC", nMarker_IsothermalNonCatalytic, Marker_IsothermalNonCatalytic, Isothermal_Temperature);
+  /* DESCRIPTION: Isothermal wall boundary marker(s)
+   Format: ( isothermal marker, wall temperature (static), ... ) */
+  addStringDoubleListOption("MARKER_ISOTHERMAL_CATALYTIC", nMarker_IsothermalCatalytic, Marker_IsothermalCatalytic, Isothermal_Temperature);
+  /* DESCRIPTION: Specified heat flux wall boundary marker(s)
+   Format: ( Heat flux marker, wall heat flux (static), ... ) */
+  addStringDoubleListOption("MARKER_HEATFLUX_NONCATALYTIC", nMarker_HeatFluxNonCatalytic, Marker_HeatFluxNonCatalytic, Heat_Flux);
+  /* DESCRIPTION: Specified heat flux wall boundary marker(s)
+   Format: ( Heat flux marker, wall heat flux (static), ... ) */
+  addStringDoubleListOption("MARKER_HEATFLUX_CATALYTIC", nMarker_HeatFluxCatalytic, Marker_HeatFluxCatalytic, Heat_Flux);
```
pcarruscag, line 1556: In the interest of simplifying the setup process for users, would it make sense to just use the normal MARKER_HEATFLUX / ISOTHERMAL, and the catalytic v non-catalytic property separately? i.e.:
```
MARKER_HEATFLUX = (left,0.0, right,0.0)
MARKER_CATALYTIC = (left)
MARKER_NONCATALYTIC = (right)
```

Common/src/CConfig.cpp
```diff
@@ -4314,6 +4364,702 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+      /*( Kind_Solver == NEMO_RANS              )) */{
+
+    bool init_err;
+    unsigned short maxEl = 0;
+    unsigned short iSpecies, jSpecies, iEl;
+    su2double mf;
+
+    switch (Kind_GasModel) {
+    case ONESPECIES:
+      /*--- Define parameters of the gas model ---*/
```
pcarruscag, line 4379: I would move this chunk of code to a separate method of config

Common/src/CConfig.cpp
```diff
@@ -4314,6 +4364,702 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
...
+      Omega00 = new su2double**[nSpecies];
+      Omega11 = new su2double**[nSpecies];
+      for (iSpecies = 0; iSpecies < nSpecies; iSpecies++) {
+        Omega00[iSpecies] = new su2double*[nSpecies];
+        Omega11[iSpecies] = new su2double*[nSpecies];
+        for (jSpecies = 0; jSpecies < nSpecies; jSpecies++) {
+          Omega00[iSpecies][jSpecies] = new su2double[4];
+          Omega11[iSpecies][jSpecies] = new su2double[4];
+        }
+      }
```
pcarruscag, line 4479: These variables should definitely use a "3D" storage type, this triple nesting of pointers is not good for performance and clutters the code a lot.

Common/src/CConfig.cpp
```diff
@@ -6306,6 +7154,61 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
       }
     }
 
+      if ((Kind_Solver == NEMO_EULER) || (Kind_Solver == NEMO_NAVIER_STOKES) //|| (Kind_Solver == NEMO_RANS) ||
+         /*(Kind_Solver == DISC_ADJ_NEMO_EULER) || (Kind_Solver == DISC_ADJ_NEMO_NAVIER_STOKES) || (Kind_Solver == DISC_ADJ_NEMO_RANS)*/) {
+      switch (Kind_TimeIntScheme_NEMO) {
+        case RUNGE_KUTTA_EXPLICIT:
+          cout << "Runge-Kutta explicit method for the flow equations." << endl;
```
pcarruscag, line 7161: This duplication could be avoided by re-using the flow options I think.

Common/src/CConfig.cpp
```diff
@@ -7200,6 +8139,10 @@ bool CConfig::GetSolid_Wall(unsigned short iMarker) const {
 
   return (Marker_All_KindBC[iMarker] == HEAT_FLUX  ||
           Marker_All_KindBC[iMarker] == ISOTHERMAL ||
+          Marker_All_KindBC[iMarker] == HEAT_FLUX_NONCATALYTIC  ||
+          Marker_All_KindBC[iMarker] == ISOTHERMAL_NONCATALYTIC ||
+          Marker_All_KindBC[iMarker] == HEAT_FLUX_CATALYTIC  ||
+          Marker_All_KindBC[iMarker] == ISOTHERMAL_CATALYTIC ||
```
pcarruscag, line 8145: If you could re-use the normal HEAT_FLUX and ISOTHERMAL enums these changes could be avoided, and this type of change is the main source of bugs in SU2, invariably some area is forgotten or someone new comes along and does not know the extra markers exist... So please please consider re-using everything you can, please :)

Common/src/CConfig.cpp
```diff
@@ -7604,6 +8557,8 @@ CConfig::~CConfig(void) {
     delete[] Marker_Supersonic_Outlet;
               delete[] Marker_Outlet;
           delete[] Marker_Isothermal;
+  delete[] Marker_IsothermalCatalytic;
+  delete[] Marker_IsothermalNonCatalytic;
```
pcarruscag, line 8561: I'm not seeing to *** type containers being deleted here, are they deleted somewhere else?

SU2_CFD/include/limiters/computeLimiters_impl.hpp
```diff
@@ -75,7 +75,7 @@ void computeLimiters_impl(CSolver* solver,
                           FieldType& limiter)
 {
   constexpr size_t MAXNDIM = 3;
-  constexpr size_t MAXNVAR = 8;
+  constexpr size_t MAXNVAR = 30;
```
pcarruscag, line 78: Make it a power of 2 for my OCD please :) 32

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -893,6 +952,22 @@ class CNumerics {
...
+   * \param[in] val_dPdU - Vector of derivatives of pressure w.r.t. conserved vars.
+   * \param[in] val_normal - Normal vector, the norm of the vector is the area of the face.
+   * \param[in] l - Tangential vector to face.
+   * \param[in] m - Tangential vector to face (mutually orthogonal to val_normal & l).
+   * \param[out] val_invp_tensor - Pointer to inverse of the P matrix.
+   */
+  void GetPMatrix(su2double *U, su2double *V, su2double *val_dPdU,
+                  su2double *val_normal, su2double *l, su2double *m,
+                  su2double **val_p_tensor) ;
+
```
pcarruscag, line 970: const specifier on all inputs and on the function itself too.

SU2_CFD/include/numerics/NEMO/NEMO_diffusion.hpp
```diff
@@ -0,0 +1,258 @@
...
+  **Mean_GradPrimVar,						/*!< \brief Mean value of the gradient. */
+  *Mean_Diffusion_Coeff, /*!< \brief Mean value of the species diffusion coefficient. */
+  Mean_Laminar_Viscosity, /*!< \brief Mean value of the viscosity. */
+  Mean_Thermal_Conductivity, /*!< \brief Mean value of the thermal conductivity. */
+  Mean_Thermal_Conductivity_ve, /*!< \brief Mean value of the vib-el. thermal conductivity. */
+
+  *ProjFlux,	/*!< \brief Projection of the viscous fluxes. */
+  dist_ij;						/*!< \brief Length of the edge and face. */
+  bool implicit; /*!< \brief Implicit calculus. */
+  CNEMOEulerVariable *variable;
```
pcarruscag, line 62: This is unusual, can you add some explanation?

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -0,0 +1,912 @@
...
+  *CoPz_Inv,         /*!< \brief z Moment coefficient (inviscid contribution) for each boundary. */
+  *CFx_Inv,          /*!< \brief x Force coefficient (inviscid contribution) for each boundary. */
+  *CFy_Inv,          /*!< \brief y Force coefficient (inviscid contribution) for each boundary. */
+  *CFz_Inv,          /*!< \brief z Force coefficient (inviscid contribution) for each boundary. */
+  *Surface_CL_Inv,   /*!< \brief Lift coefficient (inviscid contribution) for each monitoring surface. */
+  *Surface_CD_Inv,   /*!< \brief Drag coefficient (inviscid contribution) for each monitoring surface. */
+  *Surface_CSF_Inv,  /*!< \brief Side-force coefficient (inviscid contribution) for each monitoring surface. */
+  *Surface_CEff_Inv, /*!< \brief Side-force coefficient (inviscid contribution) for each monitoring surface. */
+  *Surface_CFx_Inv,  /*!< \brief x Force coefficient (inviscid contribution) for each monitoring surface. */
+  *Surface_CFy_Inv,  /*!< \brief y Force coefficient (inviscid contribution) for each monitoring surface. */
```
pcarruscag, line 83: Would it make sense to inherit from CEulerSolver to re-use these things and possible their computation? (Some mad lad (me) spent more than a couple hours cleaning this code in CEulerSolver)

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -0,0 +1,912 @@
...
+  euler_implicit,         /*!< \brief True if euler implicit scheme used. */
+  least_squares;          /*!< \brief True if computing gradients by least squares. */
+
+  su2double *Primitive,   /*!< \brief Auxiliary nPrimVar vector. */
+  *Primitive_i,           /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point i. */
+  *Primitive_j;           /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point j. */
+
+  su2double *Secondary,   /*!< \brief Auxiliary nPrimVar vector. */
+  *Secondary_i,           /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point i. */
+  *Secondary_j;           /*!< \brief Auxiliary nPrimVar vector for storing the primitive at point j. */
```
pcarruscag, line 274: Please avoid using this kind of auxiliary variable, have a look at what is done in CEulerSolver currently.

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -0,0 +1,912 @@
...
+
+  /*--- Turbomachinery Solver Variables ---*/
+  su2double *** AverageFlux,
+            ***SpanTotalFlux,
+            ***AverageVelocity,
+            ***AverageTurboVelocity,
+            ***OldAverageTurboVelocity,
+            ***ExtAverageTurboVelocity,
+             **AveragePressure,
+             **OldAveragePressure,
```
pcarruscag, line 302: Is NEMO compatible with turbomachinery?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1951,6 +1951,10 @@ class CVariable {
    */
   inline virtual CVectorOfMatrix& GetGradient_Reconstruction(void) { return Gradient; }
 
+  //virtual inline void uselesstestfunction(void) {  }
+
+
+
```
pcarruscag, line 1957: I think the name says what needs to be done here :)

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -56,19 +56,17 @@ void CIntegration::Space_Integration(CGeometry *geometry,
...
       solver_container[MainSolver]->Convective_Residual(geometry, solver_container, numerics[CONV_TERM], config, iMesh, iRKStep);
       break;
   }
 
   /*--- Compute viscous residuals ---*/
-
   solver_container[MainSolver]->Viscous_Residual(geometry, solver_container, numerics, config, iMesh, iRKStep);
 
   /*--- Compute source term residuals ---*/
-
```
pcarruscag, line 71: I guess these changes are not really needed.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -192,18 +196,23 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
       Space_Integration(geometry_fine, solver_container_fine, numerics_fine, config, iMesh, iRKStep, RunTime_EqSystem);
 
+  
       /*--- Time integration, update solution using the old solution plus the solution increment ---*/
 
       Time_Integration(geometry_fine, solver_container_fine, config, iRKStep, RunTime_EqSystem);
 
+      
+
       /*--- Send-Receive boundary conditions, and postprocessing ---*/
```
pcarruscag, line 211: Same here... can you please remove all these blank lines...

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -941,6 +941,328 @@ bool CFluidIteration::MonitorFixed_CL(COutput *output, CGeometry *geometry, CSol
...
+
+  /*--- Synchronization point before a single solver iteration. Compute the
+   wall clock time required. ---*/
+
+#ifndef HAVE_MPI
+  StartTime = su2double(clock())/su2double(CLOCKS_PER_SEC);
+#else
+  StartTime = MPI_Wtime();
+#endif
+
```
pcarruscag, line 1200: This is not needed anymore, always use SU2_MPI::Wtime()

SU2_CFD/src/iteration_structure.cpp
```diff
@@ -941,6 +941,328 @@ bool CFluidIteration::MonitorFixed_CL(COutput *output, CGeometry *geometry, CSol
   return fixed_cl_convergence;
 }
 
+CNEMOIteration::CNEMOIteration(CConfig *config) : CIteration(config) { }
+
+CNEMOIteration::~CNEMOIteration(void) { }
+
+void CNEMOIteration::Preprocess(COutput *output,
```
pcarruscag, line 948: Is a NEMO iteration really necessary? It looks suspiciously similar to the fluid iteration.

SU2_CFD/src/numerics/NEMO/NEMO_diffusion.cpp
```diff
@@ -0,0 +1,1192 @@
...
+  }
+
+  for (iVar = 0; iVar < nVar; iVar++) {
+    for (iDim = 0; iDim < nDim; iDim++) {
+      Proj_Flux_Tensor[iVar] += Flux_Tensor[iVar][iDim]*val_normal[iDim];
+    }
+  }
+}
+
+void CAvgGradCorrected_NEMO::GetViscousProjJacs(su2double *val_Mean_PrimVar,
```
pcarruscag, line 782: This is a verbatim copy of the un corrected... please don't copy code like this

SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
```diff
@@ -0,0 +1,668 @@
+/*!
+ * \file NEMO_slau.hpp
```
pcarruscag, line 2: Wrong file name?

SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
```diff
@@ -0,0 +1,668 @@
...
+
+  
+
+
+
+ // std::ofstream outfile;
+//
+ // outfile.open("prints.txt", std::ios_base::app); // append instead of overwrite
+//
+ // outfile << "val_residual[nEv]=" << val_residual[nEv] << endl;
```
pcarruscag, line 551: Don't forget to remove debug statements.

SU2_CFD/src/numerics/NEMO/convection/ausm.cpp
```diff
@@ -0,0 +1,425 @@
...
+  RHOS_INDEX    = variable->GetRhosIndex()    ;
+  RHO_INDEX     = variable->GetRhoIndex()     ;
+  P_INDEX       = variable->GetPIndex()       ;
+  T_INDEX       = variable->GetTIndex()       ;
+  TVE_INDEX     = variable->GetTveIndex()     ;
+  VEL_INDEX     = variable->GetVelIndex()     ;
+  H_INDEX       = variable->GetHIndex()       ;
+  A_INDEX       = variable->GetAIndex()       ;
+  RHOCVTR_INDEX = variable->GetRhoCvtrIndex() ;
+  RHOCVVE_INDEX = variable->GetRhoCvveIndex() ;
```
pcarruscag, line 73: If this is the only function of having a NEMOEulerVariable here then either:
Make a static function to return all these indices so that you don't need to instantiate the variable.
Alternatively, and preferably, move the index logic out of the variable and re-used it in both the variable and the numerics, you can for example have a struct with only const members, which would be safe and much less verbose.

SU2_CFD/src/numerics/NEMO/convection/ausmpwplus.cpp
```diff
@@ -0,0 +1,540 @@
...
+  for (iDim = 0; iDim < nDim; iDim++) {
+    ProjVel_i += u_i[iDim]*UnitNormal[iDim];
+    ProjVel_j += u_j[iDim]*UnitNormal[iDim];
+  }
+  sqVi = 0.0;   sqVj = 0.0;
+  for (iDim = 0; iDim < nDim; iDim++) {
+    sqVi += (u_i[iDim]-ProjVel_i*UnitNormal[iDim])
+        * (u_i[iDim]-ProjVel_i*UnitNormal[iDim]);
+    sqVj += (u_j[iDim]-ProjVel_j*UnitNormal[iDim])
+        * (u_j[iDim]-ProjVel_j*UnitNormal[iDim]);
```
pcarruscag, line 217: Compilers will optimize pow(x,2) to x*x it is not necessary to duplicate things and risk missing an index.

SU2_CFD/src/numerics/NEMO/convection/ausmplusup2.cpp
```diff
@@ -0,0 +1,576 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "../../../../include/numerics/NEMO/convection/ausmplusup2.hpp"
+
+CUpwAUSMPLUSUP2_NEMO::CUpwAUSMPLUSUP2_NEMO(unsigned short val_nDim, unsigned short val_nVar,
+                                           unsigned short val_nPrimVar,
+                                           unsigned short val_nPrimVarGrad,
+                                           CConfig *config): CNumerics (val_nDim, val_nVar, config){
```
pcarruscag, line 34: Is there nothing in common between AUSM and SLAU schemes for NEMO? even for the species part? :S I managed to re-use a ton for the ideal gas versions...

SU2_CFD/src/numerics/NEMO/convection/lax.cpp
```diff
@@ -0,0 +1,193 @@
...
+}
+
+CCentLax_NEMO::~CCentLax_NEMO(void) {
+  delete [] Diff_U;
+  delete [] MeanU;
+  delete [] MeanV;
+  delete [] MeanEve;
+  delete [] MeandPdU;
+  delete [] ProjFlux;
+}
```
pcarruscag, line 83: The variable is being leaked

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3002,6 +3002,7 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
                                 (InnerIter <= config->GetLimiterIter());
   const bool van_albada       = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE);
 
+
```
pcarruscag, line 3005: delete these new empty lines please

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -0,0 +1,5111 @@
+/*!
+ * \file CEulerSolver.hpp
+ * \brief Headers of the CEulerSolver class
+ * \author F. Palacios, T. Economon
+ * \version 7.0.3 "Blackbird"
```
pcarruscag, line 5: File version

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -0,0 +1,5111 @@
...
+    Surface_CMy[iMarker_Monitoring]  = Surface_CMy_Inv[iMarker_Monitoring];
+    Surface_CMz[iMarker_Monitoring]  = Surface_CMz_Inv[iMarker_Monitoring];
+  }
+}
+
+void CNEMOEulerSolver::Momentum_Forces(CGeometry *geometry, CConfig *config) {
+
+  unsigned long iVertex, iPoint;
+  unsigned short iDim, iMarker, Boundary, Monitoring, iMarker_Monitoring;
+  su2double *Normal = nullptr, MomentDist[3] = {0.0,0.0,0.0}, *Coord, Area,
```
pcarruscag, line 2027: We must really absolutely do something about this duplication, there is quite a bit of code here that is exactly the same as in CEulerSolver. This is an "inclusive we", I can help, and it will be a good opportunity to avoid some duplication in the incompressible solvers too.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2804,10 +2808,12 @@ void CSolver::SetSolution_Gradient_LS(CGeometry *geometry, CConfig *config, bool
   const auto& solution = base_nodes->GetSolution();
   auto& rmatrix = base_nodes->GetRmatrix();
   auto& gradient = reconstruction? base_nodes->GetGradient_Reconstruction() : base_nodes->GetGradient();
+ 
   PERIODIC_QUANTITIES kindPeriodicComm = weighted? PERIODIC_SOL_LS : PERIODIC_SOL_ULS;
 
   computeGradientsLeastSquares(this, SOLUTION_GRADIENT, kindPeriodicComm, *geometry, *config,
                                weighted, solution, 0, nVar, gradient, rmatrix);
+
```
pcarruscag, line 2816: Remove the empty new lines (I understand personal preferences but I have mine too, and there is not point creating possible merge conflicts over empty lines).

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1913,6 +1913,10 @@ void CSolver::InitiateComms(CGeometry *geometry,
       COUNT_PER_POINT  = nVar*nDim*2;
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
+    case PRIMITIVE:
+      COUNT_PER_POINT  = nPrimVar;
+      MPI_TYPE         = COMM_TYPE_DOUBLE;
+      break;  
     case PRIMITIVE_GRADIENT:
```
pcarruscag, line 1920: Do you really need the new MPI comm? Usually we compute the primitives for all points, including the halos, and so there is not need to communicate them, they will automagically be the same in different MPI processes.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -0,0 +1,1466 @@
...
+  T_INDEX       = nSpecies;
+  TVE_INDEX     = nSpecies+1;
+  VEL_INDEX     = nSpecies+2;
+  P_INDEX       = nSpecies+nDim+2;
+  RHO_INDEX     = nSpecies+nDim+3;
+  H_INDEX       = nSpecies+nDim+4;
+  A_INDEX       = nSpecies+nDim+5;
+  RHOCVTR_INDEX = nSpecies+nDim+6;
+  RHOCVVE_INDEX = nSpecies+nDim+7;
+
```
pcarruscag, line 91: Def. please consider encapsulating this logic somewhere, it is repeated in the next constructor.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -0,0 +1,1466 @@
...
+    GradV[RHOCVTR_INDEX][iDim] = 0.0;
+    for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
+      GradV[RHOCVTR_INDEX][iDim] += GradU[iSpecies][iDim]*(3.0+xi[iSpecies])/2.0 *Ru/Ms[iSpecies];
+
+    /*--- Temperature ---*/
+    // Calculate the gradient of rho*u^2
+    for (jDim = 0; jDim < nDim; jDim++) {
+      Grhou2[iDim] += 2.0/rho*(rhou[jDim]*GradU[nSpecies+jDim][iDim]) -
+          GradV[RHO_INDEX][iDim]/(rho*rho)*(GradU[nSpecies+jDim][iDim]*
+          GradU[nSpecies+jDim][iDim]);
```
pcarruscag, line 1358: Same comment about using pow(x,2), someone like me, who does not know better, needs to go index by index to see that is is actually the thing multiplied by itself, whereas with pow...

Common/include/CConfig.hpp
```diff
@@ -4441,6 +4530,14 @@ class CConfig {
    */
   unsigned short GetKind_Upwind_Flow(void) const { return Kind_Upwind_Flow; }
 
+  /*!
+   * \brief Get the kind of upwind convective numerical scheme for the NEMO equations.
+   * \note This value is obtained from the config file, and it is constant
+   *       during the computation.
+   * \return Kind of upwind convective numerical scheme for the NEMO equations.
+   */
+  unsigned short GetKind_Upwind_NEMO(void) const { return Kind_Upwind_NEMO; }
```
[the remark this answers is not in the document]
CatarinaGarbacz: It is yes. I did it this way cause I see that approach being used for the adjoint and turbulent solver. Or are these different situations?

SU2_CFD/include/variables/CVariable.hpp
CatarinaGarbacz: oops. :)

Common/include/CConfig.hpp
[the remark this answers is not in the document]
pcarruscag: They are different because those solvers need to co-exist with the flow solver, the NEMO solver is a flow solver.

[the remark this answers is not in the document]
CatarinaGarbacz: got it

Common/include/CConfig.hpp
```diff
@@ -5095,144 +5204,182 @@ class CConfig {
...
    * \return: The number of chemical reactions, read from input file
    */
-  unsigned short GetnReactions(void);
+  unsigned short GetnReactions(void) const { return nReactions; }
+
+  /*!
+   * \brief Get the array that maps chemical consituents to each chemical reaction.
+   * \return Memory location of the triple pointer to the 3-D reaction map array.
+   */
+  int ***GetReaction_Map(void) const { return Reactions; }
```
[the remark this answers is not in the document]
CatarinaGarbacz: from what I understand I should use instead CVectorOfMatrix, do you confirm ?

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -109,6 +109,61 @@ CNumerics::CNumerics(unsigned short val_nDim, unsigned short val_nVar,
 
   Dissipation_ij = 1.0;
 
+  if ((config->GetKind_Solver() == NEMO_EULER)                 ||
+      (config->GetKind_Solver() == NEMO_NAVIER_STOKES) ) { //         ||
+      //(config->GetKind_Solver() == DISC_ADJ_NEMO_EULER)        ||
+      //(config->GetKind_Solver() == DISC_ADJ_NEMO_NAVIER_STOKES) ) {
```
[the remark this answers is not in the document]
CatarinaGarbacz: not sure if I understand. could you refer to an example in the code that I can follow?

Common/include/CConfig.hpp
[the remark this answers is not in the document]
pcarruscag: Yes or something equivalent, that container is a bit "special" (it has an interface because of the way we use gradients).

SU2_CFD/src/numerics/CNumerics.cpp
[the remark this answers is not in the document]
pcarruscag: Yes for example `AddRadiation`, `GetFluidProblem`, `GetFEMSolver`, `GetStructuralProblem`, `GetSolid_Wall` and a few others in CConfig that aggregate this type of conditional.

Common/include/CConfig.hpp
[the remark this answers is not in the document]
CatarinaGarbacz: what could be something equivalent ? a pointer to MatrixType for example ? sorry, I am not yet very comfortable to memory related programming just want to make sure I understand and do things correctly, in the future as well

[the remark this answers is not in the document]
pcarruscag: No problem, thanks for being open to the comments, something like this:
```c++
/* Decorate a vector type (Storage) with 3 dimensions */
template<class Storage>
struct C3DContainerDecorator {
  using Scalar = Storage::Scalar;
  using Index = Storage::Index;
  Storage storage;
  Index M, N;

  C3DContainerDecorator() = default;

  C3DContainerDecorator(Index length, Index rows, Index cols, Scalar value = 0) {
    resize(length, rows, cols, value);
  }

  void resize(Index length, Index rows, Index cols, Scalar value = 0) {
    M = rows;
    N = cols;
    storage.resize(length*rows*cols) = value;
  }

  Scalar& operator() (Index i, Index j, Index k) { return storage(i*M*N + j*N + k); }
  const Scalar& operator() (Index i, Index j, Index k) const { return storage(i*M*N + j*N + k); }
};

/* Define an alias for a 3D int matrix, we use su2vector to store the integers contiguously
 * and the container decorator to create the access semantics we want. */
using C3DIntMatrix = C3DContainerDecorator<su2vector<int> >;
```
You can put that in C2DContainer.hpp close to CVectorOfMatrix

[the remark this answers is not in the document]
pcarruscag: The key thing is trying to do as little manual memory management as possible, we should try to encapsulate everything in types that do the work for us.

Common/include/CConfig.hpp
```diff
@@ -5095,144 +5204,182 @@ class CConfig {
...
+   * \return The heat flux.
+   */
+  su2double *GetWall_Catalycity(void) const { return Wall_Catalycity; }
 
   /*!
    * \brief Provides the number of rotational modes of energy storage
    * \return: Vector of rotational mode count
    */
-  su2double* GetRotationModes(void);
+  su2double* GetRotationModes(void) const { return RotationModes; }
```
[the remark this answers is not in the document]
CatarinaGarbacz: ok thanks. can yuo explain to me why functions like GetVelocityFreestream do not have any const keyword ? since this in an input as well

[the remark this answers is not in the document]
CatarinaGarbacz: su2double* GetVelocity_FreeStream(void) { return Velocity_FreeStream; }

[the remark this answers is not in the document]
pcarruscag: Well SU2 used to have almost no const keyword... But bit by bit we are correcting that.

WallyMaier:
@CatarinaGarbacz thanks for bringing this up with develop.   @pcarruscag thanks for helping the less skilled programmers (i.e. me) walking through these comments!

Common/include/CConfig.hpp
```diff
@@ -1141,6 +1159,37 @@ class CConfig {
   unsigned short Kind_Inlet_InterpolationType;    /*!brief type of spanwise interpolation data to use for the inlet face. */
   bool PrintInlet_InterpolatedData;               /*!brief option for printing the interpolated data file. */
 
+  /* other NEMO configure options*/
```
WallyMaier, line 1162: Just a heads up regarding these NEMO configure options will exist within the a future fluid model addition.

Common/include/CConfig.hpp
```diff
@@ -1618,7 +1667,7 @@ class CConfig {
    * \brief Get the p-norm for heat-flux objective functions (adjoint problem).
    * \return Value of the heat flux p-norm
    */
-  su2double GetPnormHeat(void);
+  su2double GetPnormHeat(void) const { return pnorm_heat; }
```
WallyMaier, line 1670: I dont believe this is used anymore.

Common/include/CConfig.hpp
```diff
@@ -5281,7 +5428,12 @@ class CConfig {
   /*!
    * \brief Indicates whether electron gas is present in the gas mixture.
    */
-  bool GetIonization(void);
+  bool GetIonization(void) const { return ionization; }
+
+  /*!
+   * \brief Indicates whether electron gas is present in the gas mixture.
```
WallyMaier, line 5434: This doc string hasn't been updated.

Common/src/CConfig.cpp
WallyMaier: I think this will be temporary as we move these routines to the fluid_model routines.

Common/src/CConfig.cpp
```diff
@@ -6020,6 +6802,70 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
...
     }
 
+    if ( (Kind_Solver == NEMO_EULER) || (Kind_Solver == NEMO_NAVIER_STOKES)// || (Kind_Solver == NEMO_RANS) ||
+         /*(Kind_Solver == DISC_ADJ_NEMO_EULER) || (Kind_Solver == DISC_ADJ_NEMO_NAVIER_STOKES) ||
+         (Kind_Solver == DISC_ADJ_NEMO_RANS)*/) {
+
+      if (Kind_ConvNumScheme_NEMO == SPACE_CENTERED) {
+        //if (Kind_Centered_NEMO == JST) {
+        //  cout << "Jameson-Schmidt-Turkel scheme (2nd order in space) for the flow inviscid terms."<< endl;
+        //  cout << "JST viscous coefficients (2nd & 4th): " << Kappa_2nd_NEMO << ", " << Kappa_4th_NEMO <<"." << endl;
```
WallyMaier, line 6812: This JST chunk can be removed.

Common/src/CConfig.cpp
```diff
@@ -6020,6 +6802,70 @@ void CConfig::SetOutput(unsigned short val_software, unsigned short val_izone) {
...
+          cout << "Lax-Friedrich scheme (1st order in space) for the flow inviscid terms."<< endl;
+          cout << "Lax viscous coefficients (1st): " << Kappa_1st_NEMO << "." << endl;
+          cout << "First order integration." << endl;
+        }
+      }
+
+      if (Kind_ConvNumScheme_NEMO == SPACE_UPWIND) {
+        if (Kind_Upwind_NEMO == ROE)   cout << "Roe (with entropy fix = "<< EntropyFix_Coeff <<") solver for the flow inviscid terms."<< endl;
+        if (Kind_Upwind_NEMO == AUSM)  cout << "AUSM solver for the flow inviscid terms."<< endl;
+        if (Kind_Upwind_NEMO == AUSMPLUSUP2)  cout << "AUSM+ -Up2 solver for the flow inviscid terms."<< endl;
```
WallyMaier, line 6825: Did AUSMPWplus get removed?

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -93,7 +96,7 @@ void computeGradientsLeastSquares(CSolver* solver,
     for (size_t iDim = 0; iDim < nDim; ++iDim)
       for (size_t jDim = 0; jDim < nDim; ++jDim)
         Rmatrix(iPoint, iDim, jDim) = 0.0;
-
+
```
WallyMaier, line 99: extra spaces here

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -34,6 +34,7 @@
 #include <cstdlib>
 
 #include "../../../Common/include/CConfig.hpp"
+#include "../variables/CNEMOEulerVariable.hpp"
```
WallyMaier, line 37: I believe this include wont be necessary with the move to fluid model.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -1316,6 +1444,40 @@ class CNumerics {
    */
   static void tql2(su2double **V, su2double *d, su2double *e, unsigned short n);
 
+  /*!
+   * \brief Set variable indexes for NEMO solver.
+   */
+
+ // inline void SetRhosIndex(unsigned short val_Index) { RHOS_INDEX = val_Index; }
```
WallyMaier, line 1451: why did these get commented?

SU2_CFD/include/numerics/NEMO/NEMO_diffusion.hpp
WallyMaier: I think this comes into an issue where the numerics portion of the code currently require functions implemented in the variable class.  Work has been to fix this in a move to mimic the fluid model class.

SU2_CFD/include/numerics/NEMO/convection/ausmplusup2.hpp
```diff
@@ -0,0 +1,92 @@
+/*!
+ * \file NEMO_slau.hpp
```
WallyMaier, line 2: I think there are some inconsistencies here with the comment block.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4635,4 +4635,6 @@ class CSolver {
   void SetVerificationSolution(unsigned short nDim,
                                unsigned short nVar,
                                CConfig        *config);
+
```
WallyMaier, line 4638: extra lines here.

TestCases/euler/wedge/SU2_CFD
```diff
@@ -0,0 +1 @@
+../../../bin/SU2_CFD
```
WallyMaier, line 1: This seems to be wrong

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -595,3 +595,33 @@ struct C3DDummyMiddleView
...
+  }
+
+  Scalar& operator() (Index i, Index j, Index k) { return storage(i*M*N + j*N + k); }
+  const Scalar& operator() (Index i, Index j, Index k) const { return storage(i*M*N + j*N + k); }
+};
+
+/* Define an alias for a 3D int matrix, we use su2vector to store the integers contiguously
+ * and the container decorator to create the access semantics we want. */
+using C3DIntMatrix = C3DContainerDecorator<su2vector<int> >;
+using C3DDoubleMatrix = C3DContainerDecorator<su2vector<double> >;
```
pcarruscag, line 627: Always `su2double` unless you absolutely need a passive type, in those cases use `passivedouble` which is defined centrally.

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -595,3 +595,33 @@ struct C3DDummyMiddleView
     return data(i,k);
   }
 };
+
+/* Decorate a vector type (Storage) with 3 dimensions */
+template<class Storage>
```
pcarruscag, line 600: These comments will need to be put in doxygen format

Common/src/CConfig.cpp
pcarruscag: More reason to make it as independent as possible.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
CatarinaGarbacz: yes this is something Tim has already talked to me in slack, I am aware of the excess of repetition and it's already in my work plan to deal with it :)

SU2_CFD/src/solvers/CSolver.cpp
CatarinaGarbacz: that's true, will be careful in the future

SU2_CFD/src/solvers/CSolver.cpp
CatarinaGarbacz: no, this should have been deleted before. well noted

Common/include/CConfig.hpp
CatarinaGarbacz: not sure if I understand what you mean

Common/src/CConfig.cpp
CatarinaGarbacz: no, this was a typo. good catch

SU2_CFD/include/numerics/CNumerics.hpp
CatarinaGarbacz: with the move to fluidmodel many unnecessary/redundant things will be removed yes, but for now it's necessary for things to work

SU2_CFD/include/numerics/CNumerics.hpp
CatarinaGarbacz: because I put them only in the constructor of the NEMO numerical methods. while they were in CNumerics.hpp that meant that these functions would be executed for any solver, but they only serve a purpose for the NEMO solver

Common/include/toolboxes/C2DContainer.hpp
CatarinaGarbacz: do you i should have C3DContainerDecorator<su2vector<**su2**double> instead?

pcarruscag: yup

Common/src/CConfig.cpp
CatarinaGarbacz: @Pedro, this is one of the steps we currently doing by moving to FluidModel, we are all agreeing here :)

Common/include/CConfig.hpp
CatarinaGarbacz: @WallyMaier pls clarify

Common/include/CConfig.hpp
CatarinaGarbacz: it seems to be used actually in CNEMONSSolver::Friction_Forces

Common/include/CConfig.hpp
WallyMaier: Sorry.  The brief for GetFrozen is the same as GetIonization

SU2_CFD/include/numerics/NEMO/NEMO_diffusion.hpp
```diff
@@ -0,0 +1,257 @@
...
+   * \brief Compute the viscous flow residual using an average of gradients.
+   * \param[out] val_residual - Pointer to the total residual.
+   * \param[out] val_Jacobian_i - Jacobian of the numerical method at node i (implicit computation).
+   * \param[out] val_Jacobian_j - Jacobian of the numerical method at node j (implicit computation).
+   * \param[in] config - Definition of the particular problem.
+   */
+  void ComputeResidual(su2double *val_residual,
+                       su2double **val_Jacobian_i,
+                       su2double **val_Jacobian_j,
+                       CConfig *config);
```
pcarruscag, line 143: You should add `override` / `final` specifiers to the new numerics classes and their methods. See for example `roe.hpp` for how they are used.

SU2_CFD/include/numerics/NEMO/convection/ausm.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+public:
+
+  /*!
+   * \brief Constructor of the class.
+   * \param[in] val_nDim - Number of dimensions of the problem.
+   * \param[in] val_nVar - Number of variables of the problem.
+   * \param[in] config - Definition of the particular problem.
+   */
+  CUpwAUSM_NEMO(unsigned short val_nDim, unsigned short val_nVar, unsigned short val_nPrimVar, unsigned short val_nPrimVarGrad, CConfig *config);
```
pcarruscag, line 62: Try to make the config const (`const CConfig*`) whenever possible.

SU2_CFD/include/numerics/flow/convection/roe.hpp
```diff
@@ -50,6 +50,9 @@ class CUpwRoeBase_Flow : public CNumerics {
   su2double** Jacobian_i = nullptr; /*!< \brief The Jacobian w.r.t. point i after computation. */
   su2double** Jacobian_j = nullptr; /*!< \brief The Jacobian w.r.t. point j after computation. */
 
+  CNEMOEulerVariable *variable;
+
+
```
pcarruscag, line 55: I guess this is not needed here.

SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -0,0 +1,110 @@
...
+  temperatures.resize(nEnergyEq,0.0);
+  energies.resize(nEnergyEq,0.0);  
+  ThermalConductivities.resize(nEnergyEq,0.0);
+
+  Kind_TransCoeffModel = config->GetKind_TransCoeffModel();
+
+}
+
+CNEMOGas::~CNEMOGas(){}
+
```
pcarruscag, line 52: If you don't need to delete anything explicitly it is best to not even declare the destructor.

SU2_CFD/src/fluid/CUserDefinedTCLib.cpp
```diff
@@ -0,0 +1,1667 @@
...
+
+  Kind_GasModel = config->GetKind_GasModel();
+
+  switch (Kind_GasModel) {
+
+  case N2:
+    /*--- Check for errors in the initialization ---*/
+    init_err = false;
+    if (nSpecies != 2) {
+      cout << "CONFIG ERROR: nSpecies mismatch between gas model & gas composition" << endl;
```
pcarruscag, line 63: These very long cases could be moved to some member function "one per case", it is usually good to have small functions, especially constructors.

SU2_CFD/src/fluid/CUserDefinedTCLib.cpp
```diff
@@ -0,0 +1,1667 @@
...
+
+}
+
+vector<su2double> CUserDefinedTCLib::GetSpeciesCvTraRot(){
+ 
+  for (iSpecies = 0; iSpecies < nHeavy; iSpecies++)
+    Cvtrs[iSpecies] = (3.0/2.0 + RotationModes[iSpecies]/2.0) * Ru/MolarMass[iSpecies];
+
+  return Cvtrs;
+}
```
pcarruscag, line 595: There might be a good reason for this, but:
When returning vectors or other expensive-to-copy types, by value as done here, it is usually better to declare the variable you want to return locally, that way the copy is avoided, instead the result is "moved" into the calling function.

If the objective is to allocate space for these vectors only once in the fluid model, then they should be returned by `const&`.

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -109,6 +103,72 @@ CNumerics::CNumerics(unsigned short val_nDim, unsigned short val_nVar,
...
+  //if ((config->GetKind_Solver() == NEMO_EULER)                 ||
+  //    (config->GetKind_Solver() == NEMO_NAVIER_STOKES) ) { //         ||
+      //(config->GetKind_Solver() == DISC_ADJ_NEMO_EULER)        ||
+      //(config->GetKind_Solver() == DISC_ADJ_NEMO_NAVIER_STOKES) ) 
+  if (config->GetNEMOProblem()){
+
+    nSpecies = nVar - nDim - 2;
+
+    hs     = new su2double[nSpecies];
+    Cvtr   = new su2double[nSpecies];
```
pcarruscag, line 115: To avoid bloating the base CNumerics I recommend moving some of these member variables and the logic to allocate them to an "intermediate" parent class between CNumerics and the NEMO classes.
You can keep the members that need to be set from the outside (to avoid inefficient virtual "setter" functions) but the "working" variables can be moved to another class without hurting performance.
You can see for example what Ruben did with the Radiation classes.

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -298,6 +356,76 @@ void CNumerics::GetInviscidProjFlux(const su2double *val_density,
   }
 
 }
+void CNumerics::GetInviscidProjFlux(const su2double *val_U,
+                                    const su2double *val_V,
+                                    const su2double *val_normal,
+                                    su2double *val_Proj_Flux) {
```
pcarruscag, line 362: And if you move the variables it would also make sense to move the methods, to try and keep things organized, we have a few `InviscidProjFlux` already.

CatarinaGarbacz:
@pcarruscag thanks for the additional feedback. some of these comments were already in my plans, but can only do one thing at a time of course. In any case it's good to see we're thinking in the same way, confirms I'm heading in the right direction, so I definitely appreciate it. Others I had not thought about it at all, even more useful! Thanks!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -0,0 +1,3172 @@
...
+    }
+  }
+
+  /*--- Warning message about non-physical points ---*/
+  if (config->GetComm_Level() == COMM_FULL) {
+#ifdef HAVE_MPI
+    SU2_MPI::Reduce(&counter_local, &counter_global, 1, MPI_UNSIGNED_LONG, MPI_SUM, MASTER_NODE, MPI_COMM_WORLD);
+#else
+    counter_global = counter_local;
+#endif
```
pcarruscag, line 264: Glad to see the drop in lines of code :)
While I'm here, These type of #ifdef HAVE_MPI blocks are very rarely needed, SU2_MPI:: does the direct copy when the code is compiled without MPI.

CatarinaGarbacz:
Lots of different updates have been done to incorporate all feedback and make improvements to this PR. It seems to be in a solid place to be merged with develop. @pcarruscag @talbring

SU2_CFD/src/numerics/CNumerics.cpp
pcarruscag: Take the above as optional, but please do the following:
Delete the commented part,
Delete the variables in this block in the destructor, currently it seems they are all being leaked (consider running valgrind to find other such issues).

SU2_CFD/include/output/CNEMOCompOutput.hpp
```diff
@@ -0,0 +1,138 @@
...
+ * \version 7.0.0 "Blackbird"
+ *
+ * The current SU2 release has been coordinated by the
+ * SU2 International Developers Society <www.su2devsociety.org>
+ * with selected contributions from the open-source community.
+ *
+ * The main research teams contributing to the current release are:
+ *  - Prof. Juan J. Alonso's group at Stanford University.
+ *  - Prof. Piero Colonna's group at Delft University of Technology.
+ *  - Prof. Nicolas R. Gauger's group at Kaiserslautern University of Technology.
```
pcarruscag, line 14: File version and entire header style are outdated.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -312,7 +317,7 @@ class CFVMFlowSolverBase : public CSolver {
    * \param[in] val_marker - Surface marker where the boundary condition is applied.
    */
   inline void BC_Euler_Wall(CGeometry* geometry, CSolver** solver_container, CNumerics* conv_numerics,
-                            CNumerics* visc_numerics, CConfig* config, unsigned short val_marker) final {
+                            CNumerics* visc_numerics, CConfig* config, unsigned short val_marker) override {
     /*--- Call the equivalent symmetry plane boundary condition. ---*/
     BC_Sym_Plane(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
```
pcarruscag, line 322: Only one of BC_Euler_Wall / BC_Sym_Plane needs to be demoted from final to override.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -29,15 +29,20 @@
 #include "../../../Common/include/omp_structure.hpp"
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
 #include "CSolver.hpp"
+#include "../variables/CNEMOEulerVariable.hpp"
+
+
```
pcarruscag, line 34: move this include to CNEMOEulerSolver (before the include of FVMBase) otherwise when you change CNEMOEulerVariable all derived solvers of FVMBase are forced to recompile.

SU2_CFD/include/solvers/CNEMONSSolver.hpp
```diff
@@ -0,0 +1,213 @@
...
+                               CNumerics *visc_numerics,
+                               CConfig *config,
+                               unsigned short val_marker);
+
+ // /*!
+ //  * \brief Compute the viscous forces and all the addimensional coefficients.
+ //  * \param[in] geometry - Geometrical definition of the problem.
+ //  * \param[in] config - Definition of the particular problem.
+ //  */
+ // void Friction_Forces(const CGeometry* geometry, const CConfig* config) final;
```
pcarruscag, line 199: delete

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -0,0 +1,550 @@
...
+ * \author F. Palacios, T. Economon, W. Maier, S.R. Copeland
+ * \version 6.2.0 "Falcon"
+ *
+ * The current SU2 release has been coordinated by the
+ * SU2 International Developers Society <www.su2devsociety.org>
+ * with selected contributions from the open-source community.
+ *
+ * The main research teams contributing to the current release are:
+ *  - Prof. Juan J. Alonso's group at Stanford University.
+ *  - Prof. Piero Colonna's group at Delft University of Technology.
```
pcarruscag, line 13: file version and rest of the license notice

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1414,6 +1419,13 @@ class CVariable {
    */
   inline virtual bool SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel) { return true; }
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   * \param[in] fluidmodel - fluid model.
+   */
+  inline virtual bool SetPrimVar(unsigned long iPoint, CNEMOGas *fluidmodel) {return false;}
```
pcarruscag, line 1427: This is not a very good idea, NEMOGas is "convertible" to FluidModel, so the first overload can be called accidentally if you have a FluidModel pointer to a NEMOGas instance (which would be legal).
It would be safer to use the `SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)` signature also for the NEMO variables and then `static_cast` (or `dynamic_cast` if you want to error check) to NEMOGas

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2886,6 +3023,11 @@ void CFluidDriver::Preprocess(unsigned long Iter) {
         for (iInst = 0; iInst < nInst[iZone]; iInst++)
           solver_container[iZone][iInst][MESH_0][FLOW_SOL]->SetInitialCondition(geometry_container[iZone][INST_0], solver_container[iZone][iInst], config_container[iZone], Iter);
       }
+      if ((config_container[iZone]->GetKind_Solver() ==  NEMO_EULER) ||
+          (config_container[iZone]->GetKind_Solver() ==  NEMO_NAVIER_STOKES)) {
```
pcarruscag, line 3027: replace by NEMOProblem method

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -257,6 +261,16 @@ void CMultizoneDriver::Preprocess(unsigned long TimeIter) {
...
     }
+    if ((config_container[iZone]->GetKind_Solver() ==  NEMO_EULER) ||
+        (config_container[iZone]->GetKind_Solver() ==  NEMO_NAVIER_STOKES) /*
+        (config_container[iZone]->GetKind_Solver() ==  NEMO_RANS)*/ ) {
+        if(!fsi) solver_container[iZone][INST_0][MESH_0][FLOW_SOL]->SetInitialCondition(geometry_container[iZone][INST_0], solver_container[iZone][INST_0], config_container[iZone], TimeIter);
+    }
+    if ((config_container[iZone]->GetKind_Solver() ==  NEMO_EULER) ||
+        (config_container[iZone]->GetKind_Solver() ==  NEMO_NAVIER_STOKES) /*
+        (config_container[iZone]->GetKind_Solver() ==  NEMO_RANS)*/ ) {
+        if(!fsi) solver_container[iZone][INST_0][MESH_0][FLOW_SOL]->SetInitialCondition(geometry_container[iZone][INST_0], solver_container[iZone][INST_0], config_container[iZone], TimeIter);
```
pcarruscag, line 272: NEMOProblem method

SU2_CFD/src/integration/CIntegration.cpp
pcarruscag: This was not resolved

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -105,6 +106,10 @@ class CVariable {
   unsigned long nSecondaryVar = 0;     /*!< \brief Number of secondary variables. */
   unsigned long nSecondaryVarGrad = 0;   /*!< \brief Number of secondaries for which a gradient is computed. */
 
+  /*!< \brief Inddex definition for NEMO variables. */
+  unsigned long RHOS_INDEX, T_INDEX, TVE_INDEX, VEL_INDEX, P_INDEX, 
+  RHO_INDEX, H_INDEX, A_INDEX, RHOCVTR_INDEX, RHOCVVE_INDEX, nSpecies;
+
```
pcarruscag, line 112: Take this out of CVariable please, it does not make sense to have it here visible to other variable classes that have nothing to do with NEMO or even flow problems.

SU2_CFD/include/fluid/CNEMOGas.hpp
```diff
@@ -0,0 +1,237 @@
...
+  
+  /*!
+   * \brief Get species T-R specific heats at constant volume.
+   */
+  virtual vector<su2double>& GetSpeciesCvTraRot(){}
+  
+  /*!
+   * \brief Get species V-E specific heats at constant volume.
+   */
+  virtual vector<su2double>& GetSpeciesCvVibEle(){}
```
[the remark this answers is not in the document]
pcarruscag: Alternatively you can return a const* (which can be nullptr).

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: Nevermind what I said, it would not work, instead delete the include from this file and do a forward declaration of CNEMOEulerVariable (`class CNEMOEulerVariable;`).

Common/src/CConfig.cpp
CatarinaGarbacz: I am not sure if I understand the suggestion @pcarruscag .  HEAT_FLUX_NONCATALYTIC , ISOTHERMAL_NONCATALYTIC , HEAT_FLUX_CATALYTIC, ISOTHERMAL_CATALYTIC are bcs different than the normal HEAT_FLUX and ISOTHERMAL, therefore need different flags in the cfg file and corresponding options in option structure, to be used somewhere else.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -195,11 +199,11 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
       /*--- Space integration ---*/
 
       Space_Integration(geometry_fine, solver_container_fine, numerics_fine, config, iMesh, iRKStep, RunTime_EqSystem);
-
+  
       /*--- Time integration, update solution using the old solution plus the solution increment ---*/
 
       Time_Integration(geometry_fine, solver_container_fine, config, iRKStep, RunTime_EqSystem);
-
+
```
pcarruscag, line 206: More spaces in this file to revert

Common/src/CConfig.cpp
pcarruscag: Ok, I thought these were small variations on the normal heatflux / isothermal boundaries (atm non-catalytic just calls the normal implementation and catalytic makes use of the normal version right?), and that it might be more intuitive for an user to specify those properties separately.
By having different boundary condition names for those, you make them incompatible with new features added to the code, for example #877. 
I do not have domain knowledge about your work, I am just trying to have the multi physics nature of SU2 in mind, the same way we have abstractions for solvers, we also need to have abstractions for options, otherwise we have a combinatoric explosion every time something new is added...

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -172,6 +170,9 @@ void CIntegration::Space_Integration(CGeometry *geometry,
           solver_container[MainSolver]->BC_HeatFlux_Wall(geometry, solver_container, conv_bound_numerics, visc_bound_numerics, config, iMarker);
         }
         break;
+      case SMOLUCHOWSKI_MAXWELL:
```
pcarruscag, line 173: By the way I do not see the catalytic boundaries being called here, maybe they got lost while merging with develop?

CatarinaGarbacz:
@pcarruscag the new and old comments have been addressed. Also the many warnings the code was generating.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2066,8 +2066,10 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
       }
     }
 
-    if ((Boundary == HEAT_FLUX) || (Boundary == ISOTHERMAL) || (Boundary == CHT_WALL_INTERFACE)) {
-      /*--- Forces initialization at each Marker ---*/
+    if ((Boundary == HEAT_FLUX) ||
+       (Boundary == ISOTHERMAL) ||
+       (Boundary == CHT_WALL_INTERFACE) || 
+       (Boundary == SMOLUCHOWSKI_MAXWELL)) {       /*--- Forces initialization at each Marker ---*/
```
pcarruscag, line 2072: @fmpmorgado please don't do this, modify "CConfig::GetViscous_Wall()" to include the new marker and use it instead.
This repetition of enums throughout the code is a major source of bugs and it is impossible to maintain, see #770.

fmpmorgado: Thank you for the comment. I believe it is working now

pcarruscag: The compilation problem is solved but you will still need to add SMOLUCHOWSKI_MAXWELL to that function.

pcarruscag (approved):
LGTM, merging after tests pass

PR #1015, opened 2020-06-03
Dividing Iteration Classes

talbring:
## Proposed Changes

This divides the iteration files and adds a factory class for their creation. A rework of the iteration class is still necessary, but I don't have time to do that at the moment.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.
- [] I have added a test case that demonstrates my contribution, if necessary.
- [] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1020, opened 2020-06-08
Quasi-Newton convergence acceleration/stabilization of discrete adjoints

### LINKED ISSUE #1021: Develop does not compile with GCC 5.4
Just as FYI, the fix is in #1020.

### LINKED ISSUE #1025: Incorrect names for limiter outputs
The compressible primitive variables are T u v w P rho H, we output the limiters for T u v w P with the names:
LIMITER_DENSITY
LIMITER_MOMENTUM-X
LIMITER_MOMENTUM-Y
LIMITER_MOMENTUM-Z
LIMITER_ENERGY
Which would be ok if we were referring to the conservative variables. Also the limiter for temperature is not relevant for ideal gases as the state is derived from the enthalpy instead.

### LINKED ISSUE #1029: Wrong communication of gradients for periodic cases
FYI
Introduced in #1009
Found by @TobiKattmann 
Fixed in #1020

pcarruscag:
## Proposed Changes
So I took a common method that folks use to converge multi physics problems expressed as fixed point iterations (the IQN-ILS) and applied it to the inner iterations of the discrete adjoint drivers (single and multizone), it seems to work well... I was having some issues when the primal solver does not converge so well for optimization edge cases, and this keeps the adjoint from diverging.
I'll post some results at some point.

Other than storing  a number of solution snapshots (20 seems like a good number) the overhead is minimal (and also provided you have Lapack/MKL, compile for OpenMP, with fast-math, and AVX support).

## Related Work
Already has #1015
Resolves #1021 
Resolves #1025
Resolves #1029

## PR Checklist


Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
+/*!
+ * \file CQuasiNewtonDriver.hpp
+ * \brief Implements a method to accelerate and stabilize the convergence
+ * of fixed point iterations, the history of past iterates is used to compute
+ * a least squares approximation of the inverse of the Jacobian, which is then
+ * used to correct the natural solution of the fixed point iteration.
+ * \note Based on the IQN-ILS method, see DOI 10.1007/s11831-013-9085-5 and
+ * references therein.
```
pcarruscag, line 8: All the implementation, explanation and references are in this file.

Common/src/CConfig.cpp
```diff
@@ -1571,6 +1571,8 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION:  Offset parameter for the buffet sensor */
   addDoubleOption("BUFFET_LAMBDA", Buffet_lambda, 0.0);
 
+  /* DESCRIPTION: Number of samples for quasi-Newton methods. */
+  addUnsignedShortOption("QUASI_NEWTON_NUM_SAMPLES", nQuasiNewtonSamples, 0);
```
pcarruscag, line 1575: The feature is activated by setting a number of samples greater than 1.

SU2_CFD/src/drivers/CDiscAdjSinglezoneDriver.cpp
```diff
@@ -187,11 +200,18 @@ void CDiscAdjSinglezoneDriver::Run() {
...
     }
 
     if (StopCalc) break;
 
+    /*--- Correct the solution with the quasi-Newton approach. ---*/
+
+    if (QNDriver.size()) {
+      GetAllSolutions(ZONE_0, true, QNDriver.FPresult());
+      SetAllSolutions(ZONE_0, true, QNDriver.compute());
+    }
```
pcarruscag, line 213: And this basically how it is used.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
...
+  Index nPtDomain = 0;                  /*!< \brief Local size of the history, considered in dot products. */
+
+  void shiftHistoryLeft() {
+    for (Index i = 1; i < X.size(); ++i) {
+      /*--- Swap instead of moving to re-use the memory of the first sample.
+       * This is why X and R are not stored as contiguous blocks of mem. ---*/
+      std::swap(X[i-1], X[i]);
+      std::swap(R[i-1], R[i]);
+    }
+  }
```
pcarruscag, line 72: For the benefit of those less familiar with c++11, there are these things called move operations, this operation of shifting matrices to the left is nearly 0 cost because swapping two `su2matrices` only swaps the pointers and sizes and does not copy the actual data.
That is why in this case X and R are not stored as a single chunk of memory.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
...
+    const auto end = std::min<Index>(nPtDomain,corr.rows())*corr.cols();
+
+    mat = Scalar(0); rhs = Scalar(0);
+
+    /*--- Tiled part of the loop. ---*/
+    Index begin = 0;
+    while (end-begin >= BLOCK_SIZE) {
+      computeNormalEquations<BLOCK_SIZE>(mat, rhs, begin);
+      begin += BLOCK_SIZE;
+    }
```
pcarruscag, line 85: For those interested in learning about performance techniques, I use something here called loop tiling, or loop blocking, or strip mining.
We need to compute A^T A where A is narrow, doing the natural "row dot column" algorithm is very inefficient because it makes poor use of the CPU cache, i.e. after doing the dot product of column 0 with itself, we go back to the beginning of column 0 to dot it with column 1, but by now the beginning of column 0 was evicted from cache (getting data from main memory is orders of magnitude slower than getting it from cache).
By tiling the loop we go over all the different combinations of columns and rows without evicting anything from the cache (since we work on much smaller sizes).
For general matrix multiplication we would need to tile both rows and columns.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  template<Index StaticSize>
+  void computeNormalEquations(su2vector<Scalar>& mat,
+                              su2vector<Scalar>& vec,
+                              Index start,
+                              Index dynSize = 0) const {
+    /*--- Select either the static or dynamic size, optimizes inner loop. ---*/
+    const auto blkSize = StaticSize? StaticSize : dynSize;
```
pcarruscag, line 113: Because we tiled the loop, we have many work chunks with a static size, and one remainder chunk with dynamic size (not known during compilation).
This method is templated on the static size, so that we can "generate" two versions, for the version where the compiler knows the size of the loop it can generate much more efficient code because:
The size is a multiple of the SIMD length, and likely also of the loop unrolling count, which means only the necessary instructions are generated (better use of instruction cache), and size checks are not needed (less branching).

oleburghardt:
Great stuff, thanks for taking the first step to integrate quasi-Newton techniques to the adjoint fixed point iterations. I'm reviewing this the next couple of days.

Just one question right away so that I'm on the right track. It seems that this implementation is based on equation 130 from Degroote's paper. Right? Though you're not using any QR decomposition for solving it? What's the approach then (assumingly in `computeNormalEquations`)?

And you're not using `std::deque` like I did for storing a window of some past residuals (i.e. V^k), looking forward to see what kind of solution you came up with.

pcarruscag:
Yes equation 130, the approach to solve equation 130, is to solve equation 130 (which are the normal equations mentioned in `computeNormalEquations`).
What I used is better than deque, as explained in one of my comments it allows shifting the vectors at zero cost.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
...
+  using Scalar = Scalar_t;
+  using Index = typename su2matrix<Scalar>::Index;
+  static_assert(std::is_floating_point<Scalar>::value,"");
+
+private:
+  using MPI_Wrapper = typename SelectMPIWrapper<Scalar>::W;
+
+  enum: size_t {BLOCK_SIZE = 1024};     /*!< \brief Loop tiling parameter. */
+  std::vector<su2matrix<Scalar> > X, R; /*!< \brief Input and residual history of the FP. */
+  su2matrix<Scalar> corr;               /*!< \brief Correction to the natural FP result. */
```
oleburghardt, line 60: Just at one intermediate point it's the correction to the natural FP result, right? It first holds the current solution, then the delta of corrected solutions, then the corrected solution. That should be noted somewhere (so that it also goes well together with the `FPresult()` code comment) or reflected in the name.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -278,6 +297,14 @@ void CDiscAdjMultizoneDriver::Run() {
                                                     solver_container, numerics_container, config_container,
                                                     surface_movement, grid_movement, FFDBox, iZone, INST_0);
 
+        /*--- Use QN driver to improve the solution. ---*/
+
+        if (QNDriver[iZone].size()) {
+          GetAllSolutions(iZone, true, QNDriver[iZone].FPresult());
+          QNDriver[iZone].compute();
```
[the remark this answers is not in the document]
pcarruscag: I don't think so, it is just a reference, I gave compute a return value for syntax sugar.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
pcarruscag: Called it work now and updated the comment.

Common/include/toolboxes/CQuasiNewtonDriver.hpp
```diff
@@ -0,0 +1,253 @@
...
+ * \class A quasi-Newton fixed-point (FP) accelerator based on IQN-ILS.
+ * \note The implementation prioritizes storage, the LS problem is solved
+ * via the normal equations as that is easy to make parallel over MPI, it
+ * may however be unstable with large sample sizes (compared to QR decomp).
+ * Usage: Allocate, store the initial solution (operator (i,j), default is 0),
+ * run the FP, store its result ("FPresult"), compute new solution, use it
+ * as the new input of the FP, run the FP, etc.
+ */
+template<class Scalar_t>
+class CQuasiNewtonDriver {
```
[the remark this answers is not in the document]
pcarruscag: I'll call it `CQuasiNewtonInvLeastSquares` then if we eventually implement more methods we can think of a generic name for the parent class, fixed point corrector or something.

pcarruscag:
Thanks for the review @oleburghardt , corrections coming up.

UnitTests/Common/toolboxes/CQuasiNewtonInvLeastSquares_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+template<class P, class Q>
+void iterate(P& p, Q& q) {
+  p.iterate(q);
+  for(int i=0; i<P::N; ++i)
+    q.FPresult(i,0) = p.sol[i];
+  q.compute();
+}
+
+TEST_CASE("QN-ILS", "[Toolboxes]") {
```
pcarruscag, line 62: @clarkpede @talbring , I've joined the unit test bandwagon, let me known if the naming and so on is appropriate (I borrowed inspiration from the other tests so I guess it's alright).

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
oleburghardt: Ok

Common/include/toolboxes/CQuasiNewtonDriver.hpp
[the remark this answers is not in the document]
oleburghardt: Alright let's do so :)

PR #1022, opened 2020-06-10
Vectorized (SIMD) Numerical Schemes

pcarruscag:
## Proposed Changes
The objective is to provide a natural (i.e. readable) way to write code that extracts all the performance modern CPU have to offer.

As described in the last update of #789 this will require completely different numerics classes, also to address other inefficiencies they have. For now all centered schemes and plain Roe have been "ported" (this also adds JST with matrix dissipation).
Most of the auxiliary CNumerics functions were also ported as consequence (inviscid fluxes, Jacobians, and so on) and so implementing new schemes only gets easier.

If you want to test this, the new numerics are called for Roe on ideal gas without low Mach corrections (enabled by option USE_VECTORIZATION=YES) and also for centered schemes.
To get good performance on gcc you need the following optimization flags:
`-O2 -funroll-loops -ffast-math -march=?? -mtune=??` Where ?? should be something with AVX (e.g. `haswell`, `skylake`).
There are AVX and AVX512 specific optimizations.
Also the OpenMP simd directive is used, at least on gcc this makes a difference (because it is stubborn at vectorizing) so -Dwith-omp=true when calling meson.py.
You can expect around 30% speedup +- 10% on problems that do not need a lot of linear solver iterations, or more on machines with AVX512.

## Related Work
#789

## PR Checklist


Common/include/parallelization/vectorization.hpp
```diff
@@ -0,0 +1,212 @@
...
+template<>
+constexpr size_t simdLen<su2double>() { return SIMD_SIZE / sizeof(passivedouble); }
+
+/*!
+ * \class Array
+ * \brief A simple SIMD type relying mostly on implicit vectorization, i.e. done by
+ * the compiler, as opposed to explicit (done via intrinsics or inline assembly).
+ */
+template<class Scalar_t, size_t N = simdLen<Scalar_t>()>
+class Array {
```
pcarruscag, line 63: The vectorization will be based on a "simd type" basically a static array class for which we overload all required operators and functions.
For now this relies just on the compiler to do "the right thing" which does require a few more flags than folks will be used to, for example https://gcc.godbolt.org/z/N4FnXM.
Relying just on the compiler makes it easier to extend functionality and makes the code compatible with un-vectorizable types.
The important thing is to hide the implementation behind a generic interface, then we can think about specializing key operations on certain types using architecture-specific intrinsics for example.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,594 @@
...
+      velocityRoe(iDim) = (R*V.j.velocity(iDim) + V.i.velocity(iDim)) / (R+1);
+    }
+    auto enthalpyRoe = (R*V.j.enthalpy + V.i.enthalpy) / (R+1);
+    auto soundSpeedRoe = sqrt((gamma-1) * (enthalpyRoe - 0.5*squaredNorm(velocityRoe)));
+
+    /*--- P tensor and inverse P tensor. ---*/
+
+    auto projVelRoe = dot(velocityRoe, unitNormal);
+    auto pMat = pMatrix(gamma, densityRoe, velocityRoe, projVelRoe, soundSpeedRoe, unitNormal);
+    auto pMatInv = pMatrixInv(gamma, densityRoe, velocityRoe, projVelRoe, soundSpeedRoe, unitNormal);
```
pcarruscag, line 509: This does not compile yet, but in theory the vectorized code will look exactly the same (well it will look better because that's what hindsight tends to do).

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,594 @@
...
+
+template<size_t nDim>
+struct CCompressibleConservatives {
+  enum : size_t {nVar = nDim+2};
+  VectorDbl<nVar> all;
+  Double& density = all(0);
+  Double& rhoEnergy = all(nDim+1);
+  FORCEINLINE Double& momentum(size_t iDim) { return all(iDim+1); }
+  FORCEINLINE Double* momentum() { return &momentum(0); }
+};
```
pcarruscag, line 224: I'm also trying to address the magic indexes (iDim+1) by defining specific types for primitive and conservative variables rather that using raw arrays.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+  FORCEINLINE Double& density() { return all(nDim+2); }
+  FORCEINLINE Double& enthalpy() { return all(nDim+3); }
+  FORCEINLINE Double& velocity(size_t iDim) { return all(iDim+1); }
+  FORCEINLINE const Double& temperature() const { return all(0); }
+  FORCEINLINE const Double& pressure() const { return all(nDim+1); }
+  FORCEINLINE const Double& density() const { return all(nDim+2); }
+  FORCEINLINE const Double& enthalpy() const { return all(nDim+3); }
+  FORCEINLINE const Double& velocity(size_t iDim) const { return all(iDim+1); }
+  FORCEINLINE const Double* velocity() const { return &velocity(0); }
+};
```
pcarruscag, line 223: I'm also trying to address the magic indexes (iDim+1) by defining specific types for primitive and conservative variables rather that using raw arrays.
Initially I thought of making `temperature, pressure, etc.` references pointing into the correct position of the storage vector (`all`)  but unfortunately that disables compiler generated constructors... so they are functions.

Common/include/toolboxes/C2DContainer.hpp
```diff
@@ -483,13 +538,31 @@ class C2DContainer :
...
+  }
+
+  /*!
+   * \brief Get a SIMD gather iterator to the inner dimension of the container.
+   */
+  template<size_t nCols, class T, size_t N>
+  FORCEINLINE CInnerIterGather<simd::Array<T,N> > innerIter(simd::Array<T,N> row) const noexcept
+  {
+    return CInnerIterGather<simd::Array<T,N> >(m_data, IsRowMajor? 1 : this->rows(), IsRowMajor? row*nCols : row);
+  }
```
pcarruscag, line 558: The only places where the "vector nature" of the SIMD type needs to be handled explicitly are when retrieving data from containers (C2DContainer and CVectorOfMatrix) and when putting it back in other containers (CSysVector and CSysMatrix).
Data is retrieved via iterators (see #789 for the reason, spoiler alert, it's for performance), the method to obtain the iterator is templated for the number of columns, i.e. the number of variables, because this is known at compile time in the new numerics (which in turn are templated for nDim). This is somewhat of a micro optimization, after benchmarking I might make it more general.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+template<size_t nDim, class Coord_t>
+FORCEINLINE VectorDbl<nDim> distanceVector(Int iPoint, Int jPoint, const Coord_t& coords) {
+  VectorDbl<nDim> vector_ij;
+  auto coord_i = coords.template innerIter<nDim>(iPoint);
+  auto coord_j = coords.template innerIter<nDim>(jPoint);
+  for (size_t iDim = 0; iDim < nDim; ++iDim) {
+    vector_ij(iDim) = 0.5 * (*(coord_j++) - *(coord_i++));
+  }
+  return vector_ij;
+}
```
pcarruscag, line 139: Those iterators are then used like this, the templating for the known size does require the somewhat baffling syntax you see above...

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+  }
+}
+
+/*!
+ * \class CRoeBase
+ * \brief Base class for Roe schemes, derived classes implement
+ *        the dissipation term in a "finalizeResidual" method.
+ */
+template<class Derived, size_t NDIM>
+class CRoeBase : public CNumericsSIMD {
```
pcarruscag, line 504: For example the base class for Roe schemes is templated on its "future" derived classes.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+    MatrixDbl<nVar> jac_i, jac_j;
+    if (implicit) {
+      jac_i = inviscidProjJac(gamma, V.i.velocity(), U.i.rhoEnergy()/U.i.density(), normal, kappa);
+      jac_j = inviscidProjJac(gamma, V.j.velocity(), U.j.rhoEnergy()/U.j.density(), normal, kappa);
+    }
+
+    /*--- Finalize in derived class (static polymorphism). ---*/
+
+    Derived::finalizeResidual(flux, jac_i, jac_j, implicit, gamma, kappa, area,
+                              unitNormal, V, U, roeAvg, lambda, pMat);
```
pcarruscag, line 622: Then we can call static methods of those derived classes like this.
Being static means we need to pass all arguments explicitly, but as mentioned above we do not want to rely on member variables for "Base-Derived communication" as that would not be thread-safe.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+  }
+};
+
+/*!
+ * \class CRoeScheme
+ * \brief Classical Roe scheme.
+ */
+template<size_t nDim>
+class CRoeScheme : public CRoeBase<CRoeScheme<nDim>,nDim> {
+private:
```
pcarruscag, line 651: Then derived classes need to "inform" the base class of who they are (curious recursion).
The code in Base::ComputeResidual will actually be instantiated (i.e. compiled and optimized) for each and every derived class, with the only difference being the details in `finalizeResidual`.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,711 @@
...
+  void ComputeFlux(unsigned long iEdge,
+                   const CConfig& config,
+                   const CGeometry& geometry,
+                   const CVariable& solution_,
+                   UpdateType updateType,
+                   CSysVector<su2double>& vector,
+                   CSysMatrix<su2mixedfloat>& matrix) const final {
+
+    const bool implicit = (config.GetKind_TimeIntScheme() == EULER_IMPLICIT);
+    const auto& solution = static_cast<const CEulerVariable&>(solution_);
```
pcarruscag, line 538: Another significant change from the current numerics is that (again for thread safety) we do not "set" anything into them, we pass them the whole variable structure (the solver `nodes`) and the numerics can read (and only read) any data it needs.
Similarly to what is done in the solvers, since we know the type of variable that goes with the numeric scheme we can `static_cast` to the derived type to avoid polymorphism.

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,777 @@
...
+ * they are not being decorated.
+ * Compile-time decoration works by specifying the base class as a template parameter.
+ * Then the class being decorated should call a method of its base class to add some
+ * contribution to the flux/source and Jacobians just before writting the results to
+ * CSysVector and CSysMatrix. The mechanism can be used to chain any number of classes
+ * at compile-time, but its main purpose is to combine convective and viscous fluxes
+ * in the most (nearly) efficient way.
+ */
+template<size_t NDIM>
+class CNumericsEmptyDecorator : public CNumericsSIMD {
```
pcarruscag, line 118: here

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
```diff
@@ -0,0 +1,777 @@
...
+ * the dissipation term in a static "finalizeFlux" method.
+ * A base class implementing "updateFlux" is accepted as template parameter.
+ * Similarly to derived, that method should update the flux and Jacobians, but
+ * whereas "finalizeFlux" is passed data prepared by CRoeBase, "updateFlux"
+ * takes the same input arguments as "ComputeFlux", i.e. it must fetch the data
+ * it needs from CVariable, etc.. Derived is meant to implement small details,
+ * Base is meant to do heavy lifting like computing viscous fluxes.
+ */
+template<class Derived, class Base>
+class CRoeBase : public Base {
```
pcarruscag, line 556: and here

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -589,17 +634,59 @@ class CSysMatrix {
...
+      auto bij = &matrix[edge_ptr(iEdge[k],0)*blkSz];
+      auto bji = &matrix[edge_ptr(iEdge[k],1)*blkSz];
+
+      /*--- Update, block i was negated during transpose in the
+       * hope the assignments below become non-temporal stores. ---*/
+      for (size_t i=0; i<blkSz; ++i) {
+        bij[i] = blk_j[k][i];
+        bji[i] = blk_i[k][i];
+      }
+    }
```
pcarruscag, line 689: The "need" for that convective/viscous fusion comes from the increased complexity of writing to CSysMatrix (so we want to do that only once) and also to create the opportunity for non-temporal stores (see e.g. https://blogs.fau.de/hager/archives/2103 for why we want these).

pcarruscag:
In terms of general look and feel this is ready for folks to have a look, there are a couple kinks to iron out:
- AD preaccumulation, @talbring is it ok to register the same variable for preaccumulation multiple times?
- Boundary numerics, it may not be possible to use these high efficiency classes for boundary terms, they are not as flexible in terms of what can be used as data sources, the current numerics take any old pointer, but these assume the data always comes from CVariable. They will probably need to complement each other.

SU2_CFD/include/numerics_simd/util.hpp
```diff
@@ -0,0 +1,181 @@
...
+
+/*!
+ * \brief Gather a matrix of variables from outer index iPoint of a 3D container.
+ */
+template<size_t nRows, size_t nCols, class Container>
+FORCEINLINE MatrixDbl<nRows,nCols> gatherVariables(Int iPoint, const Container& vars) {
+  auto x = vars.template get<MatrixDbl<nRows,nCols> >(iPoint);
+  AD::SetPreaccIn(x, nRows, nCols);
+  return x;
+}
```
pcarruscag, line 141: I'm handling pre-accumulation by registering variables as they are retrieved from the containers, which I think should make the whole process much safer.

Fusing the convective and viscous fluxes reduces the memory used by the AD tape by about 20% for a 2D RANS SST problem. The recording seems to be slightly more expensive but each iteration then runs faster.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -345,9 +346,29 @@ CEulerSolver::CEulerSolver(CGeometry *geometry, CConfig *config,
...
+
+    if (uncertain || !ideal_gas || low_mach_corr) {
+      SU2_MPI::Error("Some of the requested features are not yet "
+                     "supported with vectorization.", CURRENT_FUNCTION);
+    }
+
+    edgeNumerics = CNumericsSIMD::CreateNumerics(*config, nDim, iMesh);
+
+    if (!edgeNumerics) {
+      SU2_MPI::Error("The numerical scheme in use does not "
```
CatarinaGarbacz, line 367: typo in error message, just noticed by chance

CatarinaGarbacz:
@pcarruscag , this significant contribution on vectorization seems very valuable as one more initiative to speed up (and tidy up) the code. It's great to see SU2 evolving to make use of handy programming and C++ performance strategies. 

I am not an expert in this type of implementation, but overall it seems to be implemented in a neat and advanced way, making good use of C++ tools, functions and user defined types.

CatarinaGarbacz (commented):
@pcarruscag , this significant contribution on vectorization seems very valuable as one more initiative to speed up (and tidy up) the code. It's great to see SU2 evolving to make use of handy programming and C++ performance strategies.

I am not an expert in this type of implementation, but overall it seems to be implemented in a neat and advanced way, making good use of C++ tools, functions and user defined types.

CatarinaGarbacz (approved):

SU2_CFD/include/numerics_simd/CNumericsSIMD.hpp
pcarruscag: It will work since CEulerVariable is used in all of those in one way or another, i.e. CNSVariable is also a CEulerVariable with extra methods.

pcarruscag:
Thank you for the review Catarina, based on your comments I will try to explain the new structure better.

### The interface
These numerics classes are still abstract and so the outside world only needs to known about their interface:
```c++
class CNumericsSIMD {
public:
  virtual void ComputeFlux(...) const = 0;
};
```
Any class that wants to be a simd numerics needs to inherit from this base and in so doing is forced to implement `ComputeFlux` (because it is pure virtual).

### Static decorator pattern
Now, in this structure we don't have convective and viscous classes separate. Instead, we have convective schemes that can be "decorated" with viscous fluxes. For example:
```c++
template<int nDim>
class CCompressibleViscousFlux : public CNumericsSIMD {
protected:
  void viscousTerms(...) const {...}
};
```
Here note that this class template is derived from CNumericsSIMD but it does not implement `ComputeFlux`, thus it cannot be instantiated by itself. The idea is that convective schemes can use these viscous fluxes as base class (thereby linking them to CNumericsSIMD) to access the `viscousTerms` method (when we don't want viscous terms we just use a dummy viscous class).
Note also the template parameter `nDim`, this is because we want to create specific versions of the numerical schemes for 2D and 3D (we "want" this because it allows static allocation and unrolling loops perfectly).

Then convective classes also need to be class templates so that we can programatically change their base class:
```c++
template<class Base>
class MyConvectiveScheme : public Base {
public:
  void ComputeFlux(...) const {
    // do my own thing
    Base::viscousTerms(...);
    // update the linear system with the result
  }
};
```
Now to create an instance of this class template we do for example:
```c++
auto obj = new MyConvectiveScheme< CCompressibleViscousFlux<2> >(...);
```
which would create an object for 2D problems with viscous terms.

And so we need at least 4 instantiations of these class templates, 2D/3D with or without viscous terms, and this is done in the factory method implemented in CNumericsSIMD.cpp, which is the only cpp in this entire implementation.

### Static polymorphism
Another concept used in this implementation for efficiency is static polymorphism.
For example in the non vectorized numerics we have a family of Roe schemes since a lot of code is shared, and the only difference is how the dissipation terms are computed. There this is done with virtual functions, here we want none of that.
Virtual functions allow a parent class to dynamically call methods of its children, we want to do this statically and so we need to let the parent class know who is deriving from it.
```c++
template<class Derived>
class Parent {
public:
  void parentMethod() {
    // "I know I am also a Derived and so I can cast myself."
    auto derived = static_cast<Derived*>(this);
    // "now I can use a method of derived"
    derived->childMethod();
  }
};

// A derived class needs to inform the parent about itself
class Child : public Parent<Child> {
public:
  void childMethod() {...}
};
```

Why is this better? Note that 2 of these derived classes don't actually have the same parent, i.e. one inherits from `Parent<ChildA>` the other from `Parent<ChildB>` this means that 2 versions of Parent were instantiated specifically for each derived class, this allows code to be inlined and optimized for each, an ability lost with virtual functions.

### Putting it all together
For vectorized central schemes we have something like:
```c++
// Intermediate class for centered schemes, note the 2 template parameters
template<class Derived, class Base>
class CCenteredBase : public Base {
public:
  // Main public method implemented here making use of "Derived" and "Base".
  void ComputeFlux(...) const final {
    ... // gather data, do some computation
    derived->DissipationTerms(...); // static polymorphism
    Base::ViscousTerms(...); // static decorator
    ... // update linear system
  }
}

// A final centered scheme, which is what we instantiate, with some viscous decorator.
template<class Decorator>
class CJSTScheme : public CCenteredBase<CJSTScheme<Decorator>, Decorator> {
protected:
  void DissipationTerms(...) const {...}
}
```

PR #1030, opened 2020-06-20
Improving and cleaning element orientation checks

pcarruscag:
## Proposed Changes
General cleanup / optimization of the orientation checks.
Plus I tried to detect distorted, rather than flipped, elements separately as the node swaps done in `CPrimalGrid::Change_Orientation` can only fix flipped elements, if they are "twisted" changing the orientation does not fix the element, e.g. vtk_pixel vs vtk_quad.

## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4348,429 +4348,294 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+      if (test_1 && test_2) {
         elem[iElem]->Change_Orientation();
         quad_flip++;
       }
+      else if (test_1 || test_2) {
+        /*--- If one test fails and the other passes the
+         * element probably has serious problems. ---*/
+        SU2_OMP_ATOMIC
+        quad_error++;
+      }
```
pcarruscag, line 4414: This is the "detect distorted vs flipped" logic

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
 
+        if (test_1+test_2+test_3+test_4 >= 3) {
+          /*--- If 3 or 4 tests fail there is > 75% chance flipping is the right choice. ---*/
+          bound[iMarker][iElem_Surface]->Change_Orientation();
+          quad_flip++;
+        }
+        else if (test_1+test_2+test_3+test_4 == 2) {
+          /*--- If 50/50 we cannot be sure of what to do -> report to user.
+           * If only one test fails it is probably (75%) due to skewness or warping. ---*/
+          quad_error++;
```
pcarruscag, line 4621: For surface quads I made the decision to change orientation a bit more statistical, for stretched grids it was relatively easy to get false negatives.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
-  for (iElem = 0; iElem < nElem; iElem++) {
+void CPhysicalGeometry::Check_IntElem_Orientation(const CConfig *config) {
+
+  unsigned long tria_flip=0, quad_flip=0, tet_flip=0, prism_flip=0, hexa_flip=0, pyram_flip=0;
+  unsigned long quad_error=0, prism_error=0, hexa_error=0, pyram_error=0;
+
+  SU2_OMP_PARALLEL_(reduction(+:tria_flip,quad_flip,tet_flip,prism_flip,hexa_flip,pyram_flip)) {
+
+  /*--- Lambda to test triangles. Normal should be positive in the z direction (right hand rule). ---*/
+  auto checkTria = [&](unsigned long iElem, int Node_1, int Node_2, int Node_3) {
```
economon, line 4357: I think bool instead of auto here would be more readable

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    auto Coord_1 = nodes->GetCoord(elem[iElem]->GetNode(Node_1));
+    auto Coord_2 = nodes->GetCoord(elem[iElem]->GetNode(Node_2));
+    auto Coord_3 = nodes->GetCoord(elem[iElem]->GetNode(Node_3));
+    constexpr int nDim = 2;
+    su2double a[nDim]={0.0}, b[nDim]={0.0};
```
economon, line 4362: c arrays -> std::array

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    GeometryToolbox::Distance(nDim, Coord_2, Coord_1, a);
+    GeometryToolbox::Distance(nDim, Coord_3, Coord_1, b);
+    return a[0]*b[1]-a[1]*b[0] < 0.0;
+  };
+
+  /*--- Lambda to test tetrahedrons, volume must be positive,
+   * the normal of any face must point towards the other point. ---*/
+  auto checkTetra = [&](unsigned long iElem, int Node_1, int Node_2, int Node_3, int Node_4) {
```
economon, line 4370: and here w bool vs auto again.. I'll leave it up to you

Common/src/geometry/CPhysicalGeometry.cpp
economon, line 4357: typical comment to check the default capture on the lambda :)

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
-      for (iDim = 0; iDim < nDim; iDim++) {
-        a[iDim] = 0.5*(Coord_2[iDim]-Coord_1[iDim]);
-        b[iDim] = 0.5*(Coord_3[iDim]-Coord_1[iDim]); }
-      test = a[0]*b[1]-b[0]*a[1];
-
-      if (test < 0.0) {
+      if (checkTria(iElem,0,1,2)) {
     	  elem[iElem]->Change_Orientation();
-    	  triangle_flip++;
+    	  tria_flip++;
```
economon, line 4395: tab here.. consider running clang-format

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    auto Coord_1 = nodes->GetCoord(elem[iElem]->GetNode(Node_1));
+    auto Coord_2 = nodes->GetCoord(elem[iElem]->GetNode(Node_2));
+    auto Coord_3 = nodes->GetCoord(elem[iElem]->GetNode(Node_3));
+    auto Coord_4 = nodes->GetCoord(elem[iElem]->GetNode(Node_4));
+    constexpr int nDim = 3;
+    su2double a[nDim]={0.0}, b[nDim]={0.0}, c[nDim]={0.0}, n[nDim]={0.0};
+    GeometryToolbox::Distance(nDim, Coord_2, Coord_1, a);
+    GeometryToolbox::Distance(nDim, Coord_3, Coord_1, b);
+    GeometryToolbox::Distance(nDim, Coord_4, Coord_1, c);
+    GeometryToolbox::CrossProduct(a,b,n);
```
economon, line 4380: If not too much trouble, I think it is more clear if all of the GeometryToolbox methods return outputs rather than using in/out params, e.g., something like
```
const std::array<su2double, nDim> a = GeometryToolbox::Distance(nDim, Coord_2, Coord_1);
```

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    return a[0]*b[1]-a[1]*b[0] < 0.0;
+  };
+
+  /*--- Lambda to test tetrahedrons, volume must be positive,
+   * the normal of any face must point towards the other point. ---*/
+  auto checkTetra = [&](unsigned long iElem, int Node_1, int Node_2, int Node_3, int Node_4) {
+    auto Coord_1 = nodes->GetCoord(elem[iElem]->GetNode(Node_1));
+    auto Coord_2 = nodes->GetCoord(elem[iElem]->GetNode(Node_2));
+    auto Coord_3 = nodes->GetCoord(elem[iElem]->GetNode(Node_3));
+    auto Coord_4 = nodes->GetCoord(elem[iElem]->GetNode(Node_4));
```
economon, line 4374: const auto (here and many other places below, please give it a check throughout)

Common/src/geometry/CPhysicalGeometry.cpp
economon: Hmm.. I am not sure what is the best way to proceed here. Gut instinct says we should just throw an error if we don't know for certain what to do. Anyway we can make more precise?

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    auto Coord_3 = nodes->GetCoord(Point_3);
+    auto Coord_4 = nodes->GetCoord(Point_4);
+    constexpr int nDim = 3;
+    su2double a[nDim]={0.0}, b[nDim]={0.0}, c[nDim]={0.0}, n[nDim]={0.0};
+    GeometryToolbox::Distance(nDim, Coord_2, Coord_1, a);
+    GeometryToolbox::Distance(nDim, Coord_3, Coord_1, b);
+    GeometryToolbox::Distance(nDim, Coord_4, Coord_1, c);
+    GeometryToolbox::CrossProduct(a,b,n);
+    return GeometryToolbox::DotProduct(nDim,n,c) < 0.0;
+  };
```
economon, line 4537: If this is the same lambda as above for the other checkTetra, moving to a small separate function avoids reproduction. You could put it in an anonymous namespace at the top of the file, for instance, so it is only available in this translation unit

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
-            bound[iMarker][iElem_Surface]->Change_Orientation();
-            triangle_flip++;
-          }
+        /*--- The normal of the triangle formed by the line and domain
+         * point should point in the positive z direction. ---*/
+        constexpr int nDim = 2;
+        su2double a[nDim]={0.0}, b[nDim]={0.0};
+        GeometryToolbox::Distance(nDim, Coord_2, Coord_1, a);
+        GeometryToolbox::Distance(nDim, Coord_3, Coord_1, b);
+        bool test = a[0]*b[1]-a[1]*b[0] < 0.0;
```
economon, line 4576: this looks like checkTria - consider moving that lambda to a small function too so you can reuse if possible

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: That is not how lambdas work.

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: not possible!

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: not possible, need to capture "this"

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: I encourage you to try that for yourself and see what the compiler tells you https://gcc.godbolt.org/

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: Why?! The scope for those variables is minuscule and the functions I pass them to are const correct, contrary to 90% of SU2, I honestly feel you are trying to waste my time.

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: Deleting 50% of duplicated code is not good enough?

PR #1031, opened 2020-06-20
Small improvement to least squares gradients for non-periodic problems

pcarruscag:
## Proposed Changes
The computation was done in two loops to account for period comms in between.
When there are no periodic comms one loop is sufficient, better locality, more effective AD accumulation.
To do this the function was split in two, which is also good.

## Related Work
#594

## PR Checklist

### DIFF
--- a/SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
+++ b/SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
@@ -148,7 +148,7 @@ void computeGradientsLeastSquares(CSolver* solver,
 
     if (periodic)
     {
-      /*--- A second loop is required after periodic comms. ---*/
+      /*--- A second loop is required after periodic comms, checkpoint the preacc. ---*/
 
       for (size_t iDim = 0; iDim < nDim; ++iDim)
         for (size_t jDim = 0; jDim < nDim; ++jDim)
@@ -163,7 +163,7 @@ void computeGradientsLeastSquares(CSolver* solver,
     else {
       /*--- Periodic comms are not needed, solve the LS problem for iPoint. ---*/
 
-      solveLeastSquares(iPoint, false, nDim, varBegin, varEnd, gradient, Rmatrix);
+      solveLeastSquares(iPoint, false, nDim, varBegin, varEnd, Rmatrix, gradient);
     }
   }
 
@@ -181,7 +181,7 @@ void computeGradientsLeastSquares(CSolver* solver,
 
     SU2_OMP_FOR_DYN(chunkSize)
     for (size_t iPoint = 0; iPoint < nPointDomain; ++iPoint)
-      solveLeastSquares(iPoint, true, nDim, varBegin, varEnd, gradient, Rmatrix);
+      solveLeastSquares(iPoint, true, nDim, varBegin, varEnd, Rmatrix, gradient);
   }
 
   /*--- If no solver was provided we do not communicate ---*/
@@ -196,14 +196,17 @@ void computeGradientsLeastSquares(CSolver* solver,
 
 }
 
+/*!
+ * \brief Solve the LS problem prepared above.
+ */
 template<class GradientType, class RMatrixType>
 FORCEINLINE void solveLeastSquares(size_t iPoint,
                                    bool periodic,
                                    size_t nDim,
                                    size_t varBegin,
                                    size_t varEnd,
-                                   GradientType& gradient,
-                                   RMatrixType& Rmatrix) {
+                                   const RMatrixType& Rmatrix,
+                                   GradientType& gradient) {
   constexpr size_t MAXNDIM = 3;
 
   /*--- Entries of upper triangular matrix R. ---*/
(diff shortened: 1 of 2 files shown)


SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -152,155 +146,190 @@ void computeGradientsLeastSquares(CSolver* solver,
...
     AD::SetPreaccIn(r22);
+  }
 
-    if (r11 >= 0.0) r11 = sqrt(r11);
-    if (r11 >= 0.0) r12 /= r11; else r12 = 0.0;
-    su2double tmp = r22-r12*r12;
-    if (tmp >= 0.0) r22 = sqrt(tmp); else r22 = 0.0;
+  if (r11 >= 0.0) r11 = sqrt(r11);
+  if (r11 >= 0.0) r12 /= r11; else r12 = 0.0;
+  su2double tmp = r22-r12*r12;
```
economon, line 229: const?

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -152,155 +146,190 @@ void computeGradientsLeastSquares(CSolver* solver,
...
+ */
+template<class GradientType, class RMatrixType>
+FORCEINLINE void solveLeastSquares(size_t iPoint,
+                                   bool periodic,
+                                   size_t nDim,
+                                   size_t varBegin,
+                                   size_t varEnd,
+                                   const RMatrixType& Rmatrix,
+                                   GradientType& gradient) {
+  constexpr size_t MAXNDIM = 3;
```
economon, line 210: This seems like something we should have defined in one place, like option_structure.hpp

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -152,155 +146,190 @@ void computeGradientsLeastSquares(CSolver* solver,
...
+  if (nDim == 3) detR2 *= r33*r33;
 
-    bool singular = false;
+  /*--- Detect singular matrix ---*/
 
-    if (detR2 <= EPS) {
-      detR2 = 1.0;
-      singular = true;
-    }
+  bool singular = (detR2 <= EPS);
```
economon, line 265: I see EPS is set to 1e-16. Perhaps we should base it on std::numeric_limits<T>::epsilon()

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -152,155 +146,190 @@ void computeGradientsLeastSquares(CSolver* solver,
...
-        Smatrix[2][0] = Smatrix[0][2];
-        Smatrix[2][1] = Smatrix[1][2];
-        Smatrix[2][2] = (z33*z33)/detR2;
-      }
+      su2double z11 = r22*r33;
+      su2double z12 =-r12*r33;
+      su2double z13 = r12*r23-r13*r22;
+      su2double z22 = r11*r33;
+      su2double z23 =-r11*r23;
+      su2double z33 = r11*r22;
```
economon, line 290: check for const

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -152,155 +146,190 @@ void computeGradientsLeastSquares(CSolver* solver,
...
+ * \brief Solve the LS problem prepared above.
+ */
+template<class GradientType, class RMatrixType>
+FORCEINLINE void solveLeastSquares(size_t iPoint,
+                                   bool periodic,
+                                   size_t nDim,
+                                   size_t varBegin,
+                                   size_t varEnd,
+                                   const RMatrixType& Rmatrix,
+                                   GradientType& gradient) {
```
economon, line 209: more common practice is to use a pointer for output params at the end of the args list, i.e., GradientType* gradient

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -105,20 +107,12 @@ void computeGradientsLeastSquares(CSolver* solver,
       /*--- Distance vector from iPoint to jPoint ---*/
 
       su2double dist_ij[MAXNDIM] = {0.0};
-
-      for (size_t iDim = 0; iDim < nDim; ++iDim)
-        dist_ij[iDim] = coord_j[iDim] - coord_i[iDim];
+      GeometryToolbox::Distance(nDim, coord_j, coord_i, dist_ij);
```
economon, line 110: same suggestion to return the distance from the routine instead of an in/out arg

pcarruscag: Same reply as to why it can't be done.

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
pcarruscag: No it is not, references should be preferred to pointers.

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
pcarruscag: You realize this is not new code right?

pcarruscag:
included in #1022

PR #1038, opened 2020-07-02
Improving and cleaning element orientation checks

pcarruscag:
## Proposed Changes
Reopen #1030 after changes.

## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
 
+        if (test_1+test_2+test_3+test_4 >= 3) {
+          /*--- If 3 or 4 tests fail there is > 75% chance flipping is the right choice. ---*/
+          bound[iMarker][iElem_Surface]->Change_Orientation();
+          quad_flip++;
+        }
+        else if (test_1+test_2+test_3+test_4 == 2) {
+          /*--- If 50/50 we cannot be sure of what to do -> report to user.
+           * If only one test fails it is probably (75%) due to skewness or warping. ---*/
+          quad_error++;
```
pcarruscag, line 4621: It is probably not the best way, but a bullet proof check probably needs to go over many combinations of nodes, as for making it an hard error, I don't like the idea very much because the skewness might by due to an hard to mesh geometry, and I don't wan't people to complain I broke their cases.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4346,431 +4346,305 @@ void CPhysicalGeometry::PrepareAdjacency(CConfig *config) {
...
+    const auto Coord_2 = nodes->GetCoord(Point_2);
+    const auto Coord_3 = nodes->GetCoord(Point_3);
+    const auto Coord_4 = nodes->GetCoord(Point_4);
+    constexpr int nDim = 3;
+    su2double a[nDim]={0.0}, b[nDim]={0.0}, c[nDim]={0.0}, n[nDim]={0.0};
+    GeometryToolbox::Distance(nDim, Coord_2, Coord_1, a);
+    GeometryToolbox::Distance(nDim, Coord_3, Coord_1, b);
+    GeometryToolbox::Distance(nDim, Coord_4, Coord_1, c);
+    GeometryToolbox::CrossProduct(a,b,n);
+    return GeometryToolbox::DotProduct(nDim,n,c) < 0.0;
```
pcarruscag, line 4536: These functions are similar to the ones used to check the volume elements, I did not make this another member function of CGeometry or a free function of the geometry toolbox mostly out of laziness, but I guess some arguments in favor are to avoid crowding the CGeometry namespace, and because here what I wanted to "automate" was the boiler plate of fetching coordinates.
In any case I don't mind if someone moves this in the future.

PR #1044, opened 2020-07-08
Intermediate base class for flow solvers to avoid duplication.

pcarruscag:
## Proposed Changes
As discussed in today's developers meeting.

## Related Work
#1014

## PR Checklist
- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1050, opened 2020-07-17
[WIP] Piperno slope limiter and more flexible MUSCL scheme

bmunguia:
## Proposed Changes
This PR adds the [Piperno slope limiter](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=71E315A30221171183B3425C51A4A33B?doi=10.1.1.493.3132&rep=rep1&type=pdf) and introduces a parameter to control the relative upwind vs central differencing in the MUSCL scheme:
Screen Shot 2020-07-17 at 11 35 28 AM
where 
- X=0 gives the original unstructured formulation for second-order variable extrapolation (default)
- X=-1 gives second-order fully-upwinded MUSCL-type variable extrapolation
- X=1/2 gives third-order variable extrapolation to the cell face
- X=1/3 gives a third-order approximation to the derivative at the node (the value used in the Piperno limiter to achieve 4th order dissipation)

The reasons for these additions are
- The Piperno limiter is implemented in Inria's Wolf solver and seems to be used in a lot of their mesh adaptation work (see these references for [Euler](https://hal.inria.fr/hal-02378738v2/document) and [RANS](https://hal.inria.fr/tel-01962318/file/these_archivage_3570223.pdf))
- Some preliminary tests with the Piperno limiter have converged solutions on triangulated BL meshes that @jayantmukho and I couldn't get to converge with other limiters in SU2 (e.g. triangulated DSMA661 meshes from TMR)
- The MUSCL parameter exists in other solvers (FUN3D, CFL3D, Wolf) and is used by the NASA codes for many of the TMR cases (e.g. CFL3D uses 1/3 and FUN3D uses 1/2 for the DSMA661 case)

Further testing still needs to be done on the effect of the MUSCL parameter with other limiters, but leaving the value as default won't affect any of the test cases.

## Relevant config parameters
Piperno limiter:
```
SLOPE_LIMITER_FLOW= PIPERNO
```
MUSCL parameter:
```
% Parameter to control upwinding in MUSCL extrapolation
% Kappa = 1/3 for Piperno limiter, Default = 0 otherwise
MUSCL_KAPPA= 0.0
```

## PR Checklist


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+              Project_Grad_j *= 1.0 + (1.5*InvR_j + 1.0)*pow(InvR_j - 1.0, 3.0);
+            }
+          }
+          else {
+            Project_Grad_i *= Limiter_i[iVar];
+            Project_Grad_j *= Limiter_j[iVar];
+          }
+        }
+        Primitive_i[iVar] = V_i[iVar] + Project_Grad_i;
+        Primitive_j[iVar] = V_j[iVar] - Project_Grad_j;
```
economon, line 3181: Perhaps you can extract this into a small function that can be reused instead of repeating (can do the same with Van Albada)?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2625,12 +2625,12 @@ void CEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_con
...
-  bool muscl            = (config->GetMUSCL_Flow() || (cont_adjoint && config->GetKind_ConvNumScheme_AdjFlow() == ROE));
-  bool limiter          = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) && (InnerIter <= config->GetLimiterIter());
-  bool center           = (config->GetKind_ConvNumScheme_Flow() == SPACE_CENTERED) || (cont_adjoint && config->GetKind_ConvNumScheme_AdjFlow() == SPACE_CENTERED);
-  bool van_albada       = config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE;
+  const unsigned long InnerIter = config->GetInnerIter();
+  const bool cont_adjoint     = config->GetContinuous_Adjoint();
+  const bool muscl            = (config->GetMUSCL_Flow() || (cont_adjoint && config->GetKind_ConvNumScheme_AdjFlow() == ROE));
+  const bool limiter          = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) && (InnerIter <= config->GetLimiterIter());
+  const bool center           = (config->GetKind_ConvNumScheme_Flow() == SPACE_CENTERED) || (cont_adjoint && config->GetKind_ConvNumScheme_AdjFlow() == SPACE_CENTERED);
+  const bool edge_limiter     = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE) || (config->GetKind_SlopeLimit_Flow() == PIPERNO);
```
pcarruscag, line 2633: This logic should be moved into CConfig for the reasons mentioned in #770.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
-        else {
-          Primitive_i[iVar] = V_i[iVar] + Project_Grad_i;
-          Primitive_j[iVar] = V_j[iVar] + Project_Grad_j;
-        }
+          else if (piperno) {
+            const su2double R_i = 0.5*V_ij/(Project_Grad_i+EPS);
+            const su2double R_j = 0.5*V_ij/(Project_Grad_j+EPS);
+
+            const su2double InvR_i = (Project_Grad_i)/(0.5*V_ij+EPS);
+            const su2double InvR_j = (Project_Grad_j)/(0.5*V_ij+EPS);
```
pcarruscag, line 3148: I think adding EPS here will not protect against division by zero, also if InvR is the inverse of R...

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+            if (R_j < 0.0) {
+              Project_Grad_j = 0.0;
+            }
+            else if (R_j < 1.0) {
+              Project_Grad_j *= (3.0*pow(InvR_j, 2.0) - 6.0*InvR_j + 19.0)
+                               / (pow(InvR_j, 3.0) - 3.0*InvR_j + 18.0);
+            }
+            else {
+              Project_Grad_j *= 1.0 + (1.5*InvR_j + 1.0)*pow(InvR_j - 1.0, 3.0);
+            }
```
pcarruscag, line 3173: Please move this piece-wise function to a limiter details class (CLimiterDetails.hpp) as right now it is duplicated here and again for the incompressible solver.
Or make it one of the helper functions in that same file.

Side note, any chance this can be written without branching? Otherwise I'll have a bad time implementing it in #1022.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
         for (iDim = 0; iDim < nDim; iDim++) {
-          Project_Grad_i += Vector_ij[iDim]*Gradient_i[iVar][iDim];
-          Project_Grad_j -= Vector_ij[iDim]*Gradient_j[iVar][iDim];
+          Project_Grad_i += (1.0-Kappa)*Vector_ij[iDim]*Gradient_i[iVar][iDim];
+          Project_Grad_j += (1.0-Kappa)*Vector_ij[iDim]*Gradient_j[iVar][iDim];
         }
 
         if (limiter) {
           if (van_albada) {
-            su2double V_ij = V_j[iVar] - V_i[iVar];
```
pcarruscag, line 3135: Instead of if / if else / else, this should be a switch.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+        const su2double V_ij = V_j[iVar] - V_i[iVar];
+
+        su2double Project_Grad_i = 0.5*Kappa*V_ij;
+        su2double Project_Grad_j = 0.5*Kappa*V_ij;
 
         for (iDim = 0; iDim < nDim; iDim++) {
-          Project_Grad_i += Vector_ij[iDim]*Gradient_i[iVar][iDim];
-          Project_Grad_j -= Vector_ij[iDim]*Gradient_j[iVar][iDim];
+          Project_Grad_i += (1.0-Kappa)*Vector_ij[iDim]*Gradient_i[iVar][iDim];
+          Project_Grad_j += (1.0-Kappa)*Vector_ij[iDim]*Gradient_j[iVar][iDim];
```
pcarruscag, line 3135: It would be more efficient to bake this (1-k) into Vector_ij as it is computed.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3125,60 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
 
       for (iVar = 0; iVar < nPrimVarGrad; iVar++) {
 
-        su2double Project_Grad_i = 0.0;
-        su2double Project_Grad_j = 0.0;
+        const su2double V_ij = V_j[iVar] - V_i[iVar];
```
pcarruscag, line 3128: V_ij is not used by the point limiters, it should not be computed by default.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -100,7 +100,8 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
                         (config->GetKind_ConvNumScheme_Flow() == SPACE_UPWIND);
   /*--- Only consider flow limiters for cell-based limiters, edge-based would need to be recomputed. ---*/
   const bool limiterFlow = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) &&
-                           (config->GetKind_SlopeLimit_Flow() != VAN_ALBADA_EDGE);
+                           (config->GetKind_SlopeLimit_Flow() != VAN_ALBADA_EDGE) &&
+                           (config->GetKind_SlopeLimit_Flow() != PIPERNO);
```
pcarruscag, line 104: Same comment on moving the "edge limiter" logic to CConfig.

SU2_CFD/src/solvers/CEulerSolver.cpp
bmunguia: I moved the edge-based limiters to CLimiterDetails.hpp.

> Side note, any chance this can be written without branching?
I'm not sure how to do that since the function is piece-wise. Any suggestions?

SU2_CFD/src/solvers/CEulerSolver.cpp
bmunguia: It might not used by the limiters, but it's used by the generalized MUSCL scheme whenever Kappa != 0

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -74,11 +74,30 @@ namespace LimiterHelpers
...
   }
 
+  inline su2double vanAlbadaFunction(su2double proj, su2double delta, su2double eps)
+  {
+    return delta*(2.0*proj + delta) / (4*pow(proj, 2) + pow(delta, 2) + pow(eps, 2));
+  }
+
+  inline su2double pipernoFunction(su2double proj, su2double delta)
+  {
+    if (proj*delta <= 0.0) {
```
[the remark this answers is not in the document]
bmunguia: It should be. I was trying to guard against division by zero while incorporating `if r<=0`, but I guess this implementation is incorrect for `delta=0`

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Doh, missed that sorry, will that require some central blending of the flux Jacobians?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -11385,30 +11391,24 @@ void CEulerSolver::BC_ActDisk_VariableLoad(CGeometry *geometry, CSolver **solver
   su2double Pressure_out, Density_out,
   Pressure_in, Density_in;
 
-  su2double C[MAXNDIM], Prop_Axis[MAXNDIM], R, r[MAXNDIM], r_;
+  su2double Prop_Axis[MAXNDIM];
```
pcarruscag, line 11394: Thanks, these were fixed in #1044 already, along with the indentation of the function.

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -74,11 +74,30 @@ namespace LimiterHelpers
     return (y + delta*proj) / (y + 2*proj*proj);
   }
 
+  inline su2double vanAlbadaFunction(su2double proj, su2double delta)
+  {
+    return delta*(2.0*proj + delta) / (4*pow(proj, 2) + pow(delta, 2) + epsilon());
+  }
+
+  inline su2double pipernoFunction(su2double proj, su2double delta)
```
pcarruscag, line 82: Thanks for moving this.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3122,28 +3124,32 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+              Limiter_j[iVar] = LimiterHelpers::vanAlbadaFunction(Project_Grad_j, V_ij);
+              break;
+            case PIPERNO:
+              Limiter_i[iVar] = LimiterHelpers::pipernoFunction(Project_Grad_i, V_ij);
+              Limiter_j[iVar] = LimiterHelpers::pipernoFunction(Project_Grad_j, V_ij);
+              break;
           }
-          Primitive_i[iVar] = V_i[iVar] + Limiter_i[iVar]*Project_Grad_i;
-          Primitive_j[iVar] = V_j[iVar] + Limiter_j[iVar]*Project_Grad_j;
+          Project_Grad_i = Limiter_i[iVar]*((1.0-Kappa)*Project_Grad_i + 0.5*Kappa*V_ij);
```
pcarruscag, line 3148: Is it standard practice to apply the limiter to the sum of the projection with the central average?
I'm asking because the point limiters are based only on the projection and the average is guaranteed to be TVD.

SU2_CFD/include/limiters/CLimiterDetails.hpp
```diff
@@ -74,11 +74,29 @@ namespace LimiterHelpers
...
+  {
+    if (proj*delta <= 0.0) {
+      return 0.0;
+    }
+    else {
+      const su2double r = proj/delta;
+      return (r >= 1.0) ? su2double((3.0*pow(r, 2.0) - 6.0*r + 19.0) / (pow(r, 3.0) - 3.0*r + 18.0))
+                        : su2double(1.0 + (1.5*r + 1.0)*pow(r - 1.0, 3.0));
```
pcarruscag, line 90: Regarding making this branch-less,
r = max(0.0, proj / (delta + eps*sign(delta))); // handles div by zero and the < 0 branch.
return min("0 < r < 1", "1 < r"); // at least as they are currently defined.
where sign(x) = 1 - 2*(x<0);
(you can also leave it to me, for when I update the reconstruction logic in #1022)

SU2_CFD/src/solvers/CEulerSolver.cpp
bmunguia: > Is it standard practice to apply the limiter to the sum of the projection with the central average?

From what I've seen, it is standard. The definitions I've seen for edge limiters have the sum of the projection and central difference baked into the limiter definition.

> I'm asking because the point limiters are based only on the projection and the average is guaranteed to be TVD.

Now I'm wondering whether the point limiters should be computed with the projection and delta, or the sum and delta. The Inria papers I linked above, and some others I've found like [this NASA/NIA paper](http://www.ossanworld.com/hiroakinishikawa/My_papers/white_nishikawa_baurle_aiaa2019-0127.pdf) seem to use the sum in their limiter functions.

SU2_CFD/src/solvers/CEulerSolver.cpp
bmunguia: I guess so, because now we're feeding V_i + 0.5*k*Lim_i*(V_j - V_i) (and similarly for the right state) into CNumerics.

Should I just pass (k*Lim_i, k*Lim_j) into CNumerics and manipulate the Jacobian there?

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: That sounds like a reasonable modification.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: It would be better to pass Lim_i/j and let the numerics read k from the config.

But we should test if the central blending is beneficial, as it will make the system matrix less diagonally dominant.
Perhaps for testing only k needs to be considered (lim_i/j assumed 1), I can help testing if you hack this into Roe (I have a score to settle with some cases that did not work very well with upwind schemes)

bmunguia: I added a function `CUpwRoeBase_Flow::GetMUSCLJac()` that multiplies the Jacobians (dF/dU) by (dU/dV)\*(dV_f/dV_i)\* (dV/dU) where dV_f/dV_i is diag(1+0.5\*k\*(Lim_j-Lim_i)), and dU/dV and dV/dU are taken from Hirsch
Screen Shot 2020-07-22 at 1 31 24 PM
Screen Shot 2020-07-22 at 1 31 32 PM
I feel like the implementation isn't technically correct because the dV/dU should be evaluated at the nodal values, but from my very limited testing it doesn't seem to break anything (other than like half the regressions)

bmunguia:
It seems like the Pipero limiter isn't as promising as I initially thought, but here are some preliminary results I obtained on the DSMA 661 mesh with SA, Venkatakrishnan-Wang limiter, and varying MUSCL_KAPPA
lift_convg_dsma
drag_convg_dsma
I'm not sure what's causing the discrepancy in the lift with the NASA codes, but the drag looks pretty good, especially at higher kappa
[dsma661_sa.tar.gz](https://github.com/su2code/SU2/files/4962588/dsma661_sa.tar.gz)

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Cool, I will try it, I'm curious about the impact of the more accurate Jacobian vs a possibly simpler approximation that neglects the limiters or uses an average over all variables (which would then boil down to just scaling dF/dU and adding part of the "other" dF/dU and v.v.).
dV/dU are available in `CUpwAUSMPLUS_SLAU_Base_Flow::AccurateJacobian` (btw for those schemes the accurate Jacobian sometimes is worse, hence my curiosity) so if this ends up being the way forward we could factor them out.

pcarruscag:
> It seems like the Pipero limiter isn't as promising as I initially thought

Accuracy or stability wise? It bugs me that the limiter function starts at 0 (when r=0) while all others start at 1 and actually go above one until r becomes greater than 1. Maybe the definition of r for Piperno needs some adjustment for unstructured grids? In the paper R is the ratio of consecutive deltas, and doesn't phi(r) (the piece-wise part) need to be multiplied by another term to give the final psi(R)?

> ...especially at higher kappa

With kappa close to 1 my gut feeling says the limiters will be off everywhere (r~1 by definition) unless the gradient goes wild.
Also the left and right states will be very similar, and since the Roe dissipation term is based on their difference..., it could explain why drag converges so quickly for that problem.
With kappa that close to 1 I wonder how close Roe becomes to a central scheme with matrix dissipation.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: > other than like half the regressions

I have a script for that :smiling_imp:

bmunguia:
> Accuracy or stability wise? 

Stability-wise. On the DSMA cases I posted, the edge limiters both stall.

> It bugs me that the limiter function starts at 0 (when r=0) while all others start at 1 and actually go above one until r becomes greater than 1.

Yeah, maybe it needs to be reworked, but I thought that most limiter functions started at 0? E.g. from Venkatakrishnan

Screen Shot 2020-07-23 at 9 22 37 AM
Screen Shot 2020-07-23 at 9 20 39 AM

> Maybe the definition of r for Piperno needs some adjustment for unstructured grids? In the paper R is the ratio of consecutive deltas

He discusses that in section 5.1 of his paper. Depending on whether your upwind gradient comes from the node or the triangle upwind of the edge, you have
Screen Shot 2020-07-23 at 9 34 49 AM
I modified the upwind difference to be consistent with this definition (which is why the regressions that use van Albada failed, the rest seem to be a segfault that I'll fix soon).
The choice of r is also discussed briefly [here](https://github.com/su2code/SU2/files/4967575/r_factor_Li.pdf).

> doesn't phi(r) (the piece-wise part) need to be multiplied by another term to give the final psi(R)?

The other term (1/3 + 2/3*R) is baked into when I recompute the limited Project_Grads. Since we multiply psi(R) by the upwind difference deltaU_u, it becomes U_i + 1/2\*(1/3 deltaU_u + 2/3 deltaU_c)\*psi(R), which we get for kappa=1/3

> With kappa close to 1 my gut feeling says the limiters will be off everywhere (r~1 by definition) unless the gradient goes wild.
> Also the left and right states will be very similar, and since the Roe dissipation term is based on their difference..., it could explain why drag converges so quickly for that problem.

Yeah that makes sense to me.

pcarruscag:
They start at zero when this r is defined as the ratio of consecutive deltas, but this ratio doesn't really exist for unstructured grids.
The way I understand limiters is, if the projection is larger than the delta the limiter should kick in, if the projection is zero there is no reason to limit. If you plot the Venkatakrishnan or Van-Albada functions against proj/delta you'll see both of them do this (so does the Piperno if multiplied by the 1/3 + 2/3 delta/proj term, hence my comment about this term).

bmunguia:
Oh I see now. We could do
```
const su2double r = max(epsilon(), proj / (delta + sign*epsilon()));
return (1.0/3.0+2.0/(3.0*r))*min((3.0*pow(r, 2.0) - 6.0*r + 19.0) / (pow(r, 3.0) - 3.0*r + 18.0),
               1.0 + (1.5*r + 1.0)*pow(r - 1.0, 3.0));
```
in CLimiterDetails to get a nonzero limiter, and replace
```
Project_Grad_i = 0.5*Limiter_i[iVar]*((1.0-Kappa)*Project_Grad_i + (1.0+Kappa)*V_ij);
```
with
```
Project_Grad_i *= Limiter_i[iVar];
```
in CEulerSolver.

Similarly, the form of van Albada shown both in the original paper and Venkatakrishnan's paper is
```
((pow(delta,2)+eps2)*proj +(pow(proj,2)+eps2)*delta)  / (pow(proj, 2) + pow(delta, 2) + 2*eps2);
```
where we have
```
delta*(proj + delta) / (pow(proj, 2) + pow(delta, 2) + epsilon());
```
Should I modify that too?

pcarruscag:
I could be missing something but this version:
((pow(delta,2)+eps2)*proj +(pow(proj,2)+eps2)*delta))  / (pow(proj, 2) + pow(delta, 2) + 2*eps2);
looks like it would give you the limited projection rather than the limiter (cubic terms on top, quadratic on the bottom), unless eps2 takes on a value that disables limiting in smooth regions, I see no benefit changing how it is written (they are equivalent with eps -> 0)

bmunguia:
Ok, it looks like I can actually get convergence with Piperno on my DSMA case after including that factor of 1/3 + 2/(3r)

bmunguia:
> I've also wondered about whether the Jacobians (even dF/dU) should be evaluate at the reconstructed state (as currently done) or at the un-reconstructed... This can be easily tested in #1022 (the reconstruction is inside the numerics).

@pcarruscag coming back to this, is the plan with #1022 to exclusively use CNumericsSIMD, or keep the USE_VECTORIZATION option and have both Roe implementations available? I was thinking of merging feature_simd_numerics into this branch, and was wondering if I'd need to implement the reconstruction Jacobian for both Roe implementations or if I should just stick to CNumericsSIMD.

pcarruscag:
I think ideally we would have only the speedy stuff, but that is a very
long term goal (and ideally a community decision).
Medium term they will have to coexist due to:
1) Features I have not re-implemented like, reconstruction checks,
uncertainty stuff.
2) "Boundary numerics", I need to think of a clever way to use
CNumericsSIMD on boundaries (this should not matter for what you are
working on).

With that said, I implemented JST with matrix dissipation and did not
bother making 2 versions of it...

So, you choose, nothing would make me happier than seeing people starting
to use that framework.

SU2_CFD/src/numerics/flow/convection/roe.cpp
```diff
@@ -28,12 +28,14 @@
...
 
 CUpwRoeBase_Flow::CUpwRoeBase_Flow(unsigned short val_nDim, unsigned short val_nVar, const CConfig* config,
-                                   bool val_low_dissipation) : CNumerics(val_nDim, val_nVar, config) {
+                                   bool val_low_dissipation, bool val_muscl) : CNumerics(val_nDim, val_nVar, config) {
 
   implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
   /* A grid is defined as dynamic if there's rigid grid movement or grid deformation AND the problem is time domain */
   dynamic_grid = config->GetDynamic_Grid();
   kappa = config->GetRoe_Kappa(); // 1 is unstable
+  muscl_kappa = 0.5*config->GetMUSCL_Kappa();
```
jayantmukho, line 37: Is there a reason for multiplying the config value by 0.5?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2308,7 +2308,10 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
     /*--- Roe Turkel preconditioning ---*/
 
     if (roe_turkel) {
-      numerics->SetVelocity2_Inf(GeometryToolbox::SquaredNorm(nDim, config->GetVelocity_FreeStream()));
+      su2double sqvel = 0.0;
+      for (iDim = 0; iDim < nDim; iDim ++)
+        sqvel += pow(config->GetVelocity_FreeStream()[iDim], 2);
+      numerics->SetVelocity2_Inf(sqvel);
```
pcarruscag, line 2314:
```suggestion
      numerics->SetVelocity2_Inf(GeometryToolbox::SquaredNorm(nDim, config->GetVelocity_FreeStream()));
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -61,8 +61,8 @@ void CIncNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container
   const bool muscl = config->GetMUSCL_Flow() && (iMesh == MESH_0);
   const bool center = (config->GetKind_ConvNumScheme_Flow() == SPACE_CENTERED);
   const bool limiter = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) && (InnerIter <= config->GetLimiterIter());
-  const bool van_albada = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE);
-
+  const bool edge_limiter = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE) || (config->GetKind_SlopeLimit_Flow() == PIPERNO);
+
```
pcarruscag, line 65:
```suggestion
  const bool edge_limiter = config->GetEdgeLimiter_Flow();

```

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -94,7 +94,8 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
...
```
pcarruscag, line 98:
```suggestion
                           !config->GetEdgeLimiter_Flow();
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2308,7 +2308,10 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
     /*--- Roe Turkel preconditioning ---*/
```
pcarruscag, line 2309:
```suggestion
    su2double limiter_i[MAXNVAR] = {0.0}, limiter_j[MAXNVAR] = {0.0}, ZeroVec[MAXNVAR] = {0.0};

    /*--- Roe Turkel preconditioning ---*/

```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2334,40 +2337,64 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
+            default:
+              lim_i = nodes->GetLimiter_Primitive(iPoint, iVar);
+              lim_j = nodes->GetLimiter_Primitive(jPoint, iVar);
+              break;
+          }
+
+          /*--- Limit projection ---*/
+
+          Project_Grad_i *= lim_i;
+          Project_Grad_j *= lim_j;
```
pcarruscag, line 2393:
```suggestion
          Project_Grad_i *= lim_i;
          Project_Grad_j *= lim_j;
          limiter_i[iVar] = lim_i;
          limiter_j[iVar] = lim_j;
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2415,6 +2442,8 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
       numerics->SetPrimitive(bad_i? V_i : Primitive_i,  bad_j? V_j : Primitive_j);
       numerics->SetSecondary(bad_i? S_i : Secondary_i,  bad_j? S_j : Secondary_j);
 
+      su2double ZeroVec[MAXNDIM+3] = {0.0};
+      numerics->SetLimiter(bad_i? ZeroVec : Limiter_i, bad_j? ZeroVec : Limiter_j);
```
pcarruscag, line 2446:
```suggestion
      numerics->SetLimiter(bad_i? ZeroVec : limiter_i, bad_j? ZeroVec : limiter_j);
```

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1658,8 +1659,8 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
               if (ideal_gas) {
 
                 for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
-                  numerics[iMGlevel][FLOW_SOL][conv_term] = new CUpwRoe_Flow(nDim, nVar_Flow, config, roe_low_dissipation);
-                  numerics[iMGlevel][FLOW_SOL][conv_bound_term] = new CUpwRoe_Flow(nDim, nVar_Flow, config, false);
+                  numerics[iMGlevel][FLOW_SOL][conv_term] = new CUpwRoe_Flow(nDim, nVar_Flow, config, roe_low_dissipation, muscl_flow_jac);
+                  numerics[iMGlevel][FLOW_SOL][conv_bound_term] = new CUpwRoe_Flow(nDim, nVar_Flow, config, false, false);
```
pcarruscag, line 1663: Coarse grids never use MUSCL

SU2_CFD/src/numerics/flow/convection/roe.cpp
```diff
@@ -224,6 +311,11 @@ CNumerics::ResidualType<> CUpwRoeBase_Flow::ComputeResidual(const CConfig* confi
   if (implicit) {
     GetInviscidProjJac(Velocity_i, &Energy_i, Normal, 0.5, Jacobian_i);
     GetInviscidProjJac(Velocity_j, &Energy_j, Normal, 0.5, Jacobian_j);
+
+    if (muscl) {
+      GetMUSCLJac(muscl_kappa, Jacobian_i, Limiter_i, Limiter_j, Velocity_i, &Density_i);
```
pcarruscag, line 316: In defining this boolean in the constructor it would make sense to check the value of MUSCL kappa to check if any work needs to be done.
And possible to include USE_ACCURATE_FLUX_JACOBIANS also in that boolean.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: github does not let me get this change in but it should fix the dangling pointer issues.

Anyway... I'll leave this PR for the stale bot now

PR #1057, opened 2020-08-01
Fix underrelaxation

bmunguia:
## Proposed Changes
There was a bug in the CTurbSolver::ComputeUnderRelaxation() that would return a negative UR if the change in solution was too negative (less than the allowable decrease), which could cause the solution to head in the wrong direction.  I'm not really sure why the turbulent solver version of this function was written differently than the flow solver (maybe the intention was to have a different allowable increase and decrease but they ended up being the same in the end?), but I changed it to make it consistent with the flow solver version.

Also, as I mentioned in #1036, the UR parameters for the flow solver and turbulent solver both took the min of each other and were never reset for steady, meaning it was non-increasing.

While on the topic of UR, I was wondering what the community's thoughts are on changing the SST models to use UR instead of solution clipping. The limits seem somewhat arbitrary to me (is there a ref?), and the tests I've done on my mesh adaptation cases have performed well where I've replaced clipping with UR. I realize this is a very limited dataset so if anyone has any strong opinions of why we should be clipping, let me know.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CTurbSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSolver.cpp
@@ -660,9 +660,8 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
   /* Loop over the solution update given by relaxing the linear
    system for this nonlinear iteration. */
 
-  su2double localUnderRelaxation    =  1.00;
-  const su2double allowableDecrease = -0.99;
-  const su2double allowableIncrease =  0.99;
+  su2double localUnderRelaxation =  1.00;
+  const su2double allowableRatio =  0.99;
 
   SU2_OMP_FOR_STAT(omp_chunk_size)
   for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
@@ -674,20 +673,14 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
         /* We impose a limit on the maximum percentage that the
          turbulence variables can change over a nonlinear iteration. */
 
-        const unsigned long index = iPoint*nVar + iVar;
-        su2double ratio = LinSysSol[index]/(nodes->GetSolution(iPoint, iVar)+EPS);
-        if (ratio > allowableIncrease) {
-          localUnderRelaxation = min(allowableIncrease/ratio, localUnderRelaxation);
-        } else if (ratio < allowableDecrease) {
-          localUnderRelaxation = min(fabs(allowableDecrease)/ratio, localUnderRelaxation);
+        const unsigned long index = iPoint * nVar + iVar;
+        su2double ratio = fabs(LinSysSol[index]) / (nodes->GetSolution(iPoint, iVar) + EPS);
+        if (ratio > allowableRatio) {
+          localUnderRelaxation = min(allowableRatio / ratio, localUnderRelaxation);
         }
 
       }
     }
-
-    /* Choose the minimum factor between mean flow and turbulence. */
-
-    localUnderRelaxation = min(localUnderRelaxation, solver_container[FLOW_SOL]->GetNodes()->GetUnderRelaxation(iPoint));
 
     /* Threshold the relaxation factor in the event that there is
      a very small value. This helps avoid catastrophic crashes due
--- a/SU2_CFD/src/solvers/CTurbSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSolver.cpp
@@ -674,7 +674,7 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
          turbulence variables can change over a nonlinear iteration. */
 
         const unsigned long index = iPoint * nVar + iVar;
-        su2double ratio = fabs(LinSysSol[index]) / (nodes->GetSolution(iPoint, iVar) + EPS);
+        su2double ratio = fabs(LinSysSol[index]) / (fabs(nodes->GetSolution(iPoint, iVar)) + EPS);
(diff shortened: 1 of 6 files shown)


SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -674,21 +673,15 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
...
+          localUnderRelaxation = min(allowableRatio / ratio, localUnderRelaxation);
         }
 
       }
     }
 
-    /* Choose the minimum factor between mean flow and turbulence. */
-
-    localUnderRelaxation = min(localUnderRelaxation, solver_container[FLOW_SOL]->GetNodes()->GetUnderRelaxation(iPoint));
-
```
jayantmukho, line 691: Does this, and the change in CFVMFlowSolverBase mean that there are two different underrelaxation factors, one for the mean solver, and one for the turbulent solver?

pcarruscag:
> While on the topic of UR, I was wondering what the community's thoughts are on changing the SST models to use UR instead of solution clipping

Do you mean user-specified UR instead / on top of the automatic one? I thought the clipping was necessary to get physical solutions (at least the lower bounds).

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -674,21 +673,15 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
...
          turbulence variables can change over a nonlinear iteration. */
 
-        const unsigned long index = iPoint*nVar + iVar;
-        su2double ratio = LinSysSol[index]/(nodes->GetSolution(iPoint, iVar)+EPS);
-        if (ratio > allowableIncrease) {
-          localUnderRelaxation = min(allowableIncrease/ratio, localUnderRelaxation);
-        } else if (ratio < allowableDecrease) {
-          localUnderRelaxation = min(fabs(allowableDecrease)/ratio, localUnderRelaxation);
+        const unsigned long index = iPoint * nVar + iVar;
+        su2double ratio = fabs(LinSysSol[index]) / (nodes->GetSolution(iPoint, iVar) + EPS);
```
pcarruscag, line 677: I'd put another fabs around GetSolution, even if it is not needed for turbulence, for when this code inevitably gets copied somewhere else.

SU2_CFD/src/solvers/CTurbSolver.cpp
pcarruscag: One of them needs to go, otherwise l guess we have the cyclic dependency problem @bmunguia mentioned.

bmunguia:
> > While on the topic of UR, I was wondering what the community's thoughts are on changing the SST models to use UR instead of solution clipping
> 
> Do you mean user-specified UR instead / on top of the automatic one? I thought the clipping was necessary to get physical solutions (at least the lower bounds).

I just meant the automatic one. Maybe I need to read up more on k-omega but I thought the solution just had to be positive to be physical, and an allowable ratio < 1 should prevent any negative solutions, assuming the initial solution is positive.

But for now, I think we should probably just get this bug fix in and I can look into the solution clipping separately.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -674,21 +673,15 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
...
+        if (ratio > allowableRatio) {
+          localUnderRelaxation = min(allowableRatio / ratio, localUnderRelaxation);
         }
 
       }
     }
 
-    /* Choose the minimum factor between mean flow and turbulence. */
-
-    localUnderRelaxation = min(localUnderRelaxation, solver_container[FLOW_SOL]->GetNodes()->GetUnderRelaxation(iPoint));
```
pcarruscag, line 690: I guess some numerical issues might occur if numbers get too large/small during convergence, although that sounds more like a single precision consideration.

Is it worth keeping just this min() between flow and turbulence for now? And maybe investigate if it makes sense as part of #1036?

bmunguia: Fair enough. I'll add it back in for now since it shouldn't hurt anything

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -665,9 +665,8 @@ void CTurbSolver::ComputeUnderRelaxationFactor(CSolver **solver_container, const
   /* Loop over the solution update given by relaxing the linear
    system for this nonlinear iteration. */
 
-  su2double localUnderRelaxation    =  1.00;
-  const su2double allowableDecrease = -0.99;
-  const su2double allowableIncrease =  0.99;
+  su2double localUnderRelaxation =  1.00;
+  const su2double allowableRatio =  0.99;
```
economon, line 669: Yep, just fyi, the original intent was to allow it to increase and decrease by separate percentages. For S-A, for instance, we often have very large increases at the start of the solve, but only very small decreases, so one may want to allow for a much larger percent increase but limit the decrease to a smaller percentage. 

As for swapping out the clipping in SST for something related to under-relaxation thresholds, it could be a good idea. Worth some testing and prob a follow up PR. As you have probably seen, we are only applying the under-relaxation to some S-A variants.. I played around a little with under-relaxation with SST, but I did not have time to experiment properly (and the results seemed very sensitive to it). I do not believe the SST articles in the literature have much to say about clipping. Actually, we had been clipping S-A unnecessarily before putting in the under-relaxation and CFL adaption, which we removed, so it would be good to experiment with SST as well.

PR #1059, opened 2020-08-04
Improve partitioning load balance for hybrid grids and add advanced ParMETIS controls

pcarruscag:
## Proposed Changes
See below.

## PR Checklist

### DIFF
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -2669,13 +2669,13 @@ void CConfig::SetConfig_Options() {
   addPythonOption("NZONES");
 
   /* DESCRIPTION: ParMETIS load balancing tolerance */
-  addDoubleOption("PARMETIS_TOLERANCE", ParMETIS_tolerance, 0.05);
+  addDoubleOption("PARMETIS_TOLERANCE", ParMETIS_tolerance, 0.02);
 
   /* DESCRIPTION: ParMETIS load balancing weight for points */
-  addLongOption("PARMETIS_POINT_WEIGHT", ParMETIS_pointWgt, 1);
+  addLongOption("PARMETIS_POINT_WEIGHT", ParMETIS_pointWgt, 0);
 
   /* DESCRIPTION: ParMETIS load balancing weight for edges (equiv. to neighbors) */
-  addLongOption("PARMETIS_EDGE_WEIGHT", ParMETIS_edgeWgt, 0);
+  addLongOption("PARMETIS_EDGE_WEIGHT", ParMETIS_edgeWgt, 1);
 
   /*--- options that are used in the Hybrid RANS/LES Simulations  ---*/
   /*!\par CONFIG_CATEGORY:Hybrid_RANSLES Options\ingroup Config*/
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -2668,8 +2668,14 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Number of zones of the problem */
   addPythonOption("NZONES");
 
-  /* DESCRIPTION: Activate ParMETIS mode for testing */
-  addBoolOption("PARMETIS", ParMETIS, false);
+  /* DESCRIPTION: ParMETIS load balancing tolerance */
+  addDoubleOption("PARMETIS_TOLERANCE", ParMETIS_tolerance, 0.05);
+
+  /* DESCRIPTION: ParMETIS load balancing weight for points */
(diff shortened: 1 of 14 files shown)


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -8414,98 +8352,88 @@ void CPhysicalGeometry::SetBoundTecPlot(char mesh_filename[MAX_STRING_SIZE], boo
...
+  }
 
-    for (int i = 0; i < size; i++) {
-      tpwgts[i] = 1.0/((real_t)size);
-    }
+  /*--- For most FVM-type operations the amount of work is proportional to the
+   * number of edges, for a few however it is proportional to the number of points.
+   * Therefore, for (static) load balancing we consider a weighted function of points
+   * and number of edges (or neighbors) per point, giving more importance to the latter
+   * skews the partitioner towards evenly distributing the total number of edges. ---*/
```
pcarruscag, line 8397: Abstract

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -8414,98 +8352,88 @@ void CPhysicalGeometry::SetBoundTecPlot(char mesh_filename[MAX_STRING_SIZE], boo
...
-      vtxdist[i+1] = (idx_t)pointPartitioner.GetLastIndexOnRank(i);
-    }
+  const auto wp = config->GetParMETIS_PointWeight();
+  const auto we = config->GetParMETIS_EdgeWeight();
 
-    /*--- Calling ParMETIS ---*/
+  vector<idx_t> vwgt(nPoint);
+  for (unsigned long iPoint = 0; iPoint < nPoint; ++iPoint) {
+    vwgt[iPoint] = wp + we * (xadj[iPoint+1] - xadj[iPoint]);
+  }
```
pcarruscag, line 8405: Weight function.

Common/src/CConfig.cpp
```diff
@@ -2668,8 +2668,14 @@ void CConfig::SetConfig_Options() {
...
-  /* DESCRIPTION: Activate ParMETIS mode for testing */
-  addBoolOption("PARMETIS", ParMETIS, false);
+  /* DESCRIPTION: ParMETIS load balancing tolerance */
+  addDoubleOption("PARMETIS_TOLERANCE", ParMETIS_tolerance, 0.05);
+
+  /* DESCRIPTION: ParMETIS load balancing weight for points */
+  addLongOption("PARMETIS_POINT_WEIGHT", ParMETIS_pointWgt, 1);
+
+  /* DESCRIPTION: ParMETIS load balancing weight for edges (equiv. to neighbors) */
+  addLongOption("PARMETIS_EDGE_WEIGHT", ParMETIS_edgeWgt, 0);
```
pcarruscag, line 2678: New options, as is this is equivalent to what is done at present (un-weighted partitioning).
I will get some numbers on a large hybrid mesh and I may update the defaults to what I think is more suitable for FVM, otherwise folks will have no idea this exists.

pcarruscag:
I changed the defaults to favor an even distribution of edges per MPI rank (instead of vertices).
This is broadly equivalent for structured-ish meshes, but it improves solution times by a factor of **1.5** on the medium CRM-HL hybrid grid (running on 192 cores).

Common/src/geometry/CPhysicalGeometry.cpp
economon: If you have time, here's one suggestion to try to see if there's any benefit/difference.. ParMETIS is set up to handle multiple constraints per vertex. An alternative is to set ncon = 2 and then supply 2 contiguous weight values for each point. Something like:
```
  unsigned long idx = 0;
  for (unsigned long iPoint = 0; iPoint < nPoint; ++iPoint) {
     vwgt[idx++] = wp;
     vwgt[idx++] = we * (xadj[iPoint+1] - xadj[iPoint]);
   }
```

pcarruscag:
Thank you for the review @economon.
I just gave the multiple constraints a try on the hybrid onera m6 mesh we have in testcases, it works, both points and edges can be balanced, but the edge cuts go up by almost 50% I guess that makes sense, more constraints less minimization.
There's another reason why I like the combined function, balancing the "num neighbors" metric is not exactly the same as balancing the number of edges per partition, using a small negative point weight (-1, -2) yields better results (but the ideal value will be case dependent).

I think the next big improvement we can make in this area is to use a partitioner that is aware of the network topology, ParMETIS assumes that the communication cost between any two ranks is the same, this simplification can be very costly on large parallel machines that have a tree-like network topology, and in older clusters with slow interconnects (relative to intra node communication).

Issue #1058, opened 2020-08-03
Dynamic Library linking and Object Database


FoamScience:
Hi everyone,

First of all, I want to thank you for this marvelous project you're running! Well done!
But I have some suggestions about the "building" and design philosophy of SU2

This (programming-related) "feature request" is all about making SU2 even more extendable

## Is my feature request related to a problem?
Maybe! It's not pleasant when I'm forced to link libraries statically. I think it hinders the potential of the C++ language and makes compilation time unnecessarily longer.

For example, in [SU2_CFD.cpp](https://github.com/su2code/SU2/blob/master/SU2_CFD/src/SU2_CFD.cpp#L117) the program decides what to do with an if statement and **all drivers need to be linked at compile-time**. What if I want to write my own driver? I need to include my headers in "SU2_CFD.hpp" and rebuild the whole thing again!!!

## The solution I would like
Why not use a [Factory With Self-Registering Classes](https://www.bfilipek.com/2018/02/factory-selfregister.html) instead. Ideally, the parent class (CDriver for example) would have
a static ::new() (or ::create() ) method which will create the appropriate type (child classes) based on **user input** (aka, from config file).
Of course, this way, the table of registered classes (drivers in my example) could be extended if the user requests to
dynamically load his own library (assuming that the library contains a child class that registers itself to the table).

There is another vital related issue: Incorporating SU2 code in user-developed code is the main concern of this feature request but we also need a mechanism for getting references to all created objects. In the shared separate library, the user expects to
use already created CVariables, ... etc; how can he get those? An object database that keeps references to most involved objects is needed.

**This roughly describes what I think is needed to be done. I'm willing to help and even lead the development of this feature if it gets enough approval**.

### Alternatives?
Can't think of anything else! This feature will enable devs to effortlessly extend the software and it even facilitates the building of User Defined Functions in the future as it'll become an important feature when the project gains enough popularity.

The only other option is to leave all this behind and continue developing following the current design until the project reaches a state where it will become very hard to go back and implement this feature.

**Additional context**
1. The design I described above doesn't take into account C++ templates as I don't see them used that frequently (Is that a thing?)

2. One last thing: Please, Please, we have to stop using raw pointers like that. I ran valgrind on the "Inviscid Bump in a Channel" case and discovered that 1Megs got definitely lost (leaked). Of course it will, if raw pointers get used in that way. I fully understand that many of the skilled programmers here come from C/Fortran background but please refrain from using raw pointers everywhere
[...]

PR #1062, opened 2020-08-05
Fully working self-registering CEulerSolver with CSolverFactory

FoamScience:
## Proposed Changes
A draft PR for creating self-registering classes with a factory.
The class (CEulerSolver implemented) that wants to register to a factory (CSolverFactory implemented) should:
- Inherit from DBMember template class (new, only handles automatic-registration).
- Provide a "GetName()" member method which returns the registered name as a string.
- Provide a "CreateMethod()" member with specific signature. This function should handle the full-creation of the object (For CEulerSolver, it makes a pointer and runs Preprocessing on it).

The factory class must provide a Register() method that accepts the to-be registered class's name and a pointer to its creation method with the specific signature mentioned above.

These design constraints are all enforced in DBMember but no checks where made for now.

Instead of meddling with member functions of CSolverFactory, I chose to create the test solver (EULER) directly in CDriver.

### Results so far
1. Solver gets instantiated successfully, the solver preprocessing phase passes with no problems
2. When the program gets to integration preprocessing, it fails with out_of_range exception. But that's to be expected as no solver metadata is tracked for the newly created solver yet :)

Again, testing on the "Inviscid bump in a channel" case.

## Related Work
This PR extends the work on #1058 

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CEulerSolver.hpp
+++ b/SU2_CFD/include/solvers/CEulerSolver.hpp
@@ -28,6 +28,7 @@
 #pragma once
 
 #include "CSolver.hpp"
+#include "CSolverFactory.hpp"
 #include "../variables/CEulerVariable.hpp"
 #include "../../../Common/include/omp_structure.hpp"
 
@@ -37,7 +38,7 @@
  * a child class for each particular solver (Euler, Navier-Stokes, etc.)
  * \author F. Palacios
  */
-class CEulerSolver : public CSolver {
+class CEulerSolver : public CSolver, public DBMember<CEulerSolver, CSolverFactory> {
 protected:
   enum : size_t {MAXNDIM = 3};    /*!< \brief Max number of space dimensions, used in some static arrays. */
   enum : size_t {MAXNVAR = 12};   /*!< \brief Max number of variables, used in some static arrays. */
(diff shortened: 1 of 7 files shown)


FoamScience:
To test the solver, I'm forcing **CSolverFactory::GetSolverMeta** in commit a3911b7 to return a new SolverMetaData object specifically suitable for use with EULER solver in MULTIGRID settings. It's a temporary action of course, but it gives us a chance to evaluate the validity of the self-registered solver.

**And it works**. At least, "Diffing" the history.csv of the Inviscid bump in a channel case (ran with both master and feature branches) results in no differences.

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -37,7 +38,7 @@
  * a child class for each particular solver (Euler, Navier-Stokes, etc.)
  * \author F. Palacios
  */
-class CEulerSolver : public CSolver {
+class CEulerSolver : public CSolver, public DBMember<CEulerSolver, CSolverFactory> {
```
pcarruscag, line 41: What happens to CNSSolver? Which is derived from CEulerSolver and presumably will also need to inherit from DBMember<CNSSolver,...>.
Would it not be simpler to have a softer contract that solvers need to register themselves with some factory?.

How does this help with your goal of dynamic linking? Will you hard-code "USER_SOLVER" options into the factories?

FoamScience: 1. Every single other solver has to inherit from DBMember and provide GetName() and CreateMethod() to be registered automatically.
2. Concerning CNSSolver, I understand multiple inheritance may become confusing but it's irrelevant. Commit 37f62a2 virtually only adds those requirements to CNSSolver, and voila, it gets registered:
  - The knownSolvers map will only have 2 elements in this case: "EULER" (as CEulerSolver) and "NAVIER_STOKES" (as CNSSolver). But CNSSolver indirectly inherits from "DBMember<CEulerSolver, CSolverFactory>" so why it doesn't get registered as a CEulerSolver instead? because that only redefines CEulerSolver::is_registered (The static variable) and has nothing to do with the new solver registration :)  (I tested NS solver with the laminar flat plate case).

3. This doesn't help with dynamic linking yet; my end goal would be to take advantage of Global and shared/static Variables (instantiated well before loading libraries):
  - Macros will be helpful; mainly, "DeclareClassAsFactory" would initialize some global shared variables and "AddClassToFactory" (for example) would rewrite some of those variables and instantiate new ones to register the class
to the associated factory. This will enable automatic registration even for RunTime-loaded classes;  If it interests you, I can craft a simplified version to demonstrate the idea.

Issue #992, opened 2020-05-19
Spalart-Allmaras Model Variants


economon:
**Describe the bug**
*A clear and concise description of what the bug is and what you expect the behavior to be instead. If applicable, add screenshots to help explain your problem.*

It was brought to our attention that what we currently define as the "Standard" SA model is not using the ft2 term in the source implementation:

https://github.com/su2code/SU2/blob/20888f488752e4e8aaa5230e56c5417923c7c344/SU2_CFD/src/numerics/turbulent/turb_sources.cpp#L137

which means it is actually the SA-noft2 variant:

https://turbmodels.larc.nasa.gov/spalart.html#sanoft2

In practice, this should make little difference for most problems, but it can be important for DES/DDES (where the ft2 term is typically avoided). However, it is also important that we be clear about the different variants of the turbulence models to users, since having confidence in the models for V&V, reproducibility, etc., is so critical.

I would vote that we reinstate the "standard" model and make a new SA variant for SA-noft2 as a separate option (without duplicating a ton of code, if possible). Anyone else feel strongly about this topic?

erangit:
May I suggest a different approach? Formally the text above is correct. However, the ft2 term was originally introduced to allow a transition to occur. Not so surprisingly it is not an accurate mechanism for transitional flow. To the best of my recollection, Philippe Spalart also acknowledged it, so during the years that followed 1992, the ft2-less SA model became actually the standard. We should remember that a newbie to the code is likely to opt for the standard model. I suggest leaving this one as is, add a comment to the documentation and add a model with ft2 option. While this doesn't match  the formal naming at the larc web page (which is an excellent resource for models versions) it is user friendly

suargi:
The original Spalart-Allmaras turbulence model encloses the ft2 term: https://www.researchgate.net/publication/236888804_A_One-Equation_Turbulence_Model_for_Aerodynamic_Flows or https://turbmodels.larc.nasa.gov/spalart.html#sa. After that, some modifications were introduced and the ft2 term was neglected. But accordingly, a different name is used, i.e., Spalart-Allmaras One-Equation Model without ft2 Term.
So, I would also vote for implementing the original model along with the SA-noft2 variant.

Also, I would like to point out a fact about the current implementation of the Negative Spalart-Allmaras variant.

[...]

Issue #1051, opened 2020-07-20
TECPLOT surface file writer


suargi:
**Describe the bug**
When setting OUTPUT_FILES= (SURFACE_TECPLOT) the output file is <name>.szplt. According to the code nomenclature, this file extension should be for the binary file: SU2_CFD/src/output/CTecplotBinaryFileWriter.cpp
When using the OUTPUT_FILES= (SURFACE_TECPLOT_BINARY) I get the following error:
OUTPUT_FILES: invalid option value SURFACE_TECPLOT_BINARY. Check current SU2 options in config_template.cfg

I think the nomenclature has changed w.r.t. previous versions but the template config file did not.

SURFACE_TECPLOT which should output a ".dat" file (see SU2_CFD/src/output/CTecplotFileWriter.cpp) is now (v7.0.6) SURFACE_TECPLOT_ASCII
SURFACE_TECPLOT_BINARY which should output a ".szplt" file (see SU2_CFD/src/output/CTecplotBinaryFileWriter.cpp) is now (v7.0.6) SURFACE_TECPLOT

**Correction**
Update the config file.

 - SU2 Version: v7.0.5 (also applicable to v7.0.6)

pcarruscag:
Looks like you tracked the origin of the issues, can you open a pull request with the fixes?

suargi:
I will do that

PR #1066, opened 2020-08-14
[WIP] SA turb model consistent implementation

suargi:
Implementation of the original SA and Negative SA turbulence models as well as their no-ft2 versions

## Proposed Changes
Implementation of the original version of Spalart-Allmaras turbulence model (https://www.researchgate.net/publication/236888804_A_One-Equation_Turbulence_Model_for_Aerodynamic_Flows).
Implementation of the Spalart-Allmaras turbulence model without the ft2 term.
Implementation of the original version of Negative Spalart-Allmaras turbulence model (https://www.iccfd.org/iccfd7/assets/pdf/papers/ICCFD7-1902_paper.pdf)
Implementation of the Negative Spalart-Allmaras turbulence model without the ft2 term neither modified vorticity.

On the config file, choose:
SA:                            Original SA model
SA_NOFT2:               SA model without ft2 term (the *same* model found in the develop branch)
SA_NEG:                   Original negative SA model
SA_NEG_NOFT2:     Negative SA model (the *same* model found in the develop branch)

*They are not the same as in the develop branch, because I corrected a minor bug.*

Updated the template config file to be in accordance with the source code: turbulent models and OUTPUT_FILES tags

## Related Work
This PR resolves issues #992 and #1051.

## Regression testing
Tested on tutorials:
- Turbulent Flat Plate
- RAE2822
- Turbulent ONERA M6

## PR Checklist


SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -123,7 +123,7 @@ class CSourceBase_TurbSA : public CNumerics {
  * \class CSourcePieceWise_TurbSA
  * \brief Class for integrating the source terms of the Spalart-Allmaras turbulence model equation.
  * \ingroup SourceDiscr
- * \author A. Bueno.
+ * \author G. Suarez
```
pcarruscag, line 126: That's a bit unfair...

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1932,7 +1934,7 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
         break;
       case SPACE_UPWIND :
         for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
-          if (spalart_allmaras || neg_spalart_allmaras || e_spalart_allmaras || comp_spalart_allmaras || e_comp_spalart_allmaras ) {
+          if (spalart_allmaras || spalart_allmaras_noft2 || neg_spalart_allmaras || neg_spalart_allmaras_noft2 || e_spalart_allmaras || comp_spalart_allmaras || e_comp_spalart_allmaras ) {
```
pcarruscag, line 1937: This way of handling the new options is a no go on my part, if we do it like this we will be chasing SA bugs for months, exactly like it happened when the solver options changed. Even with SST_SUST which was just one new option I fixed a bug last month or so.

I trust you to have changed all places in the code, but other people will not be aware (because very few people follow the pull requests) and unfortunately there are 10's of branches that do not track develop closely.

So, as a bare minimum this "is SA", "is SA-neg" logic needs to be encapsulated in a single method of CConfig, and replaced everywhere.
If you want to take it further we should create a small lightweight class that condenses all SA info, including coefficients.

Actually even less work, and probably more appropriate, create a boolean option called SA_USE_FT2= YES/NO, and avoid changing all the enum's and turbulence options.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -76,6 +76,179 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSA::ComputeResidual(const CConfig
 //  AD::SetPreaccIn(TurbVar_Grad_i[0], nDim);
 //  AD::SetPreaccIn(Volume); AD::SetPreaccIn(dist_i);
 
+//  BC Transition Model variables
+  su2double vmag, rey, re_theta, re_theta_t, re_v;
+  su2double tu , nu_cr, nu_t, nu_BC, chi_1, chi_2, term1, term2, term_exponential;
+
+  // Set the boolean here depending on whether the point is closest to a rough wall or not.
+  roughwall = (roughness_i > 0.0);
+
```
pcarruscag, line 85: For something called "no ft2" this looks like wayyyyy too much code (correct me if I'm wrong, I did not do a diff).
I know for these classes we operated on a "copy first, ask questions later" basis, but that has stopped. The duplication police has entered the building :)

build/meson-logs/install-log.txt
```diff
@@ -0,0 +1,72 @@
+# List of files installed by Meson
```
pcarruscag, line 1: You will have to manually delete all these build files and commit again.

config_template.cfg
```diff
@@ -1434,9 +1434,10 @@ SOLUTION_ADJ_FILENAME= solution_adj.dat
 TABULAR_FORMAT= TECPLOT
 %
 % Files to output 
-% Possible formats : (TECPLOT, TECPLOT_BINARY, SURFACE_TECPLOT,
-%  SURFACE_TECPLOT_BINARY, CSV, SURFACE_CSV, PARAVIEW, PARAVIEW_BINARY, SURFACE_PARAVIEW, 
-%  SURFACE_PARAVIEW_BINARY, MESH, RESTART_BINARY, RESTART_ASCII, CGNS, STL)
+% Possible formats : (TECPLOT, TECPLOT_ASCII, SURFACE_TECPLOT, SURFACE_TECPLOT_ASCII,
+%  CSV, SURFACE_CSV, PARAVIEW, PARAVIEW_ASCII, PARAVIEW_LEGACY, SURFACE_PARAVIEW,
+%  SURFACE_PARAVIEW_ASCII, SURFACE_PARAVIEW_LEGACY, PARAVIEW_MULTIBLOCK, RESTART,
+%  RESTART_ASCII, CGNS, STL, STL_BINARY, MESH)
```
pcarruscag, line 1440: :+1: Do we have a "MESH" option?

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
suargi: I wrote it to put on record who has modified the original version, just in case in a future someone finds any issue with that implementation. The same for CSourcePieceWise_TurbSA_Neg.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
suargi: It is a problem with the trace. At some moment I had to start from scratch so I deleted everything and copy pasted all code on turb_sources.cpp, and implemented the modifications.

build/meson-logs/install-log.txt
suargi: My bad.

config_template.cfg
suargi: https://github.com/su2code/SU2/blob/develop/Common/include/option_structure.hpp line 1631

SU2_CFD/src/drivers/CDriver.cpp
suargi: I totally agree with that. I would suggest to follow the comment from @clarkpede on issue #992 so we are able to include more than one variation.
Maybe on the config file we can include a tag for all the SA variations. Something like:

SA_VARIATIONS: NO_FT2, COMP, MOD_VORT, etc

which are basically Boolean to identify whether we want to use no ft2, comp, modified vorticity, etc.

Then, on turb_sources to have only one class, the original SA model. Therein, include all possible variations.

config_template.cfg
pcarruscag: Indeed but what matters are the strings from 1642 to 1659, the enums are for internal use.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Yes that sounds reasonable for the options and there is precedent with the QCR modification of the stress tensor.

However, having classes that look like Swiss army knives by means of if's and switch's is a bad software practice IMO.
There are "modern" and very efficient ways to specialize such small aspects, I'm using (and have described) some in #1022, we can discuss them in the next developers meeting if you want.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
pcarruscag: I got the idea, we don't have guidelines for this, but I only replace authors if I re-write the entire thing, when I make significant changes I add myself to the list, and for fixing bugs I just let git do the tracking of who modified which line.

SU2_CFD/src/drivers/CDriver.cpp
suargi: I will take a look to that!

config_template.cfg
suargi: Okay, but "MESH" is indeed an output option, see https://github.com/su2code/SU2/blob/develop/SU2_CFD/src/output/COutput.cpp line 388. Maybe the pairing in option_structure is missing?

pcarruscag: It is only used internally in SU2_DEF I think, it does not need to be selected by users.

suargi: Good. So shall be removed from the config_template.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -76,6 +76,179 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSA::ComputeResidual(const CConfig
...
+    inv_k2_d2 = 1.0/(k2*dist_i_2);
+
+    Shat = S + TurbVar_i[0]*fv2*inv_k2_d2;
+    Shat = max(Shat, 1.0e-10);
+    inv_Shat = 1.0/Shat;
+
+//    Original SA model
+//    Production = cb1*(1.0-ft2)*Shat*TurbVar_i[0]*Volume;
+
+    if (transition) {
```
koodlyakshay, line 152: A small comment. 
It looks like the BC transition model is only valid with the no-ft2 version (https://turbmodels.larc.nasa.gov/sa-bc_1eqn.html).

suargi: Thanks for pointing it out.
"the ft2 term present in (SA) may cause an undesirable delay in transition to turbulence in the RANS region, and should be avoided".
I will take it into account.

PR #1067, opened 2020-08-16
Fixes for incompressible solver - rotating frame and convergence rate for unsteady problems

cvencro:
## Proposed Changes

The draft pull request has changes to fix couple of issues in the incompressible solver for rotating frame and convergence rate of unsteady problems (both with and without grid movement).
 
For the rotating frame, there was a confusion between conservative and primitive variables being used for the numerics.

For the convergence rate of unsteady problems, the Jacobian contributions in the dual time residual calculations had been altered to include the preconditioner. However, this slows down the convergence from only an order of 100 iterations needed for converged solution per time step to the order of 1000 iterations. Therefore this has been reverted to regain the previous faster convergence rate and further work is necessary to understand the reason for this change.

## Related Work
PR #652 

## PR Checklist


SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1527,9 +1527,9 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
 
     for (iPoint = 0; iPoint < nPointDomain; iPoint++) {
 
-      /*--- Load the conservative variables ---*/
+      /*--- Load the primitive variables ---*/
 
-      numerics->SetConservative(nodes->GetSolution(iPoint), nullptr);
+      numerics->SetPrimitive(nodes->GetPrimitive(iPoint), nullptr);
```
economon, line 1532: nice catch 👍

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -3050,7 +3050,23 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
...
+        }
+        for (iDim = 0; iDim < nDim; iDim++)
+          Jacobian_i[iDim+1][iDim+1] = Density*Jacobian_i[iDim+1][iDim+1];
+        Jacobian_i[nDim+1][nDim+1] = Density*Cp*Jacobian_i[nDim+1][nDim+1];
+
+        Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
+      }
+      // ADD old version
+
+      if (implicit && false) {
```
economon, line 3069: For the Jacobian change, let's go with the data: sounds like the version without the preconditioner is an order of magnitude better in your tests, so I would make it the default here and remove the old implementation from the code (remember we can always recover to from the git history). 

It is likely that adding the preconditioner portion is hurting the diagonal dominance of the system, or we have a bug / logic error that would be good to understand once there is time to do more analysis

TobiKattmann: The version without Preconditioner is the "exact" Jacobian dR/dV imo. I recall that (artificially) multiplying with the preconditioner (Gamma*dR/dV) was s.th. we  / I / somebody tested because there is no Jacobian contribution for the continuity equation (iVar=1). 
So we could have an unlucky case where the artificial Jacobian is hurting convergence compared to the exact one?
I looked in the Paper of Smith&Weiss for additional info without much success :)

pcarruscag: Gonna stick my nose in :) You guys can also consider an approach similar to the associated with CENTRAL_JACOBIAN_FIX_FACTOR, i.e. multiply this Jacobian contribution with a weight on the off chance some problem might benefit from it.

cvencro: Hi Pedro, I've added an option for the fix factor in centered numerics of the incompressible methods. The preconditioner acts as the multiplication factor already, but there is now an option for an additional fix factor. For a test case I'm currently using, factor below 1 has small improvements to convergence but not a significant step change so the default value for CENTRAL_INC_JACOBIAN_FIX_FACTOR in the config is 1.0. But like you say having this option will give more tweaking flexibility in the config in case it benefits some other problems.

pcarruscag: That was not what I meant... maybe I can explain better this afternoon.

pcarruscag:
@cvencro @TobiKattmann did you want to include anything else here or can it be "un drafted"?

cvencro:
Hi, there were no other changes from my side so I've removed the draft label.

If there are any other suggestions for improving the convergence (I think there is still an open discussion above), we can either act on it here or just make a note of it for future work.

pcarruscag (approved):
LGTM then, having two options with similar names is not ideal but is the default value is different I don't see another way.

pcarruscag:
One regression will likely have to be re-updated.

TobiKattmann:
> 
> 
> One regression will likely have to be re-updated.

Yep, I'll update the CHT unsteady case once the checks completed.

PR #1070, opened 2020-09-03
OptimalPropeller Python Script

EttoreSaetta:
## Proposed Changes
The OptimalPropeller script provides the optimal load distribution along the propeller radius using only global parameters based on the inviscid theory of optimal propeller developed by Glauert. It also provides the actuator disk data file needed in input for the "ACTDISK_TYPE = VARIABLE_LOAD".
This script is useful in case the user would like to use this actuator disk model, but the propeller details are not available.
In this case, using global parameters (thrust coefficient, radius, hub radius, advance ratio, free-stream velocity and blades number), it allows the use of the model.
Input: interactive.
Output: ActuatorDisk.cfg, containing part of the SU2 .cfg file; ActuatorDisk.dat, containing the propeller load distribution to be read by SU2_CFD.

Contributors:
Theoretical and Applied Aerodynamic Research Group (TAARG) of University of Naples Federico II.
EttoreSaetta, lorenzorusso07, rtogna.

## Related Work
#1027 

## PR Checklist


SU2_PY/OptimalPropeller.py
```diff
@@ -0,0 +1,371 @@
+# ==============================================================================================
```
pcarruscag, line 1: Put the SU2 license header above this one (it is a GPL requirement).

SU2_PY/OptimalPropeller.py
```diff
@@ -0,0 +1,371 @@
...
+###        Main        ###
+##########################
+# Screen output
+print('------------------ Optimal Propeller vsn 1.0.0 ------------------')
+print('| Computation of the optimal dCT/dr and dCP/dr distributions.   |')
+print('| Based on the inviscid theory of the optimal propeller.        |')
+print('|                                                               |')
+print('| This code is used to generate the actuator disk input data    |')
+print('| file needed for the VARIABLE_LOAD actuator disk type          |')
+print('| implemented in SU2 7.0.7.                                     |')
```
pcarruscag, line 73: I think the version of any component of SU2 should simply be the version of SU2.

SU2_PY/OptimalPropeller.py
```diff
@@ -0,0 +1,371 @@
...
+
+# Number of radial stations in input.
+Stations = int(input('Number of radial stations: '))
+print('')
+
+dStations = float(Stations)
+
+# Resize the vectors using the number of radial stations.
+r = [None]*Stations
+chi = [None]*Stations
```
pcarruscag, line 92: Any reason why these variables are python lists and not numpy arrays?

pcarruscag (commented):
Looks good Ettore, few small comments below.

SU2_PY/OptimalPropeller.py
EttoreSaetta: Thank you for your comment, I have added the license header as on the other .py files.

SU2_PY/OptimalPropeller.py
EttoreSaetta: Version changed. I put 7.0.7 because it should be in the next release. Is it good or should I write 7.0.6?

SU2_PY/OptimalPropeller.py
EttoreSaetta: I am sorry, but it is the first time I code in Python. I have changed these variables and now they are numpy arrays.
Thanks for this comment!

SU2_PY/OptimalPropeller.py
pcarruscag: I don't know if the next release will be 7.0.7 or 7.1.0.
Make it 7.0.6 please, there is a script that will update all 7.0.6's to the correct version before the release.

pcarruscag (approved):
LGTM but give it a couple days in case anyone else has comments.

SU2_PY/OptimalPropeller.py
EttoreSaetta: ok, I didn't know about it. Version changed to 7.0.6.

EttoreSaetta:
> LGTM but give it a couple days in case anyone else has comments.

Thank you so much for your review @pcarruscag.
I will merge it in a couple days.

SU2_PY/OptimalPropeller.py
pcarruscag: No problem, that was a genuine question, as maybe it was a requirement of some of the functions you are using.
But generally it is better to use numpy arrays for plain numbers since under the hood they are one block of memory.

PR #1071, opened 2020-09-03
Cleanup native ASCII file writers

pcarruscag:
## Proposed Changes
I started having some issues with the meshes written by SU2_DEF over night (due to some change in the cluster file system) so I cleaned up a couple of the file writers to hopefully make them more robust.
I think what we were doing (all ranks opening the file at the same time and flushing it after each write) was a bit iffy because of implementation dependent buffering aspects.

## PR Checklist


economon (approved):
Nice cleanup - open/close looks safer than flush, and we're not too worried about performance in these ASCII writers anyway (since they were already serialized which kills the performance from the start).

👍

pcarruscag:
All these ********** stale bots, codefactors and what ever are just absolutely stupid and useless.

pcarruscag:
Thanks @economon, I think the bullet proof option would be to use MPI IO, but this will have to do for now.

PR #1077, opened 2020-09-21
Fix memory issues

jblueh:
## Proposed Changes
This PR contains fixes for two memory issues.

- The first issue regards AD builds. The reinterpret cast of the dynamically allocated char array to an `su2double` array prevents `su2double` default constructor calls. In that case, the memory must be zero initialized in order for the AD index management to work correctly.
- Regarding the second issue, invalid reads occur when the pointers to the local arrays are used later in `CConfig::SetDefaults()` when the arrays are already out of scope.

## PR Checklist


pcarruscag (approved):
LGTM :+1:

PR #1078, opened 2020-09-23
NEMO ONLY (not develop): Update some structures within NEMO, more in line with develop

WallyMaier:
THIS IS A PR TO feature_NEMO.  No impact on develop

This pull requests adds preprocessing to the NEMO NS solver and commonpreprocessing functionality seen in develop.  Some functions have been extended to deal with eddy viscosity.  Regression tests are passing.

If the community finds having internal pull request distracting/cluttering I can take down.


SU2_CFD/include/numerics/NEMO/CNEMONumerics.hpp
```diff
@@ -70,6 +71,18 @@ class CNEMONumerics : public CNumerics {
...
                 const CConfig* config);
 
+  /*!
+   * \brief Constructor of the class, for turbulence applications.
+   * \param[in] val_nDim - Number of dimensions of the problem.
+   * \param[in] val_nVar - Number of variables of the problem.
+   * \param[in] config - Definition of the particular problem.
+   */
+  CNEMONumerics(unsigned short val_nDim, unsigned short val_nVar,
+                unsigned short val_nVar_NEMO,
```
CatarinaGarbacz, line 81: I think we don't need val_nVar_NEMO. val_nVar is a generic variable that stores the number of conservative variables for a given solver. I don't think there is the need to make it specific for NEMO and the way it is now, it seems simply like a repetition of val_nVar

CatarinaGarbacz: since we're here, we could also update the functions arguments in the doxy description of the function, to include val_nPrimVar, val_nPrimVarGrad

SU2_CFD/include/solvers/CNEMONSSolver.hpp
```diff
@@ -43,12 +43,22 @@
...
 
+  /*!
+   * \brief Compute the velocity^2, SoundSpeed, Pressure, Enthalpy, Viscosity.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] Output - boolean to determine whether to print output.
+   * \return - The number of non-physical points.
+   */
+  unsigned long SetPrimitive_Variables(CSolver **solver_container,
+                                       CConfig *config, bool Output);
```
CatarinaGarbacz, line 61: It seems that this function is now repeated from NEMOEulerSolver to NEMONSSolver. check out what is done about this in the vanilla solvers, and follow that implementation (I think they use virtual for EulerSolver but best to check that carefully)

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -549,4 +549,14 @@ class CNEMOEulerVariable : public CVariable {
...
+  /*!
+   * \brief Retrieves the value of the laminar viscosity in the primitive variable vector.
+   */
+  inline unsigned short GetLamViscIndex(void) { return LAM_VISC_INDEX; }
+
+  /*!
+   * \brief Retrieves the value of the eddy viscosity in the primitive variable vector.
+   */
+  inline unsigned short GetEddyViscIndex(void) { return EDDY_VISC_INDEX; }
+
```
CatarinaGarbacz, line 561: would it be possible to add this instead in the CNEMONSVariable for conceptual consistency?

SU2_CFD/src/numerics/NEMO/CNEMONumerics.cpp
```diff
@@ -71,6 +73,51 @@ CNEMONumerics::CNEMONumerics(unsigned short val_nDim, unsigned short val_nVar,
     }
 }
 
+CNEMONumerics::CNEMONumerics(unsigned short val_nDim, unsigned short val_nVar,
+                             unsigned short val_nVar_NEMO,
+                             unsigned short val_nPrimVar,
```
CatarinaGarbacz, line 78: again, I think var_nVar_NEMO is unnecessary and a repetition

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -90,6 +93,7 @@ CNEMOEulerSolver::CNEMOEulerSolver(CGeometry *geometry, CConfig *config,
 
   }
 
+  //TODO: we dont need this in NEMO delete me
```
CatarinaGarbacz, line 96: go ahead then :P

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -107,9 +111,12 @@ CNEMOEulerSolver::CNEMOEulerSolver(CGeometry *geometry, CConfig *config,
   //     U: [rho1, ..., rhoNs, rhou, rhov, rhow, rhoe, rhoeve]^T
   //     V: [rho1, ..., rhoNs, T, Tve, u, v, w, P, rho, h, a, rhoCvtr, rhoCvve]^T
   // GradV: [rho1, ..., rhoNs, T, Tve, u, v, w, P, rho, h, a, rhoCvtr, rhoCvve]^T
+  // Viscous: append [mu, mu_t]^T
   nVar         = nSpecies + nDim + 2;
-  nPrimVar     = nSpecies + nDim + 8;
+  if (navier_stokes) { nPrimVar   = nSpecies + nDim + 10; }
+  else {               nPrimVar   = nSpecies +nDim +8;    }
   nPrimVarGrad = nSpecies + nDim + 8;
+  //todo, need to clean up?
```
CatarinaGarbacz, line 119: clean up what exactly ? pls be specific in the comments since they're for everyone to understand :)

CatarinaGarbacz (commented):
@WallyMaier I gave a careful look to this PR.

First of all, I think having a concise/small-ish contribution for review is really great, makes the reviewer job incredibly easier (and therefore with more attentive reviews), than a PR with lots of big changes. So I say we keep having this approach

I think the implementation is in a good place to be merged. It definitely makes NEMO stronger with regard to being even more up to date with develop, when it comes to pre-processing. the changes that concern turbulence also seem to be implemented in a consistent and economic way.

I just made a few comments, to correct small details that I believe are not so ok. I would just ask you to address these (or perhaps justify your choices), and once this is done, I will give approval in github and we can go ahead with the merge.

SU2_CFD/include/numerics/NEMO/CNEMONumerics.hpp
WallyMaier: Good catch, Ill update the Doxy stuff.    The reason there is a second CNEMONumerics is for use in the instantiation of the NEMO turb solver, I can remove for this PR (Maybe we can chat regarding a better way to handle the issue I am having when the time comes)

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Hahah I wasnt sure, if we had any use, but Ill go ahead and delete

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: I wasnt sure if there was a "prettier" way to code this up.

SU2_CFD/include/numerics/NEMO/CNEMONumerics.hpp
CatarinaGarbacz: ok, I didn't understand this was a new constructor, I thought it was replacing the old one. but still, if val_nVar = val_nVar_NEMO, what is the reasoning here ? if there's a logical explanation, of course we can keep it, just trying to understand

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
CatarinaGarbacz: it looks good to me :)

SU2_CFD/include/numerics/NEMO/CNEMONumerics.hpp
WallyMaier: So this constructor will be instantiated with turbulence solver where nVar = 1 (or 2).  Using the current numerics structure, the number of variable vs. vector size would be mismatched. The turbulence solution with be of size 1 or 2, but the nVar loop would be 6 or 7.  But we still need the index information which to do some computations.  So instead of trying to pass that information every time an index is needed, Instead pass the number of turbulence vars (val_nVar) and the number of Nemo vars.  I hope that makes sense

That being said, there may be way easier/efficient to do this.

CatarinaGarbacz: that makes total sense @WallyMaier . now that I understand better it actually sounds like a good and simple approach.

SU2_CFD/include/solvers/CNEMONSSolver.hpp
WallyMaier: You are right, the have a virtual in the CEulerSolver.hpp.  CNEMOEulerSolver.hpp has been updated appropriately.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Done

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
WallyMaier: Done

CatarinaGarbacz (approved):

PR #1080, opened 2020-09-30
Fix develop compilation on some compilers

pcarruscag:


vdweide (approved):
Looks like a no brainer to me

Issue #823, opened 2019-11-24
Hardcoded paths in build scripts


dkavolis:
**Describe the bug**
* https://github.com/su2code/SU2/blob/15f0258a7607245be601b72fa2c158f79f01c922/SU2_PY/pySU2/meson.build#L5-L15
You should not assume that mpi4py is installed in python system directory, instead use `import mpi4py; print(mpi4py.get_include())` to get the required path directly.

* https://github.com/su2code/SU2/blob/15f0258a7607245be601b72fa2c158f79f01c922/meson.build#L114-L118
You should not assume that MKL is installed in a single master directory even though it might be true in most cases. For example, installing MKL from Ubuntu PPA will not work with SU2 since headers are installed in `/usr/bin/include/mkl` and libraries - `/usr/lib/x86_64-linux-gnu` though it does provide `pkgconfig` files.

**Desktop (please complete the following information):**
 - OS: any
 - C++ compiler and version: any
 - MPI implementation and version: any
 - SU2 Version: develop

pcarruscag:
Hi @dkavolis,
It seems like you know your stuff about build systems and associated best practices, it would be great if you could open a PR with the fixes for this, I'd be happy to test the MKL stuff on the systems I have access to.
Cheers,
Pedro

PR #1082, opened 2020-10-02
Fix #1081

### LINKED ISSUE #1081: Latest version of develop does not build in debug mode
**Describe the bug**
The latest version of develop gives undefined references when building in debug mode, e.g.
undefined reference to `CRoeBase<CRoeScheme<CCompressibleViscousFlux<2ul> >, CCompressibleViscousFlux<2ul> >::nVar'

**To Reproduce**
Build the debug version with meson using --buildtype=debug

 - OS: Ubuntu 18:04
 - C++ compiler and version: g++ 8.4.0
 - MPI implementation and version: OpenMPI version 2.1
 - SU2 Version: [e.g., v7.0.6]

pcarruscag:


vdweide (changes requested):
It works for me now in combination with both the GNU and Intel compilers. However, for Intel I get a ton of warnings the type 

warning #2196: routine is both "inline" and "noinline"

I'm not sure if these are only caused by your changes, but this is not what we want, I suppose.

Also I get the following warning with Intel
 ./Common/src/CConfig.cpp(4666): warning #823: reference is to variable "iMarker" (declared at line 3174) -- under old for-init scoping rules it would have been variable "iMarker" (declared at line 4662)

This has come in an earlier PR, but can be fixed easily by removing the unsigned short in line 4662.

Finally, when I build the AD version the Intel compiler gives the following warnings

warning #16219: Some OpenMP processing was skipped to constrain compile time. Consider overriding limits (-qoverride-limits).

./SU2_CFD/include/numerics_simd/flow/convection/centered.hpp(182): warning #15552: loop was not vectorized with "simd"

Adding -qoverride-limits does increase compile and link time significantly, but it does not solve the problem. These warnings seem to be caused by the template classes in the following two files.

SU2_CFD/include/numerics_simd/flow/convection/roe.hpp
SU2_CFD/include/numerics_simd/flow/convection/centered.hpp

I don't know whether you have vectorization in mind in combination for the AD version, but then it may be useful to split up these files, if possible.

I do not have access to a Mac, so I didn't test if on OSX.

pcarruscag:
My proposal the other day seems to have put you in quite the mood...
Do you also get a ton of "-1 converted to unsigned short"? Or is that just me?
I'll see what I can do about the rest.

> I do not have access to a Mac, so I didn't test if on OSX.

I tried clang 6 and clang 10, and had to stupefy the code because of the former, should be ok on Mac

vdweide:
I feel a bit guilty indeed regarding the review of the PR's.

I didn't see the -1 conversion warning, but I may have missed them because of all the other warnings. Do you get these from the Intel compiler as well?

pcarruscag:
Exclusively from the Intel compiler, I think, they've been around for a while now, hence my not caring much about Intel warnings.

pcarruscag:
Ok that's all I can do. There are so many warnings on AD builds with the Intel compiler (and also with gcc-10) that I don't see the point in wasting more time.

vdweide (approved):
I don't get any warnings GNU version 10 and also for intel the number of warnings are significantly less now.

PR #1091, opened 2020-10-12
Fix or silence a few Intel compilation warnings

pcarruscag:
And some other cleanup

### DIFF
--- a/Common/include/geometry/primal_grid/CPrimalGrid.hpp
+++ b/Common/include/geometry/primal_grid/CPrimalGrid.hpp
@@ -240,7 +240,7 @@ public:
   * \brief A virtual member.
   * \return The color of the element in the partitioning.
   */
-  inline virtual unsigned long GetColor(void) { return -1; }
+  inline virtual unsigned long GetColor(void) { return std::numeric_limits<unsigned long>::max(); }
 
   /*!
    * \brief Get the element global index in a parallel computation.
--- a/Common/src/geometry/CPhysicalGeometry.cpp
+++ b/Common/src/geometry/CPhysicalGeometry.cpp
@@ -3790,83 +3790,60 @@ void CPhysicalGeometry::LoadLinearlyPartitionedVolumeElements(CConfig        *co
 
   /*--- Get the volume connectivity from the mesh object. ---*/
 
-  const vector<unsigned long> &connElems =
-  mesh->GetLocalVolumeElementConnectivity();
+  auto& connElems = mesh->GetLocalVolumeElementConnectivity();
 
   /*--- Allocate space for the CGNS interior elements in our SU2 data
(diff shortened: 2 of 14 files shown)


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -3913,22 +3880,18 @@ void CPhysicalGeometry::LoadLinearlyPartitionedVolumeElements(CConfig        *co
     }
   }
 
-
   /*--- Reduce the global counts of all element types found in
    the CGNS grid with all ranks. ---*/
```
vdweide, line 3885: I may be a bit picky here, but why are multiple calls to Allreduce used here? I know this happens in quite a few places in the code, but I have never understood why this is done instead of a single call to Allreduce with an array.

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -240,7 +240,7 @@ class CPrimalGrid {
   * \brief A virtual member.
   * \return The color of the element in the partitioning.
   */
```
vdweide, line 242: Thanks for fixing these. Weird that earlier versions of the compiler did not pick up this one.

vdweide (approved):
Thanks for fixing all these warnings. Looks good to me, except maybe for the several calls to Allreduce instead of a single one in CPhysicalGeometry.cpp. But that happens in quite a few places in the code and I probably shouldn't be too picky.

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: I imagine whoever did it like that was thinking of readability of having multiple descriptive name vs one generic array.
We could have special types for that kind of data to strike a compromise, but since this particular case is not in the hot path of the code I would leave like that.

vdweide: Agree

Common/include/geometry/primal_grid/CPrimalGrid.hpp
pcarruscag: I think it is due to warning level rather than version. It does not show up at level 0, level 1+ is still too verbose on AD builds.

pcarruscag:
@vdweide compilation with level 2 warnings (-Wall -Wextra) is no longer so verbose that it is not useful.
Most of the remaining warnings are of the "not vectorized with simd" variety and they come (I suppose) from CSysVector.hpp:302, I think I'll leave them to avoid having to create an exclusion for Intel only.

If you want to fix the GCC 10 memcpy memset warnings feel free to use this PR, otherwise it is good to go on my side.

vdweide:
Good for me as well.

PR #1095, opened 2020-10-20
Axisymmetric source viscous terms

FlorianDm:
I added the viscous terms to the axisymmetric source but I am not sure it is all correct. I get very reasonable, well converged results with my cases but dissipation seems a little excessive, however I cannot say for sure because there are many other factors in my own cases. There is no test case I believe. Should I try to create one with air like this one https://turbmodels.larc.nasa.gov/axibump_val.html ?

How come in the incompressible version the turbulent heat transfer is ignored? I added it here with the Reynolds analogy as eddyvisc * cp/PrTurb * dT/dy and it seems significant in my cases (with PrTurb = 0.9).

Also, I created a generalised version of the axisymmetric source in a separate class which goes with the generalised Roe and AvgGradFlow solvers with any eos. I use it for my cases with a table based fluid model (not included here) but I am not sure it is needed for the existing fluid models because they all use a constant gamma anyway. Perhaps it is because "Pressure_i = (Gamma-1.0)*U_i[0]*(U_i[nDim+1]/U_i[0]-0.5*sq_vel)" may not be true for the "polytropic Van der Waals" and "polytropic Peng Robinson" models. If that is the case the Driver should instantiate the general one for anything other than incompressible, ideal gas and standard air.


WallyMaier:
Hi Florian, I was also working on the same thing (feature_axi).  I was unable to find a suitable case to test, and it seems we went about it in different ways.  Im happy you got something working!  Could you pass along the literature you used to create these terms?

bigfooted:
Hi Florian, great work, and important to have. The transonic bump case seems to have enough data available for proper validation. 
What are you unsure about? The actual derived source terms? If you have a document with your derivation of the source terms that would help in checking if it is correct.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1865,6 +1865,8 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
       else if (config->GetAxisymmetric() == YES) {
         if (incompressible)
           numerics[iMGlevel][FLOW_SOL][source_first_term] = new CSourceIncAxisymmetric_Flow(nDim, nVar_Flow, config);
+        if (config->GetKind_FluidModel() == 0)
+          numerics[iMGlevel][FLOW_SOL][source_first_term] = new CSourceGeneralAxisymmetric_Flow(nDim, nVar_Flow, config);
```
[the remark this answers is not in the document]
pcarruscag: Does the general (non ideal?) implementation only work for standard air? (I did not think our standard air was non ideal)
Don't compare to 0, use the appropriate enum values.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3120,6 +3120,17 @@ void CEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_contain
       /*--- Set y coordinate ---*/
       numerics->SetCoord(geometry->nodes->GetCoord(iPoint), geometry->nodes->GetCoord(iPoint));
 
+      /*--- Set primitive variables ---*/
+      numerics->SetPrimitive(nodes->GetPrimitive(iPoint), nodes->GetPrimitive(iPoint));
+
+      /*--- Set gradient of primitive variables ---*/
+      numerics->SetPrimVarGradient(nodes->GetGradient_Primitive(iPoint), nodes->GetGradient_Primitive(iPoint));
+
+      if (config->GetKind_FluidModel() == 0) {
```
pcarruscag, line 3129: again comparison with 0

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -105,6 +108,134 @@ CNumerics::ResidualType<> CSourceAxisymmetric_Flow::ComputeResidual(const CConfi
...
+
+      residual[0] -= 0.0;
+      residual[1] -= yinv*Volume*total_viscosity_i*(PrimVar_Grad_i[1][1]+PrimVar_Grad_i[2][1]
+                                                    - TWO3*PrimVar_Grad_i[2][0]);
+      residual[2] -= yinv*Volume*total_viscosity_i*(2*(PrimVar_Grad_i[2][1]-Velocity2_i*yinv)
+                                                    - TWO3*PrimVar_Grad_i[2][1]-FOUR3*Velocity2_i*yinv);
+      residual[3] -= yinv*Volume*(total_viscosity_i*(Velocity1_i*(PrimVar_Grad_i[1][1]
+                                                                  + ONE3*PrimVar_Grad_i[2][0])
+                                                                  - FOUR3*Velocity2_i*PrimVar_Grad_i[1][0]) 
+                                  + total_conductivity_i*PrimVar_Grad_i[0][1]);
```
pcarruscag, line 236: this part looks the same as for the ideal gas source, please avoid this duplication.
For example, implement this section as a method of CSourceAxisymmetric_Flow, then make the general class derived from it so it can re use the code.

pcarruscag (commented):
Thanks Florian, looks tidy, just a few comments.
Do you have a comparison between 3D and axisymmetric?
If you have time, add a regression test, it would also serve as an example for when we get questions on CFD online.

SU2_CFD/src/drivers/CDriver.cpp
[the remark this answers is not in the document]
FlorianDm: Oh yes the 0 was a placeholder but a bad idea. I made it default to the general version now for anything but ideal gas and standard air. I think that is how it should be since the non-ideal models use the generalised solvers for fluxes too

SU2_CFD/src/numerics/flow/flow_sources.cpp
FlorianDm: Okay I will see how i can do that

FlorianDm:
Hi everyone, glad this subject is of interest.

I will try to make that test case.

https://vtechworks.lib.vt.edu/handle/10919/29772, in this document on page 227 is the source term in vector notation. From there I derived as you can see in the notes but I was surprised that it ended up so simple so maybe I made some silly mistakes somewhere. In the energy equation all the v*dv/dy terms cancelled out. I wonder what is the physical meaning of that.

IMG_20201021_143238506

For the generalised inviscid part I am pretty sure it is all correct including the jacobian. You can compare with very similar terms in any generalised flux jacobian like in Glaister's paper https://www.sciencedirect.com/science/article/pii/002199918890174X

FlorianDm:
I will do my case in 3D to compare

SU2_CFD/src/drivers/CDriver.cpp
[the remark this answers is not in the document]
FlorianDm: It is completely general and works with any eos such as my table interpolation. Which btw works very well but before I can propose to integrate that I want to create a user interface in the config file to provide the data in ascii files. Standard air is of course ideal gas.

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -124,6 +131,105 @@ CNumerics::ResidualType<> CSourceAxisymmetric_Flow::ComputeResidual(const CConfi
   return ResidualType<>(residual, jacobian, nullptr);
 }
 
+void CSourceAxisymmetric_Flow::ResidualDiffusion(){
+  
+  su2double laminar_viscosity_i    = V_i[nDim+5];
```
pcarruscag, line 136: Nicely done :+1:

FlorianDm:
In this current form there is too much dissipation in most situations in my RANS simulations with k-omega-SST (in some situations it appears there is not enough, probably related to a low magnitude of du/dy). I don't have time to try any pure Navier-Stokes cases.

The source terms (convection and diffusion) for k and omega should also be implemented I suppose which should be very much analogous so I think I could try to do that. 

Otherwise I don't see what could be wrong with the flow source terms. Please let me know if you do. 

I have a couple of questions and I would really appreciate if anybody could give any input.

1. What happens when a user specifies several sources? Will only the last one be retained since they are always assigned to the same numerics[iMGlevel][FLOW_SOL][source_first_term]? This matters most for the turbulent source since there is always already another source.

2. Why is the axisymmetric source term for the incompressible variable density version what it is (the publication related to this does not give any derivation)? If it is correct, something about my derivation must be wrong. My derivation : 
    - take the equations in cylindrical coordinates (convection term as well as the tau definition, see "Bird - Transport Processes" Annex B if needed)
    - set any d/dtheta and velocity_theta (ignoring swirl) to zero 
    - rearrange into two parts, one which resembles the cartesian 2D form (r=y, z=x) and the remaining part is the source term to be added. 
- Is the above derivation wrong?

pcarruscag:
Hi Florian,
You can have a look in CDriver.cpp around line 1850, the logic is mutually exclusive, so if multiple sources are specified the order does not matter, the way things are coded determine which source is used. (I hope somewhere in CConfig we check for conflicting sources).
Note that the first term / second term is per solver, so there is actually a lot of room for different sources.

Regarding 2 there are more knowledgeable people about those topics out there than me. Maybe you can join the developers meeting tomorrow at 3pm CET (https://meet.jit.si/SU2_DevMeeting) and you can ask those questions to e.g. @economon or @TobiKattmann

FlorianDm:
Okay thank you Pedro (I meant the order in the code not in the config file). So I would just assign any k-omega axi source to numerics[iMGlevel][TURB_SOL][source_second_term].

See you tomorrow for the meeting

WallyMaier:
@FlorianDm When I was working on the feature_axi to attempt what you have done, I unsed this reference, which should match the incompressible version exactly.
[AxiSymm_1.pdf](https://github.com/su2code/SU2/files/5482565/AxiSymm_1.pdf)
[AxiSymm_2.pdf](https://github.com/su2code/SU2/files/5482567/AxiSymm_2.pdf)

@jtneedels Has ran/is running a laminar viscous wedge with your formulation of axisymmetric and 3D to do some verification.   We have also tried to run the NASA TMR axisymmetric Validation cases, but they are proving to be quite difficult.  Hope this is useful to the discussion.

FlorianDm:
@WallyMaier oh great I think this book was cited in the publication but I couldn't get access to it

pcarruscag:
@FlorianDm sorry I got my timezones mixed up, the meeting usual time is 4pm CET (3pm where I am)

TobiKattmann:
Unfortunately I am not too much of a help in this, but I hope @WallyMaier 's reference is helpful. I will follow this PR though to see if I can provide s.th. in the process :)

vdweide:
Maybe this helps a bit

2D_Axi_Part1

2D_Axi_Part2

vdweide:
Although, this seems to be exactly the same reference as @WallyMaier uploaded already....

FlorianDm:
Hi, thank you all a lot for your help. Here is my derivation : 
[SourceAxi.pdf](https://github.com/su2code/SU2/files/5506823/SourceAxi.pdf)

Mistakes have been corrected and it should be the same as your book now apart from the fact that I am neglecting the gradient of viscosity. To take it into account it needs to be evaluated and I suppose that is what those AuxVar variables are for (something like u *mu or u *mu/y I guess). This is new to me and unfortunately I do not have time to look into it atm (unless someone points me in the right direction and it's quickly done)

@WallyMaier I believe that you had taken that into account already and I just came along and cut out your work. So I should close this PR and once you open a new one I can add the generalisation for the convection terms. Were you planning on implementing the k and omega sources as well? 

What book is that btw?

WallyMaier:
@FlorianDm, I had no plans to implement the k-omega terms.  I dont think you need to close this PR, I am very happy you took this on!   I can add some of the changes I have to this branch.  The branch feature_axi has been giving us spurious results and needs some debugging.

As for the book, maybe @vdweide or @economon may know....I only have those two pages haha.

bigfooted:
@FlorianDm @WallyMaier 

I have a different version, but it looks like this is from the book of Hoffmann and Chiang, Computational Fluid Dynamics, volumeII

FlorianDm:
Aah thanks so it was cited but without the volume so I looked in volume I not knowing there was a volume II

economon:
I'm late to the party here, but just a note to say that the original implementation for the incompressible source terms are indeed from the text that @WallyMaier / @vdweide shared. It was added as part of the work in this paper (https://economon.github.io/docs/AIAA-2018-3111.pdf), but I did not test it much beyond a simple laminar channel case or really attempt to treat turbulence at the time. Thanks for putting in more effort on these terms!

WallyMaier:
@FlorianDm @economon @everyone else

I have merged Florian's derivation with my earlier derivation.  I have added the gradients with respect to viscosity.   @jtneedels is running some tests on it currently.  I can push those changes to this branch so everyone can look over them...what do you all think?

FlorianDm:
@WallyMaier fantastic! Yes please do push them to this branch

FlorianDm:
Actually I had a look at your branch and the way it is right now is not correct I believe because I had taken the other variables v and 1/y out of the derivative using the chain rule and combined them with the other derivatives to end up with the terms as they are now so only d(mu)/dy was missing. The AxiAuxVarGrad you are using is apparently d(v*mu/y)/dy so the other terms have to be different. I will change them.

So is it not possible to get simply the velocity gradient? Is there any reason not to pull the other variables out?

Is there an AxiAuxVar and an AuxVar with AuxVar being just v*mu?  

Don't we need to do a SetAuxVarGrad or something like that?

FlorianDm:
To do it with your AxiAuxVar i would suggest this: 

  ```
  residual[0] -= 0.0;
  residual[1] -= Volume*(yinv*total_viscosity_i*(PrimVar_Grad_i[1][1]+PrimVar_Grad_i[2][0]) 
                        - TWO3*AxiAuxVar_Grad_i[0][0]);
  residual[2] -= Volume*(yinv*total_viscosity_i*2*(PrimVar_Grad_i[2][1]-v*yinv)
                        - TWO3*AxiAuxVar_Grad_i[0][1]);
  residual[3] -= Volume*(yinv*(total_viscosity_i*(u*(PrimVar_Grad_i[2][0]+PrimVar_Grad_i[1][1])
                                                + v*(FOUR3*PrimVar_Grad_i[1][1]+TWO3*PrimVar_Grad_i[1][0])
                                                - TWO3*v*v*yinv)
                              - total_conductivity_i*PrimVar_Grad_i[0][1])
                        - TWO3*(AxiAuxVar_Grad_i[1][1]+AxiAuxVar_Grad_i[2][1]));
```

Could you put that into your branch and then push to this branch? I double checked the signs.. should be fine. If I commit it here I will break the compilation because these things are not defined 

And for the setters maybe combine the `if` guards so it's cleaner
```
      /*--- Set primitive variables for viscous terms and/or generalised source ---*/
      if (!ideal_gas || viscous) numerics->SetPrimitive(nodes->GetPrimitive(iPoint), nodes->GetPrimitive(iPoint));

      /*--- Set secondary variables for generalised source ---*/
      if (!ideal_gas) numerics->SetSecondary(nodes->GetSecondary(iPoint), nodes->GetSecondary(iPoint));

      if (viscous) {
        
        /*--- Set gradient of primitive variables ---*/
        numerics->SetPrimVarGradient(nodes->GetGradient_Primitive(iPoint), nodes->GetGradient_Primitive(iPoint));
      
        /*--- Set gradient of auxillary variables ---*/
        numerics->SetAxiAuxVarGrad(nodes->GetAxiAuxVarGradient(iPoint), nullptr);
      }
```

FlorianDm:
In fact i just created a new branch from yours: feature_axi_general_viscous where I made the changes so I will do a new PR from there ok?

WallyMaier:
@FlorianDm Thanks for doing all that! Unfortunately time zone differences prevented my reply haha.

FlorianDm:
Yes I was a bit impatient there. Unfortunately the results didn't change much at all for a case with turbulence. Still too much dissipation

PR #1097, opened 2020-10-23
Fix rough wall model in SST turbulence model

EttoreSaetta:
## Proposed Changes
We found a possible mistake in the tau wall computation in the rough wall boundary condition for the SST model.
In particular, in the current version, the tau wall is considered to be equal to the skin friction coefficient.
We changed it considering that
ImmaginiSU2-2.
We found this possible problem while doing a run on a NACA0012 airfoil.
The problem found is that the skin friction coefficient along the airfoil is too high compared to the theory.
The grid has 257 nodes on the airfoil, and it has been downloaded from the NASA website (https://turbmodels.larc.nasa.gov/naca0012_grids.html).
The solved equations are the RANS with the SST turbulence model, in compressible regime. M=0.1, Re=1e6, TU=10%, mut/mu=0.001
The run is dimensional.
ImmaginiSU2-1
In this figure "Rough Wall 1" is the case of tau wall computed with the current version.
"Rough Wall 2" is the case of tau wall computed after this modification.
The chosen k+ is 4 because, considering the Wilcox paper "Formulation of the k-w Turbulence Model Revisited, AIAA Journal, Vol. 46, No. 11, November 2008", the smooth wall has k+ < 5.
Moreover, the order of magnitude of omega wall is O(1e9). In case of the  "Rough Wall 1" it is of O(1e6), while in case of "Rough Wall 2" it is of O(1e9).
We don't know if the problem is somewhere else, but changing the computation of the tau wall it seems to be fixed.
 We hope this could be useful.

Authors: TAARG (Theoretical and Applied Aerodynamic Research Group), University of Naples Federico II.
Ettore Saetta, Lorenzo Russo, Renato Tognaccini, Benedetto Mele.

## Related Work
#877 

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -411,12 +411,13 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
         /*--- Set wall values ---*/
         su2double density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
         su2double laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
-
-        su2double WallShearStress = 0.0;
+        const su2double *U = config->GetVelocity_FreeStream();
+	su2double WallShearStress;
+        su2double SkinFrictionMag = 0.0;
         for (auto iDim = 0; iDim < nDim; iDim++)
-          WallShearStress += pow(solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, iDim),2.0);
+          SkinFrictionMag += pow(solver_container[FLOW_SOL]->GetCSkinFriction(val_marker, iVertex, iDim),2.0);
 
-        WallShearStress = sqrt(WallShearStress);
+        WallShearStress = sqrt(SkinFrictionMag)*0.5*density*pow(U/config->GetVelocity_Ref(), 2);
         /*--- Compute non-dimensional velocity ---*/
         su2double FrictionVel = sqrt(fabs(WallShearStress)/density);
 
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -412,7 +412,7 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
         su2double density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
         su2double laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
         const su2double *U = config->GetVelocity_FreeStream();
-	su2double WallShearStress;
+        su2double WallShearStress;
         su2double SkinFrictionMag = 0.0;
         for (auto iDim = 0; iDim < nDim; iDim++)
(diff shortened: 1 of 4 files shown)


pcarruscag (commented):
Thank you @EttoreSaetta, this fix makes sense to me, looking at how the skin friction is computed, but I'm no expert on turbulence models.

One thing I notice is that the reference velocity used to compute Cf is not always the same, see `CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces` in CFVMFlowSolverBase.inl.
And so for some problems the shear stress will still be incorrect.

Perhaps a better way to fix this issue is to store the WallShearStress computed in the flow solver and use it directly in the turbulence solvers when needed (i.e. `GetWallShearStress` instead of `GetCSkinFriction`).

Can you attend the developers meeting next week? (Wednesday 4pm CET) We can ask folks who know more about the turb models what they think.

koodlyakshay:
My bad. Thanks for noticing and fixing this @EttoreSaetta, but as Pedro said maybe there is more to it. I will also attend the meet next week and hopefully we can fix this.

EttoreSaetta:
> Thank you @EttoreSaetta, this fix makes sense to me, looking at how the skin friction is computed, but I'm no expert on turbulence models.
> 
> One thing I notice is that the reference velocity used to compute Cf is not always the same, see `CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces` in CFVMFlowSolverBase.inl.
> And so for some problems the shear stress will still be incorrect.
> 
> Perhaps a better way to fix this issue is to store the WallShearStress computed in the flow solver and use it directly in the turbulence solvers when needed (i.e. `GetWallShearStress` instead of `GetCSkinFriction`).
> 
> Can you attend the developers meeting next week? (Wednesday 4pm CET) We can ask folks who know more about the turb models what they think.

Thank you @pcarruscag for your reply.
I have seen `CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces`. Thanks also for your useful suggestion of storing the WallShearStress computed to be used directly in the turbulence solvers.
For sure I can attend to the developers meeting to talk also about this topic. I just would like to know how I can join it.
Thanks again.

Best,
Ettore

EttoreSaetta:
> My bad. Thanks for noticing and fixing this @EttoreSaetta, but as Pedro said maybe there is more to it. I will also attend the meet next week and hopefully we can fix this.

Thanks @koodlyakshay, see you at the meeting.
Best,
Ettore

pcarruscag:
This link should work https://meet.jit.si/SU2_DevMeeting

EttoreSaetta:
> This link should work https://meet.jit.si/SU2_DevMeeting

Thanks a lot, see you at the meeting.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -411,12 +411,8 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
         /*--- Set wall values ---*/
         su2double density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
         su2double laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
+        su2double WallShearStress = GetWallShearStress(val_marker, iVertex);
```
pcarruscag, line 414: The method needs to be called on the flow solver (`solver_container[FLOW_SOL]->`).

pcarruscag (approved):
:+1: Just one small comment then I think we can merge.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
EttoreSaetta: Fixed. Actually I was just doing a debug and I found that there was a problem.
Thanks for reporting it!

Issue #1000, opened 2020-05-23
SU2 will not run if built after MPICH


aditya12398:
So I built SU2 after addressing all the problems in #991, it worked just fine with openmpi. However, I would prefer to install it with mpich since I use it for different other programs as well. The built was successful but SU2 does not run and fails with the following error. I tried it with mpich v3.3.1 as well (my previous SU2 v6.2 was running on this)

```
slimshady@arch-linux-hp-probook-g3-450: SU2HOME$ SU2_CFD
[arch-linux-hp-probook-g3-450:590584] *** Process received signal ***
[arch-linux-hp-probook-g3-450:590584] Signal: Segmentation fault (11)
[arch-linux-hp-probook-g3-450:590584] Signal code: Address not mapped (1)
[arch-linux-hp-probook-g3-450:590584] Failing at address: 0x440000e8
[arch-linux-hp-probook-g3-450:590584] [ 0] /usr/lib/libc.so.6(+0x3bd70)[0x7f6d73811d70]
[arch-linux-hp-probook-g3-450:590584] [ 1] /usr/lib/openmpi/libmpi.so.40(PMPI_Comm_rank+0x37)[0x7f6d75627837]
[arch-linux-hp-probook-g3-450:590584] [ 2] SU2_CFD(+0x1023dc)[0x563dd30cd3dc]
[arch-linux-hp-probook-g3-450:590584] [ 3] /usr/lib/libc.so.6(__libc_start_main+0xf3)[0x7f6d737fd023]
[arch-linux-hp-probook-g3-450:590584] [ 4] SU2_CFD(+0x118bbe)[0x563dd30e3bbe]
[arch-linux-hp-probook-g3-450:590584] *** End of error message ***
Segmentation fault (core dumped)
```

**Desktop (please complete the following information):**
 - OS: Arch Linux x86_64
 - C++ compiler and version: g++, gcc 9.2.1
 - MPI implementation and version: OpenMPI v4.0.2; MPICH v3.3.2
 - SU2 Version: v7.0.4
 - Python Version: 3.8.1

This is my output of `mpirun --version`, if that is useful:
```
HYDRA build details:
    Version:                                 3.3.2
    Release Date:                            Tue Nov 12 21:23:16 CST 2019
    CC:                              gcc    
    CXX:                             g++    
    F77:                             gfortran   
    F90:                             gfortran   
[...]

Issue #991, opened 2020-05-18
SU2 v7.0.4 build fails while running ninja


aditya12398:
The ninja build command fails with the following error: 
```
slimshady@arch-linux-hp-probook-g3-450: SU2$ ./ninja -j4 -C build install
ninja: Entering directory `build'
[568/870] Linking target SU2_CFD/src/SU2_CFD.
FAILED: SU2_CFD/src/SU2_CFD 
[...]

PR #1102, opened 2020-11-05
Add an option to meson for extra dependencies (to fix mpich and lapack issues)

pcarruscag:
## Proposed Changes
Add option `-Dextra-deps=a,b,c` to meson to provide a simpler work around for build problems with MPICH and LAPACK.
For example if you have MPICH but it is not detected (because its pkg config name is not "mpi" but "mpich" instead), you would do `-Dcustom-mpi=true -Dextra-deps=mpich`.
If you have linker errors because the blas library does not contain LAPACK functions, you would do `-Dextra-deps=lapack`.
The option takes a comma-separated list.

I will update the documentation once this goes through.

## Related Work
#991
#1000

## PR Checklist


vdweide:
Looks good to me, but it would be nice to have the examples you mention above somewhere in the documentation. From the description in meson_options.txt I would not be able to figure it out.

pcarruscag:
Thanks Edwin, the plan is to update the build instructions on the website before the release.

pcarruscag:
@CatarinaGarbacz this should fix the compilation issues with Eigen

vdweide (approved):
Looks all good to me then.

PR #1104, opened 2020-11-10
Mesh Quality metrics output

TobiKattmann:
## Proposed Changes
Hi, I noticed that the mesh quality metrics are not available in the current output structure, so here they are. 
I removed WRT_MESH_QUALITY and just check&set in the configs postprocessing routine.
Compressible, Incompressible, Heat and Mesh Output are added here as the FEM solvers dont have the dual grid as stated in the original PR. 

**Bonus issue:** I noticed that Coordinates are not written in the Volume Output for the cases I tested (also the Quickstart) but haven't really figured out why yet. Doesn't matter whether VOLUME_OUTPUT is set with COORDINATES or not at all. I will look into this again but if you have a hint that would be appreciated💐 **Edit:** of course they are written implicitly otherwise on couldn't open a visu file but what do we have the volume output field COORDINATES for then? Maybe I am just blind, but this is not really part of this PR anyway.

Cheers, Tobi

## Related Work
Initial implementation #778 

## PR Checklist

### DIFF
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -2038,8 +2038,6 @@ void CConfig::SetConfig_Options() {
   addBoolOption("WRT_PERFORMANCE", Wrt_Performance, false);
   /* DESCRIPTION: Output the tape statistics (discrete adjoint)  \ingroup Config*/
   addBoolOption("WRT_AD_STATISTICS", Wrt_AD_Statistics, false);
-  /* DESCRIPTION: Write the mesh quality metrics to the visualization files.  \ingroup Config*/
-  addBoolOption("WRT_MESH_QUALITY", Wrt_MeshQuality, false);
     /* DESCRIPTION: Output a 1D slice of a 2D cartesian solution \ingroup Config*/
   addBoolOption("WRT_SLICE", Wrt_Slice, false);
   /*!\brief MARKER_ANALYZE_AVERAGE
@@ -3215,6 +3213,13 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
     }
   }
 
+  /*--- Check if MESH_QUALITY is requested in VOLUME_OUTPUT and set the config boolean accordingly. ---*/
+  Wrt_MeshQuality = false;
+  for (unsigned short iField = 0; iField < nVolumeOutput; iField++) {
+    if(VolumeOutput[iField].find("MESH_QUALITY") != string::npos) {
+      Wrt_MeshQuality = true;
+    }
+  }
 
   if (Kind_Solver == NAVIER_STOKES && Kind_Turb_Model != NONE){
     SU2_MPI::Error("KIND_TURB_MODEL must be NONE if SOLVER= NAVIER_STOKES", CURRENT_FUNCTION);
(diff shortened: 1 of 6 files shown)


SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -414,6 +414,11 @@ void CFlowCompOutput::SetVolumeOutputFields(CConfig *config){
     AddVolumeOutput("Q_CRITERION", "Q_Criterion", "VORTEX_IDENTIFICATION", "Value of the Q-Criterion");
   }
 
+  // Mesh quality metrics, computed in CPhysicalGeometry::ComputeMeshQualityStatistics.
+  AddVolumeOutput("ORTHOGONALITY", "Orthogonality", "MESH_QUALITY", "Orthogonality Angle (deg.)");
+  AddVolumeOutput("ASPECT_RATIO",  "Aspect_Ratio",  "MESH_QUALITY", "CV Face Area Aspect Ratio");
+  AddVolumeOutput("VOLUME_RATIO",  "Volume_Ratio",  "MESH_QUALITY", "CV Sub-Volume Ratio");
+
```
pcarruscag, line 421: Seems all logical to me, is there a common place to share these options across the different outputs?

Common/src/CConfig.cpp
```diff
@@ -2038,8 +2038,6 @@ void CConfig::SetConfig_Options() {
   addBoolOption("WRT_PERFORMANCE", Wrt_Performance, false);
   /* DESCRIPTION: Output the tape statistics (discrete adjoint)  \ingroup Config*/
   addBoolOption("WRT_AD_STATISTICS", Wrt_AD_Statistics, false);
-  /* DESCRIPTION: Write the mesh quality metrics to the visualization files.  \ingroup Config*/
-  addBoolOption("WRT_MESH_QUALITY", Wrt_MeshQuality, false);
```
pcarruscag, line 2042: There is a place in CConfig.cpp to print messages about deprecated options, look for RELAXATION_FACTOR_ADJFLOW, it would be good to explain the disappearance of the option.

SU2_CFD/src/output/CFlowCompOutput.cpp
TobiKattmann: Not that I am aware of. Even `COORD-X` etc is specifically mentioned in every output class.

Common/src/CConfig.cpp
```diff
@@ -2893,9 +2893,12 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
...
           if (!option_name.compare("RELAXATION_FACTOR_ADJFLOW"))
             newString.append("Option RELAXATION_FACTOR_ADJFLOW is now RELAXATION_FACTOR_ADJOINT, "
-                             "and it also applies to discrete adjoint problems\n.");
+                             "and it also applies to discrete adjoint problems.\n\n");
+          if (!option_name.compare("WRT_MESH_QUALITY"))
+            newString.append("WRT_MESH_QUALITY is deprecated. Use VOLUME_OUTPUT= (MESH_QUALITY, ...) instead.\n\n");
           errorString.append(newString);
           err_count++;
+          line_count++;
         continue;
```
TobiKattmann, line 2902: Due to the `continue`, `line_count` would be off by plus one for every error after the first one as the other `line_count` increment is down below.

SU2_CFD/src/output/CFlowCompOutput.cpp
pcarruscag: Ok, I guess we can leave that for the modules @talbring wants to implement.

pcarruscag (approved):

bigfooted:
Hi Tobi, Thanks for adding this (again), I will probably use this every time I do a mesh deformation!

PR #1105, opened 2020-11-13
Nan checks and better discrete adjoint default options

pcarruscag:
## Proposed Changes
Check for Nan in the monitored values of COutput (was already done for time domain).
For SU2_CFD_AD and SU2_DOT_AD the default MATH_PROBLEM becomes DISCRETE_ADJOINT
When the MATH_PROBLEM is DISCRETE_ADJOINT the linear solver and preconditioner of primal solvers is made the same as the DISCADJ_** ones, i.e. LINEAR_SOLVER and LINEAR_SOLVER_PREC are ignored (see why I think it makes sense in #942, and let me know if you think it doesn't make sense).

## Related Work
#942 

## PR Checklist


oleburghardt (approved):
Looks good to me, give it a go on Wednesday.

I've seen some nans in forward AD mode recently (for `maxHF`), maybe this helps to not always ignore them..

Common/src/CConfig.cpp
```diff
@@ -4856,6 +4861,10 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
     }
 #endif
 
+    /*--- Use the same linear solver on the primal as the one used in the adjoint. ---*/
+    Kind_Linear_Solver = Kind_DiscAdj_Linear_Solver;
+    Kind_Linear_Solver_Prec = Kind_DiscAdj_Linear_Prec;
```
TobiKattmann, line 4866: Here again that would be nice to have documented in the config_template or CConfig.

Tbh I wasn't aware that so far I have always used FGMRES+ILU in DA computations by not setting `DISCADJ_LIN_*`. I thought that the `LINEAR_SOLVER` stuff is used for the evalutaion of the linear solver during tape evaluation as well. And `grep`pring the Testcase folder showed me that only fea or fsi cases used this option.  Or am I missing something here?

But I see your point pedro in #942 and together with the `MATH_PROBLEM` which can be deletd from the `.cfg` these cases one can use the exact same config which is nice. I guess it doesn't make too much sense to use a different Linsol configuration during the taping compared to tape evaluation ... but I might ask about that on wednesday during the dev-meeting again :)

TobiKattmann (approved):

pcarruscag:
Thank you both.
@oleburghardt this will only stop the calculations for values used for monitoring, e.g. you would need a cauchy criteria on maxHF.

PR #1106, opened 2020-11-16
Axisymmetric problems :  Viscous source terms and generalised convective source terms and jacobian 

FlorianDm:
Axisymmetric problems :  Viscous source terms and generalised convective source terms and jacobian 
See https://github.com/su2code/SU2/pull/1095


SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -624,6 +627,45 @@ class CVariable {
...
+   */
+  inline void SetAxiAuxVar(unsigned long iPoint, const su2double *axiauxvar) {
+    for (unsigned long iVar = 0; iVar < 3; iVar++) AxiAuxVar(iPoint,iVar) = axiauxvar[iVar];
+  }
+
+  /*!
+   * \brief Get the entire axi aux vector of the problem.
+   * \return Reference to the axi aux matrix.
+   */
+  inline const MatrixType& GetAxiAuxVar(void) { return AxiAuxVar; }
```
pcarruscag, line 643: `const` between ) and {

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1955,6 +1997,12 @@ class CVariable {
    */
   inline virtual CVectorOfMatrix& GetGradient_Reconstruction(void) { return Gradient; }
 
+  /*!
+   * \brief A virtual member.
+   * \return Value of the primitive variables gradient.
+   */
+  inline virtual su2double **GetAxiAuxVarGradient(unsigned long iPoint) { return nullptr; }
+
```
pcarruscag, line 2005: Unless I'm missing something this does not have to be virtual?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1650,6 +1650,10 @@ void CSolver::GetCommCountAndType(const CConfig* config,
       COUNT_PER_POINT  = nDim;
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
+    case AXIAUXVAR_GRADIENT:
+      COUNT_PER_POINT  = nDim*3;
```
pcarruscag, line 1654: The 3 is a little bit magic, I would define it somewhere central like CVariable.hpp as a constexpr variable.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -80,6 +80,9 @@ class CVariable {
   VectorType AuxVar;       /*!< \brief Auxiliar variable for gradient computation. */
   MatrixType Grad_AuxVar;  /*!< \brief Gradient of the auxiliar variable. */
 
+  MatrixType AxiAuxVar;             /*!< \brief Axisymmetric auxiliar variable for gradient computation. */
+  CVectorOfMatrix Grad_AxiAuxVar;   /*!< \brief Gradient of the axisymmetric variables  of the problem. */
+
```
pcarruscag, line 85: I wonder if it would make sense to generalize AuxVar to have any number of variables, instead of adding one more "Aux thing". Do you see any conflict with how they are used?

WallyMaier: There shouldnt be any conflict...I think this is a good idea.

WallyMaier:
@FlorianDM,  I was looking through another reference that @jtneedels passed to me. (http://eprints.gla.ac.uk/184394/1/184394.pdf)

If you are still struggling with the extension to RANS: On page 56, the stress terms are altered with a couple of terms.

SU2_CFD/include/variables/CVariable.hpp
WallyMaier: Done!

SU2_CFD/include/variables/CVariable.hpp
WallyMaier: This was changed an inline only.   The "final" version in CEulersovler was removed.

FlorianDm:
@WallyMaier thank you. Interesting...

1. This reference reminded me to add the contribution of the turbulence  kinetic energy (+2/3 v rho k) in the energy equation source (see page 10 equation 2.30, pdf page 37), so that's done.

2. This reference will be helpful If I get to do the axysimmetry source terms for k and omega (by comparing equations 2.31 and 2.31 to equations B.2 and B.3) but that will be in a separate PR. In fact I am getting good results for RANS already (although no time for a thorough validation)

3. In that reference they have also done what I did before and pulled out the primitive derivatives and neglected the viscosity gradient and then came up with the same axisymmetry momentum source terms so that is comforting. However, their energy source terms are different and that made me think. What we have now is according to the first reference, so it should be fine unless there is an error in it. I find it suspicious that both aux derivatives are wrt y and none wrt x given the that we have one wrt x and one wrt y in the momentum equation... Could it be `AxiAuxVar_Grad_i[2][0]` instead of `AxiAuxVar_Grad_i[2][1]` ? The impact on results is almost unnoticeable for my case. Otherwise we are done here I think.

@pcarruscag thanks for taking the time to review!

SU2_CFD/src/solvers/CSolver.cpp
WallyMaier: Im not sure I follow here

SU2_CFD/include/variables/CVariable.hpp
WallyMaier: After looking more deeply into the code, other portions of the code use the AuxVar functons. But these are 
unrelated to the axisymmetric terms.  (CElasticity, CAdjEuler).  I think a generalization would be possible.

After there any examples of this type of generalization that I can work with?

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: Ideally this kind of info should be centralized somewhere, typically in option_structure.hpp.
e.g. const unsigned short NAXIVARS=3; or something that follows similar names we might have.
Then you use that variable in the code instead of hardcoding 3 in lots of places (3 can be confused with 3D, and if you ever need more variables...)

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: I don't think there are, but it should be just a matter of replacing the uses of AuxVar by AxiAuxVar, fixing the compilation errors, then deleting the methods dealing with AuxVar, and finally renaming AxiAuxVar by just AuxVar :) or something like that.
If it gets too messy I can help.

SU2_CFD/src/solvers/CSolver.cpp
WallyMaier: Ah i gotcha, that makes sense!

SU2_CFD/include/variables/CVariable.hpp
WallyMaier: @pcarruscag Thanks for the review and help!  I have spending some time trying to generalize the functions, but I am running into 2 main issues (I think its only 2).  The main issue moving properly handing the change from vectors to matrix of vectors.   If you had time, I would very much appreciate your help :)   

I haven't pushed the current code, since its hacked together at the moment.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -564,65 +570,62 @@ class CVariable {
...
+  inline const MatrixType& GetAuxVar(void) const { return AuxVar; }
 
   /*!
-   * \brief Get the value of the auxiliary variable.
-   * \param[in] iPoint - Point index.
-   * \return Value of the auxiliary variable.
+   * \brief Get the Aux var value at Point i, variable j.
    */
-  inline su2double GetAuxVar(unsigned long iPoint) const { return AuxVar(iPoint); }
+  inline su2double GetAuxVar(unsigned long iPoint, unsigned long iVar = 0) const { return AuxVar(iPoint,iVar); }
```
WallyMaier, line 581: Nice!

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -3107,6 +3110,34 @@ void CEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_contain
...
+      for (iPoint = 0; iPoint < nPoint; iPoint++) {
+
+        su2double yCoord          = geometry->nodes->GetCoord(iPoint, 1);
+        su2double yVelocity       = nodes->GetVelocity(iPoint,1);
+        su2double xVelocity       = nodes->GetVelocity(iPoint,0);
+        su2double Total_Viscosity = nodes->GetLaminarViscosity(iPoint) + nodes->GetEddyViscosity(iPoint);
+
+        if (yCoord > EPS){
+          su2double nu_v_on_y = Total_Viscosity*yVelocity/yCoord;
+          nodes->SetAuxVar(iPoint, 0, nu_v_on_y);
```
WallyMaier, line 3125: This is much cleaner...thanks!

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1647,7 +1647,7 @@ void CSolver::GetCommCountAndType(const CConfig* config,
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
     case AUXVAR_GRADIENT:
-      COUNT_PER_POINT  = nDim;
+      COUNT_PER_POINT  = nDim*base_nodes->GetnAuxVar();
```
WallyMaier, line 1650: I know see how to do this properly 👍

jtneedels:
@WallyMaier and I ran some verification cases for axisymmetric verification (see attached slides). Results show good agreement between planar axisymmetric and 3D dimensional results.
[AxisymmetricVerification.pptx](https://github.com/su2code/SU2/files/5655639/AxisymmetricVerification.pptx)

WallyMaier:
Thanks @jtneedels for doing some testing.    I think we saw some issues with the aerodynamic coefficient computations...but the pressure distributions looked good.  This looks good to me.  Ill leave it to @FlorianDm or @pcarruscag to make any final changes or suggestions.

Good work y'all!

WallyMaier (approved):

aeroamit (commented):
@WallyMaier & @FlorianDm , I repeated the heat flux problem, I have posted earlier and seems like issue still persist (I showed earlier that this is not the problem with 2D case with symmetry BC but pops in with axisymmetric case). I am attaching the files here for reference ....
[hflux_issue_axi_branch.pdf](https://github.com/su2code/SU2/files/5660684/hflux_issue_axi_branch.pdf)
[heatflux_issue_old.pdf](https://github.com/su2code/SU2/files/5660685/heatflux_issue_old.pdf)
[case_files.zip](https://github.com/su2code/SU2/files/5660688/case_files.zip)

WallyMaier:
@aeroamit, interesting...Im think this could possibly be related to the computation of heatflux itself (with regards to the axisymmetric formulation).  Judging from @jtneedels results, the Cp distributions look good.  It could be the scaling factor applied to the heatflux?

Ill take a look

aeroamit:
I have done some comparison cases for Cp distribution with axisymmetric formulation with older versions, they were matching well with commercial codes (I must say for the specific configurations I simulated it somehow matched). This issue came into light while trying to obtain heat flux with axisymmetric formulation.
I will also try to run this axisymmetric case with some other code and compare....

bigfooted:
@aeroamit  What does temperature do? Could you compute the heat flux from the temperature in e.g. paraview? Is the root cause the computation of the energy equation or the computation of heat flux?

pcarruscag (approved):

WallyMaier:
@aeroamit, I ran the cases you had to take a look.  While the residuals were "converged", the solution was not.  There was a massive low temperature region along the body that took many more iterations to clean.  The maximum Heat Flux was still ~6 w/m**2

aeroamit:
> @aeroamit What does temperature do? Could you compute the heat flux from the temperature in e.g. paraview? Is the root cause the computation of the energy equation or the computation of heat flux?

Hi @bigfooted, Temperature boundary condition for wall corresponds to Isothermal wall BC  (cold wall condition). This condition is applied to obtain heat flux unlike adiabatic wall (no heat transfer). 
The snapshot, I posted  from ParaView shows heat flux variation with x. This is obtained directly from surface_flow.vtu. You can simply go to Filters -> Data Analysis -> Plot data and select Points_X for X Array and heat flux in variable. Regarding computing heat flux from temperature field, I am not sure, but ParaView is having calculator utility as well as option to compute gradient of unstructured grid (from there you can obtain temperature gradients in 3 directions). 
Coming to last question, root cause of the problem - @WallyMaier has run the case yesterday, we will be posting some details soon.
Best 
Amit

aeroamit (approved):
Hi @FlorianDm and @WallyMaier, I tried this branch and ran hemisphere (M=5) case as well as one more case. I found pressure distribution to be perfectly all right and matching well with alternate code. I did not observe any issue in the stagnation region for Cp distribution. We are good to go. Thanks for the fixes.. good job..

I am just attaching a file from heat flux computation issue  case...
[SU2_hflux_issue.pdf](https://github.com/su2code/SU2/files/5685369/SU2_hflux_issue.pdf)

PR #1107, opened 2020-11-17
Heat solver fixes for primal and adjoint CHT simulations

pcarruscag:
## Proposed Changes
I think one sign was flipped in the Jacobian of the CHT boundary.
I replaced the isothermal BC for a stronger(?) alternative (flying by the seat of my pants here, if it's wrong let me know, but the solver converges better like this).
Fixes issues with deforming heat domains by generalising CDiscAdjMeshSolver::SetSensitivity to accept any solver as "target" (instead of considering always the ADJFLOW_SOL position in the solver container).

## Related Work
#951 
#1061 
(third time lucky I hope...)

## PR Checklist


SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -1143,6 +1098,12 @@ void CHeatSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **solv
 
           HeatFluxDensity   = thermal_diffusivity*(Tinterface - Tnormal_Conjugate);
           HeatFlux          = HeatFluxDensity * Area;
+
+          if (implicit) {
+
+            Jacobian_i[0][0] = -thermal_diffusivity*Area;
+            Jacobian.SubtractBlock2Diag(iPoint, Jacobian_i);
+          }
```
pcarruscag, line 1106: The sign, I also moved this so that it only applies to one of the coupling modes, to be consistent with the implementation of the heat flux wall.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -1642,7 +1597,9 @@ void CHeatSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_
 
   /*--- Solve or smooth the linear system ---*/
 
-  System.Solve(Jacobian, LinSysRes, LinSysSol, geometry, config);
+  auto iter = System.Solve(Jacobian, LinSysRes, LinSysSol, geometry, config);
+  SetIterLinSolver(iter);
+  SetResLinSolver(System.GetResidual());
```
pcarruscag, line 1602: Ah yes, it also fixes the output of the linear solver stuff.

pcarruscag:
@TobiKattmann since we were talking about it yesterday, the advantage of the "weak" formulation is that it is conservative, heat flux in equal heat flux out (because the post processed fluxes are consistent with how the boundary condition).
Obviously at the expense of the isothermal temperatures not being imposed exactly.
With the "strong" formulation, conservation can probably only be achieved if the mesh topology of the inlet/outlet boundary is the same, or at grid convergence.
Learning something new everyday, I put it back as it was.

WallyMaier (approved):
@pcarruscag Thanks for the contribution!

I looked the files changes and it looks good to me!

SU2_CFD/src/solvers/CDiscAdjMeshSolver.cpp
```diff
@@ -212,30 +210,28 @@ void CDiscAdjMeshSolver::SetSensitivity(CGeometry *geometry, CSolver **solver, C
   ExtractAdjoint_Variables(geometry, config);
 
   /*--- Store the sensitivities in the flow adjoint container ---*/
-  for (iPoint = 0; iPoint < nPoint; iPoint++) {
+  for (auto iPoint = 0u; iPoint < nPoint; iPoint++) {
```
TobiKattmann, line 213: Deleted datatype above is unsigned long, should this maybe be a `0ul` therefore instead of `0u` ? `GetSharpEdge_Distance` also takes an ul. One might end up with uint32_t either way but that is not the case on my machines where ul is uint64_t..but I guess alone 4billion points in one process is probably for the next generation of CPU's :D

Common/include/CConfig.hpp
```diff
@@ -5135,21 +5135,19 @@ class CConfig {
...
   /*!
-   * \author H. Kline
-   * \brief Get the kind of objective function. There are several options: Drag coefficient,
-   *        Lift coefficient, efficiency, etc.
-   * \note The objective function will determine the boundary condition of the adjoint problem.
-   * \return Kind of objective function.
+   * \brief Similar to GetKind_ObjFunc but returns the corresponding string.
    */
-  unsigned short GetKind_ObjFunc(unsigned short val_obj) const { return Kind_ObjFunc[val_obj]; }
+  string GetName_ObjFunc(unsigned short val_obj = 0) const;
```
TobiKattmann, line 5150: for completeness can we get the param[in] and return here as well

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -3805,10 +3805,10 @@ class CSolver {
   /*!
    * \brief A virtual member. Extract and set the geometrical sensitivity.
    * \param[in] geometry - Geometrical definition of the problem.
-   * \param[in] solver - The solver container holding all terms of the solution.
    * \param[in] config - Definition of the particular problem.
+   * \param[in] solver - The target solver for the sensitivities, optional, for when the mesh solver is used.
    */
-  inline virtual void SetSensitivity(CGeometry *geometry, CSolver **solver, CConfig *config){ }
+  inline virtual void SetSensitivity(CGeometry *geometry, CConfig *config, CSolver *target_solver = nullptr){ }
```
TobiKattmann, line 3811: `target_solver` in the comment

SU2_CFD/src/iteration/CHeatIteration.cpp
```diff
@@ -40,54 +40,6 @@ void CHeatIteration::Iterate(COutput* output, CIntegration**** integration, CGeo
                                                                     RUNTIME_HEAT_SYS, val_iZone, val_iInst);
 }
 
-void CHeatIteration::Solve(COutput* output, CIntegration**** integration, CGeometry**** geometry, CSolver***** solver,
-                           CNumerics****** numerics, CConfig** config, CSurfaceMovement** surface_movement,
-                           CVolumetricMovement*** grid_movement, CFreeFormDefBox*** FFDBox, unsigned short val_iZone,
-                           unsigned short val_iInst) {
```
TobiKattmann, line 46: nice simplification with the empty pre/postprocessing routines 👍

TobiKattmann (approved):
Thanks pedro, I did not test the jacobian influence yet but will so do so soon. 
(This Approve excludes my bits in here of course as that is not exactly the idea of a PR-review :D )

SU2_CFD/include/solvers/CDiscAdjFEASolver.hpp
```diff
@@ -176,10 +176,9 @@ class CDiscAdjFEASolver final : public CSolver {
   /*!
    * \brief Extract and set the geometrical sensitivity.
    * \param[in] geometry - Geometrical definition of the problem.
-   * \param[in] solver - The solver container holding all terms of the solution.
    * \param[in] config - Definition of the particular problem.
    */
-  void SetSensitivity(CGeometry *geometry, CSolver **solver, CConfig *config) override;
+  void SetSensitivity(CGeometry *geometry, CConfig *config, CSolver*) override;
```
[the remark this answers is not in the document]
pcarruscag: You can think of it as the default value being part of the function signature, the signature is defined by the original virtual function which is all a caller needs to know (you include CIteration.hpp to use all the other iterations), the default value is passed when calling, not used when "receiving" the call.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -699,18 +703,21 @@ void CDiscAdjMultizoneDriver::SetObjFunction(unsigned short kind_recording) {
...
+          case MOMENT_Y_COEFFICIENT:
+          case MOMENT_Z_COEFFICIENT:
+          case FORCE_X_COEFFICIENT:
+          case FORCE_Y_COEFFICIENT:
+          case FORCE_Z_COEFFICIENT:
+            FieldName = config->GetName_ObjFunc();
+            break;
 
           // Other surface-related output values
+          // The names are different than in CConfig...
```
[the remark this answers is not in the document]
pcarruscag: Yes... those name choices were not very good, it needs to be fixed at some point, but it will have implications for the python stuff.

Common/include/CConfig.hpp
pcarruscag: sorry I just find that kind of description too much fluff, one of my pet peeves with this doxy stuff is that a ton of methods describe the same arguments over and over ad nauseam and they say very little of the actual method. Less is more is what I'm getting at.

pcarruscag: For example:
`* \brief  A virtual member.`
"No **** Sherlock, I could see that from just the code..." The important is what the method should do.
Anyway, rant over.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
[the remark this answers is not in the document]
TobiKattmann: Oh yes that `historyMap.py`... maybe automatically create that idk? But that is something for another day/PR :D

pcarruscag:
Merging this to allow all tests to run so we can create a release tomorrow (there is a fix in this branch for compilation with gcc 5.3.0 so this should be part of the release). Happy to make post mortem changes if needed.

PR #1111, opened 2020-11-19
Updates to NEMO solvers

WallyMaier:
This is a pull request to merge in progress made in the NEMO solver.   The goal here is keep the pull requests to reasonable sizes.  As such, there may be multiple PRs to prevent this from being a monumental task.

This pull request addresses some of the following:
-adoption of code structure in develop
       - CommonPreproccessing, SU2 screen output, etc.
-Incorporation of Smoluchowski-Maxwell BC
-general formatting
-correcting headers and briefs
-Updated test cases

- [ x] I have added a test case that demonstrates my contribution, if necessary.
- [ x] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.


SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -138,4 +141,150 @@ su2double CNEMOGas::GetrhoCvve() {
     return rhoCvve;
 }
 
+void CNEMOGas::GetdPdU(su2double *V, vector<su2double>& val_eves, su2double *val_dPdU){
```
WallyMaier, line 144: These functions were move from CUserDefinedLibTC in preparing for Mutation++

pcarruscag (commented):
Thanks @WallyMaier I think it is a good idea to have these regular updates, especially given the attention these features have been having on CFD-online recently.
A few non-essential comments below. I'll leave the approvals for the NEMO folks.

SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -138,4 +141,150 @@ su2double CNEMOGas::GetrhoCvve() {
...
+
+  /*--- Determine the electron density (if ionized) ---*/
+  if (ionization) { rho_el = rhos[nSpecies-1]; }
+  else {            rho_el = 0.0;              }
+
+  /*--- Necessary indexes to assess primitive variables ---*/
+  unsigned long RHOS_INDEX    = 0;
+  unsigned long VEL_INDEX     = nSpecies+2;
+  unsigned long RHOCVTR_INDEX = nSpecies+nDim+6;
+  unsigned long RHOCVVE_INDEX = nSpecies+nDim+7;
```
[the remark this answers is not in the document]
WallyMaier: @pcarruscag I agree...I am adding it to the evergrowing todo list.  hah.

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -115,27 +113,9 @@ CNEMOCompOutput::~CNEMOCompOutput(void) {}
...
-    /// DESCRIPTION: Root-mean square residual of the density.
-    AddHistoryOutput("RMS_DENSITY_NO",  "rms[Rho_NO]",  ScreenOutputFormat::FIXED, "RMS_RES", "Root-mean square residual of the NO density.", HistoryFieldType::RESIDUAL);
-    /// DESCRIPTION: Root-mean square residual of the density.
-    AddHistoryOutput("RMS_DENSITY_N",   "rms[Rho_N]",   ScreenOutputFormat::FIXED, "RMS_RES", "Root-mean square residual of the N density.", HistoryFieldType::RESIDUAL);
-    /// DESCRIPTION: Root-mean square residual of the density.
-    AddHistoryOutput("RMS_DENSITY_O",   "rms[Rho_O]",   ScreenOutputFormat::FIXED, "RMS_RES", "Root-mean square residual of the O density.", HistoryFieldType::RESIDUAL);
-  }
+  for(iSpecies = 0; iSpecies < nSpecies; iSpecies++)
+    AddHistoryOutput("RMS_DENSITY_" + std::to_string(iSpecies), "rms[Rho_" + std::to_string(iSpecies) + "]",   ScreenOutputFormat::FIXED, "RMS_RES", "Root-mean square residual of the species density " + std::to_string(iSpecies) + ".", HistoryFieldType::RESIDUAL);
```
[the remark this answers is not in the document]
WallyMaier: Thats all @CatarinaGarbacz  :)

SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -138,4 +141,150 @@ su2double CNEMOGas::GetrhoCvve() {
...
+  rhoCvtr = V[RHOCVTR_INDEX];
+
+  Cvtrs              = GetSpeciesCvTraRot();
+  Enthalpy_Formation = GetSpeciesFormationEnthalpy();
+  Ref_Temperature    = GetRefTemperature();
+
+  /*--- Calculate supporting quantities ---*/
+  v2 = 0.0;
+  for (iDim = 0; iDim < nDim; iDim++)
+    v2 += V[VEL_INDEX+iDim]*V[VEL_INDEX+iDim];
```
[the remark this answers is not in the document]
WallyMaier: Does this require a CGeometry class to be passed into function?

SU2_CFD/include/fluid/CNEMOGas.hpp
```diff
@@ -168,20 +171,25 @@ class CNEMOGas : public CFluidModel {
    */
   su2double GetPressure();
 
+  /*!
+   * \brief Get gas constant.
+   */
+  su2double GetGasConstant();
```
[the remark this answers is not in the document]
WallyMaier: Alot of "get" methods within SU2 do require computations...would you suggest renaming them to compute?

SU2_CFD/src/fluid/CNEMOGas.cpp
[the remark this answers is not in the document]
WallyMaier: This may be a dumb question....
would it be okay to include the geometrytool box in the fluid model to do this for a single computation? 

The NEMO team needs to spend some time making use of the toolbox.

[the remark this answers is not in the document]
pcarruscag: I think so, it is a very small header with just some functions, the benefit is knowing immediately what is being computed from just the function name, not comments needed or anything.
You can do "using namespace GeometryToolbox;" to simplify things a bit, but do that only in cpp files (if you do it in hpp it spreads and eventually may cause ambiguous function names.

SU2_CFD/include/fluid/CNEMOGas.hpp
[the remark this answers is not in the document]
CatarinaGarbacz: in that case they should all be changed to Compute yes, cause most of them are actual computations

SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -41,17 +42,21 @@ CNEMOGas::CNEMOGas(const CConfig* config): CFluidModel(){
...
   energies.resize(nEnergyEq,0.0);  
   ThermalConductivities.resize(nEnergyEq,0.0);
 
+  Kind_GasModel        = config->GetGasModel();
   Kind_TransCoeffModel = config->GetKind_TransCoeffModel();
 
-  frozen = config->GetFrozen();
-
+  frozen               = config->GetFrozen();
+  ionization           = config->GetIonization();
```
jtneedels, line 55: Is the plan to pass in a Boolean value for ionization from the config versus having it stored in the TCLib?

TestCases/nonequilibrium/viscwedge/viscwedge.cfg
```diff
@@ -41,15 +41,16 @@ AOA= 0.0
 SIDESLIP_ANGLE= 0.0
 %
 % Free-stream pressure (101325.0 N/m^2 by default, only Euler flows)  
-FREESTREAM_PRESSURE= 101325.0
+FREESTREAM_PRESSURE= 10.0
```
jtneedels, line 44: What motivates the drop in pressure here? 10.0 Pa seems low, does that correspond to condition of Re = 10000.

jtneedels (approved):

TestCases/nonequilibrium/viscwedge/viscwedge.cfg
WallyMaier: I think @CatarinaGarbacz or @fmpmorgado can shed light here.

SU2_CFD/src/fluid/CNEMOGas.cpp
WallyMaier: Im not quite sure I follow your question.

CatarinaGarbacz: @jtneedels it is already coming the a config input, that's why it's defined as **config->GetIonization();**

TestCases/nonequilibrium/viscwedge/viscwedge.cfg
CatarinaGarbacz: that's exactly it @jtneedels , the conditions are the same as before, but in terms of pressure and temperatures, instead of pressure and reynolds, so nothing changed in reality

CatarinaGarbacz: this is on the basis of what we talked in the last meeting, that we no longer have reynolds initialization because we would need to implement iterative methods for this. I didn't make this decision (with @MarcoFossati ) out of "laziness", but more because reynolds initialization doesn't add any particular advantage, it's really just a matter of what variabeles we provide in the config.

but if people specifically want reynolds initialization back in the code, we can make it a TO-DO

jtneedels: This makes sense, I think keeping initialization in terms of pressure/temperature is fine. I just wanted to make sure I understood the reason for the change.

WallyMaier:
I've changed some 'get' functions to 'compute' to be more clear.  After some conversations with @pcarruscag Ive also added some 'const' to some 'get' functions.  More work can be done in this regard, and that process is ongoing.   I can continue doing so in this PR, or in a future one.

CatarinaGarbacz (approved):
looks like things are in a good place after addressing all comments. I say we're good for merging

PR #1113, opened 2020-11-23
Fix for nDim initialization causing macos builds in develop

WallyMaier:
The fixes a change made by #1111 
The changes passed regressions however during the macos builds in develop.
I spoke with @economon and this should fix the problem.

## Related Work
#1111


economon (approved):
Building fine now on macOS.

Btw, there are quite a few unused private field warnings in the NEMO numerics headers. Worth cleaning up in a future PR when you have some time.

PR #1114, opened 2020-11-24
Windows Git Fix

WallyMaier:
This should fix github issues I have experienced on Windows.   

It looks like somewhere down the line a blank file named '  '  was added.


TobiKattmann (approved):
Well, looks straight forward to me 👍 

I guess the file name is just a single space and the ticks `' '` are just an OS way to wrap the filename (at least in linux) as it would otherwise misinterpret a space as a separator between two files. This allows for piping the `ls` output to s.th. else without any problems. The same can be achieved by escaping via (`\` + space) so in Linux you could open the file via `vim \ ` (yeah there is one space behind the `\`). For a filename that is 2 spaces it therefore would be ( tick + space + space + tick) or ( \ + space + \ + space). 

I have seen LinuxOS's doing the tick thing, the backspace thing and our cluster does the nothing-thing (where space is not indicated at all) ... not sure how that is determined, what is best or if the user can change that.

Funny though that github does not show that file at all 🤷‍♂️ 

Using [this](https://stackoverflow.com/questions/11533199/git-find-commit-where-file-was-added): the space-file was added here f6c6266c26f4a4ca43e546422048f02c45b85e99

PR #1115, opened 2020-11-25
Cleaning parts of the legacy output

pcarruscag:
## Proposed Changes
We need to do something about these 20k lines of legacy.
ATM it is a little hard to see what might need to be ported so I'm trying to remove unused chunks to hopefully make the task easier.
@vdweide it looks like all the legacy output FEM routines are not needed, it that right?

## PR Checklist


economon:
Just in case it is helpful to find more candidates in the legacy output, @pcarruscag : in the past, I have used the cppcheck tool to identify dead code for removal. Something like 
```
cppcheck --enable=unusedFunction -DHAVE_MPI -DHAVE_CGNS SU2_CFD/src/ SU2_CFD/include/ Common/src/ Common/include/ SU2_DEF/src/ SU2_DEF/include/ SU2_GEO/src/ SU2_GEO/include/ SU2_MSH/src/ SU2_MSH/include/ SU2_SOL/src/ SU2_SOL/include/ SU2_DOT/src/ SU2_DOT/include/
```

pcarruscag:
Thanks Tom, unfortunately there are routines that are inactive only because calls to them were commented out elsewhere in the code (e.g. all the inverse design stuff), I tried to keep all of those.
I ended up deleting everything dealing with volume and surface outputs, as those were 100% not used.
Only the history / convergence routines remain, I also fixed a bug affecting harmonic balance simulations where all outputs went to the same file.

Assuming harmonic balance is not used together with turbomachinery (I guess not since there is a driver class for each) it would be easy to move it over to the new output without too much loss of functionality.

The turbomachinery stuff looks a bit harder because I don't know if it needs any of the fluid fields, it would probably be very easy for the people who work with this stuff though...

Anyway it was a big mess, now it is smaller.

WallyMaier (approved):
@pcarruscag Thanks for going through the legacy output. I think this looks good.

PR #1117, opened 2020-12-01
Further NEMO cleanup and improvements

WallyMaier:
This pull request is another in a series of pull requests to merge developments of the NEMO branch into develop without a massive PR.  

This pull requests addresses:
-Fixes non-operable schemes (ROE, MSW, LAX)
-AUSMPWPLUS remains functional but extremely unstable => flag to make unusable.
-Removes config->GetGamma() in with an appropriate function
-Renames 'Get' functions to 'Compute' where applicable

Somethings that are in the works/on the radar:
-Migrate toward Geometry:ToolBox functions (Remove GetBasis function)
-Centralize the INDEX information
 

## Related Work
This is related to feature_NEMO and previous pull requests (#1111) 

## PR Checklist


SU2_CFD/obj/Makefile.am
```diff
@@ -113,14 +113,6 @@ libSU2Core_sources = ../src/definition_structure.cpp \
   ../src/numerics/elasticity/CFEANonlinearElasticity.cpp \
   ../src/numerics/elasticity/nonlinear_models.cpp \
   ../include/numerics_simd/CNumericsSIMD.cpp \
-  ../src/numerics/NEMO/NEMO_diffusion.cpp \
-  ../src/numerics/NEMO/NEMO_sources.cpp \
-  ../src/numerics/NEMO/convection/ausm.cpp \
```
WallyMaier, line 118: These were repeated.

SU2_CFD/src/fluid/CNEMOGas.cpp
```diff
@@ -131,9 +131,48 @@ su2double CNEMOGas::ComputeGasConstant(){
   return GasConstant;
 }
 
-su2double CNEMOGas::GetrhoCvve() {
+su2double CNEMOGas::ComputeGamma(su2double *V){
 
-    Cvves = GetSpeciesCvVibEle();
```
WallyMaier, line 136: Here there are two ComputeGamma() functions at the moment:
ComputeGamma() and ComputeGamma(*V)

The plan is to move towards tracking Gamma as a numerics vector

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -1758,40 +1756,61 @@ void CNumerics::GetPrimitive2Conservative (const su2double *val_Mean_PrimVar,
   }
 }
 
-
-
 void CNumerics::CreateBasis(const su2double *val_Normal) {
+
```
WallyMaier, line 1760: This function is a prime candidate to be phased out using the Geometry toolboxes.

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
```diff
@@ -178,30 +178,31 @@ CNumerics::ResidualType<> CUpwMSW_NEMO::ComputeResidual(const CConfig *config) {
...
-                                                      epsilon*epsilon));
-  Lambda_i[nSpecies+nDim-1] = 0.5*(ProjVelst_i + Vst_i[A_INDEX] +
-                                   sqrt((ProjVelst_i + Vst_i[A_INDEX])*
-                                        (ProjVelst_i + Vst_i[A_INDEX])+
-                                        epsilon*epsilon)                );
-  Lambda_i[nSpecies+nDim]   = 0.5*(ProjVelst_i - Vst_i[A_INDEX] +
-                                   sqrt((ProjVelst_i - Vst_i[A_INDEX])*
-                                        (ProjVelst_i - Vst_i[A_INDEX]) +
-                                        epsilon*epsilon)                );
+  for (iVar = 0; iVar < nSpecies+nDim-1; iVar++)
```
WallyMaier, line 188: Dont know what happened here....Ill fix this

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
```diff
@@ -222,15 +223,15 @@ CNumerics::ResidualType<> CUpwMSW_NEMO::ComputeResidual(const CConfig *config) {
     Lambda_j[iVar]          = 0.5*(ProjVelst_j - sqrt(ProjVelst_j*ProjVelst_j +
                                                       epsilon*epsilon));
   Lambda_j[nSpecies+nDim-1] = 0.5*(ProjVelst_j + Vst_j[A_INDEX] -
```
WallyMaier, line 225: Same here

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -11947,7 +11949,8 @@ void COutputLegacy::SetResult_Files_Parallel(CSolver *****solver_container,
         cout << "Loading solution output data locally on each rank." << endl;
 
       switch (config[iZone]->GetKind_Solver()) {
-        case EULER : case NAVIER_STOKES : case RANS :
+	case EULER : case NAVIER_STOKES : case RANS :
```
WallyMaier, line 11952: not aligned

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -110,11 +110,12 @@ void CNEMONSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_containe
   }
 
   /*--- Evaluate the vorticity and strain rate magnitude ---*/
+  //TODO: THIS NEEDS TO BE UPDATED ASAP!
```
WallyMaier, line 113: Working on this...

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1109,10 +1106,18 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
 
       /*--- Calculate specific gas constant --- */
+      //TODO: Move to fluidmodel?
       GasConstant = 0.0;
       for(iSpecies = 0; iSpecies<nSpecies; iSpecies++)
         GasConstant+=UNIVERSAL_GAS_CONSTANT*1000.0/Ms[iSpecies]*nodes->GetMassFraction(iPoint,iSpecies);
-
+      
+      /*--- Compute Gamma ---*/
+      //TODO: Move to fluidmodel?
```
WallyMaier, line 1115: This is the lone (i think) instance where gamma is computed outside of fluidmodel.   Should be phased out.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -43,28 +43,28 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
                                                                     ndim,
                                                                     nvar,
                                                                     config   ),
-                                      Gradient_Reconstruction(config->GetReconstructionGradientRequired() ? Gradient_Aux : Gradient) {
+                                      Gradient_Reconstruction(config->GetReconstructionGradientRequired() ? Gradient_Aux : Gradient_Primitive) {
```
WallyMaier, line 46: Need to check this...Grad Prim is not yet implemented for NEMO_merge branch.  Its on the way...

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -5320,25 +5320,25 @@ void COutputLegacy::SetConvHistory_Body(ofstream *ConvHist_file,
                   strcat(turbo_coeff, surface_coeff);
                 }
                 SPRINTF(surface_coeff, ", %12.10f", KineticEnergyLoss[iMarker_Monitoring][nSpanWiseSections]);
-								strcat(turbo_coeff, surface_coeff);
```
WallyMaier, line 5323: These are just some fixes for codefactor. I removed the tabs and added spaces.

SU2_CFD/src/fluid/CNEMOGas.cpp
CatarinaGarbacz: *numerics variable I assume, not vector (just clarifying for potential reviewers)

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -71,7 +71,8 @@ class CNEMOEulerVariable : public CVariable {
   MatrixType dTvedU; /*!< \brief Partial derivative of vib.-el. temperature w.r.t. conserved variables. */
   MatrixType eves;   /*!< \brief energy of vib-el mode w.r.t. species. */
   MatrixType Cvves;  /*!< \brief Specific heat of vib-el mode w.r.t. species. */
-
+  VectorType Gamma;  /*!< \brief Ratio of specific heats. */
+
```
CatarinaGarbacz, line 75: why vector ? isn't it just a variable ?

SU2_CFD/src/fluid/CUserDefinedTCLib.cpp
```diff
@@ -630,13 +630,14 @@ void CUserDefinedTCLib::SetTDStateRhosTTv(vector<su2double>& val_rhos, su2double
 
 vector<su2double>& CUserDefinedTCLib::GetSpeciesCvTraRot(){
 
+  //TODO: Set this in constructor.
   for (iSpecies = 0; iSpecies < nHeavy; iSpecies++)
     Cvtrs[iSpecies] = (3.0/2.0 + RotationModes[iSpecies]/2.0) * Ru/MolarMass[iSpecies];
 
   return Cvtrs;
 }
```
CatarinaGarbacz, line 638: question: I understand the suggestion, since it is a constant variable afterall. I wonder if we want to keep it here in case we want to go for more detailed models in the future, where this wouldn't be constant. Again, not a big deal, we can always change back. Just to share thoughts

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -1758,40 +1756,61 @@ void CNumerics::GetPrimitive2Conservative (const su2double *val_Mean_PrimVar,
...
+      modm += m[iDim]*m[iDim];
+      modl += l[iDim]*l[iDim];
+    }
+    modm = sqrt(modm);
+    modl = sqrt(modl);
+    for (iDim =0 ; iDim < nDim; iDim++) {
+     l[iDim] = l[iDim]/modl;
+     m[iDim] = m[iDim]/modm;
+    }
+
```
CatarinaGarbacz, line 1813: what is this change about exactly ? just to undderstand better

SU2_CFD/src/fluid/CNEMOGas.cpp
WallyMaier: Yes, well both.  It is stored in numerics as a VariableType

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1586,7 +1583,7 @@ void CNEMOEulerSolver::SetNondimensionalization(CConfig *config, unsigned short
 
     switch(config->GetKind_FluidModel()){
     case USER_DEFINED_NONEQ:
-      ModelTable << "User defined";
+      ModelTable << "SU2 Two-Temp";
```
CatarinaGarbacz, line 1586: hi @WallyMaier  actually I don't really agree here, that's why I had changed it in the past. 2T model is the whole NEMO solver model, despite of the fluidmodel. fluid model is either mutation++ library or the hardcoded (user defined), and both use 2T model. so this is not very consistent. I suggest we change back ?

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1101,17 +1097,39 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+      su2double  tmp1, scl, Cptr;
+      su2double *Vi = nodes->GetPrimitive(iPoint);
+
+      for (unsigned short iSpecies=0; iSpecies<nSpecies; iSpecies++)
+        Mass += Vi[iSpecies]*Ms[iSpecies];
+      Cptr = Vi[RHOCVTR_INDEX]+Ru/Mass;
+      tmp1 = Cptr*(Eddy_Visc/Prandtl_Turb);
+      scl  = tmp1/ktr;
+      ktr += Cptr*(Eddy_Visc/Prandtl_Turb);
+      kve  = kve*(1.0+scl);
```
CatarinaGarbacz, line 1115: nice!

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -192,11 +192,15 @@ bool CNEMOEulerVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidMode
...
 
+  /*--- Reset solution to previous one, if nonphys ---*/
   if (nonPhys) {
     for (iVar = 0; iVar < nVar; iVar++)
       Solution(iPoint,iVar) = Solution_Old(iPoint,iVar);
   }
 
+  /*--- Set additional point quantaties ---*/
+  Gamma(iPoint) = fluidmodel->ComputeGamma();  
+
```
CatarinaGarbacz, line 203: why here? I know it doesn't make a difference in the end, but just for tydiness, shouldn't this be computed next to other fluidmodel quantities, i.e, in cons2primvar ?

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -295,23 +300,23 @@ bool CNEMOEulerVariable::Cons2PrimVar(su2double *U, su2double *V,
     } else {
       V[TVE_INDEX]   = T[1];
     }
-  }
-  else {
+  } else {
+    //TODO: can e-modes/vibe modes be active?
```
CatarinaGarbacz, line 304: no, they can't. but I don't think there's a TO-DO here.

what we did here was: for monoatomic species the concept of Tve does not exist, and if we try to compute it as usual, it gives trouble... so what we did was simply always set it to the freeestream value, as a dummy value, since it needs to be populated with something, but a 'something' that doesn't disturb the rest, so we thought freestream would do the job (even if Tve is not set in the config, which theoretically shouldn't be for monoatomic species, we have the nice 288.15 value as default)

SU2_CFD/src/variables/CNEMONSVariable.cpp
```diff
@@ -109,16 +109,21 @@ bool CNEMONSVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)
...
 
+  /*--- Reset solution to previous one, if nonphys ---*/
   if (nonPhys) {
     for (iVar = 0; iVar < nVar; iVar++)
       Solution(iPoint,iVar) = Solution_Old(iPoint,iVar);
   }
 
+  /*--- Set additional point quantaties ---*/
+  Gamma(iPoint) = fluidmodel->ComputeGamma();  
+
```
CatarinaGarbacz, line 123: same comment as before - move it to cons2privar ?

CatarinaGarbacz (commented):
Hi @WallyMaier . I really like what you've been doing here. It's really good that we have several schemes now in place. And also all the small details that never get addressed in the middle of the hussle - very nice contribution, thanks. I will keep an eye on this

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
WallyMaier: Since we loop though iPoint in SetPrimVar, the Gamma variable would be overwritten each iteration.  So instead, it is saved as a vector so we store its value at each point.

However, if there is a better solution, Im all ears :)

SU2_CFD/src/fluid/CNEMOGas.cpp
CatarinaGarbacz: where exactly ? I don't think I am following. I guess I am just confused about gamma being a vector or not. I suppose it should just be a single variable, but it seems like it's not... or it's "both" as you say. not sure what that means, pls clarify a bit more with context of the code :)

SU2_CFD/src/fluid/CUserDefinedTCLib.cpp
WallyMaier: Thats a good point.  I think either option is good.  I can update the comment

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
CatarinaGarbacz: aaaaaaaaah. sorry, my bad! got it now

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: I was debating this as well.  I feel like "User Defined" wasnt descriptive enough.    

Would be "SU2 NonEq" be better?   If not, Ill change it back

CatarinaGarbacz: well not really, cause that is the same problem, both fluid models are NonEq. The only thing that distinguishes them is really that one is hardcoded and the other one is the coupling to mutation++...

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: Yeah, I was running into issues trying to do this.  So I followed the strategy in the NS version of this.

Since Cons2PrimVar doesnt have access to iPoint, I can't directly set it.  The other option would be to update Cons2PrimVar to return gamma.   Unless Im missing something.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: I gotcha.  I wasn't sure if a gas could have its electron state excited (without ejecting electrons) in a manner that impacted the non-equilibrium models.   But I can remove.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: ok I see the problem. then I think it's ok like this. Otherwise, something we could do, or not, just for tidyness, is gather all these variables that are not primitive or conservative (gamma, dPdstuff, ccves, eves) in some sort of Secondary or someething vector, and pass it the same way we do with primitives and conservatives, instead of passing/computing one by one... depends on how much longer if you feel lile post-poning your real-work you were trying to avoid xD

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: I see you thinking, but I think - might be wrong here - that electron do not have Tve. Tve is electronic-vibrational, referring to electronic states of atoms and molecules, and vibrational states of molecules. electronic state vs. electron. I think electron either have it's own temperature, or go with translational temperature - but I am not sure, maybe @jtneedels can help here

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Okie dok.  How about?
SU2 NonEq
Mutation++ NonEq

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: Ill make it a TODO in the NEMO project! hahah.  I think this will be the last general PR.   I need to start doing the Implicit!

SU2_CFD/src/numerics/CNumerics.cpp
WallyMaier: It add the the ability for both 2D and 3D problems.

SU2_CFD/include/fluid/CNEMOGas.hpp
```diff
@@ -158,25 +158,31 @@ class CNEMOGas : public CFluidModel {
...
    */
   su2double ComputePressure();
 
   /*!
-   * \brief Get gas constant.
+   * \brief Compute gas constant.
    */
   su2double ComputeGasConstant();
 
+  /*!
```
WallyMaier, line 180: Here are the overloaded functions....I hope to fix these..

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
```diff
@@ -222,17 +223,18 @@ CNumerics::ResidualType<> CUpwMSW_NEMO::ComputeResidual(const CConfig *config) {
...
-                                        (ProjVelst_j - Vst_j[A_INDEX])+
-                                        epsilon*epsilon)                 );
+                             sqrt((ProjVelst_j - Vst_j[A_INDEX])*
+                                  (ProjVelst_j - Vst_j[A_INDEX])+
+                                                epsilon*epsilon)                 );
   Lambda_j[nSpecies+nDim+1] = 0.5*(ProjVelst_j - sqrt(ProjVelst_j*ProjVelst_j+
                                                       epsilon*epsilon));
 
   /*--- Compute projected P, invP, and Lambda ---*/
+  CreateBasis();
```
WallyMaier, line 237: This needs an arguement

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
CatarinaGarbacz: for me is ok, yes

CatarinaGarbacz:
question: the chat we had about including gamma in the numerics, so we don't need an overload function, will be included in this PR? I think it should

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
jtneedels: So my understanding is that the a monatomic species can have Tve. Since monatomic species don't have vibrational energy, that contribution to the Eve (and therefore Tve) is zero. However, variation in the configuration of electrons in the electronic energy levels does contribute to change in Tve. Equation 2.33 in Sean Copeland's thesis gives the definition of vibrational-electronic energy for general atomic/molecular species (including monatomic) as well as electrons. Using this framework, electrons are defined as only being able to carry Tve. So even if electrons are not being ejected, we still need to account for the excitation of electronic energy levels in the monatomic species.

WallyMaier:
> question: the chat we had about including gamma in the numerics, so we don't need an overload function, will be included in this PR? I think it should

Yeah, that's in the works.  I wanted to get the rest of this up in the mean time.  It should be a matter of doing a couple of Get/Set and some testing.

jtneedels (approved):

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: thanks. I am following, but 2 things:
1- I understand that electronic levels will be there for monoatomic species. but if, in the 2t model, vibrational and electronic go together, it means that if the species is monoatomic, e_ve=0 => Tve = 0 despite. so in that sense, the code should remain as it is. correct me if I'm wrong pls
2- if you agree with 1, what we could do instead is set Tve = 0 , instead of freestream, and see if it doesn't cause any trouble

jtneedels: I am not sure that we can claim (1) directly from the assumptions of the 2T model. Mode separability implies that rho * e_ve = \sum{rho_s * (e_vib,s + e_el,s}). So even if the contribution of e_vib,s is zero, the net contribution to electronic energy from a monatomic species can be non-zero if it has a non-zero electronic energy contribution. My understanding is that the assumption of equilibrium between the vibrational and electronic energy modes does not imply that if a singles species has no vibrational modes its e_ve contribution is automatically zero.

CatarinaGarbacz: you're right, I know that ahah. my brain stopped working there for a moment. Then this means we need to review this. I'll add it as a TO-DO for now, and maybe we can discuss with tomorrow with everyone, especially @fmpmorgado who is the one that introduced the monoatomic mixtures

jtneedels: Yeah, that sounds great!

SU2_CFD/src/numerics/NEMO/convection/ausmplusup2.cpp
```diff
@@ -155,8 +157,8 @@ CNumerics::ResidualType<> CUpwAUSMPLUSUP2_NEMO::ComputeResidual(const CConfig *c
   }
 
   /*--- Compute C*  ---*/
-  CstarL = sqrt(2.0*(Gamma-1.0)/(Gamma+1.0)*h_i);
-  CstarR = sqrt(2.0*(Gamma-1.0)/(Gamma+1.0)*h_j);
+  CstarL = sqrt(2.0*(Gamma_i-1.0)/(Gamma_i+1.0)*h_i);
```
WallyMaier, line 160: These are now passed/stored in via the CVariable

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -667,55 +664,59 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
 
 void CNEMOEulerSolver::SetMax_Eigenvalue(CGeometry *geometry, CConfig *config) {
```
WallyMaier, line 666: This function seemed to be causing problems beforehand with the central residual schemes.  I updated it to follow develop.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -730,14 +731,11 @@ void CNEMOEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_c
   unsigned short iVar, jVar;
   bool err;
 
-  auto numerics = numerics_container[CONV_TERM];
+  CNumerics* numerics = numerics_container[CONV_TERM];
```
WallyMaier, line 734: Centered_Residual was out-of-whack, since it isnt used very often.  It seems to be working now.

WallyMaier:
I finally removed the overload ComputeGamma function.
There were some fixes required from the LAX solver  (I removed the implicit capabilities to save time), as well as the Centered_Residual computations.  This should be in a good place now, pending a couple more checks.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -912,6 +920,7 @@ void CNEMOEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_con
         numerics->SetdTvedU(dTvedU_i, dTvedU_j);
         numerics->SetEve   (Eve_i,    Eve_j   );
         numerics->SetCvve  (Cvve_i,   Cvve_j  );
+        numerics->SetGamma (Gamma_i,  Gamma_i );
```
CatarinaGarbacz, line 923: I know we had a discussion about this and I might be forgetting something here. but it seems to me that this won't work properly. the way it is now:

Gamma_i, Gamma_j have not been updated with reconstructed variables, so they will have the same value associated to non-reconstructed variables nodes->GetGamma(iPoint)/ nodes->GetGamma(jPoint)

this is because we are not updating gamma in Cons2PrimVar. 

now, I am not saying to correct the situation, since this issue comes up because reconstruction has not been updated to Primitve variables (it still is in conservative variables).

should this be the time to include all the reconstruction in primitive vairables and make things consistent in this way?

if you do not want to mix up primitive reconstruction in this PR, can we temporarily have the gamma computation inside Cons2PrimVar for consistency ?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1867,11 +1879,12 @@ void CNEMOEulerSolver::BC_Far_Field(CGeometry *geometry, CSolver **solver_contai
...
-        visc_numerics->SetdPdU(nodes->GetdPdU(iPoint),     node_infty->GetdPdU(0));
-        visc_numerics->SetdTdU(nodes->GetdTdU(iPoint),     node_infty->GetdTdU(0));
+        visc_numerics->SetdPdU  (nodes->GetdPdU(iPoint),   node_infty->GetdPdU(0));
+        visc_numerics->SetdTdU  (nodes->GetdTdU(iPoint),   node_infty->GetdTdU(0));
         visc_numerics->SetdTvedU(nodes->GetdTvedU(iPoint), node_infty->GetdTvedU(0));
-        visc_numerics->SetEve(nodes->GetEve(iPoint),       node_infty->GetEve(0));
-        visc_numerics->SetCvve(nodes->GetCvve(iPoint),     node_infty->GetCvve(0));
+        visc_numerics->SetEve   (nodes->GetEve(iPoint),    node_infty->GetEve(0));
+        visc_numerics->SetCvve  (nodes->GetCvve(iPoint),   node_infty->GetCvve(0));
+        visc_numerics->SetGamma (nodes->GetGamma(iPoint),  node_infty->GetGamma(0));
```
CatarinaGarbacz, line 1887: correct me if this is wrong, but I don't think we need to Set gamma for viscous fluxes, since it's not used there :)

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -261,6 +261,7 @@ void CNEMONSSolver::Viscous_Residual(CGeometry *geometry,
     numerics->SetdTvedU(nodes->GetdTvedU(iPoint), nodes->GetdTvedU(jPoint));
     numerics->SetEve   (nodes->GetEve(iPoint),    nodes->GetEve(jPoint));
     numerics->SetCvve  (nodes->GetCvve(iPoint),   nodes->GetCvve(jPoint));
+    numerics->SetGamma (nodes->GetGamma(iPoint),  nodes->GetGamma(jPoint));
```
CatarinaGarbacz, line 264: same comment as before, I don't think gamma is necessary for viscous fluxes

CatarinaGarbacz (commented):
@WallyMaier I made some comments about the implementation of gamma. Once these are addressed, one way or another, it seems to me this is good to go. 

all the rest concerns the fixing of central schemes and other upwind schemes. if you have tested them and they seem to work and give good results (please confirm this), I think we're ok

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: That is correct,  the gamma passed in a MUSCL flow is not based on the reconstructed state.  

However to fix this, it is not so simple...
A gamma needs to be passed for the upwind residual to work properly -> the lazy fix would be to have a computeGamma() in the convective schemes. In this case, storing the values of Gamma is pointless, since we compute it.

If we go the route of changing to primitive reconstruction...a decent amount of work will be needed to update/test the schemes properly.  I think this would be a good step to implement, but It will take time.  

With all that, I see two options... 
1.) leave it as it is, the performance seems to be fine with it.  However, we can't say its validated...
2.) Take down the PR, make the primitive changes and make a new PR.

Im okay with either.

SU2_CFD/src/solvers/CNEMONSSolver.cpp
WallyMaier: Yup!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Nope you are right! I removed the calls that were unnecessary!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
CatarinaGarbacz: I think option 2 is preferable but without wasting the work you've already done. Given that, and sSince this PR is basically almost finished (if not finished), why don't we go with 1+2?

CatarinaGarbacz: if you're ok with that, I give approval to this PR, we merge it, and jump to 2

WallyMaier:
Im closing this for now, as I have uncovered a bug.

WallyMaier:
AUSMPWPLUS works on my local linux machine, but fails on the cluster

CatarinaGarbacz (approved):
seems like the latest changes are ok to me :) I'll give approval

TestCases/nonequilibrium/viscwedge/viscwedge.cfg
```diff
@@ -4,8 +4,8 @@
 % Case description: Mach 5 viscous flow over a 10deg wedge                     %
 % Author: C. Garbacz                                                           %
 % Institution: Strathclyde University                                          %
-% Date: 2012.10.07                                                             %
-% File Version 5.0.0 "Raven"                                                   %
+% Date: 2020.11.01                                                             %
+% File Version 7.0.7 "Blackbird"                                               %
```
pcarruscag, line 8: Run a replace all on these versions before merging please.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -1509,16 +1509,48 @@ class CNumerics {
...
+   * \param[in] val_Cvve_j - Cvve at j.
+   */
   virtual inline void SetCvve(su2double *val_Cvve_i, su2double *val_Cvve_j)       { }
 
+  /*!
+   * \brief Set the ratio of specific heats.
+   * \param[in] val_Gamma_i - Gamma at i.
+   * \param[in] val_Gamma_j - Gamma at j.
+   */
+  virtual inline void SetGamma(su2double val_Gamma_i, su2double val_Gamma_j)       { }
```
pcarruscag, line 1552: Are there tabs in these files? Maybe you can run clang-format at some point (the idea is to eventually run in on the entire code base).

WallyMaier: I only use spaces.   But there may be tabs in the file.....I can run Clang-format to find them.

TestCases/nonequilibrium/viscwedge/viscwedge.cfg
WallyMaier: Got it!

PR #902, opened 2020-03-05
[WIP] pyBeam integration into SU2 as a subproject

rsanfer:
## Proposed Changes
This PR integrates pyBeam, a non-linear open source beam solver, into the SU2 infrastructure as a subproject, and adds support for coupled FSI simulations using SU2's flow solver and pyBeam's structural solver.

Tutorials on how to use the implementation will come soon. At this point, only minor changes to the compilation and the python scripts are required. The impact to the core of the code is minor.

I will be away for a few days, but please go ahead and have a look/comment/find room for improvement in the meantime :) I will reply to your comments as soon as I'm back to office.

I also added a couple of test cases, but I haven't yet found the best way to hook them to our regression test system. I'll definitely look into it before the PR is merged, but due to personal time constraints, it'd be great if this could be integrated for 7.0.3.

## Related Work
This work is the result of a collaboration with @RoccoBombardieri and @crauno from the University Carlos III Madrid. 

Please refer to a recent [paper at EUROGEN 2019](https://www.researchgate.net/publication/335972259_Towards_an_open-source_framework_for_aero-structural_design_and_optimization_within_the_SU2_suite) with regards to this work. Full reference will be provided here upon publication in an Springer ECCOMAS book series.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1124, opened 2020-12-05
Update of the python fsi interface and new feature coupling SU2 with Nastran

Nicola-Fonzi:
## Proposed changes

Dear all,

I am very new to collaboration on a open source project, but I have been working on an update for the FSI computation in python that I think may be useful also for other researcher. Would you mind giving me a feedback about the modifications?

In few words, I have updated the python interface for fluid structure interaction to the new version of SU2, including the new driver, the new mesh deformation methods and few other modifications in the python scripts. It works with python3.

As far as the C++ code is concerned, I only added the possibility to separate the symmetry boundary conditions for the fluid solver from the ones of the mesh solver. There is an explanation of this modification in the ReadMe file in SU2_PY/FSI_tools, together with an example of application. 

The interface is still general and new structural solvers can be coupled by adding it to the fsi_computation.py file. With the present pull request I only include one structural solver, which provides a coupling with the commercial code Nastran.

I added ReadMe files to explain in detail all the modifications and also the keywords to be used in the config files. Finally, there is also a complete test case where the flutter of a flat plate is studied.

I am looking forward for your comments

Nicola

## PR checklist


pcarruscag (commented):
This looks very interesting especially if it makes it easier to interface other solvers.

One think I must ask is to move all mesh files to the TestCases repository and all the config files to TestCases folder.
You could also try to re-use the mesh files across your different examples, we should try to reduce the size of the repository.

Alternatively you can also create some tutorials (since you already have some readme files), we have a special repository for those, I think a tutorial would give your work more visibility.

I will leave some comments on the code once I go through everything.

SU2_CFD/include/iteration/CIteration.hpp
```diff
@@ -101,6 +101,18 @@ class CIteration {
...
+    * \author R. Sanchez
+    * \param[in] geometry - Geometrical definition of the problem.
+    * \param[in] solver_container - Container vector with all the solutions.
+    * \param[in] config - Definition of the particular problem.
+    * \param[in] iZone - Index of the zone.
+    * \param[in] kind_recording - Current kind of recording.
+    */
+  void SetMesh_StaticDeformation(CGeometry** geometry, CSolver** solver_container, CNumerics*** numerics_container,
+                                 CConfig* config_container, unsigned short kind_recording);
+
```
pcarruscag, line 115: Can you explain why you need a "static" deformation? Is it to freeze grid velocities during a dynamic simulation?

Nicola-Fonzi: First of all thank you very much for your comments! I will move the test cases in the appropriate location and I will try to reduce the size as much as possible. I can definitely try to prepare a tutorial, tomorrow (it is evening here) I will try to have a look at what is already present in order to follow the same style.

The static deformation is needed in the very beginning of an unsteady simulation if a non-zero initial deformation is present. As the simulation is unsteady the code would try to perform a dynamic mesh update, creating fictitious velocities as the initial velocity is set to zero everywhere. In this way I basically statically deform the mesh and then push back the solution at previous times to avoid this problem.

pcarruscag: I see that makes sense, since this is only needed on the first iteration would it be possible to achieve the same result by calling the normal mesh deformation routine and them zero-ing the grid velocities? We should avoid duplicating things as much as possible.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -678,6 +739,55 @@ void CMeshSolver::SetBoundaryDisplacements(CGeometry *geometry, CNumerics *numer
     }
   }
 
+
+  /*--- Match deform plane is not clamped ---*/
+  for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
+    if (config->GetMarker_All_Match_Deform_Mesh(iMarker) == YES) {
+      const su2double* nodeCoord[MAXNNODE_2D] = {nullptr};
+
```
pcarruscag, line 747: On the subject of duplication, the FEASolver already implements a symmetry BC, it should be possible to use that instead of duplicating its code.

Common/include/CConfig.hpp
```diff
@@ -3366,6 +3376,13 @@ class CConfig {
    */
   void SetMarker_All_Deform_Mesh(unsigned short val_marker, unsigned short val_deform) { Marker_All_Deform_Mesh[val_marker] = val_deform; }
 
+  /*!
+   * \brief Set if a marker <i>val_marker</i> allows deformation at the boundary.
+   * \param[in] val_marker - Index of the marker in which we are interested.
+   * \param[in] val_interface - 0 or 1 depending if the the marker is or not a MATCH_DEFORM_MESH marker.
+   */
```
pcarruscag, line 3383: I like the idea for the "sliding" mesh deformation at some walls but the name is probably not very intuitive, maybe MARKER_DEFORM_MESH_SYM_PLANE?

I would also reduce the methods added to CConfig to the bare minimum as it is already a gigantic class, for example `GetnMarker_Match_Deform_Mesh` does not seem to be used.

pcarruscag: Oh! And we should prevent markers from being both DEFORM_MESH and DEFORM_MESH_SYM_PLANE

Nicola-Fonzi:
Thank you for the feedbacks, I am working on the modifications you suggested. I will perform a couple of test to be sure I did not break anything.

Hope to commit the new code soon!

SU2_PY/SU2_Nastran/TestCase/FlatPlateModel.m
```diff
@@ -0,0 +1,157 @@
+%% This script uses the theodorsen theory for the flutter of a flat plate
+% Author: Giampiero Bindolino, Nicola Fonzi, Vittorio Cavalieri
+%%% Structural definition
+clear all;close all;clc
```
pcarruscag, line 4: I don't know if this is used in any of your examples, but any code file in SU2 should carry the SU2 header and licence notice, otherwise technically it is proprietary code.

SU2_PY/SU2_Nastran/pysu2_nastran.py
```diff
@@ -0,0 +1,912 @@
+#!/usr/bin/env python
+
+## \file pysu2_nastran.py
+#  \brief Structural solver tester (one or two degree of freedom) used for testing the Py wrapper for external FSI coupling.
+#  \authors Nicola Fonzi, Vittorio Cavalieri, based on the work of David Thomas
+#  \version 7.0.6 "Blackbird"
```
pcarruscag, line 6: Remember to update the versions to 7.0.8

SU2_PY/SU2_Nastran/pysu2_nastran.py
```diff
@@ -0,0 +1,912 @@
...
+    iPoint = self.markers[markerID][iVertex]
+    Vel = self.node[iPoint].GetVel()
+    return float(Vel[1])
+
+  def getInterfaceNodeVelZ(self, markerID, iVertex):
+    """ Description """
+
+    iPoint = self.markers[markerID][iVertex]
+    Vel = self.node[iPoint].GetVel()
+    return float(Vel[2])
```
pcarruscag, line 891: I would simplify this type of function to return an array with all the components. The caller can then access xyz if it wants.

pcarruscag (commented):
I had a look at the python side today.
Looks very tidy but I think some comments would help follow what is going on.
Are you planning to add doc strings to the methods? (the """ Description. """ parts) I like the idea a lot :+1:

Nicola-Fonzi:
I think I implemented all the modifications you suggested. I am sorry you see all the commits in between, unfortunately due to Covid I work outside of office and I actually use Github to push the modified code to the office PC. I was actually working on a separate branch but, for reason that I do not understand, all the commits have been moved in the merging process... sorry about that. I am still learning git.

In particular the modifications are:

-  The functions related to static mesh deformation have been removed. I only included some new lines in the python wrapper 
    that overwrite the initial velocities to zero and push back the solution.
-  I now use the BC_Sym_Plane of the FEA solver for the deformation at the symmetry plane. I had to add a flag that avoids 
    accessing LinSysReact in case of mesh deformation, as this is not initialised in that context.
-  GetnMarker_Match_Deform_Mesh is not present anymore
-  The marker has been renamed from MATCH_DEFORM_MESH to DEFORM_MESH_SYM_PLANE. All the functions have also 
    been renamed accordingly
-  I included the SU2 header in all the new files, changing the version number to 7.0.8. I did not modify the version number of 
    the files that were already present in SU2 prior to this PR. I think the merging process should take care of that, am I wrong?
-  The python functions that were separated in x,y,z component now give back an array and are merged into one function only
-  The descriptions for the methods have been added
-  The test case has been removed. I actually prepared a tutorial and all the appropriate files will be placed in the tutorial and 
     website repos. I will now perform a PR for those repos so that you can see the material.

Again thank you very much and sorry for the mess with the "internal" commits. 

Please let me know if you think I missed something

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -908,6 +909,49 @@ void CFluidDriver::SetInitialMesh() {
...
+      for (unsigned long Point_Coarse = 0; Point_Coarse < geometry_container[ZONE_0][INST_0][iMesh]->GetnPoint(); Point_Coarse++) {
+
+        /*--- Overwrite fictitious velocities ---*/
+        su2double Grid_Vel[3] = {0.0, 0.0, 0.0};
+
+        /*--- Set the grid velocity for this coarse node. ---*/
+        for (unsigned short iDim = 0; iDim < geometry_container[ZONE_0][INST_0][MESH_0]->GetnDim(); iDim++)
+          geometry_container[ZONE_0][INST_0][iMesh]->nodes->SetGridVel(Point_Coarse, iDim, Grid_Vel[iDim]);
+      }
+    }
```
pcarruscag, line 940: Here you can run the loop over grids starting from 0 which would cover the MESH_0 case.
And since the loop is over nPoint (instead of nPointDomain) the communications are not required.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -908,6 +909,49 @@ void CFluidDriver::SetInitialMesh() {
...
+
+  /*--- Push back the solution so that there is no fictious velocity at the next step. ---*/
+  solver_container[ZONE_0][INST_0][MESH_0][MESH_SOL]->GetNodes()->Set_Solution_time_n();
+  solver_container[ZONE_0][INST_0][MESH_0][MESH_SOL]->GetNodes()->Set_Solution_time_n1();
+
+  // Push back the volume
+  for (iMesh = 0; iMesh <= config_container[ZONE_0]->GetnMGLevels(); iMesh++) {
+      geometry_container[ZONE_0][INST_0][iMesh]->nodes->SetVolume_n();
+      geometry_container[ZONE_0][INST_0][iMesh]->nodes->SetVolume_nM1();
+  }
```
pcarruscag, line 951: These can be included inside the SU2_OMP_PARALLEL section

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -2788,6 +2784,12 @@ void CFEASolver::Solve_System(CGeometry *geometry, CConfig *config) {
   /*--- Solve or smooth the linear system. ---*/
 
   auto iter = System.Solve(Jacobian, LinSysRes, LinSysSol, geometry, config);
+  if (rank == MASTER_NODE){
+    if (iter >= config->GetDeform_Linear_Solver_Iter()){
+      cout<<"Reached maximum number of iterations in structural deformation solver"<<endl;
+    }
+  }
+
```
pcarruscag, line 2792: Ah I forgot the other day, remove this please, there are outputs for linear solver residuals and iterations.

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -1846,9 +1842,10 @@ void CFEASolver::BC_Sym_Plane(CGeometry *geometry, CNumerics *numerics, const CC
 
     /*--- Set and enforce 0 solution for mesh deformation ---*/
     nodes->SetBound_Disp(iPoint, axis, 0.0);
-
     LinSysSol(iPoint, axis) = 0.0;
-    LinSysReact(iPoint, axis) = 0.0;
+    if (config->GetMarker_All_Deform_Mesh_Sym_Plane(val_marker) == NO){
+      LinSysReact(iPoint, axis) = 0.0;
+    }
```
pcarruscag, line 1848: I think there might be a more elegant way to do this check, I'll get back to you.

pcarruscag (commented):
Thanks! I'll finish going over it later or tomorrow. The internal commits are no problem. They make SU2 feel lively.

SU2_CFD/src/python_wrapper_structure.cpp
pcarruscag: You can use the version of SetGridVel that takes an array, so you dont need the loop over nDim

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: I thought there was a boolean but there is not, you can use
`if (LinSysReact.GetLocSize() > 0) ...`
This way we save users some debugging if they try to use that marker type with the FEA solver.

SU2_PY/SU2_Nastran/pysu2_nastran.py
```diff
@@ -0,0 +1,909 @@
...
+    for imode in range(self.nDof):
+      line = line + str(float(self.q[imode])) + '\t' + str(float(self.qdot[imode])) + '\t' + str(float(self.qddot[imode])) + '\t'
+    line =  line + '\n'
+    histFile.write(line)
+    histFile.close()
+
+  def updateSolution(self):
+    """
+    This method updates the solution.
+    """
```
pcarruscag, line 842: Thanks for the descriptions, everything looks nice.

pcarruscag:
You need to merge develop into this branch, resolve the conflicts that git is complaining above, and then this should be good to go.
We can give it a week or so to get other opinions, maybe @rsanfer or @RoccoBombardieri would like to have a look as well.
I think the work in #902 also changes some of the same files, it is a pity if this causes conflicts... but I guess that is the nature of open source.

SU2_PY/fsi_computation.py
```diff
@@ -73,81 +70,81 @@ def main():
...
-    elif CSD_Solver == 'TESTER':
-      SolidSolver = FSI.PitchPlungeAirfoilStructuralTester.Solver(CSD_ConFile)
+    if CSD_Solver == 'AEROELASTIC':
+      from SU2_Nastran import pysu2_nastran
+      SolidSolver = pysu2_nastran.Solver(CSD_ConFile,False)
+    elif CSD_Solver == 'IMPOSED':
+      from SU2_Nastran import pysu2_nastran
+      SolidSolver = pysu2_nastran.Solver(CSD_ConFile,True)
+    else:
+      print("\n I have no idea about the solid solver you want to use")
```
pcarruscag, line 121: Did any of these solvers work? Maybe the error message could be a bit more robot-like "Invalid solid solver option" or something like that.

SU2_CFD/src/solvers/CFEASolver.cpp
Nicola-Fonzi: Great! Thank you very much, I really appreciate your help

SU2_PY/fsi_computation.py
Nicola-Fonzi: I forgot about the message, I set it at the very beginning of my project and then it remained like that. 
The other structural solvers, except for the pitchplungeairfoil, were not present. I do not know them so I am not sure they would work. In any case, I had to do some modifications to the interface to update it so I am not sure they would still work.

Nicola-Fonzi:
I was trying to merge the develop to resolve conflicts and I found an error in the compilation. I thought I had broken something so I tried to start fresh new with the commands:
`git clone https://github.com/su2code/SU2.git`
`./meson.py build -Denable-pywrapper=true --prefix=/path/to/my/folder`
`./ninja -C build install`
Even from the master branch of your repo I still find the error attached below. Is it a problem with my compiler? I could always compile with no issues before. The last one I tried from you repo was 7.0.6.
error

Nicola-Fonzi:
I use openMPI 4.0.4 on a Linux CentOS 7

pcarruscag:
What about compiler version? If you do gcc --version what do you get? (assuming it is gcc)

Nicola-Fonzi:
gcc version 4.8.5

pcarruscag:
I think that was the first gcc with official full c++11 support, and so it might not be full support...
Try pushing the merge anyway so that github can run the tests, I'll see if something can be done about 4.8.5.

Nicola-Fonzi:
Okok, I may also try to update gcc. The only issue is that it is on a server so I should do it locally and it may take a bit more time. I will test the merge and then, with more time, update gcc.

Nicola-Fonzi:
I solved for the compilation issues by upgrading gcc. Thank you for the suggestion!

pcarruscag (approved):

pcarruscag:
I'll merge this now to try to conciliate it with #902 
I will probably ask you to check if I broke anything.

Nicola-Fonzi:
Ok I will now try to run two main test cases I have. But there were no conflicting files after the last commit so I do not think there should be any issue.

Nicola-Fonzi:
Everything seems to be fine. The test cases run smoothly.

I have been following the conversation with Rocco, I will now contact him to understand how to avoid duplication and how to best integrate our works

TobiKattmann:
> 
> 
> gcc version 4.8.5

yep I had to move to gcc 5.3.0 recently from 4.8.5 actually quite recently... which would be in line  with 7.0.6 working for you. I don't recall exactly what feature that was 🤔 Beginning of October with change to 7.0.7 and the SIMD stuff

pcarruscag:
Guilty as charged.

TobiKattmann (commented):
I know I am a bit late to the party. Thanks for the contribution and adding the tutorial as well 💐 
It would be nice to have a testcase in `tutorials.py` to safeguard it against future changes. Maybe it is wise to wait for after #902 if that changes sth (but I have no clue tbh). 
Usually not everything needs its own testcase but as this adds a tutorial, so I vote here in favor of testcase.

(btw I stumbled upon [this](https://www.youtube.com/watch?v=b6CKpnzdflk) today and thought I already heard that name of the first author somewhere already :D . Big fan of Mr Brunton 👍 )

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -661,13 +662,15 @@ void CMeshSolver::SetBoundaryDisplacements(CGeometry *geometry, CNumerics *numer
   /*--- Symmetry plane is clamped, for now. ---*/
   for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
     if ((config->GetMarker_All_Deform_Mesh(iMarker) == NO) &&
+        (config->GetMarker_All_Deform_Mesh_Sym_Plane(iMarker) == NO) &&
         (config->GetMarker_All_Moving(iMarker) == NO) &&
         (config->GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE)) {
 
       BC_Clamped(geometry, numerics, config, iMarker);
     }
```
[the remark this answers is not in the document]
Nicola-Fonzi: ahaha yes I also found it quite funny when I first saw it. But it is fine now to clamp it... As the sym plane for fluid and mesh are now separated, one will explicity use the deform mesh sym plane marker if she/he wants to obtain a symmetric deformation. Otherwise, the code will clamp deformation.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -891,6 +892,32 @@ void CFluidDriver::SetInitialMesh() {
...
+    // Overwrite fictious velocities
+    for (iMesh = 0u; iMesh <= config_container[ZONE_0]->GetnMGLevels(); iMesh++) {
+      SU2_OMP_FOR_STAT(roundUpDiv(geometry_container[ZONE_0][INST_0][iMesh]->GetnPoint(),omp_get_max_threads()))
+      for (unsigned long iPoint = 0; iPoint < geometry_container[ZONE_0][INST_0][iMesh]->GetnPoint(); iPoint++) {
+
+        /*--- Overwrite fictitious velocities ---*/
+        su2double Grid_Vel[3] = {0.0, 0.0, 0.0};
+
+        /*--- Set the grid velocity for this coarse node. ---*/
+        geometry_container[ZONE_0][INST_0][iMesh]->nodes->SetGridVel(iPoint, Grid_Vel);
```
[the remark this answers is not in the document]
Nicola-Fonzi: You are right!

Nicola-Fonzi:
The most critical part is the unsteady FSI, but that actually takes its time... You think it would be fine to compare the results after just few iterations for a testcase? We could use one of the simulations in the tutorial.

Thanks for the interest in the youtube video!

SU2_CFD/src/solvers/CMeshSolver.cpp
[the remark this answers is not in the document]
pcarruscag: That was because the FEA solver did not use to have a symmetry BC.
And then when I added it, it was better to clamp the symmetry plane for my testcases :)
But now that Nicola decoupled the physical symmetry from the deformation symmetry all is well xD

pcarruscag:
All testcases run for only a few steps, not entirely to convergence, the purpose is really just to detect change.

Nicola-Fonzi:
Ok, I will prepare a test case. I will take the Mach=0.1 case from the tutorial and change it slightly. Now the coupling between fluid and structure occurs after 100 steps, I will just couple them immediately. 
The only issue with #902 I see at the moment is with the file fsi_computation.py. The others are in separate folders so they should not create conflicts...

PR #1127, opened 2020-12-10
Introduce a function to compute stress tensor

maxaehle:
## Proposed Changes
Provide a function `CNumerics::ComputeStressTensor` to compute the laminar and turbulent stress tensor. Replace the code blocks currently computing it, which exist at many locations inside the code, by function calls.

## Related Work
?

## To be discussed
Is CNumerics the best location for this function?  

4 of the 5 `rans_uq` regression tests fail (in addition to those that failed before the change already). This seems to be due to floating-point errors in `CSourcePieceWise_TurbSST::SetReynoldsStressMatrix`, `CAvgGrad_Base::SetReynoldsStressMatrix`. These little deviations (relative difference around 1e-16) are probably amplified over the 10 iterations, so that the solver takes a notably different path to the steady state solution and the intermediate logarithmic residuals differ. I checked (for the `1c` and `2c` test cases) that the converged solutions are nearly equal though. (I can supply vtu files and plots if requested.)
Should I update the stored residual values in `serial_regression.py` then?

Also I am not sure whether I have found all code blocks that can be replaced by the new function.

## PR Checklist

- [X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).


Common/include/CConfig.hpp
```diff
@@ -8826,6 +8826,14 @@ class CConfig {
    */
   bool GetPrintInlet_InterpolatedData(void) const { return PrintInlet_InterpolatedData; }
 
+  /*!
+   * \brief Get information about using the Reynolds stress tensor and mean rate of strain matrix.
+   * \return <code>TRUE</code> means that they will be used.
+   */
+  bool GetUsing_ReynoldsStress(void) const {
+    return (using_uq);
+  }
```
pcarruscag, line 8835: Why is this needed?

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -459,6 +460,35 @@ class CNumerics {
...
 
+  /*!
+   * \brief Compute the mean rate of strain matrix.
+   * \details The parameter primvargrad can be e.g. PrimVar_Grad_i or Mean_GradPrimVar.
+   * \param[in] nDim - 2 or 3
+   * \param[out] rateofstrain - Rate of strain matrix
+   * \param[in] velgrad - A velocity gradient matrix.
+   */
+  static void ComputeMeanRateOfStrainMatrix(unsigned short nDim, su2double** rateofstrain, const su2double* const* velgrad);
+
```
pcarruscag, line 471: I would generalize this further by defining the functions inline and making them templates, this way you are not constrained by the ** interface and can pass anything for which [][] works. It would save some of the temporary pointers you are creating I think.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2190,16 +2189,9 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
-      for (iDim = 0; iDim < nDim; iDim++) div_vel += Grad_Vel[iDim][iDim];
-
-      for (iDim = 0; iDim < nDim; iDim++) {
-        for (jDim = 0; jDim < nDim; jDim++) {
-          Tau[iDim][jDim] = Viscosity * (Grad_Vel[jDim][iDim] + Grad_Vel[iDim][jDim]) -
-                            TWO3 * Viscosity * div_vel * delta[iDim][jDim];
-        }
-      }
+      su2double *Tau_pointer[3] = {Tau[0], Tau[1], (nDim==3)?Tau[2]:nullptr};
+      su2double *Grad_Vel_pointer[3] = {Grad_Vel[0], Grad_Vel[1], (nDim==3)?Grad_Vel[2]:nullptr};
```
pcarruscag, line 2193: Like these ones

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -472,6 +485,59 @@ void CNumerics::GetInviscidIncProjJac(const su2double *val_density, const su2dou
...
+    rateofstrain[1][2] = 0.0;
+    rateofstrain[1][0] = rateofstrain[0][1];
+    rateofstrain[2][1] = rateofstrain[1][2];
+    rateofstrain[2][0] = rateofstrain[0][2];
+  }
+}
+
+void CNumerics::ComputeStressTensor(unsigned short nDim, su2double** stress, const su2double* const* velgrad,
+                                    su2double viscosity, su2double density, su2double turb_ke, bool reynolds3x3){
+  su2double TWO3 = 2.0/3.0;
```
pcarruscag, line 518: That is really not needed IMO writing 2./3 directly is simpler.

SU2_CFD/src/numerics/flow/flow_diffusion.cpp
```diff
@@ -123,28 +123,22 @@ void CAvgGrad_Base::SetStressTensor(const su2double *val_primvar,
...
+        tau[iDim][jDim] += (-Density) * MeanPerturbedRSM[iDim][jDim];
   } else {
-
-    for (iDim = 0 ; iDim < nDim; iDim++)
-      for (jDim = 0 ; jDim < nDim; jDim++)
-        tau[iDim][jDim] = total_viscosity*( val_gradprimvar[jDim+1][iDim] + val_gradprimvar[iDim+1][jDim] )
-                        - TWO3*total_viscosity*div_vel*delta[iDim][jDim];
+    // compute both parts in one step
+    const su2double total_viscosity = val_laminar_viscosity + val_eddy_viscosity;
+    ComputeStressTensor(nDim, tau, val_gradprimvar+1, total_viscosity, Density, 0.0); // TODO why ignore turb_ke?
```
pcarruscag, line 141: That is a good question, you could try adding it to see how much it breaks the regressions. Maybe someone knows if it is for stability reasons.

pcarruscag (commented):
Absolutely awesome, this is my favorite kind of pull request!

If you feel like it, I think the QCR modification is also repeated in some places, and there is a bug in how it is computed (see #992).

See if the template suggestion makes sense to simplify things even a bit more. It should be fine to update the failing testcases (they are very sensitive ones).

pcarruscag:
Regarding finding all code blocks, there is numerics_simd/flow/diffusion/common.hpp :: stressTensor.
But it is not very easy to conciliate pointers with the data types used in those vectorized functions.

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -472,6 +485,59 @@ void CNumerics::GetInviscidIncProjJac(const su2double *val_density, const su2dou
...
+  }
+
+  for (unsigned short iDim = 0; iDim < nDim; iDim++){
+    for (unsigned short jDim = 0; jDim < nDim; jDim++){
+      stress[iDim][jDim] =
+        viscosity * (velgrad[iDim][jDim]+velgrad[jDim][iDim])
+        - TWO3 * viscosity * divVel * (iDim==jDim)
+        - TWO3 * density * turb_ke * (iDim==jDim);
+    }
+  }
```
pcarruscag, line 532: By the way it should be a little more efficient to write this like I have done for the SIMD stuff
https://gcc.godbolt.org/z/fK8PK3
https://gcc.godbolt.org/z/eMPen6

Common/include/CConfig.hpp
maxaehle: You are right. Commit 62444fde reverts the proposed changes of the allocation of `CNumerics::MeanReynoldsStress`.

clarkpede:
Thanks for the contribution! This is a much needed change. I wanted to perform a refactoring like this myself.  The viscous (or turbulent) stress calculation is repeated many times in the code.  Changing that calculation (e.g. adding QCR or UQ) currently requires changing many files.

In my humble opinion, the stress tensor calculation should have its own class at the lowest level of the code, like the gas model (e.g. ideal gas) or the viscosity calculation (e.g. Sutherland's law).  `CNumerics` needs access to it, and `CNumerics` is meant to have minimal dependencies.   But the stress isn't really "numerics." Its part of the underlying equations we're trying to solve.  Changing the stress tensor isn't the same flavor as changing from JST to Roe.  Because the stress is part of the underlying equations, it is sometimes needed in the `CSolver` classes.  Previously, I've seen the stress tensor used in `CNSolver` boundary conditions and axisymmetric source terms.  That may still be true in the current version of the code.

The best way to make sure that both `CSolver` and `CNumerics` have access to it is to make it an independent class. Unless, of course, we want to be manually setting the six tensor elements in `CNumerics`, like we do with primitive variables.  I don't like that alternative.

SU2_CFD/src/numerics/CNumerics.cpp
clarkpede: `TWO3` is also defined in alongside the other global variables in `option_structure.hpp`.  Since `CNumerics` includes `CConfig`, which includes `option_structure.hpp`, `TWO3` should be defined already in this scope.

SU2_CFD/include/numerics/CNumerics.hpp
maxaehle: Thank you for this suggestion. In commit 078427d, I replaced the signature by e.g.

    template<class TWOINDICES_1, class TWOINDICES_2>
    inline static void ComputeStressTensor(unsigned short nDim, TWOINDICES_1& stress, const TWOINDICES_2& velgrad,
          su2double viscosity, su2double density=0.0, su2double turb_ke=0.0, bool reynolds3x3=false)

and moved the body to the h-file. Is this the way you intended?

Yes the temporary arrays of pointers were only meant to convert `su2double[3][3]` into `su2double*[3]`. I removed them now.

SU2_CFD/src/numerics/CNumerics.cpp
maxaehle: Ok I changed it in 078427d to `2./3.` as well. I also consider this more readable.

SU2_CFD/src/numerics/CNumerics.cpp
maxaehle: Thanks -> 2e90bae.

pcarruscag:
Maybe CNumerics is not the perfect place, but it is good enough for government work (there are much more misplaced things in there).

The "everything is class" OO approach applied to the lowest level of abstraction is... well I think it is terrible -- and it has taken me a mighty long time to get rid of it in CPoint and CVariable and to design alternative Numerics -- because:
- Boilerplate: Set this, get that, constructor, destructor;
- Thread safety: Those classes always end up having some mutable state that renders them thread-unsafe;
- Correctness: Many of the classes we have follow this paradigm of "pass by member variable" - I like to know what are the inputs and outputs of something just by looking at the signature;
- Slowww: Too much virtual;
- Unnecessary complexity: A case of using a canon to kill the mosquito, good code should be as simple as possible, if a function does the job than that is the level of abstraction we should use.

My introduction to C++ was also the "everything is class", then one day I read "From Mathematics to Generic Programming" and well, I started liking C++ a whole lot more.
The standard template library is incredibly successful, and it "just" provides some containers and generic algorithms which are functions. That is what we need in SU2, some decent containers and generic algorithms to operate on them.

OO and its patterns are very good high level tools to achieve encapsulation and to isolate code, which are very important for projects with millions of lines of code, but for low level things they are overkill.

Sorry for the rant, I guess I have strong feelings about tiny classes.

pcarruscag (approved):
Thank you for the changes @maxaehle , LGTM

clarkpede:
@pcarruscag I agree about OO being overkill for the stress tensor.  My main concerns are:

+ We find a way to share the stress tensor calc between solver and numerics classes.
+ We make switching between different turbulent stress models (like normal vs QCR vs UQ) painless, without if/then branches being hardcoded for every time the stress is computed.

That can be accomplished in a number of ways, including functional or data-oriented programming.  And, since I'm not the one submitting the PR, I understand that my opinions are just opinions.  Pay as much or as little attention to them as you like.

pcarruscag:
Ah! uff glad we agree (I do pay attention to what you think).
I think the first concern is more or less covered, it can always be cut and pasted somewhere else, we have sgs and wall models floating around, maybe we should have a physical_models directory.
Regarding your second point I would should have the stressTensor function deciding what modifications to perform, QCR and UQ, based on input parameters.

clarkpede:
I just poked around in `flow_sources.cpp` and `CNSSolver`.  All the places where I remember there being a hard-coded stress tensor calculation have been updated in this PR.  I couldn't find any missing calculations. Great work.

In `CNSSolver::AddDynamicGridResidualContribution`, there is still a hardcoded Jacobian calculation.  That will break if the stress tensor calculation changes independently, without also changing the Jacobian calculation. I don't think that affects this PR.  I just wanted to document that here, in case later changes are made.

maxaehle:
I searched for `TWO3` in the project and this revealed two other locations. I adapted the one in `CSolver.cpp` in commit 64bd077. There is another one in `python_wrapper_structure.cpp`, lines 400-415. There I hesitate a bit because in the `incompressible` case, `div_vel` is manually set to zero. As long as the solver has not converged, this has an effect, right?

maxaehle:
I'll wait for the regression tests to fail and then update `serial_regression.py` etc for the `rans_uq` tests.

maxaehle:
Now the tests are fine. Can I merge?

pcarruscag:
You can replace the computation in python_wrapper_structure.cpp, the benefit of having a central definition outweighs the velocity divergence thing.
Also the way that method is used expects a reasonably converged solution, so it should not make a big difference.
Moreover that assumption for incompressible flow is not correct for variable density flows, so yes please replace it if you can.
I guess the TWO3 had a purpose in the end xD

maxaehle:
I observed that sometimes only the product of the stress tensor with a vector is of interest, e.g.:

    CNumerics::ComputeStressTensor(nDim, Tau, ...)
      for (jDim = 0 ; jDim < nDim; jDim++) {
        someforce[iDim] += Tau[iDim][jDim]*Normal[jDim];
      }
    }
    // never use Tau again

Do you think it is reasonable to have another function, which takes `someforce` and `Normal` instead of `Tau` as an argument and performs the addition? This would save from allocating `Tau` and going through two pairs of nested `for` loops.

pcarruscag:
Maybe if that was in the hot path of the code it would be worth trying.
But so long as tau is of static size the compiler seems to do a good job at unrolling those loops https://gcc.godbolt.org/z/9rcrsM

maxaehle:
Ok, so I will merge the pull request now?

pcarruscag:
I think you can merge since this is a much needed change (but usually we give it some time so everyone has a chance to review).
But in this case I would say it's fine to merge. Thanks again!

TobiKattmann:
> 
> 
> But so long as tau is of static size the compiler seems to do a good job at unrolling those loops https://gcc.godbolt.org/z/9rcrsM

oh man, I totally saw the unrolled loops in assembler as well 🤥  ... maybe one day 😢

TobiKattmann:
Hi @maxaehle , I just looked over this PR (after merging 🐌 ) and thanks a lot, it looks really tidy and I appreciate the detailed function documentation. 
Unfortunately I cant help with the `TODO why ignore turb_ke?` question

SU2_CFD/src/numerics/flow/flow_diffusion.cpp
clarkpede: Just noticed this.  Where to include the turbulent kinetic energy is a messy issue.  If I'm not mistaken, @economon removed the `turb_ke` from this calculation to fix the pressure problems described in: https://github.com/su2code/SU2/issues/797

PR #1129, opened 2020-12-10
Deprecate options that did not work since v7.0.0, and other minor cleanups

TobiKattmann:
## Proposed Changes
Hi all,
Following a bit the idea of #942 this PR mostly gets rid of stuff.

- `WRT_RESIDUALS` unused
 - `WRT_LIMITERS` unused
- `WRT_CON_FREQ_DUALTIME` only used in legacy output ... so I thought I just go for it
- removed `SetMeshFile` function as it not used anywhere and CSU2MeshFileWriter is the new cool kid on the block 🛴 
- Added some options to the config_template which were missing
- Added output of the mpi rank to volume output of (in-)compressible and heat solver. I used that for debugging mostly but it might be interesting for others as well. Yes yes I know, one could think of putting COORDS, MESH_QUALITY (except for FEM) and now RANK into a centralized place to avoid duplication... next PR... I promise ... maybe
- and a few tiny things
- EDIT: yeah and some more options after my initial PR, see CConfig.cpp in `files changed` for all

There is a lot of other things to deprecate so feel free to add here.

I have to check the tutorials for these options as well... ✔️ done and open PR (see below) but not merged yet to wait for reviews. This fails to reg tests of tutorials.

(p.s. github now has a dark mode, which makes writing this on Thursday nights much nicer 🕶️)

## Related Work
none

## PR Checklist

- [x] I have added a test case that demonstrates my contribution, if necessary. **not necessary**
- [x] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary. **not necessary**


pcarruscag:
Dark mode?
"Is it possible to learn this power?"

TobiKattmann:
> 
> 
> Dark mode?
> "Is it possible to learn this power?"

🧙 Sure, if you just go to your front page (i.e. just github.com) there should be a big button directly on the right side. 
Or Settings->Appearance->Dark. Enjoy :)

pcarruscag:
What is the equivalent of DUAL_TIME in the TIME / OUTER / INNER versions we have now? TIME?
(so that we can auto-replace the stuff that is used in the testcases without breaking things too much)

With all this deprecation the next version number should probably be 7.1.0 even if I would call these things bug fixes from v6.2 to v7.0.
Going 7.0.8 -> 7.1.0 also saves the awkward 7.0.9 -> 7.1.0 (which would make people wonder if we know how version numbers work).

TobiKattmann:
Well thanks for fixing the typo. 

It seems we use `OUTPUT_WRT_FREQ` for steady and unsteady simulations. Therefore the testcase cleanup was a bit more involved. For steady cases `WRT_SOL_FREQ` had to be replaced and for unsteady ones `WRT_SOL_FREQ_DUALTIME` to follow the author's intention... and I have the feeling that nearly every testcase had these options in it. 

I have to admit I don't understand why 7.0.9 would be awkward  🤔

TobiKattmann:
I also removed your options from the tutorials @pcarruscag : only WRT_CSV_SOL was used in two cases

pcarruscag:
Because 7.0.9 + 1 = 7.0.10 xD
Thanks for taking care of the tutorials, I have a few more options coming.

Silly question, using OUTPUT_WRT_FREQ = 1 to output on every time iteration, will it cause outputs on every inner iteration for some unsteady problems that were using the DUALTIME options?
(I don't know much about unsteady options...)

TobiKattmann:
> 
> 
> Silly question, using OUTPUT_WRT_FREQ = 1 to output on every time iteration, will it cause outputs on every inner iteration for some unsteady problems that were using the DUALTIME options?
> (I don't know much about unsteady options...)
For unsteady cases output is only written at the end of each "time"-iteration (if the OUTPUT_WRT_FREQ  is matched), never inside the inner-iterations. `OUTPUT_WRT_FREQ` works differently for steady and unsteady. So that is not an issue.

pcarruscag:
@TobiKattmann  are you going after WRT_CON_FREQ now?

TobiKattmann:
yep :) maybe still today if I can figure out how and where I broke the disc_adj_fsi testcase ...

pcarruscag:
It should be fixed in my last commit.

TobiKattmann:
As there were 256 WRT_CON_FREQ in the Testcases i removed that using:
`find . type f \( -name '*.cfg' \) -exec sed --in-place '/WRT_CON_FREQ/d' {} \+`
`find . type f \( -name '*.cfg' \) -exec sed --in-place '/Writing convergence history frequency/d' {} \+`

pcarruscag:
@TobiKattmann I tried to keep the option where the default value was not used.
```python
import os

rootDir = "."

old = "WRT_CON_FREQ"
new = "SCREEN_WRT_FREQ_INNER"
default = 1

for dirName, subdirList, fileList in os.walk(rootDir):
    
    for fileName in fileList:
        if not ".cfg" in fileName:
            continue

        with open(os.path.join(rootDir,dirName,fileName),"r") as f:
            lines = f.readlines()

        newLines = []
        i = 0;
        while i < len(lines)-1:
            if lines[i+1].startswith(old):
                if not lines[i].startswith("%") or len(lines[i]) < 5:
                    newLines.append(lines[i])

                val = lines[i+1].split("=")[1]
                
                if int(val) != default:
                    newLines.append(new+"="+val)
                i += 1
            else:
                newLines.append(lines[i])
            #end
            i += 1
        #end

        if not lines[-1].startswith(old):
            newLines.append(lines[-1])

        with open(os.path.join(rootDir,dirName,fileName),"w") as f:
            f.writelines(newLines)
    #end
#end
```

That seems to have confused git, maybe because of line endings on some testcases or something.

pcarruscag (approved):

TobiKattmann:
I was too lazy on that. Sorry, I should have done that that in the first place :disappointed: 

As there were only a few non-defaults I would have gone with:
`find . type f \( -name '*.cfg' \) -exec sed --in-place '/WRT_CON_FREQ= 1/d' {} \+`
and changing the rest manually (exercising the muscle memory a bit), but that handy script might get some additional use for other options ;)

If you have some more options to deprecate in mind I can take care of some. Maybe I try to hunt down a few more over the weekend :bow_and_arrow:

TobiKattmann:
> That seems to have confused git, maybe because of line endings on some testcases or something.

you prob mean that some files a "rewritten" completely. Hm seems to happen whenever a file ends on WRT_CONV_FREQ and  does not have the default value.

TestCases/hom_euler/NACA0012_5thOrder/fem_NACA0012.cfg
```diff
@@ -185,7 +185,6 @@ SURFACE_ADJ_FILENAME= surface_adjoint
 OUTPUT_WRT_FREQ= 50000
 %
 % Writing convergence history frequency
-WRT_CON_FREQ= 1
 %
```
pcarruscag, line 188: I don't think it was "third time lucky" xD
We should probably squash-merge to compress all these attempts.

TobiKattmann: shoot I forgot the text above 🤦 
Sure I am fine with squash merge. The upside of doing it as usual would be easy reverting of single option deprecations ... but I am not nostalgic towards any of these options, so lets squash it  🎾

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -360,7 +360,7 @@ void COutputLegacy::SetConvHistory_Header(ofstream *ConvHist_file, CConfig *conf
 
   bool output_surface = (config->GetnMarker_Analyze() != 0);
   bool output_comboObj = (config->GetnObj() > 1);
-  bool output_per_surface = config->GetWrt_Surface();
+  bool output_per_surface = true;
```
TobiKattmann, line 363: Obviously this kind of change (there are just 3 in this file) takes away functionality from users of output_structure_legacy. In these cases I was always generous 👑  and enabled stuff which potentially clutters output but I am unsure whether anybody even uses this at this point

TestCases/hom_euler/NACA0012_5thOrder/fem_NACA0012.cfg
pcarruscag: 99% of what we removed was not working, I don't think we need to bring them back.

SU2_CFD/src/output/output_structure_legacy.cpp
```diff
@@ -1249,34 +1249,33 @@ void COutputLegacy::SetConvHistory_Body(ofstream *ConvHist_file,
...
+    else write_heads = true;
 
-    bool write_turbo = (((iExtIter % (config[val_iZone]->GetWrt_Con_Freq()*40)) == 0) || (iExtIter == (config[val_iZone]->GetnInner_Iter() -1)));
+    bool write_turbo = true;
 
     /*--- Analogous for dynamic problems (as of now I separate the problems, it may be worthy to do all together later on ---*/
     bool write_heads_FEM;
     if (nonlinear_analysis) write_heads_FEM = (iIntIter == 0);
-    else write_heads_FEM = (((iExtIter % (config[val_iZone]->GetWrt_Con_Freq()*40)) == 0));
+    else write_heads_FEM = true;
```
pcarruscag, line 1278: Is there nothing equivalent for Wrt_Con_Freq?

TobiKattmann: Sure ... SCREEN_WRT_FREQ_INNER

Sometimes I am not really clever, I mean... its late ... but still

SU2_CFD/src/output/output_structure_legacy.cpp
TobiKattmann: EDIT: now it is just WRT_SURFACE

SU2_CFD/src/output/output_structure_legacy.cpp
pcarruscag: It is hard to resist not deleting that entire file xD

TobiKattmann:
Ok I am happy with the changes as they are right now. I've looked over the diff once again and I have no more comments (obviously more regarding @pcarruscag commits than my own). Btw thanks for being attentive regarding my mistakes and of course for all the added commits 👍

pcarruscag:
Thank you for taking the lead (this is boring stuff, but less so if more people do it).
I also cannot think of any more options, but I might try to do something about the objective functions.

TobiKattmann:
By the way, I noticed you didn't squash-merge. Any reasons or just forgotten? No big deal though

pcarruscag:
Cus yolo.
I have no idea how much that stuff matters to the repo size, so best not to prematurely optimize.

PR #1131, opened 2020-12-11
[WIP] CGNS output

EttoreSaetta:
## Proposed Changes
Hello to everyone,
my research group and I are working on the implementation of the CGNS output format.
We think that this could be an useful feature.

For the moment, in order to provide the output file in CGNS format you should choose `CGNS` in the `OUTPUT_FILES` option.
This is a first implementation that will need some improvements.
We hope that this could be useful.

Authors: Ettore Saetta, Lorenzo Russo, Renato Tognaccini.
TAARG (Theoretical and Applied Aerodynamics Research Group), University of Naples Federico II

## PR Checklist


MicK7:
Currently the CGNS library embedded with SU2 does not handle parallel writing.
How is it handled in the code ? Is one file per MPI process generated or only main process is doing the writing ?
It should be nice to have 2 classes : CCGNSMeshWriter and a CCGNSWriter (that call the other one for the mesh but focus on solution handling). My two cents.

Good idea to start a draft on this topic.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,305 @@
...
+    cgns_err = cg_nsections(cgns_file, nbases, nzones, &nsections);
+    if(cgns_err) cg_error_print();
+    for ( i = 1; i < nsections + 1; i++)    {
+        cgns_err = cg_section_read(cgns_file, nbases, nzones, i,
+                                   ElementSectionName, &type, &start,
+                                   &end, &nbndry, &parent_flag);
+        if(cgns_err) cg_error_print();
+        cgns_err = cg_ElementDataSize(cgns_file, nbases, nzones, i, &ElementDataSize);
+        if(cgns_err) cg_error_print();
+        if(type != 20){
```
MicK7, line 205: use named constant to keep the code easy to read instead of hard coded value
20 ==> 
#define MIXED 20
if (type != MIXED)

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,305 @@
...
+    base_file = fileName;
+    cgns_err = cg_open(base_file.c_str(), CG_MODE_WRITE, &cgns_file1);
+    if(cgns_err) cg_error_print;
+    cell_dim = 3;
+    phys_dim = 3;
+    cgns_err = cg_base_write(cgns_file1, basename, cell_dim, phys_dim,  &base_number);
+    if(cgns_err) cg_error_print;
+    cgns_err = cg_zone_write(cgns_file1, base_number, zonename, *isize, Unstructured, &zone_number);
+
+    /*---    Write Coordinates    ---*/
```
MicK7, line 178: Instead of copying the mesh data it should be nice to have the possibility to create a link to the mesh file.
cg_goto(cgns_file1, base_number, "Zone_t", zone_number, "end");
std::string path = basename + "/" + zonename + "/GridCoordinates";
cg_link_write("GridCoordinates", mesh_file.c_str(), path.c_str());
...

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+    }
+
+    /*---      Close CGNS Mesh File    ---*/
+    cgns_err = cg_close(cgns_file);
+    if(cgns_err) cg_error_print();
+
+    /*---    Write Solution    ---*/
+    cgns_err = cg_sol_write(cgns_file1, base_number, zone_number, (char *)"Solution", Vertex, &cgns_flow);
```
MicK7, line 265: tips: instead of writing everywhere cgns_err = cg_... , it can be usefull to define a small function like this
void callCGNS(int error){if (error) cg_error_print();}

thus leading to the wrapping   :    callCGNS(cg_sol_write(...));

EttoreSaetta:
Hello @MicK7, thanks for your review.

> Currently the CGNS library embedded with SU2 does not handle parallel writing.
> How is it handled in the code ? Is one file per MPI process generated or only main process is doing the writing ?
Yes we notice that it does not handle parallel writing, so I decided to gather the data from the processes to the master node, so that only the main process writes the solution file.

> It should be nice to have 2 classes : CCGNSMeshWriter and a CCGNSWriter (that call the other one for the mesh but focus on solution handling). My two cents.
> 
> Good idea to start a draft on this topic.
Thanks for your suggestion, it would be more tidy. It is just a draft for the moment, we still have to work on it.

Thank you for all the other proposed changes also, I will work on it in the next few days.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -707,6 +708,29 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
+        /*--- Load and sort the output data and connectivity. ---*/
+
+        volumeDataSorter->SortConnectivity(config, geometry, true);
+        volumeDataSorter->SortOutputData();
+
+        /*--- Write a CGNS file ---*/
+        if (rank == MASTER_NODE)
+          (*fileWritingTable) << "CGNS" << fileName + CCGNSFileWriter::fileExt;
+
+        fileWriter = new CCGNSFileWriter(fileName, volumeDataSorter);
```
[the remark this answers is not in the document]
EttoreSaetta: I have introduced the SetConfig function. Now the Write_Data_CGNS() is not needed anymore.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -715,8 +739,10 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
   if (fileWriter != nullptr){
 
     /*--- Write data to file ---*/
-
-    fileWriter->Write_Data();
+   if(format == CGNS)
```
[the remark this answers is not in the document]
EttoreSaetta: Now Write_Data() works without passing config, this is connected to the upper comment.
Thanks for the suggestion.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
EttoreSaetta: MIXED defined, now it is more easy to understand.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,306 @@
...
+  for (int iPoint = 0; iPoint < nLocalVertex; iPoint++) {
+      for(int varCount = 0; varCount < nVar; varCount++){
+          buffLocData[index] = dataSorter->GetData(varCount, iPoint);
+          index++;
+      }
+  }
+
+  #ifdef HAVE_MPI
+      SU2_MPI::Barrier(MPI_COMM_WORLD);
+  #endif
```
pcarruscag, line 85: Whenever you use SU2_MPI:: the "ifdef endif" should not be necessary.
The "ifdef" is from the days when the MPI routines were called directly rather than via the SU2_MPI wrapper.

config_template.cfg
```diff
@@ -1457,7 +1457,7 @@ TABULAR_FORMAT= TECPLOT
 % Files to output 
 % Possible formats : (TECPLOT, TECPLOT_BINARY, SURFACE_TECPLOT,
 %  SURFACE_TECPLOT_BINARY, CSV, SURFACE_CSV, PARAVIEW, PARAVIEW_BINARY, SURFACE_PARAVIEW, 
-%  SURFACE_PARAVIEW_BINARY, MESH, RESTART_BINARY, RESTART_ASCII, CGNS, STL)
+%  SURFACE_PARAVIEW_BINARY, MESH, RESTART_BINARY, RESTART_ASCII, CGNS, STL, CGNS)
```
pcarruscag, line 1460: CGNS was already in the list.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,306 @@
...
+    cgsize_t A[4], B[4];
+    cgsize_t *Elements = NULL;
+    cgsize_t *NewElements = NULL;
+    PointSetType_t ptset_type;
+    BCType_t bocotype;
+    DataType_t datatype;
+    GridLocation_t location;
+    ElementType_t type;
+    DataClass_t dataclass;
+    #define MIXED 20
```
pcarruscag, line 141: It would be preferable to use a variable rather than a macro for these constants.
(because macros maybe defined somewhere else)

pcarruscag (commented):
Nice to have this addition to the code, I don't know much about CGNS so these are maybe dumb questions:
- Will this also work to output meshes after deformation with SU2_DEF? I am guessing not because we had that issue with where to put the FFD information.
- Is it possible to write the file by sections? i.e. instead of gathering all the data in a single node do a number of Send/Recv to avoid running out of memory for large problems.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
```diff
@@ -0,0 +1,306 @@
...
+            cgns_err = cg_poly_section_write(cgns_file1, base_number, zone_number,
+                                             ElementSectionName, type, start,
+                                             end, nbndry, Elements, ConnectOffset, &S);
+            if(cgns_err) cg_error_print();
+        }
+        /*---    Add Boundary Condition    ---*/
+        for (j = 0; j <= config->GetnMarker_All(); j++){
+            if (ElementSectionName == config->GetMarker_All_TagBound(j)){
+                if (config->GetMarker_All_KindBC(j) == 26){
+                    bocotype = (BCType_t)20;
```
pcarruscag, line 228: Can these constants, and a few below, be given names as well? like Mick suggested for "MIXED"

EttoreSaetta:
Hello @pcarruscag, thank you for your suggestions.

> Nice to have this addition to the code, I don't know much about CGNS so these are maybe dumb questions:
> 
>     * Will this also work to output meshes after deformation with SU2_DEF? I am guessing not because we had that issue with where to put the FFD information.

We did not think to output CGNS meshes with SU2_DEF yet. Maybe it is a development that we can take into account.
For the moment, it works just to output the flow-field in CGNS format running SU2_SOL or SU2_CFD.

>     * Is it possible to write the file by sections? i.e. instead of gathering all the data in a single node do a number of Send/Recv to avoid running out of memory for large problems.

I was thinking about problems with the memory. Maybe one possibility is to make each process to write a CGNS file and then merge them into one file (but I have never tried it).
The CGNS library currently embedded with SU2 does not support the parallel writing.

config_template.cfg
EttoreSaetta: I am sorry, I didn't notice it.
Removed.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -707,6 +708,30 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
+        if (rank == MASTER_NODE)
+          (*fileWritingTable) << "CGNS" << fileName + CCGNSFileWriter::fileExt;
+
+        fileWriter = new CCGNSFileWriter(fileName, volumeDataSorter);
+        fileWriter->SetConfig(config);
+        break;
+      }
+      else{
+        cout << "CGNS output not supported with SU2 mesh file format." << endl;
+        break;
```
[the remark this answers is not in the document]
EttoreSaetta: Thank you for that, I fixed it throwing an error.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
EttoreSaetta: Thank you, I removed the `#ifded #endif`

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
EttoreSaetta: I changed it into a variable.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
EttoreSaetta: Yes, I gave them a name declaring new variables. Now it should be more clear.

SU2_CFD/src/output/filewriter/CCGNSFileWriter.cpp
EttoreSaetta: This is a good idea, thank you for this tip.
I have defined the `callCGNS` function, now the code is more tidy.

MicK7:
- having a configuration switch to the class so that each process write its own file should work really well (already seen this stuff in other cfd packages) : it is very fast and not too memory consuming

- To cope with high memory usage, the only other solution is to enable parallel cgns in su2. But the problem may be that it requires hdf5 parallel library. @pcarruscag do you know if it is a licence issue or just technical difficulties (hdf5 use cmake and not meson) that prevent adding hdf5 to the external libraries ?

pcarruscag:
I don't know @MicK7, but for the other cgns stuff Tim Albring created a meson.build file (and for metis, parmetis, etc.)
I asked about writing in chunks because that is what we do to output the native mesh file.
In any case we don't need to solve every possible problem right now

PR #1133, opened 2020-12-13
Restructure MPI wrapper, move more files to directories

pcarruscag:
## Proposed Changes
Move a few files, mainly mpi_structure.hpp, remove the .inl and make the mpi functions const correct according to the modern versions of the MPI standard.
Compilation with ancient MPI will probably need -fpermissive.

## Related Work
#583 

## PR Checklist


TobiKattmann:
It works with gnu/5.3.0 and openMPI/3.1.6 (@bigfooted) ... but I just saw that openMPI/3.1.6 was released early 2020 so it is not that ancient anyway👴

Common/include/mpi_structure.inl
```diff
@@ -1,618 +0,0 @@
...
-}
-
-inline void CBaseMPIWrapper::Reduce(void *sendbuf, void *recvbuf, int count,
-                                Datatype datatype, Op op, int root, Comm comm) {
-  MPI_Reduce(sendbuf, recvbuf,count,datatype,op,root,comm);
-}
-
-inline void CBaseMPIWrapper::Allreduce(void *sendbuf, void *recvbuf, int count,
-                                   Datatype datatype, Op op, Comm comm) {
-  MPI_Allreduce(sendbuf,recvbuf,count,datatype,op,comm);
```
TobiKattmann, line 167: non-const sendbuf in the old implementation

Common/include/parallelization/mpi_structure.hpp
```diff
@@ -0,0 +1,687 @@
...
+  }
+
+  static inline void Reduce(const void* sendbuf, void* recvbuf, int count, Datatype datatype, Op op, int root,
+                            Comm comm) {
+    MPI_Reduce(sendbuf, recvbuf, count, datatype, op, root, comm);
+  }
+
+  static inline void Allreduce(const void* sendbuf, void* recvbuf, int count, Datatype datatype, Op op, Comm comm) {
+    MPI_Allreduce(sendbuf, recvbuf, count, datatype, op, comm);
+  }
```
TobiKattmann, line 212: here now const sendbuf. Now as long as nobody writes code before the MPI_Allreduce call that changes sendbuf we were good in terms of correctness before, right? const probably also adds a little performance gain here? 

Are those the the only changes to this hpp+inl merge?

Common/lib/Makefile.am
```diff
@@ -47,13 +47,12 @@ endif
 lib_sources = \
   ../src/fem/fem_cgns_elements.cpp \
   ../src/CConfig.cpp \
-  ../src/blas_structure.cpp \
+  ../src/linear_algebra/blas_structure.cpp \
```
TobiKattmann, line 50: good job thinking about the make build system 👍

TobiKattmann (approved):
A merge of mpi_structure hpp and inl. Moving 4 files to more appropriate places. A lot of include changes because of that and deleted trailing whitespaces. LGTM and thanks 💐

Common/include/parallelization/mpi_structure.hpp
pcarruscag: Before if we had something like:
```c++
struct A {
  int a;
  int fun() const {
    SU2_MPI::Allreduce(&a,...);
  }
};
```
It would not work because "a" is const and the compiler will not let you take a non-const address.
As we make more parts of the code const correct and use containers instead of bare pointers, these non const "sendbuffs" would be a problem.

pcarruscag: Ah! and yes those are the only things that changed.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1366,13 +1372,10 @@ void CFlowOutput::WriteForcesBreakdown(CConfig *config, CGeometry *geometry, CSo
 
     }
 
-
     /*--- Write file name with extension ---*/
 
-    string filename = config->GetBreakdown_FileName();
-    strcpy (cstr, filename.data());
```
TobiKattmann, line 1373: Ggood thing imo to replace those `strcpy` operations. I guess we could get rid of all of these (on another day in another PR ;) ), or is there any reason to not prefer directly working with strings?

TobiKattmann: Well, I see you just removed some more `strcpy`'s 👍

jblueh:
Is it ok if I add a MeDiPack update for const correctness of the scatter routines to this PR?

SU2_CFD/src/output/CFlowOutput.cpp
pcarruscag: In some places they are needed because the constructor of CConfig for example takes a char array.
Some are there because SPRINTF is used to modify the name, that would have to be replaced probably by stringstreams.
I tried to remove some obvious ones that were 100% not needed.

pcarruscag:
Go for it @jblueh.

PR #1135, opened 2020-12-16
Removing the tabs in all files

WallyMaier:
This change should have removed all the tabs in SU2.
Hopefully code-factor agrees with me.
 
This is a general maintenance pr. 


pcarruscag (approved):
Awesome :+1: I was expecting 1M lines changed xD seems we are doing alright.

TobiKattmann (approved):
Thanks Wally 💐 looks like a no🧠er. 

#808 @pcarruscag, that's prob why. There is also a little wonky bash script in <root/externals/utils/> to trim trailing whitespaces and replace tabs... poor man's clang-format

PR #1138, opened 2020-12-18
Fix compiler warnings

pcarruscag:


SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -987,8 +989,7 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
 
 
   unsigned short iDim, jDim, iVar, iSpecies;
-  unsigned short T_INDEX, TVE_INDEX, VEL_INDEX,
-      RHOCVTR_INDEX, RHOCVVE_INDEX;
+  unsigned short T_INDEX, TVE_INDEX, VEL_INDEX, RHOCVTR_INDEX;
```
pcarruscag, line 992: @WallyMaier @CatarinaGarbacz there is a bug here, RHOCVTR_INDEX is used below but not initialized.

WallyMaier: @pcarruscag Thanks for catching.   Ill take a look.

Also, thank you for catching all those spacing issues and such in NEMO!

pcarruscag: The IDE does the spaces for me :)

PR #1139, opened 2020-12-18
NEMO - transition to primitive limiter

WallyMaier:
## Proposed Changes
The PR changes the current limiter routine with the NEMO solver from a conservative variable limiter scheme to limiting on the basis of the primitive variables.

This also updates some routines within SU2-NEMO to look/behave/take advantage of the standard SU2 implementations.

I also tried tried to remove any other compiler errors.

## Related Work
This was first done with the AUSM scheme by @fmpmorgado. 
This is working towards functional implicit and dual-time capabilities in SU2-NEMO

## PR Checklist


SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -434,98 +434,108 @@ unsigned long CNEMOEulerSolver::SetPrimitive_Variables(CSolver **solver_containe
 void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_container, CConfig *config,
                                     unsigned short iMesh, unsigned long Iteration) {
 
-  su2double Area, Vol, Mean_SoundSpeed = 0.0, Mean_ProjVel = 0.0, Lambda, Local_Delta_Time,
-  Global_Delta_Time = 1E6, Global_Delta_UnstTimeND, ProjVel, ProjVel_i, ProjVel_j;
+  const bool viscous       = config->GetViscous();
```
WallyMaier, line 437: Rewrote this function to mirror the CEulerSolver implementation

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -762,178 +787,179 @@ void CNEMOEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_c
...
+      nodes->SetNon_Physical(jPoint, chk_err_j);
+
+      /*--- Get updated state, in case the point recovered after the set. ---*/
+      chk_err_i = nodes->GetNon_Physical(iPoint);
+      chk_err_j = nodes->GetNon_Physical(jPoint);
+
+      counter_local += chk_err_i + chk_err_j;
+
+      /*--- Compute Secondary variables in a thermaodynamically consistent way. ---*/
+      if (!chk_err_i) Gamma_i = nodes->ComputeConsistentExtrapolation(Primitive_i, dPdU_i, dTdU_i, dTvedU_i, Eve_i, Cvve_i);
```
WallyMaier, line 937: This function computes all the secondary variables needed.  However, d_dU* are really only needed for the implicit solver.  This is not a perfect solution, but my ears are open!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -762,178 +787,179 @@ void CNEMOEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_c
...
+      if (!chk_err_j) Gamma_j = nodes->ComputeConsistentExtrapolation(Primitive_j, dPdU_j, dTdU_j, dTvedU_j, Eve_j, Cvve_j);
+
+      /*--- Recompute Conserved variables if Roe or MSW scheme ---*/
+      if ((config->GetKind_Upwind_Flow() == ROE) || (config->GetKind_Upwind_Flow() == MSW)){
+        if (!chk_err_i) nodes->Prim2ConsVar(Conserved_i,Primitive_i);
+        if (!chk_err_j) nodes->Prim2ConsVar(Conserved_j,Primitive_j);
```
WallyMaier, line 943: This function is only required for the MSW and ROE schemes (at the moment), since we need to compute P_Matrix and GetInviscidProjFlux() with the averaged states. These functions are outdated, so work could probably be done to modifiy them.  

This function could be moved into the specific numerics routines.

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -333,19 +350,35 @@ class CNEMOEulerVariable : public CVariable {
...
+   *        in a more thermodynamically consistent way.
+   * \param[in] V - primitve variables.
+   * \param[out] d*dU - reconstructed secondaryvariables.
+   * \param[out] val_eves - reconstructed eve per species.
+   * \param[out] val_cvves - reconstructed cvve per species.
+   * \param[out] Gamma - reconstructed gamma.
+   */
+  su2double ComputeConsistentExtrapolation(su2double *V, su2double* dPdU,
+                                           su2double* dTdU, su2double* dTvedU,
+                                           su2double* val_eves,su2double* val_cvves);
```
pcarruscag, line 381: When something is an input passed by reference or pointer it should be const (likewise for the method itself if it does not change the object).
This function caught my eye because it has both a return and output parameters, and because it does not have iPoint as input, which maybe suggests this can be static? Or that V is not needed because it is already there?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1834,6 +1834,12 @@ class CVariable {
    */
   inline virtual void SetVelocity_Old(unsigned long iPoint, const su2double *val_velocity) {}
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] val_energy - Pointer to energy.
+   */
+  inline virtual void SetEnergy_Old(unsigned long iPoint, vector<su2double>& val_energy) {}
```
pcarruscag, line 1841: Looks like this is not used.

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
```diff
@@ -198,7 +198,7 @@ CNumerics::ResidualType<> CUpwMSW_NEMO::ComputeResidual(const CConfig *config) {
                                                       epsilon*epsilon));
 
   /*--- Compute projected P, invP, and Lambda ---*/
-  CreateBasis(UnitNormal);
+  CreateBasis   (UnitNormal);
   GetPMatrix    (Ust_i, Vst_i, dPdUst_i, UnitNormal, l, m, P_Tensor   );
```
pcarruscag, line 202: You will not be pleased when we eventually run clang-format... :/

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
WallyMaier: Thanks for taking a look! I appreciate the feedback and advice :)

I wasn't sure how to best handle this function.  I could pass gamma by reference as well.   That being said, I am admittedly lacking in the proper use of "static" functions.   The V vector is used to compute the dPdU, etc after its been recomputed in the primitive limiter schemes.

Please correct me if I am wrong here, V can be passed as const,  but since dPdU, eves, etc are computed in the function, they cant be const?

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
WallyMaier: Yeah....hahah. But ill change this back.

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
pcarruscag: Looking at it more closely, it does not belong in the variable class.
CVariable should be a container, if it starts being used to compute physical quantities, later down the line there will be much philosophical discussion about where to implement new things.
Granted we don't always follow this advice of "just a container" (perhaps I should fix that) but I think the general idea should be:
- CVariable: Just a container, the only info it should deal with is what variables to allocate (and how).
- CSolver: Orchestration, i.e. looping over points / edges doing very generic things.
- "Models": e.g. CNumerics, CFluidModel, and other physical models we should probably have but don't (as @clarkpede was suggesting as few days ago). These are parts of the solver that can be changed to make it do different things. These models do not have to be always classes, they can be free functions (like the ones used to compute gradients), or static methods of some class (static methods are called on the class rather than on the object, basically they are functions).

For this case in particular the thing to do would be to move these functions to the solver, I think there is some precedent for that in CEulerSolver with the consistent reconstruction stuff.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -343,4 +346,88 @@ bool CNEMOEulerVariable::Cons2PrimVar(su2double *U, su2double *V,
...
+  vector<su2double> rhos, eves, cvves;
+
+  /*--- Rename density vector ---*/
+  rhos.resize(nSpecies,0.0);
+  for (unsigned short iSpecies=0; iSpecies < nSpecies; iSpecies++ ){
+    rhos[iSpecies] = V[iSpecies];
+  }
+
+  /*--- Set new fluid state ---*/
+  fluidmodel->SetTDStateRhosTTv(rhos, V[T_INDEX], V[TVE_INDEX]);
```
pcarruscag, line 422: The other reason to have this kind of function in the solver is that, when calling it in a loop you can declare the vectors outside of the loop which saves you the dynamic allocation / deallocation every time the function is called.

SU2_CFD/include/variables/CVariable.hpp
fmpmorgado: This should not be on this commit, my bad. But it will be used in future commits for CNEMONSSolver::BC_Smoluchowski_Maxwell.

pcarruscag: Ok, for when the time comes, if the function is only used in CNEMOEuler/NSSolver, it does not need to be declared in the base CVariable, just in CNEMOEuler/NSVariable.

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
WallyMaier: That makes sense, Im onboard with that logic.

EduardoMolina:
Hi @WallyMaier ! Could you please remove the unused variables in NEMO's ausm.hpp, ausmplusup2.hpp, lax.hpp and NEMO_diffusion.hpp files ? They were introduced in a previous PR.

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -199,6 +214,12 @@ class CNEMOEulerSolver : public CFVMFlowSolverBase<CNEMOEulerVariable, COMPRESSI
    */
   void SetNondimensionalization(CConfig *config, unsigned short iMesh) final;
 
+    /*!
+   * \brief Compute the pressure at the infinity.
+   * \return Value of the pressure at the infinity.
+   */
+  inline CNEMOGas* GetFluidModel(void) const final { return FluidModel;}
+
```
pcarruscag, line 222: You need to be a little careful with this because if you call this method on a CSolver pointer to a CNEMOEulerSolver instance, the return is a pointer to CFluidModel and not to CNEMOGas, which may or may not be ok.
```c++
CNEMOEulerSolver* asolver = new CNEMOEulerSolver(...);
asolver->GetFluidModel(); // <- This is a CNEMOGas*

CSolver* bsolver = asolver;
bsolver->GetFluidModel(); // <-  But this is a CFluidModel*
```

WallyMaier: I can go through and change this.  This may cause problems in the turbulence solver eventually.

WallyMaier:
@EduardoMolina most of the variables have either been deleted or removed.  I expect most will make a return in a future PR for implicit functionality within the NEMO framework

jtneedels (approved):

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: just to clarify, why are we leaving dThing/dU and Cvves out for now ? don't we need it, for example, for Roe scheme ?

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
CatarinaGarbacz: @pcarruscag any update?

CatarinaGarbacz: can.'t we simply send the pointer FluidModel we use throughout the CNEMOEUlerSolver file ?

pcarruscag: Ah sorry, I did not see Wally's message was edited (github does not notify for that).

You can access the fluid model variable directly within the solvers, or leave the code as is.
You can worry about the turbulence when you get there, a CFluidModel can be forced to CNEMOGas with a static cast, IMO that is relatively safe thing to in places where it is known that pointer can only refer to a CNEMOGas.

CatarinaGarbacz: ok so it seems that, for now, "access the fluid model variable directly within the solvers" is the simplest solution, I think this is the same that I suggested, correct ? just to double-check

CatarinaGarbacz (approved):
@Wally Maier some comments on the PR. I think it's very good work as usual. I saw you updated the regression residuals of the wedges, good call there. just before merging, could you just fix that minor issue of the GetFluidmodel - seems that the conclusion is, we don't need to send anything since FluidModel pointer is already accessible in CNEMOEulerSolver (instantiated and initialized properly)?

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: In most cases, the computation is unneccessary, but most likely will be useful in the case of the implicit formulation.  I figured it would be easy to have these here now and remove in the next if necessary

CatarinaGarbacz: in most or in all of them? asking cause that will make the difference between this being totally functional or not

CatarinaGarbacz: so is it just for implicit ? or something else?

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
WallyMaier: Sorry, I need to be a bit better about my editting....

In any case, when trying to directly access the FluidModel, there is an issue regarding static members.  I am wondering since this function is a static member, I do need to pass the fluid model in?

WallyMaier:
@pcarruscag @CatarinaGarbacz @jtneedels I have addressed the comments and made a few additional changes.  I dont want to push this in without one last look over.  Thanks for your time :)

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -956,33 +981,140 @@ void CNEMOEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_con
...
+  /*--- Set new fluid state ---*/
+  fluidmodel->SetTDStateRhosTTv(rhos, V[T_INDEX], V[TVE_INDEX]);
+
+  /*---Compute the secondary values ---*/
+  auto it = val_eves;
+  auto& ref = fluidmodel->ComputeSpeciesEve(V[TVE_INDEX]);
+  for (auto v : ref) {
+    *it = v;  ++it;
+  }
+  val_eves  = it;
```
pcarruscag, line 1030: I would put a comment here explaining you are copy from the vector to returned by the fluid model to val_eves.
Also this last line `val_eves  = it;` can be deleted.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -956,33 +981,140 @@ void CNEMOEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_con
...
+  val_gamma = fluidmodel->ComputeGamma();
+
+  return val_gamma;
+}
+
+void CNEMOEulerSolver::RecomputeConservativeVector(su2double *U, const su2double *V) {
```
pcarruscag, line 1036: void CNEMOEulerSolver::RecomputeConservativeVector(su2double *U, const su2double *V) **const** ?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -956,33 +981,140 @@ void CNEMOEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_con
...
+  FluidModel->SetTDStateRhosTTv( rhos, V[T_INDEX], V[TVE_INDEX]);
+  const auto& Energies = FluidModel->ComputeMixtureEnergies();
+
+  /*--- Set conservative energies ---*/
+  U[nSpecies+nDim]   = V[RHO_INDEX]*(Energies[0]+0.5*sqvel);
+  U[nSpecies+nDim+1] = V[RHO_INDEX]*(Energies[1]);
+
+}
+
+bool CNEMOEulerSolver::CheckNonPhys(const su2double *V) {
```
pcarruscag, line 1073: bool CNEMOEulerSolver::CheckNonPhys(const su2double *V) **const** ?

pcarruscag (approved):
Thank you for moving those functions to the solver.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
CatarinaGarbacz: @pcarruscag if we delete  `val_eves = it;` how do we have what we want in the vector val_eves that we are trying to compute?

also, in general, can you explain this procedure? lines 1027-1029 - I don't understand this advanced c++ thing and probably it would be better to know 100% what we're doing here, thanks! :)

pcarruscag: equivalent to:
```
for (int i = 0; i < ref.size(); ++i)
  val_eves[i] = ref[i];
```
but more generic in case you change the types involved.

val_eves is a pointer, i.e. a memory location, if you write val_eves = something you are changing _where_ it points to, instead of _what_ it points to (the loop is what changes the _what_).
It would be less confusing to use a consistent type across the implementation, e.g. std::vector, instead of mixing it with c-style arrays (bare pointers), because a vector can be downgraded to pointer when it needs to be passed to some legacy interface, but to go from pointer to vector you need to copy.
Even better is to write algorithms that do not depend on a specific data structures, but rather on an interface ([], *, ++ are interfaces) that is why the algorithms in the standard template library are written in terms of iterators (but as the name implies that requires templates).

CatarinaGarbacz: ok I see there's still a lot to it generically speaking, but I understand the code we have now, thanks

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Yup, done

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: done!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Done

fmpmorgado (approved):

PR #1153, opened 2021-01-05
Cleanup adaptation

pcarruscag:
## Proposed Changes
This removes some leftovers from when we deprecated the mesh adaptation.
I had a look at the "adap" branches and it seems that the new interface has completely new scripts and these are not needed, @bmunguia correct me if I am wrong.

## PR Checklist


pcarruscag:
I'll merge this since it avoids confusion for people trying to use a feature that does not exist, it is all in one commit so it should be easy enough to revert if it becomes needed.

PR #1159, opened 2021-01-11
Reduce discrete adjoint memory usage ~25%

pcarruscag:
## Proposed Changes
On the RANS onera m6 case (43k hex grid) the tape for the geometric recording becomes 13% smaller by:
- Computing element CG's together with their contributions to the dual volumes and edge normal vectors;
- Using preaccumulation "per element" instead of for each of the individual computations done for each element.

On the hybrid grid for the same geometry we have in TestCases the reduction is 32%.

Some additional memory reduction is achieved by not storing the primal element face CG's and the edge CG's, these are only needed when building dual volumes and they are easy to recompute with the data already accessed where they are currently needed.

## Related Work
#594

## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -8460,85 +8428,72 @@ void CPhysicalGeometry::ComputeMeshQualityStatistics(CConfig *config) {
...
-          Coord_FaceElem_CG[iDim] = elem[iElem]->GetFaceCG(iFace, iDim);
-          Coord_FaceiPoint[iDim]  = nodes->GetCoord(face_iPoint, iDim);
-          Coord_FacejPoint[iDim]  = nodes->GetCoord(face_jPoint, iDim);
+          Coord_Edge_CG[iDim] = 0.5 * (Coord_FaceiPoint[iDim] + Coord_FacejPoint[iDim]);
         }
 
         /*--- Access the sub-volume of the element separately in 2D or 3D. ---*/
 
         su2double Volume_i, Volume_j;
         switch (nDim) {
```
WallyMaier, line 8487: This may be a dumb question, but what motivates a switch here versus and if/else statement?

Common/src/geometry/primal_grid/CTriangle.cpp
```diff
@@ -47,19 +47,10 @@ unsigned short CTriangle::maxNodesFace = 2;
 
 CTriangle::CTriangle(unsigned long val_point_0, unsigned long val_point_1,
            unsigned long val_point_2, unsigned short val_nDim) : CPrimalGrid() {
-  unsigned short iDim, iFace, iNeighbor_Elements;
+  unsigned short iNeighbor_Elements;
 
-  /*--- Allocate CG coordinates ---*/
   nDim = val_nDim;
-  Coord_CG = new su2double[nDim];
-  for (iDim = 0; iDim < nDim; iDim++)
```
WallyMaier, line 55: Im glad this logic is being abstracted away, good work!

SU2_GEO/src/SU2_GEO.cpp
```diff
@@ -1210,40 +1205,40 @@ int main(int argc, char *argv[]) {
...
+
       }
-      
+
     }
-    
+
     if (rank == MASTER_NODE)
       Gradient_file.close();
-
```
WallyMaier, line 1224: Wow, there was a lot of extra spaces here....

WallyMaier (approved):
@pcarruscag, this looks good to me!   The code looks much cleaner, and bloated!

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: True, I'll change it there too.
To me a switch is useful to pick one out of many, it is also a good bug prevention tool for enumerator types as the compiler will warn if you forget to cover all the enum options (we convert our enum types to unsigned short which disables those warnings unfortunately).

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -132,10 +140,15 @@ void CNSSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container, C
     SetPrimitive_Gradient_LS(geometry, config);
   }
 
-  /*--- Compute the limiter in case we need it in the turbulence model or to limit the
-   *    viscous terms (check this logic with JST and 2nd order turbulence model) ---*/
```
pcarruscag, line 136: I tried to do what this comment had been asking (for quite some time), because I think it was possible to make the solver do unnecessary work e.g. compute limiters and reconstruction gradients for centered schemes.
Also there was some mixing between flow and continuous adjoint options, which looking at the adjoint solvers code seemed redundant (it never uses the flow solver limiter for example). The regressions seem to agree.

I tested the main changes of this PR on some 3D problems and all seems fine, so I'll merge this soon for the next release.

Issue #1075, opened 2020-09-18
SURFACE_STATIC_PRESSURE not recognized as objective function


MiGaff:
The objective functions SURFACE_STATIC_PRESSURE and SURFACE_TOTAL_PRESSURE do not work with the shape_optimization.py script. 

This has been verified by running the python optimization using the test case "multiobjective_wedge" in the optimization_euler test case folder, using the provided script ("inv_wedge_ROE_multiobj.cfg")

I get the following error: 

>Traceback (most recent call last):
  File "/usr/local/bin/shape_optimization.py", line 183, in <module>
    main()
  File "/usr/local/bin/shape_optimization.py", line 98, in main
    options.nzones      )
  File "/usr/local/bin/shape_optimization.py", line 155, in shape_optimization
    project = SU2.opt.Project(config,state)
  File "/usr/local/bin/SU2/opt/project.py", line 129, in __init__
    group = historyOutFields[this_obj]['GROUP']
KeyError: 'SURFACE_TOTAL_PRESSURE'

The same happens if setting these keywords for the constraints.

My environment:

OpenSuse Leap 15.1
Python 3.6.10
gcc 7.5.0
OpenMPI 1.10.7
SU2 v7.0.6

pcarruscag:
I might have fixed the issue in #1167

pcarruscag:
Give this a try if/when you get a chance and let us know.

MiGaff:
Dear Pedro,

sorry for getting back to you so late. Yes, it works now with 7.1.0.
Thank you for the fix!

pcarruscag:
No worries, glad to hear it works (I admit I did not test).

PR #1167, opened 2021-01-17
Fix objective functions (for shape_optimization.py)

pcarruscag:
## Proposed Changes
I tried to do something about the OF's that are not working with the python scripts.
If I got it right, the OF string needs to match the name used for history/screen output, at least to allow the automatically generated history map to be used.

## Related Work
#1075

## PR Checklist


Common/src/CConfig.cpp
```diff
@@ -1594,8 +1594,6 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Evaluate a problem with engines */
   addBoolOption("ENGINE", Engine, false);
 
-  /* DESCRIPTION:  Compute buffet sensor */
-  addBoolOption("BUFFET_MONITORING", Buffet_Monitoring, false);
```
pcarruscag, line 1598: I think this was cheap enough to compute by default.

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -241,11 +241,9 @@ void CNEMOCompOutput::SetHistoryOutputFields(CConfig *config){
 
   ///   /// BEGIN_GROUP: HEAT_COEFF, DESCRIPTION: Heat coefficients on all surfaces set with MARKER_MONITORING.
   /// DESCRIPTION: Total heatflux
-  AddHistoryOutput("HEATFLUX", "HF",      ScreenOutputFormat::SCIENTIFIC, "HEAT", "Total heatflux on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
+  AddHistoryOutput("TOTAL_HEATFLUX", "HF",      ScreenOutputFormat::SCIENTIFIC, "HEAT", "Total heatflux on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
   /// DESCRIPTION: Maximal heatflux
-  AddHistoryOutput("HEATFLUX_MAX", "maxHF",    ScreenOutputFormat::SCIENTIFIC, "HEAT", "Total maximum heatflux on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
-  /// DESCRIPTION: Temperature
-  AddHistoryOutput("TEMPERATURE", "Temp", ScreenOutputFormat::SCIENTIFIC, "HEAT",  "Total avg. temperature on all surfaces set with MARKER_MONITORING.", HistoryFieldType::COEFFICIENT);
```
pcarruscag, line 248: This "temperature" output was removed from everywhere except the heat and incompressible solvers, because it is a right mess. i.e. not consistent for the different solvers because for ones it depends on marker_analyse for others on marker_monitoring... Also the name TOTAL_AVG_TEMPERATURE could as well have been ?_?_TEMPERATURE, just as descriptive.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -394,87 +394,6 @@ void CDiscAdjSolver::RegisterOutput(CGeometry *geometry, CConfig *config) {
   direct_solver->GetNodes()->RegisterSolution(input, push_index);
 }
 
-void CDiscAdjSolver::RegisterObj_Func(CConfig *config) {
-
-  /*--- Here we can add new (scalar) objective functions ---*/
-  if (config->GetnObj()==1) {
-    switch (config->GetKind_ObjFunc()) {
```
pcarruscag, line 401: This does not seem to do anything anymore.

SU2_PY/SU2/io/historyMap.py
```diff
@@ -35,50 +35,35 @@
                   'GROUP': 'FLOW_COEFF',
                   'HEADER': 'Avg_Enthalpy',
                   'TYPE': 'COEFFICIENT'},
- 'AVG_MACH': {'DESCRIPTION': 'Total average mach number on all markers set in '
```
TobiKattmann, line 38: Did you change this file manually, or did you use a script to do so?
Edit: ... well automatically as it says in the description (using updateHistoryMap.py)

TobiKattmann (commented):
Thanks a lot for this cleanup 💐 
From now on I guess the history map should never be touched by hand and only via updateHistoryMap, am I right? If yes, would it be a thing to have historyMap.py removed from the repo and automatically call updateHistoryMap.py during installation process (i.e. historyMap exists only in the build folder)?
(And of course sorry for the late comments)

Common/src/grid_movement/CVolumetricMovement.cpp
```diff
@@ -353,7 +353,7 @@ void CVolumetricMovement::ComputenNonconvexElements(CGeometry *geometry, bool Sc
         nNonconvexElements++;
       }
     }
-  } else {
+  } else if (false) {
```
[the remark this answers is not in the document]
pcarruscag: Yep

SU2_CFD/include/numerics/NEMO/NEMO_diffusion.hpp
```diff
@@ -38,8 +38,8 @@
  */
 class CAvgGrad_NEMO : public CNEMONumerics {
 private:
-  unsigned short iDim, iVar;    /*!< \brief Iterators in dimension an variable. */
```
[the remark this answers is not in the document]
pcarruscag: Using clang with --warnlevel=2

pcarruscag:
It could be a thing to run it on installation, but at the same time it is also not a big deal to run it ourselves.
The script is not particularly robust IMO because it does not check for conflicting options coming from different solvers, e.g. same name output but a slightly different header on the history file, or a different history group, etc.

PR #1168, opened 2021-01-18
[WIP] NEMO - New Symmetry BC

fmpmorgado:
## Proposed Changes
*Implementation of a new boundary condition for symmetry plane instead of using Euler Wall.
 The implementation works as follows:*
 - Calculates new viscous eigenvalues for nodes in symmetry planes;
 - ~Calculates new residuals to take into account the symmetry of the mesh;~ Only corrects momentum residual vector in current version.
 - ~Uses a modified volume to calculate the solution in symmetry planes;~ Does not require modified volume in current version.

## Related Work
#657

## PR Checklist

## Results

For the sake of demonstrating the capability of the new Symmetry Boundary condition, the image shows a comparison of pressure at the surface, using the new implementation and the old symmetry boundary condition, for a flow over a 3D cylinder and considering the lateral and bottom surfaces as symmetry plane:

### New Implementation vs Old Implementation for N2 mixture (2nd Order - NEMO_NS solver - AUSM scheme)

Screenshot from 2021-01-20 23-28-31

I have run this test without changing the viscous eigenvalues, and it converges to the same value. However, for unsteady simulations, changing the eigenvalues provides better results.

### New Implementation (on NEMO) vs Current SU2 Implementation for AIR5 mixture (2nd Order - NS solver - AUSM scheme)

image


CatarinaGarbacz (approved):
the implementation seems good to me. some of the comments need to be updated since symmetry is not a boolean. other than that, awesome contribution

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -139,6 +139,9 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
 
   bool interp = config->GetSolutionInterpolation();
 
+  /* Boolean to check if nodes are in Symmetry Plane BC */
+  symmetry.resize(nPoint) = 0;
+
```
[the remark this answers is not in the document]
fmpmorgado: Done, thanks :+1:

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -520,7 +534,14 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
...
 
       Lambda   = (Lambda_1 + Lambda_2)*Area*Area/Mean_Density;
-      nodes->AddMax_Lambda_Visc(iPoint, Lambda);
+
+      if (nodes->GetSymmetry(iPoint))
+         if(nodes->GetSymmetry(iPoint) == 1 && nodes->GetSymmetry(jPoint)) nodes->AddMax_Lambda_Visc(iPoint, 4*Lambda);
+         else if (nodes->GetSymmetry(iPoint) == 2 && nodes->GetSymmetry(jPoint) == 1) nodes->AddMax_Lambda_Visc(iPoint, 8*Lambda);
+         else if (nodes->GetSymmetry(iPoint) == 2 && nodes->GetSymmetry(jPoint) == 2) nodes->AddMax_Lambda_Visc(iPoint, 16*Lambda);
+         else nodes->AddMax_Lambda_Visc(iPoint, 2*Lambda);
+      else nodes->AddMax_Lambda_Visc(iPoint, Lambda);
```
pcarruscag, line 543: Maybe you can explain this in the next dev meeting?
This looks so complex that there has to be a simpler way.

CatarinaGarbacz: also, something missing here in this if

if(nodes->GetSymmetry(iPoint) == 1 && **nodes->GetSymmetry(jPoint)**)

fmpmorgado: @pcarruscag Sure, when will the next dev meeting be ?

The idea behind it is to mimic the eigenvalues given if using a "full" domain. Probably there is a simpler way through manipulation of the Normal values, but not sure how it can be implemented.

pcarruscag: Would doubling at the end be enough?

Wednesdays at 4pm CET https://meet.jit.si/SU2_DevMeeting

fmpmorgado: If the nodes are in a single symmetry plane, yes. A bit more complex when they are in two symmetry planes, but I will look at it.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -141,39 +141,48 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
 
   /* Under-relaxation parameter. */
   LocalCFL.resize(nPoint) = su2double(0.0);
+
+  bool interp = config->GetSolutionInterpolation();
```
WallyMaier, line 145: Can we keep interp out until it will be fully integrated into the code?

WallyMaier (commented):
@fmpmorgado thanks for doing this work!  The results are very promising, I think you should add a couple of pictures here to showcase the improvement.  

I made a comment about interp, but i think looks pretty good.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: @WallyMaier this PR is to feature_NEMO, so that's not an issue at the moment

jtneedels (approved):
Hi Fabio, thanks for implementing this, it looks really good. Yeah I would second what Wally said, maybe posting the images you showed in the slack as a comment, just to show the difference between the current and former implementations.

fmpmorgado:
For the sake of demonstrating the capability of the new Symmetry Boundary condition, the image shows a comparison of the new implementation and the old symmetry boundary condition, for  a flow over a 3D cylinder and considering the lateral and bottom surfaces as symmetry plane:

WallyMaier (approved):

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -82,6 +82,8 @@ class CNEMOEulerVariable : public CVariable {
 
   su2double Tve_Freestream; /*!< \brief Freestream vib-el temperature. */
 
+  VectorType symmetry;      /*!< \brief Nodes in symmetry planes. */
```
WallyMaier, line 85: Does this need to get initialized via a "resize"?

WallyMaier (changes requested):

WallyMaier (approved):

fmpmorgado:
@WallyMaier  @jtneedels @pcarruscag Following @CatarinaGarbacz suggestion, I have decided to do a PR directly to develop. It's missing some regression tests which have SYM_MARKER. I was expecting the thermalbath to be different, but it's strange for the flatplate not to pass since I only intended to modify NEMO part. Will check on that this afternoon.

TobiKattmann:
From the discussion today at the dev-meeting: #657 this is the PR back in the day and I hope that some thoughts on that help you with your cause. The problem was especially with viscous cases and looking at the sym_plane as is in this PR I guess that problem will appear for you as well but I might be wrong here :)

Common/include/option_structure.hpp
```diff
@@ -64,12 +64,12 @@ using MapType = CEmptyMap<T,U>;
...
-  SU2_DEF = 2,  /*!< \brief Running the SU2_DEF software. */
-  SU2_DOT = 3,  /*!< \brief Running the SU2_DOT software. */
-  SU2_MSH = 4,  /*!< \brief Running the SU2_MSH software. */
-  SU2_GEO = 5,  /*!< \brief Running the SU2_GEO software. */
-  SU2_SOL = 6   /*!< \brief Running the SU2_SOL software. */
+  SU2_CFD = 1,   /*!< \brief Running the SU2_CFD software. */
+  SU2_DEF = 2,   /*!< \brief Running the SU2_DEF software. */
+  SU2_DOT = 3,   /*!< \brief Running the SU2_DOT software. */
+  SU2_MSH = 4,   /*!< \brief Running the SU2_MSH software. */
+  SU2_GEO = 5,   /*!< \brief Running the SU2_GEO software. */
```
pcarruscag, line 71: Hmmm I deleted SU2_MSH from here recently... Delete it again please, and don't add extra spaces for no reason, that tends to create merge conflicts and that is how we end up re-adding things that were removed.

Common/src/CConfig.cpp
```diff
@@ -3058,12 +3058,12 @@ void CConfig::SetHeader(unsigned short val_software) const{
...
-    case SU2_DEF: cout << "|   |___/\\___//___|   Suite (Mesh Deformation Code)                     |" << endl; break;
-    case SU2_DOT: cout << "|   |___/\\___//___|   Suite (Gradient Projection Code)                  |" << endl; break;
-    case SU2_MSH: cout << "|   |___/\\___//___|   Suite (Mesh Adaptation Code)                      |" << endl; break;
-    case SU2_GEO: cout << "|   |___/\\___//___|   Suite (Geometry Definition Code)                  |" << endl; break;
-    case SU2_SOL: cout << "|   |___/\\___//___|   Suite (Solution Exporting Code)                   |" << endl; break;
+    case SU2_CFD:    cout << "|   |___/\\___//___|   Suite (Computational Fluid Dynamics Code)         |" << endl; break;
+    case SU2_DEF:    cout << "|   |___/\\___//___|   Suite (Mesh Deformation Code)                     |" << endl; break;
+    case SU2_DOT:    cout << "|   |___/\\___//___|   Suite (Gradient Projection Code)                  |" << endl; break;
+    case SU2_MSH:    cout << "|   |___/\\___//___|   Suite (Mesh Adaptation Code)                      |" << endl; break;
```
pcarruscag, line 3064: Same here with SU2_MSH and the spaces

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -137,15 +137,16 @@ void CIntegration::Space_Integration(CGeometry *geometry,
...
       case ELECTRODE_BOUNDARY:
         solver_container[MainSolver]->BC_Electrode(geometry, solver_container, conv_bound_numerics, config, iMarker);
         break;
       case DIELEC_BOUNDARY:
         solver_container[MainSolver]->BC_Dielec(geometry, solver_container, conv_bound_numerics, config, iMarker);
         break;
+      case SYMMETRY_PLANE:
+        if (!config->GetNEMOProblem())
+          solver_container[MainSolver]->BC_Sym_Plane(geometry, solver_container, conv_bound_numerics, visc_bound_numerics, config, iMarker);
+        break;
```
pcarruscag, line 149: Why does the order of operations matter either for NEMO or the other solvers?
We don't have to play mikado with the regression tests, if the reference residuals change for some case but it converges to the same solution we update the regression.

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -559,7 +559,7 @@ void CNEMOCompOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolv
 
 void CNEMOCompOutput::LoadSurfaceData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint, unsigned short iMarker, unsigned long iVertex){
 
-  if ((config->GetKind_Solver() == NEMO_NAVIER_STOKES)) {
+  if ((config->GetKind_Solver() == NEMO_NAVIER_STOKES && config->GetMarker_All_KindBC(iMarker) != EULER && config->GetMarker_All_KindBC(iMarker) != SYMMETRY_PLANE)) {
     SetVolumeOutputValue("SKIN_FRICTION-X", iPoint, solver[FLOW_SOL]->GetCSkinFriction(iMarker, iVertex, 0));
```
pcarruscag, line 563: Do you only want to include viscous walls? There is a boolean for that in CConfig for each marker.

configure.ac
```diff
@@ -110,7 +110,7 @@ AC_ARG_ENABLE(SOL,
   [build_SOL=$enableval], [build_SOL="yes"])
 AC_ARG_ENABLE(GEO,
     AS_HELP_STRING([--disable-GEO], [build the SU2_GEO executable (default = yes)]),
-  [build_GEO=$enableval], [build_GEO="yes"])
+  [build_GEO=$enableval], [build_GEO="yes"]) 
 AC_ARG_ENABLE(PY_WRAPPER,
```
pcarruscag, line 114: Changes to this file seem unnecessary?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -450,7 +461,7 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
   }
   SU2_OMP_BARRIER
 
-  const su2double *Normal = nullptr;
+  su2double Normal[3];
```
pcarruscag, line 464: It is better to use MAXNDIM instead of hardcoding 3 here and there.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -524,7 +542,7 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
   for (iMarker = 0; iMarker < geometry->GetnMarker(); iMarker++) {
     if ((config->GetMarker_All_KindBC(iMarker) != INTERNAL_BOUNDARY) &&
         (config->GetMarker_All_KindBC(iMarker) != PERIODIC_BOUNDARY)) {
-
+    
       SU2_OMP_FOR_STAT(OMP_MIN_SIZE)
```
pcarruscag, line 546: Check if your IDE has an option to remove trailing blank spaces, there are also scripts for that purpose in the repo.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -533,9 +551,15 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
...
 
-        Normal = geometry->vertex[iMarker][iVertex]->GetNormal();
-        Area = GeometryToolbox::Norm(nDim, Normal);
+        geometry->vertex[iMarker][iVertex]->GetNormal(Normal);
+
+        /*--- Set Normal to zero if node is in the Symmetry Plane marker ---*/
+        if(config->GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE) {          
+          for(iDim = 0; iDim < nDim; iDim++)
+            Normal[iDim]=0;
+        }
```
pcarruscag, line 560: Can't you skip the marker if it is a symmetry instead?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -567,7 +593,9 @@ void CNEMOEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contai
         Lambda_1 = (4.0/3.0)*(Mean_LaminarVisc+Mean_EddyVisc);
         Lambda_2 = (Mean_ThermalCond+Mean_ThermalCond_ve)/cv;
         Lambda   = (Lambda_1 + Lambda_2)*Area*Area/Mean_Density;
-        nodes->AddMax_Lambda_Visc(iPoint,Lambda);
+
+      if(nodes->GetSymmetry(iPoint) > 0 && config->GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE) nodes->AddMax_Lambda_Visc(iPoint, Lambda/2.0);
+      else nodes->AddMax_Lambda_Visc(iPoint, Lambda);
```
pcarruscag, line 598: I might be missing something because I am only looking at the changed lines, but, if you set the Normal to zero on symmetry planes then Lambda is 0 and all this logic is redundant, right?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1873,126 +1900,36 @@ void CNEMOEulerSolver::SetNondimensionalization(CConfig *config, unsigned short
 
 void CNEMOEulerSolver::BC_Sym_Plane(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics,
                                     CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
-
-  unsigned short iDim, jDim, iSpecies, iVar, jVar;
   unsigned long iPoint, iVertex;
+  unsigned short iDim, iVar;
+  su2double Area, Normal[3], UnitNormal[3];
+  const su2double* Residual_Old;
+  su2double *Residual = new su2double[nVar];
```
pcarruscag, line 1907: MAXNDIM and if you want you can make Residual a static array of size MAXNVAR, more practical and efficient than new delete everytime.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1873,126 +1900,36 @@ void CNEMOEulerSolver::SetNondimensionalization(CConfig *config, unsigned short
...
+    Residual_Old = LinSysRes.GetBlock(iPoint);
 
-          /////// NEW //////
-          for (iDim = 0; iDim < nDim; iDim++) {
-            Jacobian_i[nSpecies+iDim][iSpecies] = dPdU[iSpecies] * UnitNormal[iDim];
-            Jacobian_i[iSpecies][nSpecies+iDim] = cs * UnitNormal[iDim];
-          }
-        }
+    for(iDim = 0; iDim < nDim; iDim++)
+      Residual[nSpecies+iDim] = Residual_Old[nSpecies+iDim]*abs(UnitNormal[iDim]);
```
pcarruscag, line 1927: The abs(UnitNormal) looks strange, do you want to reflect it about the symmetry plane?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2906,13 +2843,130 @@ void CNEMOEulerSolver::BC_Supersonic_Outlet(CGeometry *geometry, CSolver **solut
...
+        }
+
+        /*--- Integrate over the dual-grid area ---*/
+        for (iVar = 0; iVar < nVar; iVar++)
+          for (jVar = 0; jVar < nVar; jVar++)
+            Jacobian_i[iVar][jVar] = Jacobian_i[iVar][jVar] * Area;
+
+        /*--- Apply the contribution to the system ---*/
+        Jacobian.AddBlock(iPoint,iPoint,Jacobian_i);
+
```
pcarruscag, line 2953: In cases of manipulating the diagonal of the matrix you should always use AddBlock2Diag (or something like that) see CSysMatrix.hpp for the right name.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2906,13 +2843,130 @@ void CNEMOEulerSolver::BC_Supersonic_Outlet(CGeometry *geometry, CSolver **solut
...
+  Normal     = new su2double[nDim];
+  NormalArea = new su2double[nDim];
+  Jacobian_b = new su2double*[nVar];
+  DubDu      = new su2double*[nVar];
+  u          = new su2double[nDim];
+
+  for (iVar = 0; iVar < nVar; iVar++) {
+    Jacobian_b[iVar] = new su2double[nVar];
+    DubDu[iVar] = new su2double[nVar];
+  }
```
pcarruscag, line 2868: You can initialize to zero by 
  Jacobian_b[iVar] = new su2double[nVar] **()**;

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2906,13 +2843,130 @@ void CNEMOEulerSolver::BC_Supersonic_Outlet(CGeometry *geometry, CSolver **solut
...
+  rho, cs, P, rhoE, rhoEve, conc, *u, *dPdU;
+
+  bool implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+
+  /*--- Allocate arrays ---*/
+  Normal     = new su2double[nDim];
+  NormalArea = new su2double[nDim];
+  Jacobian_b = new su2double*[nVar];
+  DubDu      = new su2double*[nVar];
+  u          = new su2double[nDim];
```
pcarruscag, line 2863: Same comment about static arrays MAXNDIM MAXNVAR

pcarruscag (changes requested):
I think it is a good idea to have regular updates to develop.
There are a few things here that give me some concerns.

fmpmorgado:
After a meeting with @TobiKattmann, I have decided to change NEMO BC_EULER_WALL to match the one in CEulerSolver.cpp, where the velocity vector and gradients are corrected.

Euler boundary condition is now called inside the symmetry plane boundary condition, before performing the residual correction
This way, I believe I am assuring for all the vector quantities to be parallel to the symmetry plane and guaranteeing the condition of symmetry according to Jiri Blazek's - "Computational fluid dynamics: principles and applications" . However, calling Euler function does not appear to substantially change the the solution substantially in the test I have made so far.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
TobiKattmann: The goal here is to set the normal component of the (momentum-)residual to zero. That is why for now it is also done after all other residual contributions so that no other BC has the chance to add a normal residual. It might be unnecessary if no contribution downstream changes s.th. on that front so I would say it is a precaution until we know better.

@fmpmorgado Is it possible that this only works if the normal vector is pointing exactly in one coordinate direction e.g. (1,0,0)? I would say that `R = R - (R*n)R`  with `R`esdiual and unit-`n`ormal is the right way. In code:
```c++
auto NormalComponent= GeometryToolbox::DotProduct(nDim, &Residual_Old[nSpecies], UnitNormal);
for(iDim = 0; iDim < nDim; iDim++)
    Residual[nSpecies+iDim] = NormalComponent * Residual_Old[nSpecies+iDim];

LinSysRes.SubtractBlock(iPoint, Residual);
```
If I am not mistaken the UnitNormal points inward here but that doesn't even matter imo and just subtracting the Block should be fine.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
TobiKattmann: Yep, I saw you already removed some trailing whitespaces in #1167 for the NEMO part😉

SU2_CFD/src/integration/CIntegration.cpp
TobiKattmann: See also my answer to one of your comments below in `CNEMOEulerSolver.cpp` .

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2906,13 +2845,294 @@ void CNEMOEulerSolver::BC_Supersonic_Outlet(CGeometry *geometry, CSolver **solut
...
+
+      /*--- Update residual value ---*/
+      LinSysRes.AddBlock(iPoint, residual);
+
+      /*--- Jacobian contribution for implicit integration. ---*/
+      if (implicit) {
+        Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
+      }
+
+      if (viscous) {
```
TobiKattmann, line 2999: The viscous part is superflous. (unless you need a slip wall I guess)

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2906,13 +2845,294 @@ void CNEMOEulerSolver::BC_Supersonic_Outlet(CGeometry *geometry, CSolver **solut
...
+
+  /*--- Allocation of primitive gradient arrays for viscous fluxes. ---*/
+  su2double** Grad_Reflected = new su2double*[nPrimVarGrad];
+  for (iVar = 0; iVar < nPrimVarGrad; iVar++) Grad_Reflected[iVar] = new su2double[nDim];
+
+  /*--- Loop over all the vertices on this boundary marker. ---*/
+
+  SU2_OMP_FOR_DYN(OMP_MIN_SIZE)
+  for (iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+    if (!preprocessed || geometry->bound_is_straight[val_marker] != true) {
```
TobiKattmann, line 2875: If you delete the viscous part then the tangential-vector computation here is not necessary as well and the section can be simplified. 
The bound_is_straight container + evaluation could be deleted in that case as well (unless we want to keep for future ideas.)

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1873,126 +1900,38 @@ void CNEMOEulerSolver::SetNondimensionalization(CConfig *config, unsigned short
...
   unsigned long iPoint, iVertex;
+  unsigned short iDim, iVar;
+  su2double Area, Normal[3], UnitNormal[3];
+  const su2double* Residual_Old;
+  su2double *Residual = new su2double[nVar];
 
-  su2double *Normal = nullptr, Area, UnitNormal[3], *NormalArea,
-  **Jacobian_b, **DubDu,
-  rho, cs, P, rhoE, rhoEve, conc, *u, *dPdU;
+  BC_Euler_Wall(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
```
TobiKattmann, line 1909: > Euler boundary condition is now called inside the symmetry plane boundary condition, before performing the residual correction
This way, I believe I am assuring for all the vector quantities to be parallel to the symmetry plane and guaranteeing the condition of symmetry according to Jiri Blazek's - "Computational fluid dynamics: principles and applications" . However, calling Euler function does not appear to substantially change the the solution substantially in the test I have made so far.

Imo this is not necessary and that part can be deleted (until Jiri's suggested implementation shows problems).

TobiKattmann (changes requested):
@fmpmorgado and I chatted and decided to try out the Version Blazek describes in his book (8.6 in the 3rd edition). 

That will also require changing some normal vectors of faces that end on a sym plane (tbd once after mesh reading) and removing components of the gradient for symmetry nodes (tbd each time they get computed). This is not done yet in this PR. 

The changes are straight forward once these two mentioned things are done as the sym_plane bc becomes rather short as it just removes the normal residual component (and nothing else). And the Euler wall stays as is so the code needs to be moved over (i.e. again separate implementations for Sym and euler wall). Keeping the viscous stuff has the potential to serve as a slip wall for viscuos flows (if I am not mistaken). 

Putting changes to the non-NEMO solvers in this PR is s.th. we could do as the normal-vector and gradient changes should be equal but it would delay the whole PR and a lot of Reg-tests would have changes. **AND** we have to see whether we have problems in the non-NEMO solvers that would be fixed with this in the first place (correct me if we already know that -> I will test that).

Together with @talbring (thanks!) I briefly looked into the [code that Jiri Blazek provides for his book](https://booksite.elsevier.com/9780080999951/examplecodes.php) and he implements the boundary he describes (... who would have thought).

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
pcarruscag: The IDE removed them for me, when I fixed the compiler warnings.

pcarruscag:
> Putting changes to the non-NEMO solvers in this PR is s.th. we could do as the normal-vector and gradient changes should be equal but it would delay the whole PR and a lot of Reg-tests would have changes

They are exactly the same type of solver, we should not have two types of symmetry.

Common/include/option_structure.hpp
fmpmorgado: @pcarruscag SU2_MSH is still there, on the last commit of develop

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: Nicely spotted. Changing it

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: @pcarruscag You are right, already corrected it.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: I removed this Euler_Wall_BC and replaced it with the one in **CFVMFlowSolverBase.inl**

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: Yes, you are right, I have removed the viscous part.

Common/include/option_structure.hpp
pcarruscag: Because someone put it back, probably because of SU2 interp, the code itself does not exist.

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -559,7 +559,7 @@ void CNEMOCompOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolv
 
 void CNEMOCompOutput::LoadSurfaceData(CConfig *config, CGeometry *geometry, CSolver **solver, unsigned long iPoint, unsigned short iMarker, unsigned long iVertex){
 
-  if ((config->GetKind_Solver() == NEMO_NAVIER_STOKES)) {
+  if (config->GetKind_Solver() == NEMO_NAVIER_STOKES && config->GetViscous_Wall(iMarker)) {
```
pcarruscag, line 562: I would also remove the check for NEMO_NAVIER_STOKES... This kind of logic causes 90% of the bugs in SU2 every time a new options is added something breaks, the simpler the logic, the easier it is to maintain the code.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -262,6 +264,58 @@ CNEMOEulerSolver::CNEMOEulerSolver(CGeometry *geometry, CConfig *config,
     }
   }
 
+
+  /*--- Correct normal directions of edges ---*/
+  for (unsigned long iMarker = 0; iMarker < geometry->GetnMarker(); iMarker++) {
+    if (config->GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE){
```
pcarruscag, line 269: We should probably move this to draft PR while you guys prototype the new symmetry implementation.
Then we can discuss how and where to implement the different changes without copy-pasting code and making sure the fix is compatible with other features of the code.

Common/include/option_structure.hpp
pcarruscag: My bad, I removed it from some python files.

PR #1162, opened 2021-01-12
Addition of Source Terms for NEMO Axisymmetric Flows

jtneedels:
## Proposed Changes
 

- Additional source terms are implemented in the residual computation for NEMO planar axisymmetric flows. This allows inviscid and viscous axisymmetric simulations. 

## Related Work

- Built upon framework in PR #1106 by FlorianDM, which introduced viscous axisymmetric source terms for non-reacting flows.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1170, opened 2021-01-18
Restructure source residual computation to fix axisymmetric chemsitry/vib source computation

jtneedels:
## Proposed Changes

- This PR addresses an issue I found in my previous implementation where the chemistry source residual was being computed incorrectly when the axisymmetric source residual was called. 
- This implementation fixes that issue, caused by using outer and inner loops. 
- This PR eliminates redundant set functions, and restructures the code blocks to minimize the number of loops. 

## Related Work

- Fixes bug found in PR #1162 

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
@@ -1146,7 +1146,7 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
     numerics->SetVolume(geometry->nodes->GetVolume(iPoint));
     numerics->SetCoord(geometry->nodes->GetCoord(iPoint),
                        geometry->nodes->GetCoord(iPoint) );
-
+  }
     /*--- Compute axisymmetric source terms (if needed) ---*/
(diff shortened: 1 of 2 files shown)


SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1180,18 +1231,6 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
       SU2_OMP_FOR_DYN(omp_chunk_size)
       for (iPoint = 0; iPoint < nPointDomain; iPoint++) {
 
-        /*--- Set solution  ---*/
-        numerics->SetConservative(nodes->GetSolution(iPoint), nodes->GetSolution(iPoint));
-
-        /*--- Set control volume ---*/
```
WallyMaier, line 1186: Nice not having to recompute these.

WallyMaier (commented):
@jtneedels thanks for the quick fix of this problem!  I think it would be nice to include some results, but otherwise, it looks good to me!

pcarruscag (approved):
Since this is a fix let's merge it for the release (later today).

PR #1174, opened 2021-01-22
Velocity transfer at fluid-structure interface

cvencro:
## Proposed Changes

This PR enforces the continuity of velocity explicitly at the fluid structure interface for each time step. The velocity was previously recalculated for the fluid nodes based on displacement history of the nodes. This is still needed for fluid only problems with deforming meshes. However, for FSI problems, the velocity at the interface can be transferred from the structural solution and propagated to the fluid nodes similarly to the displacement which ensures exact agreement at the interface at each time-step. This change has been validated using a typical FSI benchmark of square cylinder with a flexible cantilever beam, which gives good agreement in peak tip displacement and Strouhal number.
 

## PR Checklist


SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -2792,6 +2812,19 @@ void CFEASolver::PredictStruct_Displacement(CGeometry *geometry, CConfig *config
...
+  /*--- To nPointDomain: we need to communicate the predicted solution after setting it. ---*/
+  SU2_OMP_PARALLEL_(for schedule(static,omp_chunk_size))
+  for (unsigned long iPoint=0; iPoint < nPointDomain; iPoint++) {
+    nodes->SetSolution_Vel_Pred(iPoint);
+  }
+
+  InitiateComms(geometry, config, SOLUTION_VEL_PRED);
+  CompleteComms(geometry, config, SOLUTION_VEL_PRED);
+
+}
```
pcarruscag, line 2826: What is the implication of using a higher order predictor for the displacement and a 0 order (possibly inconsistent) for the velocity?

If there is no implication we can avoid all the Set/GetSolution_Vel_Pred and use the velocity directly.

pcarruscag: Can you try running the loop to nPoint and not doing the communications? (in principle it should work)

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -564,6 +567,43 @@ void CMeshSolver::UpdateDualGrid(CGeometry *geometry, CConfig *config){
...
+  }
+
+  /*--- Impose boundary conditions including bouundary velocity ---*/
+  SetBoundaryVelocities(geometry[MESH_0], numerics[FEA_TERM], config);
+
+  /*--- Solve the linear system. ---*/
+  Solve_System(geometry[MESH_0], config);
+  for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+    for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+      su2double val_vel = LinSysSol(iPoint, iDim);
```
pcarruscag, line 593: It should also be valid to run this loop to nPoint and not communicate.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -707,6 +747,59 @@ void CMeshSolver::SetBoundaryDisplacements(CGeometry *geometry, CNumerics *numer
 
 }
 
+void CMeshSolver::SetBoundaryVelocities(CGeometry *geometry, CNumerics *numerics, CConfig *config){
+
+  unsigned short iMarker;
+
+  /*--- Impose zero displacements of all non-moving surfaces (also at nodes in multiple moving/non-moving boundaries). ---*/
+  /*--- Exceptions: symmetry plane, the receive boundaries and periodic boundaries should get a different treatment. ---*/
+  for (iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
```
pcarruscag, line 756: Maybe we can give 2 modes to SetBoundaryDisplacement (displacement and velocity) to avoid duplicating the logic of what to do with each marker (which is not the same anymore btw).

SU2_CFD/src/solvers/CFEASolver.cpp
cvencro: Thanks for the review Pedro. I'm not sure about the mix of predictor orders. I've only used and tested with zero order displacement and velocity. I know there is more code to create the set/get calls, but I followed the same procedure as the displacement so that it can be adapted if needed in the future.

The change to nPoint worked and I could remove the additional communication statements here.

SU2_CFD/src/solvers/CMeshSolver.cpp
cvencro: Yes, I could switch to nPoint and remove the communication calls.

SU2_CFD/src/solvers/CMeshSolver.cpp
cvencro: Yes I think this would be better. For the velocity boundary, I just needed to set the boundary constraints for all the markers in the same way as that used for the displacement boundary but then switch BC_Deforming to BC_Velocity. So I've created a boolean flag now to switch between the two within SetBoundaryDisplacement.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -564,6 +567,39 @@ void CMeshSolver::UpdateDualGrid(CGeometry *geometry, CConfig *config){
...
+  const bool wasActive = AD::BeginPassive();
+
+  Compute_StiffMatrix(geometry[MESH_0], numerics, config);
+
+  AD::EndPassive(wasActive);
+
+  /*--- Clear residual (loses AD info), we do not want an incremental solution. ---*/
+  SU2_OMP_PARALLEL {
+    LinSysRes.SetValZero();
+    LinSysSol.SetValZero();
```
pcarruscag, line 583: In DeformMesh we should a LinSysSol.SetValZero(); for cases with grid velocity otherwise the deformation will use the velocity solution as initial conditions.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1636,12 +1636,19 @@ void CSolver::GetCommCountAndType(const CConfig* config,
       COUNT_PER_POINT  = nVar*3;
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
+    case SOLUTION_VEL_PRED:
+      COUNT_PER_POINT  = nVar;
+      MPI_TYPE         = COMM_TYPE_DOUBLE;
+      break;
```
pcarruscag, line 1642: Good to know this is not needed, makes things simpler, remove the SOLUTION_VEL_PRED please, someday I'll try to avoid also SOLUTION_PRED.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -661,7 +697,8 @@ void CMeshSolver::SetBoundaryDisplacements(CGeometry *geometry, CNumerics *numer
     if ((config->GetMarker_All_Deform_Mesh(iMarker) == YES) ||
         (config->GetMarker_All_Moving(iMarker) == YES)) {
 
-      BC_Deforming(geometry, numerics, config, iMarker);
+      if (!velocity_transfer) BC_Deforming(geometry, numerics, config, iMarker);
```
pcarruscag, line 700: Should this boolean be used to prevent the section with GetSurface_Movement(DEFORMING) from being applied when there is velocity?

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -985,6 +985,18 @@ class CSolver {
...
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
+  inline virtual void BC_Velocity(CGeometry *geometry,
+                                   CNumerics *numerics,
```
pcarruscag, line 996: Instead of this can we pass a boolean to BC deforming instead? Just because BC_velocity sounds too much like something a flow solver could have.

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -454,16 +454,23 @@ void CMultizoneDriver::Update() {
     for (auto jZone = 0u; jZone < nZone; jZone++){
       /*--- The target zone is iZone ---*/
       if (jZone != iZone){
+        /*--- Run update on structural solution to update relaxation on grid velocities before transferring data ---*/
+        if (config_container[jZone]->GetStructuralProblem()) {
+          iteration_container[jZone][INST_0]->Update(output_container[jZone], integration_container, geometry_container,
+              solver_container, numerics_container, config_container,
```
pcarruscag, line 460: Would running the Update for all zones before the transfer work?
Having exceptions for some types of problems does not look so nice IMO. And like this there is a change the update of j zone will be called for multiple i zones.

SU2_CFD/src/solvers/CFEASolver.cpp
pcarruscag: Alright let's leave it for now, good to have options.

SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -35,7 +35,7 @@ CFlowCompOutput::CFlowCompOutput(CConfig *config, unsigned short nDim) : CFlowOu
 
   turb_model = config->GetKind_Turb_Model();
   lastInnerIter = curInnerIter;
-  gridMovement = config->GetGrid_Movement();
+  gridMovement = (config->GetGrid_Movement() || config->GetDynamic_Grid());
```
pcarruscag, line 38: Just noticed in another PR that for this you only need Dynamic_Grid as it already contains Grid_Movement

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: Done

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: This should be the reason for one testcase now diverging.

pcarruscag:
I was looking a bit more into this because of the issue with having to treat the structural solver differently in the Update function.
(I gotta say I feel lucky I never had to look into time domain stuff, because 3 nested iterations... not easy).

Anyway, I think there is a problem here, the "predicted displacement" that is transferred to the fluid during outer iterations (the coupling iterations within each time iteration) is the one set in CFEASolver::SetAitken_Relaxation, which is part of the Corrector step at the end of the structural inner iterations (the ones within each outer iteration).

The "predicted velocity" is not set in SetAitken_Relaxation thus the grid velocities are lagged by one time step, since they are only set to the current velocity of the structure at the end of the time step (in the Update methods).

I have little trust in the predictor step, looking at it I thing only zero order has a chance of working and anything else will cause the fluid and structural sides to separate / overlap, I'll test the theory.

pcarruscag:
The predictor thing is fine, the fluid displacement may start different from the structural but this is corrected in the outer iterations.

I think the best solution for the velocity issue (if there is one, but it really looks like there is) would be to include ImplicitNewmark_Relaxation in the corrector step (this sets the solution to the relaxed solution and recomputes velocity and acceleration) and to transfer velocity directly, i.e. without any prediction.

cvencro:
Hi Pedro, thanks for looking into this more. I had run into inconsistencies for FSI problems with relaxation which was the reason for the domain specific calls coming into the multizone driver Update function. With the changes I introduced, the velocity at the interface was being transferred correctly with and without relaxation, so I left it there but it is a bit messy. I'll test with the modified calls for the Relaxation as you suggest. I completely agree that if we can simplify the velocity transfer by just using the velocity directly, then we should. Especially since the predicted velocity is only zero order anyway.

Nicola-Fonzi:
Dear @cvencro,
Great upgrade of the interface treatment! 
I was wondering what happens in case the structural solver is not the internal one, but rather an external solver coupled via Python. I think, in that case, the BC_velocity cannot be applied as there is information missing (we usually only transfer the displacement via python).
The solution to this, with minimum modifications, would be to slightly change the result of GetFSI_simulation. Indeed, at the moment, it checks if there is a marker_fluid_load or if there are both the structural solver and the fluid solver.
I was thinking that we may leave only the second check (i.e. if there are both the structural solver and the fluid solver), so that we avoid any problem. Do you think this is possible? Why did we actually need to check also the marker_fluid_load?

rafapalacios:
@Nicola-Fonzi, you may as well want to consider importing as well the structural velocities in the interface from your external solver. That information is always available and gives a consistent definition of the displacement and velocity on both the grid motions and the interface for dynamic FSI problems -- not to mention simplifying the evaluation of the mesh adjoints.

cvencro:
Hi @Nicola-Fonzi,

I'm not 100% sure of the reasons behind the GetFSI_Simulation criteria. I'd imagine that the nMarker_Fluid_Load might have been included there to make sure that a problem is treated as FSI only if there is transfer of load from the fluid to the structural domain, even if both zones existed independently in the config. But since the GetFSI_Simulation check is set up as an "or" rather than "and", as long as one of them is satisfied, the logic will be true for an FSI simulation, which might be sufficient for your application?

For the velocity transfer, I agree with Rafa that it would be better if you could also transfer the velocity from the external structural solver. If an external structural solver is used for dynamic analysis, it probably can output the structural velocities as well as the structural displacements? I left the methods for the recalculation of the grid velocity using finite differences in the code to support fluid-only problems with dynamic grid motion, I wouldn't suggest this as the method for FSI problems. For primal analysis, results from both methods agree with FSI benchmark cases but if you are interested in adjoint analysis at any point, transferring the velocity information from the structural domain directly to the fluid domain gives a significant simplification and improvement to the gradients.

pcarruscag:
Marker fluid load is to allow the native structural solver to receive fluid loads from python, and to extract adjoints again from python.

Similarly, it allows extracting fluid loads from the flow solver via python, and its adjoints to be set, again from python.

pcarruscag:
While I believe the velocity transfer is the way to go for adjoints and accuracy especially in incompressible flow, it should be possible to cater to both needs by basing the flow solver flow traction malarkey on Marker deform mesh, instead of marker fluid load.
Which would possibly make the setup easier as well.
If we want the flexibility to have fluid load markers independent of mesh deformation deformation, we may need a "yet another config knob".
One aspect to consider is that the displacement coming from python might be a forced motion, and so the consistency argument no longer matters.
Perhaps what we need, is a way to tell for which boundaries the velocity is known or not.
If the velocity is not known, it could be computed by finite differences at the interface and made part of the velocity-elasticity problem.
This would allow the forced motions to work together with FSI.
If the velocity is not known for any boundary, it is orders of magnitude quicker to compute it by finite differences.

Sorry for the soliloquy, but I think all y'all get what I'm sayin'

Nicola-Fonzi:
Dear all,
Thank you very much for your comments!
I agree with you, transferring the velocity is definitely the best solution in the long run. However, as Pedro mentioned, there is also the case when a certain imposed deformation is set, without solving any structural system. This is the case of a forced motion.

In the python script configuration files it is required to set the marker_fluid_load as this is the only way to make SU2 compute the tractions at that marker. Maybe, a solution to both needs, can be to make SU2 compute the tractions anyway for the wall boundary. 

In this way, if you have the velocity at the interface, you can define this interface as also marker_fluid_load, the FSI bool will be true and the new BC_velocity will be used. If you do not have the velocity at the interface, you will avoid marker_fluid_load and SU2 will not fail trying to use bc_velocity.

This will also give me more time to develop, from the python side, the required modifications to use the new interface.

Please let me know your comments

pcarruscag:
See #1197, it should be straightforward regarding the tractions, but the FSI boolean has implications in other areas of the fluid solver. I need help testing as I don't feel like deep diving into the python guts.

Nicola-Fonzi:
Thank you Pedro, I will then move on that branch and double check everything from the python side!

SU2_CFD/src/solvers/CMeshSolver.cpp
cvencro: Agreed. I've added the velocity boolean to prevent this since currently there is no support for dynamic FSI problems with additional enforced surface motion.

SU2_CFD/include/solvers/CSolver.hpp
cvencro: I've updated this to only have BC_Deforming with a boolean to handle velocity.

SU2_CFD/src/drivers/CMultizoneDriver.cpp
cvencro: When doing the initial implementation, I needed to separate the structural problem as above to update the structural solutions correctly when using relaxation. I've tested with running the update for both zones before the transfer which seems to work fine as well, so I've moved the Update call for all zones.

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -452,16 +452,15 @@ void CMultizoneDriver::Update() {
     for (auto jZone = 0u; jZone < nZone; jZone++){
       /*--- The target zone is iZone ---*/
       if (jZone != iZone){
+        iteration_container[jZone][INST_0]->Update(output_container[jZone], integration_container, geometry_container,
+            solver_container, numerics_container, config_container,
```
pcarruscag, line 456: For cases with more than 2 zones this will cause multiple "Updates" of each zone.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -866,11 +906,24 @@ void CMeshSolver::Restart_OldGeometry(CGeometry *geometry, CConfig *config) {
...
+      for (iPoint_Global = 0; iPoint_Global < geometry->GetGlobal_nPointDomain(); iPoint_Global++) {
+
+        auto iPoint_Local = geometry->GetGlobal_to_Local_Point(iPoint_Global);
+
+        if (iPoint_Local >= 0) {
+
+          for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+            if(iStep==1)
+              nodes->Set_Solution_time_n(iPoint_Local, iDim, 0.0);
+            else
```
pcarruscag, line 922: To set to zero you can simply loop over nPoint directly.
The `if(iStep...` should be outside of the loops that set to zero.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -502,11 +502,12 @@ void CMeshSolver::DeformMesh(CGeometry **geometry, CNumerics **numerics, CConfig
   /*--- Clear residual (loses AD info), we do not want an incremental solution. ---*/
   SU2_OMP_PARALLEL {
     LinSysRes.SetValZero();
+    LinSysSol.SetValZero();
```
pcarruscag, line 505: Only if time_domain and FSI please, otherwise this is very inefficiency for steady problems where the linear system gets progressively cheaper to solve.

Common/include/option_structure.hpp
```diff
@@ -2160,7 +2160,8 @@ enum MPI_QUANTITIES {
   MESH_DISPLACEMENTS   ,  /*!< \brief Mesh displacements at the interface. */
   SOLUTION_TIME_N      ,  /*!< \brief Solution at time n. */
   SOLUTION_TIME_N1     ,  /*!< \brief Solution at time n-1. */
-  PRIMITIVE               /*!< \brief Primitive solution communication. */
+  PRIMITIVE            ,  /*!< \brief Primitive solution communication. */
+  SOLUTION_VEL_PRED       /*!< \brief Velocity solution predicted communication. */
```
pcarruscag, line 2164:
```suggestion
```
I don't think either of these is used anymore

SU2_CFD/src/drivers/CMultizoneDriver.cpp
cvencro: I didn't need to change the Update call, I've reverted this back now.

SU2_CFD/src/solvers/CMeshSolver.cpp
cvencro: I've updated to loop over nPoint now. The set calls to time_n and time_n1 still depend on which step was being considered, so I've left that the same. If I've misunderstood the change you meant, please let me know or suggest the change which you think is more efficient.

SU2_CFD/src/solvers/CMeshSolver.cpp
cvencro: Yes, updated

cvencro:
Thanks for the review Pedro. Most of the regression tests pass now as well. The dynamic FSI reg tests were updated for the serial and parallel tests since the new velocity bc brings about a change for them inherently. The hybrid regression tests haven't been changed yet, but they are diverging in the tests to begin with. It is very likely that some specific calls for the hybrid implementation might be missing. I tried to follow the existing implementation style but I'm not sure what might be missing/wrong. Can you suggest anything for me to try or look for?

SU2_CFD/src/solvers/CMeshSolver.cpp
pcarruscag: Thanks that should be fine.

pcarruscag:
No problem :) the hybrid parallel should be fixed (it was a parallel section inside another).
Have you had a chance to look into the other issue I think I saw? Namely the structural velocity being transferred is not updated once per outer iteration, but rather once per time step which makes it lag (that is what it looks but I might be wrong, I explained it better on my Feb 7 comment).

cvencro:
Hi Pedro, sorry I'd forgotten to get to that one. I've just had a look through the logic again. Inside SetAiken_Relaxation, the predicted velocity is set which I think should keep the velocity update to be within the same time iteration, but if you think this still isn't actually the case let me know. Also, I've just spotted that inside Update in CFEAIteration, the logic for dynamic and/or fsi isn't complementary, so the separate "else if" handling for ImplicitNewmark_Relaxation isn't actually called at all. The SetDualTime_Solver which is called earlier in the same function in turn calls ImplicitNewmark_Relaxation so this hadn't been spotted earlier. I've removed the redundant else if statement...

cvencro:
> No problem :) the hybrid parallel should be fixed (it was a parallel section inside another).
> Have you had a chance to look into the other issue I think I saw? Namely the structural velocity being transferred is not updated once per outer iteration, but rather once per time step which makes it lag (that is what it looks but I might be wrong, I explained it better on my Feb 7 comment).

Thanks for the fix for the hybrid parallel reg test! The dynamic fsi test cases were the only remaining ones which was expected, this has been updated now.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -574,6 +579,47 @@ void CMeshSolver::UpdateDualGrid(CGeometry *geometry, CConfig *config){
 
 }
 
+void CMeshSolver::ComputeGridVelocity_FromBoundary(CGeometry **geometry, CNumerics **numerics, CConfig *config){
+
+  if (config->GetnZone() == 1)
+    SU2_MPI::Error("It is not possible to compute grid velocity from boundary velocity for single zone problems.\n"
+                   "MARKER_FLUID_LOAD should only be used for structural boundaries.", CURRENT_FUNCTION);
```
pcarruscag, line 586: @Nicola-Fonzi this should prevent this type of grid velocity method from being used when the python wrapper is used.

pcarruscag:
Oh right... It was there all along and I missed it. Hope you don't mind I made that consistent and merged the velocity prediction with the displacement prediction.
And so everything LGTM.

pcarruscag (approved):

pcarruscag:
:+1: shall we merge this PR then?

cvencro:
Yes, nothing more from me on this. Once this is merged, the comparison in PR #1260 should become simpler too. I've just updated this branch with develop, we can merge once the regression tests pass again.

PR #1176, opened 2021-01-22
Structural stress penalty function for optimization, fix some CConfig problems

pcarruscag:
## Proposed Changes
A differentiable maximum VM stress function for optimization.
Simplifies the "DoubleArray" option to avoid one array of default values and another of actual values.

## PR Checklist


Common/src/grid_movement/CSurfaceMovement.cpp
```diff
@@ -3256,7 +3256,7 @@ void CSurfaceMovement::AeroelasticDeform(CGeometry *geometry, CConfig *config, u
   if (config->GetKind_GridMovement() == AEROELASTIC_RIGID_MOTION) {
     su2double Omega, dt, psi;
     dt = config->GetDelta_UnstTimeND();
-    Omega  = (config->GetRotation_Rate(3)/config->GetOmega_Ref());
+    Omega = config->GetRotation_Rate(2)/config->GetOmega_Ref();
```
pcarruscag, line 3259: Using static types lets the compiler find bugs for us at compile time :)
Using bare-pointer does not let the compiler do anything :(

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -559,14 +559,14 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
   SetHistoryOutputValue("SURFACE_STATIC_PRESSURE", Tot_Surface_Pressure);
   SetHistoryOutputValue("AVG_DENSITY", Tot_Surface_Density);
   SetHistoryOutputValue("AVG_ENTHALPY", Tot_Surface_Enthalpy);
-  SetHistoryOutputValue("AVG_NORMALVEL", Tot_Surface_Enthalpy);
+  SetHistoryOutputValue("AVG_NORMALVEL", Tot_Surface_NormalVelocity);
```
pcarruscag, line 562: Thanks @WallyMaier , I literally changed the line below 2 days ago and missed this.

Common/src/CConfig.cpp
```diff
@@ -1262,8 +1239,8 @@ void CConfig::SetConfig_Options() {
   /*!\brief INC_DENSITY_INIT \n DESCRIPTION: Initial density for incompressible flows (1.2886 kg/m^3 by default) \ingroup Config*/
   addDoubleOption("INC_DENSITY_INIT", Inc_Density_Init, 1.2886);
   /*!\brief INC_VELOCITY_INIT \n DESCRIPTION: Initial velocity for incompressible flows (1.0,0,0 m/s by default) \ingroup Config*/
-  default_vel_inf[0] = 1.0; default_vel_inf[1] = 0.0; default_vel_inf[2] = 0.0;
```
WallyMaier, line 1265: I like the removal of all the default_

WallyMaier (approved):
This looks good to me.  I appreciate the clean-up done.  Especially getting rid of SU2_MSH!

TestCases/radiation/p1model/configp1.cfg
```diff
@@ -23,6 +23,7 @@ INC_DENSITY_MODEL= VARIABLE
 INC_ENERGY_EQUATION = YES
 INC_DENSITY_INIT= 0.00597782417156
 INC_TEMPERATURE_INIT= 288.15
+INC_VELOCITY_INIT= (0, 0, 0)
```
TobiKattmann, line 26: Why do you have to add this option here? I might have missed that in the code but if the option is not set I understood your changes that the default remains the same, but with out explicitly having containers for the default value? (default here would be 1,0,0)

Common/src/grid_movement/CSurfaceMovement.cpp
TobiKattmann: I am curious how the compiler found that now that the underlying `Rotation_Rate` is static because is has 3 elements initialized to 0. So how did the compiler know but I guess I miss sth here

TobiKattmann (commented):
Can't say much to the FEA stuff but thanks 💐  for the Config cleanup. Yet another net-code-reduction :)

Common/include/CConfig.hpp
```diff
@@ -117,6 +109,7 @@ class CConfig {
   su2double Opt_RelaxFactor;              /*!< \brief Scale factor for the line search. */
   su2double Opt_LineSearch_Bound;         /*!< \brief Bounds for the line search. */
   su2double StartTime;
+  unsigned short SmoothNumGrid;           /*!< \brief Smooth the numerical grid. */
```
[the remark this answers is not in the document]
pcarruscag: Left over from SU2_MSH I believe, there is some geometry routine that uses those variables of CPoint, but it is never called.

Common/src/CConfig.cpp
```diff
@@ -2022,23 +1994,22 @@ void CConfig::SetConfig_Options() {
   addStringListOption("MARKER_MOVING", nMarker_Moving, Marker_Moving);
   /* DESCRIPTION: Mach number (non-dimensional, based on the mesh velocity and freestream vals.) */
   addDoubleOption("MACH_MOTION", Mach_Motion, 0.0);
-  default_vel_inf[0] = 0.0; default_vel_inf[1] = 0.0; default_vel_inf[2] = 0.0;
   /* DESCRIPTION: Coordinates of the rigid motion origin */
-  addDoubleArrayOption("MOTION_ORIGIN", 3, Motion_Origin, default_vel_inf);
+  addDoubleArrayOption("MOTION_ORIGIN", 3, Motion_Origin);
```
pcarruscag, line 1998: @TobiKattmann the default velocity was not 1,0,0 because it was being re-used here as the default value for something else.

This trips a lot of people (myself included) but the actual parsing of the options does not take place in this method, it creates a list of option parsing objects but it does not read immediately. Hence the need for the config postprocessing method to check for conflicting things and so on.

Common/src/CConfig.cpp
```diff
@@ -1262,8 +1239,8 @@ void CConfig::SetConfig_Options() {
   /*!\brief INC_DENSITY_INIT \n DESCRIPTION: Initial density for incompressible flows (1.2886 kg/m^3 by default) \ingroup Config*/
   addDoubleOption("INC_DENSITY_INIT", Inc_Density_Init, 1.2886);
   /*!\brief INC_VELOCITY_INIT \n DESCRIPTION: Initial velocity for incompressible flows (1.0,0,0 m/s by default) \ingroup Config*/
-  default_vel_inf[0] = 1.0; default_vel_inf[1] = 0.0; default_vel_inf[2] = 0.0;
-  addDoubleArrayOption("INC_VELOCITY_INIT", 3, Inc_Velocity_Init, default_vel_inf);
+  vel_init[0] = 1.0; vel_init[1] = 0.0; vel_init[2] = 0.0;
+  addDoubleArrayOption("INC_VELOCITY_INIT", 3, vel_init);
```
TobiKattmann, line 1243: @pcarruscag Ok I see. This really seems so straight forward with what it replaces but it isn't

Common/src/grid_movement/CSurfaceMovement.cpp
TobiKattmann: Oh man am I blind... *insert palmface here*

pcarruscag: As blind an any programmer, we should have XYZ enums

rafapalacios:
Got my first glasses half a year into Fortran… and never looked back.

PR #1177, opened 2021-01-23
Reduce flow solver triplication (comp, incomp, nemo)

pcarruscag:
## Proposed Changes
Move more routines to the FVM flow solver base class.
At the risk of sounding like a broken record, one function means:
One test, one optimization, one debug, one fix.
Implementing new features by copy-paste-modify is like eating your legs to get to the end of a marathon, it ain't gonna work in the long run xD
Cut, Paste, Abstract, Survive

## Related Work
#1044

## PR Checklist


SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -244,6 +247,257 @@ class CFVMFlowSolverBase : public CSolver {
...
+   * \param[in] iMesh - Index of the mesh in multigrid computations.
+   * \param[in] Iteration - Value of the current iteration.
+   * \tparam SoundSpeedFunc - Function object to compute speed of sound
+   * \tparam LambdaViscFunc - Function object to compute the viscous lambda
+   * \note Both functors need to implement (nodes,iPoint,jPoint) for edges, and (nodes,iPoint) for vertices.
+   */
+  template<class SoundSpeedFunc, class LambdaViscFunc>
+  FORCEINLINE void SetTime_Step_impl(const SoundSpeedFunc& soundSpeed,
+                                     const LambdaViscFunc& lambdaVisc,
+                                     CGeometry *geometry,
```
pcarruscag, line 264: The idea is to generalize this type of function, the details of the implementation are provided by the derived solver when it instantiates (calls) the generic implementation.
There is no explicit "if incompressible" or "if nemo" or "if compressible" logic.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2330,260 +2330,50 @@ unsigned long CEulerSolver::SetPrimitive_Variables(CSolver **solver_container, C
...
-        Lambda = (Lambda_1 + Lambda_2)*Area*Area/Mean_Density;
-
-        nodes->AddMax_Lambda_Visc(iPoint, Lambda);
-
-      }
+  /*--- Define an object to compute the speed of sound. ---*/
+  struct SoundSpeed {
+    FORCEINLINE su2double operator() (const CEulerVariable& nodes, unsigned long iPoint, unsigned long jPoint) const {
+      return 0.5 * (nodes.GetSoundSpeed(iPoint) + nodes.GetSoundSpeed(jPoint));
     }
```
pcarruscag, line 2337: The function objects are defined at function scope (SetTime_Step in this example) to make very clear what are the details for each solver.
Using virtual functions or implementing these details in the corresponding variables classes would work, but it would make finding things a lot harder.

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -282,72 +297,67 @@ class CNEMOEulerSolver : public CFVMFlowSolverBase<CNEMOEulerVariable, COMPRESSI
...
-  void SetResidual_DualTime(CGeometry *geometry, CSolver **solver_container, CConfig *config,
-                            unsigned short iRKStep, unsigned short iMesh, unsigned short RunTime_EqSystem) override;
+   * \brief Update the solution using the classical RK4 scheme.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] iRKStep - Runge-Kutta step.
+   */
+  void ClassicalRK4_Iteration(CGeometry *geometry, CSolver **solver_container,
+                              CConfig *config, unsigned short iRKStep) final;
```
pcarruscag, line 338: This was missing from NEMO and the incompressible solver, any special reason?

WallyMaier: Thats interesting, there shouldn't be a reason to it is missing.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
WallyMaier: Thanks for going through this...I made some attempts to update time integration schemes in another branch but didnt push yet.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2337,132 +1857,123 @@ void CIncEulerSolver::BC_Far_Field(CGeometry *geometry, CSolver **solver_contain
...
-      V_domain = nodes->GetPrimitive(iPoint);
+    /*--- Normal vector for this vertex (negate for outward convention) ---*/
 
-      /*--- Recompute and store the velocity in the primitive variable vector. ---*/
+    geometry->vertex[val_marker][iVertex]->GetNormal(Normal);
+    for (iDim = 0; iDim < nDim; iDim++) Normal[iDim] = -Normal[iDim];
+    conv_numerics->SetNormal(Normal);
 
-      for (iDim = 0; iDim < nDim; iDim++)
-        V_infty[iDim+1] = GetVelocity_Inf(iDim);
```
WallyMaier, line 2377: These calls V_infty[iDim+1],  are one of the main differences between the NEMO solvers and everything.  Would it be wasteful to use the the NEMO indices through the code?  I haven't looked though enough of the code to know the full impact, but it seems that it could go a long way to reduce duplication.

WallyMaier (approved):
@pcarruscag this work looks great.  Ill be on the lookout to reduce these dup/triplications in future PRs.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Hmmm where to start... inline is mostly a mechanism to bypass ODR (one definition rule) if you define a function in an hpp (instead of just declaring), and include that hpp multiple times you are defining the same function multiple times which is not legal, declaring it inline makes it legal (for class methods the rules are less strict).
Because a function is defined inline, the compiler MAY decide (based on a bunch of heuristics I imagine) to copy paste its body into the places where it is used, instead of creating a function call (less indirection and more opportunity to optimize).
Most compilers will have a way for us to ask them to "just do it", i.e. forget about the heuristics and just copy the thing, in SU2 I called that FORCEINLINE.

Summary: FORCEINLINE is triple dog daring the compiler to inline the function.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: oops sorry about that, I was looking at this kind of routine and had a "what if" moment... and now it's done

SU2_CFD/src/solvers/CIncEulerSolver.cpp
pcarruscag: It would be a little yes, because the velocity offset for nemo is not known at compile-time.

But I have something that might interest you in a subsequent PR.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
WallyMaier: No apologies! Im happy its done.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
WallyMaier: Oh boy, I cant wait

TestCases/parallel_regression.py
```diff
@@ -987,7 +987,7 @@ def main():
     channel_2D.cfg_dir   = "sliding_interface/channel_2D"
     channel_2D.cfg_file  = "channel_2D_WA.cfg"
     channel_2D.test_iter = 2
-    channel_2D.test_vals = [2.000000, 0.000000, 0.398053, 0.352779, 0.405462]
```
TobiKattmann, line 990: I might have missed it in the diff but why did this Testcase change? It is only a minor change but I am curious where it was introduced

TobiKattmann (commented):
Thanks for these efforts of `Cut, Paste, Abstract, Survive` 💐 .  And yet again: sorry for being too late to the party

(I hope you don't follow all of Mr. Grylls advices, especially in the realm of refreshing beverages 😄 ) (Mr. Grylls's advice? Mr. Grylls' advice? i dont know...)

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2725,217 +2222,218 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
-    /*--- Allocate the value at the outlet ---*/
-
-    V_outlet = GetCharacPrimVar(val_marker, iVertex);
-
     iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
 
     /*--- Check if the node belongs to the domain (i.e., not a halo node) ---*/
 
-    if (geometry->nodes->GetDomain(iPoint)) {
+    if (!geometry->nodes->GetDomain(iPoint)) continue;
```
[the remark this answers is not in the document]
pcarruscag: It was just that, geometry toolbox, and avoiding those old arrays allocated at Solver level ("Residual", "Res_Conv", "Vector", "Primitive" and so on) in preparation for hybrid parallel stuff.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2725,217 +2222,218 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
-    }
   }
 
-  /*--- Free locally allocated memory ---*/
-  delete [] Normal;
-
 }
 
 void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver_container, CConfig *config,
                                         unsigned short iRKStep, unsigned short iMesh, unsigned short RunTime_EqSystem) {
```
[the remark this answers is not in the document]
pcarruscag: It is a bit more complicated because of the "if energy" and the conversion from primitive to conservative variables

TestCases/parallel_regression.py
pcarruscag: The sliding mesh cases are sensitive to very minor things

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -282,6 +1029,18 @@ class CFVMFlowSolverBase : public CSolver {
...
+   * \brief Set the total residual adding the term that comes from the Dual Time Strategy.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] iRKStep - Current step of the Runge-Kutta iteration.
+   * \param[in] iMesh - Index of the mesh in multigrid computations.
+   * \param[in] RunTime_EqSystem - System of equations which is going to be solved.
+   */
+  void SetResidual_DualTime(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iRKStep,
+                            unsigned short iMesh, unsigned short RunTime_EqSystem) override;
```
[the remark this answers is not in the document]
pcarruscag: That is not exactly what happens there, the method is not a template, the class is.
Each flow solver, incomp, nemo, comp, instantiates a different flavor of CFVMFlowSolverBase.
Each template instantiation creates a different SetResidual_DualTime that overrides the CSolver virtual function.

I don't think you can have method templates of virtual functions, as there would be no way to instantiate them, templates and overloading happen at compile time, and virtual functions at runtime.
That is why it is important to use the virtual specifiers (override, final) to guarantee that a method really overrides something (exact same signature, decide which method to call at runtime based on type) rather than overloading it (different signature, including template params, decide what to call at compile time based on arguments).

What you can have is class templates with virtual functions, each class template instantiation is like a totally different class.

pcarruscag:
Never late to the party @TobiKattmann nothing is set in stone with git. I will have more incomp changes at some point and was a bit over eager to merge to keep myself from pilling them in just one PR.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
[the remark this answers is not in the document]
TobiKattmann: 👍 ok I saw the `V2U` stuff and wondered whether that could be generalized in some way, but if the energy stuff comes on top it might not make too much sense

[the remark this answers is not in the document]
pcarruscag: Yeah it changes whether the loops run to nVar or to nVar-1... It could be done but it felt convoluted.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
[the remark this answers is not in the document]
TobiKattmann: Yep, thanks for the explanation. https://gcc.godbolt.org/z/T6555e this should better reflect the SU2 reality

Issue #1179, opened 2021-01-27
CDriver instances deadlock when initialised with subset of MPI_COMM_WORLD


aa-g:
**Describe the bug**
When creating a CDriver instance, for example through the Python wrapper, with an MPI communicator which is defined only on a subset of the processes on which the MPI library was initialised, SU2 will deadlock. This seems to be because most of the MPI communication and synchronisation operations in SU2 explicitly refer to the global `MPI_COMM_WORLD`.

The issue arose when attempting to control several Euler/RANS SU2 `CSinglezoneDriver` and `CDiscAdjSinglezoneDriver` instances in parallel (i.e. simultaneously) within a larger optimisation code, which provides each analysis with a subset of the ranks in its own MPI communicator.

Naively replacing _all_ references to `MPI_COMM_WORLD` with `SU2_MPI::GetComm()` throughout the code (except where it is defined, and where it is used to initialise the MPI wrapper in the SU2_XXX.cpp files) prevents the deadlock and results in the desired behaviour. However, I imagine that this "fix" may well cause all sorts of issues outside of my own, limited use case.

**To Reproduce**
The issue can be reproduced with any SU2_CFD configuration file and mesh, run in parallel using the Python wrapper as follows: 
```
from mpi4py import MPI
import pysu2  # or pysu2ad


filename = 'example.cfg'
zones = 1

rank = MPI.COMM_WORLD.Get_rank()
size = MPI.COMM_WORLD.Get_size()

group = MPI.COMM_WORLD.group.Excl([0])  # exclude the first rank
comm = MPI.COMM_WORLD.Create_group(group)

if comm:  # is False only on the excluded rank
    print("Running SU2 on global rank {} of {} ... ".format(rank, size))

    # create SU2 instance -- will hang during Geometry preprocessing
    instance = pysu2.CSinglezoneDriver(filename, zones, comm)

else:
    print("Doing nothing on global rank {} of {} ... ".format(rank, size))

```

**Desktop (please complete the following information):**
 - OS: Redhat 7.7
 - C++ compiler and version: g++ (GCC) 4.8.5
 - MPI implementation and version: MPICH 3.2.1
 - SU2 Version: 7.0.2 "Blackbird" (looks to be reproducible on 7.1.0 also)

pcarruscag:
All external libraries seem to take a communicator as argument. And it would be important for this to work for example for fault tolerant communication.
[...]

PR #1180, opened 2021-01-29
Support for MPI communicators which are provided externally

aa-g:
## Proposed Changes
Replace references to the global MPI communicator `MPI_COMM_WORLD` with `SU2_MPI::GetComm()` where appropriate.

This prevents deadlocking in cases where SU2 is given an MPI communicator other than `MPI_COMM_WORLD`, for example when creating an instance of `CDriver` through the Python interface.

## Related Work
Issue #1179 

## PR Checklist


Common/include/linear_algebra/CPastixWrapper.hpp
```diff
@@ -93,7 +93,7 @@ class CPastixWrapper
    * \brief Run the external solver for the task it is currently setup to execute.
    */
   void Run() {
-    dpastix(&state, MPI_COMM_WORLD, nCols, colptr.data(), rowidx.data(), values.data(),
+    dpastix(&state, SU2_MPI::GetComm(), nCols, colptr.data(), rowidx.data(), values.data(),
```
pcarruscag, line 96: I know it is a bit nit picky but can we rename GetComm() to just Comm(), just to make things shorter.

pcarruscag (commented):
Thank you for opening the PR!

Delete line 470 of mpi_structure.hpp please.

aa-g:
> Thank you for opening the PR!
> 
> Delete line 470 of mpi_structure.hpp please.

No worries. It took a while to get it going due to some issues on my side catching up with 7.1.0 (in particular, compiling against MPICH).

Regarding the deletion of the dummy `MPI_COMM_WORLD` from mpi\_structure.hpp: the value is still referenced in several places, namely line 32 of mpi\_structure.cpp, and in each of the SU2\_CFD.cpp, SU2\_DEF.cpp, SU2\_DOT.cpp, SU2\_GEO.cpp, SU2\_SOL.cpp, where `MPI_COMM_WORLD` would be the default communicator.

Common/include/linear_algebra/CPastixWrapper.hpp
aa-g: I agree that the proposed changes make the code slightly less readable.

Unfortunately, the name `SU2_MPI::Comm` is already defined in mpi_structure.hpp, line 98, as a `typedef` to `MPI_Comm`. Maybe this would have to be renamed ?

This change would also break consistency with other methods of `SU2_MPI`, such as `GetRank()` and `GetSize()`.

pcarruscag: Hmmm alright consistency is important, especially since we don't have a lot of it.

pcarruscag:
Add a method to the MPI wrappers to get the global communicator or something equivalent, the only way to make this kind of change future-compatible is to break compilation.

pcarruscag:
I merged #1178 earlier and fixed the old and new conflicts, some MPI_COMM_WORLD's will have been re-introduced in CFVMFlowSolverBase.hpp and .inl because in my previous 2 PR's I moved a lot of code.

aa-g:
I've removed the dummy `MPI_COMM_WORLD` altogether now. To keep the changes minimal, I've just changed the way that `CBaseMPIWrapper::currentComm` is initialised in `mpi_structure.cpp`, rather than introducing a new method (which would only be used once, at this location).

Now, the `CBaseMPIWrapper::currentComm` is initialised to `MPI_COMM_WORLD` if compiling with MPI, otherwise it is set to the dummy value `0`, consistent with the previous behaviour.

Common/include/parallelization/mpi_structure.cpp
```diff
@@ -27,9 +27,17 @@
...
+/* Initialise the MPI Communicator Rank and Size */
 int CBaseMPIWrapper::Rank = 0;
 int CBaseMPIWrapper::Size = 1;
+
+/* Set the default MPI Communicator */
+#ifdef HAVE_MPI
 CBaseMPIWrapper::Comm CBaseMPIWrapper::currentComm = MPI_COMM_WORLD;
+#else
+CBaseMPIWrapper::Comm CBaseMPIWrapper::currentComm = 0;  // dummy value
+#endif
```
pcarruscag, line 40: I like this solution.
The rank and size are updated by the SetComm function so everything should work ok when creating a driver from scratch.

pcarruscag (approved):

Common/include/parallelization/mpi_structure.cpp
aa-g: Glad to hear that. Is anything still missing ? Not sure how to go about the failing check for code complexity ...

pcarruscag: Not for me, I'll ask around the developers meeting tomorrow if anyone else has comments, codefactor glitches out sometimes no need to do anything.

By the way would you like to be added to su2code? You can then have branches here instead of in your fork, makes it simpler to manage future PR's you might open.

aa-g: Yes, I suppose it would be more convenient that way. Thanks !

Issue #1185, opened 2021-02-06
Edge vector for boundary numerics


maxaehle:
I wonder whether in the following code the arguments of `SetCoord` should be swapped. The code illustrates how the `visc_numerics` is initialized in `CEulerSolver::BC_Far_Field`:

    iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
    Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();

    V_domain = nodes->GetPrimitive(iPoint);
    V_infty = ...

    visc_numerics->SetCoord(geometry->nodes->GetCoord(iPoint),
                            geometry->nodes->GetCoord(Point_Normal));
    visc_numerics->SetPrimitive(V_domain, V_infty);

So `iPoint` lies on the boundary. *`Point_Normal` is its normal neighbour in the interior of the domain, and not some "dummy cell" outside. But the primitive variables at `Point_Normal` are set to the outside values.*

As far as I can tell, the `CNumerics::Coord_j` member is only used in the child classes of `CAvgGrad_Base`, in order to compute

    Edge_Vector[MAXNDIM] = {0.0},           /*!< \brief Vector from point i to point j. */

by

    Edge_Vector[iDim] = Coord_j[iDim]-Coord_i[iDim];

which is then used to call `CorrectGradient`. It has the wrong sign if the above code is incorrect. (Apart from that, `Coord_j` is used to compute the distance in the NEMO viscous residuals.)

Note that similar calls to `SetCoord` can be found at many places, e.g. in `CEulerSolver::BC_Riemann`, `CEulerSolver::BC_Giles`, their siblings in `CIncEulerSolver`. and `CTurbSASolver::BC_Inlet_Turbo`.

pcarruscag:
The gradient correction that involves the point coordinates is not active for the VISC_BOUND term of the numerics.
And we flip the normal before passing it to the numerics... Can you check if the dot product of the normal with the edge vector is positive?
Either the code has a silent bug, or it is correct but could use a comment about this :)

maxaehle:
I tried it for the NACA0012 897x257 mesh with compressible RANS (in `CAvgGrad_Flow::ComputeResidual`). 

The scalar product is negative at the far-field boundary. `Coord_i` is on the boundary, `Coord_j` is in the domain. `Normal` points outwards in the viscous numerics (but I don't know if this depends on the orientation of the boundary in the mesh file).

[...]

PR #1189, opened 2021-02-07
Fixed Coord_j for boundary viscous numerics (bug waiting to happen)

maxaehle:
## Proposed Changes
    
The visc_numerics' coordinates were often set to the coordinates
of the boundary vertex and its closest normal neighbour. Setting
them to the boundary vertex and the reflection of the closest
normal neighbour at the vertex is more systematic however.
Normally this should not change much, because the boundary numerics
do not apply gradient correction, so they only depend on the
distance between Coord_i, Coord_j.

## Related Work
#1185 

## PR Checklist

- [X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).

### DIFF
--- a/Common/include/toolboxes/geometry_toolbox.hpp
+++ b/Common/include/toolboxes/geometry_toolbox.hpp
@@ -50,6 +50,13 @@ inline void Distance(Int nDim, const T* a, const T* b, T* d) {
   for(Int i = 0; i < nDim; i++) d[i] = a[i] - b[i];
 }
 
+/*! \brief Reflect a at b: c = 2*b - a
+ */
+template<class T, typename Int>
+inline void PointPointReflect(Int nDim, const T* a, const T* b, T* d){
+  for(Int i = 0; i < nDim; i++) d[i] = 2 * b[i] - a[i];
+}
+
 /*! \return a.b */
 template<class T, typename Int>
 inline T DotProduct(Int nDim, const T* a, const T* b) {
--- a/SU2_CFD/src/solvers/CEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CEulerSolver.cpp
@@ -5155,8 +5155,10 @@ void CEulerSolver::BC_Far_Field(CGeometry *geometry, CSolver **solver_container,
         /*--- Set the normal vector and the coordinates ---*/
 
         visc_numerics->SetNormal(Normal);
-        visc_numerics->SetCoord(geometry->nodes->GetCoord(iPoint),
-                                geometry->nodes->GetCoord(Point_Normal));
+        su2double Coord_Reflected[MAXNDIM];
+        GeometryToolbox::PointPointReflect(nDim, geometry->nodes->GetCoord(Point_Normal),
+                                                 geometry->nodes->GetCoord(iPoint), Coord_Reflected);
+        visc_numerics->SetCoord(geometry->nodes->GetCoord(iPoint), Coord_Reflected);
 
         /*--- Primitive variables, and gradient ---*/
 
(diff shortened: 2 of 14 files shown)


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -8848,7 +8877,10 @@ void CEulerSolver::BC_ActDisk(CGeometry *geometry, CSolver **solver_container, C
 //        /*--- Set the normal vector and the coordinates ---*/
 //
 //        visc_numerics->SetNormal(Normal);
-//        visc_numerics->SetCoord(geometry->nodes->GetCoord(iPoint), geometry->node[iPoint_Normal]->GetCoord());
```
maxaehle, line 8851: Apart from the obvious change I also replaced `geometry->node[iPoint_Normal]->GetCoord()` by `geometry->nodes->GetCoord(Point_Normal)`.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -803,7 +809,10 @@ void CTurbSASolver::BC_Engine_Inflow(CGeometry *geometry, CSolver **solver_conta
 
 //      /*--- Viscous contribution, commented out because serious convergence problems ---*/
 //
-//      visc_numerics->SetCoord(geometry->nodes->GetCoord(iPoint), geometry->nodes->GetCoord(iPoint));
```
maxaehle, line 806: Here I also replaced the second `GetCoord(iPoint)` by `GetCoord(Point_Normal)` because it sound logical to me. I'm not sure whether this additional change is correct.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -887,7 +896,10 @@ void CTurbSASolver::BC_Engine_Exhaust(CGeometry *geometry, CSolver **solver_cont
...
```
maxaehle, line 890: Here again.
`Point_Normal` is not defined by the surrounding code, but this also applies to other commented-out sections.

Common/include/toolboxes/geometry_toolbox.hpp
```diff
@@ -50,6 +50,13 @@ inline void Distance(Int nDim, const T* a, const T* b, T* d) {
   for(Int i = 0; i < nDim; i++) d[i] = a[i] - b[i];
 }
 
+/*! \brief Reflect a at b: c = 2*b - a
+ */
+template<class T, typename Int>
+inline void PointPointReflect(Int nDim, const T* a, const T* b, T* d){
+  for(Int i = 0; i < nDim; i++) d[i] = 2 * b[i] - a[i];
+}
+
```
pcarruscag, line 59: That's the spirit :+1:

SU2_CFD/src/solvers/CTurbSASolver.cpp
pcarruscag: When the distance is 0 the viscous numerics do not compute a Jacobian, I wonder if that has anything to do with the "serious convergence problems".
I also do not know if it is correct, but it is commented out code so... it does not matter I guess.

pcarruscag (approved):

pcarruscag:
I edited your title not to scare someone reading the change log too much.
I would check (i.e. run to convergence) a few of the failed testcases especially those where CD and CL are not the same.
Small changes to only the residuals should be safe to update.

maxaehle:
I checked some of the failed testcases by letting the solution converge:

- The testcase `inc_poly_cylinder` in serial and parallel mode, and `slinc_steady` in serial mode, converge to the same solution.
- For the serial AD testcase `discadj_incomp_cylinder`, the relative difference of the sensitivities is around 1e-3. I don't know if this is a bit high?

Regarding the other testcases:

- With the turbomachinery testcases and the hybrid regression tests I'm not too familiar.
- Note that `slinc_steady` in `parallel_regression.py` is started by a pure call to `SU2_CFD`, without `mpirun` or `parallel_computation.py`. This applies to few other testcases in this script as well. So this just repeats the serial regression.
- In parallel, the failed AD testcase `discadj_incomp_cylinder` does not converge (in both versions).

I think I can update the residuals these days.

pcarruscag:
Alright thanks for checking them, hybrid regressions are the same, 1e-3 should be fine, if the case does not converge in parallel it is probably not very stable which can make adjoint results a bit "noisy".

maxaehle:
How can I fix the CodeFactor issue (complex code in `SU2_CFD/src/solvers/CIncEulerSolver.cpp`)?

PR #1192, opened 2021-02-08
Fix axi auxvargrad

FlorianDm:
This is a small fix of two bugs related to the viscous axisymmetric source terms.

1. As I had suspected in  #1106 it should be `AxiAuxVar_Grad_i[2][0]` instead of `AxiAuxVar_Grad_i[2][1]`. The book by Hoffman that was used as a reference actually contained an error. I am attaching a page from another edition of the book (thanks to @dbrezgin).

image

2. The type for the auxvar and gradient fields should be a reference `auto&` not just `auto`. i found the bug after it was pointed out to me that it was always set to zero (thanks to @dbrezgin).


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2456,8 +2456,8 @@ void CSolver::SetRotatingFrame_GCL(CGeometry *geometry, const CConfig *config) {
 
 void CSolver::SetAuxVar_Gradient_GG(CGeometry *geometry, const CConfig *config) {
 
-  const auto solution = base_nodes->GetAuxVar();
-  auto gradient = base_nodes->GetAuxVarGradient();
+  const auto& solution = base_nodes->GetAuxVar();
+  auto& gradient = base_nodes->GetAuxVarGradient();
```
pcarruscag, line 2460: indeed the not so auto thing about auto...

pcarruscag (approved):
Thank you Florian, can you put the same fix in the NEMO numerics? (if @jtneedels agrees)

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: Is there no regression for axisymmetric flow? One would think something like this could break something.

FlorianDm: I don't think there is but it did just set them to zero and we don't divide by them or anything so nothing broke just inaccurate results and perhaps not converging when it should in some cases. No compiler error or warning so i guess with auto one needs to be careful

pcarruscag: Can we have a regression that "exercises" this part of the implementation? Perhaps just modifying an existing case?

WallyMaier: @FlorianDm @pcarruscag  believe there was another issue within NEMO axisymmetric that @jtneedels fixed.  Maybe it would be a good time to merge that in as well?

jtneedels (approved):

SU2_CFD/src/solvers/CSolver.cpp
jtneedels: @FlorianDm Yes I agree with these changes, thanks for catching that! @pcarruscag Yes, as @WallyMaier mentioned I do have another small fix to pull in for axisymmetric in CNEMOEulerSolver, should I go ahead ad open a separate PR to merge that in?

FlorianDm: I will try to create a test/regression case after adding the k-omega source terms (almost done) because I think it is more useful if it is a rans case

PR #1193, opened 2021-02-09
Fix some screen output and frozen limiter logic

### LINKED ISSUE #1190: Error with wetted area calculation
By running exactly the same mesh and config flle on Linux Centos 8 with two version of SU2, it leads to two different values for the wetted area which is calculated during the initialization.

with SU2 v7.0.3:

> Finding max control volume width.
> Semi-span length = 13.5934 m.
> **Wetted area = 466.385 m^2.**
> Area projection in the x-plane = 19.4802 m^2, y-plane = 109.691 m^2, z-plane = 165.851 m^2.
> Max. coordinate in the x-direction = 34.8253 m, y-direction = 13.5934 m, z-direction = 5.44607 m.
> Min. coordinate in the x-direction = -0.00557099 m, y-direction = -13.5934 m, z-direction = -1.75235 m.

with SU2 v7.1.0:

> Finding max control volume width.
> Semi-span length = 13.5934 m.
> **Wetted area = 1.85378e-06 m^2.**
> Area projection in the x-plane = 19.4802 m^2, y-plane = 109.691 m^2, z-plane = 165.851 m^2.
> Max. coordinate in the x-direction = 34.8253 m, y-direction = 13.5934 m, z-direction = 5.44607 m.
> Min. coordinate in the x-direction = -0.00557099 m, y-direction = -13.5934 m, z-direction = -1.75235 m

Regarding my geometry and the other values given, the second one (v7.1.0) seems to be the wrong one.

pcarruscag:
## Proposed Changes
Fix #1190
Prevent possible OpenMP bug for van Albada limiter.
Fix frozen limiter logic, freezing means "do not compute", and not "do not apply".
Fix limiter outputs for turbulence variables.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/output/CFlowIncOutput.cpp
+++ b/SU2_CFD/src/output/CFlowIncOutput.cpp
@@ -436,25 +436,28 @@ void CFlowIncOutput::SetVolumeOutputFields(CConfig *config){
     break;
   }
 
-  // Limiter values
-  AddVolumeOutput("LIMITER_PRESSURE", "Limiter_Pressure", "LIMITER", "Limiter value of the pressure");
-  AddVolumeOutput("LIMITER_VELOCITY-X", "Limiter_Velocity_x", "LIMITER", "Limiter value of the x-velocity");
-  AddVolumeOutput("LIMITER_VELOCITY-Y", "Limiter_Velocity_y", "LIMITER", "Limiter value of the y-velocity");
-  if (nDim == 3)
-    AddVolumeOutput("LIMITER_VELOCITY-Z", "Limiter_Velocity_z", "LIMITER", "Limiter value of the z-velocity");
-  AddVolumeOutput("LIMITER_TEMPERATURE", "Limiter_Temperature", "LIMITER", "Limiter value of the temperature");
+  if (config->GetKind_SlopeLimit_Flow() != NO_LIMITER && config->GetKind_SlopeLimit_Flow() != VAN_ALBADA_EDGE) {
+    AddVolumeOutput("LIMITER_PRESSURE", "Limiter_Pressure", "LIMITER", "Limiter value of the pressure");
+    AddVolumeOutput("LIMITER_VELOCITY-X", "Limiter_Velocity_x", "LIMITER", "Limiter value of the x-velocity");
+    AddVolumeOutput("LIMITER_VELOCITY-Y", "Limiter_Velocity_y", "LIMITER", "Limiter value of the y-velocity");
+    if (nDim == 3)
+      AddVolumeOutput("LIMITER_VELOCITY-Z", "Limiter_Velocity_z", "LIMITER", "Limiter value of the z-velocity");
+    AddVolumeOutput("LIMITER_TEMPERATURE", "Limiter_Temperature", "LIMITER", "Limiter value of the temperature");
+  }
 
-  switch(config->GetKind_Turb_Model()){
-  case SST: case SST_SUST:
-    AddVolumeOutput("LIMITER_TKE", "Limiter_TKE", "LIMITER", "Limiter value of turb. kinetic energy.");
-    AddVolumeOutput("LIMITER_DISSIPATION", "Limiter_Omega", "LIMITER", "Limiter value of dissipation rate.");
-    break;
-  case SA: case SA_COMP: case SA_E:
-  case SA_E_COMP: case SA_NEG:
-    AddVolumeOutput("LIMITER_NU_TILDE", "Limiter_Nu_Tilde", "LIMITER", "Limiter value of Spalart–Allmaras variable.");
-    break;
-  case NONE:
-    break;
+  if (config->GetKind_SlopeLimit_Turb() != NO_LIMITER) {
+    switch(config->GetKind_Turb_Model()){
+    case SST: case SST_SUST:
+      AddVolumeOutput("LIMITER_TKE", "Limiter_TKE", "LIMITER", "Limiter value of turb. kinetic energy.");
+      AddVolumeOutput("LIMITER_DISSIPATION", "Limiter_Omega", "LIMITER", "Limiter value of dissipation rate.");
+      break;
+    case SA: case SA_COMP: case SA_E:
+    case SA_E_COMP: case SA_NEG:
+      AddVolumeOutput("LIMITER_NU_TILDE", "Limiter_Nu_Tilde", "LIMITER", "Limiter value of Spalart–Allmaras variable.");
+      break;
+    case NONE:
+      break;
+    }
   }
 
   // Hybrid RANS-LES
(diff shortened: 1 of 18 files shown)


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2431,20 +2422,22 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
...
-        if (limiter) {
-          if (van_albada) {
-            su2double V_ij = V_j[iVar] - V_i[iVar];
-            Limiter_i[iVar] = V_ij*( 2.0*Project_Grad_i + V_ij) / (4*pow(Project_Grad_i, 2) + pow(V_ij, 2) + EPS);
-            Limiter_j[iVar] = V_ij*(-2.0*Project_Grad_j + V_ij) / (4*pow(Project_Grad_j, 2) + pow(V_ij, 2) + EPS);
-          }
-          Primitive_i[iVar] = V_i[iVar] + Limiter_i[iVar]*Project_Grad_i;
-          Primitive_j[iVar] = V_j[iVar] + Limiter_j[iVar]*Project_Grad_j;
+        su2double lim_i = 1.0;
+        su2double lim_j = 1.0;
```
TobiKattmann, line 2426: Btw: The compiler (with optimization on) is clever enough to optimize out variable creation inside of these loops, right. So creating a static var inside of  an edge/point loop is fine? The unspoken rule (to some point in the past) was to put the var declaration outside of these loops. If there is no performance penalty to that than I also prefer the version you used here (I already do that better be the case).

Good job with the limiter: reads easier like that and all necessary code is close together

TobiKattmann (approved):
LGTM💐 Just a few things to point out that I think are good to adapt myself :)

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -380,41 +330,23 @@ void CFlowOutput::SetAnalyzeSurface(CSolver *solver, CGeometry *geometry, CConfi
...
-    Surface_TotalTemperature_Total[iMarker_Analyze]  = Surface_TotalTemperature_Local[iMarker_Analyze];
-    Surface_TotalPressure_Total[iMarker_Analyze]     = Surface_TotalPressure_Local[iMarker_Analyze];
-    Surface_Area_Total[iMarker_Analyze]              = Surface_Area_Local[iMarker_Analyze];
-    Surface_MassFlow_Abs_Total[iMarker_Analyze]      = Surface_MassFlow_Abs_Local[iMarker_Analyze];
-  }
-
-#endif
+  auto Allreduce = [](const vector<su2double>& src, vector<su2double>& dst) {
+    SU2_MPI::Allreduce(src.data(), dst.data(), src.size(), MPI_DOUBLE, MPI_SUM, SU2_MPI::GetComm());
+  };
```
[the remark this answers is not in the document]
pcarruscag: What I would really really is for the MPI wrapper to have overloads for vector types and deduce the mpi type and size from them... But that is a lot of work xD

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: I don't think the rule was unspoken, you had to do it for Fortran and at least early C.

Local variables are optimized away provided they do not involve dynamic allocation, or more formally that they don't have effects outside of local scope.
https://gcc.godbolt.org/z/Yx1hWT
Heap elision is tricky stuff (5 min video https://www.youtube.com/watch?v=FWD1msG8YU4&ab_channel=C%E1%90%A9%E1%90%A9WeeklyWithJasonTurner) and in general we cannot assume compilers will do it.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2253,7 +2253,7 @@ unsigned long CEulerSolver::SetPrimitive_Variables(CSolver **solver_container, c
   SU2_OMP_FOR_STAT(omp_chunk_size)
   for (unsigned long iPoint = 0; iPoint < nPoint; iPoint ++) {
 
-    /*--- Compressible flow, primitive variables nDim+5, (T, vx, vy, vz, P, rho, h, c, lamMu, eddyMu, ThCond, Cp) ---*/
+    /*--- Compressible flow, primitive variables nDim+9, (T, vx, vy, vz, P, rho, h, c, lamMu, eddyMu, ThCond, Cp) ---*/
```
TobiKattmann, line 2256: Just a thought: It would be nice to switch `P` and `T` for the comp-var to be consistent with the inc-variable. Not sure how easy that is to change though (maybe just change the `GetPressure` and `GetTemperature` call and one is good).  But maybe this is also a stupid idea.

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -90,12 +91,13 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
...
-        We need P, and rho for running the adjoint problem ---*/
+  /*--- Incompressible flow, gradients primitive variables nDim+4, (P, vx, vy, vz, T, rho, beta) ---*/
 
-  Gradient_Primitive.resize(nPoint,nPrimVarGrad,nDim,0.0);
+  if (config->GetMUSCL_Flow() || viscous) {
+    Gradient_Primitive.resize(nPoint,nPrimVarGrad,nDim,0.0);
+  }
 
-  if (config->GetReconstructionGradientRequired()) {
+  if (config->GetMUSCL_Flow() && config->GetReconstructionGradientRequired()) {
```
TobiKattmann, line 100: In which case is one true and the other isn't? MUSCL and ReconstructedGradient are close friends and I am unaware when they show up separated?

pcarruscag: You are right, I had another look at the long config post processing... what can happen is an unfortunate mix of centered and upwind options.

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Even nicer if the incompressible solvers would just use the solution directly instead of storing P u v w T twice.
Changing storage places would break all the numerics, as they get a pointer to the start of the primitives.

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -587,27 +590,31 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
...
+      SetVolumeOutputValue("LIMITER_VELOCITY-Z", iPoint, Node_Flow->GetLimiter_Primitive(iPoint, 3));
+      SetVolumeOutputValue("LIMITER_TEMPERATURE", iPoint, Node_Flow->GetLimiter_Primitive(iPoint, 4));
+    } else {
+      SetVolumeOutputValue("LIMITER_TEMPERATURE", iPoint, Node_Flow->GetLimiter_Primitive(iPoint, 3));
+    }
   }
 
-  switch(config->GetKind_Turb_Model()){
-  case SST: case SST_SUST:
-    SetVolumeOutputValue("LIMITER_TKE", iPoint, Node_Turb->GetLimiter_Primitive(iPoint, 0));
```
pcarruscag, line 602: Should be just limiter rather limiter_primitive

SU2_CFD/src/variables/CIncEulerVariable.cpp
pcarruscag: And also on coarse multigrid levels, there is no MUSCL but the reconstruction will still be true, but that is too many if's to deal with in CVariable.

SU2_CFD/src/solvers/CEulerSolver.cpp
TobiKattmann: Good to see you had a little `fun` :) and I should have thought about compiler-explorer myself to check that out. (Also means I have understand what happens in the assembler language to roughly understand what is happening.)
And interesting video on 'heap elision', I didn't know that existed.( And for the 10th time in a year I tell myself again: "Time to use clang for a bit." :) ).

PR #1194, opened 2021-02-10
Fix Multigrid for Incompressible NS with moving walls, and for NEMO NS in general

TobiKattmann:
## Proposed Changes
During multigrid prolongation the no-slip (visc.) wall BC are set again. If there is Grid velocity it was multiplied by the 0th solution entry, both inc and compr. For compr flow it is density which is correct but for inc flow that is pressure. That is corrected by overloading.
 
In a little test with `TestCases/unsteady/pitching_naca0015_rans_inc/config_incomp_turb_sa.cfg` the linear solver diverged in the first iteration before, and now it works.

Discovered by @pcarruscag and discussed in last weeks DevMeeting, so props to him.

## PR Checklist
- [x] I have added a test case that demonstrates my contribution, if necessary. *not necessary*

### DIFF
--- a/SU2_CFD/src/integration/CMultiGridIntegration.cpp
+++ b/SU2_CFD/src/integration/CMultiGridIntegration.cpp
@@ -350,7 +350,8 @@ void CMultiGridIntegration::GetProlongated_Correction(unsigned short RunTime_EqS
         /*--- For dirichlet boundary condtions, set the correction to zero.
          Note that Solution_Old stores the correction not the actual value ---*/
 
-        sol_coarse->GetNodes()->SetVelSolutionOldZero(Point_Coarse);
+        su2double zero[3] = {0.0};
+        sol_coarse->GetNodes()->SetVelocity_Old(Point_Coarse, zero);
 
       }
     }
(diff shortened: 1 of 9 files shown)


TobiKattmann:
When I want to see what what codefactor complains about I only get the filename (no linenumber or anything specifc) and the complaint (`Number of execution path through function or method` in this case). Am i just blind?
That complaint does not really help tbh and in this case I didn't change the `number of execution paths`.

pcarruscag:
I don't know if it is a code factor glitch, it's basically saying "fix the whole thing" which is a bit harsh, sometimes it shuts up after you push many commits.

pcarruscag:
Pardon the intrusion good sir, I noticed the truncated residual and co. could also be an issue for the NEMO folks.

pcarruscag (approved):

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -598,8 +598,8 @@ void CMultiGridIntegration::SetRestricted_Solution(unsigned short RunTime_EqSyst
           }
           else {
             /*--- For stationary no-slip walls, set the velocity to zero. ---*/
-
-            sol_coarse->GetNodes()->SetVelSolutionZero(Point_Coarse);
+            su2double zero[3] = {0.0};
+            sol_coarse->GetNodes()->SetVelSolutionVector(Point_Coarse, zero);
```
TobiKattmann, line 602: Thanks for this commit @pcarruscag . I forgot that that this had to be overloaded as well :( ... but just reusing the `SetVelSolutionVector` func is nice as well 👍

pcarruscag:
Thanks for fixing the fix :+1:

PR #1195, opened 2021-02-10
k-omega SST 2D axisymmetric source terms

FlorianDm:
This implements the k-omega SST source terms for 2D axisymmetric problems in a new member function of the existing source term class (it would never be used independently of it so I did not create a new class). There are some minor changes to the existing source term residual function like using that `TWO3` instead of `2.0/3.0`.

The extra source terms are the difference between the equations in 2D cartesian and 3D cylindrical coordinates under an assumption of axisymmetry without swirl. They are almost the same as those for Wilcox's k-omega shown in this image (taken from http://eprints.gla.ac.uk/184394/1/184394.pdf). 
kwcyl
Only the constants have been swapped for the blended ones and the changed definition of the eddy viscosity is taken into account (see [menter1994.pdf](https://github.com/su2code/SU2/files/5957873/menter1994.pdf)). Cross diffusion is the same as in 2D when assuming there is no azimuth gradient.

Note that these source terms have the opposite sign (RHS) of the flow equations source terms (LHS) since the block is subtracted and not added.

Looking into creating an axisymmetric RANS test/regression case

### DIFF
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -334,7 +334,7 @@ private:
   /*!
    * \brief Add contribution due to axisymmetric formulation to 2D residual
    */
-  void ResidualAxisymmetric();
+  void ResidualAxisymmetric(su2double alfa_blended);
 
 public:
   /*!
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -334,7 +334,7 @@ private:
   /*!
    * \brief Add contribution due to axisymmetric formulation to 2D residual
    */
-  void ResidualAxisymmetric(su2double alfa_blended);
+  void ResidualAxisymmetric(su2double alfa_blended, su2double zeta);
 
 public:
   /*!
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -307,8 +307,8 @@ private:
   beta_2,
   sigma_k_1,
   sigma_k_2,
-  sigma_omega_1,
-  sigma_omega_2,
+  sigma_w_1,
+  sigma_w_2,
   beta_star,
   a1;
 
@@ -335,44 +335,40 @@ private:
    */
   inline void ResidualAxisymmetric(su2double alfa_blended, su2double zeta){
 
-    if (Coord_i[1] < EPS) {
-      return;
-    }
+    if (Coord_i[1] < EPS) return;
      
-    su2double yinv, rhov;
-    su2double sigma_k_i, sigma_omega_i;
-    su2double pk_axi, pw_axi, ck_axi, cw_axi, dk_axi, dw_axi;
+    su2double yinv, rhov, k, w;
+    su2double sigma_k_i, sigma_w_i;
+    su2double pk_axi, pw_axi, cdk_axi, cdw_axi;
 
     AD::SetPreaccIn(Coord_i[1]);
+    
     yinv = 1.0/Coord_i[1];
     rhov = Density_i*V_i[2];
+    k = TurbVar_i[0];
+    w = TurbVar_i[1];
     
     /*--- Compute blended constants ---*/
-    sigma_k_i = F1_i*sigma_k_1 + (1.0 - F1_i)*sigma_k_2;
-    sigma_omega_i = F1_i*sigma_omega_1 + (1.0 - F1_i)*sigma_omega_2;
+    sigma_k_i = F1_i*sigma_k_1+(1.0-F1_i)*sigma_k_2;
+    sigma_w_i = F1_i*sigma_w_1+(1.0-F1_i)*sigma_w_2;
 
     /*--- Production ---*/
-    pk_axi = max(0.0,2.0/3.0*rhov*TurbVar_i[0]*(2.0/zeta*(yinv*V_i[2]-PrimVar_Grad_i[2][1]
-                                                                -PrimVar_Grad_i[1][0]) -1.0));
-    pw_axi = alfa_blended*zeta/TurbVar_i[0]*pk_axi;
+    pk_axi = max(0.0,2.0/3.0*rhov*k*(2.0/zeta*(yinv*V_i[2]-PrimVar_Grad_i[2][1]-PrimVar_Grad_i[1][0])-1.0));
+    pw_axi = alfa_blended*zeta/k*pk_axi;
     
-    /*--- Convection ---*/
-    ck_axi = rhov*TurbVar_i[0];
-    cw_axi = rhov*TurbVar_i[1];
+    /*--- Convection-Diffusion ---*/
+    cdk_axi = rhov*k-(Laminar_Viscosity_i+sigma_k_i*Eddy_Viscosity_i)*TurbVar_Grad_i[0][1];
+    cdw_axi = rhov*w-(Laminar_Viscosity_i+sigma_w_i*Eddy_Viscosity_i)*TurbVar_Grad_i[1][1];
     
-    /*--- Diffusion ---*/
-    dk_axi = (Laminar_Viscosity_i+sigma_k_i*Eddy_Viscosity_i)*TurbVar_Grad_i[0][1];
-    dw_axi = (Laminar_Viscosity_i+sigma_omega_i*Eddy_Viscosity_i)*TurbVar_Grad_i[1][1];
-    
-    /*--- Add all terms to the residuals ---*/
-    Residual[0] += yinv*Volume*(pk_axi-ck_axi+dk_axi);
-    Residual[1] += yinv*Volume*(pw_axi-cw_axi+dw_axi);
+    /*--- Add terms to the residuals ---*/
+    Residual[0] += yinv*Volume*(pk_axi-cdk_axi);
+    Residual[1] += yinv*Volume*(pw_axi-cdw_axi);
   
-    /*--- Add contribution to the jacobian for implicit time integration---*/
-    Jacobian_i[0][0] += yinv*Volume*(sigma_k_i/zeta*TurbVar_Grad_i[0][1]-V_i[2]);
-    Jacobian_i[0][1] -= yinv*Volume*sigma_k_i*TurbVar_i[0]*TurbVar_Grad_i[0][1]/(zeta*zeta);
-    Jacobian_i[1][0] += yinv*Volume*sigma_k_i/zeta*TurbVar_Grad_i[1][1];
-    Jacobian_i[1][1] -= yinv*Volume*(sigma_k_i*TurbVar_i[0]*TurbVar_Grad_i[1][1]/(zeta*zeta)+V_i[2]);
+    /*--- Add contribution to the jacobian for implicit time integration--- */  
+    Jacobian_i[0][0] += yinv*Volume*(sigma_k_i*TurbVar_Grad_i[0][1]/zeta-V_i[2]);
+    Jacobian_i[0][1] += yinv*Volume*(-sigma_k_i*k*TurbVar_Grad_i[0][1]/(zeta*zeta));
+    Jacobian_i[1][0] += yinv*Volume*(sigma_w_i*TurbVar_Grad_i[1][1]/zeta);
+    Jacobian_i[1][1] += yinv*Volume*(-sigma_w_i*k*TurbVar_Grad_i[1][1]/(zeta*zeta)-V_i[2]);
 
   }
 
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -338,44 +338,41 @@ private:
     if (Coord_i[1] < EPS) {
       return;
     }
+     
+    su2double yinv, rhov;
+    su2double sigma_k_i, sigma_omega_i;
+    su2double pk_axi, pw_axi, ck_axi, cw_axi, dk_axi, dw_axi;
     
-    else{
-      
-      su2double yinv, rhov;
-      su2double sigma_k_i, sigma_omega_i;
-      su2double pk_axi, pw_axi, ck_axi, cw_axi, dk_axi, dw_axi;
-      
-      yinv = 1.0/Coord_i[1];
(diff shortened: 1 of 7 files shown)


SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -761,9 +761,13 @@ CSourcePieceWise_TurbSST::CSourcePieceWise_TurbSST(unsigned short val_nDim,
 
   incompressible = (config->GetKind_Regime() == INCOMPRESSIBLE);
   sustaining_terms = (config->GetKind_Turb_Model() == SST_SUST);
+  implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
```
WallyMaier, line 764: Correct me if Im wrong here, but aren't the turbulent equations always implicit?

Also, I believe this would have to be GetKind_TimeIntScheme_Turb?

pcarruscag: It is ok to always compute the Jacobian when they are this simple, and let the solver decide if it wants to use the result or not. It can potentially solve a lot of head ache.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -816,88 +822,96 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSST::ComputeResidual(const CConfi
...
+    
+    pk = min(pk,20.0*beta_star*Density_i*TurbVar_i[1]*TurbVar_i[0]);
+    pk = max(pk,0.0);
+    
+    zeta = max(TurbVar_i[1], VorticityMag*F2_i/a1);
+    
+    /* if using UQ methodolgy, calculate production using perturbed Reynolds stress matrix */
+    
+    if (using_uq){
+      pw = PerturbedStrainMag * PerturbedStrainMag - TWO3*zeta*diverg;
```
pcarruscag, line 866: I'll be honest, I don't find the TWO3 thing any more readable than 2.0/3.0.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -922,3 +936,40 @@ void CSourcePieceWise_TurbSST::SetPerturbedStrainMag(su2double turb_ke){
   PerturbedStrainMag = sqrt(2.0*PerturbedStrainMag);
 
 }
+
+void CSourcePieceWise_TurbSST::ResidualAxisymmetric(su2double alfa_blended, su2double zeta){
+
+  if (Coord_i[1] > EPS) {
+
```
pcarruscag, line 943: In these cases I just return early from the function, or continue/break a loop to keep the indentations from growing to much.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -922,3 +936,40 @@ void CSourcePieceWise_TurbSST::SetPerturbedStrainMag(su2double turb_ke){
   PerturbedStrainMag = sqrt(2.0*PerturbedStrainMag);
 
 }
+
+void CSourcePieceWise_TurbSST::ResidualAxisymmetric(su2double alfa_blended, su2double zeta){
+
```
pcarruscag, line 941: I would define this inline (in the hpp) and let the compiler have some fun trying to re-use some of the stuff that exists in the body of ComputeResidual.
Probably not massive performance gains but they are completely free in this case.

Otherwise everything looks very nice. Are you going to add the regression in this PR?

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
FlorianDm: Alright I see... thanks. Well spotted copy paste bug!

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
FlorianDm: I had just used it to be consistent because I saw it elsewhere in the code. Personally I find it even less readable than 2.0/3.0

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
FlorianDm: ok makes sense

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
FlorianDm: Okay why not... I make it inline 

I was going to add the case in another PR but as you prefer. I reckon there is not much risk in merging this in already since the current state of affairs is no more validated or tested for performance and strictly wrong as it is incomplete. From a little testing I can see only a small effect on the results but convergence seems to be improved

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -320,12 +322,61 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+   * \brief Add contribution due to axisymmetric formulation to 2D residual
+   */
+  inline void ResidualAxisymmetric(su2double alfa_blended, su2double zeta){
+
+    if (Coord_i[1] < EPS) {
+      return;
+    }
+    
+    else{
+
```
pcarruscag, line 343: The idea of the early return was not to need the else xD

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -320,12 +322,61 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      return;
+    }
+    
+    else{
+      
+      su2double yinv, rhov;
+      su2double sigma_k_i, sigma_omega_i;
+      su2double pk_axi, pw_axi, ck_axi, cw_axi, dk_axi, dw_axi;
+      
+      yinv = 1.0/Coord_i[1];
```
pcarruscag, line 348:
```suggestion
      AD::SetPreaccIn(Coord_i[1]);
      yinv = 1.0/Coord_i[1];
```

pcarruscag: We would need that for the discrete adjoint, I think the coordinate is the only new input this section of code needs? Everything else is already used (and registered) in ComputeResidual.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
pcarruscag: We have 3 votes for 2.0/3.0 already, I sense a replace all coming.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -320,12 +322,61 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      
+      /*--- Add all terms to the residuals ---*/
+      Residual[0] += yinv*Volume*(pk_axi-ck_axi+dk_axi);
+      Residual[1] += yinv*Volume*(pw_axi-cw_axi+dw_axi);
+    
+      /*--- Add contribution to the jacobian for implicit time integration---*/
+      Jacobian_i[0][0] += yinv*Volume*(sigma_k_i/zeta*TurbVar_Grad_i[0][1]-V_i[2]);
+      Jacobian_i[0][1] -= yinv*Volume*sigma_k_i*TurbVar_i[0]*TurbVar_Grad_i[0][1]/(zeta*zeta);
+      Jacobian_i[1][0] += yinv*Volume*sigma_k_i/zeta*TurbVar_Grad_i[1][1];
+      Jacobian_i[1][1] -= yinv*Volume*(sigma_k_i*TurbVar_i[0]*TurbVar_Grad_i[1][1]/(zeta*zeta)+V_i[2]);
```
pcarruscag, line 376: All the terms added to the normal SST Jacobian are negative to avoid imbalancing the diagonal dominance of the matrix.
Do you have a feel for how large the axi symmetric contributions might be compared with the normal ones? (I guess larger near the axis).

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
pcarruscag: Got it, I think it is always best to add it together with the code, while the reference results are fresh in your mind.
If you add it later you might find yourself having to debug again because of some change somewhere else.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
FlorianDm: Oupsi haha

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
FlorianDm: They may have the same magnitude as the normal ones and may be negative or positive depending on the sign of the y velocity and the gradients. Yes the further from the axis the more the problem becomes just normal 2D. I will try different versions with nothing more added to the jacobian or only convection or only diffusion (or setting anything positive to zero perhaps but that seems no good idea).

pcarruscag: The "clamping things with min/max" to have the sign you want is very common.
We have a config option for "accurate flux Jacobians" that we use for other schemes, you can also base the decision to clamp or not on that option to make things more flexible.

pcarruscag (approved):
Aside those choices with the Jacobian all LGTM :+1:

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
clarkpede: > since the current state of affairs is no more validated or tested for performance and strictly wrong as it is incomplete. 

It's cases like these (small changes) where unit tests are handy: https://su2code.github.io/docs_v7/Writing-Unit-Tests/

FlorianDm:
About the jacobian I could not find a situation where it made a difference (I tried a lot) so I commented it out.

The residual itself does matter however. I added a test case which uses the same mesh as the NICFD nozzle tutorial. It is an axisymmetric transonic air nozzle without a shock. Pretty much any change to the axisymmetric flow or turbulence residuals would break it.

i created a pull request on the test cases repo develop branch https://github.com/su2code/TestCases/pull/59 but then I just went ahead and merged it into develop since there is no review policy

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
FlorianDm: Not sure about this but I don't really see the need for unit tests as this bit is structurally very straightforward code and its behaviour is tested in regression now anyway

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -320,12 +322,55 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+
+    /*--- Add terms to the residuals ---*/
+    Residual[0] += yinv*Volume*(pk_axi-cdk_axi);
+    Residual[1] += yinv*Volume*(pw_axi-cdw_axi);
+  
+    /*--- Add contribution to the jacobian for implicit time integration (ignore to conserve diagonal dominance) ---*/
+    //Jacobian_i[0][0] += yinv*Volume*(sigma_k_i*TurbVar_Grad_i[0][1]/zeta-V_i[2]);
+    //Jacobian_i[0][1] += yinv*Volume*(-sigma_k_i*k*TurbVar_Grad_i[0][1]/(zeta*zeta));
+    //Jacobian_i[1][0] += yinv*Volume*(sigma_w_i*TurbVar_Grad_i[1][1]/zeta);
+    //Jacobian_i[1][1] += yinv*Volume*(-sigma_w_i*k*TurbVar_Grad_i[1][1]/(zeta*zeta)-V_i[2]);
```
pcarruscag, line 371: no commented out code please

TestCases/axisymmetric_rans/air_nozzle/air_nozzle.cfg
```diff
@@ -0,0 +1,256 @@
...
+MG_POST_SMOOTH= ( 0, 0, 0, 0 )
+%
+% Jacobi implicit smoothing of the correction
+MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0 )
+%
+% Damping factor for the residual restriction
+MG_DAMP_RESTRICTION= 0.75
+%
+% Damping factor for the correction prolongation
+MG_DAMP_PROLONGATION= 0.75
```
pcarruscag, line 185: Please simplify the config by removing options that are not needed, this is to make it easier to maintain the testcases if there are changes to some options.

pcarruscag:
Thank you for adding a regression.
Why do you need both a restart_flow.dat and a solution_flow.dat file in Testcases? We should try to keep that repo small.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
FlorianDm: Ok removed

TestCases/axisymmetric_rans/air_nozzle/air_nozzle.cfg
FlorianDm: done

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
pcarruscag: Unit tests detect change rather than correctness.
It is possibly that while fixing a bug a new one is introduced, regression values change, people check a few of them (it is boring work so no one tests all of them, especially if the residuals change very little), people update the regressions...
It has happened before, it will happen again.
With unit tests it is harder for that to happen, depending on the size of the "unit"...

pcarruscag: First sentence was wrong,  text was edited.

FlorianDm: Alright I can see how that could happen but can't come up with a good unit test for this part. Keeping this in mind for the future though

clarkpede: I agree with @pcarruscag.  I brought them up because "small changes that aren't complete" is a perfect use case for unit tests.  Unit tests can be added to verify correctness before code is ready for a complete regression test.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
TobiKattmann: Make it one vote more :)

PR #1197, opened 2021-02-13
Cleanup python wrapper, replace legacy python FSI

pcarruscag:
## Proposed Changes
Trying to make the python vertex tractions and vertex traction adjoints independent of a specific marker to allow #1174 to go forward without side effects on how FSI is done via the python wrapper at the moment.

## Related Work
#1174 

## PR Checklist


SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -225,11 +225,20 @@ void CFVMFlowSolverBase<V, R>::Allocate(const CConfig& config) {
...
 
-  /*--- Only initialize when there is a Marker_Fluid_Load defined
-   *--- (this avoids overhead in all other cases while a more permanent structure is being developed) ---*/
-  if ((config.GetnMarker_Fluid_Load() > 0) && (MGLevel == MESH_0)) {
-    Alloc3D(nMarker, nVertex, nDim, VertexTraction);
-    if (config.GetDiscrete_Adjoint()) Alloc3D(nMarker, nVertex, nDim, VertexTractionAdjoint);
+  if (MGLevel == MESH_0) {
+    VertexTraction.resize(nMarker);
+    for (iMarker = 0; iMarker < nMarker; iMarker++) {
+      if (config.GetSolid_Wall(iMarker))
```
pcarruscag, line 231: The idea is to allocate and compute for all solid walls since this is not very expensive.

SU2_CFD/src/iteration/CDiscAdjFluidIteration.cpp
```diff
@@ -390,7 +389,7 @@ void CDiscAdjFluidIteration::InitializeAdjoint(CSolver***** solver, CGeometry***
     solver[iZone][iInst][MESH_0][ADJRAD_SOL]->SetAdjoint_Output(geometry[iZone][iInst][MESH_0], config[iZone]);
   }
 
-  if (interface_boundary) {
+  if (config[iZone]->GetFluidProblem()) {
     solver[iZone][iInst][MESH_0][FLOW_SOL]->SetVertexTractionsAdjoint(geometry[iZone][iInst][MESH_0], config[iZone]);
   }
```
pcarruscag, line 394: Need to check if including this in the adjoint outputs does not break multizone adjoints.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -367,68 +306,50 @@ bool CDriver::ComputeVertexForces(unsigned short iMarker, unsigned long iVertex)
...
+  /*--- Get the normal at the vertex: this normal goes inside the fluid domain. ---*/
+  Normal = geometry_container[ZONE_0][INST_0][MESH_0]->vertex[iMarker][iVertex]->GetNormal();
+  Area = GeometryToolbox::Norm(nDim, Normal);
 
-    /*--- Calculate the inviscid (pressure) part of tn in the fluid nodes (force units) ---*/
-    for (iDim = 0; iDim < nDim; iDim++) {
-     PyWrapNodalForce[iDim] = -(Pn-Pinf)*Normal[iDim];     //NB : norm(Normal) = Area
-    }
+  /*--- Get the values of pressure and viscosity ---*/
+  Pn = solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->GetNodes()->GetPressure(iPoint);
```
Nicola-Fonzi, line 317: This is a bit different from the function compute_vertex_traction in Csolver.cpp. For example, the Pinf there is equal to config->GetPressure_FreeStreamND and everything is later multiplied for a factor accounting for adimensionalisation. Should this part be modified to be the same?

Nicola-Fonzi: In any case, if we now compute all the tractions inside SU2 itself, it there still the need for this function?
I think it is only used in an old test case...

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -445,34 +366,24 @@ passivedouble CDriver::GetVertexForceDensityZ(unsigned short iMarker, unsigned l
...
 }
 
 void CDriver::SetVertexCoordY(unsigned short iMarker, unsigned long iVertex, passivedouble newPosY) {
 
-  unsigned long iPoint;
-  su2double *Coord;
-
-  iPoint = geometry_container[ZONE_0][INST_0][MESH_0]->vertex[iMarker][iVertex]->GetNode();
-  Coord = geometry_container[ZONE_0][INST_0][MESH_0]->nodes->GetCoord(iPoint);
+  auto iPoint = geometry_container[ZONE_0][INST_0][MESH_0]->vertex[iMarker][iVertex]->GetNode();
```
Nicola-Fonzi, line 377: These functions also were all intended to be used with the old mesh solver. Now that we do not provide anymore the change in coordinates to the geometry container but to the mesh solver they should not be used anymore.

SU2_CFD/src/python_wrapper_structure.cpp
pcarruscag: If it is not too much work adapt the old case we can remove this.

SU2_CFD/src/python_wrapper_structure.cpp
pcarruscag: If it does not work it should be removed.

SU2_CFD/src/python_wrapper_structure.cpp
Nicola-Fonzi: The old test case uses a FSIdriver that I cannot find anymore, not sure if it would work in any case

SU2_CFD/src/python_wrapper_structure.cpp
Nicola-Fonzi: okok, I will do that

SU2_CFD/src/python_wrapper_structure.cpp
Nicola-Fonzi: My bad, the test case is indeed used, not with this FSIdriver, but with the singlezone. It is a forced response.
I will try update the test case to the new routines

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -914,116 +616,34 @@ void CSinglezoneDriver::SetInitialMesh() {
...
                        passivedouble LoadY, passivedouble LoadZ) {
 
   unsigned long iPoint;
-  PyWrapNodalForce[0] = LoadX;
-  PyWrapNodalForce[1] = LoadY;
-  PyWrapNodalForce[2] = LoadZ;
+  vector<su2double> NodalForce (3,0.0);
+  NodalForce[0] = LoadX;
+  NodalForce[1] = LoadY;
+  NodalForce[2] = LoadZ;
```
pcarruscag, line 643: The pattern in these functions is to load the 3 separate components into an array or vector to then pass by pointer to the solver.
Since we know the number of components (from the function signature) it is better to use a fixed-size temporary for that, e.g. su2double NodalForce[3]; or array<su2double,3> NodalForce, since that avoids dynamic allocation when this function is called for each vertex.
For the functions that return small vectors we would ideally do the same, return a fixed size type (the only choice there would be array<su2double,3>) but maybe python/swig has some limitation.

Nicola-Fonzi: Thank you very much for the comment!
I do not think swig is limited in that sense, the limitation was mine not perfectly understanding the difference, I see now.
I am changing it

SU2_CFD/src/iteration/CDiscAdjFluidIteration.cpp
pcarruscag: They did not break.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -59,201 +56,127 @@ void CDriver::PythonInterface_Preprocessing(CConfig **config, CGeometry ****geom
...
-  PyWrapNodalHeatFlux[0] = 0.0;
-  PyWrapNodalHeatFlux[1] = 0.0;
-  PyWrapNodalHeatFlux[2] = 0.0;
 
 }
 
-passivedouble CDriver::Get_Drag() {
+/////////////////////////////////////////////////////////////////////////////
+/* Functions related to the global performance indices (Lift, Drag, ecc..) */
+/////////////////////////////////////////////////////////////////////////////
```
pcarruscag, line 64: Good to see those temporaries going away and the different methods organized :+1:

PR #1199, opened 2021-02-15
Fix grid velocities in dual-time RANS simulations with deforming grids (aeroelastic and legacy FSI)

pcarruscag:
## Proposed Changes
Found while debugging differences between legacy and new mesh deformation with @Nicola-Fonzi in #1197.
Explained below.

## Related Work
#1197 

## PR Checklist

### DIFF
--- a/SU2_CFD/src/integration/CIntegration.cpp
+++ b/SU2_CFD/src/integration/CIntegration.cpp
@@ -208,14 +208,10 @@ void CIntegration::Time_Integration(CGeometry *geometry, CSolver **solver_contai
 
 }
 
-void CIntegration::SetDualTime_Solver(CGeometry *geometry, CSolver *solver, CConfig *config, unsigned short iMesh) {
+void CIntegration::SetDualTime_Geometry(CGeometry *geometry, CSolver *mesh_solver, const CConfig *config, unsigned short iMesh) {
 
   SU2_OMP_PARALLEL
   {
-  /*--- Store old solution, volumes and coordinates (in case there is grid movement). ---*/
-
-  solver->GetNodes()->Set_Solution_time_n1();
-  solver->GetNodes()->Set_Solution_time_n();
 
   geometry->nodes->SetVolume_nM1();
   geometry->nodes->SetVolume_n();
@@ -224,6 +220,20 @@ void CIntegration::SetDualTime_Solver(CGeometry *geometry, CSolver *solver, CCon
     geometry->nodes->SetCoord_n1();
     geometry->nodes->SetCoord_n();
   }
+
+  if ((iMesh==MESH_0) && config->GetDeform_Mesh()) mesh_solver->SetDualTime_Mesh();
+
+  } // end SU2_OMP_PARALLEL
+}
+
+void CIntegration::SetDualTime_Solver(const CGeometry *geometry, CSolver *solver, const CConfig *config, unsigned short iMesh) {
+
+  SU2_OMP_PARALLEL
+  {
+  /*--- Store old solution, volumes and coordinates (in case there is grid movement). ---*/
+
+  solver->GetNodes()->Set_Solution_time_n1();
+  solver->GetNodes()->Set_Solution_time_n();
 
   SU2_OMP_MASTER
   solver->ResetCFLAdapt();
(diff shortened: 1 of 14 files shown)


SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -225,6 +221,20 @@ void CIntegration::SetDualTime_Solver(CGeometry *geometry, CSolver *solver, CCon
...
+}
+
+void CIntegration::SetDualTime_Solver(const CGeometry *geometry, CSolver *solver, const CConfig *config, unsigned short iMesh) {
+
+  SU2_OMP_PARALLEL
+  {
+  /*--- Store old solution, volumes and coordinates (in case there is grid movement). ---*/
+
+  solver->GetNodes()->Set_Solution_time_n1();
+  solver->GetNodes()->Set_Solution_time_n();
```
pcarruscag, line 236: This function would push the coordinates and volumes back in time.
This function is called 2 times with the same geometry in RANS simulations.
Grid velocities would then be ~1.5 times what they should be.
This was not a problem with the new mesh solver, but the volumes would also be incorrect.

Just saw I need to update the comment... FML

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -208,14 +208,10 @@ void CIntegration::Time_Integration(CGeometry *geometry, CSolver **solver_contai
 
 }
 
-void CIntegration::SetDualTime_Solver(CGeometry *geometry, CSolver *solver, CConfig *config, unsigned short iMesh) {
+void CIntegration::SetDualTime_Geometry(CGeometry *geometry, CSolver *mesh_solver, const CConfig *config, unsigned short iMesh) {
```
pcarruscag, line 211: Now there is a specific function for dual time geometry updates.

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -172,13 +172,15 @@ void CFluidIteration::Update(COutput* output, CIntegration**** integration, CGeo
...
+                                                                        config[val_iZone], iMesh);
+
       integration[val_iZone][val_iInst][FLOW_SOL]->SetConvergence(false);
     }
 
-    /*--- Update dual time solver for the dynamic mesh solver ---*/
-    if (config[val_iZone]->GetDeform_Mesh()) {
-      solver[val_iZone][val_iInst][MESH_0][MESH_SOL]->SetDualTime_Mesh();
-    }
+    SetDualTime_Aeroelastic(config[val_iZone], iMesh);
```
pcarruscag, line 183: The aeroelastic feature was also broken since it has its own dual time update which was also repeated for RANS

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -199,6 +201,15 @@ void CFluidIteration::Update(COutput* output, CIntegration**** integration, CGeo
...
     }
+
+    /*--- Update dual time solver for the weakly coupled energy equation ---*/
+
+    if (config[val_iZone]->GetWeakly_Coupled_Heat()) {
+      integration[val_iZone][val_iInst][HEAT_SOL]->SetDualTime_Solver(geometry[val_iZone][val_iInst][MESH_0],
+                                                                      solver[val_iZone][val_iInst][MESH_0][HEAT_SOL],
+                                                                      config[val_iZone], MESH_0);
+      integration[val_iZone][val_iInst][HEAT_SOL]->SetConvergence(false);
+    }
```
pcarruscag, line 212: @TobiKattmann @oleburghardt , I think this was missing for dual time weakly coupled stuff, can you see if it makes sense?

Nicola-Fonzi:
Great fix! The code is also much better organised now.

Nicola-Fonzi:
Screenshot 2021-02-15 165247

The error was spotted while updating the old python routines to the new drivers/mesh solver. In particular, the test case in py_wrapper folder, related to a flat plate plunging up and down, showed a different coefficient of lift as computed using the old and new mesh solver.

I attach here the comparison of the coefficient of lift history for three cases:

* The legacy mesh solver, after the modification in this PR
* The new mesh solver
* The legacy mesh solver, before the modification

The first two are completely coincident

SU2_CFD/src/iteration/CFluidIteration.cpp
TobiKattmann: LGTM. Although I didn't use the `WEAKLY_COUPLED_HEAT` in a long time tbh and for now do not intend to use in the future. 
I also checked: There is a Testcase `discadj_heat` using `WEAKLY_COUPLED_HEAT` (although it is of course not unsteady)

Thanks for actualy thinking of that 👍

TobiKattmann (approved):
Thanks for fixing that. Good thing to split up geometry and solver  " solution pushing" and making use of inheritance for the former and specialisation if needed for the latter. 
And the `const`ification of the code is a good thing as well 💐

SU2_CFD/src/iteration/CFluidIteration.cpp
pcarruscag: Alright, thanks for having a look. I don't blame you, there is not a lot going for weakly coupled if the flow solver still drags temperature around. Otherwise having 4 variables (in 3D) would make computers very happy.

TobiKattmann: Ok i dont get the last sentence :) Do you mean the unnecessary energy-equation that is carried around even for cases without it being active.  Or did I miss the memo that states `4>>5` for #equations. Please elaborate if you have a few moments to spare ✍️

pcarruscag: Yes that too since it makes the linear system 1.44 times larger and slower (cost of sparse operations is proportional to amount of memory used).
Then 4 is a really good number because all the blocks in the matrix would be aligned and all the operations would pack perfectly into AVX registers.
The vector blocks would also be aligned which would make prefetching (loading into cache ahead of time) effective because it operates on cache lines (and 8 doubles in one of those).
Finally the output vector could be written to with non-temporal stores (algo because of alignment) which improves caching of the input vector.
So in all these ways 4 > 5 (by about 5-10% maybe on top of the 1.44).

TobiKattmann: Oh yes sure, if forgot that computers had this thing going with powers of 2 🤦 thanks!

PR #1200, opened 2021-02-17
Fix on wall distance calculation from symmetry plane of mesh deformation

Nicola-Fonzi:
## Proposed Changes

In the calculation of the stiffness for the mesh deformation, if the symmetry plane for the deformation is too stiff, cells may get distorted and the simulation may crash.

This occurred when the wall distance stiffness was used, and the symmetry plane for the deformation was also a solid wall boundary. This small fix should avoid the issue.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CMeshSolver.cpp
+++ b/SU2_CFD/src/solvers/CMeshSolver.cpp
@@ -298,7 +298,7 @@ void CMeshSolver::SetWallDistance(CGeometry *geometry, CConfig *config) {
 
   unsigned long nVertex_SolidWall = 0;
   for(auto iMarker=0u; iMarker<config->GetnMarker_All(); ++iMarker) {
-    if(config->GetSolid_Wall(iMarker)) {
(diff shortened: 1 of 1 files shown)


SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -315,7 +315,7 @@ void CMeshSolver::SetWallDistance(CGeometry *geometry, CConfig *config) {
 
   for (unsigned long iMarker=0, ii=0, jj=0; iMarker<config->GetnMarker_All(); ++iMarker) {
 
-    if (!config->GetSolid_Wall(iMarker)) continue;
+    if (!config->GetSolid_Wall(iMarker) && !config->GetMarker_All_Deform_Mesh_Sym_Plane(iMarker)) continue;
```
pcarruscag, line 318:
```suggestion
    if (!config->GetSolid_Wall(iMarker) || config->GetMarker_All_Deform_Mesh_Sym_Plane(iMarker)) continue;
```

pcarruscag (approved):

PR #1201, opened 2021-02-18
fix typo Linael -> Linear

TobiKattmann:
## Proposed Changes
Description of `LINEAR_SOLVER_ILU_FILL_IN` wrote Linael instead of Linear .. and that was copied quite a few times
Done using `find ./ -type -f -exec sed -i -e 's/Linael/Linear/g' {} \;` in ./Testcases on Linux

As in none of the appearances the option is not using the default of zero, that option could be removed from all those testcases imo, but I put that up to discussion. As @pcarruscag mentioned in #1195 `[...]simplify the config by removing options that are not needed, this is to make it easier to maintain the testcases if there are changes to some options.`

Feel free to chip any other typos here. 

## Related Work
Seen in #1195 and annoyed me enough to open a PR, but of course not introduced there

### DIFF
--- a/config_template.cfg
+++ b/config_template.cfg
@@ -1019,7 +1019,7 @@ LINEAR_SOLVER_PREC= ILU
 % Same for discrete adjoint (JACOBI or ILU), replaces LINEAR_SOLVER_PREC in SU2_*_AD codes.
 DISCADJ_LIN_PREC= ILU
 %
-% Linael solver ILU preconditioner fill-in level (0 by default)
+% Linear solver ILU preconditioner fill-in level (0 by default)
 LINEAR_SOLVER_ILU_FILL_IN= 0
 %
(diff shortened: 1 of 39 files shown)


config_template.cfg
```diff
@@ -1019,7 +1019,7 @@ LINEAR_SOLVER_PREC= ILU
 % Same for discrete adjoint (JACOBI or ILU), replaces LINEAR_SOLVER_PREC in SU2_*_AD codes.
 DISCADJ_LIN_PREC= ILU
 %
-% Linael solver ILU preconditioner fill-in level (0 by default)
+% Linear solver ILU preconditioner fill-in level (0 by default)
 LINEAR_SOLVER_ILU_FILL_IN= 0
```
pcarruscag, line 1023: Oh no... even the template xD

WallyMaier (approved):

Issue #1155, opened 2021-01-05
fix wall model for incompressible SA and SST turbulence models


bigfooted:
**Is your feature request related to a problem? Please describe.**

The currently implemented wall model for SA is not working robustly, and needs be modified for incompressible flows. The model has not been implemented for SST but can easily be implemented from the SA wall model. The SA wall model implementation in @EduardoMolina 'feature_WallModelLES' works well with almost no modifications and the implementation can be based on this version. 

**Describe the solution you'd like**
Current implementation of wall model extended to incompressible flows for SA and SST.
Also needed is proper computation of y+ and wall shear stress in output files.
Validation test case replicating literature performance. Currently available to me: Buice & Eaton Diffuser.

**Describe alternatives you've considered**
-

**Additional context**
I performed a preliminary implementation and validation for isothermal walls (Buice & Eaton diffuser).

PR #1204, opened 2021-02-22
Fix SA and SST wall functions

bigfooted:
## Proposed Changes
 
Fixes (I think) the current implementation of the wall model. Some work still needs to be done, regarding compressible vs incompressible and validation of heated walls. 

## Related Work

related to #1155 
Also took some ideas from feature_WallModelLES

## PR Checklist


SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2535,6 +2559,10 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
   bool axisymmetric = config->GetAxisymmetric();
   bool roughwall = (config->GetnRoughWall() > 0);
   bool nemo = config->GetNEMOProblem();
+  bool wallfunctions=false;
+
+  for (iDim = 0; iDim < nDim; iDim++) RefVel2 += Velocity_Inf[iDim] * Velocity_Inf[iDim];
```
TobiKattmann, line 2564:
```suggestion
RefVel2 = GeometryToolbox::SquaredNorm(nDim, Velocity_Inf);
```
Then you can prob also make it const su2double right there and get rid of the definition above. 
And you need to include the geometry_toolbox.hpp if not already there

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  unsigned short max_iter = 100;
+  su2double tol = 1e-12;
+
+  /*--- Get the freestream velocity magnitude for non-dim. purposes ---*/
+
+  su2double *VelInf = config->GetVelocity_FreeStreamND();
+  su2double VelInfMod = 0.0;
+  for (auto iDim = 0u; iDim < nDim; iDim++)
+    VelInfMod += VelInf[iDim];
+  VelInfMod = sqrt(VelInfMod);
```
TobiKattmann, line 426:
```suggestion
  /*--- Get the freestream velocity magnitude for non-dim. purposes ---*/

const su2double VelInfMod = GeometryToolbox::Norm(nDim, config->GetVelocity_FreeStreamND());
```
Not sure if `VelInfMod += VelInf[iDim];` is correct here or if it should be the squared velinf

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+          P_Normal = nodes->GetPressure(Point_Normal);
+          T_Normal = nodes->GetTemperature(Point_Normal);
+
+          /*--- Compute the wall-parallel velocity at first point off the wall ---*/
+
+          //su2double VelNormal = 0.0;
+          //for (auto iDim = 0u; iDim < nDim; iDim++)
+          //  VelNormal += Vel[iDim] * UnitNormal[iDim];
+     
+          su2double VelNormal = GeometryToolbox::DotProduct(int(MAXNDIM), Vel, UnitNormal);
```
TobiKattmann, line 501: why `nDim` in the Norm above and here `int MAXNDIM`. Sure it makes the size known at compile time (i.e. maybe a bit faster) or are there other reasons?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+              TauElem[iDim] += tau[iDim][jDim]*UnitNormal[jDim];
+          }
+
+          /*--- Compute wall shear stress as the magnitude of the wall-tangential
+           component of the shear stress tensor---*/
+
+          TauNormal = 0.0;
+          for (auto iDim = 0u; iDim < nDim; iDim++)
+            TauNormal += TauElem[iDim] * UnitNormal[iDim];
+
```
TobiKattmann, line 570: I guess TauNormal and WallShearStress can use the geometry toolbox as well

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -480,6 +491,77 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
 }
 
+
+void CTurbSSTSolver::SetNuTilde_WF(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics,
+                                  CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
+
+  const unsigned short max_iter = 100;
+
+  /* --- tolerance has LARGE impact on convergence, do not increase this value! --- */
+  const su2double tol = 1e-12;
```
TobiKattmann, line 501: I won't 🤞 ;)

TobiKattmann (commented):
Hi, just a few things from scrolling through the code. But I like that you stay close to the given paper and also sometimes give equation numbers. Otherwise there is still a lot of commented code but that's what WIP is for ;)

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -158,6 +158,9 @@ class CFVMFlowSolverBase : public CSolver {
   su2double** CPressureTarget = nullptr;   /*!< \brief Target Pressure coefficient for each boundary and vertex. */
   su2double** YPlus = nullptr;             /*!< \brief Yplus for each boundary and vertex. */
 
+  su2double** UTau = nullptr;              /*!< \brief UTau for each boundary and vertex. */
+  su2double** EddyViscWall = nullptr;      /*!< \brief Wall Eddy Viscosity for each boundary and vertex. */
+
```
pcarruscag, line 163: I changed all these things in my blitz PR, you should be able to replace su2double** by vector<vector<su2double> >, delete the destructor part and everything should work.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -2432,7 +2435,7 @@ class CFVMFlowSolverBase : public CSolver {
    */
   inline su2double GetCSkinFriction(unsigned short val_marker, unsigned long val_vertex,
                                     unsigned short val_dim) const final {
-    return CSkinFriction[val_marker][val_dim][val_vertex];
+    return CSkinFriction[val_marker][val_vertex][val_dim];
```
pcarruscag, line 2438: I like this change, always wondered why the switched order only for this variable.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -3929,6 +3945,29 @@ class CSolver {
...
+                                    CConfig* config) { }
+
+  /*!
+   * \brief A virtual member.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   */
+  inline virtual void SetNuTilde_WF(CGeometry *geometry,
+                                    CSolver **solver_container,
```
pcarruscag, line 3965: These two do not need to be virtual I believe, they are part of the preprocessing and so they can be just private methods of their respective solvers.
Ideally abstract interfaces (like CSolver) should have as few public methods as possible, to make clear what needs to be implemented to define a "solver".

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -624,39 +624,62 @@ CNumerics::ResidualType<> CSourceWindGust::ComputeResidual(const CConfig* config
 
   u_gust = WindGust_i[0];
   v_gust = WindGust_i[1];
+// w_gust = WindGust_i[2];
 
   if (GustDir == X_DIR) {
     du_gust_dx = WindGustDer_i[0];
     du_gust_dy = WindGustDer_i[1];
+    //du_gust_dz = WindGustDer_i[2];
     du_gust_dt = WindGustDer_i[2];
```
pcarruscag, line 633: kudos for picking up "feature requests" from CFD online :+1:

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  } else if (config->GetRef_Inc_NonDim() == REFERENCE_VALUES) {
+    RefDensity = config->GetInc_Density_Ref();
+    RefVel2 = config->GetInc_Velocity_Ref() * config->GetInc_Velocity_Ref();
+  }
+
+  bool debug = true;
+  unsigned long counter;
+
+  su2double UnitNormal[3];
+  su2double **grad_primvar, tau[3][3];
```
pcarruscag, line 404: Use [MAXNDIM] instead of [3] please. It is not so much that we want to run 4D one day, but MAXNDIM and MAXNVAR make it unambiguous whether the variable is related with the geometry or the solution.
Also make grad_primvar `const su2double* const*` for peace of mind.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  su2double Density_Wall, T_Wall, P_Wall, Lam_Visc_Wall, Tau_Wall = 0.0;
+  su2double diff, Delta, grad_diff;
+  su2double U_Tau, U_Plus, Gam, Beta, Phi, Q, Y_Plus_White, Y_Plus;
+  su2double TauNormal, WallShearStress;
+  su2double Gas_Constant = config->GetGas_ConstantND();
+  su2double Cp = (Gamma / Gamma_Minus_One) * Gas_Constant;
+  su2double Eddy_Visc;
+  su2double Vector[MAXNDIM] = {0.0};
+  unsigned short max_iter = 100;
+  su2double tol = 1e-12;
```
pcarruscag, line 418: constexpr the former, const the latter.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  const su2double kappa = 0.41;
+  const su2double B = 5.5;
+  const su2double cv1_3 = 7.1*7.1*7.1;
+
+  bool converged = true;
+
+  for (auto iMarker = 0u; iMarker < config->GetnMarker_All(); iMarker++) {
+
+    if ((config->GetMarker_All_KindBC(iMarker) == HEAT_FLUX) ||
+        (config->GetMarker_All_KindBC(iMarker) == ISOTHERMAL) ) {
```
pcarruscag, line 443: Use config->GetViscous_Wall(iMarker) (or something like that) to future proof the code for new boundaries.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+        const auto Point_Normal = geometry->vertex[iMarker][iVertex]->GetNormal_Neighbor();
+
+        /*--- Check if the node belongs to the domain (i.e, not a halo node)
+         and the neighbor is not part of the physical boundary ---*/
+
+
+          /*--- Get coordinates of the current vertex and nearest normal point ---*/
+
+          su2double *Coord = geometry->nodes->GetCoord(iPoint);
+          su2double *Coord_Normal = geometry->nodes->GetCoord(Point_Normal);
```
pcarruscag, line 471: const these too

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+
+          su2double UnitNormal[MAXNDIM] = {0.0};
+          for (auto iDim = 0u; iDim < nDim; iDim++)
+            UnitNormal[iDim] = -Normal[iDim]/Area;
+
+          /*--- Get the velocity, pressure, and temperature at the nearest
+           (normal) interior point. ---*/
+          su2double Vel[MAXNDIM] = {0.0};
+          for (auto iDim = 0u; iDim < nDim; iDim++)
+            Vel[iDim] = nodes->GetVelocity(Point_Normal,iDim);
```
pcarruscag, line 490: Good, this is safer than the "iDim+1" and makes it easier to eventually extend the implementation to the other flow solvers.

SU2_CFD/src/solvers/CIncNSSolver.cpp
pcarruscag: I did that in some other places too, it was a bit of a micro optimization.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+          for (auto iDim = 0u; iDim < nDim; iDim++) {
+            for (auto jDim = 0u ; jDim < nDim; jDim++) {
+              Delta = 0.0; if (iDim == jDim) Delta = 1.0;
+              tau[iDim][jDim] = Lam_Visc_Wall*(  grad_primvar[jDim+1][iDim]
+                                               + grad_primvar[iDim+1][jDim]) -
+              TWO3*Lam_Visc_Wall*div_vel*Delta;
+            }
+            for (auto jDim = 0u; jDim < nDim; jDim++)
+              TauElem[iDim] += tau[iDim][jDim]*UnitNormal[jDim];
+          }
```
pcarruscag, line 562: There is a method in CNumerics to compute the stress tensor that can be used directly.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+
+            U_Tau = U_Tau - 0.5*(diff / grad_diff);
+
+            counter++;
+
+            /* --- if we do not converge, then something serious is going on and we stop --- */
+
+            if (counter == max_iter) {
+              SU2_MPI::Error("Y+ did not converge within the max number of iterations!",CURRENT_FUNCTION);
+            }
```
pcarruscag, line 648: Maybe we can be a bit more forgiving to avoid blowing up during initial solution transients, e.g. count the number of failed points and report at the end of the function without throwing an error.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -366,3 +378,304 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+
+            Y_Plus_White = exp((kappa/sqrt(Gam))*(asin((2.0*Gam*U_Plus - Beta)/Q) - Phi))*exp(-1.0*kappa*B);
+
+            /*--- Spalding's universal form for the BL velocity with the
+             outer velocity form of White & Christoph above. ---*/
+
+            Y_Plus = U_Plus + Y_Plus_White - (exp(-1.0*kappa*B)*
+                                              (1.0 + kappa*U_Plus + kappa*kappa*U_Plus*U_Plus/2.0 +
+                                               kappa*kappa*kappa*U_Plus*U_Plus*U_Plus/6.0));
+
```
pcarruscag, line 620: When I cleaned up the compressible solver I wrote something like:
```
su2double kUp = kappa*U_Plus;
su2double Y_Plus = U_Plus + Y_Plus_White - exp(-1.0*kappa*B) * (1.0 + kUp*(1.0 + 0.5*kUp + pow(kUp,2)/6.0));
```

pcarruscag (commented):
Yey wall functions. I know it is still WIP but few comments below.

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: The latter can't be constexpr because su2double is not a [LiteralType](https://en.cppreference.com/w/cpp/named_req/LiteralType)?

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
bigfooted: yes, I'm still getting used to using the toolbox :-) as well as getting used to not working with variable definitions at the top of the function instead of at the point where they are used for the first time.

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: Yes, I copied a line of code from elsewhere, is there really a difference in the end?

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: yep, working on replacing all of this, thanks for pointing this out!

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
bigfooted: 1e-10 was the the original value and if you look at the residual plots, changing to 1e-12 was a big improvement!

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
bigfooted: Thanks for the remark, I don't like to work with ********variables either: we have c++ for that after all. Personally, I'd like to see a complete move from new..delete to vector-only (but I conform to the local programming standard as far as that exists). So can we say that it is best practice to replace all su2double* and su2double** to vector and vector<vector> ?

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
bigfooted: Yes, it looked weird. Maybe because of cache misses for 2D arrays that do not fit in L1 cache, but I think this is now mostly dealt with by the compiler optimization?  Not sure if this is still an issue on modern architecture.

SU2_CFD/src/numerics/flow/flow_sources.cpp
bigfooted: :-)

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: Thanks, it's a habit that's hard to break...

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: oh yes, computestresstensor, I totally missed this/forgot about this!

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: yes, maybe just output a "warning: XX points did not converge"

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: vector vector if the the number of columns varies, otherwise su2activematrix which is all contiguous memory (vector vector is as bad for performance as **).

SU2_CFD/src/solvers/CIncNSSolver.cpp
pcarruscag: I confess I am not aware of all the formalities.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: ** may have a very different layout in memory than [][].
Compilers can do very little about dynamically allocated variables.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: Oh and please never conform ;)

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: it was already there and is not used, I will remove it

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -410,6 +410,17 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
 
+
+    /*--- Evaluate nu tilde at the closest point to the surface using the wall functions. ---*/
+
+    if (config->GetWall_Functions()) {
+      SU2_OMP_MASTER
+      SetNuTilde_WF(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
+      SU2_OMP_BARRIER
+      return;
+    }
```
bigfooted, line 421: @pcarruscag I have to confess that I do not know how openmp works, I'm not using it. This piece of code fails. This code was already there, so this was never working properly I guess? Do you know why this fails?

pcarruscag: You need to move it out of the iVertex loop.

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: I am disappointed Pedro 😞 jk

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: I personally wouldn't worry about this ... but maybe Pedro is preparing some big openFOAM-like 3D-everything push to get rid of all those run-time evaluated nDim loops 🤔 :D

pcarruscag: For sure
```
template<size_t nDim>
class SU2 {

};
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
bigfooted: true, so no problem to make this change.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
bigfooted: ok, that's what I suspected. Will do that...

SU2_CFD/include/solvers/CSolver.hpp
bigfooted: yes, changed it.

bigfooted:
I updated the code based on your comments. I have to rerun the validation test case that I have to see if it still matches with the experimental data.  I also have to check the compressible vs incompressible implementation.

SU2_CFD/include/solvers/CTurbSSTSolver.hpp
```diff
@@ -42,6 +42,25 @@ class CTurbSSTSolver final : public CTurbSolver {
   kine_Inf,              /*!< \brief Free-stream turbulent kinetic energy. */
   omega_Inf;             /*!< \brief Free-stream specific dissipation. */
 
+
+
+/*!
```
pcarruscag, line 47:
```suggestion
  /*!
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -586,3 +599,275 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  su2double RefDensity,RefVel2; 
+  if ((config->GetRef_Inc_NonDim() == DIMENSIONAL) || (config->GetRef_Inc_NonDim() == INITIAL_VALUES)) {
+    RefDensity = Density_Inf;
+    RefVel2 = 0.0;
+    for (auto iDim = 0u; iDim < nDim; iDim++) 
+      RefVel2 += Velocity_Inf[iDim] * Velocity_Inf[iDim];
+  } else if (config->GetRef_Inc_NonDim() == REFERENCE_VALUES) {
+    RefDensity = config->GetInc_Density_Ref();
+    RefVel2 = config->GetInc_Velocity_Ref() * config->GetInc_Velocity_Ref();
+  }
```
pcarruscag, line 619: This needs to exist in only one place, I fixed a recent bug due to having this logic in 4 different functions.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -586,3 +599,275 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+  const su2double kappa = 0.41; // put model constants in config file
+  const su2double B = 5.5;
+
+
+  for (auto iMarker = 0u; iMarker < config->GetnMarker_All(); iMarker++) {
+
+    if (!config->GetViscous_Wall(iMarker)) continue;
+
+    if ((config->GetMarker_All_KindBC(iMarker) == HEAT_FLUX) ||
+        (config->GetMarker_All_KindBC(iMarker) == ISOTHERMAL) ) {
```
pcarruscag, line 647:
```suggestion
    if (config->GetViscous_Wall(iMarker)) {
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -586,3 +599,275 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+          /*--- Compute the shear stress at the wall in the regular fashion
+           by using the stress tensor on the surface ---*/
+
+          su2double tau[MAXNDIM][MAXNDIM] = {{0.0}}, TauElem[MAXNDIM] = {0.0};
+          su2double Lam_Visc_Wall = nodes->GetLaminarViscosity(iPoint);
+          CNumerics::ComputeStressTensor(nDim, tau, nodes->GetGradient_Primitive(iPoint)+1, Lam_Visc_Wall);
+
+          for (auto iDim = 0u; iDim < nDim; iDim++) {
+            TauElem[iDim] = GeometryToolbox::DotProduct(nDim, tau[iDim], UnitNormal);
+          }
```
pcarruscag, line 750: :+1:

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -586,3 +599,275 @@ void CIncNSSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **sol
...
+
+        /*--- Check if the node belongs to the domain (i.e, not a halo node)
+         and the neighbor is not part of the physical boundary ---*/
+
+        if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+          /*--- Get coordinates of the current vertex and nearest normal point ---*/
+
+          const auto Coord = geometry->nodes->GetCoord(iPoint);
+          const auto Coord_Normal = geometry->nodes->GetCoord(Point_Normal);
```
pcarruscag, line 679: Indentation of the code after the if, i.e. move code back a couple spaces please.
the reason I started using `continue` was to not indent the code so much to the right, we only use 2 space indents, after a few of those it was difficult to see what is what.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -727,45 +727,70 @@ void CNSSolver::BC_Isothermal_Wall_Generic(CGeometry *geometry, CSolver **solver
...
 
+  //bool converged = true;
+
   for (auto iMarker = 0u; iMarker < config->GetnMarker_All(); iMarker++) {
 
     if (!config->GetViscous_Wall(iMarker)) continue;
 
+    if ((config->GetMarker_All_KindBC(iMarker) == HEAT_FLUX) ||
+        (config->GetMarker_All_KindBC(iMarker) == ISOTHERMAL) ) {
+
```
pcarruscag, line 780: viscous wall

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -583,8 +583,13 @@ void CTurbSASolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, CN
       conv_numerics->SetPrimitive(V_domain, V_inlet);
 
       /*--- Set the turbulent variable states (prescribed for an inflow) ---*/
+
+      //Solution_i[0] = nodes->GetSolution(iPoint,0);
+
```
pcarruscag, line 588: What happened here? bug?

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -374,12 +374,30 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
+
+    SU2_OMP_FOR_STAT(OMP_MIN_SIZE)
+    for (auto iVertex = 0u; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+
+      SU2_OMP_MASTER
+      SetNuTilde_WF(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
+      SU2_OMP_BARRIER
+    }
+    return;
+  }
```
pcarruscag, line 391:
```suggestion
  if (config->GetWall_Functions()) {
    SU2_OMP_MASTER
    SetNuTilde_WF(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
    SU2_OMP_BARRIER
    return;
  }
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -452,6 +470,76 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
+                                  CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
+
+  const unsigned short max_iter = 100;
+
+  /* --- tolerance has LARGE impact on convergence, do not increase this value! --- */
+  const su2double tol = 1e-12;
+
+  /*--- Typical constants from boundary layer theory ---*/
+
+  const su2double kappa = 0.4;
```
pcarruscag, line 484: Aren't some of the constants defined at solver level?

pcarruscag:
Thanks @bigfooted, I fixed the conflicts (albeit not very well judging from the broken compilation) and left some more comments some more important than others.
For me the most important ones are those regarding avoiding any and all repetition, reference factors and so on.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
bigfooted: True, not this one though. It is however also defined in the turbulence solvers with a slightly different value (0.40 vs 0.41, but in the literature you also find 5.0) I think the best is to make this a config option with a good default value, and people can change it if they really want to. In any case such magic numbers should be collected at a central location and not be hidden. It is even less hidden if it is a config option.

SU2_CFD/src/solvers/CTurbSASolver.cpp
bigfooted: not used, it was used further down in the part "commented because of serious convergence problems".

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: I compared compressible and incompressible and made it more homogeneous, and also use this in incompressible now.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2558,7 +2584,12 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
       /*--- Compute y+ and non-dimensional velocity ---*/
 
       FrictionVel = sqrt(fabs(WallShearStress[iMarker][iVertex]) / Density);
-      YPlus[iMarker][iVertex] = WallDistMod * FrictionVel / (Viscosity / Density);
+
+      /* --- in case of wall functions, we have computed YPlus in the turbulence class --- */
+      if (!wallfunctions)  
+        YPlus[iMarker][iVertex] = WallDistMod * FrictionVel / (Viscosity / Density);
+
+
```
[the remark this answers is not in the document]
bigfooted: Is there a standard oneliner that I and everybody else can run on all files to autoformat everything?

SU2_CFD/include/solvers/CNSSolver.hpp
```diff
@@ -114,7 +114,18 @@ class CNSSolver final : public CEulerSolver {
...
+   * \brief computes nu-tilde.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void SetNuTilde_WF(CGeometry *geometry,
+                                    CSolver **solver_container,
+                                    CNumerics *conv_numerics,
+                                    CNumerics *visc_numerics,
+                                    CConfig *config,
```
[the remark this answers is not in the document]
bigfooted: :s/CConfig/const CConfig/  basically?

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
TobiKattmann: Maybe document those values there in the comment as well. just for safety

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -157,6 +157,11 @@ class CFVMFlowSolverBase : public CSolver {
   vector<vector<su2double> > CPressure;         /*!< \brief Pressure coefficient for each boundary and vertex. */
   vector<vector<su2double> > CPressureTarget;   /*!< \brief Target Pressure coefficient for each boundary and vertex. */
   vector<vector<su2double> > YPlus;             /*!< \brief Yplus for each boundary and vertex. */
+  vector<vector<su2double> > UTau;              /*!< \brief UTau for each boundary and vertex. */
+  vector<vector<su2double> > EddyViscWall;      /*!< \brief Eddy viscosuty at the wall for each boundary and vertex. */
+
+
+
```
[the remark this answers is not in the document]
TobiKattmann: :D ... one can even directly commit from that suggestion ... I get the feeling the guys at github know what they're doing

SU2_CFD/src/solvers/CTurbSASolver.cpp
pcarruscag: I see, we should delete all that commented out code at some point... it is unmaintainable by definition, please remove these new lines with comments then.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1595,123 +1594,59 @@ void CTurbSASolver::SetNuTilde_WF(CGeometry *geometry, CSolver **solver_containe
...
         cout << "WARNING: Nu_tilde evaluation has not converged." << endl;
         break;
       }
     }
 
-    nodes->SetSolution_Old(Point_Normal, 0, nu_til_old);
-    LinSysRes.SetBlock_Zero(Point_Normal);
+    su2double Sol[nVar];
+    for (auto iVar = 0u; iVar < nVar; iVar++)
+      Sol[iVar] = nu_til;
```
pcarruscag, line 1642: This dynamic c-style array is not standard c++, you should use MAXNVAR which is a compile time constant, making the array static, or in the case of SA which we know to have only 1 variable... no arrays are needed.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -452,6 +470,76 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
+    omega0 = U_Tau/(sqrt(0.09)*kappa*y);                  // eq. 20 
+    omega_new = sqrt(omega0*omega0 + omega1*omega1);      // eq. 21 Nichols & Nelson
+    k_new = omega_new * Eddy_Visc/Density_Wall;           // eq. 22 Nichols & Nelson 
+                                                            // (is this the correct density? paper says rho and not rho_w)
+
+    // put some relaxation factor on the k-omega values
+    k += 0.5*(k_new - k);       
+    omega += 0.5*(omega_new - omega);       
+
+    su2double solution[nVar];
```
pcarruscag, line 527: MAXNVAR here too

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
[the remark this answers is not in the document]
pcarruscag: No we have not made it that drastic yet. If you install clang-format on your machine you can do `clang-format "file you want" > "result"` then replace the original with the formatted.
This was done to CFVMFlowSolverBase when the file was created, but it has not been done for a while now, so probably it would introduce changes that are not related with what you are working on... But you can try it and see what it looks like.

SU2_CFD/src/solvers/CTurbSASolver.cpp
bigfooted: changed nvar to MAXNVAR.

SU2_CFD/include/solvers/CNSSolver.hpp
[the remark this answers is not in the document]
pcarruscag: Yes -> https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-const

Although in this case this method does not seem to be implemented (yet?) so maybe the declaration can be removed.

SU2_CFD/include/solvers/CTurbSSTSolver.hpp
```diff
@@ -42,6 +42,25 @@ class CTurbSSTSolver final : public CTurbSolver {
...
+   * \param[in] visc_numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
+  void SetNuTilde_WF(CGeometry *geometry,
+                     CSolver **solver_container,
+                     CNumerics *conv_numerics,
+                     CNumerics *visc_numerics,
+                     CConfig *config,
+                     unsigned short val_marker);
```
pcarruscag, line 61: I would name this method after the kw variables it manipulates instead, the argument list can also probably be reduced? e.g. the numerics are not used in its implementation.

pcarruscag: Since the method is not virtual it is not forced to follow the naming and signature of any other method of any solver.

bigfooted:
I also made the von Karman constant a config option, together with the "B" constant. The von Karman constant is used in other places as well, but maybe its best to keep lets say all SA or SST coefficients as a complete set of coefficients and not have only one coefficient change value.

Common/src/CConfig.cpp
```diff
@@ -1223,6 +1223,10 @@ void CConfig::SetConfig_Options() {
   addDoubleOption("PRANDTL_LAM", Prandtl_Lam, 0.72);
   /*!\brief PRANDTL_TURB \n DESCRIPTION: Turbulent Prandtl number (0.9 (air), only for compressible flows) \n DEFAULT 0.90 \ingroup Config*/
   addDoubleOption("PRANDTL_TURB", Prandtl_Turb, 0.90);
+  /*!\brief WALLMODELKAPPA \n DESCRIPTION: von Karman constant used for the wall model \n DEFAULT 0.41 \ingroup Config*/
+  addDoubleOption("WALLMODELKAPPA", wallModelKappa, 0.41);
+  /*!\brief WALLMODELB \n DESCRIPTION: constant B used for the wall model \n DEFAULT 5.0 \ingroup Config*/
+  addDoubleOption("WALLMODELKAPPA", wallModelKappa, 0.41);
```
pcarruscag, line 1229: :+1: for putting it in the config (you missed B)
Is it common to specify a specific von Karman constant for wall models? Or should it be simply called VONKARMANCONST?

bigfooted: Ha, guess I pushed and then called it a day.

I think the value is pretty fixed, usually 0.40 or 0.41, for every turbulence model that uses it. But I  don't want to give a user the impression that you can change every occurrence of this constant in all the turbulence models in the code.  So for now, it's only for the wall model, which is reflected in the name.

pcarruscag: Makes sense

bigfooted:
update: validated with the incompressible solver with SA and SST, will add that as a validation and tutorial. To do: validation with a compressible test case.
Future work: validation for heated walls. Less data is available for validation. Let me know if you have a good test case.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2533,7 +2550,12 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
       WallShearStress[iMarker][iVertex] = 0.0;
       for (iDim = 0; iDim < nDim; iDim++) {
         TauTangent[iDim] = TauElem[iDim] - TauNormal * UnitNormal[iDim];
-        CSkinFriction[iMarker](iVertex,iDim) = TauTangent[iDim] * factorFric;
+        /* --- in case of wall functions, we have computed the skin friction in the turbulence solver --- */
+        /* --- Note that in the wall model, we switch off the computation when the computed y+ < 5    --- */
+        /* --- We put YPlus to 1.0 so we have to compute skinfriction and the actual y+ in that case as well --- */
+        if (!wallfunctions || (wallfunctions && YPlus[iMarker][iVertex] < 5.0))
```
pcarruscag, line 2556: This threshold should be defined somewhere central.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -365,12 +365,25 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
   if (WallType == ROUGH ) rough_wall = true;
 
+
+
+  /*--- Evaluate nu tilde at the closest point to the surface using the wall functions. ---*/
+
+  if (config->GetWall_Functions()) {
+    SU2_OMP_MASTER
+    SetTurbVars_WF(geometry, solver_container, config, val_marker);
+    SU2_OMP_BARRIER
```
pcarruscag, line 375: Next time you merge develop you will get some failed compilations with errors about "end macros", for example in this case you will need END_SU2_OMP_MASTER before the barrier.

pcarruscag (commented):
It looks like you committed some changes to medi.
This happens a lot when we update it because github does not do it automatically when you checkout or merge another branch. You can fix this by:
git checkout develop
git submodule update externals/medi
git checkout "this branch"
git add externals/medi
git commit

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
pcarruscag: Maybe together with other wall function parameters like number of iterations, tolerance, relaxation (even if they are not accessible via the config).

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2471,6 +2484,10 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
     su2double ForceViscous[MAXNDIM] = {0.0}, MomentViscous[MAXNDIM] = {0.0};
     su2double MomentX_Force[MAXNDIM] = {0.0}, MomentY_Force[MAXNDIM] = {0.0}, MomentZ_Force[MAXNDIM] = {0.0};
 
+    /* --- check if wall functions are used --- */
+   
+    const bool wallfunctions = (config->GetWallFunction_Treatment(Marker_Tag) != NO_WALL_FUNCTION);
```
TobiKattmann, line 2489: While you are at it: can you make this an enum class following #1258 and naming the enum e.g. `WALL_FUNCTIONS`

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
bigfooted: yes, this is a threshold from the paper, but should be user-adjustable as well...

Common/include/option_structure.hpp
```diff
@@ -935,23 +935,23 @@ static const MapType<string, ENUM_ROELOWDISS> RoeLowDiss_Map = {
...
+  NONE                      ,   /*!< \brief No wall function treatment, integration to the wall. Default behavior. */
+  STANDARD_WALL_FUNCTION    ,   /*!< \brief Standard wall function. */
+  ADAPTIVE_WALL_FUNCTION    ,   /*!< \brief Adaptive wall function. Formulation depends on y+. */
+  SCALABLE_WALL_FUNCTION    ,   /*!< \brief Scalable wall function. */
+  EQUILIBRIUM_WALL_MODEL    ,   /*!< \brief Equilibrium wall model for LES. */
+  NONEQUILIBRIUM_WALL_MODEL ,   /*!< \brief Non-equilibrium wall model for LES. */
+  LOGARITHMIC_WALL_MODEL        /*!< \brief Logarithmic law-of-the-wall model for LES. */
+};
+static const MapType<string, WALL_FUNCTIONS> Wall_Functions_Map = {
+  MakePair("NONE",                      WALL_FUNCTIONS::NONE)
```
pcarruscag, line 948: You can call the enum case NONE but let's keep the NO_WALL_FUNCTION user string.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -722,33 +722,47 @@ void CNSSolver::BC_Isothermal_Wall_Generic(CGeometry *geometry, CSolver **solver
...
   BC_Isothermal_Wall_Generic(geometry, solver_container, conv_numerics, nullptr, config, val_marker, true);
 }
 
 void CNSSolver::SetTauWall_WF(CGeometry *geometry, CSolver **solver_container, const CConfig *config) {
+/*---  The wall function implemented herein is based on Nichols and Nelson AIAAJ v32 n6 2004.
+   At this moment, the wall function is only available for adiabatic flows.
+   ---*/
+  su2double RefDensity=Density_Inf,RefVel2=0.0; 
+  if ((config->GetRef_Inc_NonDim() == DIMENSIONAL) || (config->GetRef_Inc_NonDim() == INITIAL_VALUES)) {
+    for (auto iDim = 0u; iDim < nDim; iDim++)
```
pcarruscag, line 739: Using options for the incompressible solver in the compressible solver.

SU2_CFD/include/solvers/CNSSolver.hpp
```diff
@@ -114,7 +114,18 @@ class CNSSolver final : public CEulerSolver {
...
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void SetTurbVars_WF(CGeometry *geometry,
+                                    CSolver **solver_container,
+                                    CNumerics *conv_numerics,
+                                    CNumerics *visc_numerics,
+                                    const CConfig *config,
+                                    unsigned short val_marker) ;
```
pcarruscag, line 128: It looks like this is declared but not implemented.
(I saw you removed the WIP hence my commenting) It would be good to cleanup commented out code before merging this, development nodes and such, leaving only very pressing TODO: (e.g. isothermal walls)

.vscode/launch.json
```diff
@@ -0,0 +1,46 @@
...
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "diffuser",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "/home/nijso/Codes/SU2/bin/SU2_CFD",
+            "args": ["diffuser.cfg"],
+            "stopAtEntry": false,
+            "cwd": "/home/nijso/Codes/su2cases/validation/buice_eaton_diffuser/SST_coarse",
```
pcarruscag, line 14: This file is probably not needed.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -429,6 +437,12 @@ void CFVMFlowSolverBase<V, R>::Viscous_Residual_impl(unsigned long iEdge, CGeome
     numerics->SetTurbKineticEnergy(turbNodes->GetSolution(iPoint,0),
                                    turbNodes->GetSolution(jPoint,0));
 
+  /*--- Wall shear stress values (wall functions) ---*/
+
+  numerics->SetTauWall(nodes->GetTauWall(iPoint),
+                       nodes->GetTauWall(jPoint));
```
pcarruscag, line 443: Please add or modify some test cases (comp and incomp) I will need to modify the vectorized numerics to support this and it would be helpful to have a good reference.

bigfooted: OK, I have the flat plate test case for this (Schubauer & Klebanoff), will add this

config_template.cfg
```diff
@@ -914,10 +914,10 @@ MARKER_PLOTTING = ( airfoil )
 MARKER_MONITORING = ( airfoil )
 %
 % Viscous wall markers for which wall functions must be applied. (NONE = no marker)
-% Format: ( marker name, wall function type -NO_WALL_FUNCTION, STANDARD_WALL_FUNCTION,
+% Format: ( marker name, wall function type -NONE, STANDARD_WALL_FUNCTION,
```
pcarruscag, line 917:
```suggestion
% Format: ( marker name, wall function type -NO_WALL_FUNCTION, STANDARD_WALL_FUNCTION,
```

TestCases/py_wrapper/flatPlate_unsteady_CHT/unsteady_CHT_FlatPlate_Conf.cfg
```diff
@@ -212,7 +212,7 @@ MARKER_MONITORING = ( plate )
 %
 % Viscous wall markers for which wall functions must be applied. (NONE = no marker)
 % Format: ( marker name, wall function type, ... )
-MARKER_WALL_FUNCTIONS= ( plate, NO_WALL_FUNCTION )
+MARKER_WALL_FUNCTIONS= ( plate, NONE )
```
pcarruscag, line 215:
```suggestion
MARKER_WALL_FUNCTIONS= ( plate, NO_WALL_FUNCTION )
```

config_template.cfg
```diff
@@ -914,10 +914,10 @@ MARKER_PLOTTING = ( airfoil )
...
 %           ADAPTIVE_WALL_FUNCTION, SCALABLE_WALL_FUNCTION, EQUILIBRIUM_WALL_MODEL,
 %           NONEQUILIBRIUM_WALL_MODEL-, ... )
-MARKER_WALL_FUNCTIONS= ( airfoil, NO_WALL_FUNCTION )
+MARKER_WALL_FUNCTIONS= ( airfoil, NONE )
```
pcarruscag, line 920:
```suggestion
MARKER_WALL_FUNCTIONS= ( airfoil, NO_WALL_FUNCTION )
```

SU2_CFD/src/numerics/flow/flow_diffusion.cpp
```diff
@@ -622,8 +599,24 @@ CNumerics::ResidualType<> CAvgGradInc_Flow::ComputeResidual(const CConfig* confi
...
+                        Mean_GradPrimVar+1, Mean_PrimVar[nDim+2], Mean_Eddy_Viscosity,
+                        Mean_turb_ke, MeanPerturbedRSM);
+  }
+
   /*--- Get projected flux tensor (viscous residual) ---*/
-  SetStressTensor(Mean_PrimVar, Mean_GradPrimVar, Mean_turb_ke, Mean_Laminar_Viscosity, Mean_Eddy_Viscosity);
+  SetStressTensor(Mean_PrimVar, Mean_GradPrimVar, Mean_turb_ke,
+                  Mean_Laminar_Viscosity, Mean_Eddy_Viscosity);
+  if (config->GetQCR()) AddQCR(nDim, &Mean_GradPrimVar[1], tau);
+  if (Mean_TauWall > 0) AddTauWall(UnitNormal, Mean_TauWall);
```
pcarruscag, line 619: I've been looking into making the wall functions compatible with the vectorized classes.
I got confused, however, because it looks like the tau wall was not considered in the incompressible viscous numerics.
But your validation results looked good... So, either this makes them better, or it is not needed (here and in the other 2 viscous classes that use it).

pcarruscag: In #1120 @EduardoMolina also added the "tau wall logic" to the incompressible class but he changed the implementation in the 3 classes to only consider tau wall if only one of the two points has it.
Which means that for edges connecting two wall points tau wall is not considered... might be worth re-running some tests?

bigfooted: I think in the case of RANS wall functions, the tauwall values that you mention are always overwritten by the tauwall computed at the end of CIncNSSolver::SetTauWall_WF(..). The function addtauwall is used only for the LES implementation.

pcarruscag: Tau wall is computed in that function as part of the preprocessing, to then be used in the numerics, I don't see any other place in the code where the computed tau wall can influence the solution.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -809,10 +798,12 @@ void CIncNSSolver::SetTauWall_WF(CGeometry *geometry, CSolver **solver_container
           U_Tau = U_Tau - relax*(diff / grad_diff);
 
           counter++;
-
           if (counter > max_iter) {
             notConvergedCounter++;
-            cout << "Warning: Y+ did not converge within the max number of iterations!" << endl;
```
pcarruscag: I was trying this on the CRM and these messages were showing up a lot, so I made some changes to print only one message with totals.
I also copied the "safe values" you had for the compressible version.

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -866,20 +861,19 @@ void CNSSolver::SetTauWall_WF(CGeometry *geometry, CSolver **solver_container, c
...
 
-      if (Y_Plus_Start < 5.0) {
-        cout << "skipping stress due to wall model" << endl;
-        continue;
+      if (Y_Plus_Start < Y_Plus) {
+        skipCounter++;
       }
-
-      while (fabs(diff) > tol) {
+      else while (fabs(diff) > tol) {
```
pcarruscag: I also made this logic the same as the incomp solver, it seemed that certain quantities would not be computed otherwise.
Hopefully this was the right thing to do and not the opposite.

pcarruscag (approved):
Just that test case to go I guess.

SU2_CFD/src/solvers/CIncNSSolver.cpp
bigfooted: Hi Pedro, Yes, they are present only for the first 10 iterations or so usually. Maybe we can introduce verbosity levels for such messages.

bigfooted:
Thanks @pcarruscag and @TobiKattmann ! I will add the testcase and some small things in a new PR!

SU2_CFD/src/numerics/flow/flow_sources.cpp
pcarruscag: Did you ever check if the 3D version worked? Otherwise we gained deadcode for no reason...

PR #1205, opened 2021-02-24
[WIP] Fixes to axi nemo set functions and auxvargrad in source residual

jtneedels:
fixed set functions for NEMO axisymmetric, implemented auxvargrad correction in nemo axisymmetric source residual, fixed fluid model/transport model print statement

## Proposed Changes
This PR addresses an issue found with not setting primitive and conservative variables in computing the axisymmetric source residual, as well as implementing the auxvargrad indexing correction applied in the standard air axisymmetric source residual and fixing an issue with printing the fluid and transport models. 

## Related Work
This PR is related to #1192 and #1162.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
+++ b/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
@@ -376,7 +376,7 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
                                                      +v*TWO3*(2*PrimVar_Grad_i[nSpecies+2][1]-PrimVar_Grad_i[nSpecies+2][0]
                                                      -v*yinv+rho*turb_ke_i))
                                                      -total_conductivity_i*PrimVar_Grad_i[nSpecies][1])
-                                                     -TWO3*(AuxVar_Grad_i[1][1]+AuxVar_Grad_i[2][1]));
+                                                     -TWO3*(AuxVar_Grad_i[1][1]+AuxVar_Grad_i[2][0]));
     residual[nSpecies+3] -= Volume*(yinv*(sumJeve_y -qy_ve));
   }
 
--- a/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
+++ b/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
@@ -372,12 +372,12 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
                                                                     -TWO3*AuxVar_Grad_i[0][0]);
     residual[nSpecies+1] -= Volume*(yinv*total_viscosity_i*2*(PrimVar_Grad_i[nSpecies+3][1]-v*yinv)
                                                                     -TWO3*AuxVar_Grad_i[0][1]);
-    residual[nSpecies+2] -= Volume*(yinv*(sumJhs_y + total_viscosity_i*(u*(PrimVar_Grad_i[nSpecies+3][0]+PrimVar_Grad_i[nSpecies+2][1])
+    residual[nSpecies+2] -= Volume*(yinv*(-sumJhs_y + total_viscosity_i*(u*(PrimVar_Grad_i[nSpecies+3][0]+PrimVar_Grad_i[nSpecies+2][1])
                                                      +v*TWO3*(2*PrimVar_Grad_i[nSpecies+2][1]-PrimVar_Grad_i[nSpecies+2][0]
                                                      -v*yinv+rho*turb_ke_i))
                                                      -total_conductivity_i*PrimVar_Grad_i[nSpecies][1])
                                                      -TWO3*(AuxVar_Grad_i[1][1]+AuxVar_Grad_i[2][0]));
-    residual[nSpecies+3] -= Volume*(yinv*(sumJeve_y -qy_ve));
+    residual[nSpecies+3] -= Volume*(yinv*(-sumJeve_y -qy_ve));
   }
 
 //  if (implicit) {
--- a/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
@@ -938,6 +938,21 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
         /*--- If necessary, set variables needed for viscous computation ---*/
         if (viscous) {
 
+          // TODO: NOTE: Some of these are set above. They need to be set here as well, otherwise they are passed in incorrectly.
+          // Do not remove.
+          
+          /*--- Set volume ---*/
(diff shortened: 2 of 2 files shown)


SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -938,6 +938,21 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
         /*--- If necessary, set variables needed for viscous computation ---*/
         if (viscous) {
 
+          // TODO: NOTE: Some of these are set above. They need to be set here as well, otherwise they are passed in incorrectly.
+          // Do not remove.
+          
+          /*--- Set volume ---*/
+          numerics->SetVolume(geometry->nodes->GetVolume(iPoint));
```
WallyMaier, line 945: I think this is fine.  We might need to look into a way to avoid repeating "set" functions in the future.

WallyMaier (commented):
Thanks for fixing these issues, @jtneedels.  Ill let someone else check, but LGTM.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1306,8 +1321,8 @@ void CNEMOEulerSolver::SetNondimensionalization(CConfig *config, unsigned short
     ModelTableOut <<"-- Models:"<< endl;
 
     ModelTable.AddColumn("Mixture", 25);
-    ModelTable.AddColumn("Transport Model", 25);
     ModelTable.AddColumn("Fluid Model", 25);
+    ModelTable.AddColumn("Transport Model", 25);
```
WallyMaier, line 1325: Thanks for aligning these!

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -938,6 +938,21 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
         /*--- If necessary, set variables needed for viscous computation ---*/
         if (viscous) {
 
+          // TODO: NOTE: Some of these are set above. They need to be set here as well, otherwise they are passed in incorrectly.
+          // Do not remove.
+
```
pcarruscag, line 943: This is not something that just happens.
Most likely scenario is that you have a memory leak, which should be easy to find with valgrind.
Compile a debug or debugoptimized version, choose a small testcase that uses this feature, run
valgrind SU2_CFD your_config.cfg
see where it leads you.

pcarruscag: Any luck @jtneedels ? We were planning to release 7.1.1 but we can also wait a little more.

jtneedels: @pcarruscag I haven't resolved it yet, but I am working on it more today. What is the timeline for the version 7.1.1 release? I don't want to hold it up.

pcarruscag: We don't have strict deadlines but we try to have these maintenance releases every month.
This month is shorter, so if you want a specific day lets say Monday.

jtneedels: Okay sounds good. I'm definitely seeing a memory leak, and a number of uninitialized warnings in CNEMOEulerVariable.cpp and CNEMOEulerSolver.cpp. Still working on finding what's causing this specific issue.

jtneedels: Hi Pedro, I'm having trouble pinning down the memory leak causing the issue, but it looks like it might be coming from the Auxvar_grad terms used in the source residual. I'm having trouble getting specific line numbers when running valgrind with: valgrind --leak-check=full --track-origins=yes SU2_CFD testcase.cfg and compiling with debugoptimized. Is there anything I need to add to my run command?

pcarruscag: debug instead of debug optimized might give more information... I do not know of anything else you may add.
Some of those warnings are infectious in that if something is not initialized all dependent quantities will also be a warning.
You need to start fixing from the top down. And then because of how the numerics (in general not just NEMO) are implemented the bug might be due to something else...

SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
```diff
@@ -372,12 +372,12 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
                                                                     -TWO3*AuxVar_Grad_i[0][0]);
     residual[nSpecies+1] -= Volume*(yinv*total_viscosity_i*2*(PrimVar_Grad_i[nSpecies+3][1]-v*yinv)
```
pcarruscag, line 373: ... for example in line 365 above (`sumJhs_y  += (rho*Ds[iSpecies]*GV[RHOS_INDEX+iSpecies][1] - V_i[RHOS_INDEX+iSpecies]*Vector[1]) * hs[iSpecies];`)  and also 366, where is Vector[1] populated?

jtneedels: Its populated in CNEMONumerics.cpp, should I try repopulating it in NEMO_sources?

pcarruscag: In `GetViscousProjFlux`? How does that value make its way into `CSource_NEMO::ComputeAxisymmetric` ?
Within each application of a public numerics method (like ComputeAxisymmetric or ComputeResidual) you need to be able to trace the result to the inputs, i.e. the things that are set into numerics from the outside before calling the method, or to the constructor, e.g. some constant that is stored in the constructor.
Otherwise you are computing things for one point based on values of another point.

This pattern we have in SU2 in general of "setting" followed by "compute" with very "stateful" objects (i.e. full of member variables used for intermediate results) is not very manageable unfortunately.
We have been getting away with it because most of these class hierarchies are simple, but for richer physics it is going to be headache after headache...

jtneedels: Yeah exactly in `GetViscousProjFlux`.  So in cases like this, where there aren't set/compute functions I can just call to update the value, would the best remedy be to explicitly recompute `Vector` in `CSource_NEMO::ComputeAxisymmetric`?

pcarruscag: Whenever you find yourself needing to do the same action in more than one location, make it a function.
Functions should have clear inputs and outputs.
Programming via side effects, i.e. you call a method of a class and it does some magic to a member variable that you use afterwards is a bad pattern.
The safeties the language offers to keep us from using that pattern are `const` and `static` qualifiers. `const` methods can only read from member variables, `static` can only use other static methods/variables.
"Make everything you can `const`" (every C++ author ever)

PR #1208, opened 2021-02-24
Fix skin friction coefficient

pcarruscag:
## Proposed Changes
I broke it when I defined the reference factor for aerodynamic coefficients in a central location.
Oh well... It would be good to add some regressions for output files, or to try to include more postprocessing quantities in the regressions by adding them to the screen outputs.

### DIFF
--- a/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
+++ b/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
@@ -2413,7 +2413,7 @@ class CFVMFlowSolverBase : public CSolver {
    */
   inline su2double GetCSkinFriction(unsigned short val_marker, unsigned long val_vertex,
                                     unsigned short val_dim) const final {
-    return CSkinFriction[val_marker][val_dim][val_vertex];
+    return CSkinFriction[val_marker](val_vertex,val_dim);
   }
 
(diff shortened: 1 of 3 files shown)


pcarruscag:
Looks like this bug was not released :) (Fortune favours the bold)

TobiKattmann (approved):
LGTM... and while we're at it, you mentioned yesterday an interview with a NASA guy concerning lgtm ()if I recall correctly). I would be interested in that source :)

And I guess you're onto sth concerning screen output  for reg tests... this `# last four columns` was understood as the rule I guess and sometimes that includes the iterations number :)

pcarruscag:
It was a code review by Jason Turner (https://www.youtube.com/watch?v=RbhufLudVsI&t=1743s&ab_channel=C%E1%90%A9%E1%90%A9WeeklyWithJasonTurner)

Quite a long video, I did not watch it all.

pcarruscag:
Yeah what I'm sayin is put more columns.

TobiKattmann:
Yeah, but listening to Mr Turner is surely well invested time if one is able to follow at  least some of his thoughts. But imo he does a good job in that, I sometimes tune in for cppcast and that is okayish to follow most of the time (but that is not really in depth tbh)

PR #1214, opened 2021-03-01
Hybrid Parallel AD (Part 1/?)

jblueh:
## Proposed Changes

The goal of this PR is to establish AD support for the OpenMP features of SU2.

AD support in SU2 is provided by [CoDiPack](https://www.scicomp.uni-kl.de/software/codi), which is coupled with [MeDiPack](https://www.scicomp.uni-kl.de/software/medi) for the differentiation of MPI. OpenMP support is added by coupling it in addition with [OpDiLib](https://www.scicomp.uni-kl.de/software/opdi) so that all in all, hybrid parallel AD is achieved.

This involves at least the following steps.
1. Incorporate OpDiLib into the AD workflow.
2. Establish thread-safety and parallelization of the discrete adjoint code.
3. Testing of hybrid parallel adjoints.
4. Performance optimizations.

## Related Work

OpenMP features introduced in #789, #824, #830, #834, #843, #861, #895, #975, #1178, possibly others

## PR Checklist


jblueh:
The commits so far address parts of OpDiLib's incorporation into the AD workflow and sort out some additional issues along the lines of the `disc_adj_fsi/Airfoild2d` testcase.  They provide a testing environment that is stable with respect to OpDiLib and can already be used for figuring out some of the more SU2 specific issues. However, they require that SU2 is built with a compiler with OMPT support, for example an up-to-date version of `clang++`, hence the failing CI builds.

pcarruscag:
It looks like the builds are failing because of swig (python wrapper) and a missing definition of size_t.
What does OMPT stand for? OpenMP threads? (Google was not helpful)

pcarruscag:
I see, then it should be possible to detect the version of the standard supported by the compiler and only enable the feature in that case. We do that for simd directives for compatibility (hopefully) with the MS compilers.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -477,7 +485,7 @@ void CSysMatrixComms::Complete(CSysVector<T>& x, CGeometry *geometry,
 template<class ScalarType>
 void CSysMatrix<ScalarType>::SetValZero() {
   const auto size = nnz*nVar*nEqn;
-  const auto chunk = roundUpDiv(size,omp_get_max_threads());
+  const auto chunk = roundUpDiv(size,omp_get_num_threads());
```
pcarruscag, line 488: Thank you for the best practice changes :+1:

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1818,6 +1823,7 @@ void CSolver::CompleteComms(CGeometry *geometry,
 
       SU2_OMP_MASTER
       SU2_MPI::Waitany(geometry->nP2PRecv, geometry->req_P2PRecv, &ind, &status);
+      END_SU2_OMP_MASTER
```
pcarruscag, line 1826: Is there a way in OpDiLib to debug missing end pragmas?

jblueh: There is no dedicated way yet, but I am working on it. Most of the macros involve unbalanced `{` or `}` so that improper pairing leads to incorrect code. This can be "debugged" with the compiler. However, there are also more subtle errors like using the wrong end macro. These can be "debugged" by investigating segfaults...

jblueh: OpDiLib offers a syntax checker now. It can also be used with application specific macros (as opposed to the OpDiLib ones) as they are used in SU2. I added a syntax file for this in 6aaebca2b7c3273ef365b57d8352a7339f9f6bfe. It is placed in the root directory for now, we should discuss the location. You can try it out from the root directory via
```
python externals/opdi/syntax/check.py su2omp.syntax.json Common SU2_CFD -p *.hpp *.cpp *.inl -r -s
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -318,13 +339,7 @@ class CFVMFlowSolverBase : public CSolver {
...
      *    clauses for compatibility with OpenMP 2.0 (Windows...). ---*/
 
-    SU2_OMP_MASTER
-    {
-      Min_Delta_Time = 1e30;
-      Max_Delta_Time = 0.0;
-      Global_Delta_UnstTimeND = 1e30;
-    }
-    SU2_OMP_BARRIER
+    ompMasterAssignBarrier(Min_Delta_Time,1e30, Max_Delta_Time,0.0, Global_Delta_UnstTimeND,1e30);
```
pcarruscag, line 342: Not sure if I made it better or worse :+1: :-1:

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -916,64 +920,82 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
...
+    END_SU2_OMP_MASTER
+    SU2_OMP_BARRIER
+
+    AD::FuncHelper->addToTape(CSysSolve_b<ScalarType>::Solve_b);
+    SU2_OMP_BARRIER
+
+    SU2_OMP_MASTER
     AD::EndExtFunc();
+    END_SU2_OMP_MASTER
+    SU2_OMP_BARRIER
```
pcarruscag, line 983:
```suggestion
```
Maybe this last barrier is not needed? Since end just releases the FuncHelper.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -178,64 +151,12 @@ void CDiscAdjSolver::SetRecording(CGeometry* geometry, CConfig *config){
...
-//  bool time_n_needed  = ((config->GetUnsteady_Simulation() == DT_STEPPING_1ST) ||
-//      (config->GetUnsteady_Simulation() == DT_STEPPING_2ND)),
-//  time_n1_needed = config->GetUnsteady_Simulation() == DT_STEPPING_2ND;
-
-//  unsigned long ExtIter = config->GetExtIter();
-
-  unsigned long iPoint;
-  unsigned short iDim;
-
-  /*--- Reset the solution to the initial (converged) position ---*/
```
pcarruscag, line 193: I am deleting dead code in the disc adj solvers as I move through this.
If you think I shouldn't speak now.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
jblueh: saved some lines :+1: nice C++ features :+1: fewer end macros to maintain :+1:
made the code hard to read without looking up the definition of `ompMasterAssignBarrier` :-1:

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
jblueh: If you put these changes into another pull request, they can be merged earlier.

pcarruscag: But that would have been smart... We don't do that here

pcarruscag:
Testing is going well on this side, the hybrid parallel version produces consistent results compared to serial and MPI-only.
I think the file diffs that are failing are because of the linear solver changes, together with refactoring the discrete adjoint solver.

Most of the point loops of the DiscAdjFluidIteration are now parallel and more importantly the dependencies are recorded in parallel.

How is the preaccumulation @jblueh ? Does it break because of those variables returned by value like you expected?

I still want to test the adjoint evaluation mode and get some preliminary performance numbers, seems ok on a 4 core machine, but we can integrate this PR soon (maybe as soon we have the pre-compilation checks for end macros, I can give it a look if you haven't yet?).
In fact we should integrate it soon, it does not have many lines but it touches many files...

jblueh:
Let's discuss these topics in the next dev meeting.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -3492,6 +3488,18 @@ class CSolver {
...
+   * \brief Loads the solution from the restart file.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] filename - Name of the restart file.
+   * \param[in] skipVars - Number of variables preceeding the solution.
+   */
+  void BasicLoadRestart(CGeometry *geometry,
+                        const CConfig *config,
+                        const string& filename,
+                        unsigned long skipVars);
```
pcarruscag, line 3501: I created a method in CSolver to load very simple restarts, i.e. only the solution and no additional pre/post processing. This is now used by the disc adj solvers.

SU2_CFD/src/iteration/CDiscAdjHeatIteration.cpp
```diff
@@ -227,9 +227,8 @@ void CDiscAdjHeatIteration::SetDependencies(CSolver***** solver, CGeometry**** g
 
 void CDiscAdjHeatIteration::RegisterOutput(CSolver***** solver, CGeometry**** geometry, CConfig** config,
                                            COutput* output, unsigned short iZone, unsigned short iInst) {
-  solver[iZone][iInst][MESH_0][ADJHEAT_SOL]->RegisterOutput(geometry[iZone][iInst][MESH_0], config[iZone]);
 
-  geometry[iZone][iInst][MESH_0]->RegisterOutput_Coordinates(config[iZone]);
+  solver[iZone][iInst][MESH_0][ADJHEAT_SOL]->RegisterOutput(geometry[iZone][iInst][MESH_0], config[iZone]);
```
pcarruscag, line 231: @oleburghardt I think this RegisterOutput_Coordinates was carried over from the fluid iteration but is not needed?

SU2_CFD/src/iteration/CDiscAdjFluidIteration.cpp
```diff
@@ -31,91 +31,90 @@
...
 
-        for (iMesh = 0; iMesh <= config[val_iZone]->GetnMGLevels(); iMesh++) {
-          solver[val_iZone][val_iInst][iMesh][FLOW_SOL]->GetNodes()->Set_Solution_time_n();
+        for (auto iMesh = 0u; iMesh <= config[iZone]->GetnMGLevels(); iMesh++) {
+          auto solvers = solver[iZone][iInst][iMesh];
+
+          solvers[FLOW_SOL]->GetNodes()->Set_Solution_time_n();
           if (turbulent) {
-            solver[val_iZone][val_iInst][iMesh][TURB_SOL]->GetNodes()->Set_Solution_time_n();
+            solvers[TURB_SOL]->GetNodes()->Set_Solution_time_n();
```
pcarruscag, line 101: There was "some" cleanup here to refer to solvers without [iZone][iInst][iMesh] all the time because it made me cry.

SU2_CFD/src/iteration/CDiscAdjFluidIteration.cpp
```diff
@@ -343,8 +349,11 @@ void CDiscAdjFluidIteration::Iterate(COutput* output, CIntegration**** integrati
                                      CSolver***** solver, CNumerics****** numerics, CConfig** config,
                                      CSurfaceMovement** surface_movement, CVolumetricMovement*** volume_grid_movement,
                                      CFreeFormDefBox*** FFDBox, unsigned short iZone, unsigned short iInst) {
-  bool frozen_visc = config[iZone]->GetFrozen_Visc_Disc();
-  bool heat = config[iZone]->GetWeakly_Coupled_Heat();
+
+  SU2_OMP_PARALLEL_(if(solver[iZone][iInst][MESH_0][ADJFLOW_SOL]->GetHasHybridParallel())) {
+
+  const bool frozen_visc = config[iZone]->GetFrozen_Visc_Disc();
```
pcarruscag, line 355: This type of action (ExtractAdjoint_Variables) was made hybrid parallel for the disc adj fluid iteration and solver but not for the FEA iteration because each fluid iteration is usually lighter than one structural iteration (and so it would be more of a bottleneck to run the fluid part only by one thread).

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1753,61 +1757,47 @@ void CEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_con
...
   ErrorCounter += SetPrimitive_Variables(solver_container, config);
+  SU2_OMP_BARRIER
 
-  if ((iMesh == MESH_0) && (config->GetComm_Level() == COMM_FULL)) {
-    SU2_OMP_BARRIER
-    SU2_OMP_MASTER
-    {
+  SU2_OMP_MASTER { /*--- Ops that are not OpenMP parallel go in this block. ---*/
+
+    if ((iMesh == MESH_0) && (config->GetComm_Level() == COMM_FULL)) {
```
pcarruscag, line 1770: There were some permutations here to avoid having too many END_MASTER macros, this way there is one section to put all the non OpenMP-ready preprocessing actions.

pcarruscag (commented):
This stage of the work is ready.
All seems to work fine for compressible flow problems and FEA, but testing will continue while this is up for review.
The feature is WIP and experimental but nothing should have been broken for primal+OpenMP or adjoint without OpenMP (i.e. regular).
It involves small changes to many files because of the "end macros" explained below, and for that reason we think it is better to merge into develop now, to then introduce the "yet another intermediate class" for scalar solvers.
And in future hybrid AD PR's address the remaining issues, namely:
- Regression tests;
- Fixing the preaccumulation section to be compatible with the different AD type required by hybrid parallel AD;
- Performance tunning, i.e. informing OpDiLib about sections of code where the access patterns to memory can be more aggressively exploited.

Common/src/linear_algebra/CSysSolve.cpp
jblueh: :+1:

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -972,7 +959,30 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
...
+      AD::FuncHelper->addUserData(&Jacobian);
+      AD::FuncHelper->addUserData(geometry);
+      AD::FuncHelper->addUserData(config);
+      AD::FuncHelper->addUserData(this);
+    }
+    END_SU2_OMP_MASTER
+    SU2_OMP_BARRIER
+
+    AD::FuncHelper->addToTape(CSysSolve_b<ScalarType>::Solve_b);
+    SU2_OMP_BARRIER
```
jblueh, line 980: Registering inputs, outputs and user data on one `codi::ExternalFunctionHelper` is not thread-safe. However, all participating threads add the external function to their tape. Likewise, all participating threads make calls to `AD::StartRecording` and `AD::StopRecording`.

Common/src/linear_algebra/CSysSolve_b.cpp
```diff
@@ -37,36 +37,40 @@ void CSysSolve_b<ScalarType>::Solve_b(const codi::RealReverse::Real* x, codi::Re
...
-  d->getData(geometry);
+  d->getDataByIndex(geometry, 3);
 
   const CConfig* config = nullptr;
-  d->getData(config);
+  d->getDataByIndex(config, 4);
 
   CSysSolve<ScalarType>* solver = nullptr;
-  d->getData(solver);
+  d->getDataByIndex(solver, 5);
```
jblueh, line 55: This is a thread-safe way to access external function data.

jblueh (commented):
`ad_structure.hpp` defines a clear AD interface that does not require significant changes for the application of OpDiLib. Most OpDiLib specifics can be hidden behind this interface. In addition. some parts of the code required thread-safety related changes.

SU2_CFD/src/output/filewriter/CParallelDataSorter.cpp
```diff
@@ -213,60 +193,43 @@ void CParallelDataSorter::SortOutputData() {
...
-    for (int jj = 0; jj < VARS_PER_POINT*nPoint_Recv[size]; jj++){
-      const passivedouble tmpVal = SU2_TYPE::GetValue(doubleBuffer[jj]);
-      passiveDoubleBuffer[jj] = tmpVal;
-      /*--- For some AD datatypes a call of the destructor is
-       *  necessary to properly delete the AD type ---*/
-      doubleBuffer[jj].~su2double();
-    }
-  }
-
-  /*--- Step 2: Reorder the data in the buffer --- */
```
pcarruscag, line 243: These last changes were to fix issues when the RealReverseIndex is used.
We used to have a char buffer that was used for both passive and active doubles, which required these explicit calls to the destructor, but I guess that did not work so well with mpi and all.
Now there is only passivedouble, which means that once something goes into the data sorter AD information is lost.
This would matter if some computation was performed with time averaged values, which does not seem to be the case at the moment...

SU2_CFD/include/output/filewriter/CParallelDataSorter.hpp
```diff
@@ -305,10 +316,10 @@ class CParallelDataSorter{
...
    */
   void SetUnsorted_Data(unsigned long iPoint, unsigned short iField, su2double data){
-    connSend[Index[iPoint] + iField] = data;
+    connSend[Index[iPoint] + iField] = SU2_TYPE::GetValue(data);
   }
 
-  su2double GetUnsorted_Data(unsigned long iPoint, unsigned short iField) const {
+  passivedouble GetUnsorted_Data(unsigned long iPoint, unsigned short iField) const {
     return connSend[Index[iPoint] + iField];
   }
```
pcarruscag, line 324: This is were the active-passive conversion takes place.

SU2_CFD/include/output/filewriter/CSurfaceFEMDataSorter.hpp
```diff
@@ -31,7 +31,7 @@
 
 class CSurfaceFEMDataSorter final: public CParallelDataSorter{
 
-  CFEMDataSorter* volumeSorter;                  //!< Pointer to the volume sorter instance
+  const CFEMDataSorter* volumeSorter;            //!< Pointer to the volume sorter instance
```
pcarruscag, line 34: also the usual cleanups

pcarruscag (commented):
I'll merge this soon if there are no objections.

pcarruscag (approved):

SU2_CFD/src/iteration/CDiscAdjFluidIteration.cpp
```diff
@@ -435,17 +454,19 @@ void CDiscAdjFluidIteration::RegisterInput(CSolver***** solver, CGeometry**** ge
...
 void CDiscAdjFluidIteration::SetRecording(CSolver***** solver, CGeometry**** geometry, CConfig** config,
                                           unsigned short iZone, unsigned short iInst, unsigned short kind_recording) {
-  bool frozen_visc = config[iZone]->GetFrozen_Visc_Disc();
 
-  /*--- Prepare for recording by resetting the solution to the initial converged solution ---*/
+  SU2_OMP_PARALLEL_(if(solver[iZone][iInst][MESH_0][ADJFLOW_SOL]->GetHasHybridParallel())) {
 
-  if (solver[iZone][iInst][MESH_0][ADJFEA_SOL]) {
-    solver[iZone][iInst][MESH_0][ADJFEA_SOL]->SetRecording(geometry[iZone][iInst][MESH_0], config[iZone]);
-  }
```
TobiKattmann, line 448: I already thought I dreamed of removing this code bit when I scrolled through #1257  and it was gone (because it was already deleted here) ... that conditional evaluated to true btw and it does not crash for non-FEA cases . Just c++ things

Issue #990, opened 2020-05-16
MPI partitioning issues for Turbomachinery runs


auzbaig:
The number of cores I can run Turbomachinery cases always behaves unpredictably. Here are my observations:

1. Ran my case with 1 or 2 cores it runs fine.

2. Any higher than that, it gives a segmentation fault or the following error:

---------------------- Turbomachinery Preprocessing ---------------------
Initialize Turbo Vertex Structure.
Error in "virtual void CPhysicalGeometry::ComputeNSpan(CConfig*, short unsigned int, short unsigned int, bool)":
-------------------------------------------------------------------------
 At the moment only turbomachinery with the same amount of span-wise section can be simulated
nSpan inflow 113, nSpan outflow 32510
------------------------------ Error Exit -------------------------------

3. I generate the mesh again, with the **same** NPOIN and NELEM but the mesh files are different in point numbering, and now the case runs with 3 cores as well and above that gives the same error.

Is this issue known? It may be a small fix in "CPhysicalGeometry::ComputeNSpan" to partition nSpan outflow correctly.

 - OS: CentOS 7
 - SU2 Version:  v7.0.4

economon:
I am not too familiar with that routine, but I do know that there are some checks related to periodic BCs inside. Are you using periodic BCs for your case? You might go through the routine to see that the counting of the spans is correct. Perhaps @salvovitale can comment if there's time.

Also, can you verify that the test cases in SU2/TestCases/turbomachinery/ are working for you on your system? Might check for additional options that appear in those examples.

auzbaig:
Yes, I am using periodic boundaries. However, the points are matched without any issue during initialization so I don't think the issue is there. Turbomachinery cases are working but the issue is with MPI where after a certain number of cores a case may encounter segmentation fault. I cannot predict beforehand at what number of cores it will fail, or if it will even fail or not.

stephansmit:
I had the same issue, its depended on the way the mesh is partitioned. There is an error in the calculation of NSpan (they should be the same for the inlet and outlet) when the boundary is on two different cores (points are counted twice).  I resolved it by making an extra entry in the config that specifies the number of spanwise sections.

auzbaig:
@stephansmit I think that should be able to resolve the issue temporarily. Can you let me know how you did it? Can you start a pull request for those changes to merge into [turbo_output_opt](https://github.com/su2code/SU2/tree/turbo_output_opt)?

auzbaig:
[...]

PR #1221, opened 2021-03-03
Fix calculation of number of spanwise sections

### LINKED ISSUE #1219: MPI fatal error when using more than 6 processors at Turbomachinery Preprocessing step
**### MPI partitioning issues for Turbomachinery for processor count more than 6 in mpi**

I have noticed an issue with running my turbomachinery model in SU2 version 7.04, 7.1 and version 7.1.1 during the Turbomachinery preprocessing step at the  "Initialize Turbo Vertex Structure" where it **incorrectly** calculates the number of spanwise section in zone 0 and then there is a MPI fatal error. **This happens only when I use more than 6 processors**. For less than or equal to 6 processors there is no error. I have this issue with 7.04, 7.1 and now 7.1.1. 

This issue has a previous bug report in the MPI partitioning issues for Turbomachinery runs #990

I am sure the config file is correct. I have my solutions from 6 processors, but I really need to scale up the processor count so that the restitution time is within a reasonable time.
Below is an example of where it crashes for 7 processors: 
### WITH 7 processors:
---------------------- Turbomachinery Preprocessing ---------------------

Initialize Turbo Vertex Structure.
**_Number of span-wise sections in Zone 0: 139. [This is incorrect number of span-wise sections]_**
Max number of span-wise sections among all zones: 139.
Initialize solver containers for average and performance quantities.
Compute inflow and outflow average geometric quantities.
[1:node0567] unexpected disconnect completion event from [0:node0564]
[3:node0566] unexpected disconnect completion event from [0:node0564]

### WITH 6 processors:
Initialize Turbo Vertex Structure.
**Number of span-wise sections in Zone 0: 137.  [This is the correct number of span-wise sections]**
Max number of span-wise sections among all zones: 137.
Initialize solver containers for average and performance quantities.
Compute inflow and outflow average geometric quantities.
Transfer average geometric quantities to zone 0.
Inlet area for Row 1: 30.7045 cm^2.
Oulet area for Row 1: 16.9716 cm^2.
Recomputed number of blades for Row 1: 36.
Initialize inflow and outflow average solution quantities.
Inlet flow angle for Row 1: -0.00422866°.
Outlet flow angle for Row 1: -28.9224°.
------------------------------ Begin Solver -----------------------------
.......

System settings:
**CLUSTER**
 - OS: CentOS release 6.5
 - C++ compiler and version: GCC/7.3.0-2.30
 - MPI implementation and version: compiler/intel/2018.0.1 intelmpi/2018.0.1/64 mkl/2018.0.1
 - SU2 Version: [v7.1 and 7.1.1]
Thanks
Ashvin

pcarruscag:
## Proposed Changes
Try to fix #1219 based on Stephan's insight in #990

## PR Checklist


TobiKattmann (approved):
Nice investigative work 🕵️ with the issues 👍

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -5108,53 +5102,56 @@ unsigned short iMarker, jMarker, iMarkerTP, iSpan, jSpan, kSpan = 0;
...
+              if (!nodes->GetDomain(iPoint)) continue;
+
+              PeriodicBoundary = config->GetMarker_All_PerBound(jMarker);
+              jVertex = nodes->GetVertex(iPoint, jMarker);
+
+              if ((jVertex != -1) && (PeriodicBoundary == (val_iZone + 1))){
+                coord = nodes->GetCoord(iPoint);
+                radius = sqrt(coord[0]*coord[0]+coord[1]*coord[1]);
+                switch (config->GetKind_TurboMachinery(val_iZone)){
+                case CENTRIFUGAL:
```
TobiKattmann, line 5128: `[...] as I will not debug that function any further because it has the world record for nested loops.`
@pcarruscag, 2021-03-04, Digital ink on computerscreen in #1219

TobiKattmann: Continue-man to the rescue 🦸

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -5192,46 +5178,15 @@ unsigned short iMarker, jMarker, iMarkerTP, iSpan, jSpan, kSpan = 0;
...
-      //
-      //  for (iSpan = 0; iSpan < nSpan; iSpan++){
-      //    if(rank == MASTER_NODE){
-      //      cout << setprecision(16)<<  iSpan +1 << " with a value of " <<valueSpan[iSpan]<< " at flag " << marker_flag <<endl;
-      //    }
-      //  }
+      /*--- Terrible stuff to do but so is this entire bloody function goodness me... ---*/
+      SpanWiseValue[marker_flag -1] = valueSpan;
 
+      sort(SpanWiseValue[marker_flag -1], SpanWiseValue[marker_flag -1]+nSpan);
```
TobiKattmann, line 5186: 👍

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: Info: The artist has yet to name his piece 🎨

pcarruscag: Proper thing to do would be to find the vertices this function wants to then not repeat all those conditions 3 times... But anyway, I scrolled down from this function and realized this was only a drop in the ocean.

Common/src/geometry/CPhysicalGeometry.cpp
pcarruscag: I should hope they fire anyone who implements sort "inline" especially the quadratic complexity variant.
I also hope the comment was correct because I did not try to understand if this "sort" had something different about it.

PR #1222, opened 2021-03-04
Removed CSolver::Convective_Residual

maxaehle:
## Proposed Changes
Remove `CSolver::Convective_Residual`, as it does nothing and no derived class overrides it. There is one call in `CIntegration.cpp`, which I would like to replace by an error message. 

Note that in the present version, usage of `numerics[CONV_TERM]` would not be thread-safe, it should be `numerics[CONV_TERM + omp_get_thread_num()*MAX_TERMS]` (see also #1214).

## PR Checklist

### DIFF
--- a/SU2_CFD/src/integration/CIntegration.cpp
+++ b/SU2_CFD/src/integration/CIntegration.cpp
@@ -58,9 +58,6 @@ void CIntegration::Space_Integration(CGeometry *geometry,
     case SPACE_UPWIND:
       solver_container[MainSolver]->Upwind_Residual(geometry, solver_container, numerics, config, iMesh);
       break;
-    case FINITE_ELEMENT:
-      SU2_MPI::Error("Computation of FEM convective flux in non-FEM integration attempted.", CURRENT_FUNCTION);
-      break;
   }
 
   /*--- Compute viscous residuals ---*/
--- a/SU2_CFD/src/integration/CIntegration.cpp
+++ b/SU2_CFD/src/integration/CIntegration.cpp
@@ -59,7 +59,7 @@ void CIntegration::Space_Integration(CGeometry *geometry,
       solver_container[MainSolver]->Upwind_Residual(geometry, solver_container, numerics, config, iMesh);
       break;
     case FINITE_ELEMENT:
-      solver_container[MainSolver]->Convective_Residual(geometry, solver_container, numerics[CONV_TERM], config, iMesh, iRKStep);
+      SU2_MPI::Error("Computation of FEM convective flux in non-FEM integration attempted.", CURRENT_FUNCTION);
       break;
   }
 
--- a/SU2_CFD/include/drivers/CDriver.hpp
+++ b/SU2_CFD/include/drivers/CDriver.hpp
@@ -745,7 +745,7 @@ class CFluidDriver : public CDriver {
 protected:
    unsigned long Max_Iter;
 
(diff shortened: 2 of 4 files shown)


SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -59,7 +59,7 @@ void CIntegration::Space_Integration(CGeometry *geometry,
       solver_container[MainSolver]->Upwind_Residual(geometry, solver_container, numerics, config, iMesh);
       break;
     case FINITE_ELEMENT:
-      solver_container[MainSolver]->Convective_Residual(geometry, solver_container, numerics[CONV_TERM], config, iMesh, iRKStep);
+      SU2_MPI::Error("Computation of FEM convective flux in non-FEM integration attempted.", CURRENT_FUNCTION);
       break;
```
pcarruscag, line 63: Thank you for going after dead code :+1: 
It should also be fairly safe to not cover the FINITE_ELEMENT case or just handle `default`, I doubt the execution will reach here if the wrong integration class is instantiated (multiple things would have to go wrong).
But looks good to me.

pcarruscag (approved):

SU2_CFD/src/integration/CIntegration.cpp
maxaehle: By the way, there is dead code even in `SU2_CFD.cpp`, the `else` branch is unreachable: If it were reached, then `turbo` and `harmonic_balance` would both be `false`. As `(multizone && !turbo)` has to be `false`, `multizone` must be `false` as well. But then `((!multizone && !harmonic_balance && !turbo) || (turbo && disc_adj))` is `true` ;)

maxaehle: I didn't touch this until now because I don't understand the principle behind these nested `if`-statement. Do you think the `else` branch in `SU2_CFD.cpp` should be removed?

pcarruscag: I'm not even going to try unfolding the logic, take it out or throw an error, if the tests pass it's fine by me.

pcarruscag: :spaghetti:

pcarruscag: We can even make the constructors of CFluidDriver protected since it is not meant for standalone use anymore (it exists as the base for the legacy drivers).

maxaehle: done in commit 7825b3b

maxaehle: > It should also be fairly safe to not cover the FINITE_ELEMENT case or just handle `default`, I doubt the execution will reach here if the wrong integration class is instantiated (multiple things would have to go wrong).

I agree, commit 42fab4f

TobiKattmann (approved):
Thanks @maxaehle for removing unused code 👍 Always really valuable to have those cleanups 🧹

maxaehle:
When the tests have run, I'll ignore this annoying CodeFactor test and merge, right?

TobiKattmann:
> 
> 
> When the tests have run, I'll ignore this annoying CodeFactor test and merge, right?

👍 Yep, I guess this `Complex code` without any further detail of codeFactor is a bug or sth

PR #1223, opened 2021-03-06
[WIP]: Flamelet progress variable (FPV) model

danielmayer:
## Proposed Changes
This PR implements a combustion model for laminar premixed flames in low Mach number flows. The implementation is based on a flamelet progress variable approach - i.e., the combustion chemistry is parameterized with scalar quantities for which scalar transport equations are solved. The scalars are then used to look up scalar source terms, transport -, and thermodynamic quantities from a lookup table. The lookup table is compiled from one-dimensional flame simulations using more expensive chemistry models - typically finite rate mechanisms.

In this implementation we introduce 2 additional transported scalars: the progress variable **c** and the total enthalpy **h**. All fluid properties like density, viscosity, temperature, diffusivity, reaction source terms, etc. are obtained from a lookup table that is made by the user with external 1D detailed chemistry codes like Cantera. Since temperature is now a dependent variable, we do not need to solve the regular temperature equation. However, We keep the SU2 temperature field and simply write our lookup temperature T(c,h) into this field. 

As a 'limiting case', a nonreacting passive scalar transport equation has also been implemented. All fluid properties for this passive scalar are constant, but this passive scalar can be easily extended to model nonreacting mixing flows of two fluids with different properties. This model has also been verified to work together with turbulence, although no turbulence chemistry interaction has been implemented for this WIP. 

## Related Work
This PR implements a generic scalar transport framework which can be aligned with other physics models that make use of transported scalars, e.g. turbulence models. It furthermore implements a generic lookup table framework that can be used as a basis for lookup table based physics models, such as real gas models.

## PR Checklist

## Notes
As indicated, this is work in progress. The code compiles and runs with a test case but is far from being clean. Side effects in non-flamelet simulations might occur.


bigfooted:
Great! I've added a project to github as well: 
https://github.com/su2code/SU2/projects/12
We can then keep track of what needs to be done.

pcarruscag:
See my comment in #1226 and let me know.
I would not like to see us going back to v6.2 style, it would be "going from horse to donkey" as the Portuguese say.
We can talk details tomorrow 4pm CET https://meet.jit.si/SU2_DevMeeting

SU2_CFD/src/solvers/CFlameletSolver.cpp
```diff
@@ -0,0 +1,890 @@
...
+                                          CSolver **solver_container,
+                                          CConfig *config,
+                                          unsigned short iMesh) { }
+
+unsigned long CFlameletSolver::SetPrimitive_Variables(CSolver **solver_container,
+                                                           CConfig *config,
+                                                           bool Output) {
+  
+  unsigned long iPoint, ErrorCounter = 0;
+  su2double Density, Temperature, lam_visc = 0.0, eddy_visc = 0.0, Cp;
```
pcarruscag, line 244: Continuing the conversation from yesterday.
@bigfooted @danielmayer so far as handling diffusivity for the different solvers, this looks like the solution.
Do it as part of the preprocessing, or in case it is constant it can be set in the constructor.

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
```diff
@@ -0,0 +1,583 @@
...
+void CPassiveScalarSolver::BC_Isothermal_Wall(CGeometry *geometry,
+                                              CSolver **solver_container,
+                                              CNumerics *conv_numerics,
+                                              CNumerics *visc_numerics,
+                                              CConfig *config,
+                                              unsigned short val_marker) {
+  
+  /*--- Convective fluxes across viscous walls are equal to zero. ---*/
+  
+}
```
pcarruscag, line 583: What is the reason to handle only inlet and outlet BC's in the passive and scalar solvers?

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,1102 @@
...
+
+  numerics->SetDiffusionCoeff(nodes->GetDiffusivity(iPoint),
+                              nodes->GetDiffusivity(jPoint));
+
+  /*--- Compute residual, and Jacobians ---*/
+
+  auto residual = numerics->ComputeResidual(config);
+
+  if (ReducerStrategy) {
+    EdgeFluxes.SubtractBlock(iEdge, residual);
```
pcarruscag, line 304: Cheers for bringing all this in line with the turbulence solver :+1:

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
bigfooted: For walls, we assume zero flux for the passive (nonreacting) scalars, so we don't have to do anything (I hope?). For combustion, we can impose a temperature, which is then converted to an enthalpy boundary condition for the enthalpy equation (one of the transported scalars is the equation for total enthalpy).
Other boundaries we haven't considered yet.

SU2_CFD/src/solvers/CFlameletSolver.cpp
bigfooted: yes, diffusivity is a bit messy now. For combustion, we get diffusivity from a lookup table (just like viscosity, temperature, etc, all read from a file).  For the passive scalar, I want to extend this to a mixing model.

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
pcarruscag: Then you are assuming all walls are "adiabatic" otherwise you have a direct contribution to the residual from a boundary flux, or one computed from the normal gradient in the isothermal (Dirichlet) case.

SU2_CFD/src/solvers/CFlameletSolver.cpp
pcarruscag: Ok each solver can be as messy as it wants in that primitive variables method.
But in the base solver we will avoid logic relating with the kind of solver (if passive if heat if turbulence).

Are you planning to do any more work on CScalarSolver? If not I will use what you have now to create the CFVMSolverBase and apply it to turbulence. I will then create a PR to develop, and after that is merged you can apply it here to the PassiveScalar and FlameletSolvers, @oleburghardt can then do it for the heat solver in #1226.
I think it is the best compromise to keep PR sizes manageable.
Does that sound like a plan?

Common/include/CConfig.hpp
```diff
@@ -6373,6 +6572,13 @@ class CConfig {
    */
   su2double GetInlet_Ttotal(string val_index) const;
 
+    /*!
+   * \brief Get the scalar values at an inlet boundary.
+   * \param[in] val_index - Index corresponding to the inlet boundary.
+   * \return The inlet scalar values.
+   */
+  su2double* GetInlet_ScalarVal(string val_index) const;
```
pcarruscag, line 6580: This should return const su2double* otherwise it is not really const.

Common/include/CConfig.hpp
```diff
@@ -7384,6 +7596,27 @@ class CConfig {
...
+   * \param[in] val_surface_no - Value of the NOx mass fraction.
+   */
+  //void SetSurface_Scalar(unsigned short val_imarker, su2double val_surface_scalar, unsigned short val_i_scalar){ Surface_Scalar[val_imarker][val_i_scalar] = val_surface_scalar; };
+
+  /*!
+   * \brief Set the NO mass fraction at the surface.
+   * \param[in] val_imarker - Index corresponding to the outlet boundary.
+   * \param[in] val_surface_no - Value of the NO mass fraction.
+   */
+  void SetSurface_NOx(unsigned short val_imarker, su2double val_surface_nox){ Surface_NOx[val_imarker] = val_surface_nox; };
```
pcarruscag, line 7618: Do you use this "Set" or was it implemented for completeness?
We should try to keep the config as read-only as possible, and we should especially avoid using it to store things that are solution variables (like we do with AoA in fixed CL mode, it is as bad as using global variables).

Common/include/option_structure.inl
```diff
@@ -1277,6 +1277,78 @@ public:
   }
 };
 
+class COptionInletScalar : public COptionBase {
+  string name; // identifier for the option
+  unsigned short & size;
+  string * & marker;
+  su2double ** & inletscalarval;
```
pcarruscag, line 1284: if this is a completely new type of option please use vector<vector< instead of ** and if the number of columns is constant use an su2activematrix

SU2_CFD/include/fluid/CFluidModel.hpp
```diff
@@ -119,10 +125,39 @@ class CFluidModel {
...
 
+  /*!
+   * \brief Get the source term of the transported scalar
+   */
+  virtual inline su2double GetSourceScalar(int){ return 37; }
+  
+  /*!
+  * \brief Get the number of transported scalars
+  */
+  virtual inline unsigned short GetNScalars() {return 73; }
```
pcarruscag, line 136: Magic numbers? inline before virtual (usually...)

SU2_CFD/include/fluid/CFluidFlamelet.hpp
```diff
@@ -0,0 +1,71 @@
...
+  unsigned long GetEnthFromTemp(su2double *enthalpy, su2double val_prog, su2double val_temp);
+
+  inline CLookUpTable* GetLookUpTable() {return look_up_table; }
+
+  inline su2double GetSourceEnergy() { return source_energy; }
+
+  inline su2double GetMassDiffusivity() { return mass_diffusivity; }
+
+  inline su2double GetThermalConductivity() { return Kt; }
+
```
pcarruscag, line 52: const the gets, use override or final as applicable

SU2_CFD/src/fluid/CFluidFlamelet.cpp
```diff
@@ -0,0 +1,195 @@
+
+#include "../include/fluid/CFluidFlamelet.hpp"
+#include "../include/numerics/CLookUpTable.hpp"
+
+CFluidFlamelet::CFluidFlamelet(CConfig *config, su2double value_pressure_operating) : CFluidModel() {
+
+#ifdef HAVE_MPI
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+#endif
```
pcarruscag, line 9: avoid the ifdefs, all methods of SU2_MPI will compile fine for serial compilation.

SU2_CFD/src/fluid/CFluidFlamelet.cpp
```diff
@@ -0,0 +1,195 @@
...
+    table_source_names.resize(n_table_sources);
+    table_source_names.at(I_SRC_TOT_PROG_VAR) = "ProdRateTot-PV";
+    table_source_names.at(I_SRC_TOT_CO)       = "ProdRateTot-CO";
+    table_source_names.at(I_SRC_TOT_NOX)      = "ProdRateTot-NOx";
+
+
+    config->SetTableSourceNames(table_source_names);
+
+    look_up_table = new CLookUpTable(config->GetFileNameLUT(),table_scalar_names.at(I_PROG_VAR),table_scalar_names.at(I_ENTHALPY));
+
```
pcarruscag, line 34: This does not need to be instantiated with "new" unless you are planning to have polymorphic tables?
You could just have it as member variable and have a "Initialize" method rather than just the constructor.

SU2_CFD/include/fluid/CFluidModel.hpp
```diff
@@ -119,10 +125,39 @@ class CFluidModel {
...
+  virtual inline su2double GetdSourcePVdPV() { return 0.0; }
+  virtual inline su2double GetdDensitydEnth() { return 0.0; }
+
   /*!
    * \brief Get fluid dynamic viscosity.
    */
-  su2double GetLaminarViscosity() {
+  virtual inline su2double GetLaminarViscosity() {
     LaminarViscosity->SetViscosity(Temperature, Density);
     Mu = LaminarViscosity->GetViscosity();
```
pcarruscag, line 162: I do not like this architecture because now it is double polymorphic.
i.e. the method is virtual therefore cannot be inlined, and then it makes calls to the models which are again polymorphic.

Is it possible to avoid this by creating ViscosityModels and so on that are backed by the lookup table?

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2134,6 +2171,58 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
...
+    
+    /*--- Definition of the source term integration scheme for each equation and mesh level ---*/
+    
+    for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
+      if (passive_scalar || progress_variable) {
+        numerics[iMGlevel][SCALAR_SOL][source_first_term]  = new CSourcePieceWise_transportedScalar_general(nDim, nVar_Scalar, config);
+   //if (config->GetAxisymmetric() == YES)
+      //    numerics[iMGlevel][SCALAR_SOL][SOURCE_SECOND_TERM] = new CSourceAxisymmetric_Scalar(nDim, nVar_Scalar, config);
+      //else  
+        numerics[iMGlevel][SCALAR_SOL][SOURCE_SECOND_TERM] = new CSourceNothing(nDim, nVar_Scalar, config);
```
pcarruscag, line 2213: Use the lower case variable `source_second_term` otherwise this will leak memory when using OpenMP and then segfault.
We instantiate one numerics per thread.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -297,6 +299,17 @@ void CIncEulerSolver::SetNondimensionalization(CConfig *config, unsigned short i
...
       break;
 
+    case FLAMELET_FLUID_MODEL:
+
+      config->SetGas_Constant(UNIVERSAL_GAS_CONSTANT/(config->GetMolecular_Weight()/1000.0));
+      Pressure_Thermodynamic = Density_FreeStream*Temperature_FreeStream*config->GetGas_Constant();
+      auxFluidModel = new CFluidFlamelet(config,Pressure_Thermodynamic);
+      n_scalars = auxFluidModel->GetNScalars();
+      dummy_scalar = new su2double[n_scalars]();
+      auxFluidModel->SetTDState_T(Temperature_FreeStream, dummy_scalar);
```
pcarruscag, line 309: If it is dummy, can it not be nullptr?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1081,6 +1101,8 @@ void CIncEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_cont
   const bool muscl      = (config->GetMUSCL_Flow() && (iMesh == MESH_0));
   const bool limiter    = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER);
   const bool van_albada = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE);
+  const bool energy     = config->GetEnergy_Equation();
+  const bool flame      = (config->GetKind_Scalar_Model() == PROGRESS_VARIABLE);
```
pcarruscag, line 1105: @danielmayer Please write a few words (in the PR) describing this interaction between the flamelet solver and the energy equation in incompressible flows.
Replaces? Overrides? Requires it?
If possible encapsulate the `energy && !flame` logic in GetEnergy_Equation, that would be the safest way to guarantee that new pull requests will not break, and will be compatible, with this feature.

SU2_PY/SU2/opt/project.py
```diff
@@ -160,10 +160,11 @@ def __init__( self, config, state=None ,
             # look for existing designs
             folders = glob.glob(self._design_folder)
             if len(folders)>0:
-                sys.stdout.write('Removing old designs in 10s.')
-                sys.stdout.flush()
-                if warn: time.sleep(10)
-                sys.stdout.write(' Done!\n\n')
+                for countdown in range(10, 0, -1):
```
pcarruscag, line 163: "your mission should you choose to accept it..." :)

Common/include/CConfig.hpp
bigfooted: yes, fixed it, will be in the next push

Common/include/CConfig.hpp
bigfooted: yes, in combustion we solve usually 4 transport equations: progress variable (PV), enthalpy,carbon monoxide CO and nitrogen oxides NOx. Our main objective is design optimization to minimize CO and NOx.

SU2_CFD/src/fluid/CFluidFlamelet.cpp
bigfooted: we still have to convert cfluidflamelet...

SU2_CFD/src/solvers/CIncEulerSolver.cpp
bigfooted: we now keep the temperature field but do not solve anything for temperature. We look up temperature and then store it in the temperature field.

SU2_CFD/src/solvers/CFlameletSolver.cpp
bigfooted: @pcarruscag Please go ahead and do what needs to be done. We will also merge cpassivescalarsolver.cpp and cflameletsolver.cpp as they end up being almost the same. Regarding diffusivity I think now that this should be treated in the same way as viscosity, so as part of the regular fluid model properties. After all, fluids also experience self-diffusion so even for a single component gas it makes sense to have diffusivity as a fluid property.

TobiKattmann:
Hm, good question why the builds are not running. It should not depend on the LGTM things or CodeFactor at all, they are independent.
My suggestion would be to quickly resolve the conflicts with origin/develop and push again. I could imagine that this hinders the regression tests to run. I briefly went over the conflicts and they seemed to easy to fix (on the first glance).

You also have a merge request into develop and didn't touch the regression.yml so I see no problems on that side. The Regression tests also worked in the past so lets hope the merge origin/develop helps 👍

SU2_PY/SU2/run/deform.py
```diff
@@ -85,7 +85,9 @@ def deform ( config, dv_new=None, dv_old=None ):
         return info
     
     # setup mesh name
-    suffix = 'deformed'
+    suffix = 'deform'
+    if konfig['ENABLE_REMESHING'] == 'YES':
+        suffix += 'ed'
```
TobiKattmann: @danielmayer please change that to your needs if this here doesn't fit, but the default suffix should not change for other regression cases

SU2_PY/SU2/io/state.py
```diff
@@ -258,11 +258,13 @@ def find_files(self,config):
         targetea_name = 'TargetEA.dat'
         targetcp_name = 'TargetCp.dat'
         targetheatflux_name = 'TargetHeatFlux.dat'
-        lookuptable_name = config.FILENAME_LUT
+
+        special_cases = get_specialCases(config)
+        if 'FLAMELET_FLUID_MODEL' in special_cases:
+            lookuptable_name = config.FILENAME_LUT
```
TobiKattmann: Same as 1 comment above with this change

SU2_PY/SU2/opt/project.py
TobiKattmann: "This simulation will self-destruct in ten seconds. Good luck"

TestCases/flamelet_regression.py
```diff
@@ -77,22 +77,6 @@ def main():
 
     pass_list = [ test.run_test() for test in test_list ]
 
-    ######################################
-    ### RUN SU2_SOL TESTS              ###
-    ######################################
-
-    # parallel STL output using
-    stl_writer_test                = TestCase('stl_writer_test')
```
TobiKattmann: Note to self to fix this.

TobiKattmann:
I have no clue by now why `fd_sp_pinArray_cht_2d_dp_hf` fails for surface_static_temperature only ... the discrepancy is in the realm of 1e-6 of relative error but it otherwise  works fine (and it passes for me locally). It happened often enough enough to be more than a floating point hiccup

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -516,11 +511,6 @@ void CFlowOutput::SetAnalyzeSurface(CSolver **solver, const CGeometry *geometry,
     SetHistoryOutputPerSurfaceValue("AVG_NOX", y_NOx, iMarker_Analyze);
     Tot_Surface_NOx += y_NOx;
     config->SetSurface_NOx(iMarker_Analyze, y_NOx);
-
-    su2double temp = Surface_TotalTemperature_Total[iMarker_Analyze];
-    SetHistoryOutputPerSurfaceValue("AVG_TEMP", temp, iMarker_Analyze);
-    Tot_Surface_Temp += temp;
-    config->SetSurface_Temperature(iMarker_Analyze, temp);
```
TobiKattmann: This commit 69b1e7c2aa046bca6db08c3a0222798e2d4bb170 fixed the last regression test with SURFACE_STATIC_TEMPERATURE. 

I dont see what the purpose here was because  the values from `Surface_TotalTemperature_Total` are taken which are already available as `SURFACE_TOTAL_TEMPERATURE`. The multiplication with reference Temp is missing and `config->SetSurface_Temperature(iMarker_Analyze, temp);` is used for `SURFACE_STATIC_TEMPERATURE`. This last aspect is also where the error comes in I guess. 

I removed this code which added AVG_TEMP now.

TobiKattmann:
OK the failing regression test were due to:
1. One simple wrong function call in Csolver
2. Creating new output in existing groups (AERO_COEFF, FLOW_COEFF) which invalidates some regression tests
3. pyhton code which does not test if some config variable is present or altering the python code such that it invalidates other regression tests
4. and finally the AVG_TEMP thing above

Now all these alone were somewhat minor things.

Feature_flamelet  is now up-to-date with develop with all regression test working

TobiKattmann:
Good news,
@bigfooted @danielmayer @lkusch @EvertBunschoten : Commit ecb1154 fixed the restarts for feature_multicomp which is based on feature_flamelet and from a first thought it might work for table lookup as well. Please let me know if it does so. Otherwise The interaction of the table lookup with the Turb and Flow requires some additional calls to Pre/Post routines.

@mheimgartner as discussed please merge feature_flamelet into multicomp, such that you have that fix as well.

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -904,8 +904,13 @@ void CScalarSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig
   solver[MESH_0][SCALAR_SOL]->InitiateComms(geometry[MESH_0], config, SOLUTION);
   solver[MESH_0][SCALAR_SOL]->CompleteComms(geometry[MESH_0], config, SOLUTION);
 
+  // Flow-Pre computes/sets mixture properties
   solver[MESH_0][FLOW_SOL]->Preprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
-  solver[MESH_0][SCALAR_SOL]->Postprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0);
```
TobiKattmann: Scalar_sol->post is empty, the call to preprocessing would be correct

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -126,6 +126,15 @@ void CFluidIteration::Iterate(COutput* output, CIntegration**** integration, CGe
...
                                                                        RUNTIME_SCALAR_SYS, val_iZone, val_iInst);
+
+    // In case of turbulence, the Turb-Post computes the correct eddy viscosity based on mixture-density and
+    // mixture lam-visc. In order to get the correct mixture properties, based on the just updated mass-fractions, the
+    // Flow-Pre has to be called upfront. The updated eddy-visc are copied into the flow-solver Primitive in another 
+    // Flow-Pre call which is done at the start of the next iteration.
+    if(config[val_iZone]->GetKind_Turb_Model()) {
+      solver[val_iZone][val_iInst][MESH_0][FLOW_SOL]->Preprocessing(geometry[val_iZone][val_iInst][MESH_0], solver[val_iZone][val_iInst][MESH_0], config[val_iZone], MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
+      solver[val_iZone][val_iInst][MESH_0][TURB_SOL]->Postprocessing(geometry[val_iZone][val_iInst][MESH_0], solver[val_iZone][val_iInst][MESH_0], config[val_iZone], MESH_0);
+    }
```
TobiKattmann, line 137: In the end the restart was fine but the main-iteration was missing this update of eddy viscosity to be correct (or consistent with the restart).

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -904,8 +904,13 @@ void CScalarSolver::LoadRestart(CGeometry **geometry, CSolver ***solver, CConfig
...
+  // Update eddy-visc which needs correct mixture density and mixture lam-visc. Note that after this, another Flow-Pre
+  // at the start of the Iteration sets the updated eddy-visc into the Flow-Solvers Primitives.
+  if(config->GetKind_Turb_Model()) solver[MESH_0][TURB_SOL]->Postprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0);
+  // For feature_multicomp this Scalar-Pre only computes the laminar contribution to mass diffusivity
+  solver[MESH_0][SCALAR_SOL]->Preprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
```
TobiKattmann, line 913: Note that Turb-Post (computes `eddy-visc(rho_mix, mu_lam_mix)`) and Scalar-Pre (computes `lam_Diffusivity(rho_mix, cp_mix, therm_cond_mix)`) can be switched.

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -645,12 +655,8 @@ void CScalarSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver_c
...
       Density_n = 1.0;
 
-      //if (sst_model) {
-        if (incompressible)
-          Density_n = flowNodes->GetDensity(iPoint); // Temporary fix
-      //  else
-      //    Density_n = flowNodes->GetSolution_time_n(iPoint,0);
-      //}
+      if (incompressible)
+        Density_n = flowNodes->GetDensity(iPoint);
```
TobiKattmann: This is no tvalid for the SA model, the if sst needs to stay. ... but will be replaced in #1330

SU2_CFD/src/solvers/CFlameletSolver.cpp
```diff
@@ -535,31 +535,34 @@ void CFlameletSolver::Source_Residual(CGeometry *geometry,
   CFluidModel *fluid_model_local = solver_container[FLOW_SOL]->GetFluidModel();
   
   SU2_OMP_FOR_DYN(omp_chunk_size)
-  for (auto i_point = 0u; i_point < nPointDomain; i_point++) {
+  for (auto iPoint = 0u; iPoint < nPointDomain; iPoint++) {
```
TobiKattmann, line 538: please separate cleanup commits like these parts from actual code changes. And separate them logically as well to some degree, like the strongBC thing would be its own commit. This make looking through the commits much easier than this big chunk. Doesn't need to be perfect but some care makes the overview much easier as there is some people's work depending on feature_flamelet.

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -124,7 +124,7 @@ void CFluidIteration::Iterate(COutput* output, CIntegration**** integration, CGe
 
   if (config[val_iZone]->GetKind_Scalar_Model() != NO_SCALAR_MODEL){
     config[val_iZone]->SetGlobalParam(RANS, RUNTIME_SCALAR_SYS);
-    integration[val_iZone][val_iInst][SCALAR_SOL]->MultiGrid_Iteration(geometry, solver, numerics, config, 
+    integration[val_iZone][val_iInst][SCALAR_SOL]->SingleGrid_Iteration(geometry, solver, numerics, config,
```
TobiKattmann, line 127: I already wondered why MultigridIteration for Scalar. This does make a difference, e.g. Preprocessing is also called after the iteration for Multigrid! I think in this case  the restarts (for Marks work) should be fine but I am not entirely sure for the flamelet cases. If there are potential Testcases with table lookup I highly recommend adding those here. I already put the structure in place. Now that the primal restart is working I would go for a restarted case!

bigfooted: I feel that when we have confidence in multigrid, we should activate it for the additional solvers like turbulence and scalars. But at this moment I do not see a large benefit in using multigrid because it is not robust.

SU2_CFD/src/numerics/flamelet/scalar_convection.cpp
```diff
@@ -122,19 +122,15 @@ void CUpwSca_transportedScalar_general::FinishResidualCalc(const CConfig* config
     if (implicit) {
       for (auto jVar = 0u; jVar < nVar; jVar++) {
         if (iVar == jVar) {
-          // note that for transported scalar we multiply by density
-          Jacobian_i[iVar][jVar] = a0*Density_i;
-          Jacobian_j[iVar][jVar] = a1*Density_j;
+          Jacobian_i[iVar][jVar] = a0;
+          Jacobian_j[iVar][jVar] = a1;
```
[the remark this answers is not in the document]
bigfooted: Yes you are right, now there is a bunch of unrelated features in this commit

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -329,6 +329,7 @@ CNumerics::ResidualType<> CSourceIncAxisymmetric_Flow::ComputeResidual(const CCo
                              TWO3*AuxVar_Grad_i[0][1]);
       residual[3] -= Volume*yinv*Thermal_Conductivity_i*PrimVar_Grad_i[nDim+1][1];
 
+      if (implicit) jacobian[2][2] = Volume*yinv*2.0*total_viscosity*yinv;
```
bigfooted, line 332: Note that adding the jacobian of the energy equation in the incompressible flow will affect (residuals of) all incompressible testcases that have energy equation activated!

SU2_CFD/src/solvers/CScalarSolver.cpp
bigfooted: But this piece of code is used in the transported scalar solver , not in the turbulence solver (it was copied from the turbulence solver). We should only care here about if the transported scalar is "implemented as" the SST model or the SA model. And it is implemented as the SST model because we have rho*Y just as we have rho*k and rho*omega.

TobiKattmann: right... I am getting mixed-up with CScalarSolver over at #1330 because there it is general, my bad

SU2_CFD/src/numerics/flow/flow_sources.cpp
TobiKattmann: Hmm, i am not sure that this is the root of that. This is in `CSourceIncAxisymmetric_Flow::ComputeResidual` which should only affect cases with axisymmetric flow. Also, the  Jacobian [2][2] is `d residual[2] / d U[2]` where U[2] is y-velocity, as far as I understand it. So the energy equation should not play a role here

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -122,6 +122,21 @@ void CFluidIteration::Iterate(COutput* output, CIntegration**** integration, CGe
...
+    config[val_iZone]->SetGlobalParam(RANS, RUNTIME_SCALAR_SYS);
+    integration[val_iZone][val_iInst][SCALAR_SOL]->SingleGrid_Iteration(geometry, solver, numerics, config, 
+                                                                       RUNTIME_SCALAR_SYS, val_iZone, val_iInst);
+
+    // In case of turbulence, the Turb-Post computes the correct eddy viscosity based on mixture-density and
+    // mixture lam-visc. In order to get the correct mixture properties, based on the just updated mass-fractions, the
+    // Flow-Pre has to be called upfront. The updated eddy-visc are copied into the flow-solver Primitive in another 
+    // Flow-Pre call which is done at the start of the next iteration.
+    if(config[val_iZone]->GetKind_Turb_Model()) {
+      solver[val_iZone][val_iInst][MESH_0][FLOW_SOL]->Preprocessing(geometry[val_iZone][val_iInst][MESH_0], solver[val_iZone][val_iInst][MESH_0], config[val_iZone], MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
```
TobiKattmann, line 135:
```suggestion
      solver[val_iZone][val_iInst][MESH_0][FLOW_SOL]->Preprocessing(geometry[val_iZone][val_iInst][MESH_0], solver[val_iZone][val_iInst][MESH_0], config[val_iZone], MESH_0, NO_RK_ITER, RUNTIME_FLOW_SYS, true);
```

SU2_CFD/src/numerics/flamelet/scalar_diffusion.cpp
```diff
@@ -69,7 +69,7 @@ CAvgGrad_transportedScalar::~CAvgGrad_transportedScalar(void) {
 
 CNumerics::ResidualType<> CAvgGrad_transportedScalar::ComputeResidual(const CConfig* config) {
 
- const bool inc_rans = (config->GetKind_Solver() == INC_RANS);
+ const bool inc_rans = (config->GetKind_Solver() == INC_RANS) || (config->GetKind_Solver() == DISC_ADJ_INC_RANS);
```
TobiKattmann, line 72: Eventually we might want to find a better way to do this but for now there might be other instances where this DISC_ADJ_INC_RANS has to be added. Maybe there is some stuff for in the table lookup that uses this construct

TobiKattmann:
@danielmayer @mheimgartner @bigfooted @EvertBunschoten  please note that `CScalarSolver`, `CScalarVariable` and `scalar_sources/convection/diffusion` were renamed to (basically just extended by *Legacy) `CScalarLegacySolver`, `CScalarLegacyVariable` and `scalarLegacy_sources/convection/diffusion`. So all files that you are used to do now have the `Legacy` addition in their names as I introduced the same/similar files in #1330. The 'new' ones are really new in the sense that they form the general base for scalar transport and I will create a `CSpecies*` family of files to incorporate the species transport.

So right now both concepts are present in the same branch but the Legacy will be subsequently removed/replaced by the new stuff. 

In any case: merge with the upstream feature_flamelet on github regularly. Add Testcases so that these subsequent merges do not introduce any problems I am unable to see otherwise. And if you want to push a commit here please get in touch shortly with me first.

bigfooted:
let's keep this PR for now, while we are creating a new PR, chopping this one up into smaller chunks.

PR #1226, opened 2021-03-09
Heat Transfer boundary condition

oleburghardt:
## Proposed Changes
As an alternative to `MARKER_HEATFLUX = (heat_flux_marker, q,... )`, this PR adds `MARKER_HEATTRANSFER = (heat_transfer_marker, h, Twall)` to the list of wall boundary conditions where the wall heat flux is computed by q = h*(Twall-Tref). h is the heat transfer coefficient, Tref is to be set by default or by a further option.
 
@bigfooted Is this what you've been thinking of? If so, I'd go ahead adding this to the compressible and the heat solver as well, if not, we should get this right for the incompressible solver first :-)

## PR Checklist


oleburghardt:
> If you are planning to touch the heat solver, might I suggest you, @bigfooted, @danielmayer, and me get together like the power rangers and do the scalar solver thing properwise without grafting code from 10 versions ago into develop?
> 
> The ideia would be to create a CFVMSolverBase (in the style of the one for flow solvers) to cover:
> 
>     * Turbulence solvers;
> 
>     * Heat solver;
> 
>     * Radiation solver;
> 
>     * General passive/active scalar solvers;
> 
>     * Eventual transition solvers;
> 
> 
> The starting point would be the current base turbulence solver.

I'm in. The heat solver should be easy to integrate into a generic framework if it provides a Laplacian/diffusion coefficient (which will then turn into thermal diffusivity). Apart from that, there's nothing very heat specific in the heat solver :)

Common/include/option_structure.hpp
```diff
@@ -1130,7 +1130,8 @@ enum BC_TYPE {
   ENGINE_EXHAUST = 22,        /*!< \brief Boundary nacelle exhaust. */
   RIEMANN_BOUNDARY= 24,       /*!< \brief Riemann Boundary definition. */
   ISOTHERMAL = 25,            /*!< \brief No slip isothermal wall boundary condition. */
-  HEAT_FLUX  = 26,            /*!< \brief No slip constant heat flux wall boundary condition. */
+  HEAT_FLUX = 26,             /*!< \brief No slip constant heat flux wall boundary condition. */
+  HEAT_TRANSFER = 27,         /*!< \brief No slip heat transfer boundary condition. */
```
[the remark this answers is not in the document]
oleburghardt: Right. Just to be sure, you are talking about the `GetMarker_All_KindBC(iMarker) == || ...`-stuff? Until here, `HEAT_FLUX` is just used in `Space_Integration` and `BC_Wall_Generic` and I don't know whether there's a way to circumvent those two.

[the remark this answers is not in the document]
pcarruscag: Those are fine, there are still many places in CGeometry and CVolumetricMovement, and some in CFlowOutput.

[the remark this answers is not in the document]
pcarruscag: Fixed them in #1225, you should be ok just by updating CConfig::GetViscous_Wall

bigfooted:
> > If you are planning to touch the heat solver, might I suggest you, @bigfooted, @danielmayer, and me get together like the power rangers and do the scalar solver thing properwise without grafting code from 10 versions ago into develop?
> > The ideia would be to create a CFVMSolverBase (in the style of the one for flow solvers) to cover:
> > ```
> > * Turbulence solvers;
> > 
> > * Heat solver;
> > 
> > * Radiation solver;
> > 
> > * General passive/active scalar solvers;
> > 
> > * Eventual transition solvers;
> > ```
> > 
> > 
> > The starting point would be the current base turbulence solver.
> 
> I'm in. The heat solver should be easy to integrate into a generic framework if the framework provides a Laplacian/diffusion coefficient (which will then turn into thermal diffusivity). Apart from that, there's nothing very heat specific in the heat solver :)

Me too. I'm already cleaning up our code a bit so it's easier to see the differences and similarities.

danielmayer:
> > > If you are planning to touch the heat solver, might I suggest you, @bigfooted, @danielmayer, and me get together like the power rangers and do the scalar solver thing properwise without grafting code from 10 versions ago into develop?
> > > The ideia would be to create a CFVMSolverBase (in the style of the one for flow solvers) to cover:
> > > ```
> > > * Turbulence solvers;
> > > 
> > > * Heat solver;
> > > 
> > > * Radiation solver;
> > > 
> > > * General passive/active scalar solvers;
> > > 
> > > * Eventual transition solvers;
> > > ```
> > > 
> > > 
> > > The starting point would be the current base turbulence solver.
> > 
> > 
> > I'm in. The heat solver should be easy to integrate into a generic framework if the framework provides a Laplacian/diffusion coefficient (which will then turn into thermal diffusivity). Apart from that, there's nothing very heat specific in the heat solver :)
> 
> Me too. I'm already cleaning up our code a bit so it's easier to see the differences and similarities.

Sure! Let's do this.

Common/include/option_structure.hpp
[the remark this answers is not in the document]
oleburghardt: Thanks, you've been quicker, pulling the current develop into this resolved all the remaining places that I saw.

Common/src/CConfig.cpp
```diff
@@ -2493,12 +2490,14 @@ void CConfig::SetConfig_Options() {
...
-
-  /* DESCRIPTION: Thermal diffusivity constant */
-  addDoubleOption("THERMAL_DIFFUSIVITY", Thermal_Diffusivity, 1.172E-5);
-
   /* DESCRIPTION: Thermal diffusivity constant */
-  addDoubleOption("THERMAL_DIFFUSIVITY_SOLID", Thermal_Diffusivity_Solid, 1.172E-5);
+  addDoubleOption("SOLID_THERMAL_DIFFUSIVITY", Thermal_Diffusivity_Solid, 1.172E-5);
+  /* DESCRIPTION: Thermal conductivity used for heat equation */
+  addDoubleOption("SOLID_THERMAL_CONDUCTIVITY", Thermal_Conductivity_Solid, 0.0);
+  /* DESCRIPTION: Solids temperature at freestream conditions */
```
pcarruscag, line 2497: Add the deprecated option to CConfig::SetConfig_Parsing error handling part.

Is it weird to have both diffusivity and conductivity?

oleburghardt: Ok!

It is weird to have both. For the heat solver, the only parameter that goes into computing the heat fluxes is thermal diffusivity, though one is used to specifiy the (in this case) redundant triple conductivity/density/capacity. I'm very open to remove them and leave a hint in the config template how to calculate diffusivity from them...

pcarruscag: Hmmm according to line 169 of the HeatSolver conductivity is used in some cases.
And used to set the diffusivity in the config :confused: 
I would go for a non-ambiguous way of specifying this. Hopefully consistent with the scalar solvers but also semi-backwards compatible (i.e. different option name but same quantity that needs to be specified by users).

oleburghardt: Yes but just to translate it to diffusivity. I'll go for `THERMAL_DIFFUSIVITY`, this way we can remove three ugly `SOLID_*` options :)

oleburghardt: Yes that's what I was saying. For the heat solver, one can either specify thermal conductivity alongside density and specific heat, or one goes for setting the diffusivity directly.
However when prescribing heat fluxes (as BC or in CHT), they have to be non-dimensionalized by rho*c_p, so it's maybe a better option to remove `SOLID_THERMAL_DIFFUSIVITY`as a config file option.

oleburghardt: Now that there's a energy equation in the incompressible solver, I could also remove the ungly `SOLID_*` options as the heat equation solver is active for pure solids only; using the standard `Kt_Constant` and so on.

pcarruscag: That (boundary fluxes) is a good point to raise in #1223, the wall boundaries are not implemented in the scalar solvers, only inlets and outlets (I think...).

TobiKattmann:
Reopening after chatting with @oleburghardt . I merged develop (which I hopefully made no mistakes in as there were quite some conflicts).
Left todo is imo (beyond possible further comments, possible failing tests etc): 
- adding a testcase that at best serves as a verification/sanity check (the expected results of course have to be documented for a possible future redo). I'll see what i'll find
- take a second look at the cfg option consolidation for good measure

I would prefer if this PR would not tackle the CFVMScalarSolver thing, but a first WIP-PR into that direction will come to a github repository near you soon ;)

TestCases/tutorials.py
```diff
@@ -56,6 +57,18 @@ def main():
     cht_incompressible_unsteady.tol       = 0.00001
     test_list.append(cht_incompressible_unsteady)
 
+    # CHT incompressible, 2D, 3 pins in crossflow
+    cht_incompressible           = TestCase('cht_incompressible')
+    cht_incompressible.cfg_dir   = "../Tutorials/multiphysics/steady_cht"
+    cht_incompressible.cfg_file  = "cht_2d_3cylinders.cfg"
+    cht_incompressible.test_iter = 10
```
TobiKattmann, line 64: The steady_cht case that is in the tutorials is equal to the coupled_cht/incomp_2d. steady_cht was not in the regression tests which I think is desirable for all tutorials.

My proposal for this would be:
- Make sure the same reg-case works in tutorials and Testcases
- change some cfg-values (e.g. density) of the tutorial case back to its original to be the same as in the written tutorial. Make sure the results are reasonable. (actually the results may have been created with the values from the Testcases as they were to be changed [here in the related tutorial PR](https://github.com/su2code/su2code.github.io/pull/26))
- Remove the Testcases reg-test as we dont need it twice imo

It might also makes sense to move the disc_adj_incomp_2d to tutorials as that is also done in the [tutorial](https://su2code.github.io/tutorials/Inc_Heated_Cylinders/). This will/would be done in the related [Tutorials PR](https://github.com/su2code/Tutorials/pull/16)

TobiKattmann: In case someone knows how to make the CI-pipeline clone a specific branch of su2code/tutorials ... that would be appreciated.

Common/src/CConfig.cpp
```diff
@@ -8864,12 +8893,34 @@ su2double CConfig::GetWall_HeatFlux(string val_marker) const {
...
+    if (Marker_HeatTransfer[iMarker_HeatTransfer] == val_marker) break;
+  }
+
+  return HeatTransfer_Coeff[iMarker_HeatTransfer];
+}
+
+su2double CConfig::GetWall_HeatTransfer_Temperature(string val_marker) const {
+  unsigned short iMarker_HeatTransfer = 0;
+
+  if (nMarker_HeatTransfer > 0) {
```
pcarruscag, line 8910: The "ifs" are redundant since 0 is not smaller than 0.

Side note:
We have this search pattern in every method, but technically it results in out-of-bounds access if the string is not found.

SU2_CFD/include/solvers/CIncNSSolver.hpp
```diff
@@ -157,6 +157,22 @@ class CIncNSSolver final : public CIncEulerSolver {
...
+   * \param[in] conv_numerics - Description of the numerical method.
+   * \param[in] visc_numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
+  void BC_HeatTransfer_Wall(CGeometry *geometry,
+                            CSolver **solver_container,
+                            CNumerics *conv_numerics,
+                            CNumerics *visc_numerics,
+                            CConfig *config,
```
pcarruscag, line 173: Can we constify some of these? or even better remove what we might (I did not check) not need?

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -452,7 +452,9 @@ vector<string> CDriver::GetAllCHTMarkersTag() const {
 
   //The CHT markers can be identified as the markers that are customizable with a BC type HEAT_FLUX or ISOTHERMAL.
   for(iMarker=0; iMarker<nBoundariesMarker; iMarker++){
-    if((config_container[ZONE_0]->GetMarker_All_KindBC(iMarker) == HEAT_FLUX || config_container[ZONE_0]->GetMarker_All_KindBC(iMarker) == ISOTHERMAL) && config_container[ZONE_0]->GetMarker_All_PyCustom(iMarker)){
+    if((config_container[ZONE_0]->GetMarker_All_KindBC(iMarker) == HEAT_FLUX ||
+        config_container[ZONE_0]->GetMarker_All_KindBC(iMarker) == ISOTHERMAL ||
+        config_container[ZONE_0]->GetMarker_All_KindBC(iMarker) == HEAT_TRANSFER) && config_container[ZONE_0]->GetMarker_All_PyCustom(iMarker)){
```
pcarruscag, line 457: Time to cut-paste :)

TestCases/tutorials.py
pcarruscag: In regression.yml change what is in from of "-t":
```
with:
```
Based on this: https://github.com/su2code/Docker-Builds/blob/master/test/runTests.sh

Common/src/CConfig.cpp
```diff
@@ -5072,12 +5083,12 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
   SU2_MPI::Comm_size(SU2_MPI::GetComm(), &size);
 
   /*--- Compute the total number of markers in the config file ---*/
-
+  cout << "                 nMaker heat flux (2): " << nMarker_HeatFlux << endl;
```
pcarruscag, line 5086: Debug lines?

Common/src/CConfig.cpp
```diff
@@ -2911,6 +2902,16 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
...
+          if (!option_name.compare("THERMAL_DIFFUSIVITY"))
+            newString.append("THERMAL_DIFFUSIVITY is deprecated. See the INC_ENERGY_EQUATION options instead.\n\n");
+          if (!option_name.compare("THERMAL_DIFFUSIVITY_SOLID"))
+            newString.append("THERMAL_DIFFUSIVITY_SOLID is deprecated. Set KT_CONSTANT, MATERIAL_DENSITY and SPECIFIC_HEAT_CP instead.\n\n");
+          if (!option_name.compare("SOLID_THERMAL_CONDUCTIVITY"))
+            newString.append("SOLID_THERMAL_CONDUCTIVITY is deprecated. Use KT_CONSTANT instead.\n\n");
+          if (!option_name.compare("SOLID_DENSITY"))
+            newString.append("SOLID_DENSITY is deprecated. Use MATERIAL_DENSITY instead.\n\n");
+          if (!option_name.compare("SOLID_TEMPERATURE_INIT"))
+            newString.append("SOLID_TEMPERATURE_INIT is deprecated. Use FREESTREAM_TEMPERATURE instead.\n\n");
```
pcarruscag, line 2914: We will probably call the next version 7.2.0, and so it is fine by me to make changes to options.
However, let's give this some thought based on what would make sense for the scalar solver / flamelets and so on.
We don't need to get the code perfect but "interface" stability is important.

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -459,6 +464,13 @@ void CIncNSSolver::BC_Wall_Generic(const CGeometry *geometry, const CConfig *con
         LinSysRes(iPoint, nDim+1) += scalar_factor*dot_product;
       } // if streamwise_periodic
     }
+    else if (kind_boundary == HEAT_TRANSFER) {
+
+      /*--- Apply a weak boundary condition for the energy equation.
+      Compute the residual due to the prescribed temperature and transfer coefficient. ---*/
+      LinSysRes(iPoint, nDim+1) -= Transfer_Coefficient*Tdelta*Area;
+
+    }
```
pcarruscag, line 473: If this would be valid for the compressible solver let's make it available there too please.

TestCases/tutorials.py
TobiKattmann: Yep, I am stupid. I looked at this very line and thought "this is probably affects both: Testcases and Tutorials. But I cant change one without the other" ... that why there are two params 🤦
thanks

.github/workflows/regression.yml
```diff
@@ -87,7 +87,8 @@ jobs:
       - name: Run Tests in Container
         uses: docker://su2code/test-su2:20200303
         with:
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          # -t <Tutorials-branch> -c <Testcases-branch>
+          args: -b ${{github.ref}} -t feature_heatcoeff_BC -c develop -s ${{matrix.testscript}}
```
TobiKattmann, line 91: a little service to future me (and anyone who wonders) ;)

pcarruscag: :+1: Kudos on resisting the "stale bot uprising" btw.

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: I introduced a `BC_HeatFlux_Generic` layer just as `BC_Wall_Generic` in the inc case. I did not test the results yet though, but I am somewhat confident that it will do. I'll post sth below when I have some results

SU2_CFD/include/solvers/CIncNSSolver.hpp
TobiKattmann: 👍 solver_container and *_numerics are gone now and the remaining inputs are const

Common/src/CConfig.cpp
TobiKattmann: > The "ifs" are redundant since 0 is not smaller than 0.

true... someone wanted to make really sure that the loop-body is not executed 🛑  ... 7years ago (for e.g. heatflux).

I'll remove that for hf, isothermal and ht

SU2_CFD/src/python_wrapper_structure.cpp
TobiKattmann: Hm, I am unsure whether `HEAT_TRANSFER` is meant to be in that list or if that is unnecessary. The comment above says
```c++
//The CHT markers can be identified as the markers that are customizable with a BC type HEAT_FLUX or ISOTHERMAL.
```
and in CDriver it is
```c++
/*!
   * \brief Get all the heat transfer boundary markers tags.
   * \return List of heat transfer boundary markers tags.
   */
  vector<string> GetAllCHTMarkersTag() const;
```
Personally I would maybe revert the change here alltogether as I am not sure whether sth is specifically coded to workwith/overwrite some HEAT_FLUX or ISOTHERMAL walls. A clarifying insight here would be highly appreciated 💭

Common/src/CConfig.cpp
TobiKattmann: Unfortunately I don't have the clear overview (yet, fingers crossed) and what can/needs to be generalized. 

The only thing I would like to change rn would be `KT_CONSTANT` -> `THERMAL_CONDUCTIVITY_CONSTANT` as I feel that is much more expressive.

TobiKattmann:
In order to plausibilize the results in a first step I took the fluid-domain of the streamwise periodic 2D-cht testcase .
1.  h=0 <=> adiabatic wall, i.e. heatflux=0
2. h>>0 <=~> isothermal wall with Twall = Tinfinity

which the results recover (I also checked that nondimensionalization works correct) :+1: See the archive below for mesh and cfg files. See also below for some images

Note though that the definition of the heat-transfer/convection bc above in the PR-header is not correct. the heatflux computes to:

`q=h (T_infinity - T_wall)` with

`h` the the heat transfer coefficient (to be set in config by the user)

`T_infinity` the  reference temperature at infinity (to be set in config by the user)

`T_wall` the the current local temperature of the specific FV-cell (of course done by the code)



[htc-cases.tar.gz](https://github.com/su2code/SU2/files/6893231/htc-cases.tar.gz)

The first two are for the isothermal case 2. (nondimensionalized). There is still a ~2.5K gap but I cannot go higher with h then 1e5 as I got convergence then.
image
image

For heatflux=0 and h=0 the results are (as expected) completely identical, the diff between the screen outputs is only that one uses a different bc in that `config file boundary information` tab
image

.github/workflows/regression.yml
pcarruscag: Reminder to revert

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1099,7 +1099,8 @@ void COutput::SetScreen_Output(const CConfig *config) {
           PrintingToolbox::PrintScreenFixed(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
           break;
         case ScreenOutputFormat::SCIENTIFIC:
-          PrintingToolbox::PrintScreenScientific(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
+          // SURFACE_STATIC_TEMPERATURE e.g. is written to screen twice
+          if(false)PrintingToolbox::PrintScreenScientific(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
```
pcarruscag, line 1103: Revert?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -376,24 +376,28 @@ void CIncNSSolver::BC_Wall_Generic(const CGeometry *geometry, const CConfig *con
...
+  su2double Wall_HeatFlux = 0.0, Twall = 0.0, Tinfinity = 0.0, Transfer_Coefficient = 0.0;
 
   if (kind_boundary == HEAT_FLUX)
     Wall_HeatFlux = config->GetWall_HeatFlux(Marker_Tag)/config->GetHeat_Flux_Ref();
   else if (kind_boundary == ISOTHERMAL)
     Twall = config->GetIsothermal_Temperature(Marker_Tag)/config->GetTemperature_Ref();
+  else if (kind_boundary == HEAT_TRANSFER) {
+    Transfer_Coefficient = config->GetWall_HeatTransfer_Coefficient(Marker_Tag) * config->GetTemperature_Ref()/config->GetHeat_Flux_Ref();
+    Tinfinity = config->GetWall_HeatTransfer_Temperature(Marker_Tag)/config->GetTemperature_Ref();
+  }
```
pcarruscag, line 390: switch

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -406,19 +406,42 @@ void CNSSolver::AddDynamicGridResidualContribution(unsigned long iPoint, unsigne
...
+
+void CNSSolver::BC_HeatTransfer_Wall(const CGeometry *geometry, const CConfig *config, const unsigned short val_marker) {
+
+  BC_HeatFlux_Wall_Generic(geometry, config, val_marker, HEAT_TRANSFER);
+}
+
+void CNSSolver::BC_HeatFlux_Wall_Generic(const CGeometry *geometry, const CConfig *config,
+                                         unsigned short val_marker, unsigned short kind_boundary) {
 
   /*--- Identify the boundary by string name and get the specified wall
```
pcarruscag, line 423: :+1:

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -459,6 +463,16 @@ void CIncNSSolver::BC_Wall_Generic(const CGeometry *geometry, const CConfig *con
...
       } // if streamwise_periodic
     }
+    else if (kind_boundary == HEAT_TRANSFER) {
+      Twall = nodes->GetTemperature(iPoint);
+      Wall_HeatFlux = Transfer_Coefficient * (Tinfinity - Twall);
+
+      /*--- Apply a weak boundary condition for the energy equation.
+      Compute the residual due to the prescribed temperature and transfer coefficient.
+      Note that for the Heat_Transfer wall, basically a heatflux wall that depends on the local Temperature is applied. ---*/
+      LinSysRes(iPoint, nDim+1) -= Wall_HeatFlux*Area;
```
pcarruscag, line 473: You may want to update the Jacobian accordingly, it may be why you get convergence problems with high "h"

TobiKattmann: Yes good point, I thought about that at some moment but forgot about it 👍

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1099,7 +1099,10 @@ void COutput::SetScreen_Output(const CConfig *config) {
           PrintingToolbox::PrintScreenFixed(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
           break;
         case ScreenOutputFormat::SCIENTIFIC:
-          PrintingToolbox::PrintScreenScientific(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
+          // This line causes SCIENTIFIC output from Marker_Analyze to be written to screen twice after one another.
+          // Thus ruining the nice alignment of the columns in the screen output. Requires debugging.
+          // Try e.g. SURFACE_STATIC_TEMPERATURE with incompressible cases.
+          //PrintingToolbox::PrintScreenScientific(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
```
TobiKattmann, line 1105: Concerning this review comment from @pcarruscag 

Maybe this is a way. Explaining the problem, giving example and commenting out the code. I am not sure to which degree the PerSurface output works (maybe only for history? i dont know) but as no regression test failed so at least for screen output this doesn't crash sth.

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: Good catch again, this indeed resolves the convergence issues with high values of the transfer coefficient 👍 I was now able to easily run the case (shown in a picture above) with h=1e8 instead of a max of h=1e5. That now results in a surface avg surface static temperature at exactly 400K which matches the expectation. Awesome

SU2_CFD/src/solvers/CIncNSSolver.cpp
TobiKattmann: Done

.github/workflows/regression.yml
TobiKattmann: Hm, I of course have to merge su2code/Tutorials/pull/16 first -> revert this line -> merge this PR in order to make it work.
Unfortunately this will break everyone's regression tests that do not merge develop immediately :( If I am not mistaken

Common/src/CConfig.cpp
pcarruscag: +1 on the expressiveness consideration, it also makes the "did you mean" mechanism work better.

SU2_CFD/src/output/COutput.cpp
pcarruscag, line 1105:
```suggestion
          PrintingToolbox::PrintScreenScientific(out, historyOutputPerSurface_Map.at(RequestedField)[0].value, fieldWidth);
```

pcarruscag:
The "per surface" problem starts in `COutput::CheckHistoryOutput` where the table header is not given the right number of columns. The function does not add the "per surface names" because we are using the same identifier for normal outputs and "per surface" outputs.
Moreover, the "per surface" implementation is only considering the first marker name and value for printing.
However, there is one value added per "marker analyse".
Let's fix this in a different PR.

.github/workflows/regression.yml
pcarruscag: Yep just make the change to KT_CONSTANT and then update tutorials once you are ready to merge.

pcarruscag (approved):

.github/workflows/regression.yml
TobiKattmann: 👍

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -507,6 +534,10 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
         Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
       }
 
+      if (kind_boundary == HEAT_TRANSFER){
+        Jacobian.AddVal2Diag(iPoint, nDim+1, Transfer_Coefficient*Area);
```
pcarruscag, line 538: Actually... this is not quite right for compressible flow.
You want the change in heat flux due to a change in solution variables (conservatives) incidentally the temperature depends on all of them, we should at least account for how it varies with rho*E

TobiKattmann: True! I'll fix it

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -535,7 +535,23 @@ void CNSSolver::BC_HeatFlux_Wall_Generic(const CGeometry *geometry, const CConfi
...
+
+        const su2double Density = nodes->GetDensity(iPoint);
+        const su2double Gas_Constant = config->GetGas_ConstantND();
+        const su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[1]);
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        Jacobian_i[nDim+1][0] = Transfer_Coefficient * dTdrho * Area;
+
+        /*--- Take the definition of Temp for an ideal Gas, multiply with rho/rho and derive wrt to conservative variable (rho*E). ---*/
```
pcarruscag, line 551: Division by rho * Cv

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -535,7 +535,23 @@ void CNSSolver::BC_HeatFlux_Wall_Generic(const CGeometry *geometry, const CConfi
...
+        for (auto iVar = 0u; iVar < nVar; ++iVar)
+          Jacobian_i[nDim+1][iVar] = 0.0;
+
+        const su2double Density = nodes->GetDensity(iPoint);
+        const su2double Gas_Constant = config->GetGas_ConstantND();
+        const su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[1]);
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        Jacobian_i[nDim+1][0] = Transfer_Coefficient * dTdrho * Area;
```
TobiKattmann, line 549: I did not really understand why  dT\d rho  is what is displayed there. If someone can give me a pointer that would be appreciated.

It accelerates the convergence by a bit, actually quite a respectable bit from my test with navier_stokes/cylinder

pcarruscag (commented):
You replace u by (rho u)/rho, that is also what gives derivatives wrt the momentum.

SU2_CFD/src/solvers/CNSSolver.cpp
pcarruscag: T = ((rho E)/rho - 1/2 (((rho u)/rho)^2 + ((rho v)/rho)^2 + ((rho w)/rho)^2)) / Cv
T = (U[4] / U[0] - 0.5 * ((U[1] / U[0])^2 + ... )) / Cv

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -507,6 +534,26 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
         Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
       }
 
+      if (kind_boundary == HEAT_TRANSFER){
+
+        /*--- It is necessary to zero the jacobian entries of the energy equation. ---*/
+        for (auto iVar = 0u; iVar < nVar; ++iVar)
+          Jacobian_i[nDim+1][iVar] = 0.0;
```
pcarruscag, line 541:
```suggestion
        if (!dynamic_grid)
          for (auto iVar = 0u; iVar < nVar; ++iVar)
            Jacobian_i[nDim+1][iVar] = 0.0;
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -507,6 +534,26 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        Jacobian_i[nDim+1][0] = Transfer_Coefficient * dTdrho * Area;
+
+        /*--- Take the definition of Temp for an ideal Gas, multiply with rho/rho and derive wrt to conservative variable (rho*E). ---*/
+        Jacobian_i[nDim+1][nDim+1] += Transfer_Coefficient * (Gamma-1.0)/(Gas_Constant*Density) * Area;
+
+        Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
+      }
```
pcarruscag, line 555:
```suggestion
      }
      Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -507,6 +534,26 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
         Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
```
pcarruscag, line 534:
```suggestion
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -504,9 +531,29 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
 
     if (implicit) {
       if (dynamic_grid) {
```
pcarruscag, line 533:
```suggestion
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -504,9 +531,29 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
-        Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
       }
```
pcarruscag, line 535:
```suggestion
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -504,9 +531,29 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        Jacobian_i[nDim+1][0] = Transfer_Coefficient * dTdrho * Area;
+
+        /*--- Take the definition of Temp for an ideal Gas, multiply with rho/rho and derive wrt to conservative variable (rho*E). ---*/
+        Jacobian_i[nDim+1][nDim+1] += Transfer_Coefficient * (Gamma-1.0)/(Gas_Constant*Density) * Area;
+
+      }
+      Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
```
pcarruscag, line 555:
```suggestion
      if (dynamic_grid || (kind_boundary == HEAT_TRANSFER)) {
        Jacobian.AddBlock2Diag(iPoint, Jacobian_i);
      }
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -504,9 +531,29 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+          for (auto iVar = 0u; iVar < nVar; ++iVar)
+            Jacobian_i[nDim+1][iVar] = 0.0;
+
+        const su2double Density = nodes->GetDensity(iPoint);
+        const su2double Gas_Constant = config->GetGas_ConstantND();
+        const su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[1]);
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        Jacobian_i[nDim+1][0] = Transfer_Coefficient * dTdrho * Area;
```
pcarruscag, line 546:
```suggestion
        Jacobian_i[nDim+1][0] += Transfer_Coefficient * dTdrho * Area;
```

SU2_CFD/src/solvers/CNSSolver.cpp
TobiKattmann: Excellent catch!

SU2_CFD/src/solvers/CNSSolver.cpp
TobiKattmann: I totally did not spend the last hour trying to derive the jacobian dT/drho as it is done currently in the code 🐌 (But If I had done that, I would have understood it by now)

I'll update the description and try out the jacobians wrt momentum

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
-      if (dynamic_grid) {
+      if (kind_boundary == HEAT_TRANSFER){
+
+        /*--- It is necessary to zero the jacobian entries of the energy equation. ---*/
+        if (!dynamic_grid)
+          for (auto iVar = 0u; iVar < nVar; ++iVar)
+            Jacobian_i[nDim+1][iVar] = 0.0;
+
+        const su2double Density = nodes->GetDensity(iPoint);
+        const su2double Gas_Constant = config->GetGas_ConstantND();
```
pcarruscag, line 541:
```suggestion
        const su2double oneOnCv = (Gamma - 1.0) / config->GetGas_ConstantND();
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+        const su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[1]);
```
pcarruscag, line 542:
```suggestion
        const su2double Vel2 = nodes->GetVelocity2(iPoint);
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
```
pcarruscag, line 543:
```suggestion
        const su2double dTdrho = oneOnRho * ( -Tinfinity + oneOnCv * 0.5 * Vel2);
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -541,12 +541,16 @@ void CNSSolver::BC_HeatFlux_Wall_Generic(const CGeometry *geometry, const CConfi
...
 
-        /*--- Equivalent to derivation in CNSSolver::BC_Isothermal_Wall_Generic, but no idea why that derivation is correct. ---*/
+        /*--- Total specific energy: e=c_v*T+1/2*v^2 => T=1/c_v(rho*e/rho - 1/2||rho v||^2/rho^2).
+        Together with cv=R/(gamma-1) the following Jacobian contributions for the energy equation can be derived. ---*/
         Jacobian_i[nDim+1][0] += Transfer_Coefficient * dTdrho * Area;
 
-        /*--- Take the definition of Temp for an ideal Gas, multiply with rho/rho and derive wrt to conservative variable (rho*E). ---*/
-        Jacobian_i[nDim+1][nDim+1] += Transfer_Coefficient * (Gamma-1.0)/(Gas_Constant*Density) * Area;
+        for (unsigned short iVar = 1; iVar < nVar-1; iVar++)
+          Jacobian_i[nDim+1][iVar] -= Transfer_Coefficient * dTdrhoe*nodes->GetVelocity(iPoint, iVar-1) * Area;
```
TobiKattmann, line 551: lets remove this and hope no-one notices that I am an idiot... But no-one reads PR's anyway

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
     if (implicit) {
-      if (dynamic_grid) {
+      if (kind_boundary == HEAT_TRANSFER){
+
+        /*--- It is necessary to zero the jacobian entries of the energy equation. ---*/
+        if (!dynamic_grid)
+          for (auto iVar = 0u; iVar < nVar; ++iVar)
+            Jacobian_i[nDim+1][iVar] = 0.0;
+
+        const su2double Density = nodes->GetDensity(iPoint);
```
pcarruscag, line 540:
```suggestion
        const su2double oneOnRho = 1.0 / nodes->GetDensity(iPoint);
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+        const su2double Gas_Constant = config->GetGas_ConstantND();
+        const su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[1]);
+        const su2double dTdrho = 1.0/Density * ( -Tinfinity + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+        const su2double dTdrhoe = (Gamma-1.0)/(Gas_Constant*Density);
```
pcarruscag, line 544:
```suggestion
        const su2double dTdrhoe = oneOnCv * oneOnRho;
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -503,7 +530,30 @@ void CNSSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_container
...
+
+        /*--- Total specific energy: e=c_v*T+1/2*v^2 => T=1/c_v(rho*e/rho - 1/2||rho v||^2/rho^2).
+        Together with cv=R/(gamma-1) the following Jacobian contributions for the energy equation can be derived. ---*/
+        Jacobian_i[nDim+1][0] += Transfer_Coefficient * dTdrho * Area;
+
+        for (unsigned short iVar = 1; iVar < nVar-1; iVar++)
+          Jacobian_i[nDim+1][iVar] -= Transfer_Coefficient * dTdrhoe*nodes->GetVelocity(iPoint, iVar-1) * Area;
```
pcarruscag:
```suggestion
        for (unsigned short iDim = 0; iDim < nDim; iDim++)
          Jacobian_i[nDim+1][iDim+1] -= Transfer_Coefficient * dTdrhoe * nodes->GetVelocity(iPoint, iDim) * Area;
```

SU2_CFD/src/solvers/CNSSolver.cpp
pcarruscag: The velocity is not zero if you have a moving wall, is that what you are thinking?

TobiKattmann: ... 🤦‍♂️

TobiKattmann: Re-added

pcarruscag: xD This would be something to ask Edwin or Tom next Weds, they know more about moving things.
In any case thanks for the side quest into compressible land, it turned out to be a bit more work than expected :grimacing:

TobiKattmann:
Alright, Heat-Transfer BC is good to go. Code factor does its usual thing with Complex code but the rest is green. 
Was a bit more than I anticipated but now there is a compressible version as well alongside the incompressible. 

An incompressible Testcase was added which briefly explains expected results in the extreme ends of heat transfer coefficient values.

What's is left to do for me is to update and merge su2code/su2code.github.io/pull/26 ,In there i will drop a comment or two that advertises the HeatTransfer bc.

Thanks @oleburghardt for kicking this off,  @pedro for the review and advise and @bigfooted for some chats behind the scenes

PR #1228, opened 2021-03-11
Linear solver changes to support hybrid parallel AD

pcarruscag:
## Proposed Changes
Instead of computing transposed matrix vector products, which are harder to parallelize with OpenMP, and having special handling for transposed preconditioners, the matrix is transposed in place, then any preconditioner can be applied.
Some tweaks to support better Krylov discrete adjoint.

## Related Work
#1214

## PR Checklist


Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -945,15 +937,21 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
 
     /*--- Build preconditioner for the transposed Jacobian ---*/
 
+    if (RequiresTranspose) Jacobian.TransposeInPlace();
+
```
pcarruscag, line 941: The transposition happens here.
Things that could go wrong with this approach:
- Solving the same system multiple times, without clearing the matrix or reverting the transposition in between. It does not happen atm.

I think it is worth the risk, since all preconditioners can be used now, and re-transposing the matrix if necessary is not very expensive (probably the same as applying the ILU once).

TobiKattmann: Not that I used a preconditioner besides ILU for DA cases yet, but that is a nice thing 👍

TestCases/parallel_regression_AD.py
```diff
@@ -328,7 +328,7 @@ def main():
     da_sp_pinArray_cht_2d_dp_hf.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_2d"
     da_sp_pinArray_cht_2d_dp_hf.cfg_file  = "DA_configMaster.cfg"
     da_sp_pinArray_cht_2d_dp_hf.test_iter = 100
-    da_sp_pinArray_cht_2d_dp_hf.test_vals = [-4.793283, -4.065832, -4.137121] #last 4 lines
+    da_sp_pinArray_cht_2d_dp_hf.test_vals = [-4.800583, -4.065533, -4.137185] #last 4 lines
```
TobiKattmann, line 331: Well the deviations are really tiny but both inc cases have `INC_ENERGY_EQUATION= YES` but `discadj_cht` is not here as the third case that has that... Not that I am worried but if you have an explanation why exactly those cases differ, I'd be interested?

TobiKattmann (approved):
Well, the code looks nice as always 💐  but I cannot really say much about the changes in general because I am not exactly familiar with that code :(

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -923,7 +913,9 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
 
   if(TapeActive) {
 
-    bool RequiresTranspose = !mesh_deform; // jacobian is symmetric
+    /*--- To keep the behavior of SU2_DOT, but not strictly required since jacobian is symmetric(?). ---*/
+    const bool RequiresTranspose = !mesh_deform || (config->GetKind_SU2() == SU2_DOT);
```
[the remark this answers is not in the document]
pcarruscag: Well the Jacobian is not symmetric because of how we do the boundary conditions but conjugate gradient still works so... :shrug:

Common/include/linear_algebra/CSysSolve.hpp
```diff
@@ -99,8 +99,8 @@ class CSysSolve {
   VectorType* LinSysSol_ptr;        /*!< \brief Pointer to appropriate LinSysSol (set to original or temporary in call to Solve). */
   const VectorType* LinSysRes_ptr;  /*!< \brief Pointer to appropriate LinSysRes (set to original or temporary in call to Solve). */
 
-  LinearToleranceType tol_type = LinearToleranceType::RELATIVE; /*!< \brief How the linear solvers interpret the tolerance. */
-  bool xIsZero = false;           /*!< \brief If true assume the initial solution is always 0. */
+  mutable LinearToleranceType tol_type = LinearToleranceType::ABSOLUTE; /*!< \brief How the linear solvers interpret the tolerance. */
+  mutable bool xIsZero = false;   /*!< \brief If true assume the initial solution is always 0. */
```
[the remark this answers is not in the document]
pcarruscag: Restarted FGMRES (which I put in a function) needs to force those values. But I'll do it some other way, mutable in this case is misleading as relevant state can be different after the solver is called.
I try to reserve mutable for variables that do not change the behavior of the object.

TestCases/parallel_regression_AD.py
pcarruscag: The common denominator seemed to be cases that had a significant number of linear iterations.
I ran them to convergence and there is 0 difference in final results.
I will still test these changes in my optimization cases before merging.

Common/include/option_structure.hpp
```diff
@@ -1875,21 +1875,15 @@ static const MapType<string, ENUM_FFD_BLENDING> Blending_Map = {
...
-  SMOOTHER = 8,             /*!< \brief Iterative smoother. */
-  PASTIX_LDLT = 9,          /*!< \brief PaStiX LDLT (complete) factorization. */
-  PASTIX_LU = 10,           /*!< \brief PaStiX LU (complete) factorization. */
+  CONJUGATE_GRADIENT,   /*!< \brief Preconditionated conjugate gradient method for grid deformation. */
+  FGMRES,               /*!< \brief Flexible Generalized Minimal Residual method. */
+  BCGSTAB,              /*!< \brief BCGSTAB - Biconjugate Gradient Stabilized Method (main solver). */
+  RESTARTED_FGMRES,     /*!< \brief Flexible Generalized Minimal Residual method with restart. */
+  SMOOTHER,             /*!< \brief Iterative smoother. */
+  PASTIX_LDLT,          /*!< \brief PaStiX LDLT (complete) factorization. */
+  PASTIX_LU,            /*!< \brief PaStiX LU (complete) factorization. */
```
TobiKattmann, line 1884: @bigfooted likes that :D

Common/include/linear_algebra/CSysVector.hpp
```diff
@@ -307,15 +307,12 @@ class CSysVector : public VecExpr::CVecExpr<CSysVector<ScalarType>, ScalarType>
...
-        sum = dotRes;
-        const auto mpi_type = (sizeof(ScalarType) < sizeof(double)) ? MPI_FLOAT : MPI_DOUBLE;
-        SelectMPIWrapper<ScalarType>::W::Allreduce(&sum, &dotRes, 1, mpi_type, MPI_SUM, SU2_MPI::GetComm());
-      }
+    /*--- Reduce across all mpi ranks, only master thread communicates. ---*/
+    SU2_OMP_BARRIER
+    SU2_OMP_MASTER {
+      sum = dotRes;
+      const auto mpi_type = (sizeof(ScalarType) < sizeof(double)) ? MPI_FLOAT : MPI_DOUBLE;
+      SelectMPIWrapper<ScalarType>::W::Allreduce(&sum, &dotRes, 1, mpi_type, MPI_SUM, SU2_MPI::GetComm());
```
TobiKattmann, line 315: Resolves #1241 (y)

TestCases/parallel_regression.py
```diff
@@ -1259,7 +1259,7 @@ def main():
     sp_pinArray_cht_2d_dp_hf.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_2d"
     sp_pinArray_cht_2d_dp_hf.cfg_file  = "configMaster.cfg"
     sp_pinArray_cht_2d_dp_hf.test_iter = 100
-    sp_pinArray_cht_2d_dp_hf.test_vals = [0.247022, -0.812199, -0.974877, -0.753315, 208.023676, 349.950000] #last 7 lines
+    sp_pinArray_cht_2d_dp_hf.test_vals = [0.247026, -0.811632, -0.982066, -0.753312, 208.023676, 350.180000] #last 7 lines
     sp_pinArray_cht_2d_dp_hf.su2_exec  = "mpirun -n 2 SU2_CFD"
```
TobiKattmann, line 1263: I will redo the gradient validation on my side... and I like that commit message 👍

PR #1232, opened 2021-03-14
Cleanup COption duplication

pcarruscag:
Templates are awesome

### DIFF
--- a/Common/include/option_structure.inl
+++ b/Common/include/option_structure.inl
@@ -70,220 +70,119 @@ public:
   }
 };
 
-class COptionDouble : public COptionBase {
-  su2double & field; // Reference to the fieldname
-  su2double def; // Default value
-  string name; // identifier for the option
+template<typename Scalar>
+class COptionScalar : public COptionBase {
+protected:
+  Scalar& field; // Reference to the fieldname
+  const Scalar def; // Default value
+  const string name; // identifier for the option
+  const string typeName; // name for the scalar type
 
 public:
-  COptionDouble(string option_field_name, su2double & option_field, su2double default_value) : field(option_field) {
-    this->def = default_value;
-    this->name = option_field_name;
+  COptionScalar() = delete;
+
+  COptionScalar(const string& type_name,
+                const string& option_field_name,
+                Scalar& option_field,
+                Scalar default_value) :
+    field(option_field),
+    def(default_value),
+    name(option_field_name),
+    typeName(type_name) {
   }
 
-  ~COptionDouble() override {};
+  ~COptionScalar() = default;
+
   string SetValue(vector<string> option_value) override {
     COptionBase::SetValue(option_value);
-    // check if there is more than one value
-    string out = optionCheckMultipleValues(option_value, "su2double", this->name);
-    if (out.compare("") != 0) {
-      return out;
-    }
-    istringstream is(option_value[0]);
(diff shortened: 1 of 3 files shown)


Common/include/option_structure.inl
```diff
@@ -70,220 +70,119 @@ public:
...
-    unsigned long val;
-    if (is >> val) {
-      this->field = val;
-      return "";
-    }
-    return badValue(option_value, "unsigned long", this->name);
-  }
-  void SetDefault() override {
-    this->field = this->def;
+class COptionUShort : public COptionScalar<unsigned short> {
```
pcarruscag, line 137: Summary

Common/include/option_structure.inl
```diff
@@ -378,165 +277,86 @@ public:
...
-      unsigned short val;
+      Scalar val;
       if (!(is >> val)) {
-        delete [] vals;
-        return badValue(option_value, "short", this->name);
+        delete [] field;
+        return badValue(option_value, typeName+" list", name);
       }
-      vals[i] = val;
+      field[i] = std::move(val);
```
TobiKattmann, line 321: move instead of copy for that little bit of performance I guess?

TobiKattmann (approved):
Nice clean-up using what c++11 gave us 💐

TobiKattmann:
oh and can one tell LGTM to identify .inl files for analysis as well?

Common/include/option_structure.inl
pcarruscag: I did not bully any feature of the language this time :angel:

Common/include/option_structure.inl
```diff
@@ -70,220 +70,119 @@ public:
...
-  int def; // Default value
-  string name; // identifier for the option
-
+class COptionDouble : public COptionScalar<su2double> {
 public:
-  COptionInt(string option_field_name, int & option_field, int default_value) : field(option_field) {
-    this->def = default_value;
-    this->name = option_field_name;
+  template<class... Ts>
+  COptionDouble(Ts&&... args) :
```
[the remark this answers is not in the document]
pcarruscag: Yes this only works at compile time, the variadics are just to avoid repeating the constructor arguments.
The rvalue ref is to allow this to work with literal values, using just Ts& it would not be possible to bind a literal (or any rvalue) because it does not have an address, in those cases Ts&& collapses to just Ts.

Common/include/option_structure.inl
pcarruscag: For when this deals with strings.

pcarruscag:
I did not notice that it ignored .inl... We should do that and tell it to ignore externals. It recognizes a lgtm.yml file apparently.

pcarruscag:
It does not seem to be possible to include specific files for analysis, only to exclude them from the exclusions (double negative).
But option_structure.inl seems to be analyzed atm https://lgtm.com/projects/g/su2code/SU2/latest/files/Common/include/?sort=name&dir=ASC&mode=list&showExcluded=false

Also found out we can create custom queries https://lgtm.com/help/lgtm/run-queries-database
I'm bout to be public enemy numba one :smiling_imp:

TobiKattmann:
thanks, I didn't forsee that it is a bit more involved tbh.

Oh no, Robopedro will haunt us in future 😨

PR #1236, opened 2021-03-16
Fixed values for turbulence quantities in upstream half-plane

maxaehle:
## Proposed Changes
On a user-specified upstream half-plane, fix the values of turbulence quantities to their far-field values, instead of integrating the turbulence model equations.

In the far-field flow, the production terms of SST for K and omega are zero, while destruction still takes place. Thus the turbulence quantities typically decay on their way from the far-field boundary to the airfoil, until they reach the `lowerlimit`. See for instance this turbulence kinetic energy plot on the NACA0012 mesh:
kk_sst_copr_TKE

This has been presented in [1] together with two solution approaches:
  - *Sustaining terms*, which are additional source terms compensating the destruction terms in the far-field flow. This approach has been already implemented with the `KIND_TURB_MODEL= SST_SUST` option.
    kk_sst_copr_sust_TKE
  - *Floor values*, which is equivalent to set the `lowerlimit` to the far-field values in the upstream region of the airfoil.
This PR adds floor values, but in the form of *fixed values*: The turbulence quantities are just set to the far-field values in the upstream region. This has been implemented in analogy to the strong boundary conditions.

The correction can be activated like this:

    TURB_FIXED_VALUES= YES
    TURB_FIXED_VALUES_DOMAIN= -1.0

The second parameter specifies the half-plane on which fixed values are applied. This half-plane consists of those points for which the dot product of their coordinate vector with the normalized far-field velocity vector is less than `TURB_FIXED_VALUES_DOMAIN`. 

This is the turbulence kinetic energy with AoA=0° and 10°:
kk_sst_copr_floor_TKE
kk_sst_copr_floor_aoa10_TKE

The SA model does not suffer from a decaying turbulence variable, but the proposed changes are also functional there.

[1] Philippe R. Spalart and Christopher L. Rumsey, Effective Inflow Conditions for Turbulence Models in Aerodynamic Calculations, https://arc.aiaa.org/doi/pdf/10.2514/1.29373

## Related Work
?

## PR Checklist


Common/include/CConfig.hpp
```diff
@@ -8051,6 +8054,20 @@ class CConfig {
...
+   * \brief Get whether fixed values for turbulence quantities are applied.
+   * \return <code>TRUE</code> if fixed values are applied; otherwise <code>FALSE</code>.
+   */
+  bool GetTurb_Fixed_Values(void) const { return Turb_Fixed_Values; }
+
+  /*!
+   * \brief Get shift of the upstream half-plane where fixed values for turbulence quantities are applied.
+   * \details This half-plane is given by the condition that the dot product between the
+   * coordinate vector and the normalized far-field velocity vector is less than what this
+   * function returns.
```
pcarruscag, line 8067: Elegant :+1: Hopefully not too much :)

Common/src/CConfig.cpp
```diff
@@ -2105,6 +2105,12 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Direction of the gust X or Y dir */
   addEnumOption("GUST_DIR", Gust_Dir, Gust_Dir_Map, Y_DIR);
 
+  /* Fixed values for turbulence quantities to keep them at inflow conditions. */
+  /* DESCRIPTION: Use fixed values for turbulence quantities. */
+  addBoolOption("TURB_FIXED_VALUES", Turb_Fixed_Values, false);
+  /* DESCRIPTION: Maximal scalar product of the normed far-field velocity vector and a space coordinate. */
+  addDoubleOption("TURB_FIXED_VALUES_DOMAIN", Turb_Fixed_Values_MaxScalarProd, -numeric_limits<su2double>::max());
```
pcarruscag, line 2112: use min instead of -max

SU2_CFD/include/solvers/CTurbSolver.hpp
```diff
@@ -249,6 +251,24 @@ class CTurbSolver : public CSolver {
...
+  /*!
+   * \brief Set the solution using the Freestream values.
+   * \param[in] config - Definition of the particular problem.
+   */
+  inline void SetFreeStream_Solution(const CConfig *config) final {
+    SU2_OMP_FOR_STAT(omp_chunk_size)
+    for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
+      nodes->SetSolution(iPoint, Solution_Inf);
+    }
+  }
```
pcarruscag, line 263: Awesome, more generic.
But pleeeeeeaaaaseeeee do not allocate memory with new/delete (see my love letter to containers) use a std::array of MAXNVAR or even a static array.

SU2_CFD/include/solvers/CTurbSolver.hpp
```diff
@@ -249,6 +251,24 @@ class CTurbSolver : public CSolver {
...
+
+  /*!
+   * \brief Impose fixed values to turbulence quantities.
+   * \details Turbulence quantities are set to far-field values in an upstream half-plane
+   * in order to keep them from decaying.
+   */
+  void Impose_Fixed_Values(CGeometry *geometry, CConfig *config) final;
```
pcarruscag, line 270: const const

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -508,6 +511,24 @@ void CTurbSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_conta
 
 }
 
+void CTurbSolver::Impose_Fixed_Values(CGeometry *geometry, CConfig *config){
+  SU2_OMP_FOR_DYN(omp_chunk_size)
+  for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+    if (geometry->nodes->GetDomain(iPoint)) {
```
pcarruscag, line 517: redundant as you are only going up to nPointDomain

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -508,6 +511,24 @@ void CTurbSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_conta
...
+      su2double* velocity_inf = config->GetVelocity_FreeStreamND();
```
pcarruscag, line 518: const, everything const

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -508,6 +511,24 @@ void CTurbSolver::BC_Fluid_Interface(CGeometry *geometry, CSolver **solver_conta
...
+      su2double sp =
+        GeometryToolbox::DotProduct(nDim, geometry->nodes->GetCoord(iPoint), velocity_inf)
+        / GeometryToolbox::Norm(nDim, velocity_inf);
+      if(sp < config->GetTurb_Fixed_Values_MaxScalarProd()){
+        /*--- Set the solution values and zero the residual ---*/
+        nodes->SetSolution_Old(iPoint, Solution_Inf);
+        nodes->SetSolution(iPoint, Solution_Inf);
+        LinSysRes.SetBlock_Zero(iPoint);
```
pcarruscag, line 526: I think it would be better to also manipulate the Jacobian (1's in the diagonal and desired solution in the RHS, see the wall BC's of the turbulence solvers) as that also improves its condition number.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -290,6 +294,15 @@ void CTurbSSTSolver::Source_Residual(CGeometry *geometry, CSolver **solver_conta
   SU2_OMP_FOR_DYN(omp_chunk_size)
   for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
 
+    /*--- If fixed values apply, there is no need to compute the source term. ---*/
+    if(config->GetTurb_Fixed_Values()){
+      su2double* velocity_inf = config->GetVelocity_FreeStreamND();
+      su2double sp =
+        GeometryToolbox::DotProduct(nDim, geometry->nodes->GetCoord(iPoint), velocity_inf)
+        / GeometryToolbox::Norm(nDim, velocity_inf);
```
pcarruscag, line 302: Is it worth checking the velocity is not zero somewhere?

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -130,8 +130,12 @@ CTurbSSTSolver::CTurbSSTSolver(CGeometry *geometry, CConfig *config, unsigned sh
 
   su2double VelMag2 = GeometryToolbox::SquaredNorm(nDim, VelInf);
 
-  kine_Inf  = 3.0/2.0*(VelMag2*Intensity*Intensity);
-  omega_Inf = rhoInf*kine_Inf/(muLamInf*viscRatio);
+  su2double kine_Inf  = 3.0/2.0*(VelMag2*Intensity*Intensity);
+  su2double omega_Inf = rhoInf*kine_Inf/(muLamInf*viscRatio);
+
+  Solution_Inf = new su2double[2];
```
pcarruscag, line 136: `new` :gun:

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -146,6 +146,11 @@ void CIntegration::Space_Integration(CGeometry *geometry,
     }
   }
 
+  /*--- Fixed values for turbulence quantities. ---*/
+  if(config->GetTurb_Fixed_Values()){
+    solver_container[MainSolver]->Impose_Fixed_Values(geometry, config);
+  }
```
pcarruscag, line 152: I would like something more generic here, for example call the method by default and let it decide whether it does anything or not.

At some point I want to grab all the logic of calling the different boundary conditions and move it to a single method of the solvers, to reduce the verbosity of CSolver's interface.

Common/src/CConfig.cpp
maxaehle: I'll use `lowest`

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
maxaehle: Yes, I'll check and also normalize the velocity vector outside the loop.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
maxaehle: Ok it's `su2double Solution_Inf[MAXNVAR]` now

SU2_CFD/src/solvers/CTurbSolver.cpp
maxaehle: I'll add the `Jacobian.DeleteValsRowi`

SU2_CFD/src/integration/CIntegration.cpp
maxaehle: I made some changes in 5654648. Do you think one should give this an even more generic flavor, e.g. by renaming the function `Impose_Fixed_Values`?

SU2_CFD/src/solvers/CTurbSolver.cpp
maxaehle: ok

SU2_CFD/src/solvers/CTurbSolver.cpp
maxaehle: yep

SU2_CFD/include/solvers/CTurbSolver.hpp
maxaehle: ok

SU2_CFD/include/solvers/CTurbSolver.hpp
maxaehle: Ok it's `su2double Solution_Inf[MAXNVAR]` now

maxaehle:
Do you think a test case for this is necessary?

SU2_CFD/src/integration/CIntegration.cpp
pcarruscag: It is fine for now, in the future it can become part of the "BC method" I mentioned.

pcarruscag:
Thank you for the changes.
I would say a test case is always welcome. You can simply modify an existing one, this feature is orthogonal to everything else, and then please add the new options to the config_template (with maybe the nice explanation you have in CConfig).

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -290,6 +293,15 @@ void CTurbSSTSolver::Source_Residual(CGeometry *geometry, CSolver **solver_conta
   SU2_OMP_FOR_DYN(omp_chunk_size)
   for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
 
+    /*--- If fixed values apply, there is no need to compute the source term. ---*/
+    if(config->GetTurb_Fixed_Values()){
+      const su2double* velocity_inf = config->GetVelocity_FreeStreamND();
+      su2double sp =
+        GeometryToolbox::DotProduct(nDim, geometry->nodes->GetCoord(iPoint), velocity_inf)
```
pcarruscag, line 300: Would it be appropriate to take the coordinate relative to the reference point provided in the config?
I guess in practice one would set the plane offset very large, but you never know what folks will do with your code (for better or worse).

pcarruscag (approved):
LGTM but present it in the meeting tomorrow before you merge.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
pcarruscag: Hmmmm you now have to check for zero velocity in two places as well...

maxaehle: Right ... I think I will revert the changes in `CTurbSSTSolver::Source_Residual`, so that the source residual is computed even if the residual is overwritten later.

maxaehle: How is a reference point specified in the config?

maxaehle: Probably the `REF_ORIGIN_MOMENT_X`, _Y, _Z. Yes giving the shift relative to these can be a sensible choice, let's discuss in the meeting.

maxaehle:
After looking for testcases to modify, I think it is better to have the modified testcase in addition to, and not besides, the original one. Other users might want to build their cfg file on top of the original testcase cfg file, and could be surprised by the additional feature.

maxaehle:
Now the PR is ready to be merged from my perspective.

pcarruscag:
Alright :+1:

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
```diff
@@ -0,0 +1,260 @@
...
+MARKER_PLOTTING= ( airfoil )
+%
+% Marker(s) of the surface where the functional (Cd, Cl, etc.) will be evaluated
+MARKER_MONITORING= ( airfoil )
+
+% ------------- COMMON PARAMETERS DEFINING THE NUMERICAL METHOD ---------------%
+%
+% Numerical method for spatial gradients (GREEN_GAUSS, WEIGHTED_LEAST_SQUARES)
+NUM_METHOD_GRAD= WEIGHTED_LEAST_SQUARES
+NUM_METHOD_GRAD_RECON= LEAST_SQUARES
```
TobiKattmann, line 94: Wait, I recall there were some accuracy problems with LSQ gradient. Introduced in/ and see the discussion in #790 . Was it fixed in the meantime or were the issues not that bad anyway @pcarruscag ?

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
```diff
@@ -0,0 +1,260 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: upstream turbulence quantities fixed to far-field value    %
+% Date: Mar 17th, 2021                                                         %
+% File Version 7.1.1 "Blackbird"                                               %
+%                                                                              %
```
TobiKattmann, line 7: I think it is a nice touch to put your name (or some identifier) here so it is directly clear who did this :) If it is intentional for privacy reasons then sure, but as you seem to use your actual name in github i guess thats not the issue here

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
```diff
@@ -0,0 +1,260 @@
...
+%
+% Jacobi implicit smoothing of the correction
+MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0, 0, 0 )
+%
+% Damping factor for the residual restriction
+MG_DAMP_RESTRICTION= 0.75
+%
+% Damping factor for the correction prolongation
+MG_DAMP_PROLONGATION= 0.75
+
```
TobiKattmann, line 153: I am a big fan of keeping the cfg's to its minimum. All these options (and quite a few more) can be deleted which makes potential change of cfg option much simpler down the road. Additionally if someone picks up this cfg it is a lot less overwhelming. The copy paste of complete cfg's and change of a single option is not really helpful imo. But then again here it is prob consistent with the existing Testcase which is prob in that repo (yep it is with the sst_sust)... so maybe it is the better choice here

TestCases/serial_regression.py
```diff
@@ -363,6 +363,17 @@ def main():
     turb_naca0012_sst_sust.tol       = 0.00001
     test_list.append(turb_naca0012_sst_sust)
 
+    # NACA0012 (SST, fixed values for turbulence quantities)
+    turb_naca0012_sst_fixedvalues           = TestCase('turb_naca0012_sst_fixedvalues')
+    turb_naca0012_sst_fixedvalues.cfg_dir   = "rans/naca0012"
+    turb_naca0012_sst_fixedvalues.cfg_file  = "turb_NACA0012_sst_fixedvalues.cfg"
+    turb_naca0012_sst_fixedvalues.test_iter = 10
+    turb_naca0012_sst_fixedvalues.test_vals = [-9.562435, -1.566603, 1.022029, 0.040549]
```
TobiKattmann, line 371: just a note: One can add as many screen outputs as one wants here. I think it is a good thing to add some (even arbitrary) screen outputs in order to slightly increase the coverage. It might save the e.g. LINSOL_RESIDUAL screen output in that solver one day.

config_template.cfg
```diff
@@ -192,6 +192,12 @@ FREESTREAM_VISCOSITY= 1.853E-5
 % Free-stream turbulence intensity
 FREESTREAM_TURBULENCEINTENSITY= 0.05
 %
+% Fix turbulence quantities to far-field values inside an upstream half-space
+TURB_FIXED_VALUES= NO
+%
+% Shift of the fixed values half-space, in space units in the direction of far-field velocity
+TURB_FIXED_VALUES_DOMAIN= -1.0
```
TobiKattmann, line 199: I didnt understand this explanation in the config_template and with your PR explanation and 1 min on sheet of paper it was easy ... 
`The second parameter specifies the half-plane on which fixed values are applied. This half-plane consists of those points for which the dot product of their coordinate vector with the normalized far-field velocity vector is less than TURB_FIXED_VALUES_DOMAIN.`
Imo it is ok to be verbose in the cfg_template 
(Now that I got it, this short sentence makes sense as well... 🙈 i think the space units 🚀 👨‍🚀 threw me off)

TobiKattmann (commented):
Thanks @maxaehle for this addition 💐 lgtm with good style overall 👍

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
maxaehle: I modified the `turb_NACA0012_sst_sust.cfg` file by @economon, so I was unsure which author name(s) and institution(s) to use. Would it be appropriate to write

    % Author: Max Aehle
    % Institution: TU Kaiserslautern

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
maxaehle: Yes, this is a trade-off between conciseness and demonstrating what exactly is the feature... I would like to leave it as it is

config_template.cfg
maxaehle: Yeah maybe "spatial units" or "length units" are better terms

TestCases/serial_regression.py
maxaehle: You mean to add `LINSOL_RESIDUAL` to the line

    SCREEN_OUTPUT= (INNER_ITER, RMS_DENSITY, RMS_TKE, RMS_DISSIPATION, LIFT, DRAG)

I'd then just update the new testcase, or is it better to cover it in all of the three testcases of this family (pure SST, SST_SUST, and fixedvalues)?

TestCases/rans/naca0012/turb_NACA0012_sst_fixedvalues.cfg
TobiKattmann: If one wants to concise and demonstrate what the feature is one should remove superfluous/unused options from one's config ... like those above. The gained space can be used to point out what the options do and maybe give an example what to expect from a certain change in order to increaser user intuition. It is always nicer to add stuff but that leads to PR's like #1129 because already dead options where copy-pasted around in new Testcases. 
I am ok with leaving this one as is but for future PR's all those unused (if applicable) `MG_` `_ADJ_` and other should just be deleted.

TestCases/serial_regression.py
TobiKattmann: It is a more a general thing (for future PR's) because in this case you could also just add the 5th line which would be RMS_DENSITY which would already help increasing test coverage (and maybe even INNER_ITER as 6th one). You can add as many 'test.vals' as you like and should maybe use that to safeguard some things which might otherwise go untested. So if you had
``` 
SCREEN_OUTPUT= (INNER_ITER, RMS_DENSITY, RMS_TKE, RMS_DISSIPATION, LIFT, DRAG, LINSOL_RESIDUAL)
```
you could 
```python
turb_naca0012_sst_fixedvalues.test_vals = [9, -3.456789, -9.562435, -1.566603, 1.022029, 0.040549, 0.012345]
```
with the first 2 and the last value made up.

pcarruscag: You need to leave inner iter out of the test_vals, but include it in the config.
+1 vote to all your comments.

maxaehle:
@TobiKattmann @pcarruscag: I tried to follow all of your proposals in the new branch `fix_turb_fixed_values`. I removed the multigrid stuff from the NACA 0012 SST_SUST and fixed_values cfg, but not of the pure SST one. And I added verification of LINSOL_RESIDUAL and RMS_DENSITY to all the three tests. We can open a new PR or merge this in an existing one.

Would you like to also remove the adjoint stuff from the cfgs? And right now the fixed_values test does not start from a restart solution, could that be a problem?

maxaehle:
We could also write a script that detects cfg options whose removal does not break the regression ;-)

pcarruscag:
Ah! Cool, sure open a new PR @maxaehle.
What adjoint stuff? file names and so on?
I think removing irrelevant options would be make the tests clearer, there are lots with Roe and JST options specified and vice versa which probably confuses new users.
I'm not so sure about removing all defaults thought... On one hand it would serve as regression for the default values set by CConfig, on the other it hides the tuning parameters of some methods... but then again those are more or less documented now.
:shrug:

PR #1238, opened 2021-03-18
Some fixes / cleanup of CONV_CRITERIA which has no effect and will be deprecated in 7.2.0

TobiKattmann:
## Proposed Changes
- remove unused CONV_CRITERIA from code and Testcases (223)
- remove unused Set/GetGeometryPlanes funciton
- Re-add SURFACE_STATIC_TEMPERATURE objective function for flow_solver
 
## PR Checklist


Common/src/CConfig.cpp
```diff
@@ -2887,6 +2884,8 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
             newString.append("WRT_SOL_FREQ is deprecated. Use OUTPUT_WRT_FREQ instead.\n\n");
           if (!option_name.compare("WRT_SOL_FREQ_DUALTIME"))
             newString.append("WRT_SOL_FREQ_DUALTIME is deprecated. Use OUTPUT_WRT_FREQ instead.\n\n");
+          if (!option_name.compare("CONV_CRITERIA"))
+            newString.append("CONV_CRITERIA is deprecated.\n\n");
```
pcarruscag, line 2888: Dammmm really?! I guess if something is a residual it uses the residual criteria (criterion?) and if it is a coefficient it uses Cauchy?
Mind throwing that in the message? This is another one that will break every config...

TobiKattmann: Exactly. In COutput::Convergence_Monitoring there is a distinction for `CONV_FIELD` between `HistoryFieldType::COEFFICIENT` and `HistoryFieldType::RESIDUAL`.

I'll add that information to the error msg 👍

pcarruscag: actually... do you think we should give folks a pass and just print a warning (with that exact message) and say that we will deprecate in the future? When we go to 7.2.0 (which could happen soon if we merge the hybrid parallel AD and the Scalar solvers)

TobiKattmann: Sure, that also requires readding the option. I'll try to remember to enforce that for 7.2.0

TobiKattmann: Ok I made it a warning but as it is thrown at config parsing in Postprocessing I guess not lot of people will see it anyway. 

I guess one can `git revert aa7393f` at some point in the future.

And CONV_CRITERIA is a stringOption so one can type in there whatever one likes :) no optionMap required

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -173,7 +173,7 @@ void CMultizoneDriver::StartSolver() {
   if (rank == MASTER_NODE){
     cout << endl <<"Simulation Run using the Multizone Driver" << endl;
     if (driver_config->GetTime_Domain())
-      cout << "The simulation will run for " << driver_config->GetnTime_Iter() << " time steps." << endl;
+      cout << "The simulation will run until time step " << driver_config->GetnTime_Iter() << "." << endl;
```
TobiKattmann, line 176: In SingleZoneDriver `nTimeIter - Restart_Iter` can to underflow of the unsigned long result which is now checked in CConfig

pcarruscag (approved):

TobiKattmann:
Thanks for the approval! 🦃

PR #1240, opened 2021-03-22
Continuation of #1236

maxaehle:
## Proposed Changes
Remove ununsed multigrid options in cfg file of the new NACA 0012 fixed_values testcase and the SST_SUST testcase, and some other small changes.

## Related Work
Continuation of #1236, comments by @pcarruscag  and @TobiKattmann 

## PR Checklist


maxaehle:
We could also remove the unused `*_ADJ_*` options from the cfg file

pcarruscag (approved):

maxaehle:
Commit 1bc9649 reverts my changes to those residuals in `hybrid_regression.py` which were covered by the regression tests already before. (Oddly enough, on my system the residuals of `SU2_CFD -t 2` were different from those computed by the Github pipeline. Maybe this is due to floating-point discrepancies when arithmetic operations are rearranged?)

pcarruscag:
That is possible the hybrid regression uses single precision for linear solvers.

TobiKattmann:
👍 for doing this @maxaehle . 
- I am in favor of removing the `_ADJ_` options  with the `GRAD_OBJFUNC_FILENAME`.
- As Roe is used the `JST_SENSOR_COEFF` can go. Limiter are turned off so either the slope limiter section with 2 options can be removed or be moved down below the `FLOW NUMERICAL METHOD DEFNITION`.
- `CONV_CRITERIA` will be gone once you merge develop and you can add `CONV_FIELD=RMS_DENSITY` as that is also the default. Removign the `_CAUCHY_` options is prob debatable, but in this very cfg they are unused. 
- The `CFL_ADAPT*` stuff can go (unless the values are chosen in such a way by you that they do work well, otherwise it might even do negative advertisement for an option)

I guess I am also a bit aggressive on stripping options from files, so take it with a grain of salt 🧂 .

Another aspect ,which is of course nothing to be changed in this PR and is also a bit off-topic, is that lists like `SOLVER` change over time but the list in the comment over the option in the Testcase cfg does not get updated (there might of course be good examples of people doing so ;) ). `WAVE_EQUATION` and `POISSON_EQUATION` e.g. are not available anymore, yet they are in the cfg description of `SOLVER` here... and 200 other Testcases. So a way might be to strip those and sth like 
```
% See the config_template.cfg in the code root for more information on the options
```
... although the info there are not always perfect either, but one place to maintain is easier than a couple hundred

TobiKattmann (approved):

TobiKattmann:
Hey @maxaehle , if you don't want to make further changes you can hit the merge if you like, once the tests pass.

PR #1242, opened 2021-03-23
Make "enum" : "enum class" thus removing enum numbers

TobiKattmann:
## Proposed Changes
Hi all,

The title gives it away. I imagine that piece by piece some commits are chipped in to remove more and more enum numbers. Like that it is more obvious if sth breaks because somewhere someone uses builds on knowing the actual number to do some shenanigans. 
Maybe just removing all in a big blow saves time bit #1220 already shows some limitations.

I used the following command (for streamwise periodic in this case) and looked over the results whether I notice sth fishy. Not perfect but might help catchin some easy problems.
```bash
grep -r 'NO_STREAMWISE_PERIODIC\|PRESSURE_DROP\|STREAMWISE_MASSFLOW\|Streamwise_Periodic_Map\|GetKind_Streamwise_Periodic' SU2_CFD SU2_DOT SU2_GEO SU2_DEF Common --include=*cpp --include=*hpp --include=*inl
```

So feel free to commit  your removals in here and if it breaks and one doesn't want to search for it just revert that and mention the specific enum so that no-one repeats that

Happy enum-removal  

## Related Work
@bigfooted mentioned that a couple of times and #1220 discusses some limitations

## PR Checklist


pcarruscag:
Provocative thought of the day: Removing the numbers from the enums is a bit of spring cleaning, the "modern C++" way is to use `enum class` and to never rely on implicit conversions from enum to integer type.
In any case, it should be safe to remove the numbers when they already start at zero.

TobiKattmann:
> the "modern C++" way is to use enum class and to never rely on implicit conversions from enum to integer type.

I am in. It is a little (actually quite a bit) more involved as the enum namespace has to be given now everywhere an entry is used.
I was not familiar with `enum` vs `enum class` if someone else needs an entry point [stack overflow](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum), [playin around in compiler explorer](https://gcc.godbolt.org/z/5YTd4dPoE)... oh and of course the [c++ core guidelines entry endorses the use of enum class as well](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class)

TobiKattmann:
> the "modern C++" way is to use enum class and to never rely on implicit conversions from enum to integer type.

I am in. It is a little (actually quite a bit) more involved as the enum namespace has to be given now everywhere an entry is used.
I was not familiar with `enum` vs `enum class` if someone else needs an entry point [stack overflow](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum), [playin around in compiler explorer](https://gcc.godbolt.org/z/5YTd4dPoE)... oh and of course the [c++ core guidelines entry endorses the use of enum class as well](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class)

Common/include/option_structure.inl
```diff
@@ -70,6 +70,51 @@ public:
   }
 };
 
+
+template <class Tenum>
+class COptionEnumClass : public COptionBase {
+
+  map<string, Tenum> m;
+  Tenum & field; // Reference to the fieldname
```
pcarruscag, line 78: Hmmmm not to be picky, but :)
- You can avoid duplicating this class by giving it 2 template parameters, one for the enum type, the other for the type of the field. The first type is always the enum type, the second can be either the conventional unsigned short or the enum type again. Template parameters can have default values, so you even get to do this with minimal changes.
- I never noticed that this class made a copy of the map... please make it a const reference, and pass it by reference also in the constructor.

Common/src/CConfig.cpp
```diff
@@ -355,6 +355,17 @@ void CConfig::addEnumOption(const string name, unsigned short & option_field, co
   return;
 }
 
+// enum types work differently than all of the others because there are a small number of valid
+// string entries for the type. One must also provide a list of all the valid strings of that type.
+template <class Tenum>
+void CConfig::addEnumClassOption(const string name, Tenum & option_field, const map<string, Tenum> & enum_map, Tenum default_value) {
```
TobiKattmann, line 361: I just swapped the `Tenum & option_field` from `unsigned short & option_field` and had to pull that through to `COptionEnumClass` ... it is code duplication but there is a more elegant solution without redoing more then feel free to revert and push, I am interested. 
All the `Kind_stuff` vars are unsigned shorts .. if they were their respective enum type the `addEnumClassOption` could handle those as well within the same implementation.

Common/include/option_structure.hpp
```diff
@@ -2243,15 +2243,15 @@ static const MapType<string, ENUM_VERIFICATION_SOLUTIONS> Verification_Solution_
...
+  PRESSURE_DROP, /*!< \brief Prescribed pressure drop. */
+  MASSFLOW,      /*!< \brief Prescribed massflow. */
 };
 static const MapType<string, ENUM_STREAMWISE_PERIODIC> Streamwise_Periodic_Map = {
-  MakePair("NONE",          NO_STREAMWISE_PERIODIC)
-  MakePair("PRESSURE_DROP", PRESSURE_DROP)
-  MakePair("MASSFLOW",      STREAMWISE_MASSFLOW)
+  MakePair("NONE",          ENUM_STREAMWISE_PERIODIC::NONE)
+  MakePair("PRESSURE_DROP", ENUM_STREAMWISE_PERIODIC::PRESSURE_DROP)
+  MakePair("MASSFLOW",      ENUM_STREAMWISE_PERIODIC::MASSFLOW)
```
pcarruscag, line 2254: It ends up being a little more verbose but you get to use "names" that would otherwise conflict with what is already in use for objective functions and things like that.

Common/src/CConfig.cpp
TobiKattmann: ups ... didn't see you already commented when I wrote this

Common/include/option_structure.inl
TobiKattmann: Don't hesitate to be picky 🔍 

Thats an easy and straight forward solution 👍 for the first point. The second was a bit trickier and I had to un`override` the `SetValue` function for now as this line 
```c++
Tenum val = this->m[option_value[0]];
```
is violating the const of m indirectly. So I made the member function `SetValue` const to calm down the compiler but then I had to remove the override to make it compile. I dont fully understand what is going here and there is surely a better way, so my current version is a version-to-be-improved.

TobiKattmann: ok that const ref thing compiled for me but not for github, so I reverted the change as it was not really good anyway...

Common/include/option_structure.inl
```diff
@@ -26,48 +26,50 @@
...
-    Tenum val = this->m[option_value[0]];
-    this->field = val;
+    field = static_cast<TField>(it->second);
     return "";
   }
 
-  void SetDefault() override {
-    this->field = this->def;
-  }
+  void SetDefault() override { field = static_cast<TField>(def); }
```
pcarruscag, line 72: I added the static cast but this was not smart :facepalm: (it would make enum class compatible with unsigned short...)

Common/include/option_structure.inl
```diff
@@ -186,55 +188,59 @@ public:
   }
 };
 
-template <class Tenum>
-class COptionEnumList : public COptionBase {
+template <class Tenum, class TField>
+class COptionEnumList final : public COptionBase {
```
pcarruscag, line 192: available for lists

Common/src/geometry/CGeometry.cpp
```diff
@@ -3108,7 +3108,7 @@ void CGeometry::ComputeSurf_Curvature(CConfig *config) {
 }
 
 void CGeometry::FilterValuesAtElementCG(const vector<su2double> &filter_radius,
-                                        const vector<pair<unsigned short,su2double> > &kernels,
+                                        const vector<pair<ENUM_FILTER_KERNEL,su2double> > &kernels,
```
pcarruscag, line 3111: enum class is the safe way to remove the numbers... it breaks the code everywhere :)

Common/include/option_structure.inl
pcarruscag: sorted

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1597,7 +1597,7 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
 
       /*--- Get the physical time. ---*/
       su2double time = 0.0;
-      if (config->GetTime_Marching()) time = config->GetPhysicalTime();
+      if (config->GetTime_Marching() != TIME_MARCHING::STEADY) time = config->GetPhysicalTime();
```
TobiKattmann, line 1600: Watch out for: implicitly relying that STEADY evaluates to zero/false and everything else to >0 and therefore true.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1087,7 +1087,7 @@ void CEulerSolver::SetNondimensionalization(CConfig *config, unsigned short iMes
   su2double Beta          = config->GetAoS()*PI_NUMBER/180.0;
   su2double Mach          = config->GetMach();
   su2double Reynolds      = config->GetReynolds();
-  bool unsteady           = (config->GetTime_Marching() != NO);
+  bool unsteady           = (config->GetTime_Marching() != TIME_MARCHING::STEADY);
```
TobiKattmann, line 1090: using another enum's value like NO in this case. Note to self: Maybe introduce an `inline bool isdual_time` and `isunsteady` in option_structure as it is used quite often

pcarruscag (approved):

TobiKattmann:
Merging this now to keep the PR and the number of problems, that will arise in other code, in bounds. This work will be continued in a future PR.

If your code breaks after merging this PR you most likely just have to add an `<ENUM_NAME>::` in front of your enum variable.

Thanks @pcarruscag for the idea and the setup in `option_structure.inl`

PR #1249, opened 2021-03-27
Fix some reconstruction gradient issues on periodic boundaries (when NUM_METHOD_GRAD != NUM_METHOD_GRAD_RECON))

pcarruscag:


Common/src/CConfig.cpp
```diff
@@ -4935,6 +4935,9 @@ void CConfig::SetPostprocessing(unsigned short val_software, unsigned short val_
 
   }
 
+  if (ReconstructionGradientRequired && GetFluidProblem() && Kind_ConvNumScheme_Flow == SPACE_CENTERED)
+    SU2_MPI::Error("For centered schemes the option NUM_METHOD_GRAD_RECON should not be set.", CURRENT_FUNCTION);
+
```
pcarruscag, line 4940: And all I wanted was to put this error... Ignorance is bliss.

Common/include/option_structure.hpp
```diff
@@ -2107,57 +2107,62 @@ static const MapType<string, ENUM_INPUT_REF> Input_Ref_Map = {
...
+  PERIODIC_SOL_LS     ,  /*!< \brief Solution gradient communication for weighted Least Squares (periodic only). */
+  PERIODIC_PRIM_LS    ,  /*!< \brief Primitive gradient communication for weighted Least Squares (periodic only). */
+  PERIODIC_SOL_ULS    ,  /*!< \brief Solution gradient communication for unwieghted Least Squares (periodic only). */
+  PERIODIC_PRIM_ULS   ,  /*!< \brief Primitive gradient communication for unweighted Least Squares (periodic only). */
+  PERIODIC_SOL_GG_R   ,  /*!< \brief Same but reconstruction. */
+  PERIODIC_PRIM_GG_R  ,  /*!< \brief Same but reconstruction. */
+  PERIODIC_SOL_LS_R   ,  /*!< \brief Same but reconstruction. */
+  PERIODIC_PRIM_LS_R  ,  /*!< \brief Same but reconstruction. */
+  PERIODIC_SOL_ULS_R  ,  /*!< \brief Same but reconstruction. */
+  PERIODIC_PRIM_ULS_R ,  /*!< \brief Same but reconstruction. */
```
pcarruscag, line 2128: These are needed to target the reconstruction gradient during periodic communications, otherwise not only was it not communicated, but it also corrupted the "normal" gradient for viscous and source terms.

pcarruscag, line 2128: These are needed to target the reconstruction gradient during periodic communications, otherwise not only was it not communicated, but it also corrupted the "normal" gradient for viscous and source terms.

Common/include/option_structure.hpp
```diff
@@ -2107,57 +2107,62 @@ static const MapType<string, ENUM_INPUT_REF> Input_Ref_Map = {
...
-  NEIGHBORS            = 24,  /*!< \brief Neighbor point count communication (for JST). */
-  SOLUTION_FEA         = 25,  /*!< \brief FEA solution communication. */
-  MESH_DISPLACEMENTS   = 27,  /*!< \brief Mesh displacements at the interface. */
-  SOLUTION_TIME_N      = 28,  /*!< \brief Solution at time n. */
-  SOLUTION_TIME_N1     = 29,  /*!< \brief Solution at time n-1. */
-  PRIMITIVE            = 30   /*!< \brief Primitive solution communication. */
+  SOLUTION             ,  /*!< \brief Conservative solution communication. */
+  SOLUTION_OLD         ,  /*!< \brief Conservative solution old communication. */
+  SOLUTION_GRADIENT    ,  /*!< \brief Conservative solution gradient communication. */
+  SOLUTION_GRAD_REC    ,  /*!< \brief Conservative solution reconstruction gradient communication. */
```
pcarruscag, line 2143: We used to communicate (normal MPI, not the periodic stuff) normal+reconstruction gradient at the same time, even when reconstruction==normal, not anymore.

SU2_CFD/include/variables/CEulerVariable.hpp
```diff
@@ -135,38 +125,19 @@ class CEulerVariable : public CVariable {
...
-  }
-
-  /*!
-   * \brief Set the gradient of the primitive variables.
-   * \param[in] iVar - Index of the variable.
-   * \param[in] value - Value of the gradient.
-   */
-  inline void SetLimiter_Primitive(unsigned long iPoint, unsigned long iVar, su2double value) final {
-    Limiter_Primitive(iPoint,iVar) = value;
-  }
```
pcarruscag, line 155: This type of Set was removed, the comms operate on the entire container, which also makes things more efficient by avoiding virtual access in every (iPoint,iVar,iDim) triplet.

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -151,29 +153,7 @@ void CNEMONSSolver::SetPrimitive_Gradient_GG(CGeometry *geometry, const CConfig
 
   const auto& primitives = nodes->GetPrimitive_Aux();
 
-  computeGradientsGreenGauss(this, PRIMITIVE_GRADIENT, PERIODIC_PRIM_GG, *geometry,
-                             *config, primitives, 0, nPrimVarGrad, gradient);
-}
-
-void CNEMONSSolver::SetPrimitive_Gradient_LS(CGeometry *geometry, const CConfig *config, bool reconstruction) {
```
pcarruscag, line 158: This was the same as the base method.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1359,50 +1220,23 @@ void CSolver::CompletePeriodicComms(CGeometry *geometry,
...
-                }
-              }
+                  (*gradient)(iPoint, iVar, iDim) += bufDRecv[buf_offset+iVar*nDim+iDim];
 
               break;
 
+            case PERIODIC_SOL_LS: case PERIODIC_SOL_ULS:
+            case PERIODIC_SOL_LS_R: case PERIODIC_SOL_ULS_R:
             case PERIODIC_PRIM_LS: case PERIODIC_PRIM_ULS:
+            case PERIODIC_PRIM_LS_R: case PERIODIC_PRIM_ULS_R:
```
pcarruscag, line 1239: We have more comm types but less comm code, because a gradient is a gradient and a limiter is a limiter.
If you review these changes be on the look out for uses of nVar and nPrimVarGrad instead of something that looks more generic like COUNT_PER_POINT, ICOUNT, or nVarGrad (I have introduced bugs like that in the past with this type of refactoring...)

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -744,16 +782,16 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
             /*--- Get conservative solution and rotate if necessary. ---*/
 
-            for (iVar = 0; iVar < nVar; iVar++)
-              rotPrim_i[iVar] = base_nodes->GetSolution(iPoint, iVar);
+            for (iVar = 0; iVar < ICOUNT; iVar++)
+              rotPrim_i[iVar] = field(iPoint, iVar);
 
             if (rotate_periodic) {
-              Rotate(zeros, &base_nodes->GetSolution(iPoint)[1], &rotPrim_i[1]);
+              Rotate(zeros, &field(iPoint,1), &rotPrim_i[1]);
```
pcarruscag, line 789: Periodic comms with rotation will not work for NEMO.
This velocity offset should be generalized.

SU2_CFD/src/solvers/CSolver.cpp
TobiKattmann: And you tell me that now that I am close to the end of the diff... ok I scroll through it again

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1831,8 +1622,9 @@ void CSolver::CompleteComms(CGeometry *geometry,
               base_nodes->SetUnd_Lapl(iPoint, iVar, bufDRecv[buf_offset+iVar]);
             break;
           case SOLUTION_LIMITER:
-            for (iVar = 0; iVar < nVar; iVar++)
-              base_nodes->SetLimiter(iPoint, iVar, bufDRecv[buf_offset+iVar]);
+          case PRIMITIVE_LIMITER:
+            for (iVar = 0; iVar < COUNT_PER_POINT; iVar++)
+              (*limiter)(iPoint,iVar) = bufDRecv[buf_offset+iVar];
```
TobiKattmann, line 1627: Whats that? Is `(*limiter)` a cast, no. Or are the brackets of `(iPoint,iVar)` an access operator for the limiter container that is created a bit above (yep thats more likely)
```c++
auto limiter = &base_nodes->GetLimiter();
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2312,28 +2086,33 @@ void CSolver::SetSolution_Gradient_GG(CGeometry *geometry, const CConfig *config
...
 
   const auto& solution = base_nodes->GetSolution();
   auto& rmatrix = base_nodes->GetRmatrix();
   auto& gradient = reconstruction? base_nodes->GetGradient_Reconstruction() : base_nodes->GetGradient();
-  PERIODIC_QUANTITIES kindPeriodicComm = weighted? PERIODIC_SOL_LS : PERIODIC_SOL_ULS;
+  const auto comm = reconstruction? SOLUTION_GRAD_REC : SOLUTION_GRADIENT;
 
-  computeGradientsLeastSquares(this, SOLUTION_GRADIENT, kindPeriodicComm, *geometry, *config,
-                               weighted, solution, 0, nVar, gradient, rmatrix);
+  computeGradientsLeastSquares(this, comm, commPer, *geometry, *config, weighted, solution, 0, nVar, gradient, rmatrix);
```
TobiKattmann, line 2115: Is this a friendly `nVar`? a bit above as well

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1547,11 +1325,13 @@ void CSolver::GetCommCountAndType(const CConfig* config,
...
     case SOLUTION_GRADIENT:
-      COUNT_PER_POINT  = nVar*nDim*2;
+    case SOLUTION_GRAD_REC:
+      COUNT_PER_POINT  = nVar*nDim;
       MPI_TYPE         = COMM_TYPE_DOUBLE;
       break;
     case PRIMITIVE_GRADIENT:
-      COUNT_PER_POINT  = nPrimVarGrad*nDim*2;
+    case PRIMITIVE_GRAD_REC:
+      COUNT_PER_POINT  = nPrimVarGrad*nDim;
```
TobiKattmann, line 1334: nVar and nPrimVarGrad?

TobiKattmann (approved):
Thanks for this PR💐  No major comments and mostly questions

Common/include/option_structure.hpp
```diff
@@ -2107,57 +2107,62 @@ static const MapType<string, ENUM_INPUT_REF> Input_Ref_Map = {
  * \brief Vertex-based quantities exchanged during periodic marker communications.
  */
 enum PERIODIC_QUANTITIES {
```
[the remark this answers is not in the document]
TobiKattmann: ... bonus points if you do it anyway

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -266,6 +282,56 @@ void CSolver::GetPeriodicCommCountAndType(const CConfig* config,
   }
 }
 
+namespace PeriodicCommHelpers {
```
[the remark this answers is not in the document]
TobiKattmann: ah.. code duplication. reading commit msgs helps

SU2_CFD/include/variables/CAdjEulerVariable.hpp
```diff
@@ -197,6 +175,7 @@ class CAdjEulerVariable : public CVariable {
    * \brief Get the reconstruction gradient for variables at all points.
    * \return Reference to reconstruction gradient.
    */
-  inline CVectorOfMatrix& GetGradient_Reconstruction(void) final { return Gradient_Reconstruction; }
+  inline CVectorOfMatrix& GetGradient_Reconstruction() final { return Gradient_Reconstruction; }
+  inline const CVectorOfMatrix& GetGradient_Reconstruction() const final { return Gradient_Reconstruction; }
```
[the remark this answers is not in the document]
pcarruscag: The compiler tries the non-const first, always, is "nodes" is const in that context the the const version is used and auto evaluates to "const CVectorOfMatrix". auto drops references (but not pointers) and so & needs to be added explicitly, otherwise you would get a local copy.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1396,6 +1231,12 @@ class CVariable {
   inline virtual bool SetPrimVar(unsigned long iPoint, su2double Density_Inf, su2double Viscosity_Inf,
                                  su2double eddy_visc, su2double turb_ke, CConfig *config) {return true; }
 
+  /*!
+   * \brief Get the primitive variables for all points.
+   * \return Reference to primitives.
+   */
+  inline virtual const MatrixType& GetPrimitive() const { AssertOverride(); return Solution; }
```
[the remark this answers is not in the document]
pcarruscag: It would force every derived to implement.

SU2_CFD/src/solvers/CSolver.cpp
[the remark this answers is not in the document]
pcarruscag: Making sure the "one definition rule" is not broken. like that issue with the interpolation from a while back. CSolver has enough methods as is, hence a free function in a helper namespace https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-pure

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: dereferencing a pointer, I'll change it to be consistent with the other changes.

Common/include/option_structure.hpp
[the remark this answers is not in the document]
pcarruscag: rain check.
These enums can be almost entirely removed by implementing the comms such that they operate on entire containers.
Then the solver just passes the container instead of the enum. Periodics have extra logic but some cases can also be removed.
Too tired to do that now, though

TobiKattmann:
@pcarruscag Wait, why did you merge this when the last commit made so many regression tests fail? I guess you did you not wait until they were finished? Not cool though and I hope this was not intentional
From what I've seen it is not super obvious how the tests are linked (especially the parallel_AD ones) but the commit is not that big so I only really suspect the removed enum entries of all ....

pcarruscag:
Apparently they fail if one merges early, develop is fine https://github.com/su2code/SU2/actions/runs/705683698

pcarruscag:
The merged branch is deleted and then the test action is not able to checkout
"error: pathspec 'refs/pull/1249/merge' did not match any file(s) known to git."
Because the regressions work by cloning su2 master and then doing "git checkout"
There was zero chance of breaking any testcases with those changes.

TobiKattmann:
... well ... let me quickly stop the air strike on your position :D
somehow it managed to exactly fail the streamwise periodic stuff (as they were added more recently then it makes sense)
I did not check check the release management 👍 so thanks for the info

> There was zero chance of breaking any testcases with those changes

I thought that as well, but I wouldn't be the first time I thought that and was proven otherwise ... :)

PR #1256, opened 2021-04-07
Fix memory leaks in CHeatSolver

maxaehle:
## Proposed Changes
The following members of `CHeatSolver` were allocated but never freed: `Primitive_Flow_i`, `Primitive_Flow_j`, `HeatFlux_per_Marker`, `AverageT_per_Marker`, `Surface_Areas`, `ConjugateVar`.

I propose to correct this in the spirit of #1225, using containers (`std::vector`) instead of pointers. There is a bit of additional cleanup.

## Related Work
#1225 

## PR Checklist

### DIFF
--- a/Common/include/containers/C2DContainer.hpp
+++ b/Common/include/containers/C2DContainer.hpp
@@ -624,39 +624,3 @@ using su2activematrix = su2matrix<su2double>;
 using su2passivevector = su2vector<passivedouble>;
 using su2passivematrix = su2matrix<passivedouble>;
 
-/*--- Functions to allocate 2D/3D data. ---*/
-
-namespace container_helpers {
-
-/*!
- * \brief Allocate a vector of varying-size vectors and initialize with some value.
- * \param[in] M - the first index is >=0 and < M
- * \param[in] N - the second index is >=0 and < N[first index]
- * \param[in,out] X - the vector of vectors
- * \param[in] val - the value for initialization, default is 0
- * \tparam T - type of an element
- */
-template<typename T>
-inline void Alloc2D(unsigned long M, const std::vector<unsigned long>& N, std::vector<std::vector<T> >& X, T val = 0) {
-  X.resize(M);
-  for(unsigned long i = 0; i < M; ++i){
-    X[i].resize(N[i], val);
-  }
-}
-
-/*!
- * \brief Allocate a vector of matrices with varying row count, and initialize with some value.
- * \param[in] M - the first index is >=0 and < M
- * \param[in] N - the second index is >=0 and < N[first index]
- * \param[in] P - the third index is >=0 and < P
- * \param[in,out] X - the vector of matrices
- * \param[in] val - the value for initialization, default is 0
- */
-inline void Alloc3D(unsigned long M, const std::vector<unsigned long>& N, unsigned long P, std::vector<su2activematrix>& X, su2double val=0) {
-  X.resize(M);
-  for(unsigned long i = 0; i < M; ++i){
-    X[i].resize(N[i],P) = su2double(val);
-  }
-}
-
-}
--- a/Common/include/containers/C2DContainer.hpp
+++ b/Common/include/containers/C2DContainer.hpp
@@ -33,6 +33,7 @@
 
 #include <utility>
 #include <type_traits>
+#include <vector>
 
 /*!
  * \enum StorageType
@@ -348,6 +349,7 @@ public:
 #undef MATRIX_ACCESSORS
 #undef VECTOR_ACCESSORS
 #undef UNIV_ACCESSORS
+
 }
 
 /*!
@@ -621,3 +623,40 @@ using su2activematrix = su2matrix<su2double>;
 
 using su2passivevector = su2vector<passivedouble>;
 using su2passivematrix = su2matrix<passivedouble>;
+
+/*--- Functions to allocate 2D/3D data. ---*/
+
+namespace container_helpers {
+
+/*!
+ * \brief Allocate a vector of varying-size vectors and initialize with some value.
+ * \param[in] M - the first index is >=0 and < M
+ * \param[in] N - the second index is >=0 and < N[first index]
+ * \param[in,out] X - the vector of vectors
+ * \param[in] val - the value for initialization, default is 0
+ * \tparam T - type of an element
+ */
+template<typename T>
+inline void Alloc2D(unsigned long M, const std::vector<unsigned long>& N, std::vector<std::vector<T> >& X, T val = 0) {
+  X.resize(M);
+  for(unsigned long i = 0; i < M; ++i){
+    X[i].resize(N[i], val);
+  }
+}
+
+/*!
+ * \brief Allocate a vector of matrices with varying row count, and initialize with some value.
+ * \param[in] M - the first index is >=0 and < M
+ * \param[in] N - the second index is >=0 and < N[first index]
+ * \param[in] P - the third index is >=0 and < P
+ * \param[in,out] X - the vector of matrices
+ * \param[in] val - the value for initialization, default is 0
+ */
+inline void Alloc3D(unsigned long M, const std::vector<unsigned long>& N, unsigned long P, std::vector<su2activematrix>& X, su2double val=0) {
+  X.resize(M);
+  for(unsigned long i = 0; i < M; ++i){
+    X[i].resize(N[i],P) = su2double(val);
+  }
+};
+
+}
--- a/Common/include/containers/C2DContainer.hpp
+++ b/Common/include/containers/C2DContainer.hpp
@@ -33,7 +33,6 @@
 
 #include <utility>
 #include <type_traits>
-#include <vector>
 
 /*!
  * \enum StorageType
@@ -51,7 +50,7 @@ enum SizeType : size_t {DynamicSize=0};
 
 /*--- Namespace to "hide" helper classes and
  functions used by the container class. ---*/
-namespace container_helpers
+namespace container_details
 {
 /*!
  * \class AccessorImpl
@@ -148,7 +147,9 @@ public:
   bool empty() const noexcept {return size()==0;}                       \
   Scalar_t* data() noexcept {return m_data;}                            \
   const Scalar_t* data() const noexcept {return m_data;}                \
+  Scalar_t* begin() noexcept {return data();}                           \
   const Scalar_t* begin() const noexcept {return data();}               \
+  Scalar_t* end() noexcept {return data()+size();}                      \
   const Scalar_t* end() const noexcept {return data()+size();}
 
   /*!
@@ -360,7 +361,6 @@ public:
 #undef MATRIX_ACCESSORS
 #undef VECTOR_ACCESSORS
 #undef UNIV_ACCESSORS
-
 }
 
 /*!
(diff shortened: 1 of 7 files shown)


SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -69,6 +65,27 @@ CHeatSolver::CHeatSolver(CGeometry *geometry, CConfig *config, unsigned short iM
...
+    X.resize(M);
+    for(unsigned long i = 0; i < M; ++i) X[i].resize(N[i], 0.0);
+  };
+
+  /*--- Allocates a 3D array with variable "middle" sizes and init to 0. ---*/
+
+  auto Alloc3D = [](unsigned long M, const vector<unsigned long>& N, unsigned long P, vector<su2activematrix>& X) {
+    X.resize(M);
+    for(unsigned long i = 0; i < M; ++i) X[i].resize(N[i],P) = su2double(0.0);
+  };
```
maxaehle, line 87: This is taken from the `CEulerSolver` constructor. Maybe `AllocND` could be defined in `CSolver.hpp` already?

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -667,7 +649,7 @@ void CHeatSolver::Set_Heatflux_Areas(CGeometry *geometry, CConfig *config) {
     }
   }
 
-  SU2_MPI::Allreduce(Local_Surface_Areas, Surface_Areas, config->GetnMarker_HeatFlux(), MPI_DOUBLE, MPI_SUM, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(Local_Surface_Areas, Surface_Areas.data(), config->GetnMarker_HeatFlux(), MPI_DOUBLE, MPI_SUM, SU2_MPI::GetComm());
```
maxaehle, line 652: Could this be problematic (MPI + containers)?

SU2_CFD/include/solvers/CHeatSolver.hpp
```diff
@@ -38,12 +38,23 @@
...
+  static constexpr size_t MAXNDIM = 3; /*!< \brief Max number of space dimensions, used in some static arrays. */
+  static constexpr size_t MAXNVAR = 1; /*!< \brief Max number of variables, for static arrays. */
+
   unsigned short nVarFlow, nMarker, CurrentMesh;
-  su2double **HeatFlux, *HeatFlux_per_Marker, *Surface_HF, Total_HeatFlux, AllBound_HeatFlux,
-            *AverageT_per_Marker, Total_AverageT, AllBound_AverageT,
-            *Primitive, *Primitive_Flow_i, *Primitive_Flow_j,
-            *Surface_Areas, Total_HeatFlux_Areas, Total_HeatFlux_Areas_Monitor;
-  su2double ***ConjugateVar, ***InterfaceVar;
+  vector<vector<su2double> > HeatFlux;
```
maxaehle, line 45: Maybe someone could add `/*!< \brief comments */`.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -989,7 +965,7 @@ void CHeatSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **solv
                || (config->GetKind_Solver() == DISC_ADJ_INC_NAVIER_STOKES)
                || (config->GetKind_Solver() == DISC_ADJ_INC_RANS));
 
-  su2double *Normal = new su2double[nDim];
+  su2double* Normal = nullptr;
```
TobiKattmann, line 968:
```suggestion
  su2double Normal[MAXNDIM];
```

SU2_CFD/src/solvers/CHeatSolver.cpp
TobiKattmann: not that I am aware of :)

SU2_CFD/include/solvers/CHeatSolver.hpp
TobiKattmann: Hmm yes I can do that, I already worked on the output of the heatsolver ... but I put that on the long bench. I am not sure if all those containers are necessary. 
I'll add a commit tomorrow 👍

SU2_CFD/src/solvers/CHeatSolver.cpp
TobiKattmann: If you can avoid code duplication, I am in for moving that to CSolver 👍

TobiKattmann (approved):
Thanks for this @maxaehle 💐 lgtm

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: Free function in container helpers or something

SU2_CFD/src/solvers/CHeatSolver.cpp
maxaehle: related change in commit 9f78e46

SU2_CFD/src/solvers/CHeatSolver.cpp
maxaehle: done in commit 3f65e8e

Common/include/containers/C2DContainer.hpp
```diff
@@ -621,3 +623,40 @@ using su2activematrix = su2matrix<su2double>;
...
+
+/*--- Functions to allocate 2D/3D data. ---*/
+
+namespace container_helpers {
+
+/*!
+ * \brief Allocate a vector of varying-size vectors and initialize with some value.
+ * \param[in] M - the first index is >=0 and < M
+ * \param[in] N - the second index is >=0 and < N[first index]
+ * \param[in,out] X - the vector of vectors
```
pcarruscag, line 635: Not on this file please, container_decorators.hpp (sorry I thought I had named it helpers).
Also the general convention we have for the few namespaces in the code is to use CamelCase.

Common/include/containers/C2DContainer.hpp
```diff
@@ -621,3 +623,40 @@ using su2activematrix = su2matrix<su2double>;
...
+ * \param[in] val - the value for initialization, default is 0
+ * \tparam T - type of an element
+ */
+template<typename T>
+inline void Alloc2D(unsigned long M, const std::vector<unsigned long>& N, std::vector<std::vector<T> >& X, T val = 0) {
```
pcarruscag, line 640: You could make the entire `std::vector<unsigned long>` also a template parameter, so that we can pass other types.
Potentially we retrieve "M" from the size of "N", same for Alloc3D

Common/include/containers/C2DContainer.hpp
```diff
@@ -621,3 +623,40 @@ using su2activematrix = su2matrix<su2double>;
...
+ * \param[in] M - the first index is >=0 and < M
+ * \param[in] N - the second index is >=0 and < N[first index]
+ * \param[in] P - the third index is >=0 and < P
+ * \param[in,out] X - the vector of matrices
+ * \param[in] val - the value for initialization, default is 0
+ */
+inline void Alloc3D(unsigned long M, const std::vector<unsigned long>& N, unsigned long P, std::vector<su2activematrix>& X, su2double val=0) {
+  X.resize(M);
+  for(unsigned long i = 0; i < M; ++i){
+    X[i].resize(N[i],P) = su2double(val);
```
pcarruscag, line 658: Same comments as before, and you can also template the type using `su2matrix<T>`.
You may need a separate template for "val" and then cast val to T for the assignment.
(otherwise you would need `decay_t<T>` for the type of of val to allow implicit conversion).

pcarruscag: Or instead of su2matrix<T> you can template the entire matrix (say as Matrix) to support also column major storage etc.
Then if you need the type you use `typename Matrix::Scalar`

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: There is no problem, memory is memory, but when doing this I prefer to use the `.size()` of the container for the MPI count rather than reading it from config or somewhere else.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -1002,7 +962,7 @@ void CHeatSolver::BC_ConjugateHeat_Interface(CGeometry *geometry, CSolver **solv
 
       if (geometry->nodes->GetDomain(iPoint)) {
 
-        Normal = geometry->vertex[val_marker][iVertex]->GetNormal();
+        su2double const* Normal = geometry->vertex[val_marker][iVertex]->GetNormal();
```
pcarruscag, line 965: This is probably more correct but in most places we have `const su2double*` or `const auto`.
Just because `su2double const*` and `su2double* const` are very different and may get confused...

Common/include/containers/C2DContainer.hpp
maxaehle: Maybe you remember correctly, there is a `namespace container_helpers` also in the top of `C2DContainer.hpp`.

pcarruscag: Damn... :facepalm: naming really is the hardest problem in programming.
The idea for that container_helpers namespace is to have "private" details (I should have just called it that...) used to define C2DContainer, because some elements of its implementation change when both sizes are dynamic, or when some or all sizes are known at compile time.
Maybe don't use a namespace for the Alloc functions, but give them longer names, like AllocVectorOfVectors.

maxaehle: `daad1d0`

Common/include/containers/C2DContainer.hpp
maxaehle: in f32f9ee

SU2_CFD/src/solvers/CHeatSolver.cpp
maxaehle: in f32f9ee

Common/include/containers/container_decorators.hpp
```diff
@@ -205,3 +205,60 @@ struct C3DDummyMiddleView
...
+  }
+}
+
+/*!
+ * \overload Derive outer size from index vector.
+ */
+template<class IndexVector, class VectorOfVector, class Scalar = int>
+inline void AllocVectorOfVectors(const IndexVector& N, VectorOfVector& X, Scalar val = 0) {
+  auto M = N.size();
+  AllocVectorOfVectors(M, N, X, val);
```
pcarruscag, line 235: Good idea with the overload :+1:

pcarruscag (approved):

Common/include/containers/container_decorators.hpp
```diff
@@ -205,3 +205,60 @@ struct C3DDummyMiddleView
...
+ * \param[in] val - the value for initialization, default is 0
+ * \tparam IndexVector - type of N
+ * \tparam VectorOfVector - type of X
+ */
+template<class IndexVector, class VectorOfVector, class Scalar = int>
+inline void AllocVectorOfVectors(size_t M, const IndexVector& N, VectorOfVector& X, Scalar val = 0) {
+  X.resize(M);
+  for(size_t i = 0; i < M; ++i){
+    X[i].resize(N[i]);
+    for (auto& x : X[i]) x = val;
```
pcarruscag, line 225: I changed it a bit to try to make it compatible with more types. If I broke anything I'll fix it.
Thanks for all the cleanup.

PR #1258, opened 2021-04-07
Part 2: Make enum -> enum class

TobiKattmann:
Hi all, 

this is the continuation of #1242 . Change of the `enum`'s in `option_structure.hpp` into `enum class`. This has the benefit of no implicit conversion to int and therefore increased safety with the enum type -> [c++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#enum3-prefer-class-enums-over-plain-enums).

If you want to sponsor a commit (which I highly recommend ;) ) then a good approach is to 
1. make your `enum` of choice an `enum class` and remove the associated numbers. 
2. Rename here if reasonable, i,.e. `ENUM_INLET_TYPE` can be just `INLET_TYPE`. Also the option names to not be redundant so e.g. `INLET_TYPE::PRESSURE_INLET_TYPE` could become  `INLET_TYPE::PRESSURE`
3. Add a `NONE` or `NO_THING` option if needed. Often the global `NONE` option is and that made a lot of people very angry and has been widely regarded as a bad move 🧙 
2. then you go to `CConfig.c/hpp` and change the option type and everything there (because in the `CConfig` class the variable is usually defined). Find the `GetKind_yourOption` and search the code base for that `Get` function.
3. Once all search results and the following function bodies are changed just compile and fix until the code builds
4. Special care for switch statements: If not all options are present make a `default; break;` line -> otherwise the automated builds will fail on a warning (clang is used, I found gcc does only complain with an additional flag)
4. commit at best just 1 commit with a clear message and push. That makes reverting easier if it fails and introduced some non-obvious error
 
Have fun :)

## Related Work
#1242

## PR Checklist


Common/include/option_structure.hpp
```diff
@@ -2070,11 +2070,12 @@ static const MapType<string, ENUM_DIRECTDIFF_VAR> DirectDiff_Var_Map = {
 };
 
 
-enum ENUM_RECORDING {
-  SOLUTION_VARIABLES = 1,
-  MESH_COORDS = 2,
-  MESH_DEFORM = 3,
-  SOLUTION_AND_MESH = 4
+enum class RECORDING {
+  NO_RECORDING,
```
TobiKattmann, line 2074: Added a `NO_RECORDING` option as the global `NONE` was used

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -235,9 +235,9 @@ void CDiscAdjMultizoneDriver::Run() {
 
   /*--- Evaluate the objective function gradient w.r.t. the solutions of all zones. ---*/
 
-  SetRecording(NONE, Kind_Tape::OBJECTIVE_FUNCTION_TAPE, ZONE_0);
-  SetRecording(SOLUTION_VARIABLES, Kind_Tape::OBJECTIVE_FUNCTION_TAPE, ZONE_0);
-  RecordingState = NONE;
+  SetRecording(RECORDING::NO_RECORDING, Kind_Tape::OBJECTIVE_FUNCTION_TAPE, ZONE_0);
```
TobiKattmann, line 238: see here

pcarruscag: We can try to make the "NONE" case either expressive or concise, for example the NONE of time marching is steady.
The NONE recording serves to clear indices.

pcarruscag: We can try to make the "NONE" case either expressive or concise, for example the NONE of time marching is steady.
The NONE recording serves to clear indices.

TobiKattmann: Exactly. For TIME_MARCHING that is done 👍 
Only add a NONE-type option if it not already expressed by another option by an elimination process

TobiKattmann: oh and 👍 for the addition :)

pcarruscag: What I'm say is RECORDING::NO_RECORDING is not a very good name IMO

TobiKattmann: I'll change it to RECORDING::CLEAR_INDICES (tomorrow). That should be better?
(Sorry I read the comment with a vanished attention span 😴 :) )

pcarruscag: I think so, gives more info at least, but hey... I'm sleeping on the job too.

Common/include/option_structure.hpp
```diff
@@ -469,20 +469,6 @@ const int EL_PYRAM = 2;   /*!< \brief Elements of five nodes (3D). */
 const int EL_PRISM = 3;   /*!< \brief Elements of six nodes (3D). */
 
 
-/*!
- * \brief Types of mathematical problem to solve
- */
-enum ENUM_MATH_PROBLEM {
```
TobiKattmann, line 503: not used any more, nice 👍

pcarruscag (approved):

TobiKattmann:
Codefactor is just complainig about the usual CIncEulerSolver. All other test pass, so I'll merge this and continue in Part III. 
Coming to a repository near you soon. 🎦

PR #1259, opened 2021-04-10
Fixes in history output for time averaged and multizone problems

cvencro:
## Proposed Changes
Improvements to time averaged history output for dynamic multizone problems, which now also include the structural zone. The averaging logic has been updated to include values only for the final solution in each time iteration.

To test the fixes in this PR, time averaged history output has been added to the existing dynamic fsi test case.

## PR Checklist


pcarruscag:
You could modify `CWindowedAverage::addValue` to only "push back" more values if it detects a change in current time iteration, otherwise it simple overwrites the last value in the history.
Then you could get rid of the entire logic in `SetUpdate_Averages` simply making it true or false (less logic is the way for less bugs).

SU2_CFD/src/output/COutput.cpp
```diff
@@ -248,6 +248,25 @@ void COutput::SetHistory_Output(CGeometry *geometry,
...
+
+    /*--- Write the screen header---------------------------------------------------------------------------*/
+    write_header = WriteScreen_Header(config);
+    if (write_header) SetScreen_Header(config);
+
+    /*--- Write the screen output---------------------------------------------------------------------------*/
+    write_screen = WriteScreen_Output(config);
+    if (write_screen) SetScreen_Output(config);
+
+  }
```
pcarruscag, line 269: This block of code is showing up quite a bit, make it a private function (like Postprocess_HistoryData) please.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -217,7 +217,7 @@ void COutput::SetHistory_Output(CGeometry *geometry,
 
   /*--- Output using only the master node ---*/
 
-  if (rank == MASTER_NODE && !noWriting) {
+  if (rank == MASTER_NODE && !noWriting && !config->GetMultizone_Problem()) {
```
pcarruscag, line 220: Won't this stop inner iterations from being written to history/screen?

cvencro:
Hi Pedro, thanks for the idea to update the windowing directly! I've updated addValue such that the values are added for new time only (replaces existing values if it is still the same time iteration). This is a lot simpler and very happy to remove the convoluted logic. The SetUpdate_Averages was still necessary though to pass the regression test for unsteady_cylinder_windowed_average.

SU2_CFD/src/output/COutput.cpp
cvencro: Yes, that was true. The first inner iteration for a new time iteration still updated the averaged value using the previous logic attempt if this was called. This is irrelevant now with the update to windowing directly, so I've reverted this.

SU2_CFD/src/output/COutput.cpp
cvencro: Yes agreed, I've added a function, OutputScreenAndHistory, which takes care of these.

SU2_CFD/src/output/tools/CWindowingTools.cpp
```diff
@@ -74,7 +74,20 @@ void CWindowedAverage::Reset(){
...
+      if (winTimeIter[winTimeIter.size()-1] == curTimeIter) {
+        values.pop_back();
+        winTimeIter.pop_back();
+      }
+    }
+
+    /*--- Update time iteration and value ---*/
+    winTimeIter.push_back(curTimeIter);
+    values.push_back(valIn);
+  }
```
pcarruscag, line 90:
```suggestion
  if (curTimeIter < startIter) return;
  curTimeIter -= startIter;
  if (curTimeIter == values.size()) values.push_back(valIn);
  else {
    assert(curTimeIter == values.size()-1 && "Cannot go back or advance more than one timestep.");
    values.back() = valIn;
  }
```
This should do without storing the time iterations.

SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -726,7 +726,7 @@ bool CFlowCompOutput::SetInit_Residuals(CConfig *config){
 
 bool CFlowCompOutput::SetUpdate_Averages(CConfig *config){
 
-  return true;
+  return (config->GetTime_Marching() != TIME_MARCHING::STEADY && (curInnerIter == config->GetnInner_Iter() - 1 || convergence));
```
pcarruscag: It is strange that would break the testcase, I'll have a look.

pcarruscag: Your initial change was correct, the logic was artificially delaying the update of the average to the end of the timestep, and the timedomain regressions check the first inner iteration by default.

pcarruscag (approved):

cvencro:
Hi Pedro, thanks for updating the regression test and all the tidy up as well.

TobiKattmann:
Hi @cvencro, thanks for those changes 💐 I added very small changes to another PR

PR #1264, opened 2021-04-16
Fix inlet profile file loading when not restarting unsteady problems

pcarruscag:
## Proposed Changes
When not restarting unsteady problems, set file iteration to 0 (instead of the -1/-2 logic).

## Related Work
https://www.cfd-online.com/Forums/su2/235485-parabolic-velocity-inlet-profile-error.html

## PR Checklist


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3316,29 +3316,23 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
   unsigned long Marker_Counter = 0;
 
-  bool turbulent = (config->GetKind_Solver() == RANS ||
-                    config->GetKind_Solver() == INC_RANS ||
-                    config->GetKind_Solver() == ADJ_RANS ||
-                    config->GetKind_Solver() == DISC_ADJ_RANS ||
-                    config->GetKind_Solver() == DISC_ADJ_INC_RANS);
-
   unsigned short nVar_Turb = 0;
-  if (turbulent)
```
bigfooted, line 3330: yes! Sanity checks should be (and are in this case) done in cconfig, so we can safely assume that the solver is a turbulence solver.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3316,29 +3316,23 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+    case NONE:
+      break;
+    case SA: case SA_NEG: case SA_E: case SA_COMP: case SA_E_COMP:
+      nVar_Turb = 1;
+      break;
+    case SST: case SST_SUST:
+      nVar_Turb = 2;
+      break;
+    default:
+      SU2_MPI::Error("Specified turbulence model unavailable or none selected", CURRENT_FUNCTION);
```
bigfooted, line 3333: "none selected" is covered by the switch.

pcarruscag: Not sure what you mean, we still need to cover the case when there is no turbulence solver.
I cleaned it up more to avoid the switch.

bigfooted (approved):
Looks fine to me. Also like the introduction of the c++11 range based for loop. We should use that more often.

PR #1266, opened 2021-04-17
Multi grid output

pcarruscag:
## Proposed Changes
New output to visualize the coarse grids.
It is based on coloring nodes, which is a little difficult to visualize with Paraview since it interpolates to elements, anyway:
image

## PR Checklist


pcarruscag:
Last week folks mentioned, during the meeting, that this would be helpful, so I'll take it as approved.

Common/src/CConfig.cpp
```diff
@@ -4996,9 +5004,9 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
   if (GetGasModel() == "ARGON") monoatomic = true;
 
   // This option is deprecated. After a grace period until 7.2.0 the usage warning should become an error.
-  if(OptionIsSet("CONV_CRITERIA")) {
-    cout << string("\n\nWARNING: CONV_CRITERIA is deprecated. SU2 will choose the criteria automatically based on the CONV_FIELD.\n") +
-            string("RESIDUAL for any RMS_* BGS_* value. CAUCHY for coefficients like DRAG etc.\n\n");
+  if(OptionIsSet("CONV_CRITERIA") && rank == MASTER_NODE) {
```
TobiKattmann, line 5007: makes it less annoying, sure, but that was the purpose to some degree :D (totally did that on purpose ... 🤥 )

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -2735,6 +2735,41 @@ bool CFlowOutput::WriteVolume_Output(CConfig *config, unsigned long Iter, bool f
   return false || force_writing;
 }
 
+void CFlowOutput::AddCommonFVMOutputs(const CConfig *config) {
```
TobiKattmann, line 2738: 👍 Special reason to not move `COORD-XYZ`  here?

TobiKattmann (commented):
💐 Thanks, also for the little stuff around

SU2_CFD/src/output/CFlowOutput.cpp
pcarruscag: Order matters I think.

Common/src/CConfig.cpp
pcarruscag: Yeah you can't miss the message if you run 1000 cores

SU2_CFD/src/output/CFlowOutput.cpp
TobiKattmann: yep it does ... it needs to be the first, at least for paraview output

Issue #1265, opened 2021-04-17
inlet profile from file is not initialized from config.cfg


bigfooted:
When SU2 creates an inlet profile file, the values in the file are set to 0.0, expect for the vertex coordinates. The values used to be initialized from MARKER_INLET in the config file. Any particular reason this initialization disappeared? I propose to add this again, and also to add a line to inlet.dat with the names of the variables in the columns so it's easier for users to see what's in the columns.

bigfooted:
implementation result:

```
NMARK= 2
MARKER_TAG= inlet_air
NROW=32
NCOL=8
# COORD-X               COORD-Y                 TEMPERATURE             VELOCITY                NORMAL-X                NORMAL-Y                TKE                     DISSIPATION
-2.500000000000000e-01  5.000000000000000e-02   3.000000000000000e+02   5.000000000000000e+00   1.000000000000000e+00   0.000000000000000e+00   1.500000000000000e-02   1.000000000000000e+02
-2.500000000000000e-01  4.838709677419802e-02   3.000000000000000e+02   5.000000000000000e+00   1.000000000000000e+00   0.000000000000000e+00   1.500000000000000e-02   1.000000000000000e+02


```

pcarruscag:
Sounds good so long as backwards compatibility is kept and the names for the columns and default values are passed into CMarkerProfileReaderFVM in a way that does not require it to do anything problem-specific.

bigfooted:
yes, it works with and without the additional column information.

bigfooted:
closed by #1268

Issue #1263, opened 2021-04-15
wrong turbulence intensity for transition models


bigfooted:
**Describe the bug**
The (freestream) turbulence intensity is defined as u'/U_mean, and it is used like that to define e.g. freestream tke at far field boundaries. However, transition models use the percentage value of turbulent intensity, so 10 x u'/U_mean. This factor of 100 is missing in SU2, e.g. in transitionLM or the SST transition model. This means that you cannot consistently use a transition model (input turbulence intensity in %) together with an inlet boundary for instance (turbulence intensity at inlet should be a fraction, so 0.01*percentage-value) Actually, in the description of the LM model, the factor of 100 is 
explicitly mentioned when computing intensity from tke: 

https://turbmodels.larc.nasa.gov/langtrymenter_4eqn.html

I think the best is to keep the standard definition that the input turbulence intensity is a fractional value, e.g. 0.05 or 0.10, and that we convert to the percentage value (in this example 5 or 10) in the transition models.

see also the discussion on this issue on:
https://www.cfd-online.com/Forums/su2/235365-free-stream-turbulence-intensity-option.html

bigfooted:
closed by #1268

PR #1267, opened 2021-04-18
Fix turbulence input

bigfooted:
## Proposed Changes

1. fixed factor of 100 in the transition turbulence models
2. changed initialization of inlet profile from 0.0 to values taken from config file. I guess this was done to make this routine more general (for other boundaries) in the future. However, I think generating a default file that works out of the box is better. 
3. added a descriptive comment line to the inlet profile file (one for each marker) with the names of the columns. This line is intended to make editing easier for the user. This comment line is not mandatory and is ignored by the profile reader. 
3. do not write residuals of temperature to output file when energy equation is off

## Related Work
Related to issue #1265 and #1263

## PR Checklist


sametcaka:
Hello,
Sorry for a very very late comment, but if I did not get it wrong, it seems like the turbulence intensity value specified in the config file is now multiplied by 100 in the turb_sources.cpp at line 143 after this commit. 

Generally, the users tend to use the percent value (because nearly every paper on the subject talk in percentages) for turbulence intensitirs, so for example, if the experimental conditions say 3% for the turbulence intensity, the option in the config file should be 3.0. But now with this 100* modification, the users must enter 0.03, is that correct? 

If it is correct, then transition test case/tutorial config files should be updated as well. Because the transition onset correlations (Menter or Abu-Ghannam) uses the percent value. 

All in all, however, I suggest that we should remove the factor of 100, so the value in the config file is in percentages directly, as it was before. 

Regards,
Samet

bigfooted:
Thank you for the careful scrutiny. The input in su2 is turbulence intensity, which is defined as u'/U. Instead of changing the 40 turbulence testcases that use turbulence intensity, I suggest we change the one testcase that uses the turbulence intensity percentage 100*u'/U (the T3A flat plate).

For what it's worth: openfoam uses turbulence intensity u'/U and ANSYS Fluent uses 100*u'/U, so both notations are used.

PR #1268, opened 2021-04-18
Fix dissipation in transition model and update inlet profile (initial profile from config)

bigfooted:
## Proposed Changes

 

1. fixed factor of 100 in the transition turbulence models
2. changed initialization of inlet profile from 0.0 to values taken from config file. I guess this was done to make this routine more general (for other boundaries) in the future. However, I think generating a default file that works out of the box is better.
3. added a descriptive comment line to the inlet profile file (one for each marker) with the names of the columns. This line is intended to make editing easier for the user. This comment line is not mandatory and is ignored by the profile reader.
4. do not write residuals of temperature to output file when energy equation is off

## Related Work
Related to #1263 and #1265 

## PR Checklist


SU2_CFD/include/CMarkerProfileReaderFVM.hpp
```diff
@@ -111,6 +114,7 @@ class CMarkerProfileReaderFVM {
    */
   CMarkerProfileReaderFVM(CGeometry      *val_geometry,
                           CConfig        *val_config,
+                          CSolver        **val_solver,
                           string         val_filename,
```
pcarruscag, line 118: I'm very sorry but this is not the way. Having generic code that is orthogonal to each other is a good thing, and we need to move in that direction, always.

Prepare the column names and default values outside of this class, in the method that uses it for example.
Then give this class a method to pass in that information, when it is not passed it can have default values like "Var1" "Var2", etc.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
```diff
@@ -428,13 +459,68 @@ void CMarkerProfileReaderFVM::WriteMarkerProfileTemplate() {
...
+      switch (numberOfTurbVars) {
+        case 0:
+          /*--- no turbulence model---*/
+          break;
+        case 1:
+          /*--- 1-equation turbulence model: SA ---*/
+          turb_val[0] = config->GetNuFactor_FreeStream()*config->GetViscosity_FreeStreamND()/config->GetDensity_FreeStreamND();
+          node_file << "NU_TILDE   " << setw(24);
+          break;
+        case 2:
```
pcarruscag, line 513: This type of assumption makes it hard to extend the code, always prefer text (i.e. enums) to numbers.

SU2_CFD/src/numerics/turbulent/turb_sources.cpp
```diff
@@ -139,7 +139,8 @@ CNumerics::ResidualType<> CSourcePieceWise_TurbSA::ComputeResidual(const CConfig
       const su2double chi_1 = 0.002;
       const su2double chi_2 = 50.0;
 
-      su2double tu = config->GetTurbulenceIntensity_FreeStream();
+      /*--- turbulence intensity is u'/U so we multiply by 100 to get percentage ---*/
+      su2double tu = 100.0 * config->GetTurbulenceIntensity_FreeStream();
```
pcarruscag, line 143: :+1: every now and then CFD-online helps us find some bugs :)

pcarruscag (changes requested):

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
bigfooted: I would have preferred that we have the number of solvers available somewhere, and we can then take the numvars of each solver and do a summation. I now need numvar_turb, later also numvar_scalar, probably others later as well (radiation?). 
Do we have the number of solvers available?

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
bigfooted: so maybe introduce ONE_EQUATION_MODEL, TWO_EQUATION_MODEL? Because I do not like having SA_model1 OR SA_model2 OR SA_model256 everywhere when the only thing we want to know is the number of turbulence variables.

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
bigfooted: I removed val_solver

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
bigfooted: I made an update, but now we have case SA:case SA_NEG etc... here.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
```diff
@@ -428,13 +458,72 @@ void CMarkerProfileReaderFVM::WriteMarkerProfileTemplate() {
...
+      else 
+        node_file << "# COORD-X  " << setw(24) << "COORD-Y    " << setw(24) << "COORD-Z    " << setw(24); 
+
+      INLET_TYPE Kind_Inlet = config->GetKind_Inc_Inlet(Marker_Tag);
+
+      if (compressible) {
+        node_file << "TEMPERATURE" << setw(24) << "PRESSURE   " << setw(24); 
+      }
+      else {
+        switch (Kind_Inlet) {
```
pcarruscag, line 484: What I meant is that this class should not have anything specific to any type of physics.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
pcarruscag: I do not like either -> #770 
It is for that reason we have methods in CConfig (and some in option_structure) like GetSolid_Wall, GetViscous_Wall, GetFluidProblem, GetStructuralProblem, etc.
Like I've said in that issue that no one cared about, repeating that kind of switch statement is one of the main causes of bugs we have in SU2.

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
pcarruscag: The solver container has a maximum number of solvers and there are fixed positions for different types of solver.
You could, for example, add a method to solvers to be able to ask the solver "do you like inlet profiles? what are the name of the variables you want from it? what default values should they have?"
Then you can loop over all possible solvers and ask them that (if they exist i.e. not nullptr), that gets passed down to the profile reader (as lists of strings, whose content is irrelevant for a class that just reads a table).

SU2_CFD/include/CMarkerProfileReaderFVM.hpp
```diff
@@ -46,6 +46,7 @@
 #include "../../Common/include/parallelization/mpi_structure.hpp"
 #include "../../Common/include/CConfig.hpp"
 #include "../../Common/include/geometry/CGeometry.hpp"
+#include "../../SU2_CFD/include/solvers/CSolver.hpp"
```
pcarruscag, line 49:
```suggestion
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3320,25 +3320,7 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
 
   unsigned long Marker_Counter = 0;
 
-  bool turbulent = (config->GetKind_Solver() == RANS ||
```
pcarruscag, line 3323: We are going to end up with changes to the same functions here and in #1264 maybe you could review that PR since you are also fixing inlet profile issues.

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
bigfooted: We could construct the entire comment line and a vector (of vectors) of default values outside of the function (so in loadInletProfile in csolver) and pass that to the reader, is that what you mean?

bigfooted: I moved the physics from profilereader to csolver.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3327,9 +3335,87 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+
+    columnValue << t_total << "\t" << p_total <<"\t";
+    for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+      columnValue << flow_dir[iDim] <<"\t";
+    }
+
+    columnName << "# COORD-X  " << setw(24) << "COORD-Y    " << setw(24);
+    if(nDim==3) columnName << "COORD-Z    " << setw(24); 
+    
+    if (compressible){
```
pcarruscag, line 3362: This is ok for now, but if more solvers start using inlet profiles this is not a scalable solution.
The suggestion was to have this block of code as a method of different solvers, so that you do not have to figure out the type of solver based on configuration options (and consequently the type of inlet, etc. etc.).
```solver->GetInletVariableNames()``` something like that.
The motivation for polymorphism and abstract classes is to allow centralizing this type of knowledge, i.e. "what type of object is this?", on a single place, the instantiation of the object.
Enough philosophy...

**Review by pcarruscag — APPROVED** [review]

**Comment by pcarruscag** [comment]
By the way please give the PR a more descriptive title that reflects the transition fix and the column names feature.
So that folks know what to expect from the change log.

File: SU2_CFD/src/solvers/CSolver.cpp (hunk 7 above)
Comment (bigfooted) [reply SU2_CFD/src/solvers/CSolver.cpp:3362]: yes, I prefer having something like this at the level of the solver container so we can construct this information. After all, very soon we will have passive scalar, combustion variables, nemo combustion variables,...

File: SU2_CFD/src/CMarkerProfileReaderFVM.cpp [hunk 8]
```diff
@@ -45,6 +47,8 @@ CMarkerProfileReaderFVM::CMarkerProfileReaderFVM(CGeometry      *val_geometry,
   filename     = val_filename;
   markerType   = val_kind_marker;
   numberOfVars = val_number_vars;
+  columnNames  = val_columnNames;
+  columnValues = val_columnValues;
```
TobiKattmann, line 51: I guess this could be done with an initializer list for the ctor .. but I am unsure if one is to be preferred over the other. The core guidelines dont seem to have a strong opinion either. [Payground](https://gcc.godbolt.org/z/eraGd1qEz)

SU2_CFD/src/CMarkerProfileReaderFVM.cpp
```diff
@@ -95,24 +105,40 @@ void CMarkerProfileReaderFVM::ReadMarkerProfile() {
...
+        /*--- Skip the data. This is read in the next loop. ---*/
+        
+        for (unsigned long iRow = 0; iRow < (numberOfRowsInProfile[iMarker]-skip); iRow++) getline (profile_file, text_line);
 
       }
     } else {
-      SU2_MPI::Error("While opening profile file, no \"NMARK=\" specification was found", CURRENT_FUNCTION);
+      //cout << "inlet profile reader is ignoring line: " << text_line << endl;
     }
   }
```
TobiKattmann, line 134: in that case I would remove the whole `else ` statement ... unless there are further plans

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -593,9 +594,11 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
   SetVolumeOutputValue("RES_VELOCITY-Y", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 2));
   if (nDim == 3){
     SetVolumeOutputValue("RES_VELOCITY-Z", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 3));
-    SetVolumeOutputValue("RES_TEMPERATURE", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 4));
+    if (config->GetEnergy_Equation())  
+      SetVolumeOutputValue("RES_TEMPERATURE", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 4));
   } else {
-    SetVolumeOutputValue("RES_TEMPERATURE", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 3));
+    if (config->GetEnergy_Equation())  
+      SetVolumeOutputValue("RES_TEMPERATURE", iPoint, solver[FLOW_SOL]->LinSysRes(iPoint, 3));
```
TobiKattmann, line 601: I guess one could move this `config->GetEnergy_Equation()` out of the `if nDim` and make the acces `solver[FLOW_SOL]->LinSysRes(iPoint, nDim+1)`

TobiKattmann (commented):
A bit late, but thanks for the transition-model bugfix and the inlet-marker-stuff 💐

PR #1269, opened 2021-04-23
"Did you mean...?" (make config errors more helpful)

pcarruscag:
## Proposed Changes
suggest an alternative for unrecognized options
list the enum alternatives when an invalid one is used

### DIFF
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -2889,6 +2889,8 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
             newString.append(string("CONV_CRITERIA is deprecated. SU2 will choose the criteria automatically based on the CONV_FIELD.\n") +
                              string("RESIDUAL for any RMS_* BGS_* value. CAUCHY for coefficients like DRAG etc.\n\n"));*/
           else {
+            /*--- Find the most likely candidate for the unrecognized option, based on the length
+             of start and end character sequences shared by candidates and the option. ---*/
             auto countMatchChars = [&option_name](const string& candidate) {
               const size_t sz1 = option_name.size(), sz2 = candidate.size();
               size_t nMatch = 0;
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -2807,8 +2807,8 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
 
   string errorString;
 
+  const int max_err_count = 30; // Maximum number of errors to print before stopping
   int err_count = 0;  // How many errors have we found in the config file
-  int max_err_count = 30; // Maximum number of errors to print before stopping
   int line_count = 1;
 
   map<string, bool> included_options;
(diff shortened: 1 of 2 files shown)


Common/src/CConfig.cpp
```diff
@@ -2862,34 +2860,61 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
           if (!option_name.compare("RELAXATION_FACTOR_ADJFLOW"))
             newString.append("Option RELAXATION_FACTOR_ADJFLOW is now RELAXATION_FACTOR_ADJOINT, "
                              "and it also applies to discrete adjoint problems.\n\n");
-          if (!option_name.compare("WRT_MESH_QUALITY"))
+          else if (!option_name.compare("WRT_MESH_QUALITY"))
```
WallyMaier, line 2863: Wouldn't this prevent multiple warnings from being printed? Not sure if that's the point (or Im misunderstanding).

pcarruscag: No, those checks for deprecated options were already mutually exclusive.
You get multiple messages from the outer loop over all options in the config, and so you will also get multiple "did you mean?"s if you misspell more than one option.

WallyMaier: Ahh I missed the outer loop. Thanks.

WallyMaier (approved):
This looks good to me!  Thanks for updating, should be way more user friendly now.

Common/src/CConfig.cpp
```diff
@@ -2862,34 +2860,61 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
...
             newString.append("WRT_SOL_FREQ is deprecated. Use OUTPUT_WRT_FREQ instead.\n\n");
-          if (!option_name.compare("WRT_SOL_FREQ_DUALTIME"))
+          else if (!option_name.compare("WRT_SOL_FREQ_DUALTIME"))
             newString.append("WRT_SOL_FREQ_DUALTIME is deprecated. Use OUTPUT_WRT_FREQ instead.\n\n");
           // This option is deprecated. After a grace period until 7.2.0 the usage warning should become an error.
-          /*if (!option_name.compare("CONV_CRITERIA"))
+          /*else if (!option_name.compare("CONV_CRITERIA"))
             newString.append(string("CONV_CRITERIA is deprecated. SU2 will choose the criteria automatically based on the CONV_FIELD.\n") +
                              string("RESIDUAL for any RMS_* BGS_* value. CAUCHY for coefficients like DRAG etc.\n\n"));*/
+          else {
```
pcarruscag, line 2891:
```suggestion
          else {
            /*--- Find the most likely candidate for the unrecognized option, based on the length
             of start and end character sequences shared by candidates and the option. ---*/
```

Common/src/CConfig.cpp
```diff
@@ -4401,11 +4425,11 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
   if (DirectDiff != NO_DERIVATIVE) {
 #ifndef CODI_FORWARD_TYPE
-      if (Kind_SU2 == SU2_COMPONENT::SU2_CFD) {
-        SU2_MPI::Error(string("SU2_CFD: Config option DIRECT_DIFF= YES requires AD or complex support!\n") +
-                       string("Please use SU2_CFD_DIRECTDIFF (configuration/compilation is done using the preconfigure.py script)."),
-                       CURRENT_FUNCTION);
-      }
+    if (Kind_SU2 == SU2_COMPONENT::SU2_CFD) {
+      SU2_MPI::Error(string("SU2_CFD: Config option DIRECT_DIFF= YES requires AD or complex support!\n") +
+                     string("Please use SU2_CFD_DIRECTDIFF (configuration/compilation is done using the preconfigure.py script)."),
```
TobiKattmann, line 4430: please make this `meson.py` in #1272

PR #1271, opened 2021-04-27
Fixing version numbers 

WallyMaier:
## Proposed Changes
This PR fixes the version numbers of files and test cases up to the 7.1.1. (Mostly in NEMO)   All files should be consistent now, but I may have missed some. 

I am not sure if the test cases were intentionally left old, but I can change that.

EDIT:
After taking a look through, I used my best judgement regarding what is up-to-date.  The biggest issue I had was in the sliding_interface sections where configs for Kind_Interpolation= Weight-Average and Nearest-Neighbor.  In this case I assumed that these options or used interchangeably among test cases, that I can update both _NN and _WA.   I did similar things for some SST vs SA options.   Below are the only cases that Im not sure if they are being used/up-to-date with develop: 

fea_fsi/SquareCyl_Beam/
incomp_rans/rough_wall/   [pretty sure this is up-to-date, but not used in regressions]
optimization_euler/steady_inverse_design/
sliding_interface/incompressible_unsteady/  [Commented out, so I left this one alone]
sliding_interface/single_stage/
turbomachinery/centrifugal_*/   

I also went ahead and removed:
rans/naca0012/*_sa_neg   [SA_neg is already used in turb_NACA0012_sa.cfg] ?
rans/naca0012/*_sa_binary

## Related Work
This is a chore.

## PR Checklist


WallyMaier:
It looks likes my attempts to align the test cases were a failure...tabs :(    

I can go through and update those to make it look nice, if everything else looks good.

pcarruscag (commented):
Are all of these part of the regressions? i.e. do they really run with 7.1.1? Otherwise it would be important to keep the reference with which they are known to run.

WallyMaier:
Sounds good.  I guess this begs the questions of should we be carrying test cases that dont run with the most recent version of the code?

WallyMaier:
I made some comments above regarding the cases.

pcarruscag (approved):
I guess the answer to that is it depends... ideally all of them would work.

TobiKattmann (approved):
>  Otherwise it would be important to keep the reference with which they are known to run.

that would obviously nice ... but during the config option cleanups, options were also removed in cfg's that might just run with older code, i.e. even with the correct version i might fail out of the box. 
Thanks @WallyMaier for this 💐

PR #1277, opened 2021-05-01
Interpolate restart file when it does not match the mesh

pcarruscag:
Method described below.
No config options needed.
Takes about a minute to interpolate an 8M node mesh to a 70M mesh on 1000 cores.
May or may not add OpenMP to this which would improve scalability.
Restarts are read twice for RANS problems, which is not great for this type of feature...


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3094,6 +3114,181 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
...
+
+  /* Challenges:
+   *  - Do not use too much memory by gathering the restart data in all ranks.
+   *  - Do not repeat too many computations in all ranks.
+   * Solution?:
+   *  - Build a local ADT for the domain points (not the restart points).
+   *  - Find the closest target point for each donor, which does not match all targets.
+   *  - "Diffuse" the data to neighbor points.
+   *  Complexity is approx. Nlt + (Nlt + Nd) log(Nlt) where Nlt is the LOCAL number
+   *  of target points, Nd the TOTAL number of donors. */
```
pcarruscag, line 3138: method

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2882,16 +2884,18 @@ void CSolver::Read_SU2_Restart_ASCII(CGeometry *geometry, const CConfig *config,
     }
   }
 
+  if (iPoint_Global != geometry->GetGlobal_nPointDomain())
+    SU2_MPI::Error("The solution file does not match the mesh.",CURRENT_FUNCTION);
```
pcarruscag, line 2888: does not work for ASCII

bigfooted:
> Method described below.
> No config options needed.
> Takes about a minute to interpolate an 8M node mesh to a 70M mesh on 1000 cores.
> May or may not add OpenMP to this which would improve scalability.
> Restarts are read twice for RANS problems, which is not great for this type of feature...

This is a great addition, I am now doing this outside of SU2 with some python code, but having this native in SU2 is really great! I haven't looked at the details yet, but I guess the assumption is that the geometry stay the same, we only have a finer mesh? Also, the 'restarts are read twice for RANS' is that a general issue when having additional solvers (like heat, scalar, nemo)?
Last thing (for now): Does it work when using an inlet.dat boundary file?

SU2_CFD/src/solvers/CSolver.cpp
bigfooted: will you support it for this PR, or this is for later?

pcarruscag: It is a bit messy to do it, besides, if someone is restarting from ASCII the problem is probably not very large.
(every mpi rank needs to read the entire ASCII file because lines are defined by the \n character)

pcarruscag:
There is no assumption, but if the geometry is very different you'll end up with free-stream values in the boundary layer and vice versa.
The restarts are read once per solver, so, if you have Incomp RANS, with weakly coupled heat, a scalar, and mesh deformation... 5 times :)
I expect it to work with everything (to the extent that a matching restart does). Except maybe for problems that update the grid coordinates based on the contents of the restart.

bigfooted:
> The restarts are read once per solver, so, if you have Incomp RANS, with weakly coupled heat, a scalar, and mesh deformation... 5 times :)
Well, it seems fast enough, so it does not pose an immediate problem. I guess the solution to this is to move the reading of the data 'higher up', first get all the solvers and variables, then read the file and populate the solver variables, something like that? 
Anyway, I think the way it is done now looks good.

bigfooted (approved):
looks good to me.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2925,17 +2929,18 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
                    string("possible with the READ_BINARY_RESTART option."), CURRENT_FUNCTION);
   }
 
-  /*--- Store the number of fields to be read for clarity. ---*/
+  /*--- Store the number of fields and points to be read for clarity. ---*/
 
-  nFields = Restart_Vars[1];
+  const unsigned long nFields = Restart_Vars[1];
+  const unsigned long nPointFile = Restart_Vars[2];
```
TobiKattmann, line 2936: Any specific reason to take the unsigned long datatype here for nFields? Imo that is not necessary as the number of fields is typically <100. And Restart_Vars consists of int's 
That also applies to `nPointFile` which replaces `geometry->GetnPointDomain()` which actually returns an unsigned long ... but now it cant be more than an int? Not that I plan to use more than 2 Billion points anytime soon though 
Or make `Restart_Vars` above a `new unsigned long`

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3094,6 +3115,182 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
...
+    /*--- Repeat while all points are not mapped. ---*/
+    done = true;
+    for (auto iPoint = 0ul; iPoint < nPointDomain; ++iPoint) {
+      isMapped[iPoint] = nDonor[iPoint];
+      done &= nDonor[iPoint];
+    }
+  }
+  } // everything goes out of scope except "localVars"
+
+  /*--- Move to Restart_Data in "funny" order. ---*/
```
TobiKattmann, line 3299: ... my favourite order. 
Well, I honestly did not attempt to understand this implementation 🙈 but I would be interested to why 'funny' (I guess the access pattern into localVars is rather random? but i have no clue tbh)

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3094,6 +3115,182 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
...
+
+  Restart_Data = new passivedouble[nPointDomain*nFields];
+  Restart_Vars[2] = nPointDomain;
+
+  int counter = 0;
+  for (auto iPoint_Global = 0ul; iPoint_Global < geometry->GetGlobal_nPointDomain(); ++iPoint_Global) {
+    const auto iPoint = geometry->GetGlobal_to_Local_Point(iPoint_Global);
+    if (iPoint >= 0) {
```
TobiKattmann, line 3307: When is this not the case? For Halo-points?

TobiKattmann (approved):
Thanks 💐 this is surely a neat feature, especially for grid-independence studies. 
Do you plan to add/ adapt a Testcase?

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: These days I just use the largest one, to avoid those "narrow conversion" warnings and so on.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3094,6 +3115,182 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
...
+
+    /*--- Find the closest target for each donor. ---*/
+
+    vector<su2double> targetDist(nPointDonorMax);
+    vector<unsigned long> iTarget(nPointDonorMax);
+
+    for (auto iDonor = 0ul; iDonor < donorVars.rows(); ++iDonor) {
+      int r=0;
+      adt.DetermineNearestNode(donorVars[iDonor], targetDist[iDonor], iTarget[iDonor], r);
+    }
```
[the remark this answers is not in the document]
pcarruscag: The function writes to that variable.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: I guess formally what that does is sort by global index, which is how data is stored in the files. I'll update that comment.

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: it is only true if iPoint_Global is one of the domain points.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3094,6 +3116,205 @@ void CSolver::Read_SU2_Restart_Binary(CGeometry *geometry, const CConfig *config
...
+    SU2_OMP_ATOMIC
+    done &= myDone;
+
+    SU2_OMP_BARRIER
+  }
+  END_SU2_OMP_PARALLEL
+
+  } // everything goes out of scope except "localVars"
+
+  /*--- Move to Restart_Data in "funny" order. ---*/
```
pcarruscag, line 3299:
```suggestion
  /*--- Move to Restart_Data in ascending order of global index, which is how a matching restart would have been read. ---*/
```

SU2_CFD/src/solvers/CSolver.cpp
TobiKattmann: Ok. For nPointFile you are narrowing it to the version before by giving it an int value ... but then again this is not an issue

SU2_CFD/src/solvers/CSolver.cpp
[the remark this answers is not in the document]
TobiKattmann: Ah, the function takes an `int&` so you cannot give an rvalue like zero directly would be. `r` obbviously goes out of scope and is not used afterwards

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: What?

pcarruscag: Anywho, testcase has been added, thanks for keeping me honest :blush:

TobiKattmann: You replace `geometry->GetnPointDomain()` (which returns an unsigned long) with `nPointFile` which is an unsigned long as well. So far so good. But `const unsigned long nPointFile = Restart_Vars[2];` where `Restart_Vars` is an array of ints, i.e. `nPointFile` cannot be larger than int -> that is the 'narrowing [to the code] version of before' which I mean.

TestCases/navierstokes/flatplate/lam_flatplate_unst.cfg
```diff
@@ -0,0 +1,89 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: Test interp. restart, and auto time-step for dual-time     %
```
TobiKattmann, line 4: 👍

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: Ah ok. I was thinking of this LGTM warning:
`
fixed alerts:
1 for Comparison of narrow type with wide type in loop condition
`
Comparison... I wrote conversion... doh! Same difference though, narrowing would be converting nPointFile to something shorter than int, nothing is lost by going up.

PR #1281, opened 2021-05-08
Chore in CHeatSolver.cpp

TobiKattmann:
## Proposed Changes
Cleanup CHeatSolver.cpp
- define loop vars in the loop
- make stuff const auto
- remove unnecessary vars - or move definition&allocation where needed
- some minor logical changes which hopefully are equivalent ( and therefore don't change the reg test 🤞 )
- Fix primal unsteady Heat restarts

@oleburghardt might be interested in these changes ;)

Can be seen as a follow up to #1256 of @maxaehle (although I dont specifically fix memory leaks)

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CHeatSolver.hpp
+++ b/SU2_CFD/include/solvers/CHeatSolver.hpp
@@ -41,8 +41,8 @@ protected:
   static constexpr size_t MAXNDIM = 3; /*!< \brief Max number of space dimensions, used in some static arrays. */
   static constexpr size_t MAXNVAR = 1; /*!< \brief Max number of variables, for static arrays. */
 
-  bool flow; /*!< \brief Use solver as a scalar transport equation of Temperature for the inc solver. */
-  bool heat_equation; /*!< \brief use solver for heat conduction in solids. */
+  const bool flow; /*!< \brief Use solver as a scalar transport equation of Temperature for the inc solver. */
+  const bool heat_equation; /*!< \brief use solver for heat conduction in solids. */
 
   unsigned short nVarFlow, nMarker;
   vector<vector<su2double> > HeatFlux;
@@ -67,11 +67,6 @@ protected:
   inline CVariable* GetBaseClassPointerToNodes() override { return nodes; }
 
 public:
-
-  /*!
-   * \brief Constructor of the class.
-   */
-  CHeatSolver(void);
 
   /*!
    * \brief Constructor of the class.
--- a/SU2_CFD/include/solvers/CHeatSolver.hpp
+++ b/SU2_CFD/include/solvers/CHeatSolver.hpp
@@ -41,7 +41,10 @@ protected:
   static constexpr size_t MAXNDIM = 3; /*!< \brief Max number of space dimensions, used in some static arrays. */
   static constexpr size_t MAXNVAR = 1; /*!< \brief Max number of variables, for static arrays. */
 
-  unsigned short nVarFlow, nMarker, CurrentMesh;
+  bool flow; /*!< \brief Use solver as a scalar transport equation of Temperature for the inc solver. */
+  bool heat_equation; /*!< \brief use solver for heat conduction in solids. */
+
+  unsigned short nVarFlow, nMarker;
   vector<vector<su2double> > HeatFlux;
   vector<su2double> HeatFlux_per_Marker;
   su2double Total_HeatFlux;
--- a/SU2_CFD/src/solvers/CHeatSolver.cpp
+++ b/SU2_CFD/src/solvers/CHeatSolver.cpp
@@ -32,21 +32,16 @@ CHeatSolver::CHeatSolver(void) : CSolver() { }
 
 CHeatSolver::CHeatSolver(CGeometry *geometry, CConfig *config, unsigned short iMesh) : CSolver() {
 
-  unsigned short iVar, iDim, nLineLets, iMarker;
+  flow = ((config->GetKind_Solver() == INC_NAVIER_STOKES) ||
+          (config->GetKind_Solver() == INC_RANS) ||
+          (config->GetKind_Solver() == DISC_ADJ_INC_NAVIER_STOKES) ||
+          (config->GetKind_Solver() == DISC_ADJ_INC_RANS));
 
-  bool multizone = config->GetMultizone_Problem();
-
-  int rank = MASTER_NODE;
-
-  bool heat_equation = ((config->GetKind_Solver() == HEAT_EQUATION) ||
-                        (config->GetKind_Solver() == DISC_ADJ_HEAT));
+  heat_equation = ((config->GetKind_Solver() == HEAT_EQUATION) ||
+                   (config->GetKind_Solver() == DISC_ADJ_HEAT));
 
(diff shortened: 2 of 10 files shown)


SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -32,22 +32,17 @@ CHeatSolver::CHeatSolver(void) : CSolver() { }
...
+          (config->GetKind_Solver() == DISC_ADJ_INC_RANS));
 
-  bool multizone = config->GetMultizone_Problem();
-
-  int rank = MASTER_NODE;
-
-  bool heat_equation = ((config->GetKind_Solver() == HEAT_EQUATION) ||
-                        (config->GetKind_Solver() == DISC_ADJ_HEAT));
+  heat_equation = ((config->GetKind_Solver() == HEAT_EQUATION) ||
+                   (config->GetKind_Solver() == DISC_ADJ_HEAT));
```
pcarruscag, line 41: GetFluidProblem, GetHeatProblem ? make the booleans const and use the initializer list to set them forever?

TobiKattmann: One has to set the initializer list for const class-members for all constructors obviously. that cost me a few minutes. I deleted the empty constructor as I couldn't find it being used, i.e. `grep -ri CHeatSolver()` was empty.

TobiKattmann: Concerning`GetFluidProblem`: I guess one could make a `GetIncFluidProblem` to be fully correct, but that should be catched elsewhere.

pcarruscag: Should not matter for this.

SU2_CFD/include/solvers/CHeatSolver.hpp
```diff
@@ -41,7 +41,10 @@ class CHeatSolver final : public CSolver {
   static constexpr size_t MAXNDIM = 3; /*!< \brief Max number of space dimensions, used in some static arrays. */
   static constexpr size_t MAXNVAR = 1; /*!< \brief Max number of variables, for static arrays. */
 
-  unsigned short nVarFlow, nMarker, CurrentMesh;
+  const bool flow; /*!< \brief Use solver as a scalar transport equation of Temperature for the inc solver. */
+  const bool heat_equation; /*!< \brief use solver for heat conduction in solids. */
```
pcarruscag, line 45: Or you could have given them default values, then the default constructors use those.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -420,23 +352,23 @@ void CHeatSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
...
-  unsigned short iDim, iVar;
-  unsigned long iEdge, iPoint, jPoint;
-  bool flow = ((config->GetKind_Solver() == INC_NAVIER_STOKES)
-               || (config->GetKind_Solver() == INC_RANS)
-               || (config->GetKind_Solver() == DISC_ADJ_INC_NAVIER_STOKES)
-               || (config->GetKind_Solver() == DISC_ADJ_INC_RANS));
-  bool muscl = (config->GetMUSCL_Heat());
+
+  vector<su2double> Vector_i(nDim, 0.0);
+  vector<su2double> Vector_j(nDim, 0.0);
```
pcarruscag, line 357: nope, static of size MAXNDIM.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -420,23 +352,23 @@ void CHeatSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
...
 
   if(flow) {
 
     nVarFlow = solver_container[FLOW_SOL]->GetnVar();
 
-    for (iEdge = 0; iEdge < geometry->GetnEdge(); iEdge++) {
+    /*--- Define some auxiliary vectors related to the primitive flow solution ---*/
+    vector<su2double> Primitive_Flow_i(nVarFlow, 0.0);
+    vector<su2double> Primitive_Flow_j(nVarFlow, 0.0);
```
pcarruscag, line 365: I get it that you want to clean this, but we might as well work towards a better scalar solver framework (as we have talked about), because all these things are going to be deleted...

SU2_CFD/src/solvers/CHeatSolver.cpp
TobiKattmann: As you see, I wanted to be faithful ⛪ ...but the devil himself sent `maybe-unitialized`-Warnings to derail my train of good intentions. 

(I wanted a 1-line solution to allocate and initialize with sth and I wasn't aware that `<static array>= {x}` not only sets the first value to x but actually sets all other elements to zero. Had I known that back in the day, i would have not sinned)

SU2_CFD/src/solvers/CHeatSolver.cpp
TobiKattmann: What do you mean with `all these things`?
- just these containers? - they are now no more class data so i wouldn't matter as it only lives in this very narrow scope. 
- If you mean the transported temperature scalar for incompressible flow alltogehter I fully agree. I didnt't want to open the discussion here but let's go for it : 

Can we delete the weakly coupled heat solver for incompressible flow in its entirety? From my incomplete perspective: yes

Or do you mean the whole heatsolver - with solid heat conduction as a special case of a (non-)transported scalar?

SU2_CFD/include/solvers/CHeatSolver.hpp
TobiKattmann: hm, I would actually prefer to make the new-constructor-author aware that he needs to take care of that ... but on the other hand one might expect that kind of awareness. But that's more of a general idea. Thanks for hint ... didn't have that on my watch

pcarruscag: Equally valid, just pointing out alternatives :)

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: Fair enough. I mean methods like UpwindResidual, ViscousResidual, SetTimeStep, DualTimeResidual, etc. etc. will have a common implementation across all scalar solvers.

TobiKattmann:
Fixed unsteady Heat restarts. Dotted blue line simulation runs trough and the solid blue is restarted (from the red ... which is identical to the dotted blue up to this point ... so it was kinda unnecessary :facepalm: ). 

before 4990753 
imperfectUnsteadyCHTrestart
after 
perfectUnsteadyCHTrestart

SU2_CFD/src/solvers/CHeatSolver.cpp
TobiKattmann: i wont go further in that matter in this PR though and I'd suggest to talk about this maybe briefly on Wednesday

pcarruscag (approved):

TobiKattmann:
Merging this in now as the restart fix makes my life a bit easier in conjunction with #1260 . I am more than happy to address any comments, change requests etc after this was merged because this wasn't open very long tbh

PR #1284, opened 2021-05-09
Hybrid Parallel AD (Part 2/?)

pcarruscag:
## Proposed Changes
1 - Local indices for solution variables (which are overwritten during recording), even in singlezone problems (it was already used in multizone), to make things consistent for the two drivers.

2 - SU2_TYPE::GetDerivative was based on a global counter and a shared vector of input indices, which meant derivatives had to be extracted in the same order they were registered, and made the whole process not thread-safe. This is gone, if an input variable is overwritten during recording its (initial) index needs to be managed explicitly.

3 - Simplify and improve the restart logic because of unsteady FSI issues (not an Hybrid AD topic, but I had enough branches).

## Related Work
Continues #1214 

## PR Checklist


TobiKattmann:
Do you plan to add any hybrid parallel AD stuff here? I would prefer if this was a follow-up to #1260 where the Single zone DA input index storing is changed to the multizone strategy with explicit index-containers owned by the solver. You know ... smaller PR, faster merging

I will also add a preliminary unsteady cht adjoint case myself for coverage. Then we could also move the `!CrossTerm` of ExtractAdjoint_Solution thing in this PR and see if it breaks stuff.

pcarruscag:
To some extent this is a fix for hybrid parallel AD, because the single zone driver used the version of RegisterInput that pushed indices into that global "inputValues" vector, and that would not work with OpenMP.
This PR will not be very big anyway. You can test your testcase here, or in #1260, or in a different PR, and them we test the cross term thing.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -362,25 +362,25 @@ void CDiscAdjSolver::ExtractAdjoint_Solution(CGeometry *geometry, CConfig *confi
   END_SU2_OMP_MASTER
 
   /*--- Extract and store the adjoint of the primal solution at time n ---*/
-  if (time_n_needed) {
+  if (time_n_needed && !CrossTerm) {
```
TobiKattmann, line 365: As discussed in #1260 . I tested my cht case which gave identical residuals so I would expect it is the same for fsi, but as there is a testcase there is no guessing necessary :)

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -344,6 +341,11 @@ void CDiscAdjSolver::ExtractAdjoint_Solution(CGeometry *geometry, CConfig *confi
   }
   END_SU2_OMP_FOR
 
+  /*--- Residuals and time_n terms are not needed when evaluating multizone cross terms. ---*/
+  if (CrossTerm) return;
```
TobiKattmann, line 345: well ... if thought about this way via return as well but went with the smaller change ... i guess you merge-conflicted that away(?)

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -344,6 +341,11 @@ void CDiscAdjSolver::ExtractAdjoint_Solution(CGeometry *geometry, CConfig *confi
...
+
+  SetResToZero();
```
TobiKattmann, line 347: Could we have an unwanted addition to the residual because of moving the `SetResToZero` below the CrossTerm-return? The residual should compare the (full) adjoint solution so aren't we missing the External (crossterms + dualTime terms) here anyway. I guess just the dualTime Terms are constant so we can leave them out. Or is it correct now as CrossTerms are extracted first. I have to look that up

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -297,32 +297,29 @@ void CDiscAdjSolver::RegisterOutput(CGeometry *geometry, CConfig *config) {
...
 
-  if(!config->GetMultizone_Problem()) nodes->Set_OldSolution();
+  if (!config->GetMultizone_Problem()) nodes->Set_OldSolution();
 
   SU2_OMP_FOR_STAT(omp_chunk_size)
   for (auto iPoint = 0ul; iPoint < nPoint; iPoint++) {
 
     /*--- Extract the adjoint solution ---*/
 
+    su2double Solution[MAXNVAR] = {0.0};
```
TobiKattmann, line 322: moving that container was necessary to have one per thread, right? so it wasn't correct for potential omp adjoint before?

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
TobiKattmann: Well, ... this commit lived a short life of 30 minutes

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
pcarruscag: no, just making it more local, same idea as iPoint.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
pcarruscag: No, the residual should measure the norm of the residual of the (linear) adjoint equations, otherwise it would measure stagnation instead of convergence.
It just so happens that differences between consecutive fixed point iterations give you that residual (because the fixed point is the same as a Richardson iteration for the right-preconditioned adjoint equations).
The single zone uses differences of the total terms, the multizone uses differences of the product between total terms and the Jacobian of the iteration (it is easier to do it that way), this is equivalent since the right hand side is constant during inner iterations.

TobiKattmann:
Did you expect that your introduced changes fail the regression test for fsi related cases? If so, could you point out what bits are responsible

pcarruscag:
I'm trying to fix the unsteady FSI restarts.

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -593,10 +600,18 @@ void CMeshSolver::ComputeGridVelocity_FromBoundary(CGeometry **geometry, CNumeri
...
   /*--- Clear residual (loses AD info), we do not want an incremental solution. ---*/
   SU2_OMP_PARALLEL {
     LinSysRes.SetValZero();
-    LinSysSol.SetValZero();
+
+    SU2_OMP_FOR_STAT(omp_chunk_size)
+    for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++)
+      for (unsigned short iDim = 0; iDim < nDim; iDim++)
+        LinSysSol(iPoint, iDim) = geometry[MESH_0]->nodes->GetGridVel(iPoint)[iDim] * velRef;
+    END_SU2_OMP_FOR
```
pcarruscag, line 614: @TobiKattmann for FSI the mesh solver solves two systems, displacement and velocity, these can be initialized to reduced their cost.
This also reduces the cost of clearing indices and recording the primal iteration, in those cases the initialization should be perfect.

TestCases/serial_regression.py
```diff
@@ -1903,7 +1903,7 @@ def main():
     pywrapper_rigidMotion.cfg_dir       = "py_wrapper/flatPlate_rigidMotion"
     pywrapper_rigidMotion.cfg_file      = "flatPlate_rigidMotion_Conf.cfg"
     pywrapper_rigidMotion.test_iter     = 5
-    pywrapper_rigidMotion.test_vals     = [-1.551335, 2.295594, 0.350050, 0.093081]
+    pywrapper_rigidMotion.test_vals     = [-1.614170, 2.242953, 0.350050, 0.093137]
```
pcarruscag, line 1906: these changed because the volumes at time n and time n-1 are now set equal to the initial value (changes in CDriver::GeometryPreprocessing and in CFVMFlowSolverBase<V, R>::PushSolutionBackInTime)

The other dynamic FSI cases changed because of that and because of initializing the grid-velocity-elasticity-problem.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -140,28 +140,22 @@ CDriver::CDriver(char* confFile, unsigned short val_nZone, SU2_Comm MPICommunica
...
-    iteration_container[iZone]   = new CIteration*    [nInst[iZone]];
-    solver_container[iZone]      = new CSolver***     [nInst[iZone]];
-    integration_container[iZone] = new CIntegration** [nInst[iZone]];
-    numerics_container[iZone]    = new CNumerics****  [nInst[iZone]];
-    grid_movement[iZone]         = new CVolumetricMovement* [nInst[iZone]];
+    geometry_container[iZone]    = new CGeometry**    [nInst[iZone]] ();
+    iteration_container[iZone]   = new CIteration*    [nInst[iZone]] ();
+    solver_container[iZone]      = new CSolver***     [nInst[iZone]] ();
+    integration_container[iZone] = new CIntegration** [nInst[iZone]] ();
+    numerics_container[iZone]    = new CNumerics****  [nInst[iZone]] ();
```
TobiKattmann, line 147: C++ understanding question: With `()` you call the default constructor, right? What is the benefit of that? Does it automatically initialize with `nullptr` as that set-ting was deleted below?

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1137,170 +1101,56 @@ void CDriver::Inlet_Preprocessing(CSolver ***solver, CGeometry **geometry,
...
-      if (time_domain) {
-        if (config->GetRestart()) val_iter = config->GetRestart_Iter()-1;
-        else val_iter = config->GetUnst_AdjointIter()-1;
+    for (auto iSol = 0u; iSol < MAX_SOLS; ++iSol) {
+      auto sol = solver[MESH_0][iSol];
+      if (sol && !sol->GetAdjoint()) {
+        /*--- Note that the mesh solver always loads the most recent file (and not -2). ---*/
+        SU2_OMP_PARALLEL_(if(sol->GetHasHybridParallel()))
+        sol->LoadRestart(geometry, solver, config, val_iter + (iSol==MESH_SOL && dt_step_2nd), update_geo);
+        END_SU2_OMP_PARALLEL
```
TobiKattmann, line 1133: Awesome cleanup of the restart logic🧹

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -239,12 +239,12 @@ void CMultizoneDriver::Preprocess(unsigned long TimeIter) {
 
     /*--- Set the initial condition for EULER/N-S/RANS ---------------------------------------------*/
     /*--- For FSI, the initial conditions are set, after the mesh has been moved. --------------------------------------*/
-    if (!fsi && !config_container[iZone]->GetDiscrete_Adjoint() && config_container[iZone]->GetFluidProblem()) {
+    if (!fsi && config_container[iZone]->GetFluidProblem()) {
```
pcarruscag, line 242: I don't think the multizone driver is going to be recorded anytime soon so this was redundant.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -293,45 +292,35 @@ void CDiscAdjSolver::RegisterVariables(CGeometry *geometry, CConfig *config, boo
...
     /*--- Extract the adjoint solution ---*/
 
-    if(config->GetMultizone_Problem()) {
-      direct_solver->GetNodes()->GetAdjointSolution_LocalIndex(iPoint,Solution);
-    }
-    else {
-      direct_solver->GetNodes()->GetAdjointSolution(iPoint,Solution);
-    }
+    su2double Solution[MAXNVAR] = {0.0};
+    direct_solver->GetNodes()->GetAdjointSolution(iPoint,Solution);
```
TobiKattmann, line 323: Nice to have the input-index & output-index storing-strategies consolidated between single and multizone adjoint 👍

SU2_CFD/src/solvers/CDiscAdjFEASolver.cpp
```diff
@@ -418,24 +367,18 @@ void CDiscAdjFEASolver::SetSensitivity(CGeometry *geometry, CConfig *config, CSo
...
-      if(config->GetMultizone_Problem()) {
-        Sensitivity = geometry->nodes->GetAdjointSolution(iPoint, iDim);
-      }
-      else {
-        Sensitivity = SU2_TYPE::GetDerivative(Coord[iDim]);
-        /*--- Set the index manually to zero. ---*/
-        AD::ResetInput(Coord[iDim]);
-      }
+      su2double Sensitivity = geometry->nodes->GetAdjointSolution(iPoint, iDim);
+      AD::ResetInput(Coord[iDim]);
```
pcarruscag, line 379: For whatever reason this reset is fundamental for the correctness of multiple recordings with different inputs.
Reseting the indices (explicitly) during the CLEAR_INDICES recording does not work :mage:

SU2_CFD/src/solvers/CDiscAdjMeshSolver.cpp
```diff
@@ -98,18 +98,19 @@ void CDiscAdjMeshSolver::SetRecording(CGeometry* geometry, CConfig *config){
...
+  direct_solver->GetNodes()->Register_MeshCoord();
+
 }
 
 void CDiscAdjMeshSolver::RegisterVariables(CGeometry *geometry, CConfig *config, bool reset){
 
+  /*--- Register boundary displacements as input.
+   * Except for FSI, where they are determined by the FEA solver. ---*/
+
+  if (config->GetFSI_Simulation()) return;
```
pcarruscag, line 111: Registering intermediate variables as input also causes issues with the multiple recordings.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1137,170 +1101,56 @@ void CDriver::Inlet_Preprocessing(CSolver ***solver, CGeometry **geometry,
...
-    case ADJ_NAVIER_STOKES : ns = true; turbulent = (config->GetKind_Turb_Model() != NONE); adj_ns = true; break;
-    case ADJ_RANS : ns = true; turbulent = true; adj_ns = true; adj_turb = (!config->GetFrozen_Visc_Cont()); break;
-    case DISC_ADJ_EULER: case DISC_ADJ_INC_EULER: euler = true; disc_adj = true; break;
-    case DISC_ADJ_NAVIER_STOKES: case DISC_ADJ_INC_NAVIER_STOKES: ns = true; disc_adj = true; heat = config->GetWeakly_Coupled_Heat(); break;
-    case DISC_ADJ_RANS: case DISC_ADJ_INC_RANS: ns = true; turbulent = true; disc_adj = true; disc_adj_turb = (!config->GetFrozen_Visc_Disc()); heat = config->GetWeakly_Coupled_Heat(); break;
-    case DISC_ADJ_FEM_EULER: fem_euler = true; disc_adj = true; break;
-    case DISC_ADJ_FEM_NS: fem_ns = true; disc_adj = true; break;
-    case DISC_ADJ_FEM_RANS: fem_ns = true; turbulent = true; disc_adj = true; disc_adj_turb = (!config->GetFrozen_Visc_Disc()); break;
-    case DISC_ADJ_FEM: fem = true; disc_adj_fem = true; break;
-    case DISC_ADJ_HEAT: heat = true; disc_adj_heat = true; break;
```
pcarruscag, line 1210: :rocket: :fireworks: :skull_and_crossbones:

TestCases/hybrid_regression.py
```diff
@@ -646,7 +646,7 @@ def main():
     fsi2d.cfg_dir   = "fea_fsi/WallChannel_2d"
     fsi2d.cfg_file  = "configFSI.cfg"
     fsi2d.test_iter = 4
-    fsi2d.test_vals = [4.000000, 0.000000, -3.768521, -4.159940]
+    fsi2d.test_vals = [4, 0, -3.743230, -4.133462]
```
TobiKattmann, line 649: Of course it would be nice here to have more coverage on the screen output vals.

TobiKattmann (approved):
Thanks for the adjoint consolidation of index storing things and all the other stuff 💐 . Imo this helps quite a bit to understand the DA solver (single & multizone). I have mostly question for my own understanding below. (I also have to come back to some of your previous comments/asnwers which I have to understand ;) )

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Yes

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2112,80 +2127,43 @@ class CVariable {
   /*!
    * \brief Register the variables in the solution array as input/output variable.
    * \param[in] input - input or output variables.
-   * \param[in] push_index - boolean whether we want to push the index or save it in a member variable.
    */
-  void RegisterSolution(bool input, bool push_index = true);
```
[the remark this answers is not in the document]
pcarruscag: It is still used by a few variables.

SU2_CFD/src/drivers/CDriver.cpp
pcarruscag: Perhaps the only advantage of the solver container being an array.

SU2_CFD/include/variables/CMeshVariable.hpp
```diff
@@ -93,9 +93,11 @@ class CMeshVariable : public CVariable {
   /*!
    * \brief Recover the value of the adjoint of the mesh coordinates.
    */
-  inline void GetAdjoint_MeshCoord(unsigned long iPoint, su2double *adj_mesh) const final {
-    for (unsigned long iDim = 0; iDim < nDim; iDim++)
-      adj_mesh[iDim] = SU2_TYPE::GetDerivative(Mesh_Coord(iPoint,iDim));
+  inline void GetAdjoint_MeshCoord(unsigned long iPoint, su2double *adj_mesh) final {
+    for (unsigned long iDim = 0; iDim < nDim; iDim++) {
+      adj_mesh[iDim] = AD::GetDerivative(AD_InputIndex(iPoint,iDim));
+      AD::ResetInput(Mesh_Coord(iPoint,iDim));
```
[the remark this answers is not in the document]
pcarruscag: I think the trend is that if something is an input the index is not cleared by the dummy recording (since nothing is written to an input) and so you have to clear it explicitly.
Now, I guess this is only strictly required if sometimes (i.e. some recordings) you register that variable and sometimes you don't. If you always register it then the index is always valid, but if you leave it with "dangling" indices then you have a problem.
From my comment above it also seems that the location where this reset is performed matters... (but I might also have changed two things at the same time while testing).

SU2_CFD/include/variables/CVariable.hpp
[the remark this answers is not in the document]
TobiKattmann: ah yes, by everything in `CDiscAdjSolver::RegisterVariables` i.e. Mach, Freestream values etc. As far as I see all those vars are unaffected by the fixed-point update so their original value/variable stays in memory -> so it should be possible to go without that logic . That would of course mean the `GetDerivative(const su2double& data)` has to be changed from 
```c++
  FORCEINLINE passivedouble GetDerivative(const su2double& data) {
    return AD::getGlobalTape().getGradient(AD::inputValues[AD::adjointVectorPosition++]);
  }
```
to the one below
```c++
FORCEINLINE passivedouble GetDerivative(const su2double& data) {return data.getGradient();}
```
RN there is also this fake input `const su2double& data` where one thinks one could directly extract fom a variable like so
```c++
Local_Sens_Mach  = SU2_TYPE::GetDerivative(Mach);
```
which gives a false sense of whats happening because one would expect it works like the proposed one (I certainly did). It should also made clear in those two functions (register & extract) that the order of registering has to match the order of extracting. that is currently not the case.

In case the `inputValues` is kept I would like to see the `GetDerivative(const su2double& data) ` be removed, added explanation in Register/Extract-Variables and maybe switch the default `push_index=True` to `False` in `AD::RegisterInput` to have this additional way to say "look-what-that-boolean-does" and then everyone should be aware that extract the same way you register is what should be done

(When I say `one should do sth` in the text or anywhere I am happy to do it myself but I'll ask first if I overlook sth in my ideas)

TestCases/serial_regression.py
TobiKattmann: 👍 for the explanation

SU2_CFD/src/solvers/CDiscAdjFEASolver.cpp
TobiKattmann: If @jblueh is at the DEV-meeting I will prob ask about this ResetInput and CLEAR_INDICES-recording (again 😞 🐔 🧠 ) to make sure I get how and why things work

SU2_CFD/src/drivers/CDriver.cpp
TobiKattmann: ... this kind of cleanup will also makes the potential change to `enum class Kind_Solver` less tedious

SU2_CFD/include/variables/CVariable.hpp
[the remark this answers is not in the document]
pcarruscag: Ah yes, I remember the day I saw that for the first time... Sadly I cannot reproduce the words I hurled at the screen here xD
Anyway, let's see if it works...

SU2_DOT/src/SU2_DOT.cpp
```diff
@@ -712,21 +710,13 @@ void SetProjection_AD(CGeometry *geometry, CConfig *config, CSurfaceMovement *su
...
 
-      /*--- Initilization with su2double resets the index ---*/
+      config->SetDV_Value(iDV, iDV_Value, 0.0);
 
-      DV_Value = 0.0;
-
-      AD::RegisterInput(DV_Value);
-
-      config->SetDV_Value(iDV, iDV_Value, DV_Value);
+      AD::RegisterInput(config->GetDV_Value(iDV, iDV_Value));
```
TobiKattmann, line 719: I wondered what made this break before and I can only think that before the local `DV_VALUE` was registered and below we now(!) extracted what is in `config->GetDV_Value(...)`. Before this PR it accessed the hidden `inputValues` so everything was good (no matter the input value to `SU2_TYPE::GetDerivative(...)`). 
```c++
      DV_Value = config->GetDV_Value(iDV, iDV_Value);
      my_Gradient = SU2_TYPE::GetDerivative(DV_Value);
```
Now you changed it to only work with the `DV_Value` that lives in the config and that is consistent with this change up here. 

The problem before this very commit was probably index management optimization because `config->SetDV_Value(iDV, iDV_Value, DV_Value);` before was just an assignment. But that is sth I would ask @jblueh in the Dev-meeting as well

TestCases/disc_adj_fsi/dyn_fsi/grad_dv.opt.ref
```diff
@@ -1,9 +1,9 @@
...
-0	-3.465632121754074e-03
-1	-1.844003360219930e-03
-2	-7.925070240684198e-04
-3	-2.742895873196515e-04
-4	-2.738093165742900e-04
-5	-7.890548338696854e-04
-6	-1.831172969426724e-03
-7	-3.431365309796604e-03
+0	-3.461460672772851e-03
+1	-1.841786315630031e-03
```
TobiKattmann, line 3: I know the changes are marginal and probably ok, but if this is not a fully converged adjoint and DOT step (not just 20 SU2_CFD_AD steps and then SU2_DOT_AD) I would prefer to have that checked on the 'converged' gradient. Maybe you did this already

TobiKattmann (approved):
Man this is good stuff 💐 Also without any problems (and modifications) with Register/Extract_Variable. I hope there is at least some Testcases that had e.g. AOA sensitivity in their screen output? (Did you maybe even check that)

And if I understood it correctly that this change revealed, that in two occasions 
```c++
  FORCEINLINE passivedouble GetDerivative(const su2double& data) {
    return AD::getGlobalTape().getGradient(AD::inputValues[AD::adjointVectorPosition++]);
  }
```
fooled people with its fake input  value.

Thanks again! As this is now at `+500 -1000` and quite some changes some to the DA machine room I would personally prefer to have this merged without any major other updates. Also kudos for updating the PR explanation on top 👍

TestCases/disc_adj_fsi/dyn_fsi/grad_dv.opt.ref
pcarruscag: I changed the settings of the case to make it converge better/faster.

SU2_DOT/src/SU2_DOT.cpp
pcarruscag: We were using always the same variable (declared at the top of the function).
On each loop iteration the =0 part would reset the previously registered index, then on extraction we were also writing over the same variable.

Common/include/basic_types/ad_structure.hpp
```diff
@@ -299,7 +299,7 @@ namespace AD{
 
   FORCEINLINE void RegisterOutput(su2double& data) {AD::getGlobalTape().registerOutput(data);}
 
-  FORCEINLINE void ResetInput(su2double &data) {data.getGradientData() = su2double::GradientData();}
+  FORCEINLINE void ResetInput(su2double &data) {data = data.getValue();}
```
TobiKattmann, line 302: 👍 now it is fully equivalent with what is done during CLEAR_INDICES recording.

TobiKattmann: at least from the standpoint that you clear the index by writing to the primal variable

jblueh: Thanks for adding this already @pcarruscag.

TobiKattmann: The fastest keyboard in the west 🤠

pcarruscag:
Back to WIP, sorry @TobiKattmann.
It may be worth git cherry-picking 19d3329da5b596c19a463e0ef8fbeae72e1f02e0 to see if it fixes the issues @oleburghardt found.

Common/include/code_config.hpp
```diff
@@ -92,7 +92,7 @@ using su2double = codi::RealReversePrimal;
 #elif CODI_PRIMAL_INDEX_TAPE
 using su2double = codi::RealReversePrimalIndex;
 #else
-using su2double = codi::RealReverse;
+using su2double = codi::RealReverseIndex;
```
TobiKattmann, line 95: by the way concerning these preprocessor statements? This is the code snippet from `code_config.hpp`
```c++
#if defined(CODI_REVERSE_TYPE) // reverse mode AD
#include "codi.hpp"
#include "codi/tools/dataStore.hpp"

#ifndef CODI_INDEX_TAPE
#define CODI_INDEX_TAPE 0
#endif
#ifndef CODI_PRIMAL_TAPE
#define CODI_PRIMAL_TAPE 0
#endif
#ifndef CODI_PRIMAL_INDEX_TAPE
#define CODI_PRIMAL_INDEX_TAPE 0
#endif

#if defined(HAVE_OMP)
using su2double = codi::RealReverseIndexParallel;
#else
#if CODI_INDEX_TAPE
using su2double = codi::RealReverseIndex;
#elif CODI_PRIMAL_TAPE
using su2double = codi::RealReversePrimal;
#elif CODI_PRIMAL_INDEX_TAPE
using su2double = codi::RealReversePrimalIndex;
#else
using su2double = codi::RealReverse;
#endif
#endif
```
In AD mode I can see in the build files that the flag `-DCODI_REVERSE_TYPE` is given so that code will be excecuted. Now all those three other vars `CODI_INDEX_TAPE ` `CODI_PRIMAL_TAPE` and `CODI_PRIMAL_INDEX_TAPE` are set to 0 (and none of those can be found in the build file ... or are #defined elsewhere) so in the following part will use the `#else` clause ... which you changed. Is it then the `JacobiTape` or now `JacobiIndexTape`?

jblueh: `codi::RealReverse` uses the `JacobiTape` and `codi::RealReverseIndex` uses the `JacobiIndexTape`. The change to `codi::RealReverseIndex` is a temporary way of using that type in the CI pipeline, this will be undone before merging.

TobiKattmann: Ok so the `codi::RealReverseIndex` using the `JacobiIndexTape` is just for test purposes now? Sure, there is curiosity, but this change will be reverted: What is the idea with this test? (Or am I fully missing the point here)

I also guess that this RealReverse change introduced the failing regression test not the 3 updated Codi lines?

pcarruscag: ... who are you and what have you done with the Tobi that was the meeting yesterday?
(Johannes is working on the issues the preaccumulation has with the RealReverseIndex type, which is one of the steps required to make it work with hybrid parallel AD) the codi lines are the fix, the failing tests are some other problem in the FEA solver)

TobiKattmann: ... attention span issues 🐔 😞 
but thanks for the explanation.

TobiKattmann:
Well thanks for reverting the latest two commits with RealReverseIndex and the CoDiPack update and merging this. I guess that work will go on in another PR

PR #1286, opened 2021-05-11
New data structure for multidimensional data (+some bugfixes)

### LINKED ISSUE #1293: compilation with GCC v11.1 fails
**Describe the bug**
When compiling the master branch or v7.1.1 of SU2, as described in the [Quick compilation guide](https://su2code.github.io/docs_v7/Build-SU2-Linux-MacOS/#quick-compilation-guide) with gcc version 11.1.0, the ninja command fails, saying that `'numeric_limits' is not a member of 'std'`.

**Likely reason**
There were [header dependency changes in GCC 11](https://www.gnu.org/software/gcc/gcc-11/porting_to.html#header-dep-changes). Probably because of this, the header `<limits>` is not (recursively) `#include`d from the file

    Common/include/geometry/primal_grid/CPrimalGrid.hpp

any more. Thus `std::numeric_limits` has not been declared when it is used in line 243.

**Solution**
`#include <limits>` in `CPrimalGrid.hpp`

maxaehle:
**Edit**: Besides introducing the `NdFlattener` data structure, this PR applies the following bugfixes:
- correct retrieval of the wall roughness in a multizone setup, the original intention of this PR
- the non-MPI substitute for `SU2_MPI::Allgatherv` and `Alltoallv` now takes the displacement into account
- fix #1293, adding a missing `#include` statement 

Below is the original PR description.

## Proposed Changes
I think that currently the roughness at the closest wall is retrieved in a wrong way for multizone setups. The current structure is like this:

    def CGeometry::ComputeWallDistance:
        for each iZone:
             construct an ADT which stores the markers of iZone
             for each jZone:
                 call CPhysicalGeometry::SetWallDistance to jZone geometry which \
                 reduces the stored wall distances of jZone's points according to iZone's markers in the ADT and\
                 sets wall roughnesses of jZone's points according to data from jZone's geometry

The issue I perceive here is that the closest wall in the ADT, identified by the `markerID`, belongs to `iZone`. `iZone`'s geometry and config must therefore be an additional argument to `CPhysicalGeometry::SetWallDistance`. We must not use the wall roughness from `jZone`'s geometry and config associated to the markerID, as it is currently done.

## Related Work
?
## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -11041,12 +11038,15 @@ void CPhysicalGeometry::SetWallDistance(const CConfig *config, CADTElemClass *Wa
...
+      if(dist < nodes->GetWall_Distance(iPoint)){
+        nodes->SetWall_Distance(iPoint, dist);
 
-      if (config->GetnRoughWall() > 0) {
-        auto index = GlobalMarkerStorageDispl[rankID] + markerID;
-        auto localRoughness = GlobalRoughness_Height[index];
-        nodes->SetRoughnessHeight(iPoint, localRoughness);
+        if (config->GetnRoughWall() > 0) {
+          const CPhysicalGeometry* ph_geometry = dynamic_cast<const CPhysicalGeometry*>(geometry);
+          auto index = ph_geometry->GlobalMarkerStorageDispl[rankID] + markerID;
```
pcarruscag, line 11046: Thanks Max, I agree with your assessment of the issue, but perhaps this could be done avoiding passing geometries into geometry just to retrieve a couple of fields (which then requires casting those geometries).

For example, CPhysicalGeometry::SetGlobalMarkerRoughness could become a free function, or a member of CConfig, that returns a custom type with interface `(rank,marker)`, objects of this type would be prepared in CGeometry::ComputeWallDistance and passed to SetWallDistance instead of entire geometries (if they are `.empty()` the roughness would not be set).

Alternatively, this would be more generic, and possibly useful for the future, we could pass `jZone` to SetWallDistance and store, in the nodes (i.e. CPoint), `zone,rank,marker` associated with the wall distance. Then, a separate method e.g. `SetWallRoughness`, would use this info to populate the wall roughness. This method would be called from CGeometry::ComputeWallDistance and it would also make use of above change to CPhysicalGeometry::SetGlobalMarkerRoughness (i.e. "freeing it" or moving to CConfig).

Minimum requirement IMO would be to add a virtual method to CGeometry to obtain the roughness from (rank,marker) instead of dynamic cast to access member variables.

maxaehle:
Ok, now `CPoint` stores information on the closest wall element location in vectors `ClosestWall_{Rank,Zone,Marker,Elem}`. `CPoint::SetWallRoughness` performs the lookup, inserting the location into the map (rank,zone,marker)->(roughness) that is provided to `CPoint::SetWallRoughness` as an argument of a new templated type `NdFlattener`.

I wrote the new file `Common/include/ndflattener.hpp` some time ago and I think it is a nice generalization of the

 > custom type with interface (rank, [zone,] marker)

you're referring to. In short, it allows to

 - serialize pointer-to-pointer-to... arrays into 1D arrays, as many as there are indices. The array corresponding to the last index stores the data and the others store indices.
 - And to `MPI_Allgatherv` them, adding a new index for the rank.

Using it detaches a bit from the original goal of this PR (which was to fix some illegal reads reported by valgrind, because the wrong array of roughnesses was accessed) but I think it's a useful structure. Probably I should also add some tests.

maxaehle:
I found only one testcase for rough walls, in `TestCases/incomp_rans/rough_flatplate/rough_flatplate_incomp.cfg`. It's not contained in the regression tests.

pcarruscag (commented):
Thanks Max, I think this is a nice addition but if you have time, it would be good to make it more obvious how to use it (see below).

Common/src/geometry/CGeometry.cpp
```diff
@@ -3947,5 +3942,36 @@ void CGeometry::ComputeWallDistance(const CConfig* const* config_container, CGeo
...
+    if(!allEmpty){
+      /*--- Store all wall roughnesses in a common data structure. ---*/
+      // prepare parameters for collective communication
+      MPI_Environment<decltype(&(SU2_MPI::Allgatherv)), decltype(MPI_INT), decltype(SU2_MPI::GetComm())> mpi_env;
+      mpi_env.MPI_Allgatherv = &(SU2_MPI::Allgatherv);
+      mpi_env.mpi_index = MPI_UNSIGNED_LONG;
+      mpi_env.mpi_data = MPI_DOUBLE;
+      mpi_env.comm=SU2_MPI::GetComm();
+      SU2_MPI::Comm_rank(mpi_env.comm, &(mpi_env.rank));
+      SU2_MPI::Comm_size(mpi_env.comm, &(mpi_env.size));
```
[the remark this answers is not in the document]
maxaehle: Yes -> `::Get_Nd_MPI_Env()`

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+    for(Index iChild=0; iChild<nChild; iChild++){
+      data[iNode] = f.second(iChild);
+      iNode++;
+    }
+  }
+  
+  /*=== Construct with Allgatherv ===*/
+protected:
+  template<typename MPI_Allgatherv_type, typename MPI_Datatype_type, typename MPI_Communicator_type>
+  void count_g(MPI_Environment<MPI_Allgatherv_type, MPI_Datatype_type, MPI_Communicator_type> const& mpi_env,
```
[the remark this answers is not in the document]
maxaehle: Yes, I changed it to a single template parameter `MPI_Environment_type` everywhere.

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+    }
+    result << "]";
+    return result.str();
+  }
+
+public:
+  NdFlattener(void) {}
+
+  virtual ~NdFlattener(void) {
+    if(data!=nullptr)
```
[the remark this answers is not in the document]
maxaehle: I use `std::vector` now for the data and index arrays. The `Nodes_all` is still a pointer-of-pointers array, but I guess I can make this a vector too.

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+  typedef CurrentLayer LowestLayer;
+
+private:
+  Index nNodes=0;
+  Index iNode=0;
+  Data* data=nullptr;
+
+  /*=== Outputting ===*/
+protected:
+  std::string toPythonString_fromto(Index from, Index to) const {
```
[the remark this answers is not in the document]
maxaehle: Ok

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+  Index iNode=0;
+
+  /*! \brief Indices in the lower layer's indices or data array */
+  Index* indices=nullptr;
+
+  /*! \brief Return a const pointer to the base class (one layer down).
+   */
+  Base const* cast_to_Base() const {
+    return (Base const*)this;
+  }
```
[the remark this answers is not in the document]
maxaehle: Yes, no need for such a function. Replaced by `static_cast` everywhere.

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+   *
+   * Like this: [[1, 2], [10, 20, 30]]
+   * \param[in] from - Beginning of the representation in the indices array.
+   * \param[in] to - Ending of the representation in the indices array.
+   */
+  std::string toPythonString_fromto(Index from, Index to) const {
+    std::string result = "[";
+    for(Index i=from; i<to; ){
+      result += Base::toPythonString_fromto(indices[i], indices[i+1]);
+      if(++i<to) result += ", ";
```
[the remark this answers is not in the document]
maxaehle: ok

Common/include/ndflattener.hpp
```diff
@@ -0,0 +1,558 @@
...
+      Index last_entry_to_be_shifted = Nodes_all_k_cumulated[r+1]-1;
+      Index shift = indices[ first_entry_to_be_shifted - 1];
+      for(Index i=first_entry_to_be_shifted; i<=last_entry_to_be_shifted; i++){
+        indices[ i ] += shift;
+      }
+    }
+    delete[] Nodes_all_K_as_int;
+    delete[] Nodes_all_k_cumulated;
+
+    Base::set_g(mpi_env, Nodes_all, local_version->cast_to_Base());
```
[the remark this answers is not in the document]
maxaehle: replaced by static_cast

Common/src/geometry/dual_grid/CPoint.cpp
```diff
@@ -179,3 +185,14 @@ void CPoint::SetCoord_Old() {
...
 
 void CPoint::SetCoord_SumZero() { parallelSet(Coord_Sum.size(), 0.0, Coord_Sum.data()); }
+
+void CPoint::SetWallRoughness(NdFlattener<su2double,3,unsigned long> const& roughness){
+  for (unsigned long iPoint=0; iPoint<GlobalIndex.size(); ++iPoint) {
+    int rankID = ClosestWall_Rank[iPoint];
+    unsigned short zoneID = ClosestWall_Zone[iPoint];
+    unsigned short markerID = ClosestWall_Marker[iPoint];
+    if(rankID != -1){
+      SetRoughnessHeight(iPoint, roughness.get(rankID, zoneID, markerID));
```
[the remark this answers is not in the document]
maxaehle: This was quite some work but now the []...[] interface is supported as the standard way of accessing data elements. You may also provide less than all indices and get an `IndexAccumulator` object to which you can pass additional parameters.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+      //const typename Nd_type::Index new_offset = this->nd->GetIndices()[this->offset+i];
+      //const typename Nd_type::Index new_size = this->nd->GetIndices()[this->offset+i+1] - new_offset;
+      return LookupType(static_cast<Nd_Base_type*>(this->nd),this->nd->GetIndices()[this->offset+i]);
+    }
+  };
+
+  template<size_t _N, typename _Nd_type>
+  class IndexAccumulator_Checked : public IndexAccumulator<_N,_Nd_type>{
+  public:
+    static constexpr size_t N = _N;
```
pcarruscag, line 233: I understand the motivation, but this duplication is probably not justified.
Add size to the "normal IndexAccumulator" since that would be useful to allow range-based loops.
Always assert(i<=size) (since it goes away in optimized builds).
Add a bool template param (i.e. Check) and if Check is true you do the MPI:Error thing.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    /*! \brief Read one more index, checking whether it is in the range dictated by the NdFlattener and
+     * previous indices.
+     * \param[in] i - Index.
+     */
+    LookupType operator[] (typename Nd_type::Index i) {
+      if(i>=_size){
+        SU2_MPI::Error("NdFlattener: Index out of range.", CURRENT_FUNCTION);
+      }
+      const typename Nd_type::Index new_offset = this->nd->GetIndices()[this->offset+i];
+      const typename Nd_type::Index new_size = this->nd->GetIndices()[this->offset+i+1] - new_offset;
```
pcarruscag, line 257: auto all you can (we don't want people saying templates are hard to read)

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    }
+
+    typename Nd_type::Index size() const {
+      return _size;
+    }
+  };
+
+} // namespace helpers
+
+inline helpers::NdFlattener_MPI_Environment<decltype(&(SU2_MPI::Allgatherv)), decltype(MPI_INT), decltype(SU2_MPI::GetComm())>
```
pcarruscag, line 336: using Something = NdFlattener_MPI_Environment<decltype(&(SU2_MPI::Allgatherv)), decltype(MPI_INT), decltype(SU2_MPI::GetComm())>
To avoid repeating the shopping list.

Comm and Type have typedefs in SU2_MPI, no need for decltype.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+protected:
+  /*! \brief Determine the space required for reading the 'recursive function'.
+   *
+   * \param f - the 'recursive function'
+   */
+  template<class f_type>
+  void count_f(f_type f) {
+    Index nChild = f.first;
+    for(Index iChild=0; iChild<nChild; iChild++){
+      nNodes++;
```
pcarruscag, line 522: nNodes = nChild ?

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    for(Index iChild=0; iChild<nChild; iChild++){
+      Base::set_f(f.second(iChild), refresh);
+      if(!refresh){
+        indices[iNode+1] = indices[iNode] + f.second(iChild).first;
+      } else {
+        if( indices[iNode+1] != indices[iNode] + f.second(iChild).first ){
+          SU2_MPI::Error("NdFlattener: Structure has changed, cannot refresh.", CURRENT_FUNCTION);
+        }
+      }
+      iNode++;
```
pcarruscag, line 546: iNode = iChild ?

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    Index** Nodes_all = new Index*[K]; // [k][r] is number of all nodes in layer (k+1), rank r in the new structure
+    for(size_t k=0; k<K; k++)
+      Nodes_all[k] = nullptr;
+    Nodes_all[K-1] = new Index[mpi_env.size]; // {1, 1, ..., 1}
+    int* displs = new int[mpi_env.size]; // {0, 1, ..., size-1}
+    int* ones = new int[mpi_env.size]; // {1,1, ...}
+    for(int r=0; r<mpi_env.size; r++){
+      nNodes += Nodes_all[K-1][r] = 1;
+      displs[r] = r;
+      ones[r] = 1;
```
pcarruscag, line 575: Please use vector for these as well, it makes it so much cleaner to initialize.

Nodes_all is rectangular... please use a matrix type

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+   *
+   * The new first index will encode the rank of the process. Data is exchanged in MPI::Allgatherv-style
+   * collective communication.
+   * \param[in] mpi_env - The MPI environment used for communication.
+   * \param[in] local_version - The local NdFlattener structure with (K-1) indices.
+   */
+  template<typename MPI_Environment_type>
+  void initialize( 
+    MPI_Environment_type const& mpi_env,
+    Base const* local_version
```
pcarruscag, line 564: Reference instead of pointer?

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+  }
+  /*! \brief Data look-up, non-const version.
+   */
+  Data& operator[](Index i0) {
+    return data[i0];
+  }
+  /*! \brief Data look-up, const version.
+   */
+  const Data& operator[](Index i0) const {
+    return data[0];
```
pcarruscag, line 829: looks suspicious

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    // gather numbers of nodes in the current layer from all processes
+    mpi_env.MPI_Allgatherv( &(local_version->nNodes), 1, mpi_env.mpi_index, Nodes_all[0], ones, displs, mpi_env.mpi_index, mpi_env.comm );
+    for(int r=0; r<mpi_env.size; r++){
+      nNodes += Nodes_all[0][r];
+    }
+  }
+
+  template<typename MPI_Environment_type>
+  void set_g(MPI_Environment_type const& mpi_env,
+         Index** Nodes_all,
```
pcarruscag, line 795: const ?

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+
+    int* Nodes_all_0_as_int = new int[mpi_env.size];
+    int* Nodes_all_0_cumulated = new int[mpi_env.size+1];
+    Nodes_all_0_cumulated[0] = 0;
+    for(int r=0; r<mpi_env.size; r++){
+      Nodes_all_0_cumulated[r+1] = Nodes_all_0_cumulated[r] + Nodes_all[0][r];
+      Nodes_all_0_as_int[r] = Nodes_all[0][r];
+    }
+
+    mpi_env.MPI_Allgatherv( local_version->data.data(), Nodes_all[0][mpi_env.rank], mpi_env.mpi_data, data.data(), Nodes_all_0_as_int, Nodes_all_0_cumulated, mpi_env.mpi_data, mpi_env.comm );
```
pcarruscag, line 808: please run clang_format on this file once you are done.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+  template<typename MPI_Environment_type>
+  void set_g(MPI_Environment_type const& mpi_env,
+         Index** Nodes_all,
+         CurrentLayer const* local_version )
+  { 
+
+    int* Nodes_all_K_as_int = new int[mpi_env.size];
+    int* Nodes_all_k_cumulated = new int[mpi_env.size+1]; // [r] is number of nodes in the current layer, summed over all processes with rank below r 
+    // plus one. Used as displacements in Allgatherv, but we do not want to transfer the initial zeros and rather the last element of indices, 
+    // which is the local nNodes of the layer below. Note that MPI needs indices of type 'int'.
```
pcarruscag, line 665: basically the counts and displacements of Allgatherv

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+  void count_g(MPI_Environment_type const& mpi_env,
+         Index** Nodes_all,
+         CurrentLayer const* local_version,
+         int const* displs, int const* ones )
+  { 
+    assert( Nodes_all[K-1]==nullptr);
+    Nodes_all[K-1] = new Index[mpi_env.size];
+    nNodes = 0;
+    // gather numbers of nodes in the current layer from all processes
+    mpi_env.MPI_Allgatherv( &(local_version->nNodes), 1, mpi_env.mpi_index, Nodes_all[K-1], ones, displs, mpi_env.mpi_index, mpi_env.comm );
```
pcarruscag, line 643: Why not use MPI_Allgather and avoid the boilerplate "ones" and "displs"?

pcarruscag (commented):
Alright, another batch of comments, it looks like there is some room for simplification.

Common/src/geometry/CGeometry.cpp
```diff
@@ -3942,5 +3942,26 @@ void CGeometry::ComputeWallDistance(const CConfig* const* config_container, CGeo
...
+      auto roughness_f =
+        make_pair( nZone, [config_container,geometry_container,iInst](unsigned long iZone){
+          const CConfig* config = config_container[iZone];
+          return make_pair( geometry_container[iZone][iInst][MESH_0]->GetnMarker() , [config](unsigned long iMarker){
+            return config->GetWallRoughnessProperties(config->GetMarker_All_TagBound(iMarker)).second;
+          });
+        });
+      NdFlattener<2,su2double,unsigned long> roughness_local(roughness_f);
+      // [rank][iZone][iMarker] -> roughness
+      NdFlattener<3,su2double,unsigned long> roughness_global(Get_Nd_MPI_Env(), &(roughness_local));
```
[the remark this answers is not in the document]
maxaehle: yep

Common/include/fem/fem_geometry_structure.hpp
```diff
@@ -1227,11 +1227,14 @@ class CMeshFEM_DG: public CMeshFEM {
...
-   * \param[in] WallADT - The ADT to compute the wall distance
+   * \brief Reduce the wall distance based on an previously constructed ADT.
+   * \details The ADT might belong to another zone, giving rise to lower wall distances
+   * than those already stored.
+   * \param[in] WallADT - The ADT to reduce the wall distance
+   * \param[in] config - Config of this geometry (not the ADT zone's geometry)
+   * \param[in] iZone - ignored
    */
-  void SetWallDistance(const CConfig *config, CADTElemClass* WallADT) override;
+  void SetWallDistance(CADTElemClass* WallADT, const CConfig* config, unsigned short iZone=numeric_limits<unsigned short>::max()) override;
```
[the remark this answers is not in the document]
maxaehle: ok

Common/include/geometry/dual_grid/CPoint.hpp
```diff
@@ -85,7 +86,14 @@ class CPoint {
   su2vector<bool> Agglomerate;                 /*!< \brief This flag indicates if the element has been agglomerated. */
 
   su2vector<unsigned short> nNeighbor;    /*!< \brief Number of neighbors, needed by some numerical methods. */
+
+  /*--- Closest element on a viscous wall, and distance to it. ---*/
   su2activevector Wall_Distance;          /*!< \brief Distance to the nearest wall. */
+  su2vector<int> ClosestWall_Rank; /*!< \brief Rank of process holding the closest wall element. */
+  su2vector<unsigned short> ClosestWall_Zone; /*!< \brief Zone index of closest wall element. */
+  su2vector<unsigned short> ClosestWall_Marker; /*!< \brief Marker index of closest wall element, for given rank and zone index. */
+  su2vector<unsigned long> ClosestWall_Elem; /*!< \brief Element index of closest wall element, for givenrank, zone and marker index. */
```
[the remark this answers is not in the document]
maxaehle: Yes, go for it

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+ */
+
+
+template<size_t K, typename Data=su2double, typename Index=unsigned long>
+class NdFlattener;
+
+namespace helpers {
+  template<typename MPI_Allgatherv_type, typename MPI_Datatype_type, typename MPI_Communicator_type>
+  struct NdFlattener_MPI_Environment {
+    MPI_Allgatherv_type MPI_Allgatherv;
```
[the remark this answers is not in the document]
maxaehle: I changed it to `MPI_Allgatherv_fun`

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+   *  - The case N==2 is much like the case N>3 but an additional function data() should be provided.
+   *  - For N>3, more indices have to be read, return an IndexAccumulator<N-1>.
+   * \tparam N - Number of missing parameters
+   * \tparam Nd_type - Type of the accessed NdFlattener, should be NdFlattener<N,...>
+   */
+  /*! \class IndexAccumulator_Base
+   * \brief Parent class of IndexAccumulator.
+   * \details IndexAccumulator provides the operator[] method.
+   */
+  template<size_t _N, typename _Nd_type>
```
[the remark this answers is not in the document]
maxaehle: Replaced \_TemplateParam by TemplateParam\_ and Some_type by Some_t

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+      nd(nd), offset(offset) {}
+
+  };
+
+  template<size_t _N, typename _Nd_type>
+  class IndexAccumulator : public IndexAccumulator_Base<_N,_Nd_type>{
+  public:
+    static constexpr size_t N = _N;
+    using Nd_type = _Nd_type;
+    typedef IndexAccumulator_Base<_N,_Nd_type> Base;
```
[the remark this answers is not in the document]
maxaehle: I substituted it

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+
+    /*== Conditional is part of the standard library since C++14, keep it here for C++11 compatibility. ==*/
+    template<bool B, class T, class F> struct conditional { typedef T type; };
+    template<class T, class F> struct conditional<false, T, F> { typedef F type; };
+
+  protected:
+    Nd_type* nd; /*!< \brief The accessed NdFlattener. */
+    const typename Nd_type::Index offset; /*!< \brief Index in the currently accessed layer. */
+    //const typename Nd_type::Index size; /*!< \brief Exclusive upper bound for the next index. */
+    IndexAccumulator_Base(Nd_type* nd, typename Nd_type::Index offset):
```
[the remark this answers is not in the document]
maxaehle: Ok. Now it keeps `IndexAccumulator_Base::nd` by reference instead of pointer. This does also apply to the `local_version` for `NdFlattener` construction via collective communication.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+   */
+  template<size_t _N, typename _Nd_type>
+  class IndexAccumulator_Base {
+  public:
+    static constexpr size_t N = _N;
+    using Nd_type = _Nd_type;
+
+    /*== Conditional is part of the standard library since C++14, keep it here for C++11 compatibility. ==*/
+    template<bool B, class T, class F> struct conditional { typedef T type; };
+    template<class T, class F> struct conditional<false, T, F> { typedef F type; };
```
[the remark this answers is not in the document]
maxaehle: done

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+    /*! Return type of operator[]. */
+    using LookupType = IndexAccumulator<N-1, Nd_Base_type>;
+
+    /*! \brief Read one more index.
+     * \param[in] i - Index.
+     */
+    LookupType operator[] (typename Nd_type::Index i) {
+      //const typename Nd_type::Index new_offset = this->nd->GetIndices()[this->offset+i];
+      //const typename Nd_type::Index new_size = this->nd->GetIndices()[this->offset+i+1] - new_offset;
+      return LookupType(static_cast<Nd_Base_type*>(this->nd),this->nd->GetIndices()[this->offset+i]);
```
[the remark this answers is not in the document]
maxaehle: yes, this makes it a lot easier to read

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,846 @@
...
+      const typename Nd_type::Base,
+      typename Nd_type::Base
+    >::type;
+    /*! Return type of operator[]. */
+    using LookupType = IndexAccumulator<N-1, Nd_Base_type>;
+
+    /*! \brief Read one more index.
+     * \param[in] i - Index.
+     */
+    LookupType operator[] (typename Nd_type::Index i) {
```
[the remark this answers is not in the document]
maxaehle: Actually, it was possible to make `IndexAccumulator<N,...>::operator[]` and also `IndexAccumulator<1,...>::data` always const methods. The accumulator just knows how to access an `NdFlattener` (and which one), and this information does not change along with the access.  
The exact type of the `NdFlattener`, including a `const`-qualifier, is a template argument to `IndexAccumulator`. This constness is preserved by `operator[]` and eventually used to determine whether the reference/pointer to the data should be const.

Common/include/toolboxes/ndflattener.hpp
maxaehle: Ok, I "merged" the IndexAccumulator_Checked into IndexAccumulator and IndexAccumulator_Base

Common/include/toolboxes/ndflattener.hpp
maxaehle: Now with the `typename Nd_type::` and `this->` gone, it's readable and I left it.

Common/include/toolboxes/ndflattener.hpp
maxaehle: I decided to remove those template arguments completely, in the context of SU2 there is no need for more generality.

Common/include/toolboxes/ndflattener.hpp
maxaehle: No. `nNodes` is the length of the indices (or data) array (+-1), `nChild` is the number of indices (or data elements) "below" a particular partial index tuple. The former is the sum of the latter, so to speak.

Common/include/toolboxes/ndflattener.hpp
maxaehle: No, for the same reason as above. `iNode` is set to 0 only when initializing or refreshing the NdFlattener. Whenever in this depth-first search you descent to the first child node, the `iChild` in some layer is set to 0.

Common/include/toolboxes/ndflattener.hpp
maxaehle: done

Common/include/toolboxes/ndflattener.hpp
maxaehle: Yes, this was possible

Common/include/toolboxes/ndflattener.hpp
maxaehle: Good spot! There were no tests for `NdFlattener<1>`, let alone for `const NdFlattener`. So I extended the test set a bit.

Common/include/toolboxes/ndflattener.hpp
maxaehle: Yes, now it's a `su2matrix<Index_t> const& Nodes_all`

Common/include/toolboxes/ndflattener.hpp
maxaehle: done

Common/include/toolboxes/ndflattener.hpp
maxaehle: yes

Common/include/toolboxes/ndflattener.hpp
maxaehle: Good idea, I changed it

maxaehle:
Thanks @pcarruscag for your comments! I have now incorporated most of them. What is missing from the first batch is a good way of initialization instead of the recursive lambda expressions, because I'm lacking completely convincing ideas for this. 

The most comfortable interface I can think of is a function
- whose parameters are all indices at once, and
- which returns a pair/structure of one `size_t` and one `Data_t`.
- If the index tuple is valid (in range), the `size_t` component is arbitrary and the `Data_t` component is the data element.
- If the index tuple is invalid, the `size_t` component specifies the first out-of-range index and the `Data_t` component is arbitrary.

So the lambda expression would look like this:

    [captures](unsigned long i1, ..., unsigned long iK){
      if( i1 >= ... ) return Nd_Range(1);
      if( i2 >= ... ) return Nd_Range(2);
      ...
      return Nd_Value( ... );
    }

Do you consider this interface sufficiently comfortable? Anyway I'll have to think about how to implement this...

maxaehle:
I'm happy that the strange segmentation fault in the discadj_heat testcase in `serial_regression_AD.py` seems to be gone ;)

Common/include/toolboxes/ndflattener.hpp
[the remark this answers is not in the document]
pcarruscag: Great, less chance for dangling pointer issues.

pcarruscag:
I don't mind the "pairs" by themselves, it fits nicely with how the NdFlattener works, i.e. its efficient to split them and pass them down to lower layers.
The only slight problem is that "first" and "second" are not a super expressive interface, but I think that is ok for now.
If it ever becomes needed you can create some adapter to convert something with operator (i,j,....) into pairs of pairs behind the scenes, but for the current use that would not be an option anyway.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,751 @@
...
+
+template <size_t K, typename Data_t_ = su2double, typename Index_t_ = unsigned long>
+class NdFlattener;
+
+/*! \struct Nd_MPI_Environment
+ * \brief Contains information for the collective communication of NdFlatteners.
+ *
+ * The default arguments in the constructor are chosen in a sensible way:
+ * - To communicate su2double data, just call ND_MPI_Environment().
+ * - To communicate unsigned long data, call Nd_MPI_Environment(MPI_UNSIGNED_LONG).
```
pcarruscag, line 161: Nice :+1:

Common/include/toolboxes/ndflattener.hpp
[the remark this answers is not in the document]
pcarruscag: I get that but if [ ] is const this should const-qualify the nd& kept by base which should then only be able to return const&... :thinking:
Have you tried modifying data via the IndexAccumulator?

UnitTests/Common/toolboxes/ndflattener_tests.cpp
```diff
@@ -0,0 +1,140 @@
...
+  REQUIRE( nd2.size() == 2 );
+  REQUIRE( nd2[0][0] == 0.0 );
+  REQUIRE( nd2[0][1] == 1.0 );
+  REQUIRE( nd2[1].size() == 3 + rank );
+  for(int i=0; i<3+rank; i++){
+    REQUIRE( nd2[1][i] == 2.0 + rank + i );
+  }
+
+  /*-- Modify structure. --*/
+  nd2[0][0] = 0.7;
```
pcarruscag, line 67: you have, nevermind then

Common/include/toolboxes/ndflattener.hpp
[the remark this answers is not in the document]
pcarruscag: const method qualifies "this" but not the member, that is why it works...
However, this would allow a const IndexAccumulator to be changed https://gcc.godbolt.org/z/chvfrhEc6.

Common/include/toolboxes/ndflattener.hpp
pcarruscag: Ok it is only equal for the top level.

Common/include/toolboxes/ndflattener.hpp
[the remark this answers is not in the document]
maxaehle: This is how I understand what is happening:

In a const method `X::f()`, `this` is not `X*` but `X const*`, making any member read-only.

Anyway, a reference member, like `IndexAccumulator<N,Nd_t>::nd` of type `Nd_t&`, is always read-only in the sense that it cannot be reassigned to point to another variable. Whether or not the variable to which it points can be changed, does not depend on the constness of `*this`. It depends only on whether the type `Nd_t` is const, which can be found out with `std::is_const<Nd_t>::value`.

If the `IndexAccumulator` was generated by the `operator[]` from a non-const `NdFlattener`, it's an (non-const) `IndexAccumulator<N, NdFlattener<...> >` so its member `nd` is a reference to a non-const `NdFlattener`, so after supplying enough indices we get a reference/pointer to the `IndexAccumulator<1, NdFlattener<...> >::LookupType` aka `Data_t`.

If the `IndexAccumulator` was generated by the `operator[] const` from a const `NdFlattener`, it's an (non-const) `IndexAccumulator<N, const NdFlattener<...> >` so its member `nd` is a reference to a `const NdFlattener`, so after supplying enough indices we get a reference/pointer to the `IndexAccumulator<1, const NdFlattener<...> >::LookupType` aka `const Data_t`.

To simulate this in the godbolt example, replacing `vector<int>` by `const vector<int>` raises a compilation error "assignment of read-only location". In `ndflattener_tests.cpp`, I tried writing to `nd3_const[0][0][0]` and it did not compile. I don't know whether we can make a unit test for "does not compile".

UnitTests/Common/toolboxes/ndflattener_tests.cpp
```diff
@@ -51,6 +56,8 @@ TEST_CASE("NdFlattener Test", "[NdFlattener]"){
   NdFlattener<2> nd2;
   nd2.initialize_or_refresh(f);
 
+  setIntoConstObject(nd2[0]);
```
maxaehle, line 59: Do you mean this:

    const NdFlattener<2>& nd2_const = nd2;
    setIntoConstObject(nd2_const[0]);

pcarruscag: No, I mean exactly what I wrote.
If you have any object and you pass it to a function by const& that function cannot be able to change it.
You need const and non-const versions of operator [] in the Accumulator class.

maxaehle: The statement `obj[0] = 1;` does not change the const `IndexAccumulator obj`, it changes the non-const `NdFlattener obj.nd`. So IMO it is perfectly fine that the code compiles. If `nd2` was const, the code would not compile. This is precisely the behaviour that I intended.

Are you saying that this behaviour is not good? I.e., that a function with a const argument `obj` that implements the [] interface should never be allowed to do `obj[0]=1;`, no matter what `operator[]` actually does? 

If yes, I'm fine with changing this. As you say, essentially there should be both methods `operator[]`, `operator[] const` returning `LookupType&` and `LookupType const&`, respectively (and the same for `data`). But let's discuss this in the developer meeting tomorrow.

pcarruscag: To be clear, I think that behavior is unsafe and incorrect.
Constness should be derived from current context and not from how the object was created initially.
If a function is allowed to modify the state of something that is supposed to be read only, it is not const-correct and it is very difficult to guarantee thread safety.
Please change it so we can move on, enough time has been put into this already.

pcarruscag (approved):

UnitTests/Common/toolboxes/ndflattener_tests.cpp
maxaehle: Ok. f0e16b0 failed to build, as expected. I removed the forbidden write access in 722a4dc.

You're mentioning thread safety, should we put `SU2_OMP_MASTER { ... } END_SU2_OMP_MASTER` around the code in `initialize_or_refresh`?

pcarruscag: Thanks. No we leave that for the users of the class otherwise thread local objects would not be possible.

pcarruscag (approved):
The UB sanitizer had nothing to say, so I think the cases that changed might just be due to solar flares or something.

TobiKattmann (commented):
Alright @maxaehle, thanks for this escalation of a bug fix 💐 
I really did not understand much of the implementation details tbh so my question are more of a general nature. That said, the implementation looks fine otherwise and I appreciate the extensive commenting. I forego the approval because I simply feel I didn't understand enough to do so.

You have a `Usage` section and `Implementation details` in the head documentation of ndflattener.hpp and I would love to see a `When/why to use | What is my purpose` section. So as far as I understand the ndflattener can be used to (deep)copy specified data that is behind multidimensional (pointer-to-pointer-... ) access (e.g. rank,zone,marker) into a new datastructure that inherits the same access pattern to the outside and internally is a set of a 1D arrays. The data to be copied might be part a small part of a larger class like geometry so we would also  "extract" the data from such (more complicated, and less contiguous) structure while keeping the multidimensional acces logic. 

So it is always useful when i just want to use very specific data which is 'hidden' in a larger class/dataset but need to keep the accessing logic. I can extract the data with the ndflattener at a top level and pass just this data I need the a method which needs it. Like that I dont have to pass all kinds of class pointers down.

let me know it any of this makes sense 🐌

Common/include/geometry/dual_grid/CPoint.hpp
```diff
@@ -857,4 +877,19 @@ class CPoint {
...
+   * \brief Set wall roughnesses according to stored closest wall information.
+   * \param[in] roughness - Mapping [rank][zone][marker] -> roughness
+   */
+  template<typename Roughness_type>
+  void SetWallRoughness(Roughness_type const& roughness){
+    for (unsigned long iPoint=0; iPoint<GlobalIndex.size(); ++iPoint) {
+      auto rankID = ClosestWall_Rank[iPoint];
+      auto zoneID = ClosestWall_Zone[iPoint];
+      auto markerID = ClosestWall_Marker[iPoint];
+      if(rankID >= 0){
```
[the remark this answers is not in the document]
maxaehle: It's mainly safety. The `ClosestWall_Rank` is initialized by -1. If there is no viscous wall, it will stay -1. Right now this code is only called from `CGeometry::ComputeWallDistance`, and only if there is a viscous wall, so there is no immediate danger IMO. But I think we should keep it safe.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,793 @@
...
+ * You can access the NdFlattener via a "[]...[]" interface that resembles multidimensional arrays
+ * or pointer-to-pointer-... arrays. If you provide all of the K indices, this returns a reference
+ * to the data element. If the NdFlattener is declared const, the reference is also const. If you
+ * provide less than K indices, you obtain an IndexAccumulator object to which you can later pass more
+ * indices. With respect to the above example:
+ *
+ *     std::cout << nd_global[rank][zone][marker];
+ *     auto nd_g_rank = nd_global[rank];
+ *     nd_g_rank[zone][marker] += 1.0;
+ *
```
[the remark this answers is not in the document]
maxaehle: I added a second output line with example value. This does also demonstrate that writing to the IndexAccumulator does not invalidate it (in case somebody asks).

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -802,6 +802,8 @@ class CNumerics {
    * \param[in] val_dist_j - Value of of the roughness of the nearest wall from point j
    */
   void SetRoughness(su2double val_roughness_i, su2double val_roughness_j) {
+    if(val_roughness_i!=0.0) SU2_MPI::Error("val_roughness_i!=0.0",CURRENT_FUNCTION);
+    if(val_roughness_j!=0.0) SU2_MPI::Error("val_roughness_j!=0.0",CURRENT_FUNCTION);
     roughness_i = val_roughness_i;
     roughness_j = val_roughness_j;
   }
```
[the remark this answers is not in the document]
maxaehle: This was just a temporary addition in 1244638 to see whether it works as intended ;) There is no roughness in the regression tests yet, so these errors should never occur; yet, they do. I will further investigate this and report in the main discussion.

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -0,0 +1,828 @@
...
+   * \returns true if the NdFlattener has been initialized
+   */
+  bool initialized() { return nNodes > 0; }
+
+ protected:
+  /*! \brief Allocate the indices array after \a nNodes has been determined.
+   */
+  void allocate() {
+    indices.reserve(nNodes + 1);
+    indices[0] = 0;
```
pcarruscag, line 505: I believe this use of vector::reserve is not correct, reserve changes the capacity of the vector but not its size.
Technically indices[0] is out of bounds (not in memory but in "logic").
Whether this is the problem or not IDK.

Common/include/geometry/dual_grid/CPoint.hpp
```diff
@@ -88,7 +89,14 @@ class CPoint {
   su2vector<bool> Agglomerate;                 /*!< \brief This flag indicates if the element has been agglomerated. */
 
   su2vector<unsigned short> nNeighbor;    /*!< \brief Number of neighbors, needed by some numerical methods. */
+
+  /*--- Closest element on a viscous wall, and distance to it. ---*/
   su2activevector Wall_Distance;          /*!< \brief Distance to the nearest wall. */
+  su2vector<int> ClosestWall_Rank; /*!< \brief Rank of process holding the closest wall element. */
+  su2vector<unsigned short> ClosestWall_Zone; /*!< \brief Zone index of closest wall element. */
```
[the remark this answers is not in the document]
maxaehle: The code cites
 > Aupoix, B. and Spalart, P. R., "Extensions of the Spalart-Allmaras Turbulence Model to Account for Wall Roughness,"
 > International Journal of Heat and Fluid Flow, Vol. 24, 2003, pp. 454-462.
 > See https://turbmodels.larc.nasa.gov/spalart.html#sarough for detailed explanation.

The TMR page says that you require the " conventional Nikuradse sand roughness scale height" at the wall. In the old implementation each marker is allowed to have its specific roughness, and for integration at one point, the roughness of the closest marker is chosen. As I just wanted to fix the bug and I'm not familiar with the theory, I kept this level of generality.

I encountered the bug in setup with two fluid zones connected by a sliding mesh interface. 

The point is that in `CPhysicalGeometry::SetWallDistance`, the call to `WallADT::DetermineNearestElement` gives the closest wall marker in the `iZone` (which is not the zone in which `this CPhysicalGeometry` lives in, generally). It is therefore wrong to access the wall roughness array of `this CPhysicalGeometry` with the marker ID that belongs to a different zone. In my case, valgrind reported an invalid read, because there were different numbers of markers in both zones.

Maybe the dependency on the rank is irrelevant, I don't know if all markers have the same marker IDs in all ranks... I don't care ;)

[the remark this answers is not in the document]
pcarruscag: They don't. iMarker only makes sense locally and different ranks may have different boundaries.
If you try to do any MPI as part of a boundary condition you will have a bad time.

SU2_CFD/include/numerics/CNumerics.hpp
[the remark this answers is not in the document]
TobiKattmann: ok, so just temporary debugging output/error 👍

Common/include/toolboxes/ndflattener.hpp
```diff
@@ -37,8 +37,40 @@
...
+/*! \page ndflattener_usage NdFlattener usage
+ *
+ * # What is an NdFlattener?
+ *
+ * NdFlattener is a container to store a deep copy of multidimensional data in a single object that provides a
+ * generic []...[] interface. The data and their structure are stored in few contiguous arrays, allowing for an
+ * efficient MPI gathering operation.
+ *
+ * # Which problem does it solve?
+ * In SU2, there is a lot of multidimensional data: data depending on K>1 indices, like a wall property
```
TobiKattmann, line 49: Thanks for added parts 👍 
A lot of work for the wall roughness feature that is not used (to my knowledge) but hopefully finds some good usage in the future 🔮 it certainly is one the better documented code bits :)

maxaehle:
8372309 fixed the `MPI_Allgatherv` function used in a non-MPI build. Before, it ignored the displacement (which is 1 for the index arrays). This was revealed by experiments with the CI pipeline in #1292. Now the tests succeed.

PR #1294, opened 2021-06-01
Hybrid Parallel AD (Part 3/?)

jblueh:
## Proposed Changes
1. Fix preaccumulation issues with `RealReverseIndex`.
2. ~~Fix~~ Identify issues regarding index reuse and partial tape evaluations.
3. Add regression tests for hybrid parallel AD (subset of parallel_regression_AD).
4. Re-enable parallel preaccumulation.
5. Shared reading optimizations.

## Related Work
continues #1284

## PR Checklist


TobiKattmann:
Part 3, but the branch-name is is `hybrid_parallel_ad4` ... who are you trying to fool?

(jk :D , that mismatch was already in the previous PR)

jblueh:
f501dc168cd7ef4a4d1f81140a0959a9d4382128 fixes divergence of the `disc_adj_fsi` case among the parallel regression tests that occurred when using the CoDiPack type `RealReverseIndex` instead of `RealReverse`.

Since the bug is not obvious from the commit and there might be a better way to fix this, let me put this up for discussion.

In `CDiscAdjMultizoneDriver::SetRecording`, the tape is recorded part by part. The part of the code that is at fault involves three of them.
- `DEPENDENCIES -> OBJECTIVE_FUNCTION`: an index is last used on a right hand side
- `OBJECTIVE_FUNCTION -> TRANSFER`: the index is available for reuse and used on the left hand side
- `recording for a zone`: the index is used on a right hand side

In `CDiscAdjMultizoneDriver.cpp::ComputeAdjoints`, however, the `TRANSFER -> OBJECTIVE_FUNCTION` part of the evaluation is skipped in some cases. After the `recording for a zone` evaluation accumulated adjoints for the said index, the adjoint reset that would be performed in the skipped tape part is missing. Instead, the tape evaluation `DEPENDENCIES -> OBJECTIVE_FUNCTION` receives the adjoints for this index, which is wrong because the index is associated with a different primal variable in this part of the tape.

Have there been issues regarding dependencies between tape parts previously? Are there mechanisms established that address such issues?

If someone wants to give it a try, I added a build option for the tape type in 183c3ca7a3235c8293cf2242c66c82d35a3922c7. You can switch to the index tape by calling meson with `-Dcodi-tape=JacobianIndex`.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -898,8 +898,12 @@ void CDiscAdjMultizoneDriver::ComputeAdjoints(unsigned short iZone, bool eval_tr
    *    on the last inner iteration. Structural problems have some minor issue and we
    *    need to evaluate this section on every iteration. ---*/
```
jblueh, line 899: Can someone comment on the "minor issue" that prevented skipping parts of the evaluation already when using a linear tape?

pcarruscag: Right, that part of the tape is skipped for inner iterations because it is not needed but it is expensive when there is mesh deformation (much more expensive that one fluid iteration).
However... Doing that skipping for structural problems results in a spike of the residuals on the last iteration (where the transfer section is evaluated). The derivatives do not change but the change in residuals was annoying to monitor convergence.
Because the structural problem does not have mesh deformation, it was "ok" to always evaluate the transfer to avoid that increase in residuals :shrug:

meson.build
```diff
@@ -53,8 +53,18 @@ endif
...
   codi_dep = [declare_dependency(include_directories: 'externals/codi/include')]
-  codi_rev_args = '-DCODI_REVERSE_TYPE'
-  codi_for_args = '-DCODI_FORWARD_TYPE'
+  codi_rev_args = ['-DCODI_REVERSE_TYPE']
+  codi_for_args = ['-DCODI_FORWARD_TYPE']
+endif
+
+if get_option('enable-autodiff')
+  if get_option('codi-tape') == 'JacobianIndex'
+    codi_rev_args += '-DCODI_INDEX_TAPE'
```
pcarruscag, line 62: :+1:

jblueh:
Parallel preaccumulations must not share any variables. In particular, they must not have any common inputs and outputs. I suspect (judging from the code) that the preaccumulations treated in 781092a3533caf36589426e99293c2b20d332726 violate this pattern, and that this is the reason why excluding them makes several hybrid parallel AD tests pass. Other preaccumulations might be affected as well. The correctness of preaccumulations inside numerics objects, for example, depends on the values of pointers like `V_i`, `V_j`. Other preaccumulations might look wrong but are ok because common variables are prevented by color loops.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
jblueh: Then this performance optimization should be kept. While the change fixes the `disc_adj_fsi` test case, it does not close all gaps in tape evaluations and the principle problem remains. For now, I will `assert` that index AD types do not support multiple zones.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1535,6 +1539,10 @@ void CFVMFlowSolverBase<V, R>::EdgeFluxResidual(const CGeometry *geometry,
     InstantiateEdgeNumerics(solvers, config);
   }
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+#endif
```
pcarruscag, line 1544: We will do it like this to avoid different logic between primal and adjoint.
I have updated the warning message to reflect the higher memory usage in case the "ReducerStrategy" is used.
The places I modified now are also the places where the adjoint evaluation mode that does not use atomics could in theory be enabled.

jblueh: Don't you find short-circuiting hard to read?

jblueh:
The computed values of `discadj_arina2k` in `hybrid_regression_AD` are suspiciously close to the expected values in `parallel_regression_AD`. Could the reference values be wrong? If I test with `-Denable-mixedprec=false` locally, the same holds true for `discadj_fea`.

Common/include/basic_types/ad_structure.hpp
```diff
@@ -262,6 +262,18 @@ namespace AD{
...
+   * \brief Pause the use of preaccumulation.
+   * \return True if preaccumulation was active.
+   */
+  inline bool PausePreaccumulation() { return false; }
+
+  /*!
+   * \brief Resume the use of preaccumulation.
+   * \param[in] wasActive - Whether preaccumulation was active before pausing.
+   */
+  inline void ResumePreaccumulation(bool wasActive) {}
```
pcarruscag, line 275:
```suggestion
  inline void ResumePreaccumulation(bool wasActive) {}

  /*!
   * \brief Begin a hybrid parallel adjoint evaluation mode that assumes an inherently safe reverse path.
   */
  inline void StartNoSharedReading() {}

  /*!
   * \brief End the "no shared reading" adjoint evaluation mode.
   */
  inline void EndNoSharedReading() {}
```

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7701,7 +7701,9 @@ void CPhysicalGeometry::SetBoundControlVolume(const CConfig *config, unsigned sh
 
       const auto nNodes = bound[iMarker][iElem]->GetnNodes();
 
+#ifndef HAVE_OPDI
       AD::StartPreacc();
+#endif
```
pcarruscag, line 7706:
```suggestion
      /*--- Cannot preaccumulate if hybrid parallel due to shared reading. ---*/
      if (omp_get_num_threads() == 1) AD::StartPreacc();
```

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -76,7 +76,9 @@ void computeGradientsGreenGauss(CSolver* solver,
   {
     auto nodes = geometry.nodes;
 
+#ifndef HAVE_OPDI
     AD::StartPreacc();
+#endif
```
pcarruscag, line 81:
```suggestion
    /*--- Cannot preaccumulate if hybrid parallel due to shared reading. ---*/
    if (omp_get_num_threads() == 1) AD::StartPreacc();
```

SU2_CFD/include/gradients/computeGradientsLeastSquares.hpp
```diff
@@ -203,7 +203,9 @@ void computeGradientsLeastSquares(CSolver* solver,
     auto nodes = geometry.nodes;
     const auto coord_i = nodes->GetCoord(iPoint);
 
+#ifndef HAVE_OPDI
     AD::StartPreacc();
+#endif
```
pcarruscag, line 208:
```suggestion
    /*--- Cannot preaccumulate if hybrid parallel due to shared reading. ---*/
    if (omp_get_num_threads() == 1) AD::StartPreacc();
```

SU2_CFD/include/limiters/computeLimiters_impl.hpp
```diff
@@ -132,7 +132,9 @@ void computeLimiters_impl(CSolver* solver,
     auto nodes = geometry.nodes;
     const auto coord_i = nodes->GetCoord(iPoint);
 
+#ifndef HAVE_OPDI
     AD::StartPreacc();
+#endif
```
pcarruscag, line 137:
```suggestion
    /*--- Cannot preaccumulate if hybrid parallel due to shared reading. ---*/
    if (omp_get_num_threads() == 1) AD::StartPreacc();
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1535,6 +1539,12 @@ void CFVMFlowSolverBase<V, R>::EdgeFluxResidual(const CGeometry *geometry,
     InstantiateEdgeNumerics(solvers, config);
   }
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+  if (!ReducerStrategy)
+    AD::StartNoSharedReading();
+#endif
```
pcarruscag, line 1546:
```suggestion
  /*--- For hybrid parallel AD, pause preaccumulation if there is shared reading of
  * variables, otherwise switch to the faster adjoint evaluation mode. ---*/
  bool pausePreacc = false;
  if (ReducerStrategy) pausePreacc = AD::PausePreaccumulation();
  else AD::StartNoSharedReading();
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1992,6 +1992,12 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
   su2double Primitive_i[MAXNVAR] = {0.0}, Primitive_j[MAXNVAR] = {0.0};
   su2double Secondary_i[MAXNVAR] = {0.0}, Secondary_j[MAXNVAR] = {0.0};
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+  if (!ReducerStrategy)
+    AD::StartNoSharedReading();
+#endif
```
pcarruscag, line 1999:
```suggestion
  /*--- For hybrid parallel AD, pause preaccumulation if there is shared reading of
  * variables, otherwise switch to the faster adjoint evaluation mode. ---*/
  bool pausePreacc = false;
  if (ReducerStrategy) pausePreacc = AD::PausePreaccumulation();
  else AD::StartNoSharedReading();
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1018,6 +1018,12 @@ void CIncEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_co
   bool implicit    = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
   bool jst_scheme  = ((config->GetKind_Centered_Flow() == JST) && (iMesh == MESH_0));
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+  if (!ReducerStrategy)
+    AD::StartNoSharedReading();
+#endif
```
pcarruscag, line 1025:
```suggestion
  /*--- For hybrid parallel AD, pause preaccumulation if there is shared reading of
  * variables, otherwise switch to the faster adjoint evaluation mode. ---*/
  bool pausePreacc = false;
  if (ReducerStrategy) pausePreacc = AD::PausePreaccumulation();
  else AD::StartNoSharedReading();
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1110,6 +1122,12 @@ void CIncEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_cont
   const bool limiter    = (config->GetKind_SlopeLimit_Flow() != NO_LIMITER);
   const bool van_albada = (config->GetKind_SlopeLimit_Flow() == VAN_ALBADA_EDGE);
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+  if (!ReducerStrategy)
+    AD::StartNoSharedReading();
+#endif
```
pcarruscag, line 1129:
```suggestion
  /*--- For hybrid parallel AD, pause preaccumulation if there is shared reading of
  * variables, otherwise switch to the faster adjoint evaluation mode. ---*/
  bool pausePreacc = false;
  if (ReducerStrategy) pausePreacc = AD::PausePreaccumulation();
  else AD::StartNoSharedReading();
```

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -106,6 +106,12 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
   su2double solution_i[MAXNVAR] = {0.0}, flowPrimVar_i[MAXNVARFLOW] = {0.0};
   su2double solution_j[MAXNVAR] = {0.0}, flowPrimVar_j[MAXNVARFLOW] = {0.0};
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = ReducerStrategy && AD::PausePreaccumulation();
+  if (!ReducerStrategy)
+    AD::StartNoSharedReading();
+#endif
```
pcarruscag, line 113:
```suggestion
  /*--- For hybrid parallel AD, pause preaccumulation if there is shared reading of
  * variables, otherwise switch to the faster adjoint evaluation mode. ---*/
  bool pausePreacc = false;
  if (ReducerStrategy) pausePreacc = AD::PausePreaccumulation();
  else AD::StartNoSharedReading();
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1557,6 +1567,12 @@ void CFVMFlowSolverBase<V, R>::EdgeFluxResidual(const CGeometry *geometry,
     END_SU2_OMP_FOR
   }
 
+#ifdef HAVE_OPDI
+  AD::ResumePreaccumulation(preaccEnabled);
+  if (!ReducerStrategy)
+    AD::EndNoSharedReading();
+#endif
```
pcarruscag, line 1574:
```suggestion
  /*--- Restore preaccumulation and adjoint evaluation state. ---*/
  AD::ResumePreaccumulation(pausePreacc);
  if (!ReducerStrategy) AD::EndNoSharedReading();
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2176,6 +2182,12 @@ void CEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_contain
   END_SU2_OMP_FOR
   } // end color loop
 
+#ifdef HAVE_OPDI
+  AD::ResumePreaccumulation(preaccEnabled);
+  if (!ReducerStrategy)
+    AD::EndNoSharedReading();
+#endif
```
pcarruscag, line 2189:
```suggestion
  /*--- Restore preaccumulation and adjoint evaluation state. ---*/
  AD::ResumePreaccumulation(pausePreacc);
  if (!ReducerStrategy) AD::EndNoSharedReading();
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1082,6 +1088,12 @@ void CIncEulerSolver::Centered_Residual(CGeometry *geometry, CSolver **solver_co
   END_SU2_OMP_FOR
   } // end color loop
 
+#ifdef HAVE_OPDI
+  AD::ResumePreaccumulation(preaccEnabled);
+  if (!ReducerStrategy)
+    AD::EndNoSharedReading();
+#endif
```
pcarruscag, line 1095:
```suggestion
  /*--- Restore preaccumulation and adjoint evaluation state. ---*/
  AD::ResumePreaccumulation(pausePreacc);
  if (!ReducerStrategy) AD::EndNoSharedReading();
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1250,6 +1268,12 @@ void CIncEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_cont
...
```
pcarruscag, line 1275:
```suggestion
  /*--- Restore preaccumulation and adjoint evaluation state. ---*/
  AD::ResumePreaccumulation(pausePreacc);
  if (!ReducerStrategy) AD::EndNoSharedReading();
```

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -232,6 +238,12 @@ void CTurbSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
   END_SU2_OMP_FOR
   } // end color loop
 
+#ifdef HAVE_OPDI
+  AD::ResumePreaccumulation(preaccEnabled);
+  if (!ReducerStrategy)
+    AD::EndNoSharedReading();
+#endif
```
pcarruscag, line 245:
```suggestion
  /*--- Restore preaccumulation and adjoint evaluation state. ---*/
  AD::ResumePreaccumulation(pausePreacc);
  if (!ReducerStrategy) AD::EndNoSharedReading();
```

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -76,6 +76,10 @@ void CIntegration::Space_Integration(CGeometry *geometry,
   CNumerics* conv_bound_numerics = numerics[CONV_BOUND_TERM + omp_get_thread_num()*MAX_TERMS];
   CNumerics* visc_bound_numerics = numerics[VISC_BOUND_TERM + omp_get_thread_num()*MAX_TERMS];
 
+#ifdef HAVE_OPDI
+  const auto preaccEnabled = AD::PausePreaccumulation();
+#endif
```
pcarruscag, line 81:
```suggestion
  /*--- Pause preaccumulation in boundary conditions for hybrid parallel AD. ---*/
  /// TODO: Check if this is really needed.
  const auto pausePreacc = (omp_get_num_threads() > 1) && AD::PausePreaccumulation();
```

SU2_CFD/src/integration/CIntegration.cpp
```diff
@@ -178,6 +182,10 @@ void CIntegration::Space_Integration(CGeometry *geometry,
     solver_container[MainSolver]->BC_Periodic(geometry, solver_container, conv_bound_numerics, config);
   }
 
+#ifdef HAVE_OPDI
+  AD::ResumePreaccumulation(preaccEnabled);
+#endif
```
pcarruscag, line 187:
```suggestion
  AD::ResumePreaccumulation(pausePreacc);
```

pcarruscag (commented):
You are right, short-circuiting is not the most natural thing (the name itself makes it sound undesirable), but it allows the state flag to be const...
Have a look at the changes below, I think we can do without the #ifdefs since "ReducerStrategy" is never true when there is only 1 thread or when the code is compiled without OpenMP.
(I also removed most short circuits).

Common/include/basic_types/ad_structure.hpp
```diff
@@ -522,6 +539,39 @@ namespace AD{
...
+    SU2_OMP_MASTER
+    PreaccEnabled = true;
+    END_SU2_OMP_MASTER
+    SU2_OMP_BARRIER
+  }
+
+  FORCEINLINE void StartNoSharedReading() {
+#ifdef HAVE_OPDI
+    opdi::logic->setAdjointAccessMode(opdi::LogicInterface::AdjointAccessMode::Classical);
+    opdi::logic->addReverseBarrier();
```
pcarruscag, line 565: Thanks for this, I think when I tried it I forgot the reverse barrier.
If performance testing goes well this mode can also be enabled in the `::Source_Residual` routines of the solvers.

jblueh:
The suggestions look good to me! :+1:

jblueh:
> The computed values of `discadj_arina2k` in `hybrid_regression_AD` are suspiciously close to the expected values in `parallel_regression_AD`. Could the reference values be wrong? If I test with `-Denable-mixedprec=false` locally, the same holds true for `discadj_fea`.


I just saw that `discadj_arina2k` was addressed in #1311, so I suppose that its reference values are indeed outdated.

TestCases/hybrid_regression_AD.py
```diff
@@ -0,0 +1,238 @@
...
+    discadj_cylinder3D.test_iter = 5
+    discadj_cylinder3D.test_vals = [-3.730673, -3.832084, -0.000000, 0.000000]
+    test_list.append(discadj_cylinder3D)
+
+    # Arina nozzle 2D
+    discadj_arina2k              = TestCase('discadj_arina2k')
+    discadj_arina2k.cfg_dir      = "disc_adj_euler/arina2k"
+    discadj_arina2k.cfg_file     = "Arina2KRS.cfg"
+    discadj_arina2k.test_iter    = 20
+    discadj_arina2k.test_vals    = [2.107806, 1.574254, 47250.000000, 0.000000]
```
pcarruscag, line 66:
```suggestion
    discadj_arina2k.test_vals    = [-3.087876, -3.481506, 0.068878, 0.000000]
```

pcarruscag:
I was wondering where the massive difference in reference values came from, indeed it is from other changes that affected that case, maybe discadj_fea is just "floating point" noise due to preacc on vs off.

TestCases/hybrid_regression_AD.py
```diff
@@ -0,0 +1,238 @@
...
+    ###################################
+    ### Structural Adjoint          ###
+    ###################################
+   
+    # Structural model
+    discadj_fea           = TestCase('discadj_fea')
+    discadj_fea.cfg_dir   = "disc_adj_fea"
+    discadj_fea.cfg_file  = "configAD_fem.cfg" 
+    discadj_fea.test_iter = 4
+    discadj_fea.test_vals = [2.183540, 2.071459, -0.000363, -8.655000]
```
pcarruscag, line 206:
```suggestion
    discadj_fea.test_vals = [1.774569, 1.928023, -0.000364, -8.690300]
```
Seems like these values are repeatable after preaccumulation was enabled.

jblueh: I have not observed any variations either

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -923,6 +923,7 @@ unsigned long CSysSolve<ScalarType>::Solve(CSysMatrix<ScalarType> & Jacobian, co
     Iterations = IterLinSol;
   }
   END_SU2_OMP_MASTER
+  SU2_OMP_BARRIER
```
pcarruscag, line 926: I don't think this is strictly necessary (for Residual and Iterations at least) because it is also the master thread that will request those fields to store the results in CSolver after the linear system is solved. Did you run into some issues?

jblueh: The *suspected* bug was the following: If the master thread is significantly ahead, the shared variables `LinSysRes_ptr` and `LinSysSol_ptr` are reset in `HandleTemporariesOut` although other threads still need them. The code looks like a barrier to prevent this is missing, but after rechecking I think the one in `CSysVector::PassiveCopy` called from `HandleTemporariesOut` does the trick, and the corner case of `(const void*)this == (const void*)&other` that skips this barrier is irrelevant for the part of code in question?

jblueh: resolved in fcc39ce28bd66586a9854e97af43a65fdb6d2c82 as discussed in the dev meeting

jblueh:
Turns out that the deadlocks I observed were actually not an SU2 issue at all but a problem of OpenMPI 3.1 on certain partitions of our cluster.

Common/include/basic_types/ad_structure.hpp
jblueh: Not for all parts though, can it? `CEulerSolver::SourceResidual`, for example, has a computational path that leads to `computeGradientsGreenGauss` where preacc had to be disabled due to shared reading.

pcarruscag: Only for the iPoint loops involving numerics where we "Set" data based on iPoint alone.

jblueh: I agree. I added this for the `Source_Residual` routines that use OpenMP in 2830dea1c0da1cc6d6a53bf0a3403723de5d2ba7.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1656,6 +1658,8 @@ void CFVMFlowSolverBase<V, FlowRegime>::SetResidual_DualTime(CGeometry *geometry
     }
     END_SU2_OMP_FOR
 
+    AD::EndNoSharedReading();
+
```
jblueh, line 1662: Some of these shared reading optimizations depend on `TimeStep` being passive. Does this always hold true?

pcarruscag: It's a fair-enough assumption, I added some more over the Primitive loops and removed some over smaller loops where the performance benefit might not justify the increased maintenance.

pcarruscag (approved):
So the only thing left is to decide whether hybrid AD compiles by default if -Dwith-omp=true or if a special flag is added?

jblueh:
> So the only thing left is to decide whether hybrid AD compiles by default if -Dwith-omp=true or if a special flag is added?

You mean a flag like `-Dwith-opdi`? This would only be meaningful for reverse AD + OpenMP. Conversely, reverse AD + OpenMP are not very meaningful without OpDiLib (you could probably execute it with `-t 1`, but then why bother using OpenMP in the first place). So if that was the question, I would say we leave it the way it is at the moment.

pcarruscag:
The omp simd directive is used in the reverse mode for passive linear algebra things, not a huge deal.

jblueh:
I see, you can't drop `-Dwith-omp` without losing vectorization. For a vectorized serial code, you want to be able to compile with OpenMP but make everything else, or at least AD, behave as if you wouldn't?

pcarruscag:
We don't really lose vectorization... the compiler will still generate it eventually, it will just be a bit less consistent across compilers and what they include in -O2 or -O3.
This is mostly for vector operations which probably don't take a huge time anyway, to justify another flag.
Maybe FORCE_OPDI_OMPT_BACKEND could be used to encode an "off-switch"?

jblueh:
Have a look at a9466bb158ca8675352bf3b4b887cbf82b7f28ff for a suggestion on how to disable OpDiLib within the scope of the existing build options (casually tested).

To clarify the AD implications: If you use OpDiLib, you will have atomic adjoints where no shared reading optimizations are applied, even with `-t 1`. Disabling OpDiLib - if we add that option - will amend this. The presence of `_OPENMP`, however, makes SU2 use the thread-safe CoDiPack type and enables thread-safety augmentations in CoDiPack. Therefore, a reverse AD + OpenMP - OpDiLib build is not the same as a classical serial reverse AD build that is vectorized one way or the other.

If we add this, reverse AD + OpenMP - OpDiLib should also print an error for `-t 2` or higher.

Since I am not 100% sure about it yet, could you clarify again what it is that you want to achieve in the end, and why?

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -880,6 +880,12 @@ void CDiscAdjMultizoneDriver::SetAdj_ObjFunction() {
 
 void CDiscAdjMultizoneDriver::ComputeAdjoints(unsigned short iZone, bool eval_transfer) {
 
+#if defined(CODI_INDEX_TAPE) || defined(HAVE_OPDI)
+  if (nZone > 1) {
```
pcarruscag, line 884:
```suggestion
  if (nZone > 1 && rank == MASTER_NODE) {
```

pcarruscag:
Right, let's get this moving so we can have 7.2.0

PR #1290, opened 2021-05-19
CFVMOutput & Streamwise+spanwise periodic

TobiKattmann:
## Proposed Changes
Hi all,

- Little change to make streamwise and spanwise periodic simulations possible (in CIncNSSolver.cpp). 
- Some cleanup in CConjugateHeatInterface
- Added a CFVMOutput class that only has a CommonFVMoutput for Volume output (COORD-XYZ (fresh addition), MESH_QUALITY, Multigrid visualization and RANK)
- in the output folder(s) there were quite some wrong doxygen `\file <filename>`, i changed that. I renamed `CAdjFlowOutput.hpp` ->  `CAdjFlowCompOutput.hpp` as the cpp as that name and that class as well.
- little changes/additions to unst CHT gradient validation case 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

Issue #1229, opened 2021-03-12
Mesh type used in optimization and numerical simulation


tuzitt:
**Describe the bug**
     hello, I use SU2 to do shape optimization for M6 wing. I discover SU2_GEO can not identify the section shape correctly if I use the structured grid.  But it can correctly identify the section shape，airfoil，when I use the unstructured grid.  Does it means the unstructured grid is the only mesh type can be used in shape optimization?   
      I use a unstructured grid do shape optimization, my objective is minimizi drag with fixed lift,  I got a  model with less drag in SU2. I used the SU2_CFD module to verify whether the optimized model really reduces drag with more volume mesh. I got a nice result. Compared to the original model, the optimized model has reduced drag. But when the same grids, original model grid and optimized model grid,  is used for numerical simulation in Fluent, I got a bad result. there is almost no drag reduction, and drag, lift and pitching moment are all smaller than SU2. Is there something wrong with the grid or something else?  
      I used SU2 to simulate the structural grid of the CRM model downloaded from the https://aiaa-dpw.larc.nasa.gov. The drag and lift  are consistent with the result of Fluent. But pitching moment is 5 times of that calculated by Fluent. I think maybe there is something wrong with my configuration.  But I got a relatively correct pitching moment when I use the same configuration to similate  with a unstructured grid. This makes me doubt the mesh type.
      I am so confused with these problems. I will appreciate very much if  someone can give me some advice？

- SU2 Version: [e.g., v6.2.0]

PR #1295, opened 2021-06-02
Fix restart logic in python FSI

Nicola-Fonzi:
Dear all,

This PR resolves an issue with the updated restart logic. Further, it also provides a couple of modifications which will be required in a future PR aimed at the automatic identification of the aerodynamic system.

## PR Checklist


SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -525,7 +525,7 @@ class CDriver {
    * \param[in] iVertex - Vertex identifier.
    * \return Unit normal (vector) at the vertex.
    */
-  vector<passivedouble> GetVertexUnitNormal(unsigned short iMarker, unsigned long iVertex) const;
+  vector<passivedouble> GetVertexNormal(unsigned short iMarker, unsigned long iVertex, bool unitNormal) const;
```
pcarruscag, line 528:
```suggestion
  vector<passivedouble> GetVertexNormal(unsigned short iMarker, unsigned long iVertex, bool unitNormal = false) const;

  inline vector<passivedouble> GetVertexUnitNormal(unsigned short iMarker, unsigned long iVertex) const {
    return GetVertexNormal(iMarker, iVertex, true);
  }
```

IDK if swig understands inline methods (?) but it would be good to maintain some stability of the python interface.
(unless this method (GetVertexUnitNormal) was new and never released before).

Nicola-Fonzi: I see your point and agree with you

Nicola-Fonzi:
The unsteady CHT regression is failing, but I cannot understand why. I should not have touched parts of the code that influence that regression test. May I ask you if you have any idea of the reason for this fail?

pcarruscag:
@TobiKattmann any idea about what may be happening with the unsteady cht tutorial?

TobiKattmann:
> 
> 
> @TobiKattmann any idea about what may be happening with the unsteady cht tutorial?

Oh man I am so sorry, It was already failing in #1290  that was merged 3 days ago by me. So it is not your problem @Nicola-Fonzi . 

Let me explain how that happened (not to justify that or anything): I wanted to move the unsteady CHT case to the tutorial.py file because it is, in fact, a tutorial. As I didn't have a PR open that time I put that into #1226 because it fit into the PR-scope. The PR was not worked on ever since so I now thought I just squeeze it into #1290 to have that covered.
The testcase was not failing in #1226 so I cherry-picked that into #1290 and obviously didn't check if it still works. Code Factor was always giving me red-crosses for all the commits so I didn't pay much attention to that :( 👎 

I apologize again and will fix that asap and see what is going on

pcarruscag (approved):

pcarruscag:
Hmmm looks like we have an issue with the setup of tutorials.py then, because the merge button should not show up as green with failing regressions.
The code quality tests are not hard requirements but everything else should be.

Nicola-Fonzi:
Indeed the tutorial.py is not a required test.
Anyway, I am happy I did not break anything. Do you think we can merge this?

TobiKattmann:
Ok, I think I understood and fixed what happened (and did 2 more stupid mistakes).

In #1226 2 things happened:  
1. I moved the reg-test to tutorials and therefore I had to add files to /su2code/Tutorials/12 of course. There I changed some values (Linear Solver Iter, Outer_Iter, Time_Iter, to name the most important) because the case until then did not get unsteady. Because the original case worked in #1226 before my commit and I only changed cfg-options over at Tutorials and touched no more than tutorials.py -> there was no unwanted introduced code changed that is overlooked!
2. in #1229 three cfg option names for heat were changed ... so I changed them over at /su2code/Tutorials as well cause I thought that the test have to pass in #1229 and /su2code/Tutorials contains the cfg's as well (other than for the usual Testcases). I also thought that #1229 would be merged in quick succession so that premature change wouldn't hurt much ... well that merge didnt happen until today.

Point 2. above bit me when I cherry picked in #1290 which introduced the failing test visible here. Because with that I changed from the Testcases-test to the Tutorial-test which already had the new cfg-names in it ... which are not in the code yet! Changing that back gave the correct values on my machine and I get the correct values with current develop wrt to the old test (which can be seen in the diff of #1226). Like so I am confident that no code changed the regression test values but rather incautious Tobi.

The 2 mistakes I made were:
1. I pushed the changes (different cfg-option-names) directly to develop without a PR. That is prob not too bad tbh but I would have preferred a PR
2. I re-ran the tests for the last commit of this PR to check but I guess it is better to leave that alone and just do that somewhere else in a new PR just to test that ... not sure if that works out

TobiKattmann:
I rerun https://github.com/su2code/SU2/runs/2804205189 in the open PR #1294 which will tell us if this issue is resolved as this commit had this very tutorial case failing as well.

Issue #1262, opened 2021-04-13
Expose mesh deformation (SU2_DEF) and AD gradient projection (SU2_DOT_AD) to the pysu2 Python wrapper


aa-g:
**Context**
I am working in a parallel MDO framework where one or several instances of SU2 (corresponding to different operating cases) can coupled with other solvers and tools, to perform coupled aero-structural shape optimisation. 

The framework initialises MPI and assigns processors to the different solvers and tools as required. The SU2 instances are initialised and controlled through Python, and given a communicator for the relevant subset of the MPI processes. 

**Issue**
Shape optimisation with SU2 (using discrete adjoints) require four operations at each major optimisation iteration:

1. `SU2_DEF`: Deform the mesh according to the new design variables (or _design deformations_, in contrast to e.g. the _structural deformations_ in an aeroelastic analysis), and output the deformed mesh to file,
2. `SU2_CFD`: Update the primal solution, which may include a mesh deformation due to structural deformations in an aeroelastic analysis, and output the converged state to file,
3. `SU2_CFD_AD`: Update the discrete adjoint solution corresponding to the converged primal solution, once for each objective function of interest, and output the converged adjoints to file,
4. `SU2_DOT_AD`: Calculate the surface sensititivities from the discrete adjoint solution (from what I can gather, this step is required in the discrete-adjoint sensitivity analysis to account for the in-direct influence surface deformations have on the objective through the volume deformation; essentially a differentiation of `SU2_DEF`), and output these to file.

Currently, the SU2 Python wrapper exposes _driver_ classes which allow control of the execution of steps 2 and 3 described above, namely `CSinglezoneDriver` and `CDiscAdjSinglezonedriver` (among other variants). For steps 1 and 4 however, there is no such option: these steps must be executed by a "command-line call" in a newly spawned sub-process, as is done in the SU2 Python tools, in `SU2.run.DEF` and `SU2.run.DOT` respectively.

However, spawning new -- parallel -- processes using `subprocess` from within an MPI environment is "dangerous", besides being rather rude towards workload managers such as Slurm, making this an unviable solution in this kind of application.

**Proposed solution**
If at all reasonable, I would like to have the functionality of `SU2_DEF` and `SU2_DOT_AD` exposed to the Python wrappers `pysu2` and `pysu2ad`, ideally using an approach which is consistent with the existing implementation for `SU2_CFD` and `SU2_CFD_AD`.

If I understand correctly, this change would require something along the lines of: 

[...]

PR #1300, opened 2021-06-10
Refactorization of SU2 Grid Deformation for Python

patelha57:
## Proposed Changes
This pull request represents progress towards exposing the mesh deformation routines and their adjoint counterparts to Python through pysu2 and pysu2ad, respectively. This refactorization introduced:

1. `CDriverBase`
       a. Parent class for `CDriver`, `CDeformationDriver`, `CDiscAdjDeformationDriver`
       b. Common methods regarding geometry, markers, MPI, etc.

2. `CDeformationDriver` and `CDiscAdjDeformationDriver`

In addition, the `CDeformationDriver` was modified to use the new mesh deformation solver `CMeshSolver`. The integration of `CMeshSolver` required introducing the `numerics_container` and `solver_container`, as well as minor changes to `CMeshSolver` (e.g., `DeformMesh` and `SetMesh_Stiffness`). Both the legacy mesh deformation solver `SU2_DEF` and the newer `CMeshSolver` can be used with these new drivers, depending on how the `DEFORM_MESH` option is set in the SU2 config file. 

## Related Work
These efforts are related to Issue #1262. The purpose of this work is to enable the coupling of SU2 with external solvers like Nastran, particularly for aerostructural MDAO. The goal is to apply this for coupled discrete adjoint sensitivities. 

In the future, these driver classes will be extended to include more methods accessible via Python (i.e., for setting or extracting variables, config parameters, etc.). Currently `CDiscAdjDeformationDriver` is not integrated with `CMeshSolver`, but this will be addressed in the future. 

## PR Checklist


SU2_DEF/include/drivers/CDiscAdjDeformationDriver.hpp
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#define ENABLE_MAPS
+#include "../../../Common/include/CConfig.hpp"
+#undef ENABLE_MAPS
```
pcarruscag, line 32: we should only do this ENABLE_MAPS in .cpp files to keep it from spreading.
(the const static maps we have in option_structure become a compilation problem if used everywhere)

SU2_DEF/src/meson.build
```diff
@@ -1,38 +1,93 @@
...
-                                        'output/filewriter/CParaviewXMLFileWriter.cpp',
-                                        'output/filewriter/CParaviewVTMFileWriter.cpp',
-                                        'output/filewriter/CSU2MeshFileWriter.cpp',
-                                        'limiters/CLimiterDetails.cpp'])
+su2_cfd_obj = su2_cfd_lib.extract_objects(['solvers/CSolver.cpp',
+                                      'solvers/CBaselineSolver.cpp',
+                                      'CMarkerProfileReaderFVM.cpp',
+                                      'output/COutput.cpp',
+                                      'output/tools/CWindowingTools.cpp',
+                                      'output/CMeshOutput.cpp',
```
pcarruscag, line 14: you can simplify this by linking with the entire su2_cfd_lib library

SU2_DEF/src/drivers/CDiscAdjDeformationDriver.cpp
```diff
@@ -0,0 +1,945 @@
...
+    }
+}
+
+void CDiscAdjDeformationDriver::Geometrical_Preprocessing() {
+
+    /*--- Loop over all zones to initialize the various classes. In most
+     cases, nZone is equal to one. This represents the solution of a partial
+     differential equation on a single block, unstructured mesh. ---*/
+
+     for (iZone = 0; iZone < nZone; iZone++) {
```
pcarruscag, line 152: If possible it would be good to share methods like these between the two deformation drivers, for example DiscAdj could inherit from the normal, or the two could share a base, etc.

SU2_DEF/src/drivers/CDiscAdjDeformationDriver.cpp
```diff
@@ -0,0 +1,945 @@
...
+    driver_config = nullptr;
+
+    if (rank == MASTER_NODE)
+        cout << "\n------------------------- Solver Postprocessing -------------------------" << endl;
+
+    if (geometry_container != nullptr) {
+        for (iZone = 0; iZone < nZone; iZone++) {
+            if (geometry_container[iZone] != nullptr) {
+                for (iInst = 0; iInst < nInst[iZone]; iInst++){
+                    if (geometry_container[iZone][iInst] != nullptr) {
```
pcarruscag, line 361: These inner-most null-checks are not needed and we have been removing them from the rest of the code.

SU2_DEF/src/drivers/CDiscAdjDeformationDriver.cpp
```diff
@@ -0,0 +1,945 @@
...
+
+    nDV_Value =  config->GetnDV_Value(iDV);
+
+    for (iDV_Value = 0; iDV_Value < nDV_Value; iDV_Value++){
+
+      /*--- Initilization with su2double resets the index ---*/
+
+      DV_Value = 0.0;
+
+      AD::RegisterInput(DV_Value);
```
pcarruscag, line 698: You will need to apply some of the recent changes I made to SU2_DOT as this does not work anymore (it was a logic error for AD to do it like this).

pcarruscag (commented):
Looking good :+1:

TobiKattmann:
fyi
1. The failing tutorials.py regression test are solved elsewhere so that will disappear upon the next pushed commit
2. The Complex Code error of CodeFactor can be (most likely) disregarded as a false positive. It does so for eg CSolver.cpp already so i guess we have ourselves a new hypochondriac-file

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -32,6 +32,7 @@
 
 #include <iostream>
 #include <vector>
+#include <limits>
```
aa-g, line 35: This was giving me an error after updgrading to GCC 11.1.0

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4205,8 +4205,16 @@ class CSolver {
    * \param[in] config - Definition of the particular problem.
    * \param[in] referenceCoord - Determine if the mesh is deformed from the reference or from the current coordinates.
    */
-  inline virtual void SetMesh_Stiffness(CGeometry **geometry,
-                                        CNumerics **numerics,
+  inline virtual void DeformMesh(CGeometry *geometry,
```
aa-g, line 4225: The main issue in integrating the CMeshSolver in CDeformationDriver was that several routines in the solver implementation are specifically written to handle multigrid. To circumvent these function calls, it seemed simpler as a first approach to overload the DeformMesh function so that, if a simple pointer to a geometry is given (rather than a double pointer), a multigrid-free implementation is used.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4205,8 +4205,16 @@ class CSolver {
...
+                                 CNumerics **numerics,
+                                 CConfig *config) { }
+
+  /*!
+   * \brief A virtual member.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] referenceCoord - Determine if the mesh is deformed from the reference or from the current coordinates.
+   */
+  inline virtual void SetMesh_Stiffness(CNumerics **numerics,
```
aa-g, line 4234: Here, a `CGeometry**` was expected, but not used. To make the function compatible with `CDeformationDriver` which does not have multigrid, this argument was removed (at the cost of some consistency with the signatures of other methods, maybe).

SU2_CFD/src/solvers/CMeshSolver.cpp
```diff
@@ -545,6 +545,63 @@ void CMeshSolver::DeformMesh(CGeometry **geometry, CNumerics **numerics, CConfig
 
 }
 
+void CMeshSolver::DeformMesh(CGeometry *geometry, CNumerics **numerics, CConfig *config){
+
+  if (multizone) nodes->Set_BGSSolution_k();
+
+  /*--- Capture a few MPI dependencies for AD. ---*/
+  geometry->InitiateComms(geometry, config, COORDINATES);
```
pcarruscag, line 553: So, what we do in such cases it to create a protected implementation function, e.g. DeformMesh_impl which in addition to a **geometry, etc., takes a boolean for MG, then the two DeformMesh overloads call the implementation function.

It does not have to be exactly like that, but the moral of the story is that if you ever see the need to copy-paste, cut-paste instead.
BTW I think this is a bit out of date with develop?

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,804 @@
...
+    }
+  }
+}
+
+void CDeformationDriver::Postprocessing() {
+
+    if (rank == MASTER_NODE)
+      cout << endl <<"------------------------- Solver Postprocessing -------------------------" << endl;
+
+    delete driver_config;
```
pcarruscag, line 550: make this the destructor and remove the Postprocessing, it should clear the LGTM warning you are getting at the moment.

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,804 @@
...
+    allBoundariesMap[Marker_Tag] = iMarker;
+  }
+
+  return allBoundariesMap;
+}
+
+map<string, string> CDeformationDriver::GetAllBoundaryMarkersType() const {
+
+  map<string, string> allBoundariesTypeMap;
+  unsigned short iMarker, KindBC;
```
pcarruscag, line 652: I see what you mean with the duplication of python wrapper methods now... This needs a better solution, either this class needs to be derived from CDriver, or this class and CDriver need to share a new base class that provides this functions.
You can move these new files to SU2_CFD/src/drivers, if it makes that more natural.
Please give this some thought because duplication causes very stubborn bugs to remove.

SU2_DEF/src/meson.build
```diff
@@ -1,38 +1,40 @@
...
+  su2_def_lib_ad = static_library('SU2core_DEF_AD',
+                                  su2_def_src,
+                                  install : false,
+                                  dependencies : [su2_deps, codi_dep, commonAD_dep],
+                                  cpp_args:  [default_warning_flags, su2_cpp_args, codi_rev_args])
+
+  su2_def_dep_ad = declare_dependency(link_with: su2_def_lib_ad,
+                                     include_directories: su2_def_include)
+
+endif
```
pcarruscag, line 40: Is this needed?

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,804 @@
...
+
+  rank = SU2_MPI::GetRank();
+  size = SU2_MPI::GetSize();
+
+  /*--- Copy the config filename ---*/
+  strcpy(config_file_name, confFile);
+
+  /*--- Initialize the configuration of the driver ---*/
+  driver_config = nullptr;
+  driver_config = new CConfig(config_file_name, SU2_COMPONENT::SU2_DEF);
```
pcarruscag, line 58:
```suggestion
  driver_config = new CConfig(config_file_name, SU2_COMPONENT::SU2_DEF);
```

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,804 @@
...
+        solver_container[iZone]       = nullptr;
+        numerics_container[iZone]     = nullptr;
+    }
+}
+
+void CDeformationDriver::Input_Preprocessing() {
+
+    /*--- Initialize a char to store the zone filename ---*/
+    char zone_file_name[MAX_STRING_SIZE];
+
```
pcarruscag, line 139: Run clang-format on the new files at some point please. We have been doing that for new files to make everything consistent.
The format file is in the root of the repo.

SU2_DEF/src/meson.build
patelha57: Can you please elaborate on which parts are redundant or need to be addressed?

pcarruscag: There is no SU2_DEF_AD, so what is this "def_lib_ad" for? Is it SU2_DOT_AD but renamed?

patelha57:
Hello @pcarruscag @TobiKattmann and SU2 developers,

We have been busy making several code updates, performing cleanup, etc. for this PR. Can you please review and provide feedback when you get the chance? Thank you in advance!

pcarruscag:
Hi guys, indeed I noticed the final push over the last weeks. Nice work 👍 

A couple practical comments:
- It would be helpful if you guys could point out the main changes to the python API (this Wednesday maybe?), and then by defining a doxygen group so that doxygen can be used to generate user friendly API documentation.
- Please go over our old comments and mark as resolved anything you did already.
- Run clang-format on new files and check indentation changes in existing files, I think I saw a few.

pcarruscag (commented):
One more, there are changes to the submodules (codi, etc ) that need to be undone.

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,651 @@
...
+        solver_container[iZone][INST_0][MESH_0][MESH_SOL]->SetMesh_Stiffness(numerics_container[iZone][INST_0][MESH_0][MESH_SOL], config_container[iZone]);
+        
+        /*--- Deform the volume grid around the new boundary locations ---*/
+        
+        solver_container[iZone][INST_0][MESH_0][MESH_SOL]->DeformMesh(geometry_container[iZone][INST_0][MESH_0], numerics_container[iZone][INST_0][MESH_0][MESH_SOL], config_container[iZone]);
+        
+    }
+}
+
+void CDeformationDriver::Update_Legacy() {
```
WallyMaier, line 337: Is there a need for a legacy update?

Common/include/drivers/CDriverBase.hpp
```diff
@@ -0,0 +1,525 @@
+/*!
+ * \file CDriverBase.hpp
+ * \brief Base class template for all drivers.
+ * \author H. Patel, A. Gastaldi
+ * \version 7.3.0 "Blackbird"
```
WallyMaier, line 5: Make sure to update these headers with version numbers and dates

meson.build
```diff
@@ -231,10 +231,10 @@ endif
 subdir('Common/src')
 # compile SU2_CFD executable
 subdir('SU2_CFD/src')
-# compile SU2_DOT executable
-subdir('SU2_DOT/src')
 # compile SU2_DEF executable
 subdir('SU2_DEF/src')
+# compile SU2_DOT executable
```
WallyMaier, line 236: why did this move?

Common/include/drivers/CDriverBase.hpp
```diff
@@ -0,0 +1,525 @@
...
+     */
+    void SetContainers_Null();
+    
+    /*!
+     * \brief Read in the config and mesh files.
+     */
+    void Input_Preprocessing(CConfig **&config, CConfig *&driver_config);
+    
+    /*!
+     * \brief Construction of the edge-based data structure and the multigrid structure.
```
WallyMaier, line 499: We include the params[in] here as well

meson.build
patelha57: Because `SU2_DOT` is now dependent on `CDiscAdjDeformationDriver (SU2_DEF_AD, previously SU2_DOT_AD)`. The dependency required the build order to be swapped

Common/include/drivers/CDriverBase.hpp
```diff
@@ -0,0 +1,640 @@
...
+ * \author H. Patel, A. Gastaldi
+ * \version 7.4.0 "Blackbird"
+ *
+ * SU2 Project Website: https://su2code.github.io
+ *
+ * The SU2 Project is maintained by the SU2 Foundation
+ * (http://su2foundation.org)
+ *
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
+ *
```
WallyMaier, line 13:
```suggestion
 * Copyright 2012-2022, SU2 Contributors (cf. AUTHORS.md)
 *
```

WallyMaier: Make sure you get all these.

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1289 @@
...
+ * \brief Base class template for all drivers.
+ * \author H. Patel, A. Gastaldi
+ * \version 7.4.0 "Blackbird"
+ *
+ * SU2 Project Website: https://su2code.github.io
+ *
+ * The SU2 Project is maintained by the SU2 Foundation
+ * (http://su2foundation.org)
+ *
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
```
WallyMaier, line 12: anotha one

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1289 @@
...
+
+map<string, string> CDriverBase::GetMarkerTypes() const {
+  map<string, string> typeMap;
+  string type;
+
+  for (auto iMarker = 0u; iMarker < main_config->GetnMarker_All(); iMarker++) {
+    auto tag = main_config->GetMarker_All_TagBound(iMarker);
+    auto kindBC = main_config->GetMarker_All_KindBC(iMarker);
+
+    switch (kindBC) {
```
WallyMaier, line 217: Will this work with Reimann/Giles/Catalytic walls?

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -797,69 +682,69 @@ class CDriver {
  * \author T. Economon, G. Gori
  */
 class CFluidDriver : public CDriver {
-
+
```
WallyMaier, line 685: make sure to run clang format on your files, so we dont have these empty spaces

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -924,61 +809,61 @@ class CTurbomachineryDriver : public CFluidDriver {
...
-  /*!
-   * \brief Reset the convergence flag (set to false) of the solver for the Harmonic Balance.
-   */
-  void ResetConvergence() override;
+    
+    /*!
+     * \brief Constructor of the class.
+     * \param[in] confFile - Configuration file name.
+     * \param[in] val_nZone - Total number of zones.
+     * \param[in] val_nDim - Number of dimensions.
```
WallyMaier, line 824: unneeded

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -31,865 +31,644 @@
...
+            
+            if (rank == MASTER_NODE) cout << endl << "----------------- Python Interface Preprocessing ( Zone "<< iZone <<" ) -----------------" << endl;
+            
+            if (rank == MASTER_NODE) cout << "Setting customized boundary conditions for zone " << iZone << endl;
+            for (iMesh = 0; iMesh <= config[iZone]->GetnMGLevels(); iMesh++) {
+                geometry[iZone][INST_0][iMesh]->SetCustomBoundary(config[iZone]);
+            }
+            geometry[iZone][INST_0][MESH_0]->UpdateCustomBoundaryConditions(geometry[iZone][INST_0], config[iZone]);
+            
+            if ((config[iZone]->GetKind_Solver() == MAIN_SOLVER::EULER) ||
```
WallyMaier, line 51: Is there a reason we didn't include the INC solver beforehand?   Should we add INC and NEMO here as well?

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -31,865 +31,644 @@
...
+    vector<su2double> GradT (3,0.0);
+    vector<su2double> HeatFlux (3,0.0);
+    vector<passivedouble> HeatFluxPassive (3,0.0);
+    
+    bool compressible = (config_container[ZONE_0]->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE);
+    
+    iPoint = geometry_container[ZONE_0][INST_0][MESH_0]->vertex[iMarker][iVertex]->GetNode();
+    
+    if(compressible){
+        laminar_viscosity    = solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
```
WallyMaier, line 238:
```suggestion
    if(compressible){
        //TODO: This will not work for NEMO or INC(?) cases.
        laminar_viscosity    = solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
```

WallyMaier: Or can this grab the values in CFVMSolverBase.inl?

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -31,865 +31,644 @@
...
-  return HeatFluxPassive;
+    
+    HeatFluxPassive[0] = SU2_TYPE::GetValue(HeatFlux[0]);
+    HeatFluxPassive[1] = SU2_TYPE::GetValue(HeatFlux[1]);
+    HeatFluxPassive[2] = SU2_TYPE::GetValue(HeatFlux[2]);
+    
+    return HeatFluxPassive;
 }
 
 passivedouble CDriver::GetVertexNormalHeatFlux(unsigned short iMarker, unsigned long iVertex) const{
```
WallyMaier, line 253: This function looks like it has a lot of shared code from GetVertexHeatFlux

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -31,865 +31,644 @@
...
-      break;
-
-    default:
-      break;
+    
+    for(iZone = 0; iZone < nZone; iZone++) {
+        switch (config_container[iZone]->GetKind_Solver()) {
+                
+            case MAIN_SOLVER::EULER: case MAIN_SOLVER::NAVIER_STOKES: case MAIN_SOLVER::RANS:
+            case MAIN_SOLVER::INC_EULER: case MAIN_SOLVER::INC_NAVIER_STOKES: case MAIN_SOLVER::INC_RANS:
```
WallyMaier, line 389: Can you NEMO cases?

SU2_DEF/include/drivers/CDeformationDriver.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * \brief Headers of the main subroutines for driving the mesh deformation.
+ * \author A. Gastaldi, H. Patel
+ * \version 7.4.0 "Blackbird"
+ *
+ * SU2 Project Website: https://su2code.github.io
+ *
+ * The SU2 Project is maintained by the SU2 Foundation
+ * (http://su2foundation.org)
+ *
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
```
WallyMaier, line 12: Anotha one

SU2_DEF/include/drivers/CDiscAdjDeformationDriver.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * \brief Headers of the main subroutines for driving the projection of sensitivities.
+ * \author T. Economon, H. Kline, R. Sanchez, A. Gastaldi, H. Patel
+ * \version 7.4.0 "Blackbird"
+ *
+ * SU2 Project Website: https://su2code.github.io
+ *
+ * The SU2 Project is maintained by the SU2 Foundation
+ * (http://su2foundation.org)
+ *
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
```
WallyMaier, line 12: Anotha one

TestCases/py_wrapper/flatPlate_rigidMotion/launch_flatPlate_rigidMotion.py
```diff
@@ -104,9 +104,8 @@ def main():
   # Extract the initial position of each node on the moving marker
   CoordX = np.zeros(nVertex_MovingMarker)
   CoordY = np.zeros(nVertex_MovingMarker)
-  CoordZ = np.zeros(nVertex_MovingMarker)
   for iVertex in range(nVertex_MovingMarker):
-    CoordX[iVertex], CoordY[iVertex], CoordZ[iVertex] = SU2Driver.GetInitialMeshCoord(MovingMarkerID, iVertex)
+    CoordX[iVertex], CoordY[iVertex] = SU2Driver.GetMarkerInitialCoordinates(MovingMarkerID, iVertex)
```
WallyMaier, line 108: why did we lose the 3D?

SU2_DEF/src/drivers/CDeformationDriver.cpp
```diff
@@ -0,0 +1,614 @@
...
+ * \brief Main subroutines for driving the mesh deformation.
+ * \author A. Gastaldi, H. Patel
+ * \version 7.4.0 "Blackbird"
+ *
+ * SU2 Project Website: https://su2code.github.io
+ *
+ * The SU2 Project is maintained by the SU2 Foundation
+ * (http://su2foundation.org)
+ *
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
```
WallyMaier, line 12: Anotha 1

SU2_CFD/src/python_wrapper_structure.cpp
patelha57: True. We have addressed this in PR #1750 for `feature_pysu2_fsi_adjoint` branch, which includes all the enhancements and code cleanup to `python_wrapper_structure.cpp`!

SU2_CFD/src/python_wrapper_structure.cpp
patelha57: I am not sure why they were not included before... but I've now added INC and NEMO support.

TestCases/py_wrapper/flatPlate_rigidMotion/launch_flatPlate_rigidMotion.py
patelha57: The older `GetInitialMeshCoord()` required all 3 components, even for a 2D problem. Our `GetMarkerInitialCoordinates()` accounts for the problem definition and only returns 2 components in this case

SU2_DEF/src/drivers/CDiscAdjDeformationDriver.cpp
```diff
@@ -0,0 +1,1065 @@
...
+
+      /*--- Build the grid data structures using the ParMETIS coloring. ---*/
+
+      if (fem_solver) {
+        switch (config_container[iZone]->GetKind_FEM_Flow()) {
+          case DG:
+            geometry_container[iZone][iInst] = new CGeometry*[nMesh];
+            geometry_container[iZone][iInst][MESH_0] = new CMeshFEM_DG(geometry_aux, config_container[iZone]);
+            break;
+        }
```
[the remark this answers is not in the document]
WallyMaier: Probably best to add a default case here

Nicola-Fonzi:
Hi and thank you for this PR, I think it will be really useful.
I noticed that you removed several functions in the python wrapper, and also renamed others. This will break the compatibility with the fsi module. May I ask you why you had to remove it? We can maybe find a solution that keeps the possibility to use the fsi module as well

patelha57:
> Hi and thank you for this PR, I think it will be really useful. I noticed that you removed several functions in the python wrapper, and also renamed others. This will break the compatibility with the fsi module. May I ask you why you had to remove it? We can maybe find a solution that keeps the possibility to use the fsi module as well

Hi Nicola! Are you asking about methods relating to number of vertices, global indices, coordinates, deformable marker tags, etc.? If so, we have not removed those and they are still available through the Python wrapper. These functions were relocated to [CDriverBase](https://github.com/su2code/SU2/blob/feature_pysu2_DEF/Common/include/drivers/CDriverBase.hpp) so they can be inherited by all other drivers. These changes should not break compatibility with the FSI module (aside from some function renaming).

patelha57:
Hi @pcarruscag and @TobiKattmann. Can you please review this PR whenever you get the chance? Thank you very much!

SU2_DEF/include/drivers/CDiscAdjDeformationDriver.hpp
pcarruscag: It looks like some of my old comments are still applicable. Given that this was in the backburner for a long time maybe you can take a look at those comments first.

SU2_DEF/include/drivers/CDiscAdjDeformationDriver.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include "../../../Common/include/fem/fem_geometry_structure.hpp"
+#include "../../../Common/include/geometry/CGeometry.hpp"
+#include "../../../Common/include/grid_movement/CSurfaceMovement.hpp"
+#include "../../../Common/include/grid_movement/CVolumetricMovement.hpp"
+#include "../../../Common/include/parallelization/mpi_structure.hpp"
+#include "../../../SU2_CFD/include/numerics/CGradSmoothing.hpp"
+#include "../../../SU2_CFD/include/output/CBaselineOutput.hpp"
+#include "../../../SU2_CFD/include/output/COutput.hpp"
+#include "../../../SU2_CFD/include/solvers/CBaselineSolver.hpp"
+#include "../../../SU2_CFD/include/solvers/CGradientSmoothingSolver.hpp"
```
pcarruscag, line 44: Why do you need to include some many things here? Reduce to what you need and move to the cpp when possible.

pcarruscag (commented):
In general, this is a (challenging) cut-and-paste job, so there's not much to review. Do you know why the tests are failing?

pcarruscag:
Also be sure to go over the CodeQL alerts

Nicola-Fonzi:
Hi @patelha57 , yes I meant those functions. I could not find them because I did a fast ctrl+f search in the files and it did not work because of the renaming. I should have read more carefully. I think it would have been better to keep the old names, as renaming tends to create small issues here and there. However, I see that you fixed that almost everywhere, also in the python test cases. Thank you!
Probably only the fsi_interface and fsi_computation files are missing

Nicola-Fonzi:
Thank you very much for taking care of the changes

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c feature_FFD -s ${{matrix.testscript}}
```
WallyMaier, line 131: I know we talked about this before, but I am making a note here  just to be sure.

I forgot to change this back in a recent PR.

Common/include/drivers/CDriverBase.hpp
```diff
@@ -0,0 +1,640 @@
...
+
+#pragma once
+
+#include "../../../SU2_CFD/include/numerics/CNumerics.hpp"
+#include "../../../SU2_CFD/include/output/COutput.hpp"
+#include "../../../SU2_CFD/include/solvers/CSolver.hpp"
+#include "../CConfig.hpp"
+
+class CDriverBase {
+ protected:
```
pcarruscag, line 36: This should be in SU2_CFD instead of Common, and it is missing doxygen docs

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+  return FFDBox[ZONE_0][iFFDBox]->GetnCornerPoints();
+}
+
+unsigned short CDriverBase::GetNumberFFDBoxControlPoints(unsigned short iFFDBox) const {
+  if (GetNumberFFDBoxes() == 0) {
+    SU2_MPI::Error("FFD Box definition not found in the mesh file.", CURRENT_FUNCTION);
+  }
+  if (iFFDBox >= GetNumberFFDBoxes()) {
+    SU2_MPI::Error("FFD Box index exceeds size.", CURRENT_FUNCTION);
+  }
```
pcarruscag, line 99: Too much copy paste, put the checks in a function

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+
+  vector<unsigned long> values;
+
+  for (auto iPoint = 0ul; iPoint < GetNumberFFDBoxSurfacePoints(iFFDBox); iPoint++) {
+    values.push_back(FFDBox[ZONE_0][iFFDBox]->Get_VertexIndex(iPoint));
+  }
+  return values;
+}
```
pcarruscag, line 145: Making the driver the interface to the FFD does not feel like the best idea, because you need to basically duplicate the FFD interface here.
Would it not be possible to wrap the FFD with swig? And then in python do `driver.GetFFD(iFFD).GetNumberSurfacePoints()` and so on?
Can you stash this for a different PR?

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+    values.push_back(main_geometry->nodes->GetGlobalIndex(iPoint));
+  }
+
+  return values;
+}
+
+vector<vector<unsigned long>> CDriverBase::GetMarkerElementConnectivities(unsigned short iMarker) const {
+  const auto nBound = GetNumberMarkerElements(iMarker);
+
+  vector<vector<unsigned long>> values;
```
pcarruscag, line 462: Is it really worth having this type of function in the interface? This would be one line of python and could be stored in a numpy array which would probably be more efficient than the list of lists than vector<vector> maps to?

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+
+  for (auto iBound = 0ul; iBound < nBound; iBound++) {
+    values.push_back(GetMarkerElementConnectivities(iMarker, iBound));
+  }
+
+  return values;
+}
+
+vector<unsigned long> CDriverBase::GetMarkerElementConnectivities(unsigned short iMarker, unsigned long iBound) const {
```
pcarruscag, line 471: This designation is not consistent with C++ and it is misleading, change to ElementNodes please.

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+  vector<unsigned long> values;
+
+  for (auto iVertex = 0ul; iVertex < nVertex; iVertex++) {
+    values.push_back(GetMarkerVertexIndices(iMarker, iVertex));
+  }
+
+  return values;
+}
+
+unsigned long CDriverBase::GetMarkerVertexIndices(unsigned short iMarker, unsigned long iVertex) const {
```
pcarruscag, line 539: Same, Node instead of Index

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+  vector<unsigned long> values;
+
+  for (auto iPoint = 0ul; iPoint < nPoint; iPoint++) {
+    values.push_back(GetVertexIDs(iPoint));
+  }
+
+  return values;
+}
+
+unsigned long CDriverBase::GetVertexIDs(unsigned long iPoint) const {
```
pcarruscag, line 559: Again, why have a different name than in c++? In SU2 we use vertex to refer to boundary nodes.

SU2_CFD/src/python_wrapper_structure.cpp
```diff
@@ -31,865 +31,650 @@
...
+    
+    unsigned short val_iZone = ZONE_0;
+    unsigned short FinestMesh = config_container[val_iZone]->GetFinestMesh();
+    su2double CDrag, factor, val_Drag;
+    
+    /*--- Calculate drag force based on drag coefficient ---*/
+    factor = solver_container[val_iZone][INST_0][FinestMesh][FLOW_SOL]->GetAeroCoeffsReferenceForce();
+    CDrag = solver_container[val_iZone][INST_0][FinestMesh][FLOW_SOL]->GetTotal_CD();
+    
+    val_Drag = CDrag*factor;
```
pcarruscag, line 81: Fix this file please, or run clang-format it is difficult to see changes like this.

Common/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,1393 @@
...
+  solver_container[ZONE_0][INST_0][MESH_0][MESH_SOL]->InitiateComms(main_geometry, main_config, MESH_DISPLACEMENTS);
+  solver_container[ZONE_0][INST_0][MESH_0][MESH_SOL]->CompleteComms(main_geometry, main_config, MESH_DISPLACEMENTS);
+}
+
+void CDriverBase::ReadFFDInfo(CGeometry* geometry, CConfig* config, CFreeFormDefBox** FFDBox, string val_mesh_filename) {
+  string text_line, iTag;
+  ifstream mesh_file;
+  su2double CPcoord[3], coord[] = {0, 0, 0};
+  unsigned short degree[3], iFFDBox, iCornerPoints, iControlPoints, iMarker, iDegree, jDegree, kDegree, iChar,
+      LevelFFDBox, nParentFFDBox, iParentFFDBox, nChildFFDBox, iChildFFDBox, nMarker, *nCornerPoints, *nControlPoints;
```
pcarruscag, line 956: Copy pasted

pcarruscag (changes requested):

Common/include/drivers/CDriverBase.hpp
patelha57: `CDriverBase` is in Common because it is not exclusively used for SU2_CFD drivers. The SU2_DEF module has `CDeformationDriver` and `CDiscAdjDeformationDriver` that also inherit from `CDriverBase`. In the future, other SU2 modules can also build off of `CDriverBase`

patelha57: Noted about doxygen docs!

Common/src/drivers/CDriverBase.cpp
patelha57: Sure, we can discuss more streamlined ways to expose the FFD functionality to Python in the future. I will remove the FFD functions from the current PR and move it as part of PR #1750

Common/src/drivers/CDriverBase.cpp
patelha57: As discussed in a comment above, I will remove the FFD functions in CDriverBase from the current PR and move it as part of PR #1750. This function will be removed

Common/include/drivers/CDriverBase.hpp
pcarruscag: But it is including things from SU2_CFD... which defeats the purpose of having Common.
Move this to SU2_CFD and make the deformation drivers depend on SU2_CFD.

patelha57: CDriverBase was moved to SU2_CFD directory in commit [5c1ec48](https://github.com/su2code/SU2/pull/1300/commits/5c1ec48ef8f712b67afe88faa8011ace8282cfc6)

Common/src/drivers/CDriverBase.cpp
patelha57: FFD functions removed in commit [3c8f5cd](https://github.com/su2code/SU2/pull/1300/commits/3c8f5cd84eea3a870c6f3b0bf2b74fdbdfead498)

Common/src/drivers/CDriverBase.cpp
patelha57: Removed in commit [3c8f5cd](https://github.com/su2code/SU2/pull/1300/commits/3c8f5cd84eea3a870c6f3b0bf2b74fdbdfead498)

Common/src/drivers/CDriverBase.cpp
patelha57: Updated function to `GetNodes(iPoint)` in [4353a26](https://github.com/su2code/SU2/pull/1300/commits/4353a260a68413ef1f3f2068120e5eb1768b1cbf)

Common/src/drivers/CDriverBase.cpp
patelha57: Updated function to `GetMarkerVertices(iMarker, iVertex)` in [4353a26](https://github.com/su2code/SU2/pull/1300/commits/4353a260a68413ef1f3f2068120e5eb1768b1cbf)

Common/src/drivers/CDriverBase.cpp
patelha57: Updated function to `GetMarkerElementNodes(iMarker, iElem)` in [4353a26](https://github.com/su2code/SU2/pull/1300/commits/4353a260a68413ef1f3f2068120e5eb1768b1cbf)

SU2_CFD/src/python_wrapper_structure.cpp
patelha57: I am not sure... I haven't touched this code. We added some new functions related to laminar viscosity on the feature_pysu2_fsi_adjoint branch, so I will double check for PR #1750.

SU2_DEF/src/drivers/CDiscAdjDeformationDriver.cpp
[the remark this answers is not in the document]
patelha57: What would be the default case for this? There are other usages similar to this in CDriver.cpp and SU2_SOL.cpp without a default case. 

Should I replace all with a simple if statement or just ignore this QL alert?

patelha57:
Okay the regression tests are all finally passing! 
Can you advise on what to do about the CodeQL alerts? Any other comments or suggestions in general?

SU2_PY/FSI_tools/FSIInterface.py
```diff
@@ -565,8 +565,12 @@ def interfaceMapping(self,FluidSolver, SolidSolver, FSI_config):
             # Note that the fluid solver is separated in more processors outside the python script
             # thus when, from a core, we request for the vertices on the interface, we only obtain
             # those in that core
-            GlobalIndex = FluidSolver.GetVertexGlobalIndex(self.fluidInterfaceIdentifier, iVertex)
-            posx, posy, posz = FluidSolver.GetInitialMeshCoord(self.fluidInterfaceIdentifier, iVertex)
+            GlobalIndex = FluidSolver.GetMarkerNodes(self.fluidInterfaceIdentifier, iVertex)
+            if self.nDim == 2:
+                posx, posy = FluidSolver.GetMarkerInitialCoordinates(self.fluidInterfaceIdentifier, iVertex)
```
WallyMaier, line 570: Is GetMarkerInitialCoordinates a new function?
Can the function return all coordinates (instead of this statement in another python script?

patelha57: It is an updated implementation of what used to be `GetInitialMeshCoord()` in python_wrapper_structure.cpp. `GetMarkerInitialCoordinates()` can be found in [CDriverBase.cpp](https://github.com/su2code/SU2/blob/feature_pysu2_DEF/SU2_CFD/src/drivers/CDriverBase.cpp#L474), and it only returns the components that are relevant to the number of dimensions in the problem.

SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
  * \file driver_structure.hpp
```
pcarruscag, line 2:
```suggestion
 * \file CDriver.hpp
```

SU2_CFD/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,756 @@
...
+  for (auto iNode = 0u; iNode < nNode; iNode++) {
+    unsigned long iPoint = main_geometry->bound[iMarker][iElem]->GetNode(iNode);
+
+    values.push_back(main_geometry->nodes->GetGlobalIndex(iPoint));
+  }
+
+  return values;
+}
+
+unsigned long CDriverBase::GetNumberNodes() const { return main_geometry->GetnPoint(); }
```
pcarruscag, line 296: I commented on this already, why are the names different from c++?

SU2_CFD/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,756 @@
...
+  unsigned long nHalo = 0;
+
+  for (auto iPoint = 0ul; iPoint < nPoint; iPoint++) {
+    if (!(main_geometry->nodes->GetDomain(iPoint))) {
+      nHalo += 1;
+    }
+  }
+
+  return nHalo;
+}
```
pcarruscag, line 317: The number of halos is nPoint - nPointDomain by definition.

SU2_CFD/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,756 @@
...
+  }
+
+  return values;
+}
+
+bool CDriverBase::GetMarkerDomain(unsigned short iMarker, unsigned long iVertex) const {
+  auto iPoint = GetMarkerVertices(iMarker, iVertex);
+
+  return main_geometry->nodes->GetDomain(iPoint);
+}
```
pcarruscag, line 428: This API has too much boilerplate.
Either implement one function that returns a vector, or two functions, count + point-wise access.
And please just **mirror the c++ interface**...
For example, if you return the nodes of elements as global indices how do you read the solution at those nodes?
You are trying to make things too simple and convenient, which is good, except then it only works for one or two applications, exactly like our optimization scripts!
Do not apply additional transformations to the connectivity (domain bool and global index) which can otherwise be done as a single line in python... Because in doing so you automatically need to write one version of the function per transformation you want to offer to users, for every function!
Which results in too much code, and less functionality.
Just mirror the parts of the c++ interface that you think are useful, without adding more boilerplate to them, which can be expressed in single lines of python...

pcarruscag:
You can leave the "uncontrolled path" warnings

SU2_CFD/src/drivers/CDriverBase.cpp
patelha57: I will write some thoughts here but we can bring this up for discussion at a dev meeting.

The new Python API requirements emerged from efforts to couple SU2 with external solvers in MDAO frameworks. The functions implemented in this PR and to a larger extent in PR #1750 (which I consider to be part 2 of the Python API refactoring) were necessary for those coupling efforts. These new functions were also found to be critical for standalone SU2 Python use, like performing residual-based adjoint sensitivity verification, geometry manipulations, and parameter sweeps for far-field conditions

Our experience through two independent research efforts led to specific requirements for the Python interface:
- Enable easy in-memory integration of primal and adjoint solvers.
- Increased modularity with a highly flexible API for execution and queries.
- Reduce startup development costs for future Python-based efforts. These include the creation of modular interfaces allowing SU2 to be driven by another framework and providing access to appropriate data structure in a useable format.
- Intuitive interface. Facilitate searchability of functions without in-depth experience with SU2 and its source code. This is also why some function names have been deliberately chosen to not directly mirror the C++ interface. We gave careful consideration for the standardization of the function naming towards this goal.

These requirements shaped the current Python API. Yes there is boilerplate code, but we think the convenience and the flexibility that is provided to users for a variety of applications is worth it

patelha57: > For example, if you return the nodes of elements as global indices how do you read the solution at those nodes? 

PR #1750 has new functions in python_wrapper_stucture.cpp that return the solution. There are two overloaded variations of the same function called `GetStates()`, and one of them takes an input argument as the global node index

patelha57: > Do not apply additional transformations to the connectivity (domain bool and global index) which can otherwise be done as a single line in python... Because in doing so you automatically need to write one version of the function per transformation you want to offer to users, for every function! Which results in too much code, and less functionality.

Not sure what you mean by the additional transformations. Can you please elaborate?

Even if it can be done in one line of Python, the whole motivation behind these API updates is to remove that obstacle for users. If they want connectivities, they will get them in a suitable data format directly from the API. We found these new functions to be vital for our research applications and other SU2 Python users should find them to be useful as well

pcarruscag:
I'll try to break down each request to make it clearer.

Request: Use nomenclature consistent with the C++ interface (point instead of node, etc.)
Reason: People who know the C++ interface will know how to use the python wrapper straight away, people who learn how to use the python wrapper will be able to understand the C++ implementation (without needing a translation layer in their mind all the time)

Request: Return connectivity/adjacency as local indices instead of global
Reason: This is the natural way of referencing data in an MPI code, local indices correspond directly to data stored in the partition thereby making it more efficient to access (without global-to-local conversions, which are much more expensive than local-to-global). Furthermore, the local ordering strategy improves the efficiency of loops (over edges, neighboring points, etc.)

Request: Do not apply unnecessary "transformations" to the connectivity/adjacency as part of the API
Reason: Increases the maintenance burden and makes for a less versatile API. By returning the local indices that form a marker or element, it is trivial to retrieve any other data for those indices (the "transformation"). For example, the solution at those points, the point coordinates, whether the points are halo or domain points, the global indices, etc.
This way, by adding a **single** function to the API e.g. get the global index of a point of element, the functionality of the API grows a lot more because the user can apply the function to **any** connectivity/adjacency function.

Request: Do not offer too many overloads of the same function
Reason: Increases the maintenance burden and it's returning the data in structures that are not very efficient, namely vector of vectors. It is much easier to offer domain-wide data access on the python side, as a very small function that can be part of wrapper utilities (i.e. live only in python but not c++). It may even be more efficient since you make a numpy array/matrix instead of lists of lists.

SU2_CFD/src/drivers/CDriverBase.cpp
```diff
@@ -0,0 +1,756 @@
...
+  for (auto iNode = 0u; iNode < nNode; iNode++) {
+    unsigned long iPoint = main_geometry->elem[iElem]->GetNode(iNode);
+
+    values.push_back(main_geometry->nodes->GetGlobalIndex(iPoint));
+  }
+
+  return values;
+}
+
+vector<vector<unsigned long>> CDriverBase::GetMarkerElementNodes(unsigned short iMarker) const {
```
pcarruscag, line 266: I checked for myself and in python this is converted into tuple of tuples

pcarruscag:
I've been testing this a bit this afternoon, things look ok.
We can revisit the accessor functions for the entire domain once we can return the container by reference instead of copying the data first.
We can also look into code-generating accessors for markers for whatever volume accessors we define, once the names stabilize a bit, there are some inconsistencies in the naming of the old python wrapper functions.

This needs regression tests now.

pcarruscag (approved):

RudiFeiman:
I'm trying to build SU2 with NVIDIA HPC compiler.
I found error with meson about dependency problem.
I found cond_config.hpp without NVIDIA HPC macro check. Does this mean SU2 didn't consider NVIDIA HPC compiler?

PR #1310, opened 2021-06-24
Feature libROM for reduced order modelling

jtlau:
## Proposed Changes
Adding support to use the external library libROM (https://github.com/LLNL/libROM) for reduced order modeling with SU2.

libROM captures solution data and either computes the proper orthogonal decomposition (POD) over the data, or saves the solution "snapshot" in an hdf5 file for later POD computation. The POD matrix is also saved in an hdf5 file.

## Related Work
There are a couple projects using libROM:
- For projection-based model order reduction (PROM) in the branch feature_MOR
- It is also being used with the NEMO solver to further test the PROM capabilities
- libROM was used in a prior project that used the POD matrix as input to a neural network (to be published)

- libROM also has other features, like help choosing the hyper-parameter nodes (a PROM step) and under development are black-box modeling methods like dynamic mode decomposition (DMD)

Please let me know if you have any questions / comments / recommendations!

I have tested this on macOS and linux. 

## PR Checklist

## Steps to Compile:

1. Compile libROM
2. Compile SU2 with the following meson options:
`./meson.py build --prefix=/your/prefix -Denable-librom=true -Dlibrom_root=/path/to/libROM`
3. You may need to add the environment variable:
 `export LD_LIBRARY_PATH=/path/to/libROM/build/:$LD_LIBRARY_PATH`
4. You can test by adding `SAVE_LIBROM=YES` to the Quickstart config


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+      u_basis_generator->takeSample(u, t, dt);
+      delete[] u;
+   }
+   
+  /*--- End collection of data and save POD ---*/
+  
+   if (converged) {
+
+      if (!unsteady) {
+         double* u = new double[nPointDomain*nVar];
```
pcarruscag, line 4297: Do you want this feature to work with the discrete adjoint or does it only make sense to use with primal solvers?
(because there might be simpler ways to make this copy)

config_template.cfg
```diff
@@ -1692,3 +1692,20 @@ DEFINITION_DV= ( 1, 1.0 | airfoil | 0, 0.05 ); ( 1, 1.0 | airfoil | 0, 0.10 ); (
 %
 % Use combined objective within gradient evaluation: may reduce cost to compute gradients when using the adjoint formulation.
 OPT_COMBINE_OBJECTIVE = NO
+%
+% --------------------- LIBROM PARAMETERS -----------------------%
+%
```
pcarruscag, line 1697: Maybe leave a link to the library or any relevant info/presentation here.

meson.build
```diff
@@ -169,6 +169,21 @@ elif get_option('enable-openblas')
 
 endif
 
+if get_option('enable-librom')
+   
+   assert(get_option('librom_root')!='',
+       'Must specify librom folder (-Dlibrom_root=path/to/libROM)')
```
pcarruscag, line 175: What is their default installation folder?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -188,6 +191,9 @@ CSolver::~CSolver(void) {
 
   delete VerificationSolution;
 
+#ifdef HAVE_LIBROM
+  if (u_basis_generator != nullptr) u_basis_generator = nullptr;
+#endif
```
pcarruscag, line 196:
```suggestion
```
Not needed.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4326,6 +4337,17 @@ class CSolver {
    * \return Struct holding 4 su2doubles.
    */
   virtual StreamwisePeriodicValues GetStreamwisePeriodicValues() const { return StreamwisePeriodicValues(); }
+  
+#ifdef HAVE_LIBROM
```
pcarruscag, line 4341:
```suggestion
```
Minimize #ifdefs by providing empty methods when the feature is not active.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4326,6 +4337,17 @@ class CSolver {
...
+  /*!
+   * \brief Save snapshot or POD data using libROM
+   * \param[in] solver - Container vector with all the solutions.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] converged - Whether or not solution has converged.
+  */
+  void SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged);
+#endif
```
pcarruscag, line 4350:
```suggestion
```

SU2_CFD/src/drivers/CSinglezoneDriver.cpp
```diff
@@ -93,6 +93,13 @@ void CSinglezoneDriver::StartSolver() {
     /*--- Output the solution in files. ---*/
 
     Output(TimeIter);
+    
+    /*--- Save iteration solution for libROM ---*/
+    #ifdef HAVE_LIBROM
+        if (config_container[MESH_0]->GetSave_libROM()) {
+          solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->SavelibROM(solver_container[ZONE_0][INST_0][MESH_0], geometry_container[ZONE_0][INST_0][MESH_0], config_container[ZONE_0], StopCalc);
+        }
+    #endif
```
pcarruscag, line 102:
```suggestion
    if (config_container[MESH_0]->GetSave_libROM()) {
      solver_container[ZONE_0][INST_0][MESH_0][FLOW_SOL]->SavelibROM(solver_container[ZONE_0][INST_0][MESH_0], geometry_container[ZONE_0][INST_0][MESH_0], config_container[ZONE_0], StopCalc);
    }
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
                    string("It could be empty lines at the end of the file."), CURRENT_FUNCTION);
   }
 }
+
+#ifdef HAVE_LIBROM
+void CSolver::SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged) {
+
```
pcarruscag, line 4221:
```suggestion
void CSolver::SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged) {
#ifdef HAVE_LIBROM

```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+         u_basis_generator->writeSnapshot();
+      }
+      std::cout << "Computing SVD" << std::endl;
+      int rom_dim = u_basis_generator->getSpatialBasis()->numColumns();
+      std::cout << "Basis dimension: " << rom_dim << std::endl;
+      u_basis_generator->endSamples();
+      std::cout << "ROM Sampling ended" << std::endl;
+   }
+}
+#endif
```
pcarruscag, line 4323:
```suggestion
#else
  SU2_MPI::Error("SU2 was not compiled with libROM support.", CURRENT_FUNCTION);
#endif
}
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
                    string("It could be empty lines at the end of the file."), CURRENT_FUNCTION);
   }
 }
+
+#ifdef HAVE_LIBROM
+void CSolver::SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged) {
+  
+  unsigned long iPoint, total_index;
+  unsigned short iVar;
```
pcarruscag, line 4223:
```suggestion
```
Use local variables for loop counts etc.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+  
+  bool unsteady            = (config->GetTime_Marching() != TIME_MARCHING::STEADY);
+  string filename          = config->GetlibROMbase_FileName();
+  unsigned short pod_basis = config->GetKind_PODBasis();
+  unsigned long TimeIter   = config->GetTimeIter();
+  unsigned long nTimeIter  = config->GetnTime_Iter();
+  int maxBasisDim          = config->GetMax_BasisDim();
+  int dim = int(nPointDomain * nVar);
```
pcarruscag, line 4231: Better check for unsteady with GetTime_Domain() (or something similar, can't recall the exact name).

const, or const auto as appropriate, the things you get from the config.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+  bool incremental = false;
+  
+  // Get solver nodes
+  CVariable* nodes = GetNodes();
```
pcarruscag, line 4235: In CSolver you have base_nodes

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4215,109 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+    
+    /*--- Define SVD basis generator ---*/
+    int timesteps = (int)(nTimeIter - TimeIter);
+    CAROM::Options svd_options = CAROM::Options(dim, timesteps, -1,
+                                                false, true).setMaxBasisDimension(int(maxBasisDim));
+    
+    if (pod_basis == STATIC_POD) {
+      std::cout << "Creating static basis generator." << std::endl;
+      if (unsteady) {
+        std::cout << "Incremental basis generator recommended for unsteady simulations." << std::endl;
```
pcarruscag, line 4247: Your messages will be printed multiple times in MPI runs, and you will have multiple "basis generators", is that ok? Or are you looking for a better integration with MPI, we can help with that.

Common/include/CConfig.hpp
```diff
@@ -9292,4 +9300,39 @@ class CConfig {
...
+   * \brief Get maximum number of POD basis dimensions (default: 100).
+   * \return Maximum number of POD basis vectors.
+   */
+  unsigned short GetMax_BasisDim(void) const { return maxBasisDim; }
+  
+  /*!
+   * \brief Get number of POD modes desired.
+   * \return Number of POD modes desired.
+   */
+  //unsigned short GetnPOD_Modes(void) const { return nPOD_Modes; }
```
pcarruscag, line 9331: No commented out code please.

Common/include/option_structure.hpp
```diff
@@ -2225,6 +2225,18 @@ struct StreamwisePeriodicValues {
   su2double Streamwise_Periodic_InletTemperature;   /*!< \brief Area avg static Temp [K] at the periodic inlet. Used for adaptive outlet heatsink. */
 };
 
+/*!
+ * \brief Type of POD basis generation (for use with libROM)
+ */
+enum ENUM_POD_KIND {
+  STATIC_POD = 1,            /*!< \brief Use static SVD for POD basis generation. */
+  INCREMENTAL_POD = 2        /*!< \brief Use incremental SVD for POD basis generation. */
+};
```
pcarruscag, line 2234:
```suggestion
enum class POD_KIND {
  STATIC,            /*!< \brief Use static SVD for POD basis generation. */
  INCREMENTAL        /*!< \brief Use incremental SVD for POD basis generation. */
};
```
We have made a change recently to use `enum class` when possible to keep up with c++ best practices.
You will need POD_KIND instead of unsigned short in some places and e.g. POD_KIND::STATIC instead of STATIC_POD. Easy changes.

pcarruscag (commented):
Hi Jessie, thank you for this contribution.
It is not my specialty but I recall one or two interesting presentations.
Some general questions before we get into code details:
Are the results of this ready to use? Or do they need some additional post-processing steps to make sense? If so it would be good to have an example with any scripts needed.
Our current gold standard has been set by @TobiKattmann, see TestCases/incomp_navierstokes/streamwise_periodic.

If you were already planning to add a tutorial then nevermind the above.

jtlau:
Thanks Pedro for the suggestions! I'm working on the changes now.

meson.build
jtlau: What I'm looking for here is where the user decided to install libROM. The default location of the shared library is in libROM_install_directory/build/

jtlau: There is not a pkg-config file, but that's something I could request from the libROM folks

SU2_CFD/src/solvers/CSolver.cpp
jtlau: This feature would only make sense with the primal solvers... in that case, is this way OK or is there a better way?

SU2_CFD/src/solvers/CSolver.cpp
jtlau: Oh yes, I meant for the messages to only be printed once. I updated those with (rank == MASTER_NODE) if statements. The multiple basis generators is intentional

jtlau:
There are some post-processing steps involved, so I think it would be a good idea for me to include an example. 

The features in this PR are one step to a bigger projection-based ROM process. The code for the bigger process is in feature_MOR, and will take some time for me to prepare the code from "make it work" to "actually good code". 

However, this feature-PR can be used by itself by anyone who wants to efficiently compute the POD of SU2 simulation data, and then use the POD for other modeling strategies.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4211,112 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+      for (iPoint = 0; iPoint < nPointDomain; iPoint++) {
+        unsigned long globalPoint = geometry->nodes->GetGlobalIndex(iPoint);
+        auto Coord = geometry->nodes->GetCoord(iPoint);
+        f << Coord[0] << ", " << Coord[1] << ", " << globalPoint << "\n";
+      }
+    
+    f.close();
+  }
+
+  if (unsteady) {
```
jtlau, line 4270: I need to add an option here to capture snapshots at intervals

meson.build
pcarruscag: Got it, even if they add pkg-config support it may be good to have this fallback.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4211,111 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+  
+    if (!unsteady) {
+       double* u = new double[nPointDomain*nVar];
+       for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+         for (unsigned short iVar = 0; iVar < nVar; iVar++) {
+             unsigned long total_index = iPoint*nVar + iVar;
+             u[total_index] = base_nodes->GetSolution(iPoint, iVar);
+         }
+       }
+
```
pcarruscag, line 4295:
```suggestion
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4211,111 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+         for (unsigned short iVar = 0; iVar < nVar; iVar++) {
+             unsigned long total_index = iPoint*nVar + iVar;
+             u[total_index] = base_nodes->GetSolution(iPoint, iVar);
+         }
+       }
+ 
+       // dt is different for each node, so just use a placeholder dt
+       double dt = base_nodes->GetDelta_Time(0);
+       double t = dt*TimeIter;
+       u_basis_generator->takeSample(u, t, dt);
```
pcarruscag, line 4299:
```suggestion
       u_basis_generator->takeSample(base_nodes->GetSolution().data(), t, dt);
```

This should work but it will not compile with AD, hence the modified #ifdef.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4211,111 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
                    string("It could be empty lines at the end of the file."), CURRENT_FUNCTION);
   }
 }
+
+void CSolver::SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged) {
+  
+#ifdef HAVE_LIBROM
```
pcarruscag, line 4217:
```suggestion
#if defined(HAVE_LIBROM) && !defined(CODI_FORWARD_TYPE) && !defined(CODI_REVERSE_TYPE)
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -119,6 +119,9 @@ CSolver::CSolver(bool mesh_deform_mode) : System(mesh_deform_mode) {
   nPrimVarGrad = 0;
   nPrimVar     = 0;
 
+#ifdef HAVE_LIBROM
+  u_basis_generator  = nullptr;
+#endif
```
pcarruscag, line 124:
```suggestion
```
The default constructor of unique_ptr already initializes as nullptr.

pcarruscag (commented):
Yup a short example would be nice, but if there is more work coming we can wait until then.

Common/include/option_structure.hpp
```diff
@@ -2225,6 +2225,18 @@ struct StreamwisePeriodicValues {
...
 };
 
+/*!
+ * \brief Type of POD basis generation (for use with libROM)
+ */
+enum class POD_KIND {
+  STATIC,            /*!< \brief Use static SVD for POD basis generation. */
+  INCREMENTAL,       /*!< \brief Use incremental SVD for POD basis generation. */
+};
+static const MapType<string, POD_KIND> POD_Map = {
```
pcarruscag, line 2235:
```suggestion
static const MapType<std::string, POD_KIND> POD_Map = {
```

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4326,7 +4337,15 @@ class CSolver {
...
-
+  
+  /*!
+   * \brief Save snapshot or POD data using libROM
+   * \param[in] solver - Container vector with all the solutions.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] converged - Whether or not solution has converged.
+  */
+  void SavelibROM(CSolver** solver, CGeometry *geometry, CConfig *config, bool converged);
```
pcarruscag, line 4348:
```suggestion
   * \param[in] geometry - Geometrical definition of the problem.
   * \param[in] config - Definition of the particular problem.
   * \param[in] converged - Whether or not solution has converged.
  */
  void SavelibROM(const CGeometry *geometry, const CConfig *config, bool converged);
```
It looks like you don't need the solver container in that function? Or are you planning to add turbulence variables at some point?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+      }
+    
+    f.close();
+  }
+
+  if (unsteady && (TimeIter % save_freq == 0)) {
+    // give solution and time steps to libROM:
+    double dt = config->GetDelta_UnstTime();
+    double t =  config->GetCurrent_UnstTime();
+    u_basis_generator->takeSample(const_cast<double*>(base_nodes->GetSolution().data()), t, dt);
```
pcarruscag, line 4269: It sounded like you were in touch with the libROM folks, you might ask them to make the inputs of their functions `const` (because having to const_cast is not ideal).
```suggestion
    u_basis_generator->takeSample(const_cast<su2double*>(base_nodes->GetSolution().data()), t, dt);
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+        f << Coord[0] << ", " << Coord[1] << ", " << globalPoint << "\n";
+      }
+    
+    f.close();
+  }
+
+  if (unsteady && (TimeIter % save_freq == 0)) {
+    // give solution and time steps to libROM:
+    double dt = config->GetDelta_UnstTime();
+    double t =  config->GetCurrent_UnstTime();
```
pcarruscag, line 4268:
```suggestion
    su2double dt = config->GetDelta_UnstTime();
    su2double t =  config->GetCurrent_UnstTime();
```
We never use `double` explicitly unless extremely necessary, for consistency with the rest of the code, and to avoid confusion it is better to always try to use su2double.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+   
+  /*--- End collection of data and save POD ---*/
+  
+  if (converged) {
+  
+    if (!unsteady) {
+       // dt is different for each node, so just use a placeholder dt
+       double dt = base_nodes->GetDelta_Time(0);
+       double t = dt*TimeIter;
+       u_basis_generator->takeSample(const_cast<double*>(base_nodes->GetSolution().data()), t, dt);
```
pcarruscag, line 4280:
```suggestion
       su2double dt = base_nodes->GetDelta_Time(0);
       su2double t = dt*TimeIter;
       u_basis_generator->takeSample(const_cast<su2double*>(base_nodes->GetSolution().data()), t, dt);
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+  const unsigned long nTimeIter  = config->GetnTime_Iter();
+  const int maxBasisDim          = config->GetMax_BasisDim();
+  const int save_freq            = config->GetRom_SaveFreq();
+  int dim = int(nPointDomain * nVar);
+  bool incremental = false;
+
+  if (!u_basis_generator) {
+    
+    /*--- Define SVD basis generator ---*/
+    int timesteps = (int)(nTimeIter - TimeIter);
```
pcarruscag, line 4227:
```suggestion
    auto timesteps = static_cast<int>(nTimeIter - TimeIter);
```
The preferred syntax according to modern C++ experts.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+    
+    u_basis_generator.reset(new CAROM::BasisGenerator(
+      svd_options, incremental,
+      filename));
+    
+    // Print nodes for each rank for now
+    std::cout << "nPointDomain: " << nPointDomain << " for rank: " << rank << std::endl;
+    
+    // Save mesh ordering
+    std::ofstream f;
```
pcarruscag, line 4254: Do you mean to print this to the file? Or really to the screen?
If the latter, it may be better to gather and do all the printing by the master node (in order).
Or instead terminate the lines with "\n" instead of "endl", because "endl" forces a flush and you may end up with mixedup lines ("\n" does not guarantee that you won't, it just makes it less likely).

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -4209,3 +4208,93 @@ void CSolver::BasicLoadRestart(CGeometry *geometry, const CConfig *config, const
...
+    f.open(filename + "_mesh_" + to_string(rank) + ".csv");
+      for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+        unsigned long globalPoint = geometry->nodes->GetGlobalIndex(iPoint);
+        auto Coord = geometry->nodes->GetCoord(iPoint);
+        f << Coord[0] << ", " << Coord[1] << ", " << globalPoint << "\n";
```
pcarruscag, line 4259: No 3D applications for this?

pcarruscag (commented):
Thank you for the changes, just a couple more things I noticed.

SU2_CFD/src/solvers/CSolver.cpp
jtlau: Yes! It should have been general for either 2D or 3D apps. Fixed

SU2_CFD/include/solvers/CSolver.hpp
jtlau: Good catch - it is not needed

SU2_CFD/src/solvers/CSolver.cpp
jtlau: I realized this part was unnecessary... the value nPointDomain for that rank is the same as # of data points in each of the .csv files created just below. I was using this for debugging mostly

pcarruscag (approved):
LGTM

PR #1311, opened 2021-06-25
Fix inverse design Cp function

pcarruscag:
## Proposed Changes
It had mpi problems and others: https://www.cfd-online.com/Forums/su2/236677-mpi-returns-wrong-objective-function-values.html

The target Cp file was also not formatted correctly, the residuals will change because of that.

## PR Checklist


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -695,123 +695,96 @@ void CFlowOutput::SetRotatingFrameCoefficients(CConfig *config, CSolver *flow_so
...
-            Point2Vertex[iPoint][0] = iMarker;
-            Point2Vertex[iPoint][1] = iVertex;
-            PointInDomain[iPoint] = true;
-            solver->SetCPressureTarget(iMarker, iVertex, 0.0);
-          }
-
-        }
-      }
-    }
+  if (!Surface_file.good()) {
```
WallyMaier, line 715: Nice!

WallyMaier (approved):
LGTM!

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -695,123 +695,96 @@ void CFlowOutput::SetRotatingFrameCoefficients(CConfig *config, CSolver *flow_so
...
+        if (!geometry->nodes->GetDomain(iPoint)) continue;
 
-          Area = GeometryToolbox::Norm(nDim, Normal);
+        const auto Cp = solver->GetCPressure(iMarker, iVertex);
+        const auto CpTarget = solver->GetCPressureTarget(iMarker, iVertex);
 
-          PressDiff += Area * (CpTarget - Cp) * (CpTarget - Cp);
-        }
+        const auto Normal = geometry->vertex[iMarker][iVertex]->GetNormal();
+        const auto Area = GeometryToolbox::Norm(nDim, Normal);
```
TobiKattmann, line 776: Just a note: We need to be careful on future PR's with these su2double auto's because like so as a single assignment  it is correct but is looks straight forward e.g. to extend it to `const auto Cp_rho = solver->GetCPressure(iMarker, iVertex) * rho` (rho falling from the sky) which would then be incorrect because the auto for su2double active then evaluates to an expression.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -695,123 +695,96 @@ void CFlowOutput::SetRotatingFrameCoefficients(CConfig *config, CSolver *flow_so
...
+    const auto Boundary = config->GetMarker_All_KindBC(iMarker);
 
-          Normal = geometry->vertex[iMarker][iVertex]->GetNormal();
+    if (config->GetSolid_Wall(iMarker) || (Boundary == NEARFIELD_BOUNDARY)) {
+      for (auto iVertex = 0ul; iVertex < geometry->GetnVertex(iMarker); iVertex++) {
 
-          Cp = solver->GetCPressure(iMarker, iVertex);
-          CpTarget = solver->GetCPressureTarget(iMarker, iVertex);
+        const auto iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
+        if (!geometry->nodes->GetDomain(iPoint)) continue;
```
TobiKattmann, line 770: Fai (For anyone's interest): this is the missing check of domain points that is mentioned in the cfd-online post

TobiKattmann (commented):
(I missed the last bus to this PR 🚏 🚌  ... so I am arriving just now 🐌   )

Thanks for getting into the details of this implementation and fixing this @pcarruscag 💐

TestCases/parallel_regression_AD.py
```diff
@@ -69,9 +69,9 @@ def main():
     discadj_arina2k.cfg_dir      = "disc_adj_euler/arina2k"
     discadj_arina2k.cfg_file     = "Arina2KRS.cfg"
     discadj_arina2k.test_iter    = 20
-    discadj_arina2k.test_vals    = [2.189902, 1.635938, 47258.000000, 0.000000]
+    discadj_arina2k.test_vals    = [-3.111181, -3.501516, 6.8705e-02, 0]
```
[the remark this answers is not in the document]
pcarruscag: garbage in garbage out as with all things

SU2_CFD/src/output/CFlowOutput.cpp
pcarruscag: Ah! Because this function is always called and the result is always available as one of the outputs :shrug: and so it cannot throw an error.

PR #1312, opened 2021-06-27
Another charge against pointer to pointer

pcarruscag:
## Proposed Changes
Introduce a "matrix view" type to avoid ** for example when passing gradients of primitives into the numerics.

## PR Checklist


Common/include/containers/container_decorators.hpp
```diff
@@ -30,6 +30,43 @@
...
+  CMatrixView(T* ptr = nullptr, Index cols = 0) : m_ptr(ptr), m_cols(cols) {}
+
+  template<class U> friend class CMatrixView;
+  template<class U>
+  CMatrixView(const CMatrixView<U>& other) : m_ptr(other.m_ptr), m_cols(other.m_cols) {}
+
+  explicit CMatrixView(const su2matrix<Scalar>& mat) : m_ptr(mat.data()), m_cols(mat.cols()) {}
+
+  template<class U = T, su2enable_if<!std::is_const<U>::value> = 0>
+  explicit CMatrixView(su2matrix<Scalar>& mat) : m_ptr(mat.data()), m_cols(mat.cols()) {}
```
pcarruscag, line 56:
```suggestion
  explicit CMatrixView(su2matrix<Scalar>& mat) : m_ptr(mat.data()), m_cols(mat.cols()) {}

  template<class U = T, su2enable_if<std::is_const<U>::value> = 0>
  explicit CMatrixView(const su2matrix<Scalar>& mat) : m_ptr(mat.data()), m_cols(mat.cols()) {}
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1260,7 +1259,7 @@ void CFVMFlowSolverBase<V, R>::BC_Sym_Plane(CGeometry* geometry, CSolver** solve
                 GradNormVel[iDim] * UnitNormal[iVar] + GradTangVel[iDim] * Tangential[iVar];
 
         /*--- Set the primitive gradients of the boundary and reflected state. ---*/
-        visc_numerics->SetPrimVarGradient(nodes->GetGradient_Primitive(iPoint), Grad_Reflected);
+        visc_numerics->SetPrimVarGradient(nodes->GetGradient_Primitive(iPoint), CMatrixView<su2double>(Grad_Reflected));
```
pcarruscag, line 1262: Matrix types have an explicit conversion to this "matrix view", i.e. you can pass `su2activematrix` to CNumerics instead of a `su2double**`

SU2_CFD/src/interfaces/fsi/CFlowTractionInterface.cpp
```diff
@@ -189,7 +189,7 @@ void CFlowTractionInterface::GetDonor_Variable(CSolver *flow_solution, CGeometry
     su2double Viscosity = flow_nodes->GetLaminarViscosity(Point_Flow);
 
     su2double tau[3][3];
-    CNumerics::ComputeStressTensor(nVar, tau, flow_nodes->GetGradient_Primitive(Point_Flow)+1,Viscosity);
+    CNumerics::ComputeStressTensor(nVar, tau, flow_nodes->GetGradient_Primitive(Point_Flow,1), Viscosity);
```
pcarruscag, line 192: Preferred way to get the gradient with an offset, although the "+" arithmetic still works.

SU2_CFD/src/output/CFlowCompOutput.cpp
```diff
@@ -553,7 +553,7 @@ void CFlowCompOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolv
     } else {
       SetVolumeOutputValue("VORTICITY", iPoint, Node_Flow->GetVorticity(iPoint)[2]);
     }
-    SetVolumeOutputValue("Q_CRITERION", iPoint, GetQ_Criterion(&(Node_Flow->GetGradient_Primitive(iPoint)[1])));
+    SetVolumeOutputValue("Q_CRITERION", iPoint, GetQ_Criterion(Node_Flow->GetGradient_Primitive(iPoint,1)));
```
pcarruscag, line 556: But this (taking the address) will no longer work.

Common/include/containers/container_decorators.hpp
```diff
@@ -105,42 +157,11 @@ class C3DContainerDecorator {
...
-
-/*!
- * \class CVectorOfMatrix
- * \brief This contrived container is used to store small matrices in a contiguous manner
- *        but still present the "su2double**" interface to the outside world.
- *        The "interface" part should be replaced by something more efficient, e.g. a "matrix view".
- */
-class CVectorOfMatrix: public C3DDoubleMatrix {
-private:
-  su2matrix<Scalar*> interface;
```
pcarruscag, line 121: The main motivation for this change is to store "vectors of matrices" i.e. gradients in a contiguous way without needing an extra matrix of pointers to provide the `su2double**` compatibility with other areas of the code.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -659,6 +627,58 @@ class CNumerics {
     }
   }
 
+  /*!
+   * \brief Perturb the Reynolds stress tensor based on parameters.
```
pcarruscag, line 631:
```suggestion
   * \brief Project average gradient onto normal (with or w/o correction) for viscous fluxes of scalar quantities.
```

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -659,6 +627,58 @@ class CNumerics {
     }
   }
 
+  /*!
+   * \brief Project average gradient onto normal (with or w/o correction) for viscous fluxes of scalar quantities.
```
pcarruscag, line 631: To reduce duplication

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1053,8 +1053,7 @@ void CFVMFlowSolverBase<V, R>::BC_Sym_Plane(CGeometry* geometry, CSolver** solve
       Tangential[MAXNDIM] = {0.0}, GradNormVel[MAXNDIM] = {0.0}, GradTangVel[MAXNDIM] = {0.0};
 
   /*--- Allocation of primitive gradient arrays for viscous fluxes. ---*/
-  su2double** Grad_Reflected = new su2double*[nPrimVarGrad];
-  for (iVar = 0; iVar < nPrimVarGrad; iVar++) Grad_Reflected[iVar] = new su2double[nDim];
+  su2activematrix Grad_Reflected(nPrimVarGrad, nDim);
```
WallyMaier, line 1056: This looks much nicer, and less of a memory headache. 👍

SU2_CFD/src/numerics/NEMO/convection/msw.cpp
```diff
@@ -198,7 +198,8 @@ CNumerics::ResidualType<> CUpwMSW_NEMO::ComputeResidual(const CConfig *config) {
                                                       epsilon*epsilon));
 
   /*--- Compute projected P, invP, and Lambda ---*/
-  CreateBasis(UnitNormal);
+  su2double l[MAXNDIM], m[MAXNDIM];
```
WallyMaier, line 201: thanks for untangling some of this stuff.  Its on the list to go through and completely abstract away....eventually.

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -674,7 +674,7 @@ CNumerics::ResidualType<> CSourceWindGust::ComputeResidual(const CConfig* config
   smx = rho*(du_gust_dt + (u+u_gust)*du_gust_dx + (v+v_gust)*du_gust_dy);
   smy = rho*(dv_gust_dt + (u+u_gust)*dv_gust_dx + (v+v_gust)*dv_gust_dy);
   //smz = rho*(dw_gust_dt + (u+u_gust)*dw_gust_dx + (v+v_gust)*dw_gust_dy) + (w+w_gust)*dw_gust_dz;
-
```
WallyMaier, line 677: 👍

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -91,20 +87,21 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
 
   /*--- Adding artificial dissipation to stabilize the centered scheme ---*/
 
-  Diff_Lapl[0] = Und_Lapl_i[0]-Und_Lapl_j[0];
+  su2double Diff_Lapl = Und_Lapl_i[0]-Und_Lapl_j[0];
+  su2double Area2 = GeometryToolbox::SquaredNorm(nDim, Normal);
 
-  SoundSpeed_i = sqrt(ProjVelocity_i*ProjVelocity_i + (BetaInc2_i/DensityInc_i)*Area*Area);
-  SoundSpeed_j = sqrt(ProjVelocity_j*ProjVelocity_j + (BetaInc2_j/DensityInc_j)*Area*Area);
+  su2double SoundSpeed_i = sqrt(pow(ProjVelocity_i,2) + (BetaInc2_i/DensityInc_i)*Area2);
```
WallyMaier, line 93: Just for my own knowledge, is pow(x,2) superior to x*x?

WallyMaier (approved):
@pcarruscag thanks for the work!

This looks good to me.

SU2_CFD/src/numerics/heat.cpp
pcarruscag: I think it is more expressive, saves you the trouble of reading two variable names to make sure they are the same.
In terms of performance, the pow function is horrible, however, pow(x,2) always gets optimized to x*x.
Larger integer powers are only optimized in that way if some -ffast-math optimizations are allowed, since e.g. x*x*x is not a strict way to compute pow(x,3) (as per the floating-point standard).
That type of optimization is fairly innocuous in double-precision, the intel compilers actually do them by default.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -659,6 +627,58 @@ class CNumerics {
...
+   */
+  template<class Vec1, class Vec2, class Mat>
+  FORCEINLINE static su2double ComputeProjectedGradient(int nDim, int nVar, const Vec1& normal,
+                                                        const Vec1& coord_i, const Vec1& coord_j,
+                                                        const Mat& grad_i, const Mat& grad_j,
+                                                        bool correct,
+                                                        const Vec2& var_i, const Vec2& var_j,
+                                                        su2double* projNormal,
+                                                        su2double* projCorrected) {
+    nDim = (nDim > 2)? 3 : 2;
```
pcarruscag, line 654: This keeps the compiler from going crazy with loop unrolling thinking that nDim is large, and then creating unrolled code that is never used.

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1932,8 +1932,8 @@ void CDriver::Numerics_Preprocessing(CConfig *config, CGeometry **geometry, CSol
     /*--- Definition of the viscous scheme for each equation and mesh level ---*/
     for (iMGlevel = 0; iMGlevel <= config->GetnMGLevels(); iMGlevel++) {
 
-      numerics[iMGlevel][HEAT_SOL][visc_term] = new CAvgGradCorrected_Heat(nDim, nVar_Heat, config);
-      numerics[iMGlevel][HEAT_SOL][visc_bound_term] = new CAvgGrad_Heat(nDim, nVar_Heat, config);
+      numerics[iMGlevel][HEAT_SOL][visc_term] = new CAvgGrad_Heat(nDim, nVar_Heat, config, true);
+      numerics[iMGlevel][HEAT_SOL][visc_bound_term] = new CAvgGrad_Heat(nDim, nVar_Heat, config, false);
```
TobiKattmann, line 1936: nice

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -777,7 +777,7 @@ CNumerics::ResidualType<> CSourceIncStreamwisePeriodic_Outlet::ComputeResidual(c
 
   /*--- Force the area avg inlet Temp to match the Inc_Temperature_Init with additional residual contribution ---*/
   const su2double delta_T = SPvals.Streamwise_Periodic_InletTemperature - config->GetInc_Temperature_Init()/config->GetTemperature_Ref();
-  residual[nDim+1] += 0.5 * abs(local_Massflow) * SpecificHeat_i * delta_T;
+  residual[nDim+1] += 0.5 * abs(local_Massflow) * Cp_i * delta_T;
```
TobiKattmann, line 780: yep, it was me 👮 :prison: (that emoji doesnt exists :( )

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -71,16 +56,27 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
...
+      ProjVelocity_i += V_i[iDim+1]*Normal[iDim];
+      ProjVelocity_j += V_j[iDim+1]*Normal[iDim];
+    }
   }
-  Area = GeometryToolbox::Norm(nDim, Normal);
 
   /*--- Computing the second order centered scheme part ---*/
 
-  ProjVelocity = 0.5*(ProjVelocity_i+ProjVelocity_j);
+  su2double ProjVelocity = 0.5*(ProjVelocity_i+ProjVelocity_j);
```
TobiKattmann, line 79:
```suggestion
  const su2double ProjVelocity = 0.5*(ProjVelocity_i+ProjVelocity_j);
```

TobiKattmann: any reason to not const all of those? Does the compiler do the necessary optimization anyway because the scope is so narrow?

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -91,20 +87,21 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
 
   /*--- Adding artificial dissipation to stabilize the centered scheme ---*/
 
-  Diff_Lapl[0] = Und_Lapl_i[0]-Und_Lapl_j[0];
+  su2double Diff_Lapl = Und_Lapl_i[0]-Und_Lapl_j[0];
+  su2double Area2 = GeometryToolbox::SquaredNorm(nDim, Normal);
```
TobiKattmann, line 91:
```suggestion
  const su2double Diff_Lapl = Und_Lapl_i[0]-Und_Lapl_j[0];
  const su2double Area2 = GeometryToolbox::SquaredNorm(nDim, Normal);
```

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -91,20 +87,21 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
...
 
-  SoundSpeed_i = sqrt(ProjVelocity_i*ProjVelocity_i + (BetaInc2_i/DensityInc_i)*Area*Area);
-  SoundSpeed_j = sqrt(ProjVelocity_j*ProjVelocity_j + (BetaInc2_j/DensityInc_j)*Area*Area);
+  su2double SoundSpeed_i = sqrt(pow(ProjVelocity_i,2) + (BetaInc2_i/DensityInc_i)*Area2);
+  su2double SoundSpeed_j = sqrt(pow(ProjVelocity_j,2) + (BetaInc2_j/DensityInc_j)*Area2);
```
TobiKattmann, line 94:
```suggestion
  const su2double SoundSpeed_i = sqrt(pow(ProjVelocity_i,2) + (BetaInc2_i/DensityInc_i)*Area2);
  const su2double SoundSpeed_j = sqrt(pow(ProjVelocity_j,2) + (BetaInc2_j/DensityInc_j)*Area2);
```

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -91,20 +87,21 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
...
 
-  Local_Lambda_i = fabs(ProjVelocity_i)+SoundSpeed_i;
-  Local_Lambda_j = fabs(ProjVelocity_j)+SoundSpeed_j;
-  MeanLambda = 0.5*(Local_Lambda_i+Local_Lambda_j);
+  su2double Local_Lambda_i = fabs(ProjVelocity_i)+SoundSpeed_i;
+  su2double Local_Lambda_j = fabs(ProjVelocity_j)+SoundSpeed_j;
+  su2double MeanLambda = 0.5*(Local_Lambda_i+Local_Lambda_j);
```
TobiKattmann, line 98:
```suggestion
  const su2double Local_Lambda_i = fabs(ProjVelocity_i)+SoundSpeed_i;
  const su2double Local_Lambda_j = fabs(ProjVelocity_j)+SoundSpeed_j;
  const su2double MeanLambda = 0.5*(Local_Lambda_i+Local_Lambda_j);
```

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -91,20 +87,21 @@ void CCentSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jac
...
 
-  val_residual[0] += -Param_Kappa_4*Diff_Lapl[0]*MeanLambda;
+  val_residual[0] += -Param_Kappa_4*Diff_Lapl*MeanLambda;
 
   if (implicit) {
-    cte_0 = Param_Kappa_4*su2double(Neighbor_i+1)*MeanLambda;
-    cte_1 = Param_Kappa_4*su2double(Neighbor_j+1)*MeanLambda;
+    su2double cte_0 = Param_Kappa_4*su2double(Neighbor_i+1)*MeanLambda;
+    su2double cte_1 = Param_Kappa_4*su2double(Neighbor_j+1)*MeanLambda;
```
TobiKattmann, line 104:
```suggestion
    const su2double cte_0 = Param_Kappa_4*su2double(Neighbor_i+1)*MeanLambda;
    const su2double cte_1 = Param_Kappa_4*su2double(Neighbor_j+1)*MeanLambda;
```

TobiKattmann: github wouldnt let me do the whole thing at once, that why it is so chunky

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -149,14 +131,24 @@ void CUpwSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jaco
...
-  for (iDim = 0; iDim < nDim; iDim++) {
-    Velocity_i[iDim] = V_i[iDim+1];
-    Velocity_j[iDim] = V_j[iDim+1];
-    q_ij += 0.5*(Velocity_i[iDim]+Velocity_j[iDim])*Normal[iDim];
+  su2double q_ij = 0.0;
+
+  if (dynamic_grid) {
+    for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+      su2double Velocity_i = V_i[iDim+1] - GridVel_i[iDim];
+      su2double Velocity_j = V_j[iDim+1] - GridVel_j[iDim];
```
TobiKattmann, line 139:
```suggestion
      const su2double Velocity_i = V_i[iDim+1] - GridVel_i[iDim];
      const su2double Velocity_j = V_j[iDim+1] - GridVel_j[iDim];
```

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -149,14 +131,24 @@ void CUpwSca_Heat::ComputeResidual(su2double *val_residual, su2double **val_Jaco
...
+  else {
+    for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+      q_ij += 0.5*(V_i[iDim+1]+V_j[iDim+1])*Normal[iDim];
+    }
   }
 
-  a0 = 0.5*(q_ij+fabs(q_ij));
-  a1 = 0.5*(q_ij-fabs(q_ij));
+  su2double a0 = 0.5*(q_ij+fabs(q_ij));
+  su2double a1 = 0.5*(q_ij-fabs(q_ij));
```
TobiKattmann, line 150:
```suggestion
  const su2double a0 = 0.5*(q_ij+fabs(q_ij));
  const su2double a1 = 0.5*(q_ij-fabs(q_ij));
```

SU2_CFD/src/numerics/heat.cpp
```diff
@@ -204,31 +180,15 @@ void CAvgGrad_Heat::ComputeResidual(su2double *val_residual, su2double **Jacobia
   AD::SetPreaccIn(ConsVar_Grad_i[0],nDim); AD::SetPreaccIn(ConsVar_Grad_j[0],nDim);
   AD::SetPreaccIn(Thermal_Diffusivity_i); AD::SetPreaccIn(Thermal_Conductivity_j);
 
-  Thermal_Diffusivity_Mean = 0.5*(Thermal_Diffusivity_i + Thermal_Diffusivity_j);
+  su2double NormalGrad[nVar], CorrectedGrad[nVar];
 
-  /*--- Compute vector going from iPoint to jPoint ---*/
+  auto proj_vector_ij = ComputeProjectedGradient(nDim, nVar, Normal, Coord_i, Coord_j, ConsVar_Grad_i,
+                                                 ConsVar_Grad_j, correct, &Temp_i, &Temp_j,
+                                                 NormalGrad, CorrectedGrad);
```
TobiKattmann, line 187: cut-paste-spezialize  in action :)

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -377,8 +377,8 @@ void CHeatSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
       V_i = solver_container[FLOW_SOL]->GetNodes()->GetPrimitive(iPoint);
       V_j = solver_container[FLOW_SOL]->GetNodes()->GetPrimitive(jPoint);
 
-      Temp_i_Grad = nodes->GetGradient(iPoint);
-      Temp_j_Grad = nodes->GetGradient(jPoint);
+      const auto Temp_i_Grad = nodes->GetGradient(iPoint);
+      const auto Temp_j_Grad = nodes->GetGradient(jPoint);
```
TobiKattmann, line 381: now auto evaluates to the MatrixView type, right (that returns const values on its own)?

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -392,10 +392,10 @@ void CHeatSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_containe
...
         }
 
-        Gradient_i = solver_container[FLOW_SOL]->GetNodes()->GetGradient_Reconstruction(iPoint);
-        Gradient_j = solver_container[FLOW_SOL]->GetNodes()->GetGradient_Reconstruction(jPoint);
-        Temp_i_Grad = nodes->GetGradient_Reconstruction(iPoint);
-        Temp_j_Grad = nodes->GetGradient_Reconstruction(jPoint);
+        const auto Gradient_i = solver_container[FLOW_SOL]->GetNodes()->GetGradient_Reconstruction(iPoint);
+        const auto Gradient_j = solver_container[FLOW_SOL]->GetNodes()->GetGradient_Reconstruction(jPoint);
+        const auto Temp_i_Grad = nodes->GetGradient_Reconstruction(iPoint);
+        const auto Temp_j_Grad = nodes->GetGradient_Reconstruction(jPoint);
```
TobiKattmann, line 398: Temp_i_Grad already defined in this scope (on the "block above") a few lines above. maybe `Temp_i_Grad_Recon` ? Isnt this somewhat dangerous because if someone deletes these lines (for some reason) there will still be a a valid Temp_i_Grad object floating around that does not return what one actually wants? ... one of  the worst kinds of error

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_2d/of_grad_findiff.csv.ref
```diff
@@ -1,2 +1,2 @@
...
+0               , 0.0             , -99999.96982514858, 3.330700000025998e-08, 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , -0.04999999997368221, -1.1100000001884e-08, -2.069999999187999       , 0.0                     , 2.12000000054946    , 3.7100000016554446            , 330.00000030369847        , -39.999997625272954          , 315.0000011942211        , -40.000008993956726         , -1.400000004814217   , -149.9999996212864, 0.0                , -469.9999976764957, 1e-08
```
TobiKattmann, line 2: 🔎 what's going on here? (breakfast 🐞  obviously, but jokes aside) was this a bug? I guess it is 1e-8 land ...

TestCases/serial_regression.py
```diff
@@ -345,7 +345,7 @@ def main():
     turb_naca0012_sst.cfg_dir   = "rans/naca0012"
     turb_naca0012_sst.cfg_file  = "turb_NACA0012_sst.cfg"
     turb_naca0012_sst.test_iter = 10
-    turb_naca0012_sst.test_vals = [-11.451011, -12.798258, -5.863895, 1.049989, 0.019163, -1.925028]
+    turb_naca0012_sst.test_vals = [-11.451010, -12.798258, -5.863895, 1.049989, 0.019163, -1.925018]
```
TobiKattmann, line 348: I have a bit of trouble in finding an explanation from the changes why these few have marginal differences ... because I would then expect more to fail ... but maybe you can shed some light on this if you have an idea

TobiKattmann (approved):
Thanks pedro for this 💐  which is a bit more of just a charge against pointer-to-pointer. This ComputeProjectedGradient func and the simplification of the viscous terms numerics for the heat solver. And the const-ing and cleanup. 
I have a few things below, none of which are dealbreakers (mostly questions I guess)

SU2_CFD/include/numerics/CNumerics.hpp
pcarruscag: I should add an assert though, for 2 or 3.
Yes compiler explorer

Common/include/containers/container_decorators.hpp
```diff
@@ -30,6 +30,43 @@
...
+  explicit CMatrixView(const su2matrix<Scalar>& mat) : m_ptr(mat.data()), m_cols(mat.cols()) {}
+
+  const Scalar* operator[] (Index i) const noexcept { return &m_ptr[i*m_cols]; }
+  const Scalar& operator() (Index i, Index j) const noexcept { return m_ptr[i*m_cols + j]; }
+
+  template<class U = T, su2enable_if<!std::is_const<U>::value> = 0>
+  Scalar* operator[] (Index i) noexcept { return &m_ptr[i*m_cols]; }
+
+  template<class U = T, su2enable_if<!std::is_const<U>::value> = 0>
+  Scalar& operator() (Index i, Index j) noexcept { return m_ptr[i*m_cols + j]; }
```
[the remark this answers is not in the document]
pcarruscag: Hmmm no..
If the matrix view object is used in a const context then it returns const refs to su2double. Otherwise it will "try" to return non const ref.
However, if the view was returned by a const container, it cannot do that. Hence the use of sfinae to disable the non const version.

Common/include/containers/container_decorators.hpp
```diff
@@ -30,6 +30,43 @@
...
+
+  friend CMatrixView operator+ (CMatrixView mv, Index incr) { return CMatrixView(mv[incr], mv.m_cols); }
```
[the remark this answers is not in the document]
pcarruscag: Well the type is not really important in this case, just the interface. And there is too much of this +stuff, so less work for me xD.

Common/include/containers/container_decorators.hpp
```diff
@@ -78,6 +118,18 @@ class C3DContainerDecorator {
   Scalar& operator() (Index i, Index j, Index k) noexcept { return m_storage(i, j*m_innerSz + k); }
   const Scalar& operator() (Index i, Index j, Index k) const noexcept { return m_storage(i, j*m_innerSz + k); }
 
+  /*!
+   * \brief Matrix access.
+   */
+  Matrix operator[] (Index i) noexcept { return Matrix(m_storage[i], m_innerSz); }
+  ConstMatrix operator[] (Index i) const noexcept { return ConstMatrix(m_storage[i], m_innerSz); }
```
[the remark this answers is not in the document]
pcarruscag: I think I was influenced by Eigen (inception style)

SU2_CFD/src/numerics/heat.cpp
pcarruscag: I don't think compilers look for const to optimize, because you have things like const_cast as part of legal c++.
I was lazy... You can go ahead and constify if you want but the method is small and safe.

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: Correct. Before it was a const pointer that could return non const

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: I may have missed it. I'll circle back to this.

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_2d/of_grad_findiff.csv.ref
pcarruscag: My different handling of division by zero maybe, or the optimizability of 2 or 3

SU2_CFD/include/numerics/CNumerics.hpp
pcarruscag, line 654:
```suggestion
    assert(nDim == 2 || nDim == 3);
    nDim = (nDim > 2)? 3 : 2;
```

SU2_CFD/src/numerics/heat.cpp
TobiKattmann: Ok you already went ahead yourself 👍

SU2_CFD/include/numerics/CNumerics.hpp
TobiKattmann: Of course this is a valid check, but can you motivate that a bit why you think it is additionally necessary here? I guess if we would do this kind of sanity checks of sizes in more places, we would have quite a bit of them. At first glance that looks like a bit overkill but maybe i just dont understand it.

And might this already be enough for the compiler to not unroll further because it puts a hard constraint on the nDim value

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_2d/of_grad_findiff.csv.ref
TobiKattmann: Can you give me place to look for the different handling of division by zero 🔬 (so I dont have to think for myself)

SU2_CFD/include/numerics/CNumerics.hpp
pcarruscag: Because the code is only right if called with nDim = 2 or 3, and it is good practice to assert that kind of assumption to facilitate debugging, keep in mind that asserts are no-ops in release builds!
This is also why they cannot be used to inform the compiler that we expect 2 or 3 -> https://gcc.godbolt.org/z/nbPv8KhPo

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -659,6 +627,59 @@ class CNumerics {
...
+    assert(nDim == 2 || nDim == 3);
+    nDim = (nDim > 2)? 3 : 2;
+    su2double edgeVec[MAXNDIM], dist_ij_2 = 0.0, proj_vector_ij = 0.0;
+
+    for (int iDim = 0; iDim < nDim; iDim++) {
+      edgeVec[iDim] = coord_j[iDim] - coord_i[iDim];
+      dist_ij_2 += pow(edgeVec[iDim], 2);
+      proj_vector_ij += edgeVec[iDim] * normal[iDim];
+    }
+    proj_vector_ij /= max(dist_ij_2,EPS);
```
pcarruscag, line 663: div by 0

SU2_CFD/src/solvers/CHeatSolver.cpp
pcarruscag: Well, before the scope of that variable was even larger... This type of reconstruction logic should be shared by scalar solvers in the future, so I won't lose too much time with this.

PR #1326, opened 2021-07-12
Fix for axisymmetric terms in NEMO + general NEMO updates

WallyMaier:
## Proposed Changes
-This work is focused on fixing some of the axisymmetric loop structures within NEMO, which prevented residual evolution.   This also changes a viscous planar wedge regression to an axisymmetric case.  

-Code updates to make NEMO more-in-line with SU2 code style.  i.e. variable declarations, SU2 MPI errors, etc.

-Fixes bug in vibrational-electronic functionality for CSU2TCLib

-Updating ionization/nheavy loops so CSU2TCLib and CMutationTCLib are consistent.   

## Related Work
This addresses/builds on #1106 #1205 

## PR Checklist


SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -1670,23 +1665,23 @@ void CNEMOEulerSolver::BC_Far_Field(CGeometry *geometry, CSolver **solver_contai
 
         /*--- Species diffusion coefficients ---*/
         visc_numerics->SetDiffusionCoeff(nodes->GetDiffusionCoeff(iPoint),
-                                         node_infty->GetDiffusionCoeff(0) );
+                                         nodes->GetDiffusionCoeff(iPoint));
```
WallyMaier, line 1668: This is more consistent with the baseline SU2.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -815,45 +815,71 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
...
   /*--- Initialize the error counter ---*/
-  eAxi_local = 0;
-  eChm_local = 0;
-  eVib_local = 0;
+  unsigned long eAxi_local = 0;
+  unsigned long eChm_local = 0;
+  unsigned long eVib_local = 0;
 
   /*--- Initialize the source residual to zero ---*/
   for (iVar = 0; iVar < nVar; iVar++) Residual[iVar] = 0.0;
```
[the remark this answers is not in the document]
pcarruscag: That is unlikely, the NEMO solver does not declare itself to be OpenMP-ready (via method GetHasHybridParallel() -> true) and so it will not be called multi-threaded by the Integration classes.
Also if you do not compile with -Dwith-omp=enabled (or true, I can never recall) all those SU2_OMP_? macros are no-ops.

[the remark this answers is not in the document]
WallyMaier: So this was a super complex problem to solve....
I had the gradient computation inside the for loop.  So as you can imagine, things went south quite quickly.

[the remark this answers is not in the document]
pcarruscag: Actually you did point me to an hybrid parallel bug...

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -815,45 +815,70 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
...
 
+  /*--- Preprocess viscous axisymm variables (if necessary) ---*/
+  if (axisymm && viscous) {
+
+    for (iPoint = 0; iPoint < nPoint; iPoint++) {
```
pcarruscag, line 841:
```suggestion
    SU2_OMP_FOR_STAT(omp_chunk_size)
    for (iPoint = 0; iPoint < nPoint; iPoint++) {
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -815,45 +815,70 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
...
+      su2double xVelocity       = nodes->GetVelocity(iPoint,0);
+      su2double Total_Viscosity = nodes->GetLaminarViscosity(iPoint) + nodes->GetEddyViscosity(iPoint);
+
+      if (yCoord > EPS){
+        su2double nu_v_on_y = Total_Viscosity*yVelocity/yCoord;
+        nodes->SetAuxVar(iPoint, 0, nu_v_on_y);
+        nodes->SetAuxVar(iPoint, 1, nu_v_on_y*yVelocity);
+        nodes->SetAuxVar(iPoint, 2, nu_v_on_y*xVelocity);
+      }
+   }
```
pcarruscag, line 853:
```suggestion
   }
   END_SU2_OMP_FOR
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -815,45 +815,70 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
...
+
+   /*--- Compute the auxiliary variable gradient with GG or WLS. ---*/
+   if (config->GetKind_Gradient_Method() == GREEN_GAUSS) {
+     SetAuxVar_Gradient_GG(geometry, config);
+   }
+   if (config->GetKind_Gradient_Method() == WEIGHTED_LEAST_SQUARES) {
+     SetAuxVar_Gradient_LS(geometry, config);
+   }
```
pcarruscag: Move this to CFVMFlowSolverBase.inl and share it between CEulerSolver and CNEMOEulerSolver please (the bug is in CEuler solver, not here)

SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
```diff
@@ -353,31 +353,31 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
...
-    su2double qy_ve                = kve*GV[TVE_INDEX][1];
+    su2double qy_t                 = -total_conductivity_i*GV[T_INDEX][1];
+    su2double qy_ve                = -kve*GV[TVE_INDEX][1];
 
     /*--- Enthalpy and vib-el energy transport due to y-direction diffusion---*/
     for (iSpecies = 0; iSpecies < nHeavy; iSpecies++) {
-      sumJhs_y  += (rho*Ds[iSpecies]*GV[RHOS_INDEX+iSpecies][1] - V_i[RHOS_INDEX+iSpecies]*Vector[1]) * hs[iSpecies];
-      sumJeve_y += (rho*Ds[iSpecies]*GV[RHOS_INDEX+iSpecies][1] - V_i[RHOS_INDEX+iSpecies]*Vector[1]) * eve_i[iSpecies];
+      sumJhs_y  += -(rho*Ds[iSpecies]*GV[RHOS_INDEX+iSpecies][1] - V_i[RHOS_INDEX+iSpecies]*Vector[1]) * hs[iSpecies];
+      sumJeve_y += -(rho*Ds[iSpecies]*GV[RHOS_INDEX+iSpecies][1] - V_i[RHOS_INDEX+iSpecies]*Vector[1]) * eve_i[iSpecies];
```
pcarruscag, line 367: Have a look at #1312, I think there was a logic error here with how "Vector" is used which I tried to fix.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
pcarruscag: Yes, the logic is more or less: if the method does not have to be a template then it is declared in the .hpp and implemented in .inl of CFVMSolverBase.
If it has to be a template, or you want it to be "inlineable", then it has to go in the .hpp.
But both files are headers, the only thing that changes is the "visibility", e.g. CEulerSolver includes the hpp but not the inl.
At the moment the inl is included by the NS solvers, at the top of the associated .cpp's you'll see
```c++
/*--- Explicit instantiation of the parent class of CIncEulerSolver,
 *    to spread the compilation over two cpp files. ---*/
template class CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME::INCOMPRESSIBLE>;
```
That also means that if you change something in the .inl only the NS solvers recompile, but not the Euler solvers (which sounded like a good idea because e.g. CEulerSolver is a reasonably large file).

The incompressible version looks different enough to be either its own method or left as is.
My rule of thumb for such master versions is not to have physics-based conditionals in there (e.g. "if compressible"), if necessary, that type of detail should be provided from the outside, and not already present in the body of the "master" version.
In other words, we want those versions to be generic in the sense they allow externally provided specialization, not in the sense they already try to cover all possible options.
I hope this makes some sense.

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -154,10 +154,10 @@ void CSourceAxisymmetric_Flow::ResidualDiffusion(){
   residual[2] -= Volume*(yinv*total_viscosity_i*2*(PrimVar_Grad_i[2][1]-v*yinv)
                          -TWO3*AuxVar_Grad_i[0][1]);
   residual[3] -= Volume*(yinv*(total_viscosity_i*(u*(PrimVar_Grad_i[2][0]+PrimVar_Grad_i[1][1])
-                                                  +v*TWO3*(2*PrimVar_Grad_i[1][1]-PrimVar_Grad_i[1][0]
-                                                           -v*yinv+U_i[0]*turb_ke_i))
-                               -total_conductivity_i*PrimVar_Grad_i[0][1])
-                         -TWO3*(AuxVar_Grad_i[1][1]+AuxVar_Grad_i[2][0]));
+                                                 +v*TWO3*(2*PrimVar_Grad_i[1][1]-PrimVar_Grad_i[1][0]
+                                                 -v*yinv+U_i[0]*turb_ke_i))
+                                                 +total_conductivity_i*PrimVar_Grad_i[0][1])
```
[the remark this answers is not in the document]
FlorianDm: Indeed I believe it should actually be +k dT/dy then. Thanks for spotting this. The test case i made breaks so I will rerun it with this branch and add the change to the PR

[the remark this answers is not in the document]
FlorianDm: Ok so I reran the test case with this branch. I guess I should wait until this PR is to be merged to push the changes in the regression files to the TestCases repo so any other open PRs don't get failed regressions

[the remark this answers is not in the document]
FlorianDm: or is the rest ready to go? If it may take a while maybe I better do the fix now in another PR in concert with the TestCases changes so it can be merged straight away as a simple fix and then you just merge the latest develop in here.. let me know what you think is best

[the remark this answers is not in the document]
WallyMaier: @FlorianDm Glad you could confirm!  This PR should be done on Monday/Tuesday (hopefully) and Ill update the regression tests.  However, if you want to go ahead to address it case before then, thats fine as well!

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -767,15 +767,15 @@ vector<su2double>& CSU2TCLib::ComputeSpeciesEve(su2double val_T){
         Ev = 0.0;
       /*--- Calculate electronic energy ---*/
       num = 0.0;
-      denom = ElDegeneracy[iSpecies][0] * exp(-CharElTemp[iSpecies][0]/val_T);
+      denom = ElDegeneracy[iSpecies][0] * exp(CharElTemp[iSpecies][0]/val_T);
```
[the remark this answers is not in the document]
WallyMaier: Note to myself: This should be a negative, I negative should change in ComputeMixtureEnergies.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: Done!

WallyMaier:
Everything is up-to-date. I just need to update the TestCase folder....not sure what that order is...

TestCases/hybrid_regression.py
```diff
@@ -227,7 +227,7 @@ def main():
     axi_rans_air_nozzle.cfg_dir   = "axisymmetric_rans/air_nozzle"
     axi_rans_air_nozzle.cfg_file  = "air_nozzle.cfg"
     axi_rans_air_nozzle.test_iter = 10
-    axi_rans_air_nozzle.test_vals = [-12.094937, -6.622043, -8.814412, -2.393288]
+    axi_rans_air_nozzle.test_vals = [-6.348077, -0.827162, -2.241982, 2.313210]
```
FlorianDm, line 230: Just to clarify, the way I had done the regressions was with RESTART=YES in the config file where the restart file in TestCases solution_flow.dat was already converged to rho<1e-12 so it would just do 10 iterations and then those values I took. I see you did it differently since the values are only at 1e-6. I guess it does not matter but just wanted to clarify

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -277,6 +277,11 @@ class CFVMFlowSolverBase : public CSolver {
    */
   void SumEdgeFluxes(const CGeometry* geometry);
 
+  /*!
+   * \brief Computes and sets the required auxilliary vars (and gradients) for axisymmetric flow.
+   */
+  void PrepareAndConputeAuxVarGrad(CGeometry *geometry, const CConfig* config);
```
pcarruscag, line 283:
```suggestion
  void ComputeAxisymmetricAuxGradients(CGeometry *geometry, const CConfig* config);
```
I'm not aware of the proper naming but let's go for something more descriptive.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2923,3 +2923,32 @@ su2double CFVMFlowSolverBase<V,R>::EvaluateCommonObjFunc(const CConfig& config)
 
   return objFun;
 }
+
+template <class V, ENUM_REGIME FlowRegime>
+void CFVMFlowSolverBase<V, FlowRegime>::PrepareAndConputeAuxVarGrad(CGeometry *geometry, const CConfig* config) {
```
pcarruscag, line 2928:
```suggestion
void CFVMFlowSolverBase<V, FlowRegime>::ComputeAxisymmetricAuxGradients(CGeometry *geometry, const CConfig* config) {
```
Nicely done! Thanks.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -2339,30 +2339,7 @@ void CEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_contain
...
-
-      /*--- Compute the auxiliary variable gradient with GG or WLS. ---*/
-      if (config->GetKind_Gradient_Method() == GREEN_GAUSS) {
-        SetAuxVar_Gradient_GG(geometry, config);
-      }
-      if (config->GetKind_Gradient_Method() == WEIGHTED_LEAST_SQUARES) {
-        SetAuxVar_Gradient_LS(geometry, config);
-      }
-
+      PrepareAndConputeAuxVarGrad(geometry, config);
```
pcarruscag, line 2342:
```suggestion
      ComputeAxisymmetricAuxGradients(geometry, config);
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -815,45 +815,49 @@ void CNEMOEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_con
...
+  unsigned long eAxi_local = 0;
+  unsigned long eChm_local = 0;
+  unsigned long eVib_local = 0;
 
   /*--- Initialize the source residual to zero ---*/
   for (iVar = 0; iVar < nVar; iVar++) Residual[iVar] = 0.0;
 
+  /*--- Preprocess viscous axisymm variables (if necessary) ---*/
+  if (axisymm && viscous) {
+    PrepareAndConputeAuxVarGrad(geometry,config);
```
pcarruscag, line 840:
```suggestion
    ComputeAxisymmetricAuxGradients(geometry,config);
```

pcarruscag (approved):

pcarruscag (commented):
The order is to merge into testcases first just before merging this branch (then you can manually trigger the tests to re-run).

TestCases/hybrid_regression.py
WallyMaier: @FlorianDm, thanks for the clarification.  I only updated the values in the regression test, so it still should be a restart.

PR #1330, opened 2021-07-21
Base class for transported scalars

TobiKattmann:
## Proposed Changes
Hi all,

The general idea for this PR can be cited from @pcarruscag in #1226:
```
The ideia would be to create a CFVMSolverBase (in the style of the one for flow solvers) to cover:

    Turbulence solvers;
    Heat solver;
    Radiation solver;
    General passive/active scalar solvers;
    Eventual transition solvers;

The starting point would be the current base turbulence solver.
```
 
Right now the only things I did were (i.e. don't fear the large diff)
- create  CScalarSolver/Variable files/classes and scalar_convection/diffusion/sources files and fill them with the contents of the CTurb* or turb_* files.
- Make CScalarSolver/Variable the parent class of CTurbSolver/Variable
- Delete pretty much everything from CTurb* or turb_* files except for some constructors

So far it compiled and some local regression testing showed no problems, lets see what the CI pipeline has to say about this

Now with this additional class introduced the actual works begins:
- figure out what is a reasonable interface in CScalar* to unify transported species and turb-models (and the other solvers probably)
- code this common interface and push all turb-specific things back into the respective files (that is why I didn't `git mv` )

In order to keep this somewhat limited I would focus on porting the Turbsolvers to the new structure (however that might look, but looking at #1223 is what I'll do first) and maybe a simple transported passive scalar or so. We want the turb solver to be integrated under that Scalar-class which is arguably what most/all people use, that's why I want to have in this work at the beginning. We also have quite a few RANS Testcases so it is directly visible if something in the actual computation is changed.

In case there are comments, ideas foreseeable stepping stones from experience please let me know here.

## Related Work
#777 by @economon which is the initial push towards this separate scalar solver class for additionally transported scalars (not merged)
#1223 by @danielmayer @bigfooted  which builds up on #777 and extends it to its needs (but does not support the turb solvers e.g.)
#1226 by @oleburghardt - In this PR is a discussion on creating the structure which this (and following) PR('s) try to achieve
#1332 by @mheimgartner - implementing multi component flow (or analytic mixing properties of said multicomp flow). That PR is built upon #1223 but will merge this PR (once it is done) and throw out all additional changes in order to be merged into develop without relying on feature_flamelet

## PR Checklist


SU2_CFD/src/numerics/scalar/scalar_sources.cpp
```diff
@@ -0,0 +1,930 @@
+/*!
+ * \file turb_sources.cpp
+ * \brief Implementation of numerics classes for integration of
+ *        turbulence source-terms.
```
pcarruscag, line 4: Again :+1: for taking the plunge on this.
I would not have moved the sources, I think these will always be very specific to each derived solver.
It would be better for navigation to keep it in the turbulence folder.
TBH I would not have moved anything that was not immediately necessary (all numerics in this case) because if someone comes along with changes to those classes....

TobiKattmann: My motivation to move everything was more of a test/first-step to see if I break anything with that -> inserting the CScalar layer and moving things there. Turbulent sources will certainly move back and everything else specific will move back at some point.

> code this common interface and push all turb-specific things back into the respective files (that is why I didn't git mv )

I still have to figure out how a suitable generalized interface looks like and based on that I will reject the temporary residence permits for some stuff (e.g. all turb sources) 👍

TobiKattmann: But by all means, I highly appreciate all sorts of comments as I there is lots of things to do wrong so I am glad for additional brains to look at this 🧠 👀

pcarruscag: Sure but you only need to worry about the solver and variables. The numerics are fairly independent (for now).
Minimize the number of changes per PR is all I'm saying. I think that's a guideline we all endorse :)

TobiKattmann:
Still WIP but with that in the title the regression status behind each commit is always "unfinished" because WIP-check-bot does that. Now that we have Draft-PR I guess we dont WIP anymore tbh

pcarruscag:
It helps me with the manual edit of the release notes, to know what was not merged yet.

SU2_CFD/src/numerics/scalar/scalar_sources.cpp
TobiKattmann: > TBH I would not have moved anything that was not immediately necessary (all numerics in this case) because if someone comes along with changes to those classes....

well, yes that happenend ... should have listened directly

TobiKattmann:
Things done:
- Create CScalarSolver/Variable classes that now sits between CSolver/Variable and CTurbSolver/Variable
- Move variables/methods to CScalar-land for which one can reasonably assume that they will be useful for future children of the CScalar-family
- That implies, that the rest remains in Turbulence land mostly(!) untouched
- Mostly, because I renamed (Set)TurbVar[-,Gradient] variables/methods in CNumerics which result in changing that all over turb_diff/conv/sources. They hear by the name (Set)ScalarVar* now
And that's it for now. I added some personal comments in the code here and there that will be removed.

Next up (coming to a PR near you soon):
- add CSpeciesSolver/Variable with appropriate Upwind/Viscous_Residual and Boundary conditions
- The first step will be a truly Passive Species transport but then ...
- ...introduce the mixture properties of @mheimgartner from #1332 i.e. make the transport equation 'active'. This comes along with some more necessary changes to CConfig. I am not sure yet what is the best order of tackling that. We'll see...
- @mheimgartner work concentrates on 1 additional transport equation. The idea is of course to make it as general as possible but it might be wiser to just go with 1 for the beginning.
- For that setup (2 components mixing) we'll want to get validated gradients going
- The more general way will be needed to merge with @danielmayer #1223 but that would be the step after the above mentioned

@bigfooted that i what I currently have in mind but feel free to chime in here.

SU2_CFD/include/variables/CTurbVariable.hpp
```diff
@@ -27,22 +27,19 @@
...
 /*!
  * \class CTurbVariable
  * \brief Base class for defining the variables of the turbulence model.
  * \ingroup Turbulence_Model
  * \author A. Bueno.
  */
-class CTurbVariable : public CVariable {
+class CTurbVariable : public CScalarVariable {
 protected:
   VectorType muT;         /*!< \brief Eddy viscosity. */
```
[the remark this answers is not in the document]
pcarruscag: The solvers see the nodes as CScalarVariable, but the methods to access muT are in CTurbVariable.
The compiler does not know that the type of nodes will be CTurbVariable at runtime and so it cannot de-virtualize the access to variables that are not in CScalarVariable.

By making CScalarSolver a class template you can instantiate it with different types for "nodes", i.e. in practice you end up with multiple CScalarSolver "tuned" for each of the derived classes. CScalarSolver<CTurbVariable>, CScalarSolver<CHeatVariable> (which are not the same solver anymore, but that is fine because we want this as a building block.

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,958 @@
...
+  EdgeColoring[0] = DummyGridColor<>(geometry->GetnEdge());
+#endif
+}
+
+CScalarSolver::~CScalarSolver() {
+
+  delete nodes;
+
+}
+
```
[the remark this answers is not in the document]
pcarruscag: Indeed that is annoying, I think I've put su2double in front of all flow solver vars because of that. Otherwise the problem comes back, or we have to disable clang-format around certain sections.

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,958 @@
...
+  /*--- Menter's first blending function (only SST)---*/
+  if ((config->GetKind_Turb_Model() == SST) || (config->GetKind_Turb_Model() == SST_SUST))
+    numerics->SetF1blending(nodes->GetF1blending(iPoint),
+                            nodes->GetF1blending(jPoint));
+
+  /*--- Roughness heights. ---*/
+  if (config->GetKind_Turb_Model() == SA)
+    numerics->SetRoughness(geometry->nodes->GetRoughnessHeight(iPoint),
+                           geometry->nodes->GetRoughnessHeight(jPoint));
```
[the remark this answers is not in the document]
pcarruscag: Something along the lines of what we are doing to compute time-steps in the flow solver, to keep it nice and speedy.

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,958 @@
...
+        for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+
+          su2double density = flowNodes->GetDensity(iPoint);
+          su2double density_old = density;
+
+          if (compressible)
+            density_old = flowNodes->GetSolution_Old(iPoint,0);
+
+          for (unsigned short iVar = 0; iVar < nVar; iVar++) {
+            nodes->AddConservativeSolution(iPoint, iVar,
```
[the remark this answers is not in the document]
pcarruscag: Ah... yet another variant, well we can always override those very special cases. It is probably better to have one or two exceptions than to shoe-horn every solver into an interface that will be inefficient in 90% of cases (every extra variable is a little more memory for AD for example).

SU2_CFD/src/solvers/CScalarSolver.cpp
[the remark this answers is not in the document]
pcarruscag: > The 1 space indentation of protected seems weird as well

It's the price to pay for peace of mind with formatting, it's easier to get used to that than to every personal style in SU2 :laughing:

SU2_CFD/src/solvers/CScalarSolver.cpp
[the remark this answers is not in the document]
TobiKattmann: I added a `bool Conservative` to Cscalarsolver as class data and used the more general `AddConservativeSolution` for both, the clipping limits are default set to su2doule::min/max in case a solver does not use this (e.g. SA)

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -48,11 +48,13 @@ class CScalarSolver : public CSolver {
 
   unsigned long omp_chunk_size; /*!< \brief Chunk size used in light point loops. */
 
-  su2double lowerlimit[MAXNVAR] = {0.0}, /*!< \brief contains lower limits for turbulence variables. */
-      upperlimit[MAXNVAR] = {0.0};       /*!< \brief contains upper limits for turbulence variables. */
+  su2double lowerlimit[MAXNVAR] = {std::numeric_limits<su2double>::lowest()}; /*!< \brief contains lower limits for turbulence variables. Note that ::min() returns the smallest positive value for floats. */
+  su2double upperlimit[MAXNVAR] = {std::numeric_limits<su2double>::max()}; /*!< \brief contains upper limits for turbulence variables. */
```
pcarruscag, line 52: This does not do what you think it does :) after the first you have zeros.

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -48,7 +48,7 @@ class CScalarSolver : public CSolver {
 
   unsigned long omp_chunk_size; /*!< \brief Chunk size used in light point loops. */
 
-  su2double lowerlimit[MAXNVAR] = {std::numeric_limits<su2double>::min()}; /*!< \brief contains lower limits for turbulence variables. */
+  su2double lowerlimit[MAXNVAR] = {std::numeric_limits<su2double>::lowest()}; /*!< \brief contains lower limits for turbulence variables. Note that ::min() returns the smallest positive value for floats. */
```
TobiKattmann, line 51: I wasn't aware of that. Maybe that is a noteworthy info.

SU2_CFD/include/solvers/CScalarSolver.hpp
TobiKattmann: hm, yes. it didn't break on me because SA only has 1 var (but some solver in the future might have more and does not explicitly set those). I'll fix it

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -450,12 +448,17 @@ void CScalarSolver::SetResidual_DualTime(CGeometry* geometry, CSolver** solver_c
...
+  su2double Density_nM1 = 1.0, Density_n = 1.0, Density_nP1 = 1.0;
   su2double Volume_nM1, Volume_nP1;
-  su2double Density_nM1, Density_n, Density_nP1;
   const su2double *Normal = nullptr, *GridVel_i = nullptr, *GridVel_j = nullptr;
   su2double Residual_GCL;
 
+  const int nvar = nVar;
+  auto FactorTimesVector = [nvar](su2double factor, su2double* vec) {
+    for (int iVar = 0; iVar < nvar; iVar++) vec[iVar] = factor * vec[iVar];
+  };
```
TobiKattmann, line 460: Maybe this is already implemented somewhere centrally? (Note that I wanted to avoid the term Scalar ...)

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -372,40 +373,26 @@ void CScalarSolver::CompleteImplicitIteration(CGeometry* geometry, CSolver** sol
...
+    /*--- Update the turbulent solution. Only SST variants are clipped. ---*/
 
-          if (compressible) density_old = flowNodes->GetSolution_Old(iPoint, 0);
+    SU2_OMP_FOR_STAT(omp_chunk_size)
+    for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++) {
+      /*--- Multiply the Solution var with density to get the conservative transported quantity, if necessary. ---*/
+      if (Conservative) {
+        density = flowNodes->GetDensity(iPoint);
+        density_old = compressible ? flowNodes->GetSolution_Old(iPoint, 0) : density;
+      }
```
pcarruscag, line 387: That will work I guess, but it throws away optimizations.
Better use if (conservative) (outside of the loop over variables) to call either AddConservativeSolution or AddClippedSolution.

SU2_CFD/src/solvers/CScalarSolver.cpp
TobiKattmann: nvm, removed that

SU2_CFD/src/solvers/CScalarSolver.cpp
TobiKattmann: Ok I removed that but I just kept AddConservativeSolution renamed to AddClippedSolution and made the factor between Solution and Transported variable default to 1. Thus the difference between the if(Conservative) ...} else {..} is really just the density. See 817186f826785e43d5aa8ae397846b39698dabbd . Let me know what you think

SU2_CFD/include/solvers/CScalarSolver.hpp
TobiKattmann: Now it is done via a loop in the CScalarVar ctor, so each derived solver can overwrite that

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,811 @@
...
+                (Density_nP1 * U_time_nP1[iVar] - Density_n * U_time_n[iVar]) * (Volume_nP1 / TimeStep);
+          if (second_order)
+            LinSysRes(iPoint, iVar) +=
+                (Density_nP1 * U_time_nP1[iVar] - Density_n * U_time_n[iVar]) * (3.0 * Volume_nP1 / (2.0 * TimeStep)) +
+                (Density_nM1 * U_time_nM1[iVar] - Density_n * U_time_n[iVar]) * (Volume_nM1 / (2.0 * TimeStep));
+        }
+
+      } else {
+        for (iVar = 0; iVar < nVar; iVar++) {
+          if (first_order) LinSysRes(iPoint, iVar) += (U_time_nP1[iVar] - U_time_n[iVar]) * (Volume_nP1 / TimeStep);
```
pcarruscag, line 671: and here

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,811 @@
...
+    } else {
+      Read_SU2_Restart_ASCII(geometry[MESH_0], config, restart_filename);
+    }
+
+    /*--- Skip flow variables ---*/
+
+    unsigned short skipVars = 0;
+
+    if (nDim == 2) skipVars += 6;
+    if (nDim == 3) skipVars += 8;
```
pcarruscag, line 717:
```suggestion
    unsigned short skipVars = nDim + solver[MESH_0][FLOW_SOL]->GetnVar();
```

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1591,3 +1588,54 @@ void CTurbSASolver::SetUniformInlet(const CConfig* config, unsigned short iMarke
...
+void CTurbSASolver::ComputeUnderRelaxationFactor(const CConfig *config) {
+
+  /* Apply the turbulent under-relaxation to the SA variants. The
+   SA_NEG model is more robust due to allowing for negative nu_tilde,
+   so the under-relaxation is not applied to that variant. */
+
+  bool sa_model = ((config->GetKind_Turb_Model() == SA)        ||
+                   (config->GetKind_Turb_Model() == SA_E)      ||
+                   (config->GetKind_Turb_Model() == SA_COMP)   ||
+                   (config->GetKind_Turb_Model() == SA_E_COMP));
```
pcarruscag, line 1601:
```suggestion
  if (config->GetKind_Turb_Model() == SA_NEG) return;
```

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1591,3 +1588,54 @@ void CTurbSASolver::SetUniformInlet(const CConfig* config, unsigned short iMarke
...
+          turbulence variables can change over a nonlinear iteration. */
+
+        const unsigned long index = iPoint * nVar + iVar;
+        su2double ratio = fabs(LinSysSol[index]) / (fabs(nodes->GetSolution(iPoint, iVar)) + EPS);
+        if (ratio > allowableRatio) {
+          localUnderRelaxation = min(allowableRatio / ratio, localUnderRelaxation);
+        }
+
+      }
+    }
```
pcarruscag, line 1626:
```suggestion
    su2double ratio = fabs(LinSysSol[iPoint]) / (fabs(nodes->GetSolution(iPoint, 0)) + EPS);
    /* We impose a limit on the maximum percentage that the
      turbulence variables can change over a nonlinear iteration. */
    if (ratio > allowableRatio) {
      localUnderRelaxation = min(allowableRatio / ratio, localUnderRelaxation);
    }
```

TobiKattmann: yep, sa has indeed only 1 variable 👍

SU2_CFD/src/solvers/CTurbSASolver.cpp
TobiKattmann: Although I am not fully sold on reversing these kinds of logics, who knows what is about to come

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+          if (Conservative) {
+            if (incompressible)
+              Density_n = flowNodes->GetDensity(iPoint);  // Temporary fix
+            else
+              Density_n = flowNodes->GetSolution_time_n(iPoint, 0);
+
+            for (iVar = 0; iVar < nVar; iVar++) LinSysRes(iPoint, iVar) += Density_n * U_time_n[iVar] * Residual_GCL;
+          } else {
+            for (iVar = 0; iVar < nVar; iVar++) LinSysRes(iPoint, iVar) += U_time_n[iVar] * Residual_GCL;
```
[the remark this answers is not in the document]
TobiKattmann: 👍 yes, I was a bit lazy for the dynamic grid portion

SU2_CFD/include/variables/CScalarVariable.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+/*!
+ * \class CScalarVariable
+ * \brief Base class for defining the variables of the turbulence model.
+ * \ingroup Turbulence_Model
+ * \author A. Bueno.
+ */
+class CScalarVariable : public CVariable {
+ protected:
+  MatrixType HB_Source; /*!< \brief Harmonic Balance source term. */
```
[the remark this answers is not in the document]
TobiKattmann: 👍 yep they were in the SA-Variable, so they weren't accessible for the SST solver as well

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+    if (config->GetRead_Binary_Restart()) {
+      Read_SU2_Restart_Binary(geometry[MESH_0], config, restart_filename);
+    } else {
+      Read_SU2_Restart_ASCII(geometry[MESH_0], config, restart_filename);
+    }
+
+    /*--- Skip flow variables ---*/
+
+    unsigned short skipVars = nDim + solver[MESH_0][FLOW_SOL]->GetnVar();
```
TobiKattmann, line 705: `CScalarSolver::LoadRestart` will work like this only in a 'choose one' fashion. For Turb and Species Transport There needs to be a clear hierarchy (which is given by the ouptut of course). Although not strictly necessary (I think), keeping the order of how the linear systems are solved might be good for OCD-patients. 
I personally would keep this method right here because the majority stays the same. idk have a `skipVars = DetermineSkpVars(config)` method

SU2_CFD/src/solvers/CScalarSolver.cpp
```diff
@@ -0,0 +1,799 @@
...
+  SU2_OMP_BARRIER
+
+  /*--- MPI solution and compute the eddy viscosity ---*/
+
+  solver[MESH_0][TURB_SOL]->InitiateComms(geometry[MESH_0], config, SOLUTION);
+  solver[MESH_0][TURB_SOL]->CompleteComms(geometry[MESH_0], config, SOLUTION);
+
+  solver[MESH_0][FLOW_SOL]->Preprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0, NO_RK_ITER,
+                                          RUNTIME_FLOW_SYS, false);
+  solver[MESH_0][TURB_SOL]->Postprocessing(geometry[MESH_0], solver[MESH_0], config, MESH_0);
```
TobiKattmann, line 759: The other things that is solver specific part is especially what to Pre/Postprocess after solution loading to stay consistent with the 'normal' iteration. For species transport  with turb this gets a tiny bit more involved and with 4 solvers and more cross-dependencies one might have to do quite a bunch of calls to get everything updated.
(and the SOLVER_IDX needs to be specialized)

SU2_CFD/include/variables/CTurbVariable.hpp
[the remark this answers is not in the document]
TobiKattmann: I did a little debugging on restarts over at feature_flamelet and feature_multicomp, so this will be a bit delayed 🐌 sorry for that

SU2_CFD/src/solvers/CScalarSolver.cpp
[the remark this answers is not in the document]
TobiKattmann: Implemented in https://github.com/su2code/SU2/pull/1330/commits/3e975d3a69f495dfc7507246c816012b1b520fde let me know whether this is what you had in mind 👍

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -289,6 +286,24 @@ void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_contain
...
+                                     CNumerics* numerics, CConfig* config) {
+
+  /*--- Define an object to set solver specific numerics contribution. ---*/
+  struct SolverSpecificNumerics {
+
+    FORCEINLINE void operator() (const CGeometry& geometry, CNumerics& numerics, CConfig& config, const CScalarVariable& nodes, unsigned long iPoint, unsigned long jPoint) const {
+      /*--- Roughness heights. ---*/
+      if (config.GetKind_Turb_Model() == SA)
+        numerics.SetRoughness(geometry.nodes->GetRoughnessHeight(iPoint), geometry.nodes->GetRoughnessHeight(jPoint));
+    }
```
TobiKattmann, line 299: I know this is probably redundant, but I kept it for know to see whether the regression test pass. (this commit shouldn't change anything logically)
Is there any reason why surface roughness and other SA-variants should not work together

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -0,0 +1,294 @@
...
+    numerics->SetPrimitive(flowNodes->GetPrimitive(iPoint), flowNodes->GetPrimitive(jPoint));
+
+    /*--- Turbulent variables w/o reconstruction, and its gradients ---*/
+
+    numerics->SetScalarVar(nodes->GetSolution(iPoint), nodes->GetSolution(jPoint));
+    numerics->SetScalarVarGradient(nodes->GetGradient(iPoint), nodes->GetGradient(jPoint));
+
+    /*--- Call Numerics contribution which are Solver-Specifc. Implemented in the caller: Viscous_Residual.  ---*/
+
+    SolverSpecificNumerics(*geometry, *numerics, *config, *nodes, iPoint, jPoint);
```
pcarruscag, line 123: I would have passed just iPoint / jPoint because the caller has everything already.

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -0,0 +1,294 @@
...
+   * \brief Compute the viscous flux for the turbulent equation at a particular edge.
+   * \param[in] iEdge - Edge for which we want to compute the flux
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \note Calls a generic implementation after defining a SolverSpecificNumerics object.
+   */
+  inline virtual void Viscous_Residual(unsigned long iEdge, CGeometry* geometry, CSolver** solver_container,
+                                       CNumerics* numerics, CConfig* config) {}
```
pcarruscag, line 150: You can give it default behavior by passing an empty solver-specific object.
If you see strong reasons against that, make this pure virtual so that no one forgets this needs to be implemented in derived classes.

SU2_CFD/src/solvers/CTurbSASolver.cpp
pcarruscag: It is not implemented for other types of SA source, unfortunately.
Is there an incompatibility @koodlyakshay ?
The work of @suargi might fix this.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -289,6 +286,24 @@ void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_contain
...
+
+  } SolverSpecificNumerics;
```
pcarruscag, line 301: e.g. if you reduce the signature to iPoint/jPoint, also note that you would need to template the "nodes" of the struct for when you template the scalar solver. If you let the caller "capture" them that is not necessary.
```suggestion
  auto SolverSpecificNumerics = [&](unsigned long iPoint, unsigned long jPoint) {
    if (config->GetKind_Turb_Model() == SA)
      numerics->SetRoughness(geometry->nodes->GetRoughnessHeight(iPoint),
                             geometry->nodes->GetRoughnessHeight(jPoint));
  };
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -280,6 +275,25 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
...
+void CTurbSSTSolver::Viscous_Residual(unsigned long iEdge, CGeometry* geometry, CSolver** solver_container,
+                                     CNumerics* numerics, CConfig* config) {
+
+  /*--- Define an object to set solver specific numerics contribution. ---*/
+  struct SolverSpecificNumerics {
+
+    FORCEINLINE void operator() (const CGeometry& geometry, CNumerics& numerics, CConfig& config, const CScalarVariable& nodes, unsigned long iPoint, unsigned long jPoint) const {
+      /*--- Menter's first blending function (only SST)---*/
+      if ((config.GetKind_Turb_Model() == SST) || (config.GetKind_Turb_Model() == SST_SUST))
+        numerics.SetF1blending(nodes.GetF1blending(iPoint), nodes.GetF1blending(jPoint));
```
pcarruscag, line 287:
```suggestion
      numerics.SetF1blending(nodes.GetF1blending(iPoint), nodes.GetF1blending(jPoint));
```

pcarruscag (commented):
Yep that's the idea, nice :+1: 
Just a few things that may simplify the process a bit.

SU2_CFD/src/solvers/CTurbSASolver.cpp
TobiKattmann: 👍 ok so a lambda function instead of struct. That is indeed nicer to just capture everything that is used inside by reference

SU2_CFD/src/solvers/CTurbSASolver.cpp
koodlyakshay: I know the roughness model works with the noft2 version (which we have in SU2 now) and the standard version of SA. I have not seen it used with other versions anywhere.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
TobiKattmann: also done in https://github.com/su2code/SU2/pull/1330/commits/3e86076ad4641710db97f5aa38cd23af11b419ae

SU2_CFD/include/solvers/CScalarSolver.hpp
TobiKattmann: I went with the empty solver specific implementation instead of pure virtual. I guess as of right now all solvers will need to their own stuff so one could pure virtual imo but maybe there will be a gneric Diffusivity at one point and then this might come in handy
Done in https://github.com/su2code/SU2/pull/1330/commits/1d8f520a9da2907b105b3f47f2fcd05e75d424c9

SU2_CFD/src/solvers/CTurbSASolver.cpp
TobiKattmann: Ok thanks, so the condition `if SA` can stay 👍

SU2_CFD/src/solvers/CTurbSASolver.cpp
pcarruscag: yep, if all you need is "operator ()", lambdas are simpler I think.

SU2_CFD/include/variables/CScalarVariable.hpp
```diff
@@ -0,0 +1,93 @@
...
+  inline CVectorOfMatrix& GetGradient_Reconstruction() final { return Gradient_Reconstruction; }
+  inline const CVectorOfMatrix& GetGradient_Reconstruction() const final { return Gradient_Reconstruction; }
+
+  /*!
+   * \brief Set the harmonic balance source term.
+   * \param[in] iPoint - Point index.
+   * \param[in] iVar - Index of the variable.
+   * \param[in] source - Value of the harmonic balance source term. for the index <i>iVar</i>.
+   */
+  inline void SetHarmonicBalance_Source(unsigned long iPoint, unsigned long iVar, su2double source) override { HB_Source(iPoint,iVar) = source; }
```
pcarruscag, line 84:
```suggestion
  inline void SetHarmonicBalance_Source(unsigned long iPoint, unsigned long iVar, su2double source) final { HB_Source(iPoint,iVar) = source; }
```

SU2_CFD/include/variables/CScalarVariable.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  /*!
+   * \brief Get the harmonic balance source term.
+   * \param[in] iPoint - Point index.
+   * \param[in] iVar - Index of the variable.
+   * \return Value of the harmonic balance source term for the index <i>val_var</i>.
+   */
+  inline su2double GetHarmonicBalance_Source(unsigned long iPoint, unsigned long iVar) const override { return HB_Source(iPoint,iVar); }
```
pcarruscag, line 92:
```suggestion
  inline su2double GetHarmonicBalance_Source(unsigned long iPoint, unsigned long iVar) const final { return HB_Source(iPoint,iVar); }
```

SU2_CFD/include/variables/CScalarVariable.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+#pragma once
+
+#include "CVariable.hpp"
+
+/*!
+ * \class CScalarVariable
+ * \brief Base class for defining the variables of the turbulence model.
+ * \ingroup Turbulence_Model
+ * \author A. Bueno.
```
pcarruscag, line 36:
```suggestion
 * \brief Base class for defining the shared variables of scalar solvers.
```

SU2_CFD/include/variables/CScalarVariable.hpp
```diff
@@ -0,0 +1,93 @@
+/*!
+ * \file CScalarVariable.hpp
+ * \brief Base class for defining the variables of the turbulence model.
```
pcarruscag, line 3:
```suggestion
 * \brief Base class for defining the shared variables of scalar solvers.
```

SU2_CFD/include/variables/CTurbVariable.hpp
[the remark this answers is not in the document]
TobiKattmann: Should I go for a similar approach as `CFVMFlowSolverBase.hpp/inl` with the explicit instantiation of the Solver i need in the derived solvers? i.e. `template class CScalarSolver<CTurbVariable>;` at the top of CTurbVariable.cpp?

alternatively I could include CScalarSolver.cpp (or rename to .inl) at the end of its hpp?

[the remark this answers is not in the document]
pcarruscag: Do it like CFVMFlowSolverBase.hpp and .inl please.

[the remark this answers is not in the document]
TobiKattmann: Implemented in ca16276d50b2dc99ea40c511aa6df70b141b986c Let me know if that is what you think should be done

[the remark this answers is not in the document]
TobiKattmann: Just for my own documentation: [template use in implementation file](https://stackoverflow.com/questions/18186878/invalid-use-of-template-name-without-an-argument-list) / [templates implemented in header files](https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file)

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -30,9 +30,9 @@
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
```
pcarruscag, line 31:
```suggestion
#include "../../include/solvers/CScalarSolver.inl"

```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -25,11 +25,15 @@
...
 
+#include "../../include/solvers/CScalarSolver.inl"
 #include "../../include/solvers/CTurbSSTSolver.hpp"
 #include "../../include/variables/CTurbSSTVariable.hpp"
 #include "../../../Common/include/parallelization/omp_structure.hpp"
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
 
+/*--- Explicit instantiation of the parent class of CTurbSolver,
+ *    to spread the compilation over multiple cpp files. ---*/
+template class CScalarSolver<CTurbVariable>;
```
pcarruscag, line 36:
```suggestion
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -25,11 +25,15 @@
  * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "../../include/solvers/CScalarSolver.inl"
```
pcarruscag, line 28:
```suggestion
```

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1591,3 +1606,43 @@ void CTurbSASolver::SetUniformInlet(const CConfig* config, unsigned short iMarke
...
+    if (localUnderRelaxation < 1e-10) localUnderRelaxation = 0.0;
+
+    /* Store the under-relaxation factor for this point. */
+
+    nodes->SetUnderRelaxation(iPoint, localUnderRelaxation);
+
+  }
+  END_SU2_OMP_FOR
+
+}
```
TobiKattmann, line 1648:
```suggestion
}

```

SU2_CFD/src/solvers/CTransLMSolver.cpp
```diff
@@ -25,18 +25,22 @@
...
+/*--- Explicit instantiation of the parent class of CTurbSolver,
+ *    to spread the compilation over multiple cpp files. ---*/
+template class CScalarSolver<CTurbVariable>;
+
 /*---  This is the implementation of the Langtry-Menter transition model.
        The main reference for this model is:Langtry, Menter, AIAA J. 47(12) 2009 
        DOI: https://doi.org/10.2514/1.42362 ---*/
        
-CTransLMSolver::CTransLMSolver(void) : CTurbSolver() {}
+CTransLMSolver::CTransLMSolver(void) : CTurbSolver(true) {} ///Note: TransLM seems to rho*gamma, rho*Re_sigma as Solution variables
```
TobiKattmann, line 41:
```suggestion
// Note: TransLM seems to use rho*gamma, rho*Re_sigma as Solution variables, thus Conservative=true
CTransLMSolver::CTransLMSolver(void) : CTurbSolver(true) {}
```

SU2_CFD/src/solvers/CTransLMSolver.cpp
```diff
@@ -25,11 +25,15 @@
  * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
  */
 
-
+#include "../../include/solvers/CScalarSolver.inl"
```
TobiKattmann, line 28:
```suggestion
```

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -0,0 +1,301 @@
...
+#include "../../../Common/include/parallelization/omp_structure.hpp"
+#include "../variables/CScalarVariable.hpp"
+#include "CSolver.hpp"
+
+/*!
+ * \class CScalarSolver
+ * \brief Main class for defining a scalar solver.
+ * \tparam VariableType - Class of variable used by the solver inheriting from this template.
+ */
+template <class TVariable>
```
TobiKattmann, line 40:
```suggestion
template <class VariableType>
```

SU2_CFD/include/solvers/CScalarSolver.hpp
```diff
@@ -35,10 +34,10 @@
 
 /*!
  * \class CScalarSolver
- * \brief Main class for defining the turbulence model solver.
- * \ingroup Turbulence_Model
- * \author A. Bueno.
+ * \brief Main class for defining a scalar solver.
+ * \tparam TVariable - Class of C*Variable of the specific solver
  */
+template<class TVariable>
```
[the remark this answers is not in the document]
TobiKattmann: I forget to add a this to the batch ... and now had to re-add this suggestions because it is not visible in the changed files any more 🤷

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -30,9 +29,11 @@
 #include "../../../Common/include/parallelization/omp_structure.hpp"
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
 
-CScalarSolver::CScalarSolver(bool conservative) : CSolver(), Conservative(conservative) {}
+template<class TVariable>
+CScalarSolver<TVariable>::CScalarSolver(bool conservative) : CSolver(), Conservative(conservative) {}
```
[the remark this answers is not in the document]
TobiKattmann: I went with <VariableType> instead of <V> to use the same as in the hpp. Maybe it is a bit more consistent this way idk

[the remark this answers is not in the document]
TobiKattmann: See  6f8e69a

pcarruscag (approved):
Awesome work :+1: can't wait to see the crash diet that some solvers are about to go on.

TobiKattmann:
Thanks @pcarruscag for making this PR a lot better.

@bigfooted @danielmayer @mheimgartner 
To layout future steps as discussed after this merge: 
1. Merge these changes into feature_flamelet (and then feature_multicomp). Therefore the `CScalar*` things that are currently  present in these branches will be renamed to CLegacyScalar* to not overwrite files. This has the advantage that those branches are kept up-to-date with develop.
2. Implement a rudimentary (passive) CSpeciesSolver (specifically does not contain multicomponent mixing from #1332 ) and merge that. Finding a good (validation) Testcase would be great here.
3. Extend the CSpeciesSolver by #1332 mixing models. Either by cutting away all the flamelet specific parts or by porting the additions to a new PR.

Additionally the heat and radiation solver should be ported into this new structure, but that is independent work.

PR #1332, opened 2021-07-23
[WIP] Composition dependent fluid properties for multicomponent flow 

mheimgartner:
## Proposed Changes
Hi all, 

This draft PR contains all work on composition dependent fluid properties for multicomponent flow. 

What I did so far: 

•	Implemented new CFluidScalar class (which inherits from CFluidModel) and can be accessed via a new fluid model: MIXTURE_FLUID_MODEL. 
•	Verified the current implementation by comparing with Fluent.   
•	Added regression test case for two component incomp_rans flow with variable fluid properties. 

Future work:

•	Fix variable specific heat capacity and convergence issues in case of axisymmetric flow.  
•	Add possibility to have more than a single transported scalar (i.e. mixture containing more than two components). 
•	Remove flamelet implementation from feature_multicomp in order to have this work self-contained. **Note:** Currently this is a PR into feature_flamelet to make the diff easier

In case there are comments, ideas, foreseeable stepping stones from experience please let me know. 

## Related Work
#1223 by @danielmayer @bigfooted which feature_multicomp currently branches from. 
#1330 by @TobiKattmann which cuts the scalar transport necessary for this work out to merge it separately into develop

## PR Checklist


TobiKattmann:
@mheimgartner first of all congrats to this openend PR

For the builds in the tests, compiler warnings are treated like errors. I.e. the code compiles and works on your side but the test-builds won't if certain warnings are present (like unused variables). 
For more information where to look, see the `Details` button next to the individual builds. Click and wait a moment ... that will automatically scroll down for you. 
The failing builds will be for the same code bits probably, so looking at BaseNoMPI for the start should suffice.

To look in the `Details` and change code from there might be tedious so you can see the warnings locally by setting `./meson.py <your build folder> ... --warnlevel=3 ...`. Work through the warnings until there are none left -> commit the changes and push. That will probably calm the compiler-warnings-are-errors gods and grant you access to the regression test

SU2_CFD/include/fluid/CUnityLewisDiffusivity.hpp
```diff
@@ -0,0 +1,58 @@
...
+  CUnityLewisDiffusivity() {}
+ 
+  su2double GetDiffusivity() const override {return diff_;}
+
+  /*!
+   * \brief Set diffusivity.
+   */
+  void SetDiffusivity(su2double T, su2double rho, su2double mu_lam, su2double mu_turb, su2double cp, su2double kt) override { 
+    diff_ = kt / (Lewis * rho * cp); // is this possible? cp is calculated with mass fracties while wilke is calculated with mole fractions?
+    // also: don't kt, rho, cp need to be constant (take values of air?) want je hebt ook geen mixture lewis getal dus je moet wel k, cp, rho pakken van elke species.
```
TobiKattmann, line 51: as this is a draft PR is it of course the idea to keep own thoughts in here. But please in English so that us non-dutchies can follow 😉

Common/src/CConfig.cpp
```diff
@@ -3651,20 +3678,70 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
     }
   }
 
+  if (Molecular_Weight == nullptr){
+    Molecular_Weight = new su2double[1];
+    Molecular_Weight[0] = 28.96;
```
TobiKattmann, line 3683: There is a bit of hardcoding going here. Even for developing purposes it might be good if we tackle that

SU2_CFD/src/fluid/CFluidModel.cpp
```diff
@@ -61,6 +65,8 @@ void CFluidModel::SetLaminarViscosityModel(const CConfig* config) {
       SU2_MPI::Error("Viscosity model not available.", CURRENT_FUNCTION);
       break;
   }
+//  }
+  // else {}
```
TobiKattmann, line 69: Now that you have this nice 'diff-view' you might want to go over it a bit and get rid of this commented out code, unnecessarily added lines and white spaces whenever you know that its superfluous. That helps keeping the diff smaller and bit more focussed on your work. Also splitting overlong lines at suitable points is a good idea here and there.

I think that kind of cleanup work is not too hard but will help you having less work down the road.

WallyMaier:
Hi @mheimgartner, thanks for the PR.   I think this is a good use of the fluidmodel classes.  There seems to be a lot of potential to generalize some aspects of Mixture_Fluid_Model and CNEMOGas.   Im not sure what the best plan to approach this may be, but I think its something worth discussing.

bigfooted:
> Hi @mheimgartner, thanks for the PR. I think this is a good use of the fluidmodel classes. There seems to be a lot of potential to generalize some aspects of Mixture_Fluid_Model and CNEMOGas. Im not sure what the best plan to approach this may be, but I think its something worth discussing.

@WallyMaier I think there might be some overlap in fluid properties like mixing models as well, but I do not know what information you have available in a library like mutation++. I assume that it has general fluid property calculations and mixing rules available that could be used (wilke rule, kinetic models, etc)?

WallyMaier:
> > Hi @mheimgartner, thanks for the PR. I think this is a good use of the fluidmodel classes. There seems to be a lot of potential to generalize some aspects of Mixture_Fluid_Model and CNEMOGas. Im not sure what the best plan to approach this may be, but I think its something worth discussing.
> 
> @WallyMaier I think there might be some overlap in fluid properties like mixing models as well, but I do not know what information you have available in a library like mutation++. I assume that it has general fluid property calculations and mixing rules available that could be used (wilke rule, kinetic models, etc)?

@bigfooted Thats correct!  Both the CNEMOGas fluid models (SU2TCLib and CMutation++) have thermodynamic properties and transport properties.  I imagine there would be a smart way to reduce overlap, though Im not sure what the best approach would be.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1870,7 +1904,7 @@ void CIncEulerSolver::SetPreconditioner(const CConfig *config, unsigned long iPo
 
     if (energy && !flame) Preconditioner[nDim+1][0] = Cp*Temperature/BetaInc2;
     else        Preconditioner[nDim+1][0] = 0.0;
-
+
```
pcarruscag, line 1907: Set up your IDE to clear trailing spaces

TestCases/incomp_rans/multicomponentflow_variablefluidproperties/c6_cms_venturi_conference.cfg
```diff
@@ -0,0 +1,383 @@
...
+% Adaptive CFL number (NO, YES)
+CFL_ADAPT= NO
+
+% Parameters of the adaptive CFL number (factor down, factor up, CFL min value,
+%                                        CFL max value )
+CFL_ADAPT_PARAM= ( 1.5, 0.5, 1.0, 100.0 )
+
+
+% Runge-Kutta alpha coefficients
+RK_ALPHA_COEFF= ( 0.66667, 0.66667, 1.000000 )
```
pcarruscag, line 233:
```suggestion
```
Remove redundant options from the config.

pcarruscag (changes requested):
Some comments below to keep in mind when you continue the implementation.

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <vector>
+#include <memory>
+
+#include "CFluidModel.hpp"
+
+class CFluidScalar : public CFluidModel {
```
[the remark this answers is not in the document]
mheimgartner: Thank you Pedro for all your comments, really appreciate the help! Question regarding the CFluidScalar class which I made final, is it better to use override or final for its members, e.g.:

void SetLaminarViscosityModel(const CConfig* config) override;
vs
void SetLaminarViscosityModel(const CConfig* config) final;

If a member function is declared without virtual and with final, it means that it already overrides a virtual function in base class, so I figure it might be preferred to use final ? Also, do I need to state final  after every function member in CFluidScalar?

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -0,0 +1,69 @@
...
+  std::vector<su2double> moleFractions;  
+  std::vector<su2double> molarMasses;
+  std::vector<su2double> laminarViscosity;
+  std::vector<su2double> specificHeat; 
+  // std::vector<su2double> laminarViscosity = {1.099e-05, 1.823e-05}; 
+  // std::vector<su2double> laminarViscosity = {1.126e-05, 1.849e-05}; //corresponds with Fluent 
+  std::vector<su2double> laminarthermalConductivity; 
+  
+  unique_ptr<CViscosityModel> LaminarViscosityPointers[100];  
+  unique_ptr<CConductivityModel> ThermalConductivityPointers[100];
```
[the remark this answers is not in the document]
mheimgartner: Yes I hardcoded this because I don't know how to set the size corresponding to the number of species in the mixture. 
Right now, I changed it to e.g.:
std::vector<std::unique_ptr<CViscosityModel>> LaminarViscosityPointers;

and kept void CFluidScalar::SetLaminarViscosityModel(const CConfig* config) {} in CFluidScalar.cpp the same. Is this the correct way?

Common/include/CConfig.hpp
```diff
@@ -1603,13 +1605,15 @@ class CConfig {
    * \brief Get the value of the molecular weight for an incompressible ideal gas (g/mol).
    * \return Value of the molecular weight for an incompressible ideal gas (g/mol).
    */
-  su2double GetMolecular_Weight(void) const { return Molecular_Weight; }
-
+  su2double GetMolecular_Weight(void) const { return Molecular_Weight[0]; }
+  su2double GetMolecular_Weight (unsigned short val_index) const { return Molecular_Weight [val_index]; }
```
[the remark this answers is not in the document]
mheimgartner: 👍

Common/src/CConfig.cpp
```diff
@@ -3651,20 +3678,70 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
 
   if (SystemMeasurements == US) {
 
     /* Correct the viscosities, if they contain the default SI values. */
-    if(fabs(Mu_Constant-1.716E-5) < 1.0E-15) Mu_Constant /= 47.88025898;
-    if(fabs(Mu_Ref-1.716E-5)      < 1.0E-15) Mu_Ref      /= 47.88025898;
+    for(unsigned short iVar = 0; iVar < n_species; iVar++){
+      if(fabs(Mu_Constant[iVar]-1.716E-5) < 1.0E-15) Mu_Constant[iVar] /= 47.88025898;
+
+      if(fabs(Mu_Ref[iVar]-1.716E-5)      < 1.0E-15) Mu_Ref[iVar]      /= 47.88025898;
```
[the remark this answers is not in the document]
mheimgartner: I did not change all hardcoded values since I don't know exactly what these values resemble...

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -0,0 +1,69 @@
...
+  // unique_ptr<CViscosityModel[]> LaminarViscosityPointers(new CViscosityModel[100]); 
+  // auto LaminarViscosityPointers = std::make_unique<CViscosityModel[]>(100);
+
+  CFluidModel *FluidModel = nullptr;
+
+ public:
+
+  CFluidScalar(CConfig *config, su2double value_pressure_operating);
+
+  ~CFluidScalar() {};
```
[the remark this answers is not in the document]
mheimgartner: 👍

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  for(int iVar = 0; iVar < n_species_mixture; iVar++){
+    mixtureMolarMass += massFractions[iVar] / molarMasses[iVar]; 
+  }
+
+  for(int iVar = 0; iVar < n_species_mixture; iVar++){
+    moleFractions.at(iVar) = (massFractions[iVar] / molarMasses[iVar]) / mixtureMolarMass;
+  }
+  
+  return moleFractions;
```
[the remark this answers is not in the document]
mheimgartner: 👍

SU2_CFD/src/solvers/CIncEulerSolver.cpp
mheimgartner: 👍

TestCases/incomp_rans/multicomponentflow_variablefluidproperties/c6_cms_venturi_conference.cfg
mheimgartner: Will do 👍

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,277 @@
...
+  SetThermalConductivityModel(config); 
+}
+
+std::vector<su2double> CFluidScalar::massToMoleFractions(su2double* val_scalars){
+  su2double mixtureMolarMass = 0.0; 
+  su2double mixtureFractionDenumerator = 0.0; 
+  
+  // change if val_scalars becomes array of scalar su2double* 
+  massFractions.at(0) = val_scalars[0];
+  massFractions.at(1) = 1 - val_scalars[0];
```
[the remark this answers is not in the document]
mheimgartner: Oh no specific reason, thanks for the heads up 👍

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,277 @@
...
+su2double CFluidScalar::wilkeViscosity(su2double* val_scalars){
+  std::vector<su2double> phi; 
+  std::vector<su2double> wilkeNumerator;
+  std::vector<su2double> wilkeDenumeratorSum;
+  su2double wilkeDenumerator = 0.0; 
+  wilkeDenumeratorSum.clear();
+  wilkeNumerator.clear(); 
+  su2double viscosityMixture = 0.0;
+   
+  massToMoleFractions(val_scalars);
```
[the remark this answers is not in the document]
mheimgartner: Right now I only moved `massToMoleFractions` to `SetTDState_T`. I did not make the method `const` because `laminarViscosity` is altered by `wilkeViscosity`. I placed `laminarViscosity` within the function because the `laminarViscosity` value may change at every grid point, e.g. when using Sutherland law in case of non-isothermal flow. Is there a better approach to do this?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -504,15 +500,13 @@ void CIncEulerSolver::SetNondimensionalization(CConfig *config, unsigned short i
 
       case MIXTURE_FLUID_MODEL:
         fluidModel   = new CFluidScalar(config, Pressure_Thermodynamic);
-	n_scalars    = fluidModel->GetNScalars();
+	      n_scalars    = fluidModel->GetNScalars();
         dummy_scalar = new su2double[n_scalars]();
```
TobiKattmann: you should check the brakepads on your spacebar or tab

TobiKattmann:
I hope to have fixed all regression testcases over at feature_flamelet so please merge and see whether we can get some ✔️'s here as well

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+  switch (config->GetKind_ViscosityModel()) {
+    case VISCOSITYMODEL::CONSTANT:
+    /* Build a list of LaminarViscosity pointers to be used in e.g. wilkeViscosity to get the species viscosities. */
+      for (int iVar = 0; iVar < n_species_mixture; iVar++){
+        LaminarViscosityPointers[iVar] = std::unique_ptr<CConstantViscosity>(new CConstantViscosity(config->GetMu_Constant(iVar)));
+      }
+      break;
+    case VISCOSITYMODEL::SUTHERLAND:
+      for (int iVar = 0; iVar < n_species_mixture; iVar++){
+        LaminarViscosityPointers[iVar] = std::unique_ptr<CSutherland>(new CSutherland(config->GetMu_Ref(iVar), config->GetMu_Temperature_Ref(iVar), config->GetMu_S(iVar)));
```
TobiKattmann, line 57: is it possible that different species have different viscosity models? 

Currently not, it seems, but in case sth like `config->GetKind_ViscosityModel(>>iVar<<)` shows up we would need to swap the order of switch and for loop.

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+std::vector<su2double>& CFluidScalar::massToMoleFractions(const su2double * const val_scalars){
+  su2double mixtureMolarMass {0.0};
+
+  /* Change if val_scalars becomes array of scalar su2double*. */
+  massFractions[0] = val_scalars[0];
+  massFractions[1] = 1 - val_scalars[0];
+
+  for(int iVar = 0; iVar < n_species_mixture; iVar++){
+    mixtureMolarMass += massFractions[iVar] / molarMasses[iVar];
+  }
```
TobiKattmann, line 126: The top bit is hardcoded for `n_species_mixture=2` overall species, and the bottom for arbitrary many. Lets keeps in mind that at some point I think it is better to keep it open for arbitrary many species

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+  wilkeNumerator.clear();
+
+  for (int iVar = 0; iVar < n_species_mixture; iVar++){
+    ThermalConductivityPointers[iVar]->SetConductivity(Temperature, Density, Mu, Mu_Turb, Cp);
+    laminarthermalConductivity.at(iVar) = ThermalConductivityPointers[iVar]->GetConductivity();
+  }
+
+  for(int i = 0; i < n_species_mixture; i++){
+    for(int j = 0; j < n_species_mixture; j++){
+      phi.push_back(pow((1 + pow((laminarViscosity[i]) / (laminarViscosity[j]),0.5) * pow(molarMasses[j] / molarMasses[i],0.25)),2) / pow(8 * (1 + molarMasses[i] / molarMasses[j]),0.5));
```
TobiKattmann, line 211:
```suggestion
      phi.push_back(pow(1 + sqrt(laminarViscosity[i] / laminarViscosity[j]) * pow(molarMasses[j] / molarMasses[i],1./4),2) / sqrt(8 * (1 + molarMasses[i] / molarMasses[j])));
```

TobiKattmann: I wouldnt use sqrt() over pow(,0.5) for readabilty. And 1./4 is maybe a bit more intuitive than 0.25 (but that is debatable). Removed some superfluous braces as well, please check though

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+    wilkeDenumerator = 0.0;
+    phi.clear();
+    wilkeNumerator.push_back(moleFractions[i] * laminarthermalConductivity[i]);
+    conductivityMixture += wilkeNumerator[i] / wilkeDenumeratorSum[i];
+  }
+  return conductivityMixture;
+}
+
+unsigned long CFluidScalar::SetTDState_T(const su2double val_temperature, su2double * const val_scalars){
+  MeanMolecularWeight = 1/(val_scalars[0]/(molarMasses[0]/1000) + (1-val_scalars[0])/(molarMasses[1]/1000));
```
TobiKattmann, line 224: I have the feeling with this multicompent-stuff there is a bit of `Mean...` computation going on. I would propose to extract those computations in functions that return the mean so that they could be used in multiple places ... but even if it is just used in one place. i.e. some 
```c++
su2double ??::ComputeMeanMolecularWeight(...) { ...; return MeanMolecularWeight}
```

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Cp = specificHeat[0] * val_scalars[0] + specificHeat[1] * (1- val_scalars[0]);
+  Cp = 1009.39;
+  // Cp = 1009.3;
+  Cv = Cp/1.4;
```
TobiKattmann, line 229:
```suggestion
  constexpr su2double CpAir350Kelvin = 1009.39;
  constexpr su2double RatioSpecificHeatsAir = 1.4;
  Cp = CpAir350Kelvin ;
  Cv = Cp/RatioSpecificHeatsAir ;
```

TobiKattmann: I know this is was just intermediate hardcoding but lets make sure we use 'more readable hardcoding'

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -0,0 +1,245 @@
...
+  if(wilke){
+    Mu  = wilkeViscosity(val_scalars);
+  }
+  else if(davidson){
+    Mu = davidsonViscosity(val_scalars);
+  }
+
+  Kt = wilkeConductivity(val_scalars);
+
+  return 0;
```
TobiKattmann, line 244: I guess you didn't write the function prototype but returning sth without use is not good. But if it comes from feature_flamelet do not change

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -892,7 +905,7 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
   if(config->GetKind_TimeIntScheme_Flow()==EULER_IMPLICIT){
     SetVolumeOutputValue("TIMESTEP", iPoint, Node_Flow->GetDelta_Time(iPoint));
   }
-    
+
```
TobiKattmann, line 908: please do not run TrimTrailingWhitespaces on those large files, it just inflates the diff and is done best in seperate PR's or at the very end. (And yes I am aware of the Hide trailing whitespaces option on github ;) ) You might get merge conflicts just because of that

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
```diff
@@ -236,10 +238,41 @@ CPassiveScalarSolver::~CPassiveScalarSolver(void) {
...
+                        (config->GetInnerIter() <= config->GetLimiterIter());
+
+  for (auto i_point = 0u; i_point < nPoint; i_point++) {
+
+    CFluidModel * fluid_model_local = solver_container[FLOW_SOL]->GetFluidModel();
+    su2double * scalars = nodes->GetSolution(i_point);
+
+    su2double Temperature = solver_container[FLOW_SOL]->GetNodes()->GetTemperature(i_point);
+
+    n_not_in_domain += fluid_model_local->SetTDState_T(Temperature, scalars); /*--- first argument (temperature) is not used ---*/
```
TobiKattmann, line 255: ah, I guess for lookuptables this returns how many points were outside the triangulation @danielmayer ?

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
```diff
@@ -326,85 +359,97 @@ void CPassiveScalarSolver::SetInitialCondition(CGeometry **geometry,
...
       for (iVar = 0; iVar < nVar; iVar++) {
-        
+
         total_index = iPoint*nVar+iVar;
-        
+
         su2double c = nodes->GetSolution(iPoint,iVar);
-        
+        dRhodC = -Density*((MolecularWeight[1]-MolecularWeight[0])/(MolecularWeight[0]+(MolecularWeight[1]-MolecularWeight[0])*c));
+
```
TobiKattmann, line 428: @bigfooted @danielmayer we should also consolidate our nomenclature and be consistent throughout code and equations. Massfractions `c` or `Y` for example?

Common/src/CConfig.cpp
```diff
@@ -4730,7 +4730,8 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
 
   if (Kind_Solver == INC_EULER || Kind_Solver == INC_NAVIER_STOKES || Kind_Solver == INC_RANS) {
     if ((Kind_DensityModel == INC_DENSITYMODEL::CONSTANT) || (Kind_DensityModel == INC_DENSITYMODEL::BOUSSINESQ))
-      Kind_FluidModel = CONSTANT_DENSITY;
+      if (Kind_FluidModel != CONSTANT_DENSITY)
+        SU2_MPI::Error("Incompressible problems with DENSITY_MODEL = CONSTANT or DENSITY_MODEL = BOUSSINESQ must use FLUID_MODEL = CONSTANT_DENSITY.", CURRENT_FUNCTION);
```
TobiKattmann: Problem here was that the default `FLUID_MODEL `is `STANDARD_AIR `and the failing reg cases relied upon this code bit  changing the not-set `FLUID_MODEL `to `CONSTANT_DENSITY`. I prefer being more strict in the config (erroring instead of silently changing), because this would also silently change the  `FLUID_MODEL` back to `CONSTANT_DENSITY` although the user provided a different  `FLUID_MODEL`. This, to some extent, also forces one to learn more about what combinations are reasonable and why others aren't. 

This wont be handeled here but maybe somewhere else

Common/src/CConfig.cpp
```diff
@@ -3675,20 +3685,145 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
+      }
+      break;
+    case VISCOSITYMODEL::SUTHERLAND:
+      if ((nMu_Ref != n_species) || (nMu_Temperature_Ref != n_species) || (nMu_S != n_species)) {
+        SU2_MPI::Error("The use of MIXTURE_FLUID_MODEL requires the number of entries for MU_REF, MU_T_REF and SUTHERLAND_CONSTANT,\n"
+                       "to be equal to the number of entries of SCALAR_INIT + 1", CURRENT_FUNCTION);
+      }
+      break;
+    default:
+      SU2_MPI::Error("Viscosity model not available.", CURRENT_FUNCTION);
```
mheimgartner, line 3776:
```suggestion
if (n_species != 1)      SU2_MPI::Error("Viscosity model not available.", CURRENT_FUNCTION);
```

Common/src/CConfig.cpp
```diff
@@ -3675,20 +3685,145 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
+        }
+      } else {
+          if (nPrandtl_Lam != n_species) {
+            SU2_MPI::Error("The use of MIXTURE_FLUID_MODEL requires the number of entries for PRANDTL_LAM,\n"
+                           "to be equal to the number of entries of SCALAR_INIT + 1", CURRENT_FUNCTION);
+          } 
+      }
+      break;
+    default:
+      SU2_MPI::Error("Conductivity model not available.", CURRENT_FUNCTION);
```
mheimgartner, line 3808:
```suggestion
if (n_species != 1)      SU2_MPI::Error("Conductivity model not available.", CURRENT_FUNCTION);
```

SU2_CFD/src/fluid/CFluidScalar.cpp
mheimgartner: Yes I agree, I will change that somewhere down the road.

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -221,12 +221,13 @@ su2double CFluidScalar::wilkeConductivity(const su2double * const val_scalars){
 }
 
 unsigned long CFluidScalar::SetTDState_T(const su2double val_temperature, su2double * const val_scalars){
-  MeanMolecularWeight = 1/(val_scalars[0]/(molarMasses[0]/1000) + (1-val_scalars[0])/(molarMasses[1]/1000));
+  CFluidModel::GetMeanMolecularWeight(molarMasses, val_scalars);
```
TobiKattmann, line 224: Not sure how in how many places GetMeanMolecularWeight will be used and whether it justifies having it as class data. For now I would prefer taking it out of the class and only computing when it is needed and returning the value ... it even is `su2double GetMeanMolecularWeight` right now. In the current state it would make more sense to have `void GetMeanMolecularWeight`

so here:
```c++
const su2double MeanMolecularWeight = GetMeanMolecularWeight(molarMasses, val_scalars);
```
Then GetMeanMolecularWeight needs to be under protected methods I guess

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -73,6 +73,8 @@ void CFlowOutput::AddAnalyzeSurfaceOutput(const CConfig *config){
   AddHistoryOutput("AVG_CO",                   "Avg_CO",                    ScreenOutputFormat::SCIENTIFIC, "SPECIES_COEFF", "Total average mass fraction of CO on all markers set in MARKER_ANALYZE", HistoryFieldType::COEFFICIENT);
```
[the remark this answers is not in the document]
mheimgartner: Yes that is probably a good idea to have a look at somewhere in the near future...

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -7,6 +7,7 @@
 
 class CFluidScalar final : public CFluidModel {
 private:
+  unsigned short n_scalars = 0.0;                     /*!< \brief Number of transported scalars. */
   unsigned short n_species_mixture = 0.0;             /*!< \brief Number of species in mixture. */
```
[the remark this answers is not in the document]
mheimgartner: I think it's best to fix this at a later stage when nVar is introduced throughout the whole scalar base.

SU2_CFD/src/output/CFlowOutput.cpp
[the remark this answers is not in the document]
TobiKattmann: That definitely has to leave with something like a `SPECIES_NAMES= CO, ...` list that gets automatically appended

Common/src/CConfig.cpp
```diff
@@ -3675,20 +3685,145 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
+  }
+
+  if (Prandtl_Turb == nullptr){
+    Prandtl_Turb = new su2double[1];
+    Prandtl_Turb[0] = Prandtl_Turb_Default;
+    nPrandtl_Turb = 1;
+  }
+
+  /*--- Check whether inputs for MIXTURE_FLUID_MODEL are correctly specified. ---*/
+  unsigned short n_species = 1; //TODO TK:: make it static?
```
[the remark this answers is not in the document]
TobiKattmann: Subsequent to #1330 there will be SpeciesSolver that will purely operate on the basis of nVar (i.e. nspecies=nVar+1) and nspecies will at most be used in very local scopes

SU2_CFD/src/numerics/flow/flow_sources.cpp
```diff
@@ -313,22 +313,32 @@ CNumerics::ResidualType<> CSourceIncAxisymmetric_Flow::ComputeResidual(const CCo
...
       /*--- Viscous terms. ---*/
 
       residual[0] -= 0.0;
-      residual[1] -= Volume*(yinv*tau[0][1] - TWO3*AuxVar_Grad_i[0][0]);
-      residual[2] -= Volume*(yinv*2.0*total_viscosity*PrimVar_Grad_i[2][1] -
-                             yinv* yinv*2.0*total_viscosity*Velocity_i[1] -
-                             TWO3*AuxVar_Grad_i[0][1]);
+      //residual[1] -= Volume*(yinv*tau[0][1] - TWO3*AuxVar_Grad_i[0][0]);
+      //residual[2] -= Volume*(yinv*2.0*total_viscosity*PrimVar_Grad_i[2][1] -
+      //                       yinv* yinv*2.0*total_viscosity*Velocity_i[1] -
```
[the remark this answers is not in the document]
TobiKattmann: I dont understand it either? @mheimgartner 
(maybe removing the viscous contribution?)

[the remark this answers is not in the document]
bigfooted: This was my doing, I did some test where I removed the terms related to  the divergence of velocity, since they should be 0 for low mach cases. I did not find any big changes in convergence. I didn't do a lot of tests, but I think it is safe to simply keep the nabla.V terms.

SU2_CFD/src/output/CFlowOutput.cpp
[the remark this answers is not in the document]
bigfooted: I have a local implementation of this...

SU2_CFD/src/solvers/CPassiveScalarSolver.cpp
```diff
@@ -237,9 +241,49 @@ void CPassiveScalarSolver::Preprocessing(CGeometry *geometry, CSolver **solver_c
...
+    for(auto iVar = 0u; iVar < nVar; iVar++){
+      nodes->SetDiffusivity(i_point, mass_diffusivity, iVar);
+    }
+
+    // for(auto iVar = 0u; iVar < nVar; iVar++){
+    //   nodes->SetDiffusivity(i_point, fluid_model_local->GetMassDiffusivity(), iVar);
+    // }
+
+    if (!Output) LinSysRes.SetBlock_Zero(i_point);
```
TobiKattmann, line 282: Why was this added? Not even 10 lines below there is this code which should do the same. please enlighten me
```c++
  if (!ReducerStrategy && !Output) {
    LinSysRes.SetValZero();
    LinSysRes.SetValZero();
    if (implicit) Jacobian.SetValZero();
    if (implicit) Jacobian.SetValZero();
    else {SU2_OMP_BARRIER}
    else {SU2_OMP_BARRIER}
  }
  }
```

TobiKattmann:
```suggestion
```

TestCases/multicomp_regression.py
```diff
@@ -0,0 +1,102 @@
...
+    multicompflow_variableprop_venturi_planar_outlettargetmassflowrate.timeout   = 1600
+    multicompflow_variableprop_venturi_planar_outlettargetmassflowrate.new_output = True
+    multicompflow_variableprop_venturi_planar_outlettargetmassflowrate.tol       = 0.00001
+    test_list.append(multicompflow_variableprop_venturi_planar_outlettargetmassflowrate)
+
+    multicompflow_variableprop_venturi_axisymmetric_velocityinlets           = TestCase('multicompflow_variableprop_venturi_axisymmetric_velocityinlets')
+    multicompflow_variableprop_venturi_axisymmetric_velocityinlets.cfg_dir   = "incomp_rans/multicomponentflow_variablefluidproperties"
+    multicompflow_variableprop_venturi_axisymmetric_velocityinlets.cfg_file  = "C6_pneumatic_venturi_axisymmetric_velocityinlets.cfg"
+    multicompflow_variableprop_venturi_axisymmetric_velocityinlets.test_iter = 5
+    multicompflow_variableprop_venturi_axisymmetric_velocityinlets.test_vals = [-5.658917, -4.475284, -4.442651, -5.811396, 20, -1.853076, 4, -3.547922]
```
TobiKattmann, line 72:
```suggestion
    multicompflow_variableprop_venturi_axisymmetric_velocityinlets.test_vals = [-5.658917, -4.475284, -4.442651, -5.811413, 20, -1.853076, 4, -3.547922]
```

TobiKattmann:
The SST-2003 changes are in 33f9591 and were reverted directly afterwards in 8208c9e such that we have a commit with the necessary changes but they are not applied in the latest version of this code. This was done to revisit and finish this at a later stage.

PR #1340, opened 2021-07-30
SU2-NEMO - Optimize initialization time

fmpmorgado:
## Proposed Changes

Optimize the SU2-NEMO solver initialization time 

## Related Work
None that I remember

## PR Checklist

- [ X ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [ X ] I have added a test case that demonstrates my contribution, if necessary.
- [ X ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

## Work Overview

I have optimized the initialization time of SU2-NEMO. You can check the time comparison between the old and new version at the table below.

image

Some steps that I have done:
### CNEMOEulerVariable.cpp 
* Removed constant calculations inside a for loop
* Passed the Solution_Old = Solution to outside the loop

### CNEMOEulerSolver.cpp
* Removed the "for" loop that fills the Nodes Primitive Fields (already done in the Preprocessing phase)
* Removed the error counter (already done in the Preprocessing phase)

### CSolverFactory.cpp
* Removed the Preprocessing step after the CNEMOEulerSolver instantiation (**This is done before the start of the iterative process, regardless. Additionally, the Navier-Stokes do not have any preprocess phase after the instantiation, why is that ?** )

The residuals may slightly different, because of the removal of the for loop inside CNEMOEulerSolver.cpp. Notice that inside that for loop, the soundspeed is obtained using the function FluidModel->GetSoundSpeed(), which gives a slight different value than the one calculated inside CNEMOEulerVariable.cpp using FluidModel->ComputeSoundSpeed(). This means that the regression tests may fail.

### DIFF
--- a/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
@@ -256,7 +256,7 @@ void CNEMOEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver
       config->SetNonphysical_Points(ErrorCounter);
       
       if ((rank == MASTER_NODE) && (ErrorCounter != 0))
-        cout << "Warning. The original solution contains "<< ErrorCounter << " points that are not physical." << endl;
+        cout << "Warning. The initial solution contains "<< ErrorCounter << " points that are not physical." << endl;
   }
 
   /*--- Artificial dissipation ---*/
--- a/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
@@ -46,6 +46,7 @@ CNEMOEulerSolver::CNEMOEulerSolver(CGeometry *geometry, CConfig *config,
     description = "Euler";
   }
 
+  vector<su2double> Energies_Inf;
   unsigned long iPoint, counter_local, counter_global = 0;
   unsigned short iDim, iMarker, iSpecies, nLineLets;
   unsigned short nZone = geometry->GetnZone();
@@ -60,6 +61,8 @@ CNEMOEulerSolver::CNEMOEulerSolver(CGeometry *geometry, CConfig *config,
   string filename_ = "flow";
 
   bool nonPhys;
+
+  Energies_Inf.resize(2);
 
   /*--- Store the multigrid level. ---*/
   MGLevel = iMesh;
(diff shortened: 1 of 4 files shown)


SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -47,7 +47,8 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
 
   vector<su2double> energies;
   unsigned short iDim, iSpecies;
-  su2double soundspeed, sqvel, rho;
+  su2double soundspeed, rho;
+  su2double sqvel = 0.0;
```
pcarruscag, line 51:
```suggestion
```

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -47,7 +47,8 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
 
   vector<su2double> energies;
```
pcarruscag, line 48:
```suggestion
```

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -143,23 +144,20 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
...
+  /*--- Set mixture state ---*/
+  fluidmodel->SetTDStatePTTv(val_pressure, val_massfrac, val_temperature, val_temperature_ve);
+
+  /*--- Compute necessary quantities ---*/
+  rho = fluidmodel->GetDensity();
+  soundspeed = fluidmodel->ComputeSoundSpeed();
+  for (iDim = 0; iDim < nDim; iDim++){
+    sqvel += val_mach[iDim]*soundspeed * val_mach[iDim]*soundspeed;
+  }
+  energies = fluidmodel->ComputeMixtureEnergies();
```
pcarruscag, line 156:
```suggestion
  const su2double rho = fluidmodel->GetDensity();
  const su2double soundspeed = fluidmodel->ComputeSoundSpeed();
  const su2double sqvel = GeometryToolbox::SquaredNorm(nDim, val_mach) * pow(soundspeed,2);
  const auto& energies = fluidmodel->ComputeMixtureEnergies();
```
You may need the include for geometry_toolbox.hpp.
Be aware of the methods in CNEMOGas that return by reference, those little copies of std::vector add up and it is an easy fix.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -303,7 +254,9 @@ void CNEMOEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver
       unsigned long tmp = ErrorCounter;
       SU2_MPI::Allreduce(&tmp, &ErrorCounter, 1, MPI_UNSIGNED_LONG, MPI_SUM, SU2_MPI::GetComm());
       config->SetNonphysical_Points(ErrorCounter);
-
+      
+      if ((rank == MASTER_NODE) && (ErrorCounter != 0))
+        cout << "Warning. The original solution contains "<< ErrorCounter << " points that are not physical." << endl;
```
pcarruscag, line 259: original -> initial ?

SU2_CFD/src/solvers/CSolverFactory.cpp
```diff
@@ -429,7 +429,7 @@ CSolver* CSolverFactory::CreateFlowSolver(SUB_SOLVER_TYPE kindFlowSolver, CSolve
       break;
     case SUB_SOLVER_TYPE::NEMO_EULER:
       flowSolver = new CNEMOEulerSolver(geometry, config, iMGLevel);
-      flowSolver->Preprocessing(geometry, solver, config, iMGLevel, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
+      //flowSolver->Preprocessing(geometry, solver, config, iMGLevel, NO_RK_ITER, RUNTIME_FLOW_SYS, false);
```
pcarruscag, line 432: Don't leave commented code please.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -168,14 +166,9 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
 
     Solution(iPoint,nSpecies+nDim)       = rho*(energies[0]+0.5*sqvel);
     Solution(iPoint,nSpecies+nDim+1)     = rho*(energies[1]);
-
-    Solution_Old = Solution;
```
pcarruscag, line 172: If you suspect this may be happening somewhere else I can tell you how to instrument the code to find it.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: Yes, just copied the warning message that was there before. Will change it.

SU2_CFD/src/solvers/CSolverFactory.cpp
fmpmorgado: I left as a comment first, because I am still not sure if I should delete this line of code or not. Since the CEulerSolver uses it.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
fmpmorgado: Thanks Pedro, I still need to check what is happening at each iteration, but first wanted to sort out the initialization issue

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
fmpmorgado: That is true, and may be the main reason of the slowness of SU2-NEMO iterations. I will try to address that in my next PR.

pcarruscag (approved):

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
pcarruscag: :+1: You can have a look at #716 (scroll down) for how to use perf-tools to profile the code and point you to the hot spots "first rule of performance is that you need to measure it". There should be hotter areas in NEMO than vector copies.

WallyMaier:
Thanks @fmpmorgado!   NEMO definitely needs some speed up! This looks good!

Regarding the preprocessing, why is the ComputeSoundSpeed() function returning slightly different results?

fmpmorgado:
> Thanks @fmpmorgado! NEMO definitely needs some speed up! This looks good!
> 
> Regarding the preprocessing, why is the ComputeSoundSpeed() function returning slightly different results?

So, commenting as it was the version of NEMO before my changes:

Inside the CNEMOEulerVariable, you set the fluidmodel with the freestream conditions written in the config file and use the ComputeSoundSpeed().

However, in the CNEMOEulerSolver, in the deleted loop, you had GetSoundSpeed(), which retrieves the solution already estabilished in the Primitive Matrix. The Primitive Matrix was estabilished inside the SetPrimVar() function. However, inside the function, we set the fluidmodel with slightly different values than the ones used in the config file, because we first run the iterative process to find the translational and vibrational temperature with the given energies. This iterative process provides slightly different temperatures than the ones in the config file, thus computing slightly different sound speeds

jtneedels (approved):

PR #1341, opened 2021-07-31
Fix "per-surface" outputs

### LINKED ISSUE #1336: "Per surface" outputs do not work
The "per surface" problem starts in `COutput::CheckHistoryOutput` where the table header is not given the right number of columns. The function does not add the "per surface names" because we are using the same identifier for normal outputs and "per surface" outputs.
Moreover, the "per surface" implementation is only considering the first marker name and value for printing.
However, there is one value added per "marker analyse".
Let's fix this in a different PR.

_Originally posted by @pcarruscag in https://github.com/su2code/SU2/issues/1226#issuecomment-889067006_

pcarruscag:
Fixes #1336

### DIFF
--- a/SU2_CFD/include/output/COutput.hpp
+++ b/SU2_CFD/include/output/COutput.hpp
@@ -57,19 +57,21 @@ protected:
 
   /*----------------------------- General ----------------------------*/
 
-  int rank,     /*!< \brief MPI Rank. */
-  size;         /*!< \brief MPI Size. */
+  const int rank;     /*!< \brief MPI Rank. */
+  const int size;     /*!< \brief MPI Size. */
 
-  unsigned short nDim;   /*!< \brief Physical Dimension */
+  const unsigned short nDim;   /*!< \brief Physical Dimension */
 
-  bool multiZone,       /*!< \brief Boolean to store whether we are running a multizone problem */
-  gridMovement,         /*!< \brief Boolean to store whether we have grid movement enabled */
-  femOutput;            /*!< \brief Boolean to store whether we should use the FEM routines */
+  const bool multiZone;     /*!< \brief Boolean to store whether we are running a multizone problem */
+  const bool gridMovement;  /*!< \brief Boolean to store whether we have grid movement enabled */
+  const bool femOutput;     /*!< \brief Boolean to store whether we should use the FEM routines */
+  const bool si_units;
+  const bool us_units;
 
   /*----------------------------- Screen and history output ----------------------------*/
 
+  const unsigned short fieldWidth = 12; /*!< \brief Width of each column for the screen output (hardcoded for now) */
   string historySep;              /*!< \brief Character which separates values in the history file */
-  unsigned short fieldWidth;      /*!< \brief Width of each column for the screen output (hardcoded for now) */
   bool noWriting;                 /*!< \brief Boolean indicating whether a screen/history output should be written */
   unsigned long curTimeIter,      /*!< \brief Current value of the time iteration index */
   curAbsTimeIter,                 /*!< \brief Current value of the time iteration index */
@@ -250,7 +252,7 @@ public:
   /*!
    * \brief Constructor of the class.
    */
-  COutput(CConfig *config, unsigned short nDim, bool femOutput);
+  COutput(const CConfig *config, unsigned short nDim, bool femOutput);
 
   /*!
    * \brief Preprocess the volume output by setting the requested volume output fields.
--- a/SU2_CFD/include/output/COutput.hpp
+++ b/SU2_CFD/include/output/COutput.hpp
@@ -532,8 +532,9 @@ protected:
    * \param[in] value - The new value of this field.
    */
   inline void SetHistoryOutputValue(string name, su2double value){
-    if (historyOutput_Map.count(name) > 0){
-      historyOutput_Map[name].value = value;
+    auto it = historyOutput_Map.find(name);
+    if (it != historyOutput_Map.end()){
+      it->second.value = value;
     } else {
       SU2_MPI::Error(string("Cannot find output field with name ") + name, CURRENT_FUNCTION);
     }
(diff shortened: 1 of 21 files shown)


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -843,23 +838,24 @@ void CFlowOutput::WriteMetaData(const CConfig *config){
   meta_file.close();
 }
 
-void CFlowOutput::WriteForcesBreakdown(CConfig *config, CGeometry *geometry, CSolver **solver_container){
+void CFlowOutput::WriteForcesBreakdown(const CConfig* config, const CSolver* flow_solver) const {
```
pcarruscag, line 841: Most of that last commit was trying to make this function less ugly.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -843,23 +838,24 @@ void CFlowOutput::WriteMetaData(const CConfig *config){
...
+  const bool compressible = (config->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE);
+  const bool incompressible = (config->GetKind_Regime() == ENUM_REGIME::INCOMPRESSIBLE);
+  const bool unsteady = config->GetTime_Domain();
+  const bool viscous = config->GetViscous();
+  const bool dynamic_grid = config->GetDynamic_Grid();
+  const bool gravity = config->GetGravityForce();
+  const bool turbulent = config->GetKind_Solver() == RANS;
+  const bool fixed_cl = config->GetFixed_CL_Mode();
+  const auto Kind_Solver = config->GetKind_Solver();
   const auto Kind_Turb_Model = config->GetKind_Turb_Model();
```
pcarruscag, line 856:
```suggestion
  const auto Kind_Turb_Model = config->GetKind_Turb_Model();
  const bool turbulent = Kind_Turb_Model != NONE;
  const bool fixed_cl = config->GetFixed_CL_Mode();
  const auto Kind_Solver = config->GetKind_Solver();
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -46,15 +46,17 @@
...
-COutput::COutput(CConfig *config, unsigned short nDim, bool fem_output): femOutput(fem_output) {
+COutput::COutput(const CConfig *config, unsigned short ndim, bool fem_output):
+  rank(SU2_MPI::GetRank()),
+  size(SU2_MPI::GetSize()),
+  nDim(ndim),
+  multiZone(config->GetMultizone_Problem()),
+  gridMovement(config->GetGrid_Movement()),
+  femOutput(fem_output),
+  si_units(config->GetSystemMeasurements() == SI),
+  us_units(config->GetSystemMeasurements() == US) {
```
TobiKattmann, line 57: 👍 nice to avoid these repetitions

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1497,39 +1480,39 @@ void CFlowOutput::WriteForcesBreakdown(const CConfig* config, const CSolver* flo
         break;
 
       case INC_IDEAL_GAS:
-        file << "Fluid Model: INC_IDEAL_GAS " << endl;
-        file << "Variable density incompressible flow using ideal gas law." << endl;
-        file << "Density is a function of temperature (constant thermodynamic pressure)." << endl;
-        file << "Specific heat at constant pressure (Cp): " << config->GetSpecific_Heat_Cp() << " N.m/kg.K." << endl;
-        file << "Molecular weight : " << config->GetMolecular_Weight() << " g/mol" << endl;
-        file << "Specific gas constant: " << config->GetGas_Constant() << " N.m/kg.K." << endl;
+        file << "Fluid Model: INC_IDEAL_GAS \n";
```
TobiKattmann, line 1483: I once saw a video from Mr Turner that using endl to line break is much slower compared to \n. Was that you primary motivation or are there other considerations to this?

Ok it was actually so important to him that it was [Ep 7 of c++ weekly](https://www.youtube.com/watch?v=GMqQOEZYVJQ) back in the good'ol days

TobiKattmann (approved):
Well thanks for this fix(es) and massive cleanup 💐 Someone at some point will surely appreciate that tedious work!

> Moreover, the "per surface" implementation is only considering the first marker name and value for printing.
> However, there is one value added per "marker analyse".

Can you point me to where that went wrong real quick. I was not able to spot that while going over 😬

TestCases/navierstokes/flatplate/lam_flatplate.cfg
```diff
@@ -87,6 +87,7 @@ MARKER_PLOTTING= ( wall )
 %
 % Marker(s) of the surface where the functional (Cd, Cl, etc.) will be evaluated
 MARKER_MONITORING= ( wall )
+MARKER_ANALYZE= (inlet, outlet)
```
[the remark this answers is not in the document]
pcarruscag: We don't seem to.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -57,19 +57,21 @@ class COutput {
...
-  femOutput;            /*!< \brief Boolean to store whether we should use the FEM routines */
+  const bool multiZone;     /*!< \brief Boolean to store whether we are running a multizone problem */
+  const bool gridMovement;  /*!< \brief Boolean to store whether we have grid movement enabled */
+  const bool femOutput;     /*!< \brief Boolean to store whether we should use the FEM routines */
+  const bool si_units;
+  const bool us_units;
 
   /*----------------------------- Screen and history output ----------------------------*/
 
+  const unsigned short fieldWidth = 12; /*!< \brief Width of each column for the screen output (hardcoded for now) */
```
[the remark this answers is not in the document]
pcarruscag: Me neither :) it's the column width it does not affect the history, only screen.

SU2_CFD/src/output/CFlowOutput.cpp
pcarruscag: Yep that as well, and also clang-format looked nicer.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1069,44 +1061,48 @@ void COutput::SetScreen_Header(const CConfig *config) {
...
+            PrintingToolbox::PrintScreenScientific(out, field.value, fieldWidth);
+            break;
+          case ScreenOutputFormat::PERCENT:
+            PrintingToolbox::PrintScreenPercent(out, field.value, fieldWidth);
+            break;
+        }
+        (*convergenceTable) << out.str();
       }
     }
-    (*convergenceTable) << out.str();
```
pcarruscag, line 1109: We were re-using this string stream and that created the lumped values without separator.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1069,44 +1061,48 @@ void COutput::SetScreen_Header(const CConfig *config) {
...
+      for (const auto& field : it2->second) {
+        stringstream out;
+        switch (field.screenFormat) {
+          case ScreenOutputFormat::INTEGER:
+            PrintingToolbox::PrintScreenInteger(out, SU2_TYPE::Int(field.value), fieldWidth);
+            break;
+          case ScreenOutputFormat::FIXED:
+            PrintingToolbox::PrintScreenFixed(out, field.value, fieldWidth);
+            break;
+          case ScreenOutputFormat::SCIENTIFIC:
```
pcarruscag, line 1096: I should have made this block a lambda or something for re-use, one day...

PR #1344, opened 2021-08-03
Fix slidingmesh for SA

maxaehle:
## Proposed Changes
`CSlidingInterface::GetDonor_Variable` needs to access the `Primitive` of flow variables and the `Solution` of turbulent variables. Which one is the case, is currently decided by checking `donor_solution->GetnPrimVar() == 2` (true->turbulent, false->flow). If the SA turbulence model is used, the non-overidden "virtual placeholder" `CVariable::GetPrimitive(unsigned long,unsigned long) const` returning zero is incorrectly called.

We propose to fix this by additionally checking for `donor_solution->GetnPrimVar() == 1`.

## Related Work
@maxaehle, @mcmorelli, @BeckettZhou, @EduardoMolina

## PR Checklist


SU2_CFD/src/interfaces/cfd/CSlidingInterface.cpp
```diff
@@ -51,20 +51,24 @@ void CSlidingInterface::GetDonor_Variable(CSolver *donor_solution, CGeometry *do
   nDonorVar = donor_solution->GetnPrimVar();
 
   /*---  the number of primitive variables is set to two by default for the turbulent solver ---*/
-  bool turbulent = (nDonorVar == 2) ;
+  bool SA, SST;
+  SA  = (nDonorVar == 1);
```
WallyMaier, line 55: can you do "bool SA = (nDonor==1) ", etc?

SU2_CFD/src/interfaces/cfd/CSlidingInterface.cpp
```diff
@@ -51,20 +51,24 @@ void CSlidingInterface::GetDonor_Variable(CSolver *donor_solution, CGeometry *do
...
-
-    /*---  for turbulent solver retrieve solution and set it as the donor variable ---*/
+  if (SA){
+    /*---  for SA turbulent solver retrieve solution and set it as the donor variable ---*/
+    Donor_Variable[0] = donor_solution->GetNodes()->GetSolution(Point_Donor,0);
+  }
+  else if (SST){
+    /*---  for SST turbulent solver retrieve solution and set it as the donor variable ---*/
     Donor_Variable[0] = donor_solution->GetNodes()->GetSolution(Point_Donor,0);
     Donor_Variable[1] = donor_solution->GetNodes()->GetSolution(Point_Donor,1);
```
pcarruscag, line 65: Do a loop over nVar...

pcarruscag (changes requested):

maxaehle:
I followed @pcarruscag 's request in commit 12974ab.
BTW it's very confusing that the SA/SST turbulent solvers have `nPrimVar==1,2` but `GetPrimitive` is not overridden and thus returns zero. So another fix to the bug considered in this PR would be to override `GetPrimitive` on the turbulent solvers such that it forwards to `GetSolution`?

SU2_CFD/src/interfaces/cfd/CSlidingInterface.cpp
```diff
@@ -47,24 +47,18 @@ void CSlidingInterface::GetDonor_Variable(CSolver *donor_solution, CGeometry *do
...
                                           unsigned long Vertex_Donor, unsigned long Point_Donor) {
 
-  unsigned short iVar, nDonorVar;
-  nDonorVar = donor_solution->GetnPrimVar();
-
-  /*---  the number of primitive variables is set to two by default for the turbulent solver ---*/
-  bool turbulent = (nDonorVar == 2) ;
-
+  const unsigned short nDonorVar = donor_solution->GetnPrimVar();
+  const bool turbulent = (nDonorVar==1) || (nDonorVar==2); // SA or SST
```
pcarruscag, line 51:
```suggestion
  const auto nDonorVar = donor_solution->GetnPrimVar();
  /// TODO: Replace with approach compatible with any number of variables (e.g. encapsulate in a "solver info" object).
  const bool turbulent = nDonorVar <= 2;
```

pcarruscag (approved):
Thanks Max, I missed the solution v primitive part of the problem.

pcarruscag:
I don't think forwarding GetPrimitive to GetSolution by default is a good option, a "Slinding state" layer could be appropriate but that does not need to be fixed right now.

SU2_CFD/src/interfaces/cfd/CSlidingInterface.cpp
maxaehle: Yes, `GetnPrimVar()` seems to be non-zero on all flow and turbulence solvers.

maxaehle:
By "Sliding State layer", do you mean a virtual `CVariable::GetSlidingMeshDonorVariable` that returns 0 by default, but forwards to `GetSolution` on the flow and `GetPrimitive` on the turb variables?

pcarruscag:
I would add that layer to the solver so that the interpolation does not have to bypass it by going straight to the nodes.

maxaehle:
Ok thanks Pedro!

PR #1346, opened 2021-08-04
Fix the neighbor-finding in `CInterpolator::ReconstructBoundary`

maxaehle:
## Proposed Changes
`CInterpolator::ReconstructBoundary` builds arrays `Buffer_Receive_StartLinkedNodes`, `_LinkedNodes`,... that store the connectivity of the marker, i.e. the list of neighbor vertices for each vertex.

In the previous version of the code, two vertices were considered "neighbors" if they were connected through an edge of a volume element ("volume neighbors"). We propose to consider vertices as "neighbors" only if they are connected through an edge of a surface element ("surface neighbors").

These two notions are equivalent for smooth surfaces and clearly surface neighbors are always volume neighbors as well. However, at surfaces with sharp edges, two vertices can be volume neighbors without being surface neighbors. This happens when the edge of the volume element connecting the vertices lies inside the domain.

`CInterpolator::ReconstructBoundary` is called from `CSlidingMesh::SetTransferCoeff`, which then hands the connectivity structure to `CSlidingMesh::Build_3D_surface_element`. The latter function forms a circular ordering of the neighbors of each vertex s: e.g. if there are four neighbors a,b,c,d of s with connectivity a-b,c-d,b-c,a-d then it performs an iteration over e.g. a-b-c-d. If we understand "neighbors" as "surface neighbors", the neighbors of any vertex s can be brought into a "closed circle" or "open chain" ordering where each of them is neighbor only to its direct predecessor and successor. This is not true for the volume neighbor relation however, and in this case the iteration might enter an infinite loop e-a-b-c-d-a-b-... (until the access to `element[ iElementNode ]` segfaults). That's approximately how we (@maxaehle,@mcmorelli,@BeckettZhou,@EduardoMolina) noticed the bug.
 
## Related Work
#1344 

## PR Checklist


Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
-#ifdef HAVE_MPI
-  int nProcessor = size, iRank;
-  unsigned long iTmp2, tmp_index, tmp_index_2;
-#endif
-
-  /*--- Copy coordinates and point to the auxiliar vector ---*/
+  // coordinates of all domain vertices on the marker
+  su2double *Buffer_Send_Coord           = new su2double     [ nLocalVertex * nDim ];
+  // global point IDs of all domain vertices on the marker
+  long *Buffer_Send_GlobalPoint = new long [ nVertex ];
```
pcarruscag, line 209: If you have some time can you convert these and others to `su2activematrix` `su2vector<long>` `su2activevector`?

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
-    Buffer_Send_nLinkedNodes[iVertex] = 0;
-    Aux_Send_Map[iVertex]             = nullptr;
-
     iPoint = geom->vertex[val_marker][iVertex]->GetNode();
-
     if (geom->nodes->GetDomain(iPoint)) {
-      Buffer_Send_GlobalPoint[nLocalVertex] = geom->nodes->GetGlobalIndex(iPoint);
-
+      unsigned long iLocalVertex = iVertex_to_iLocalVertex[iVertex];
+      Buffer_Send_GlobalPoint[iLocalVertex] = (long) geom->nodes->GetGlobalIndex(iPoint);
```
pcarruscag, line 220: no C-style casts please, use `static_cast`

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
-      nNodes = 0;
-
-      for (jEdge = 0; jEdge < nEdges; jEdge++){
-        EdgeIndex = geom->nodes->GetEdge(iPoint, jEdge);
-
-        if( iPoint == geom->edges->GetNode(EdgeIndex,0) )
-          dPoint = geom->edges->GetNode(EdgeIndex,1);
-        else
-          dPoint = geom->edges->GetNode(EdgeIndex,0);
+        Buffer_Send_Coord[iLocalVertex*nDim+iDim] = geom->nodes->GetCoord(iPoint, iDim);
```
pcarruscag, line 222: this for example would be a lot cleaner with a matrix type.

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
+  su2double *Buffer_Send_Coord           = new su2double     [ nLocalVertex * nDim ];
+  // global point IDs of all domain vertices on the marker
+  long *Buffer_Send_GlobalPoint = new long [ nVertex ];
 
-  nGlobalVertex     = 0;
-  nLocalVertex      = 0;
-  nLocalLinkedNodes = 0;
+  // Assign to each domain vertex on the marker, identified by local point ID,
+  // a set of surface-neighbor vertices on the marker, identified by global point ID.
+  map<unsigned long, forward_list<unsigned long>*> neighbors;
```
pcarruscag, line 213: This looks like overkill...

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
+    CPrimalGrid* elem = geom->bound[val_marker][iElem];
+    for(unsigned short iNode=0; iNode<elem->GetnNodes(); iNode++){
+      iPoint = elem->GetNode(iNode);
+      if (geom->nodes->GetDomain(iPoint)) {
+        forward_list<unsigned long>* neighb = neighbors.at(iPoint);
+        for(unsigned short iNeighbor=0; iNeighbor<elem->GetnNeighbor_Nodes(iNode); iNeighbor++){
+          unsigned long jPoint = elem->GetNode( elem->GetNeighbor_Nodes(iNode,iNeighbor) );
+          unsigned long jPoint_global = geom->nodes->GetGlobalIndex(jPoint);
+          if( std::find(std::begin(*neighb), std::end(*neighb), jPoint_global) == std::end(*neighb) ){
+            neighb->emplace_front( jPoint_global );
```
pcarruscag, line 238: I think what you want is an unordered_map of unordered_sets...
Linear searches over a list are reallyyyyy expensive, ask me how I know.
Also if you always `emplace_front` you would want a stack, which is a vector upside down.

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,110 +171,117 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
-    for (jEdge = 0; jEdge < Buffer_Send_nLinkedNodes[iVertex]; jEdge++){
-      Buffer_Send_LinkedNodes[nLocalLinkedNodes] = Aux_Send_Map[iVertex][jEdge];
-      nLocalLinkedNodes++;
+  for (iVertex = 0; iVertex < nVertex; iVertex++) {
+    iPoint = geom->vertex[val_marker][iVertex]->GetNode();
+    if (geom->nodes->GetDomain(iPoint)) {
+      unsigned long iLocalVertex = iVertex_to_iLocalVertex[iVertex];
+      Buffer_Send_nLinkedNodes[iLocalVertex] = std::count_if(
+        std::begin(*neighbors[iPoint]), std::end(*neighbors[iPoint]),
+        [](unsigned long i){return true;} );
```
pcarruscag, line 256: And what are we doing here? Counting the elements in a list? Doesn't look very expressive :confused:

pcarruscag (commented):
Thank you for looking into this, these interpolation functions can get quite complicated.
I think there is some room to simplify the new algorithm a bit, and to improve the old code if you have some time.
Just before merging this PR please run clang-format on CInterpolator.cpp.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: I started with `iVertex_to_iLocalVertex` (332ca23, easy) and `Buffer_Send_Coord`,`Buffer_Receive_Coord` (30ac1aa, this was difficult).

Common/src/interface_interpolation/CNearestNeighbor.cpp
```diff
@@ -169,10 +169,7 @@ void CNearestNeighbor::SetTransferCoeff(const CConfig* const* config) {
     }
     END_SU2_OMP_PARALLEL
 
-    delete[] Buffer_Send_Coord;
     delete[] Buffer_Send_GlobalPoint;
-
-    delete[] Buffer_Receive_Coord;
```
maxaehle, line 175: The 1:1 replacement of `delete[] Buffer_Send_Coord` would probably be `Buffer_Send_Coord.resize(0,0)`. However this function seems to be called repeatedly in the `channel_2D` testcase, so let's keep the buffers.
In `CIsoparametric.cpp`, `CRadialBasisFunction.cpp` such `delete[]`s do also occur, and I also left them away although I don't know whether the buffers are used repeatedly.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: ok, changed in 19d7a95

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -171,126 +170,135 @@ void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
...
-      nLocalVertex++;
     }
   }
 
-  unsigned long *Buffer_Send_LinkedNodes = new unsigned long [ nLocalLinkedNodes ];
-
+  // numbers of surface-neighbors of all domain vertices on the marker
+  unsigned long *Buffer_Send_nLinkedNodes       = new unsigned long [ nLocalVertex ];
+  // cumsum of Buffer_Send_nLinkedNodes
+  unsigned long *Buffer_Send_StartLinkedNodes   = new unsigned long [ nLocalVertex ];
```
pcarruscag, line 247: Ok great! I meant mostly the local arrays on the function you are working on, but thanks for taking it further.

Common/include/interface_interpolation/CSlidingMesh.hpp
```diff
@@ -64,7 +64,7 @@ class CSlidingMesh final : public CInterpolator {
    * \return Number of points included in the element.
    */
   static int Build_3D_surface_element(const unsigned long *map, const unsigned long *startIndex,
-                                      const unsigned long* nNeighbor, const su2double *coord,
+                                      const unsigned long* nNeighbor, su2activematrix const& coord,
```
pcarruscag, line 67:
```suggestion
                                      const unsigned long* nNeighbor, const su2activematrix& coord,
```
I know you like this ordering but this is not how most of the code is written... And it will probably be changed if you run clang-format.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: I changed the type of `neighbors` from `map<unsigned long, forward_list<unsigned long>*>` to `unordered_map<unsigned long, unordered_set<unsigned long> >` and `emplace_front` by `insert`, in commit 480d9a3.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: You're right, this is fixed in 480d9a3 also

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -87,7 +87,7 @@ void CInterpolator::Collect_VertexInfo(int markDonor, int markTarget,
 
   for (iVertex = 0; iVertex < MaxLocalVertex_Donor; iVertex++) Buffer_Send_GlobalPoint[iVertex] = -1;
 
-  for (iVertex = 0; iVertex < MaxLocalVertex_Donor*nDim; iVertex++) Buffer_Send_Coord[iVertex] = 0.0;
+  Buffer_Send_Coord.resize(nVertexDonor,nDim);
```
pcarruscag, line 91: If zero-init is needed you can do 
```suggestion
  Buffer_Send_Coord.resize(nVertexDonor,nDim) = su2double(0.0);
```

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -97,15 +97,14 @@ void CInterpolator::Collect_VertexInfo(int markDonor, int markTarget,
     if (donor_geometry->nodes->GetDomain(iPointDonor)) {
       Buffer_Send_GlobalPoint[iLocalVertexDonor] = donor_geometry->nodes->GetGlobalIndex(iPointDonor);
       for (iDim = 0; iDim < nDim; iDim++)
-        Buffer_Send_Coord[iLocalVertexDonor*nDim+iDim] = donor_geometry->nodes->GetCoord(iPointDonor, iDim);
+        Buffer_Send_Coord[iLocalVertexDonor][iDim] = donor_geometry->nodes->GetCoord(iPointDonor, iDim);
```
pcarruscag, line 101:
```suggestion
        Buffer_Send_Coord(iLocalVertexDonor, iDim) = donor_geometry->nodes->GetCoord(iPointDonor, iDim);
```
The preferred syntax for su2matrix is (i, j) because this will check the range of each index in a debug build.
Operator [] returns a pointer, because we had to keep things compatible with numerics and so on.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: Thanks for the hint! Zero-initialization is actually not needed here, we just write the coordinates into the buffer afterwards.

Common/src/interface_interpolation/CNearestNeighbor.cpp
pcarruscag: Or you assign an empty matrix to it, which is more STL style since STL containers usually don't "resize down"
```
Buffer_Send_Coord = su2activematrix();
```

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: ok, changed in 7cc51eb

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: in 612be75 I changed the type to `unsigned long` anyway

pcarruscag (approved):
Thanks, run clang-format on CInterpolator.cpp before merging please, it's a bit messy with unaligned comments, etc.

maxaehle:
The previous commit 4eeb6f9 replaces `std::unordered_set` by `std::set` for the collection of neighbors of a vertex. 

For an `unordered_set`, it is not specified in which order the elements show up in the iteration over all neighbors. Through floating-point arithmetics, the order probably influences the residuals in the regression tests; I think few testcases failed because a different order was used. With `unordered_set`, testcases would fail already if another implementation of the standard library is used.

In contrast, `set` is sorted by `std::less` by default.

`unordered_set` is also used in other places in the code, where it is ok:
- `Point_Map` in `CPhysicalGeometry::DistributeColoring`: only used to sort into `Global2Local`
- `processedMarkers` in `CFlowTractionInterface::ComputeVertexAreas`: not iterated through
- `markerPoints` in `CFEASolver::Set_VertexEliminationSchedule`: not iterated through

maxaehle:
Only the testcases `channel_2D`, `channel_3D` fail. Both use sliding mesh interfaces, but there are also other sliding mesh testcases that do not fail. The residuals are still very close and also the flow fields look similar. I thus adapted the stored residuals in 9d8f194. Let me know if you would like to investigate the changed residuals further.

pcarruscag:
No need to dig anymore. And nevermind the failing hybrid regression (it looks like those tests were waiting to be merged to become flaky 😠 )

maxaehle:
That's really strange: The hybrid AD testcase `discadj_incomp_turb_NACA0012_sst` was ok after commit c68cf13, then I changed the stored residuals of unrelated testcases in 9d8f194 and the testcase fails... I'll run clang-format now, let's see what happens ;)

maxaehle:
I restarted the CI pipeline for 6d9e103. In the first run, there was an additional error that mesh files were missing (??). Now in the second run, the regression tests pass...

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,253 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
-
-  for (iVertex = 0; iVertex < nLocalVertex; iVertex++){
-    for (jEdge = 0; jEdge < Buffer_Send_nLinkedNodes[iVertex]; jEdge++){
-      Buffer_Send_LinkedNodes[nLocalLinkedNodes] = Aux_Send_Map[iVertex][jEdge];
-      nLocalLinkedNodes++;
+  for (iVertex = 0; iVertex < nVertex; iVertex++) {
+    iPoint = geom->vertex[val_marker][iVertex]->GetNode();
+    if (geom->nodes->GetDomain(iPoint)) {
+      unsigned long iLocalVertex = iVertex_to_iLocalVertex[iVertex];
+      Buffer_Send_nLinkedNodes[iLocalVertex] = neighbors[iPoint].size();
```
pcarruscag, line 235: It's safer to use .at() with maps if you are not planning to modify.
Also eases debugging if an index is not there and should.

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,253 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
-    for (iVertex = 0; iVertex < nLocalVertex; iVertex++){
-      Buffer_Receive_GlobalPoint[iVertex]      = Buffer_Send_GlobalPoint[iVertex];
-      Buffer_Receive_Proc[iVertex]             = MASTER_NODE;
-      Buffer_Receive_nLinkedNodes[iVertex]     = Buffer_Send_nLinkedNodes[iVertex];
+  int nProcessor = size, iRank;
+  unsigned long received_nLocalLinkedNodes, received_nLocalVertex;
+
+  if (rank == MASTER_NODE) {
+    /*--- "Receive" from master process, i.e. copy. ---*/
+    for (unsigned long iVertex = 0; iVertex < nLocalVertex; iVertex++)
```
pcarruscag, line 277: To copy you can assign the matrix type directly which is more efficient than looping over each dimension

maxaehle: I want to copy the data of `Buffer_Send_Coord` into the beginning of `Buffer_Receive_Coord`. `operator=` would resize (shrink) the receive buffer container and resizing (making larger) afterwards might invalidate the content?

pcarruscag: Fair enough, I thought they were the same size all the time.

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: thanks, faf2b3b

TobiKattmann:
> 
> 
> That's really strange: The hybrid AD testcase `discadj_incomp_turb_NACA0012_sst` was ok after commit [c68cf13](https://github.com/su2code/SU2/commit/c68cf13552bc1412b98bda281a92f0b4bd4799ba), then I changed the stored residuals of unrelated testcases in [9d8f194](https://github.com/su2code/SU2/commit/9d8f194eeb4523440b51b3ff4c08018492974198) and the testcase fails... I'll run clang-format now, let's see what happens ;)

I had a similar thing over at the flamelet branch and a re-run of the regression tests solved that :)

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,253 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
 
-        if( iPoint == geom->edges->GetNode(EdgeIndex,0) )
-          dPoint = geom->edges->GetNode(EdgeIndex,1);
-        else
-          dPoint = geom->edges->GetNode(EdgeIndex,0);
+  /*--- Define or initialize them. ---*/
+  for (iVertex = 0; iVertex < nVertex; iVertex++) {
+    iPoint = geom->vertex[val_marker][iVertex]->GetNode();
+    if (geom->nodes->GetDomain(iPoint)) {
+      unsigned long iLocalVertex = iVertex_to_iLocalVertex[iVertex];
```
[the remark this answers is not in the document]
TobiKattmann: not sure if the const for the iPoint is possible but that would be prefered. 

This approach would be appreciated in the rest the blocks as well not just here

TobiKattmann (approved):
Thanks @maxaehle 💐 ... just a little nagging in a few places.

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,253 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
 }
 
-void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker){
-
-  CGeometry *geom = Geometry[val_zone][INST_0][MESH_0];
+void CInterpolator::ReconstructBoundary(unsigned long val_zone, int val_marker) {
+  CGeometry* geom = Geometry[val_zone][INST_0][MESH_0];
 
-  unsigned long iVertex, jVertex, kVertex;
+  unsigned long iVertex, kVertex;
```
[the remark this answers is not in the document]
TobiKattmann: Here i mean of course not the change you did, but keeping i,kVertex

Common/src/interface_interpolation/CInterpolator.cpp
TobiKattmann: @maxaehle please dont forget to ask pedro `how he knows`. I would be interested in that

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -52,18 +52,15 @@ bool CInterpolator::CheckInterfaceBoundary(int markDonor, int markTarget) {
 }
 
 bool CInterpolator::CheckZonesInterface(const CConfig* donor, const CConfig* target) {
-
   /*--- Loop over all interface markers to find if the 2 zones share any interface boundary. ---*/
-  for (auto iInter = 0; iInter < (donor->GetMarker_n_ZoneInterface()/2); iInter++) {
-    if (CheckInterfaceBoundary(donor->FindInterfaceMarker(iInter), target->FindInterfaceMarker(iInter)))
-      return true;
+  for (auto iInter = 0; iInter < (donor->GetMarker_n_ZoneInterface() / 2); iInter++) {
+    if (CheckInterfaceBoundary(donor->FindInterfaceMarker(iInter), target->FindInterfaceMarker(iInter))) return true;
```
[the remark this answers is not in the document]
pcarruscag: It will do that, yes.

Common/src/interface_interpolation/CSlidingMesh.cpp
```diff
@@ -286,10 +291,10 @@ void CSlidingMesh::SetTransferCoeff(const CConfig* const* config) {
...
 
             for(iDim = 0; iDim < nDim; iDim++){
-              donor_iMidEdge_point[iDim] = ( DonorPoint_Coord[ donor_forward_point  * nDim + iDim] +
-                                             DonorPoint_Coord[ donor_iPoint * nDim + iDim] ) / 2;
-              donor_jMidEdge_point[iDim] = ( DonorPoint_Coord[ donor_backward_point * nDim + iDim] +
-                                             DonorPoint_Coord[ donor_iPoint * nDim + iDim] ) / 2;
+              donor_iMidEdge_point[iDim] = ( DonorPoint_Coord[ donor_forward_point ][ iDim] +
+                                             DonorPoint_Coord[ donor_iPoint ][ iDim] ) / 2;
+              donor_jMidEdge_point[iDim] = ( DonorPoint_Coord[ donor_backward_point ][ iDim] +
+                                             DonorPoint_Coord[ donor_iPoint ][ iDim] ) / 2;
```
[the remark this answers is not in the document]
pcarruscag: The right thing there would be operator ( , ) because Max replaced pointers with containers.
If we just want to clean the spaces then run clang-format and call it a day.

Common/src/interface_interpolation/CInterpolator.cpp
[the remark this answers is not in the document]
pcarruscag: Those are either on the stack or probably don't even leave the registers.

pcarruscag:
Hybrid AD tests are a bit flaky ATM :shrug: you can merge things without those tests passing, but if more than 1 fails let me know.

Common/src/interface_interpolation/CSlidingMesh.cpp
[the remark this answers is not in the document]
maxaehle: I forgot to replace by `operator(i,j)` here, fixed in 475c6971

Common/src/interface_interpolation/CInterpolator.cpp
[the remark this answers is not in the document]
maxaehle: I replaced `unsigned long` by `const auto` at some locations

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,249 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
+      neighbors.insert(pair<unsigned long, set<unsigned long> >(iPoint, set<unsigned long>()));
+    }
+  }
 
-        if ( geom->nodes->GetVertex(dPoint, val_marker) != -1 ){
-          Aux_Send_Map[nLocalVertex][nNodes] = geom->nodes->GetGlobalIndex(dPoint);
-          nNodes++;
+  /*--- Define the neighbors map. ---*/
+  for (unsigned long iElem = 0; iElem < nElems; iElem++) {
+    CPrimalGrid* elem = geom->bound[val_marker][iElem];
```
maxaehle, line 207: This cannot be made `const CPrimalGrid* elem` at the moment because `CPrimalGrid::GetNode` etc are not marked as `const`.

Common/src/interface_interpolation/CInterpolator.cpp
```diff
@@ -143,253 +130,253 @@ unsigned long CInterpolator::Collect_ElementInfo(int markDonor, unsigned short n
...
+  // Assign to each domain vertex on the marker, identified by local point ID,
+  // a set of surface-neighbor vertices on the marker, identified by global point ID.
+  unordered_map<unsigned long, set<unsigned long> > neighbors;
 
-        if( iPoint == geom->edges->GetNode(EdgeIndex,0) )
-          dPoint = geom->edges->GetNode(EdgeIndex,1);
-        else
-          dPoint = geom->edges->GetNode(EdgeIndex,0);
+  /*--- Define or initialize them. ---*/
+  for (iVertex = 0; iVertex < nVertex; iVertex++) {
```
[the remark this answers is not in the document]
maxaehle: Changed in da41cdb for this particular commit. I used `//` also to document local variables like this:

    // coordinates of all domain vertices on the marker
    su2activematrix Buffer_Send_Coord(nLocalVertex, nDim);
    // global point IDs of all domain vertices on the marker
    su2vector<unsigned long> Buffer_Send_GlobalPoint(nVertex);
which is ok, right?

Common/src/interface_interpolation/CInterpolator.cpp
TobiKattmann: I guess/assume you could make the various `GetNode` functions const :) ... and then `const auto elem` ... but this goes a little beyond PR scope so up to you

Common/src/interface_interpolation/CSlidingMesh.cpp
```diff
@@ -245,10 +245,10 @@ void CSlidingMesh::SetTransferCoeff(const CConfig* const* config) {
 
           dTMP = 0;
           for(iDim = 0; iDim < nDim; iDim++){
-            target_iMidEdge_point[iDim] = ( TargetPoint_Coord[ target_segment[0] ][ iDim ] +
-                                            target_geometry->nodes->GetCoord( target_iPoint , iDim) ) / 2;
-            target_jMidEdge_point[iDim] = ( TargetPoint_Coord[ target_segment[1] ][ iDim ] +
-                                            target_geometry->nodes->GetCoord( target_iPoint , iDim) ) / 2;
+            target_iMidEdge_point[iDim] = ( TargetPoint_Coord(target_segment[0], iDim ) +
+                                            target_geometry->nodes->GetCoord( target_iPoint , iDim) ) / 2.;
```
TobiKattmann, line 249: I am curious: Did you get a compiler notification or so on this `2.` or was it just personal preference?

Common/src/interface_interpolation/CInterpolator.cpp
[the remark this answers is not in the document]
maxaehle: I completely agree with @TobiKattmann that these big declaration blocks in the beginning of a function are not very meaningful to the reader. As according to @pcarruscag there is no performance concern, I improved this in `CInterpolator::ReconstructBoundary` and  `CSlidingMesh::Build_3D_surface_element`.

Common/src/interface_interpolation/CSlidingMesh.cpp
maxaehle: Personal preference. Once I spent a lot of time finding out that 1/2==0.

maxaehle:
> > That's really strange: The hybrid AD testcase `discadj_incomp_turb_NACA0012_sst` was ok after commit [c68cf13](https://github.com/su2code/SU2/commit/c68cf13552bc1412b98bda281a92f0b4bd4799ba), then I changed the stored residuals of unrelated testcases in [9d8f194](https://github.com/su2code/SU2/commit/9d8f194eeb4523440b51b3ff4c08018492974198) and the testcase fails... I'll run clang-format now, let's see what happens ;)
> 
> I had a similar thing over at the flamelet branch and a re-run of the regression tests solved that :)

Seems to be like that here as well:
After e95cdfa, only `transonic_stator` in `hybrid_regression_AD.py` failed.
After 475c697, only `discadj_incomp_turb_NACA0012_sa` in `hybrid_regression_AD.py` failed.
After b704f9e no test in `hybrid_regression_AD.py` failed.

Common/include/interface_interpolation/CInterpolator.hpp
```diff
@@ -28,7 +28,10 @@
 
 #include "../../include/basic_types/datatype_structure.hpp"
 #include "../../include/containers/C2DContainer.hpp"
+#include "../../include/containers/container_decorators.hpp"
 #include <vector>
+#include <forward_list>
```
pcarruscag, line 33:
```suggestion
```

maxaehle:
After 24d81f5, only `ea_naca64206` in `parallel_regression_AD.py` failed

Common/src/interface_interpolation/CInterpolator.cpp
maxaehle: #1352

PR #1351, opened 2021-08-15
Delete dead-code for "nearfield" and "interface" boundaries

pcarruscag:
I think nearfield was a kind of internal interface between two boundaries with the same points, so not a "do nothing" group of points but more like an actuator disk without any jump.
Anyway, it was all incomplete or broken and there was no documentation in the code, for a change...
To anyone in the future the commit to revert is 265bfe8814ed345e9ab6288265e80858b00ae59e


SU2_CFD/src/solvers/CAdjEulerSolver.cpp
```diff
@@ -1183,311 +898,6 @@ void CAdjEulerSolver::SetForceProj_Vector(CGeometry *geometry, CSolver **solver_
...
-              }
-
-              if (AngleInt <= 60) {
-                iColumn = IndexNF_inv[AngleInt];
-
-                /*--- An azimuthal angle is not defined... this happens with MG levels ---*/
-
-                if (iColumn < 0.0) {
-                  if (IndexNF_inv[AngleInt+1] > 0) { iColumn = IndexNF_inv[AngleInt+1]; goto end; }
-                  if (IndexNF_inv[AngleInt-1] > 0) { iColumn = IndexNF_inv[AngleInt-1]; goto end; }
```
pcarruscag, line 1329: I was almost surprised :smile:

pcarruscag:
@snow54 I think the solution file for your testcase will have to be updated, can you have a look please to make sure the discrete adjoint still works?
This should also fix the problems you had with the Venkatakrishnan limiter.

SU2_CFD/src/solvers/CAdjEulerSolver.cpp
TobiKattmann: lol, that code looks like someone really hated `else if`

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -640,10 +640,9 @@ void CDriver::Geometrical_Preprocessing(CConfig* config, CGeometry **&geometry,
 
   geometry[MESH_0]->SetPositive_ZArea(config);
 
-  /*--- Set the near-field, interface and actuator disk boundary conditions, if necessary. ---*/
+  /*--- Set the and actuator disk boundary conditions, if necessary. ---*/
```
TobiKattmann, line 643:
```suggestion
  /*--- Set the actuator disk boundary conditions, if necessary. ---*/
```

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -4347,7 +4309,7 @@ class CSolver {
    * \return Struct holding 4 su2doubles.
    */
   virtual StreamwisePeriodicValues GetStreamwisePeriodicValues() const { return StreamwisePeriodicValues(); }
-  
+
```
TobiKattmann, line 4312: 'It wasn't me' - Shaggy

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1253,332 +1253,6 @@ void CTurbSASolver::BC_Inlet_Turbo(CGeometry *geometry, CSolver **solver_contain
 
 }
 
-void CTurbSASolver::BC_Interface_Boundary(CGeometry *geometry, CSolver **solver_container, CNumerics *numerics,
-                                          CConfig *config, unsigned short val_marker) {
-
-  //  unsigned long iVertex, iPoint, jPoint;
-  //  unsigned short iVar, iDim;
```
TobiKattmann, line 1260: A pro-move I learned from a dear colleague in a company I wont name here: throw an error in the function, and if no-one complains for long enough, delete the code 👍 ... this full commenting of the function content is kinda the same

TobiKattmann (approved):
Well, you had me at `+23 −1,962` ... 💐 I do not have much to say other than thanks for this cleanup \(^~^)/

Once yuki  confirms that everything works this is good to go from my side

... And 'now' you complain about complex code! Get your things together CodeFactor. (*proceeds to hang off employee of the month Feb2020 picture)

snow54:
@pcarruscag I have made a [pull request](https://github.com/su2code/TestCases/pull/72) to update solution file for the test case.

You have deleted codes that create WeightNF.dat so I have deleted it from python codes as well. (I tried to push as a suggestion but it didn't work as I expected).

> I think nearfield was a kind of internal interface between two boundaries with the same points, so not a "do nothing" group of points but more like an actuator disk without any jump.

I thought so when I was trying to make it work in ver.6 but, when I tried with mesh with two boundaries with the same points, it didn't work.

pcarruscag:
Thank you for updating the solution file @snow54 , there's quite a big difference in the adjoint residuals, do the final derivatives still make sense? Is the flow field noticeably different (for the better, e.g. smoother? or equivalent to not having the boundary at all?) now that nearfield is treated as an internal boundary?
If so let's update the residuals and merge.

I'm sorry that you had to spend time fixing that MPI code... But at least we found out we could clean all this obsolete code.

snow54:
> Thank you for updating the solution file @snow54 , there's quite a big difference in the adjoint residuals, do the final derivatives still make sense? Is the flow field noticeably different (for the better, e.g. smoother? or equivalent to not having the boundary at all?) now that nearfield is treated as an internal boundary?
> If so let's update the residuals and merge.

I think we can update the residuals.

Gradients between adjoint and finite difference match quite well as shown below. I haven't finished calculating for all design variables, but I think it is enough.

Comparison_gradient

In terms of flow field, the capture below is from the current develop branch. The boundary between structured grid and unstructured gird is a nearfield boundary. An object is above this capture and pressure wave propagates from there. Since this grid is inclined by Mach angle, pressure distribution on the nearfield should be fairly similar but this capture shows some strange pressure disturbance.

Limiter: VAN_ALBADA_EDGE
NF_before

Limiter: VENKATAKRISHNAN_WANG
NF_before_VEN

The capture below is from the branch with this PR. The issue I mentioned above does not exist.

Limiter: VAN_ALBADA_EDGE
NF_after

I think the residuals for direct solver will be different like the adjoint if you run it for some hundreds more iterations (currently, the test case has only 20 iterations). However, since the nearfield boundary is a bit far from an object and it takes some iterations for pressure waves to reach the nearfield boundary.

This PR also solves an issue with VENKATAKRISHNAN_WANG limiter. It seems to be much easier for convergence than VAN_ALBADA_EDGE, so it is fairly useful. I still have a residual un-match issue with my bigger mesh but I believe it is coming from something else.

> I'm sorry that you had to spend time fixing that MPI code... But at least we found out we could clean all this obsolete code.

Yeah, I noticed it had been deleted. It's OK. It was still a good opportunity for me to learn how MPI works.

PR #1352, opened 2021-08-16
Refactor CPrimalGrid classes

maxaehle:
## Proposed Changes
- Qualify member functions `CPrimalGrid::Get...` as const
- `CPrimalGridFEM::GetCornerPointsAllFaces` did set the member `nFaces`. To make it const, I introduced `CPrimalGridFEM::elementtype_to_nFaces` and initialized `nFaces` in the constructor.
- Now we can use `const CPrimalGrid* elem` in `CInterpolator::ReconstructBoundary` (#1346).
- **EDIT**: see more changes below

## PR Checklist


Common/src/geometry/primal_grid/CPrimalGridFEM.cpp
```diff
@@ -99,6 +101,18 @@ CPrimalGridFEM::CPrimalGridFEM(unsigned long  val_elemGlobalID, unsigned short v
...
+    case TRIANGLE: return 3;
+    case QUADRILATERAL: return 4;
+    case TETRAHEDRON: return 4;
+    case PYRAMID: return 5;
+    case PRISM: return 5;
+    case HEXAHEDRON: return 6;
+    default: SU2_MPI::Error("Invalid elementType.", CURRENT_FUNCTION); return 0;
+  }
+}
+
```
pcarruscag, line 115:
```suggestion
```

Common/include/geometry/primal_grid/CPrimalGridFEM.hpp
```diff
@@ -213,8 +213,16 @@ class CPrimalGridFEM final: public CPrimalGrid {
...
                                unsigned short nPointsPerFace[],
-                               unsigned long  faceConn[6][4]) override;
+                               unsigned long  faceConn[6][4]) const override;
+private:
+  /*!
+   * \brief Get the number of faces of the element.
+   * \param[in] elementType - element type
+   * \return number of faces
+   */
+  static unsigned short elementtype_to_nFaces(unsigned short elementType);
```
pcarruscag, line 223:
```suggestion
  static inline unsigned short nFacesOfElementType(unsigned short elementType) {
    switch (elementType) {
      case TRIANGLE: return 3;
      case QUADRILATERAL: return 4;
      case TETRAHEDRON: return 4;
      case PYRAMID: return 5;
      case PRISM: return 5;
      case HEXAHEDRON: return 6;
      default: SU2_MPI::Error("Invalid elementType.", CURRENT_FUNCTION); return 0;
    }
  }
```
Use naming conventions that more or less follow what is done.

pcarruscag (approved):

WallyMaier (approved):
This LGTM!

Common/include/geometry/primal_grid/CPrimalGridFEM.hpp
```diff
@@ -213,8 +213,26 @@ class CPrimalGridFEM final: public CPrimalGrid {
...
+   * \return number of faces
+   */
+  static inline unsigned short nFacesOfElementType(unsigned short elementType) {
+    switch (elementType) {
+      case TRIANGLE: return 3;
+      case QUADRILATERAL: return 4;
+      case TETRAHEDRON: return 4;
+      case PYRAMID: return 5;
+      case PRISM: return 5;
+      case HEXAHEDRON: return 6;
```
TobiKattmann, line 230: Note1: in option_structure.hpp we have
```c++
const unsigned short N_POINTS_TRIANGLE = 3;
```
so we could do the same for N_FACES_*

Note2 off-topic: I already wanted to plug some enum class change here, but the VTK element types actually have some fixed numbers attached to them ... so one might even state that over there.

TobiKattmann (approved):
👍

Common/include/geometry/primal_grid/CPrimalGridFEM.hpp
maxaehle: In addition to note 1, the classes `CLine`, `CTriangle` etc have private static members `nFaces`, `nNodes` replicating the information another time. We could also make those public (and constexpr), and use them in `nFacesOfElementType` as well as to replace `N_POINTS_TRIANGLE` wherever possible. I'll try this if you don't object @TobiKattmann @WallyMaier @pcarruscag .

pcarruscag: The objective of those classes being virtual (well, any class) is that you don't need their declaration everywhere.
So you would move N_FACES_TRIANGLE to e.g. option_structure (or to primal grid) and include it in CTriangle.cpp, and not include CTriangle everywhere you need N_FACES_TRIANGLE.

That family of classes could use some work, it is taking the worst of two worlds in some cases.
We can talk about that if you're interested in changing some of the older parts of SU2 :smile:

TobiKattmann: Ok, so I guess the only reasonable approach is then for e.g. for `TRIANGLE` to add `const unsigned short N_FACES_TRIANGLE=3` to option structure.hpp and then find - replace all `3`s in the codebase 🤝

pcarruscag: :facepalm:

Common/include/option_structure.hpp
```diff
@@ -127,6 +127,13 @@ const unsigned short N_POINTS_HEXAHEDRON = 8;    /*!< \brief General output & CG
 const unsigned short N_POINTS_PYRAMID = 5;       /*!< \brief General output & CGNS defines. */
 const unsigned short N_POINTS_PRISM = 6;         /*!< \brief General output & CGNS defines. */
 enum: unsigned short{N_POINTS_MAXIMUM = 8};      /*!< \brief Max. out of the above, used for static arrays, keep it up to date. */
+const unsigned short N_FACES_TRIANGLE = 3;       /*!< \brief General output & CGNS defines. */
+const unsigned short N_FACES_QUADRILATERAL = 4;  /*!< \brief General output & CGNS defines. */
+const unsigned short N_FACES_TETRAHEDRON = 4;    /*!< \brief General output & CGNS defines. */
+const unsigned short N_FACES_PYRAMID = 5;        /*!< \brief General output & CGNS defines. */
+const unsigned short N_FACES_PRISM = 5;          /*!< \brief General output & CGNS defines. */
+const unsigned short N_FACES_HEXAHEDRON = 6;      /*!< \brief General output & CGNS defines. */
```
TobiKattmann, line 136:
```suggestion
const unsigned short N_FACES_PYRAMID = 5;        /*!< \brief General output & CGNS defines. */
const unsigned short N_FACES_PRISM = 5;          /*!< \brief General output & CGNS defines. */
const unsigned short N_FACES_HEXAHEDRON = 6;     /*!< \brief General output & CGNS defines. */
```

TobiKattmann: 🔬

Common/include/geometry/primal_grid/CPrimalGridFEM.hpp
maxaehle: I added `N_FACES_*` and use the constants in option_structure.hpp in `nFacesOfElementType` and for the static members `nNodes`/`nFaces` of CLine etc. The numbers are hard-coded at many other locations, both in CLine etc (e.g. dimensions of `Faces`, `Neighbor_Nodes`,...) and outside. 
I am interested in how these classes could be improved  (I'll ask in some developer meeting, though maybe I cannot attend next week), but regarding this PR I'd like to finish it here ;)

pcarruscag: :+1: Regarding finishing PR's, I think it's fair to assume that if someone approves you don't have to go over their wishlist (unless they express the approval is conditional on that).

Ok, we can also talk via slack if you want. Or open a discussion, although it would take me longer to explain via text than to code the small changes I'm thinking :smile:

maxaehle:
I had a chat with @pcarruscag in the meantime. Summary of the recent changes:
- All the non-FEM classes derived from CPrimalGrid (CLine, CTriangle etc) declared the static members for the connectivity themselves. So they all had to override the virtual getter functions for the connectivity themselves, resulting in a lot of code duplication. -> Now we have an intermediate class `CPrimalGridWithConnectivity<Connectivity>` that overrides them once, accessing the static members of `Connectivity` = `CLineConnectivity`, `CTriangleConnectivity` etc.
- `type*` array -> `std::vector<type>`. As a special case for a fixed-size boolean array, `std::bitset`.
- removed unused variables and functions

Common/src/geometry/primal_grid/CHexahedron.cpp
```diff
@@ -26,47 +26,32 @@
...
-  unsigned short iNeighbor_Elements;
-
-  nDim = 3;
-
+             unsigned long val_point_6, unsigned long val_point_7):
+  CPrimalGridWithConnectivity<CHexahedronConnectivity>(false)
+{
   /*--- Allocate and define face structure of the element ---*/
-  Nodes = new unsigned long[nNodes];
+  Nodes.resize(GetnNodes());
```
maxaehle, line 47: The member function `GetNodes()` 
- is declared virtual in `CPrimalGrid`, 
- has a final override in the derived class `CPrimalGridWithConnectivity<CHexahedronConnectivity>`, and 
- is used in the constructor of the derived class `CHexahedron`.

I know that virtual functions cannot be used in the constructor as freely as afterwards, because base classes are constructed before derived classes. However in this case, I think it should be fine because the virtual version of `GetnNodes`, that we want to use in the constructor, is defined in a base class (and moreover, is marked final there).

pcarruscag: It's fine, you can call the virtual members of parent classes, if they are final there is no chance of errors.

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -45,50 +46,60 @@
  */
 class CPrimalGrid {
 protected:
-  unsigned long *Nodes;         /*!< \brief Vector to store the global nodes of an element. */
+  vector<unsigned long> Nodes;         /*!< \brief Global node indices of the element. */
   unsigned long GlobalIndex;    /*!< \brief The global index of an element. */
-  long *Neighbor_Elements;      /*!< \brief Vector to store the elements surronding an element. */
-  short *PeriodIndexNeighbors;  /*!< \brief Vector to store the periodic index of a neighbor.
+  vector<long> Neighbor_Elements;      /*!< \brief Vector to store the elements surronding an element. */
+  vector<short> PeriodIndexNeighbors;  /*!< \brief Vector to store the periodic index of a neighbor.
```
pcarruscag, line 52: This is not optimum with respect to size because std::vector stores 3 pointers to track capacity, however here we do not need "push_back" and so we are wasting 3 pointers.
Also, the "size" of Neighbor_Elements and PeriodIndexNeighbors is the same.

The most efficient way to store PeriodIndexNeighbors is as a static array of N_FACES_MAXIMUM since that will only take 12 bytes (then move it down in the member list). Possibly this could use even less space @vdweide ? (this doc string doesn't add anything to the name of the var...)

For Nodes and Neighbor_Elements use a unique_ptr.

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -45,50 +46,60 @@
  */
 class CPrimalGrid {
 protected:
-  unsigned long *Nodes;         /*!< \brief Vector to store the global nodes of an element. */
+  vector<unsigned long> Nodes;         /*!< \brief Global node indices of the element. */
   unsigned long GlobalIndex;    /*!< \brief The global index of an element. */
```
pcarruscag, line 50: Boundary elements do not have a global index, but they have a DomainElement and vice versa.
So we can (if c++ allows) do:
```
union {
  unsigned long GlobalIndex;
  unsigned long DomainElement;
};
```

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -451,3 +434,65 @@ class CPrimalGrid {
...
+  inline unsigned short GetnNodesFace(unsigned short val_face) const final {
+    return Connectivity::nNodesFace[val_face];
+  }
+
+  inline unsigned short GetMaxNodesFace() const final {
+    return Connectivity::maxNodesFace;
+  }
+
+  inline unsigned short GetFaces(unsigned short val_face, unsigned short val_index) const final {
+    return Connectivity::Faces[val_face][val_index];
```
maxaehle, line 482:
```suggestion
    assert(val_face < GetnFaces() && val_index < GetnNodesFace(val_face));
    return Connectivity::Faces[val_face][val_index];
```

Common/include/geometry/primal_grid/CPrimalGrid.hpp
```diff
@@ -451,3 +434,65 @@ class CPrimalGrid {
...
+  inline unsigned short GetnNodesFace(unsigned short val_face) const final {
+    return Connectivity::nNodesFace[val_face];
+  }
+
+  inline unsigned short GetMaxNodesFace(void) const final {
+    return Connectivity::maxNodesFace;
+  }
+
+  inline unsigned short GetFaces(unsigned short val_face, unsigned short val_index) const final {
+    return Connectivity::Faces[val_face][val_index];
```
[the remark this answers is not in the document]
maxaehle: I inserted a stricter condition after the `&&`, `val_index < GetnNodesFace(val_face)`

PR #1357, opened 2021-08-22
Updates of SU2-python

Nicola-Fonzi:
## Proposed Changes
Dear all,

Here are some updates for the python fsi interface.

Namely:

- Better use of mpi4py, using always numpy array (more computationally efficient)
- Fix of the restarted coupled simulation
- Automatic use of default values, in case of wrong inputs
- Introduction of a sort of dry run, which actually prints out the mapped modes and the normals of the aerodynamic surface mesh. This is yet another step towards the automatic creation of ROM from aerodynamic response
- Fix of small bugs

## PR Checklist


SU2_PY/FSI_tools/FSIInterface.py
```diff
@@ -2128,3 +2127,74 @@ def SteadyFSI(self, FSI_config,FluidSolver, SolidSolver):
...
+          if myid == self.rootProcess:  # The root process contains the solid solver for sure
+              modesNumber = np.array(int(SolidSolver.getNumberOfModes()))
+          else:
+              modesNumber = np.empty(1, dtype=np.int)
+
+          self.comm.Bcast(modesNumber, root=self.rootProcess)
+
+          for mode in range(np.asscalar(modesNumber)):
+              self.MPIPrint("Setting mode {} active".format(mode))
+              if myid in self.solidSolverProcessors:
```
pcarruscag, line 2178: and in this block

Nicola-Fonzi:
Dear @pcarruscag, thank you very much for the review. I implemented the modification you suggested and corrected the indentation (sorry about that...)

pcarruscag (approved):
Thanks @Nicola-Fonzi :+1:

SU2_PY/SU2_Nastran/pysu2_nastran.py
```diff
@@ -943,24 +984,27 @@ def applyload(self, iVertex, fx, fy, fz):
...
-    L = list(self.markers)
-    return L[0]
+    return self.FSI_marker
 
   def getNumberOfSolidInterfaceNodes(self, markerID):
 
     return len(self.markers[markerID])
 
-  def getInterfaceNodeGlobalIndex(self, markerID, iVertex):
+  def getVertexGlobalIndex(self, markerID, iVertex):   # TODO This solver is serial, thus global=local
```
TobiKattmann, line 1007: Are the TODO on your list to be tackled next? Is parallel use presented in those setups? I would rather error-check and document the state then to write a TODO in the code

SU2_PY/SU2_Nastran/pysu2_nastran.py
```diff
@@ -774,12 +855,14 @@ def __temporalIteration(self,time):
       self.a += (1-self.alpha_f)/(1-self.alpha_m)*self.qddot
     else:
       if self.ImposedMotionToSet:
+        if self.Config["RESTART_SOL"] == "NO":                                   # If yes we already set it in the __setRestart function
+          self.timeStartCoupling = time
```
TobiKattmann, line 859:
```suggestion
        if self.Config["RESTART_SOL"] == "NO":
          # If yes we already set it in the __setRestart function
          self.timeStartCoupling = time

```

SU2_PY/FSI_tools/FSIInterface.py
```diff
@@ -1247,16 +1239,16 @@ def interpolateSolidPositionOnFluidMesh(self, FSI_config):
           if myid == self.rootProcess:
             for iProc in self.fluidInterfaceProcessors:
               sendBuff = {}
-              for key in self.FluidHaloNodeList[iProc].keys():
-                globalIndex = self.fluidIndexing[key]
+              for key in self.FluidHaloNodeList[iProc].keys():                  # The key is the SU2 global
+                globalIndex = self.fluidIndexing[key]                           # This is the interface global, not the SU2 global
```
TobiKattmann, line 1243: Not that I would understand what is going on but `# The key is the SU2 global` sounds like there is sth missing.

TobiKattmann (commented):
Thanks @Nicola-Fonzi 💐 of course I cannot say anything content-wise but here are a few little spots I can nag about

SU2_PY/SU2_Nastran/pysu2_nastran.py
TobiKattmann: I am not sure whether a comment 20 spaces away is the best thing todo, I would rather have it close to its partner lines

SU2_PY/SU2_Nastran/pysu2_nastran.py
Nicola-Fonzi: Hi Tobi! Thank you very much for your time. Yes, you are right... TODOs in the code are ugly. They are definitely being tackled right now, this is why they are there, but I can remove them and just keep a note.

SU2_PY/FSI_tools/FSIInterface.py
Nicola-Fonzi: you mean that it should be "the keys are..."?

SU2_PY/SU2_Nastran/pysu2_nastran.py
Nicola-Fonzi: agree

SU2_PY/FSI_tools/FSIInterface.py
TobiKattmann: Hm, no I think more sth like 
```python
# The key is the SU2 global >node ID<
```
in case that makes any sense... That is the whole point :) I dont know what `SU2 global` alone should mean. I just had the feeling there was a noun missing (like e.g. `node ID`), not that I know it is missing. 

Treat it more like a question: Is this sentence correct? If yes, then leave it as is 👍

Nicola-Fonzi: I see what you mean. No the comment was actually saying that each key in the dictionary consists of the global ID of an interface node. However, from your question, I feel it was not really clear... I try to clarify

Nicola-Fonzi: Please have a look at #1359

PR #1359, opened 2021-08-24
Update SU2-python 2.0

Nicola-Fonzi:
## Proposed Changes
Just small changes suggested by @TobiKattmann (Thank you) in #1357

## PR Checklist

### DIFF
--- a/SU2_PY/FSI_tools/FSIInterface.py
+++ b/SU2_PY/FSI_tools/FSIInterface.py
@@ -1239,7 +1239,7 @@ class Interface:
           if myid == self.rootProcess:
             for iProc in self.fluidInterfaceProcessors:
               sendBuff = {}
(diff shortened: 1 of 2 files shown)


TobiKattmann (approved):
Thanks @Nicola-Fonzi 💐  

As my review was too late, it would have been totally fine to just do a local commit and share (and reference it) with the next PR 👍 

Just one more question: Can the features that should be run only in serial mode, be run in parallel by accident without throwing some error like `The combination ... only allows for serial execution of the code.`. If not I would appreciate those kinds of checks that prevent use out of spec.

Nicola-Fonzi:
I definitely closed #1357 too fast, so no problem. I tried to implement all your suggestions. Regarding the possibility of running by error in parallel the structural solver, this is not possible. That part of the code is meant to solve a small system of structural equations, so it is only implemented in serial mode.

Issue #1345, opened 2021-08-03
a possible bug for loading grid points for a zone


dingxiwang:
**Describe the bug**
When loading grid points for the ith zone, it seems that grid points for all zones after the ith zone will be loaded as well.
This grid point loading is implemented in the function of void CSU2ASCIIMeshReaderFVM::ReadPointCoordinates() 

The code snippet below is doing the function. I think there should be a break for the while loop once the grid points for the required zone are loaded.

  /*--- Open the mesh file and jump to our zone. ---*/

  mesh_file.open(meshFilename, ios::in);
  FastForwardToMyZone();

  /*--- Read the point coordinates into our data structure. ---*/

  string text_line;
  string::size_type position;
  while (getline (mesh_file, text_line)) {

    position = text_line.find ("NPOIN=",0);
    if (position != string::npos) {

 - SU2 Version:7.1.1

It should be noted that this bug is not expected to cause wrong solutions, but lead to unnecessary extra time cost and memory consumption probably.

pcarruscag:
I think you are right.
You joined the dev meeting today as we were ending right? Maybe the European summertime misled you.
Can you open a pull request for this? There are other parts of that function that are not very efficient.

dingxiwang:
Aha, yes, Nitish from Delft University of Technology sent me the link and I opened it at that time.
I started looking into SU2 very recently and may join the development meeting next time.

PR #1361, opened 2021-08-24
Cleanup ASCII mesh reader

pcarruscag:
## Proposed Changes
Reduce duplication and maybe make it a bit faster.

We were reading the entire mesh file in four passes (open-close, open-close, ...) one for metadata (NDIME, NELEM, etc.) one for points, one for elements, and then one for markers.
Now we have a maximum of 2 (meta + reading) and if the points are listed before the elements we do it in 1 pass (there are some meshes like that in TestCases).
If we need to split actuator disk surfaces there are more passes, don't ask me how many because I'm not going to touch that function (it is probably not used often).

We still do a lot of reading for single-file multizone meshes to get to each zone, and usually we put smaller meshes at the end right :sweat_smile: I won't get into that now.

## Related Work
#1345 

## PR Checklist


Common/src/geometry/meshreader/CSU2ASCIIMeshReaderFVM.cpp
```diff
@@ -122,128 +129,140 @@ void CSU2ASCIIMeshReaderFVM::ReadMetadata() {
...
 
-    position = text_line.find ("NMARK=",0);
-    if (position != string::npos) {
+    if (!foundNMARK && text_line.find ("NMARK=",0) != string::npos) {
       text_line.erase (0,6);
       numberOfMarkers = atoi(text_line.c_str());
+
+      if (current_section_idx != 2) {
+        SU2_MPI::Error("Markers must be listed after points and elements in the SU2 mesh file.", CURRENT_FUNCTION);
+      }
```
pcarruscag, line 230: This PR introduces this restriction. It could be avoided if we stop checking for v6-style periodicity in the mesh file.

Common/src/toolboxes/CLinearPartitioner.cpp
```diff
@@ -80,12 +80,7 @@ unsigned long CLinearPartitioner::GetRankContainingIndex(unsigned long index) co
 
   /*--- Initial guess ---*/
 
-  unsigned long iProcessor = index/sizeOnRank[0];
-
-  /*--- Guard against going over size. ---*/
-
-  if (iProcessor >= (unsigned long)size)
-    iProcessor = (unsigned long)size-1;
+  unsigned long iProcessor = min<unsigned long>(index/sizeOnRank[0], size-1);
```
WallyMaier, line 83: 👍

Common/src/geometry/meshreader/CSU2ASCIIMeshReaderFVM.cpp
```diff
@@ -718,705 +740,304 @@ void CSU2ASCIIMeshReaderFVM::SplitActuatorDiskSurface() {
...
+      if (text_line.find("NMARK=",0) == string::npos) continue;
+    }
 
-              case TRIANGLE:
+    for (unsigned short iMarker = 0; iMarker < numberOfMarkers; ++iMarker) {
+      getline (mesh_file, text_line);
+      text_line.erase (0,11);
+      string::size_type position;
+
+      for (unsigned short iChar = 0; iChar < 20; iChar++) {
```
WallyMaier, line 924: Where does 20 come from?

WallyMaier (commented):
Thanks for this work, Pedro!  This looks good to me!

Common/src/geometry/meshreader/CSU2ASCIIMeshReaderFVM.cpp
```diff
@@ -718,705 +740,304 @@ void CSU2ASCIIMeshReaderFVM::SplitActuatorDiskSurface() {
...
-        string::size_type position;
-
-        for (unsigned short iChar = 0; iChar < 20; iChar++) {
-          position = text_line.find( " ", 0 );
-          if (position != string::npos) text_line.erase (position,1);
-          position = text_line.find( "\r", 0 );
-          if (position != string::npos) text_line.erase (position,1);
-          position = text_line.find( "\n", 0 );
-          if (position != string::npos) text_line.erase (position,1);
-        }
```
pcarruscag, line 1222: Comes from the ancient ones, I suppose we strip up to 20 spaces anywhere from the marker names :shrug: 
I doubt the \r and \n ever do anything, because these are used by getline to know what is a line.
This should not be performance or correctness critical, so if it ain't broke...

WallyMaier (approved):
LGTM!

Issue #1365, opened 2021-09-01
Axisymmetric viscous source term bug in energy equation


mskim0401:
Dear all,

In my opinion, there is a bug in axisymmetric viscous source term of energy equation.

![AxiSymm_3](https://user-images.githubusercontent.com/74228020/131637195-699fc230-e499-4b32-80a4-d309fa6adb01.png)
From the equation D-55 of reference in PR#1192 , the energy equation should be 
`...
+v*TWO3*(2*PrimVar_Grad_i[2][1]-PrimVar_Grad_i[1][0]-v*yinv))
...`

below is current code in `void CSourceAxisymmetric_Flow::ResidualDiffusion()` 
`residual[3] -= Volume*(yinv*(total_viscosity_i*(u*(PrimVar_Grad_i[2][0]+PrimVar_Grad_i[1][1])
                                                 +v*TWO3*(2*PrimVar_Grad_i[1][1]-PrimVar_Grad_i[1][0]
                                                 -v*yinv+U_i[0]*turb_ke_i))
                                                 +total_conductivity_i*PrimVar_Grad_i[0][1])
                                                 -TWO3*(AuxVar_Grad_i[1][1]+AuxVar_Grad_i[2][0]));` 

I appreciate if I miss something.
Thank you.

FlorianDm:
Oh thank you very much for pointing this out. I don't know how it happened but I will address it right away

FlorianDm:
fixed https://github.com/su2code/SU2/pull/1366

PR #1366, opened 2021-09-01
fix axisym source term + update regression cases

FlorianDm:
## Proposed Changes

fix https://github.com/su2code/SU2/issues/1365

### DIFF
--- a/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
+++ b/SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
@@ -378,7 +378,7 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
     residual[nSpecies+1] -= Volume*(yinv*total_viscosity_i*2*(GV[nSpecies+3][1]-v*yinv)
                                                              -TWO3*AuxVar_Grad_i[0][1]);
     residual[nSpecies+2] -= Volume*(yinv*(-sumJhs_y + total_viscosity_i*(u*(GV[nSpecies+3][0]+GV[nSpecies+2][1])
-                                                                        +v*TWO3*(2*GV[nSpecies+2][1]-GV[nSpecies+2][0]
+                                                                        +v*TWO3*(2*GV[nSpecies+3][1]-GV[nSpecies+2][0]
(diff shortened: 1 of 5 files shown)


FlorianDm:
I updated the restart file in the TestCases directory so I suppose regressions apart from this branch will break until this PR is merged. Sorry

pcarruscag:
I have some vague memory of the textbook having a typo?

WallyMaier:
@pcarruscag I also have a vague memory of that.  But I couldn't seem to find it in the previous discussions....

FlorianDm:
Yes it did but it's unrelated. Surprised this bug here could creep in. For the regression it seems like it is still using the old restart file and not the one I pushed to TestCases master a few hours ago. Not sure what I am missing here

FlorianDm:
@WallyMaier OK will do

WallyMaier:
@FlorianDm, not sure the issue either.  You could just do a "cold" start for the regression?

pcarruscag:
Don't push things to master directly
create a PR to develop and then merge, so we can keep track of things.
It's develop you need not master

FlorianDm:
Yes just realised that sorry
Will fix it

SU2_CFD/src/numerics/NEMO/NEMO_sources.cpp
```diff
@@ -378,7 +378,7 @@ CNumerics::ResidualType<> CSource_NEMO::ComputeAxisymmetric(const CConfig *confi
     residual[nSpecies+1] -= Volume*(yinv*total_viscosity_i*2*(GV[nSpecies+3][1]-v*yinv)
                                                              -TWO3*AuxVar_Grad_i[0][1]);
     residual[nSpecies+2] -= Volume*(yinv*(-sumJhs_y + total_viscosity_i*(u*(GV[nSpecies+3][0]+GV[nSpecies+2][1])
-                                                                        +v*TWO3*(2*GV[nSpecies+2][1]-GV[nSpecies+2][0]
+                                                                        +v*TWO3*(2*GV[nSpecies+3][1]-GV[nSpecies+2][0]
```
WallyMaier, line 381: Thanks!

WallyMaier (approved):
This looks good to me! Im still surprised how this slipped through.

PR #1375, opened 2021-09-19
Intermediate class of variable for all flow solver variables + Fix for MUSCL in NEMO solver

pcarruscag:
## Proposed Changes
```
CVariable -> CFlowVariable -> CEulerVariable -> CNSVariable
```
**Likewise for incompressible and NEMO.**

**CFlowVariable** stores, allocates, and gives access to, primitives, their gradients and limiters, and a few other odd things.
For now this is just cleanup but it can be used to give non-virtual access to these flow variables to other solvers.
For example, the scalar solver #1330 can static_cast or dynamic_cast the nodes of the flow solver to access some quantities more efficiently.
I'll leave that for after #1330 is merged.

## PR Checklist


SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -73,39 +67,18 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
...
-    }
-  }
-
-  /*--- Allocate undivided laplacian (centered) and limiter (upwind)---*/
-  if (config->GetKind_ConvNumScheme_Flow() == SPACE_CENTERED)
-    Undivided_Laplacian.resize(nPoint,nVar);
-
   /*--- Always allocate the slope limiter,
    and the auxiliar variables (check the logic - JST with 2nd order Turb model - ) ---*/
   Limiter.resize(nPoint,nVar) = su2double(0.0);
```
pcarruscag, line 79: @WallyMaier, this may be leftover from when NEMO reconstructed the conservatives?

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
```diff
@@ -117,31 +90,8 @@ CNEMOEulerVariable::CNEMOEulerVariable(su2double val_pressure,
   Gamma.resize(nPoint)           = su2double(0.0);
 
   /*--- Compressible flow, gradients primitive variables ---*/
-  Gradient_Primitive.resize(nPoint,nPrimVarGrad,nDim,0.0);
   Gradient.resize(nPoint,nVar,nDim,0.0);
```
pcarruscag, line 93: And this too?

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -0,0 +1,226 @@
+/*!
+ * \file CFlowVariable.hpp
+ * \brief Class for defining the common variables of flow solvers solver.
```
pcarruscag, line 3:
```suggestion
 * \brief Class for defining the common variables of flow solvers.
```

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -0,0 +1,226 @@
...
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "CVariable.hpp"
+
+/*!
+ * \class CFlowVariable
+ * \brief Class for defining the common variables of flow solvers solver.
```
pcarruscag, line 33:
```suggestion
 * \brief Class for defining the common variables of flow solvers.
```

SU2_CFD/src/meson.build
```diff
@@ -71,6 +71,7 @@ su2_cfd_src += files(['variables/CIncNSVariable.cpp',
                       'variables/CVariable.cpp',
                       'variables/CNSVariable.cpp',
                       'variables/CAdjTurbVariable.cpp',
+                      'variables/CFlowVariable.cpp'
```
pcarruscag, line 74:
```suggestion
                      'variables/CFlowVariable.cpp',
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -567,22 +557,22 @@ void CNEMOEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_con
...
         }
 
-        if (limiter) {
-          if (van_albada) {
-            su2double V_ij = V_j[iVar] - V_i[iVar];
-            Limiter_i[iVar] = V_ij*( 2.0*Project_Grad_i + V_ij) / (4*pow(Project_Grad_i, 2) + pow(V_ij, 2) + EPS);
-            Limiter_j[iVar] = V_ij*(-2.0*Project_Grad_j + V_ij) / (4*pow(Project_Grad_j, 2) + pow(V_ij, 2) + EPS);
-          }
-          if (lim_i > Limiter_i[iVar] && Limiter_i[iVar] != 0) lim_i = Limiter_i[iVar];
-          if (lim_j > Limiter_j[iVar] && Limiter_j[iVar] != 0) lim_j = Limiter_j[iVar];
```
pcarruscag, line 577: This was forcing 1st order because lim_i and lim_j were init to 0

CatarinaGarbacz: @fmpmorgado you were working on this before, maybe you can double-check if this is ok?

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: i think this was there just because we had been trying to mirror other solvers. but in NEMO this seems to not to be used actually. @WallyMaier can you confirm ?

CatarinaGarbacz: in any case the change to CFlowVariable seems not to harm anything

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
CatarinaGarbacz: this is used for primitive reconstruction, and for diffusion terms in the NS solver. I believe that the move to CFlowVariable seems ok

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
fmpmorgado: Thank you for pointing this out @pcarruscag . This has been corrected on feature_NEMO but looks like the fix was not pushed to develop.

Also, @WallyMaier following your last year commit on this part that I just noticed now. As it is, we are using a more conservative approach in feature_NEMO than in develop. In feature_NEMO we are using the same limiter for all the primitive variables (the limiter with lowest value), while in develop we are using a different limiter for each primitive.

For now, I don't see any issues, but we may want to keep this in mind.

pcarruscag: Ok, I can update the residuals of the two cases failing then? They should also work with 2nd order? Otherwise I can switch the config to 1st order but that is not ideal for coverage.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: @pcarruscag @CatarinaGarbacz This is left over from when the other solvers were mimicked a long time ago.  Before I gave up on implementing some centered schemes. This should be superfluous.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: I am skeptical that "Gradient" is used anymore.  After a quick look, I believe it may be unused.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: @fmpmorgado feature_NEMO is very out-of-date with develop, so It may be a good idea to update that to stay consistent. 

@pcarruscag, the residuals can be updated, we should be able to do 2nd order.  Thanks for finding this bug!

WallyMaier:
@pcarruscag LGTM!  Thanks for putting in the work and abstracting more of the repeated routines.

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -483,52 +323,52 @@ class CNEMOEulerVariable : public CVariable {
   /*!
    * \brief Retrieves the value of the species density in the primitive variable vector.
    */
-  inline unsigned short GetRhosIndex(void) { return RHOS_INDEX; }
+  inline unsigned short GetRhosIndex() const { return RHOS_INDEX; }
```
WallyMaier, line 326: Thanks! 👍

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -377,15 +377,17 @@ void CNEMOCompOutput::SetVolumeOutputFields(CConfig *config){
...
+    AddVolumeOutput("LIMITER_DENSITY", "Limiter_Density", "LIMITER", "Limiter value of the density");
+    AddVolumeOutput("LIMITER_MOMENTUM-X", "Limiter_Momentum_x", "LIMITER", "Limiter value of the x-momentum");
+    AddVolumeOutput("LIMITER_MOMENTUM-Y", "Limiter_Momentum_y", "LIMITER", "Limiter value of the y-momentum");
+    if (nDim == 3)
+      AddVolumeOutput("LIMITER_MOMENTUM-Z", "Limiter_Momentum_z", "LIMITER", "Limiter value of the z-momentum");
+    AddVolumeOutput("LIMITER_ENERGY", "Limiter_Energy", "LIMITER", "Limiter value of the energy");
+  }
 
-  if (config->GetKind_SlopeLimit_Flow() != NO_LIMITER) {
+  if (config->GetKind_SlopeLimit_Turb() != NO_LIMITER) {
```
WallyMaier, line 390: Good catch!

WallyMaier (approved):

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: But "Gradient" does not appear to be used.

SU2_CFD/src/variables/CNEMOEulerVariable.cpp
WallyMaier: After futher looking through, it seems the only time NEMO uses "Limiter" is as a bool.

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -0,0 +1,226 @@
...
+   * \param[in] iVar - Index of the variable.
+   * \return Value of the primitive variable for the index <i>iVar</i>.
+   */
+  inline su2double GetPrimitive(unsigned long iPoint, unsigned long iVar) const final { return Primitive(iPoint,iVar); }
+
+  /*!
+   * \brief Get all the primitive variables of the problem.
+   * \return Pointer to the primitive variable vector.
+   */
+  inline su2double *GetPrimitive(unsigned long iPoint) final { return Primitive[iPoint]; }
```
TobiKattmann, line 77: I think you did not run clang-format over this the latest file 🤨  `pointerAlignement: left` is my reason here :)

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -0,0 +1,226 @@
...
+
+  /*!
+   * \brief Set the value of the primitive variables.
+   * \param[in] iVar - Index of the variable.
+   * \param[in] iVar - Index of the variable.
+   * \return Set the value of the primitive variable for the index <i>iVar</i>.
+   */
+  inline void SetPrimitive(unsigned long iPoint, unsigned long iVar, su2double val_prim) final {
+    Primitive(iPoint,iVar) = val_prim;
+  }
```
TobiKattmann, line 93: How do we want to go about the doxygen documentation? In many cases, especially the `param[in]` list is incorrect. Do we want to a) fix it: tedious work (but not too much here) b) Remove all `iPoint, iVar, iDim, ...` and only keep the interesting ones or c) keep as is d) do sth else?

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
```diff
@@ -27,48 +27,34 @@
...
 
-#include "CVariable.hpp"
+#include "CFlowVariable.hpp"
 #include "../fluid/CNEMOGas.hpp"
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
 
 /*!
  * \class CNEMOEulerVariable
  * \brief Main class for defining the variables of the NEMO Euler's solver.
+ * \note Primitive variables (rhos_s, T, Tve, ...)
```
TobiKattmann, line 37: Can somone from the NEMO-team add the `...` the the `Gradients of Primitives` part like in the other solvers @WallyMaier @CatarinaGarbacz , or is this not really applicable here?

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -29,22 +29,17 @@
...
-  if (config->GetAxisymmetric()) {
-    nAuxVar = 1;
-    AuxVar.resize(nPoint,nAuxVar) = su2double(0.0);
-    Grad_AuxVar.resize(nPoint,nAuxVar,nDim);
-  }
-
   /*--- Allocate memory for the AuxVar+gradient of eddy viscosity mu_t ---*/
-  if (config->GetStreamwise_Periodic_Temperature() && (config->GetKind_Turb_Model() != NONE)) {
+  if (config->GetAxisymmetric() ||
+      (config->GetStreamwise_Periodic_Temperature() && (config->GetKind_Turb_Model() != NONE))) {
```
TobiKattmann, line 42: Is it eddy viscosity the AuxVar(Grad) for the axisymmetric cases as well ❔

TestCases/parallel_regression.py
```diff
@@ -83,7 +83,7 @@ def main():
     visc_cone.cfg_dir   = "nonequilibrium/axi_visccone"
     visc_cone.cfg_file  = "axi_visccone.cfg"
     visc_cone.test_iter = 10
-    visc_cone.test_vals = [-5.173078, -5.697841, -20.831296, -20.719164, -23.419769, -1.564084, -2.069040, 2.203924, -2.590729]
+    visc_cone.test_vals = [-5.175785, -5.700548, -20.705025, -20.592408, -22.851499, -1.540045, -2.104353, 2.197906, -2.576183]
```
TobiKattmann, line 86: so many regression values ❤️

TobiKattmann (approved):
Thanks @pcarruscag 💐 Some generalization is always nice. I only a have a few minor points. 

I'll try to finish #1330 (template the ScalarLayer on the specialized type of variable), so this can be merged

SU2_CFD/src/variables/CIncNSVariable.cpp
pcarruscag: idk

SU2_CFD/include/variables/CFlowVariable.hpp
pcarruscag: something consistent

TobiKattmann: updated in  aecc95f

SU2_CFD/include/variables/CNEMOEulerVariable.hpp
WallyMaier: @TobiKattmann I can fill out the primitives.  I need to double check how much of the Gradient/Limiter primitive is filled out.

WallyMaier, line 37:
```suggestion
 * \note Primitive variables (rhos_s, T, Tve, vx, vy, vw, P, rho, h, rhoCvtr, rhoCvve)
```

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -29,22 +29,17 @@
...
   DES_LengthScale.resize(nPoint) = su2double(0.0);
   Max_Lambda_Visc.resize(nPoint);
 
-  if (config->GetAxisymmetric()) {
-    nAuxVar = 1;
-    AuxVar.resize(nPoint,nAuxVar) = su2double(0.0);
-    Grad_AuxVar.resize(nPoint,nAuxVar,nDim);
-  }
-
   /*--- Allocate memory for the AuxVar+gradient of eddy viscosity mu_t ---*/
```
TobiKattmann, line 40:
```suggestion
  /*--- Allocate memory for the AuxVar+gradient. See CIncEulerSolver::Source_Residual:
   * Axisymmetric: total-viscosity * y-vel / y-coord
   * Streamwise Periodic: eddy viscosity mu_t ---*/
```

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1391,137 +1385,141 @@ void CTurbSASolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CC
...
-            deltaDDES = geometry->nodes->GetMaxLength(iPoint);
+        const su2double deltaDDES = geometry->nodes->GetMaxLength(iPoint);
+
+        su2double delta[MAXNDIM] = {}, ratioOmega[MAXNDIM] = {};
+
+        for (const auto jPoint : geometry->nodes->GetPoints(iPoint)) {
+          const auto coord_j = geometry->nodes->GetCoord(jPoint);
+          for (auto iDim = 0u; iDim < nDim; iDim++){
+            const su2double deltaAux = abs(coord_j[iDim] - coord_i[iDim]);
+            delta[iDim] = max(delta[iDim], deltaAux);
```
TobiKattmann, line 1445: I currently use gnu-5.3.0 on our cluster which gives me an `array subscript is above array bounds`-warning for the `max(delta[iDim]...`. [This](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59124) makes me believe that it is problem maybe including this version number.

using `unsigned short` instead of `auto` fixes this unjustified warning. I know it is not an actual problem but I would change that over at #1382 if ok

Issue #763, opened 2019-08-21
Periodic boundary conditions and simple timestepping don't combine well


clarkpede:
**Describe the bug**

Periodic boundary conditions work well for steady problems and dual-time-stepping.  But when I try to use time-stepping (not dual-time-stepping), the periodic boundary condition shows some artifacts.  For a problem that's homogeneous in the periodic direction, the transient solution is not homogeneous in the periodic direction.

**To Reproduce**

To verify the problem, I looked at the "navierstokes/poiseuille" test case.  This case is homogeneous in the y direction, and has periodic boundaries.  Instead of treating it as a steady-state problem, I did time-marching wit simple time-stepping.  I then compared the results with the same simulation on v6.2.0 (before PR #652 changed the periodic boundaries). The y-momentum is very similar in both cases. However, the x-momentum shows inhomogeneity on the develop branch:

![comparison](https://user-images.githubusercontent.com/13340225/63468140-bc674380-c42c-11e9-976a-ba94def43bd8.gif)

Here are the files to recreate my test case:
[periodic_timestepping.tar.gz](https://github.com/su2code/SU2/files/3527333/periodic_timestepping.tar.gz)

**My Analysis**

One of my colleagues pointed out that `CSolver::InitiatePeriodicComms` and `CSolver::CompletePeriodicComms` only communicate the residual and the Jacobian of point "i" with respect to point "i".  The Jacobian of point "i" with respect to point "j" is *not* communicated.  This means that a different problem is being passed to the linear solver at the periodic nodes.  In v6.2.0, a halo node approach was used, so the linear solver problem was identical for all nodes along a homogeneous direction.

If this analysis is correct, then this could be fixed by communicating the full Jacobian (with respect to all neighbors) at the periodic points.

**Desktop**
 - OS: Ubuntu 18.04
 - C++ compiler and version: icc 19.0.2.187
 - MPI implementation and version: mpich2/3.2
 - SU2 Version: develop branch

economon:
Nice analysis @clarkpede. It is true that we simplify the Jacobians at the periodic boundaries, mostly to avoid issues with adding entries to the Jacobian from the neighbors that potentially do not live on our rank and to keep communication costs low (those neighbors are treated explicitly). This could be changed to communicate the full Jacobian.. but I am not sure it is worth the effort/cost.

[...]

PR #1382, opened 2021-09-25
Add periodic boundary conditions to the heat-equation solver

TobiKattmann:
Dear friends of the Heat Solver and Periodic Boundary conditions, 

I feared no exhausting work and added periodic BC to the Heat Solver.

Stock up your supplies, cancel meetings for today and dive into the diff to review this piece.

Ok for real, I don't know if that is all that has to be done? I did some early testing with mixed results, I noticed that I had to turn down CFL for the solid zone (1e4 in non-periodic case-> 1) to get a solution that seems to respect the periodicity. If the CFL is too large then I get some spurious temperature and weird residual pattern. Any comments are appreciated. This was just a first shot so Im gonna do more digging  on that issue

## Related Work
Periodic bc was initially implemented in #652 

## PR Checklist


TobiKattmann:
(I will add info later but I am in a bit of a hurry currently) with CFL=1 
image

with CFL=1e3
image

residuals with CFL=1e3
image

pcarruscag:
Have a look at my last comment in #763, I think there is a better way of handling linear system periodicity instead of what we do, which may allow you to run at higher CFL (and with simpler periodic comms).

But at the moment you are missing the PERIODIC_IMPLICIT comms after solving the linear system, see CompleteImplicitIteration_impl in CFVMFlowSolverBase.hpp.

TobiKattmann:
Thanks @pcarruscag , I had to do a few additional things together with the PERIODIC_IMPLICIT communication to make it work, but that also brought me on the right track 👍 Now it works really nicely from my point of view.

TobiKattmann:
Here the "new" solution of the above case:
image

And the Residual_temperature at the periodic border. Not that the very same pattern can be seen on the non-periodic pin upstream. Also for me this kind of blotchy residual pattern is a sign for me that the case fully converged without some "bad" spot of high residual:
image

And a picture of a CHT case with now added periodicity in the solids. Above with periodicity and below without. 
image

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -1274,11 +1301,12 @@ void CHeatSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_
...
-    if (nodes->GetDelta_Time(iPoint) != 0.0) {
-
+    const su2double dt = nodes->GetDelta_Time(iPoint);
+    if (dt != 0.0) {
+      /*--- For nodes on periodic boundaries, add the respective partner volume. ---*/
       // Identical for flow and heat
-      const su2double Delta = geometry->nodes->GetVolume(iPoint) / nodes->GetDelta_Time(iPoint);
-      Jacobian.AddVal2Diag(iPoint, Delta);
+      const su2double Vol = geometry->nodes->GetVolume(iPoint) + geometry->nodes->GetPeriodicVolume(iPoint);
+      Jacobian.AddVal2Diag(iPoint, Vol / dt);
```
TobiKattmann, line 1309: I made this bit consistent with the Scalar Solver.. which will a possible port later a bit

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -159,6 +159,19 @@ CHeatSolver::CHeatSolver(CGeometry *geometry, CConfig *config, unsigned short iM
...
+  for (unsigned short iPeriodic = 1; iPeriodic <= config->GetnMarker_Periodic() / 2; iPeriodic++) {
+    InitiatePeriodicComms(geometry, config, iPeriodic, PERIODIC_VOLUME);
+    CompletePeriodicComms(geometry, config, iPeriodic, PERIODIC_VOLUME);
+    InitiatePeriodicComms(geometry, config, iPeriodic, PERIODIC_NEIGHBORS);
+    CompletePeriodicComms(geometry, config, iPeriodic, PERIODIC_NEIGHBORS);
+  }
+  /*--- Store if implicit scheme is used. This has implications on the Residual and Jacobian handling for periodic
+   * boundaries  ---*/
+  const bool euler_implicit = (config->GetKind_TimeIntScheme_Heat() == EULER_IMPLICIT);
+  SetImplicitPeriodic(euler_implicit);
```
TobiKattmann, line 172: The ScalarSolver is currently not doing that. I think there it is enough as the flow solver already does that. But if a ScalarSolver-child is standalone then it prob has to be incorporated. But this is more of a guess than safe info

TobiKattmann:
@NAnand-TUD We talked about this briefly in a recent dev-meeting, so that you know :)

TobiKattmann:
I'll adapt one of the existing Testcases to use this feature

pcarruscag (approved):
LGTM

pcarruscag:
With what CFL can you run now?

TobiKattmann:
> With what CFL can you run now?

For the solid only case I had the feeling that I could go arbitrarily high. 1e8 and 1e12 had no difference.

For the cht case I kept the cfl of 1e4 from the non-periodic case. There I will try a bit more to see what's possible.

NAnand-TUD:
@TobiKattmann Thanks for sharing! I will be testing this part of the code with our test case soon.

TobiKattmann:
> 
> 
> I'll adapt one of the existing Testcases to use this feature

Well that was a lie. I thought it would be a good idea to adapt the 2D cht streamwise periodic case but I forgot that the used mesh is used for a streamwise periodic primal, FD and DA case as well as HeatTransfer BC Testcase (and that is a bit of work tbh). And for the periodic BC I had to change the mesh (adding new markers). As we do not have a single standalone (always run in a CHT setup) heat-equation Testcase, I went ahead and cut out the solid zone only and made that a Testcase. I already merged the gmsh geo and .su2 over at su2code/Testcases/pull/76 . If this is not liked I can of course do sth different. Note that the new Testcase takes ~2s to complete its 750 iterations until convergence.

TobiKattmann:
I found solid only and CHT cases to take ~10x the iterations to reach the same convergence levels if periodic boundary conditions are used in the solid (left: using periodic BC, right: not using periodic BC). This is not really ideal tbh, Maybe I have not found a suitable setup yet (I modifed some things, Linear Solver iter, CFL, Gradient Method) or there is a reasonable explanation for this in how the bc is handeled (Therefore I would have to take a look into #763 I guess ;) )

image

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1440,7 +1440,7 @@ void CTurbSASolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CC
 
         for (const auto jPoint : geometry->nodes->GetPoints(iPoint)) {
           const auto coord_j = geometry->nodes->GetCoord(jPoint);
-          for (auto iDim = 0u; iDim < nDim; iDim++){
+          for (unsigned short iDim = 0u; iDim < nDim; iDim++){
```
pcarruscag, line 1443: Revert please. This is inconsistent and it is not su2's problem.

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -1326,6 +1354,12 @@ void CHeatSolver::ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_
     }
   }
 
+  /*--- Synchronize the solution between master and passive periodic nnodes after the linear solve. ---*/
```
TobiKattmann, line 1357:
```suggestion
  /*--- Synchronize the solution between master and passive periodic nodes after the linear solve. ---*/
```

WallyMaier (commented):
LGTM

PR #1383, opened 2021-09-29
Addition of species diffusion term to surface heat flux

jtneedels:
## Proposed Changes

- Addition of species diffusion terms to surface energy balance in NEMO, important for highly dissociated flows
- Addition of vector for s species, "enthalpys", to variable to store values accessed in solver.
- General cleanup/commenting of the NEMO heat flux computation in CFVMFlowSolverBase.inl.

## PR Checklist


jtneedels:
Any ideas on why this might be failing for the one incompressible regression test? I don't think I modified any files that would affect it.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -937,6 +937,22 @@ class CVariable {
    */
   inline virtual su2double GetMassFraction(unsigned long iPoint, unsigned long val_Species) const { return 0.0; }
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   * \param[in] val_Species - Index of species s.
+   * \return Value of the mass fraction of species s.
+   */
+  inline virtual su2double GetEve(unsigned long iPoint) const { return 0.0; }
```
pcarruscag, line 946: Looks unused

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -937,6 +937,22 @@ class CVariable {
...
+
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   * \param[in] val_Species - Index of species s.
+   * \return Value of the mass fraction of species s.
+   */
+  inline virtual su2double* GetEnthalpys(unsigned long iPoint) { return nullptr; }
```
pcarruscag, line 954: Move to CNEMOEulerVariable please, since it looks this only needs to be visible in the solver

SU2_CFD/src/variables/CNEMONSVariable.cpp
```diff
@@ -102,6 +104,14 @@ bool CNEMONSVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)
   for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
     DiffusionCoeff(iPoint, iSpecies) = Ds[iSpecies];
 
+  su2double T   =  Primitive(iPoint,nSpecies);
+  su2double Tve =  Primitive(iPoint,nSpecies+1);
+
+  su2double* val_eves = GetEve(iPoint);
+  hs = fluidmodel->ComputeSpeciesEnthalpy(T, Tve, val_eves);
+  for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
```
pcarruscag, line 112: Try changing the local "iVars" and "iSpecies" declared in the NEMO variable member functions to unsigned long to see if we can fix some of those LGTM warnings. (e.g. line 82 of this file)

SU2_CFD/include/variables/CNEMONSVariable.hpp
```diff
@@ -46,8 +46,10 @@ class CNEMONSVariable final : public CNEMOEulerVariable {
   VectorType LaminarViscosity;  /*!< \brief Viscosity of the fluid. */
   VectorType ThermalCond;       /*!< \brief T-R thermal conductivity of the gas mixture. */
   VectorType ThermalCond_ve;    /*!< \brief V-E thermal conductivity of the gas mixture. */
-  vector<su2double> thermalconductivities;
-  vector<su2double> Ds;
+  vector<su2double> thermalconductivities;   /*!< \brief Species thermal conductivities. */
+  vector<su2double> Ds;    /*!< \brief Species diffusion coefficients. */
+  vector<su2double> hs;    /*!< \brief Species enthalpies. */
```
pcarruscag, line 51: This is not a good way of doing this.
The methods in the fluid model return vectors by reference, but by assigning to these member variables a copy is forced.
1) Try removing these declarations.

SU2_CFD/src/variables/CNEMONSVariable.cpp
```diff
@@ -102,6 +104,14 @@ bool CNEMONSVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)
   for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
     DiffusionCoeff(iPoint, iSpecies) = Ds[iSpecies];
 
+  su2double T   =  Primitive(iPoint,nSpecies);
+  su2double Tve =  Primitive(iPoint,nSpecies+1);
+
+  su2double* val_eves = GetEve(iPoint);
+  hs = fluidmodel->ComputeSpeciesEnthalpy(T, Tve, val_eves);
```
pcarruscag, line 111: 2)
```suggestion
  const auto& hs = fluidmodel->ComputeSpeciesEnthalpy(T, Tve, val_eves);
```
Like this there is no copy.
And similarly for the others I mentioned.

SU2_CFD/include/variables/CVariable.hpp
jtneedels: Yes, you're right, this was vestigial and isn't necessary. Removing it.

SU2_CFD/include/variables/CVariable.hpp
jtneedels: Sorry, should I move the virtual function def from CVariable -> CNEMOEulerVariable, or should I move the function def from CNEMONSVariable -> CNEMOEulerVariable?

SU2_CFD/src/variables/CNEMONSVariable.cpp
jtneedels: Sure will do. Out of curiosity, why does this throw a warning? Is it because if the iVar is unsigned short, there are values that can be stored in an unsigned long that it cannot hold?

pcarruscag: That's the only reason I can think of, it is general good practice to avoid mixing types.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: What error? In FVMFlowSolverBase the nodes are of type CNEMOEulerVariable, so this should work.

SU2_CFD/include/variables/CNEMONSVariable.hpp
```diff
@@ -46,8 +46,8 @@ class CNEMONSVariable final : public CNEMOEulerVariable {
   VectorType LaminarViscosity;  /*!< \brief Viscosity of the fluid. */
   VectorType ThermalCond;       /*!< \brief T-R thermal conductivity of the gas mixture. */
   VectorType ThermalCond_ve;    /*!< \brief V-E thermal conductivity of the gas mixture. */
-  vector<su2double> thermalconductivities;
-  vector<su2double> Ds;
+  vector<su2double> thermalconductivities;   /*!< \brief Species thermal conductivities. */
```
pcarruscag, line 49: Do the same with `thermalconductivities` "while you are here" please.

SU2_CFD/include/variables/CVariable.hpp
jtneedels: This is the error I get on compiling:
`In file included from ../SU2_CFD/src/solvers/CNSSolver.cpp:32:
../SU2_CFD/src/solvers/../../include/solvers/CFVMFlowSolverBase.inl:2592:53: error: no member named ‘GetEnthalpys’ in ‘CEulerVariable’
    const auto hs           = nodes->GetEnthalpys(iPoint);
                       ~~~~~ ^
../SU2_CFD/src/solvers/CNSSolver.cpp:36:16: note: in instantiation of member function ‘CFVMFlowSolverBase<CEulerVariable, ENUM_REGIME::COMPRESSIBLE>::Friction_Forces’ requested here
template class CFVMFlowSolverBase<CEulerVariable, ENUM_REGIME::COMPRESSIBLE>;`

pcarruscag: Dohh :facepalm: sorry my bad. You can leave it in CVariable then.

SU2_CFD/include/variables/CNEMONSVariable.hpp
```diff
@@ -100,6 +99,12 @@ class CNEMONSVariable final : public CNEMOEulerVariable {
    */
   inline su2double* GetDiffusionCoeff(unsigned long iPoint) override { return DiffusionCoeff[iPoint]; }
 
+    /*!
+   * \brief Get the species enthalpy.
+   * \return Value of the species enthalpy.
+   */
```
pcarruscag, line 105:
```suggestion
  /*!
   * \brief Get the species enthalpy.
   * \return Value of the species enthalpy.
   */
```

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -937,6 +937,14 @@ class CVariable {
    */
   inline virtual su2double GetMassFraction(unsigned long iPoint, unsigned long val_Species) const { return 0.0; }
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   * \param[in] val_Species - Index of species s.
+   * \return Value of enthalpy of species s.
+   */
```
pcarruscag, line 945:
```suggestion
  /*!
   * \brief Get the species enthalpy.
   * \return Value of the species enthalpy.
   */
```

SU2_CFD/src/variables/CNEMONSVariable.cpp
```diff
@@ -98,13 +99,21 @@ bool CNEMONSVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)
...
+
+  su2double* val_eves = GetEve(iPoint);
+  const auto& hs = fluidmodel->ComputeSpeciesEnthalpy(T, Tve, val_eves);
+  for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
+    Enthalpys(iPoint, iSpecies) = hs[iSpecies];
+  
   LaminarViscosity(iPoint) = fluidmodel->GetViscosity();
 
-  thermalconductivities    = fluidmodel->GetThermalConductivities();
+  const auto& thermalconductivities    = fluidmodel->GetThermalConductivities();
```
pcarruscag, line 116:
```suggestion
  const auto& thermalconductivities = fluidmodel->GetThermalConductivities();
```

pcarruscag (approved):
LGTM code-wise

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2581,16 +2582,36 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
+
         const auto thermal_conductivity_tr = nodes->GetThermalConductivity(iPoint);
         const auto thermal_conductivity_ve = nodes->GetThermalConductivity_ve(iPoint);
         const auto Grad_PrimVar            = nodes->GetGradient_Primitive(iPoint);
+        const auto PrimVar                 = nodes->GetPrimitive(iPoint);
+        const auto Ds                      = nodes->GetDiffusionCoeff(iPoint);
+        const auto hs                      = nodes->GetEnthalpys(iPoint);
+
+        su2double rho                      = PrimVar[RHO_INDEX]; 
+        su2double dYsn;
```
pcarruscag, line 2595: Actually...
```suggestion
        const su2double rho = PrimVar[RHO_INDEX]; 
```

pcarruscag: declare the variables locally as you need them

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2581,16 +2582,36 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
+
+        /*--- Compute enthalpy transport to surface due to mass diffusion ---*/ 
+        su2double sumJhs = 0.0;
+        for (iDim = 0; iDim < nDim; iDim++) {
+          for (iSpecies = 0; iSpecies < nSpecies; iSpecies++) {
+            dYsn = Grad_PrimVar[RHOS_INDEX+iSpecies][iDim]/rho*UnitNormal[iDim];
+            sumJhs += rho*Ds[iSpecies]*dYsn*hs[iSpecies];
+          }
+        }
```
pcarruscag, line 2604: Clean this up please, you're dividing by rho, then multiplying on the line below.
And use the geometry toolbox for dot products with the unit normal, saves the loop over nDim (which is in the wrong order, should be inside the species loop).

pcarruscag (changes requested):

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
jtneedels: Ok, I'll do the same for the compressible and incompressible heat fluxes as well. It looks like there are a lot of iDim loops that can be replaced in friction forces. I can open a separate PR to replace those, would that be a good idea?

pcarruscag: No need to fix all of them (unless you really want to of course) not introducing new ones is enough :)

jtneedels: Haha sounds good, I'll do those two since it's quick. Is the benefit to using GeometryToolbox for dot products because it reduces lines of code, and should be faster since its an inline function?

jtneedels:
I'm getting a warning that GradTemperature may not be uninitialized, but it is. Any idea why this is popping up all of the sudden?

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2527,26 +2527,21 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
       }
-      WallShearStress[iMarker][iVertex] = sqrt(WallShearStress[iMarker][iVertex]);
+      WallShearStress[iMarker][iVertex] = GeometryToolbox::Norm(nDim, TauTangent);
 
       for (iDim = 0; iDim < nDim; iDim++) WallDist[iDim] = (Coord[iDim] - Coord_Normal[iDim]);
-      WallDistMod = 0.0;
-      for (iDim = 0; iDim < nDim; iDim++) WallDistMod += WallDist[iDim] * WallDist[iDim];
-      WallDistMod = sqrt(WallDistMod);
+      
+      WallDistMod = GeometryToolbox::Norm(nDim, WallDist);
```
WallyMaier, line 2544:
```suggestion
      su2double WallDistMod = GeometryToolbox::Norm(nDim, WallDist);
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2562,35 +2557,41 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
+        const auto& Grad_PrimVar            = nodes->GetGradient_Primitive(iPoint);
+        const auto& Ds                      = nodes->GetDiffusionCoeff(iPoint);
+        const auto& hs                      = nodes->GetEnthalpys(iPoint);
+
+        /*--- Compute enthalpy transport to surface due to mass diffusion ---*/ 
+        su2double sumJhs = 0.0;
+        for (iSpecies = 0; iSpecies < nSpecies; iSpecies++)
+          sumJhs += Ds[iSpecies]*hs[iSpecies]*GeometryToolbox::DotProduct(nDim, Grad_PrimVar[RHOS_INDEX+iSpecies], UnitNormal);
+        
+        dTn   = GeometryToolbox::DotProduct(nDim, Grad_PrimVar[T_INDEX], UnitNormal);
```
WallyMaier, line 2589:
```suggestion
        su2double dTn   = GeometryToolbox::DotProduct(nDim, Grad_PrimVar[T_INDEX], UnitNormal);
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
pcarruscag: Aside from cleanup / expressiveness (i.e. reading a word instead of code), we gain some safety by being able to:
`const su2double a = GeometryToolbox::Something(...)`
Instead of having to allow `a` to be modified.
On the performance side, one of the main sources of inefficiency in SU2 is that the compilers "don't know" that loops over nDim, nVar, nSpecies, are super small, and they often make them inefficient by applying optimizations that are only profitable for large loops.
By encapsulating those loops over nDim we have a single place where we can inform the compilers (via all sorts of compiler-specific directives or in the limit by stating "nDim is 2 or 3") we are not doing this yet.

pcarruscag (approved):
Thanks for the additional cleanup :+1:

PR #1389, opened 2021-10-03
Cleanup in NEMO

jtneedels:

## Proposed Changes

- General cleanup/optimization in NEMO solver/numerics

 

## PR Checklist

### DIFF
--- a/SU2_CFD/src/numerics/NEMO/NEMO_diffusion.cpp
+++ b/SU2_CFD/src/numerics/NEMO/NEMO_diffusion.cpp
@@ -233,7 +233,6 @@ CAvgGradCorrected_NEMO::~CAvgGradCorrected_NEMO(void) {
 CNumerics::ResidualType<> CAvgGradCorrected_NEMO::ComputeResidual(const CConfig *config) {
 
   unsigned short iSpecies;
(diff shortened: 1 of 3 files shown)


pcarruscag (approved):

SU2_CFD/src/numerics/NEMO/NEMO_diffusion.cpp
```diff
@@ -336,5 +328,4 @@ CNumerics::ResidualType<> CAvgGradCorrected_NEMO::ComputeResidual(const CConfig
   }
 
   return ResidualType<>(Flux, nullptr, nullptr);
-}
-
+}
```
TobiKattmann, line 331: I am not sure whether there are negative consequences of not having a newline at the end of the file, but if GitHub warns about this one might as well just add a newline :)

jtneedels: Yes good point, it looks like this is mainly to differentiate a file with/without the newline in the diff, but I guess can also cause issues in some cases if files are concatenated. I'm going to open up a second PR to clean up the NEMO convective schemes, I'll address this fix.

Issue #1364, opened 2021-08-27
SA and SST turbulence model consistent implementation with literature, and handling more options


pcarruscag:
_Originally posted by @suargi in https://github.com/su2code/SU2/issues/1066#issuecomment-906376697_

Currently each variation/correction model has its unique name identification in SU2. In order to simultaneously combine multiple correction models to a given turbulence model I suggest the following:
- The config file option KIND_TURB_MODEL should specify only which turbulence model to use, i.e. SA or SST.
- Introduce an additional config file field TURB_MODEL_CORRECTIONS to indicate all variations/corrections to be considered. 

Here an example,
% Specify turbulent model (NONE, SA, SST)
KIND_TURB_MODEL= SA
% Turbulence model corrections (NONE, SA-EDW, SA-NOFT2, SA-COMP, SA-NEG, SA-QCR2000, SST-SUST)
TURB_MODEL_CORRECTIONS= SA-EDW, SA-NEG

What do you think?

suargi:
As discussed I will follow your suggestions and implement the new procedure of specifying the turbulence models guaranteeing backwards compatibility of the config file.              
                                                                                                                                                                                    
Besides, the baseline or standalone Spalart-Allmaras turbulence model currently implemented in SU2 is actually its ft2=0 variation, a.k.a noft2. To keep compatibility with previous SU2 versions results it is preferable to keep the current nomenclature, i.e., SA will refer to the SA-noft2 variant. In that sense, to introduce the actual baseline version an additional variant should be implemented, yesft2 :), corresponding to the standard model found in the literature. I will make sure to properly document this matter in the SU2 website.

erangit:
The ft2 term was a "not a very successful" attempt to allow computation of transitional flow or relaminarization, where a region exists without turbulent diffusion. During the time that passed since the nineties, it was commonly considered ineffective. I think Philipe also agrees with that. That way, though historically it is not correct, I think the basic SA model should be with ft2=0
I appreciate your effort. I put some order and logic in a combination that creates a lot of misunderstanding by many
Best,
Eran

HexFluid:
@pcarruscag

Hi Pedro, 
While I am trying to figure out the implementation of SA model in SU2, I found in Ver 7.2.1 /SU2/SU2_CFD/src/numerics/turbulent/turb_sources.cpp, line 103:

  if (rotating_frame) { Omega += 2.0*min(0.0, StrainMag_i-Omega); }

[...]

PR #1413, opened 2021-10-22
Refactor SA source terms to modularize the specification of model variants and correction terms

suargi:
## Pull request description
The idea of this pull request is to provide a new code structure to implement the turbulence models in SU2.  In the new approach, "all" variations/correction from a turbulence model should be compatible with each other and the user can use them simultaneously. Most of these variations/corrections only affect to the source terms, thus it forms the main contribution to this pull request.

## Short history summary
Originally the idea behind such an approach is given by @clarkpede in [here](https://github.com/su2code/SU2/issues/992#issuecomment-652446915).
The code implementation was suggested by @pcarruscag in [here](https://github.com/su2code/SU2/pull/1066#discussion_r470975915).
For the user interface, this is being or was discussed in [here](https://github.com/su2code/SU2/issues/1364#issue-981474153).

Some time ago, I made a PR #1066 to correct the SA model in SU2. Further I had the intention to implement the above presented approach to the SA for that pull request. However since the the presented approach can also be applied to the [SST](https://github.com/su2code/SU2/discussions/1403#discussion-3625259) or any turbulence model, I decided to create a new pull request.

_Please note that this is a very draft pull request, it is not meant to work at all._

## Documentation
Down below I will explain the "theory" behind it as well as the code structure that I will try to follow, just in case someone wants to replicate it for the SST.

Regarding the user interface, I did not work on that yet. I will address it at the end.

### Theory
The Spalart-Allamaras correction/variation models to be implemented in SU2 are listed below. One can find a detailed description for example in [NASA TMR website](https://turbmodels.larc.nasa.gov/spalart.html).
- Spalart-Allmaras standard
- Spalart-Allmaras negative
- Spalart-Allmaras **yes** ft2 term
- Spalart-Allmaras mixing layer compressibility correction
- Spalart-Allmaras Edwards modification
- Spalart-Allmaras QCR-2000

The first step is to identify the core, i.e., the mathematical formulation that is shared/common between all of them. Below I provide a very short description of their main characteristics.

#### Spalart-Allmaras standard
In the literature one can find a correction that sets the ft2-term to zero. However, the current baseline model in SU2 sets this ft2-term to zero. As discussed in [here](https://github.com/su2code/SU2/issues/1364#issuecomment-916890178) in order to keep compatibility with previous SU2 versions results it is preferable to keep the current nomenclature, i.e., SA will refer to the SA-noft2 variant.

#### Spalart-Allmaras negative
The model is the same as the "baseline" version when the turbulence variable $\tilde \nu$ is greater than or equal to zero. The modified vorticity is modified to prevent it from acquiring negative values. When $\tilde \nu$ is negative a slightly different integral equation is solved. The viscous fluxes are redefined as well.

The boundary conditions are the same as for the standard S-A.

#### Spalart-Allmaras nonzero ft2 term
This is the approach corresponds to the faithful baseline. The equations are the same as for the "baseline" version (SA), except we take the term $f_{t2}$ into account.

#### Spalart-Allmaras mixing layer compressibility correction
This version is the same as for the "baseline" version except that an additional term is included on the right hand side of the equation.

#### Spalart-Allmaras Edwards modification
This version is the same as for the "baseline" version, except that $f_{t2}$ is ignored, and two variables are redefined: the modified vorticity $\tilde{S}$ and $r$.

Note that this correction is not compatible with the negative SA model. Since there is no literature on how to handle both corrections simultaneously, the community decision was to prompt an error when trying to use them.

#### Spalart-Allmaras QCR-2000
The QCR correction modifies the turbulent stresses definition.

### Code structure
The approach that I decided to follow is:
The baseline/standalone SA class will have several templates parameters. Each of them will correspond to one of those variables that are subject to change in a SA correction/variation. Then, when creating the SA turbulence model, one just needs to replace each of these templates by the appropriate class.

For those corrections/variations that just introduce an additional source term, I think it is better to implement them using decorators. Hence multiple corrections/variations introducing source terms can be simultaneously used.

Here is a minimal example:

:bangbang: IMPORTANT: the following code served as a first draft and orientation. The final version has slightly changed to improve efficiency, reusability and readability. :bangbang:

```c++
template<class ft2_c, class ModVort_c>
class SA_Base : public CNumerics {
public:
  SA_Base() {}

  void computeResidual() final {
    ft2 = ft2_c::Compute()
    S_hat = ModVort_c::Compute(2.0,10)
  }
};

// the term "bsl" stands for baseline
class ft2_bsl     { static double Compute() { return 0.0; } }; 
class ft2_nonzero { static double Copmute() { return ...; } };
class ModVort_bsl { static double Compute(const double x, const int y){ return x*y; } };
class ModVort_Neg { static double Compute(const double x, const int y){ return x*y; } };
class ModVort_Edw { static double Compute(const double x, const int y){ return x+y; } };

/*-- Numerics preprocessing --*/
// From the config file (.cfg) determine which corrections/variations we have to use: ft2_variant, ModVort_variant, etc
// Create Frankenstein SA
if(ft2_variant==0 && ModVort_variant==0) numerics[i] = new SA_Base<ft2_bsl, ModVort_bsl>();
if(ft2_variant==0 && ModVort_variant==1) numerics[i] = new SA_Base<ft2_bsl, ModVort_Neg>();
if(ft2_variant==0 && ModVort_variant==2) numerics[i] = new SA_Base<ft2_bsl, ModVort_Edw>();
if(ft2_variant==1 && ModVort_variant==0) numerics[i] = new SA_Base<ft2_nonzero, ModVort_bsl>();
if(ft2_variant==1 && ModVort_variant==1) numerics[i] = new SA_Base<ft2_nonzero, ModVort_Neg>();
// so on

// Proceed to compute the turbulent sources residual as always
```

## Related Work
This PR closes the PR #1066.
This PR closes the issue #1364.
This PR will hopefully close the discussion #1403.

## PR Checklist


TestCases/serial_regression.py
```diff
@@ -703,6 +703,94 @@ def main():
...
+    turbmod_sa_qcr_rae2822.cfg_dir   = "turbulence_models/sa/rae2822"
+    turbmod_sa_qcr_rae2822.cfg_file  = "turb_SA_QCR_RAE2822.cfg"
+    turbmod_sa_qcr_rae2822.test_iter = 20
+    turbmod_sa_qcr_rae2822.test_vals = [-2.020123, -5.269330, 0.807147, 0.060499]
+    turbmod_sa_qcr_rae2822.su2_exec  = "SU2_CFD"
+    turbmod_sa_qcr_rae2822.timeout   = 1600
+    turbmod_sa_qcr_rae2822.new_output = True
+    turbmod_sa_qcr_rae2822.tol       = 0.00001
+    test_list.append(turbmod_sa_qcr_rae2822)
+
```
pcarruscag, line 793:
```suggestion
```

SU2_CFD/include/numerics/turbulent/turb_sources_new.hpp
```diff
@@ -0,0 +1,662 @@
...
+};
+/*
+ *  ResidualType<> ComputeResidual(const CConfig* config) final {
+ *    CommonVariables data;
+ *    ...
+ *    data.Ji = ...
+ *    ...
+ *    ft2_class::get(data, ft2, d_ft2);
+ */
+/* This way you can also dispense with templating "get". */
```
pcarruscag, line 334: I think this will work.

suargi: As @pcarruscag indicated: The structure I have implemented in  #3dd64f44e37b8b4ada9a31c5d450b722d3ef8554 for the turb_sources.hpp causes a cyclic dependency between the turbulence base and the variants classes. To break that one needs to template the `get` method in the variants classes instead of the class, for example the procedure he suggested in #2f60dbb4cae19aa15b9b6578043838f8eefd39b0

SU2_CFD/include/numerics/turbulent/turb_sources_new.hpp
```diff
@@ -143,520 +143,373 @@ class CSourceBase_TurbSA : public CNumerics {
...
+   template <class Base>
+   static void get(CommonVariables &model_var) {
 
-      dr = (Shat - ScalarVar_i[0] * dShat) * inv_Shat * inv_Shat * inv_k2_d2;
-      if (r == 10.0) dr = 0.0;
-      dg = dr * (1. + cw2 * (6.0 * pow(r, 5.0) - 1.0));
-      dfw = dg * glim * (1. - g_6 / (g_6 + cw3_6));
-      Jacobian_i[0] -= cw1 * (dfw * ScalarVar_i[0] + 2.0 * fw) * ScalarVar_i[0] / dist_i_2 * Volume;
+      using Base::PrimVar_Grad_i;
+      using Base::idx;
```
pcarruscag, line 264: Looking good @suargi :+1:, thanks for defining the common vars struct already.

If you template `get` on `Base` you need to pass something to `get` from which `Base` can be deduced, and the `using` declarations here will not give access to those variables.
Best solution IMO, do not template on base (or anything else to keep it simple) and pass what you need to `get`.

suargi: Thanks. In 123302d8acb96c34e672afb428129d1da3a8b4ca I have restructured the `get` methods to only pass those necessary variables instead of the class.

pcarruscag:
There is already a lot of nice work here in identifying the meaningful differences between different SA variants and reducing code duplication, so let's checkpoint this into develop before some other changes are made to turbulence models.
The alternative way to specify corrections/variants in the config can be done in a different PR.

suargi:
Really nice work @pcarruscag :+1: 
It looks good to me. I agree with

> There is already a lot of nice work here in identifying the meaningful differences between different SA variants and reducing code duplication, so let's checkpoint this into develop before some other changes are made to turbulence models. The alternative way to specify corrections/variants in the config can be done in a different PR.

TobiKattmann (approved):
Thanks @suargi (and @pcarruscag) for untangling these SA turb models :bouquet: 

I have just a few comments. overall everything looks quite tidy. Together with a precise documentation in https://github.com/su2code/su2code.github.io/pull/59 this is a great contribution :+1:

TestCases/parallel_regression.py
```diff
@@ -710,7 +710,7 @@ def main():
     turbmod_sa_comp_edw_rae2822.cfg_dir   = "turbulence_models/sa/rae2822"
     turbmod_sa_comp_edw_rae2822.cfg_file  = "turb_SA_COMP_EDW_RAE2822.cfg"
     turbmod_sa_comp_edw_rae2822.test_iter = 20
-    turbmod_sa_comp_edw_rae2822.test_vals = [-2.004687, 0.742306, 0.497310, -5.290769, 0.809485, 0.062036]
+    turbmod_sa_comp_edw_rae2822.test_vals = [-2.004685, 0.742307, 0.497311, -5.290750, 0.809487, 0.062045]
```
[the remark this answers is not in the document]
pcarruscag: Just noise I think.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -28,331 +26,567 @@
...
-   * \return Value of the intermittency.
-   */
-  inline su2double GetGammaBC(void) const final { return Gamma_BC; }
+    Omega::get(Vorticity_i, nDim, PrimVar_Grad_i + idx.Velocity(), var);
 
-  /*!
-   * \brief  ______________.
-   */
-  inline su2double GetDestruction(void) const final { return Destruction; }
+    /// TODO: Make this one of the template parameters?
```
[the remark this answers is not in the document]
pcarruscag: PR's are not easy to search either, I'll link back to this.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -28,331 +26,567 @@
...
-   * \param[in] val_nVar - Number of variables of the problem.
+   * \param[in] nDim - Number of dimensions of the problem.
    * \param[in] config - Definition of the particular problem.
    */
-  CSourceBase_TurbSA(unsigned short val_nDim, unsigned short val_nVar, const CConfig* config);
+  CSourceBase_TurbSA(unsigned short nDim, unsigned short, const CConfig* config)
+      : CNumerics(nDim, 1, config),
+        idx(nDim, config->GetnSpecies()),
+        rotating_frame(config->GetRotating_Frame()),
+        transition(config->GetKind_Trans_Model() == TURB_TRANS_MODEL::BC) {
```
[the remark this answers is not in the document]
pcarruscag: I also did not know but the LM model works in combination with SST, SA-LM and SST-BC do not "exist".

TobiKattmann:
hmm, [](https://github.com/su2code/SU2/pull/1413/commits/0f4977f4b761df87737ac283518870fb7cbbd466)[0f4977f](https://github.com/su2code/SU2/pull/1413/commits/0f4977f4b761df87737ac283518870fb7cbbd466) must have introduced sth that made the two (streamwise periodic) testcases fail. I think the defining feature here is, that they use the SST-turb model.

But looking at the mentioned commit, the changes look equal to me :thinking:

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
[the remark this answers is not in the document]
suargi: Since the plan is to close this pull request once it is merged, writing a Todo's list at the end does not seem to be the best approach (In my opinion). I would suggest to include such a Todo's list in #1487 under the Turbulence models section.

Besides, the complementary notes from @pcarruscag inside the code can be helpful for future PR.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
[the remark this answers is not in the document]
suargi: It seems that other transition models for SA might be implemented in a near future, see #1496. So, it would be a friendly correction to modify that line.

I will add a TODO annotation in the code.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -35,18 +35,19 @@
...
-  const su2double sigma = 2.0 / 3.0;
-  const su2double cb2 = 0.622;
-  const su2double cb2_sigma = cb2 / sigma;
-  const su2double cw1 = cb1 / k2 + (1 + cb2) / sigma;
-  const su2double cr1 = 0.5;
+  const su2double
+   cv1_3 = pow(7.1, 3.0),
+   k2    = pow(0.41, 2.0),
+   cb1   = 0.1355,
+   cw2   = 0.3,
```
pcarruscag, line 42: revert please, it is clang-format or nothing

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -1,16 +1,14 @@
...
+ * \brief Numerics classes for integration of source terms in turbulence problems.
  * \version 7.3.0 "Blackbird"
  *
  * SU2 Project Website: https://su2code.github.io
  *
  * The SU2 Project is maintained by the SU2 Foundation
  * (http://su2foundation.org)
  *
- * Copyright 2012-2022, SU2 Contributors (cf. AUTHORS.md)
+ * Copyright 2012-2021, SU2 Contributors (cf. AUTHORS.md)
```
pcarruscag, line 11:
```suggestion
 * Copyright 2012-2022, SU2 Contributors (cf. AUTHORS.md)
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -28,407 +26,762 @@
...
 };
 
+}  // namespace detail
+
+/* =============================================================================
+ * SPALART-ALLMARAS CLASSES
+ * ============================================================================*/
+
+/// TODO: Factory method to create combinations of the different variations based on the config.
+/// See PR #1413, the combinations exposed should follow https://turbmodels.larc.nasa.gov/spalart.html
```
pcarruscag, line 508: Main TODO for a future PR

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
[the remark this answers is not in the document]
TobiKattmann: > I will add a TODO annotation in the code.

:disappointed:

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
TobiKattmann: That was the reason to have a `const double` in front of all of those @suargi, in order to have clang-format not collapse these together

TestCases/parallel_regression.py
```diff
@@ -1428,7 +1428,7 @@ def main():
     sp_pinArray_3d_cht_mf_hf_tp.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_3d"
     sp_pinArray_3d_cht_mf_hf_tp.cfg_file  = "configMaster.cfg"
     sp_pinArray_3d_cht_mf_hf_tp.test_iter = 30
-    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [-13.380025, -7.476945, -7.025285, -0.009675, 99.879812, 4.1920e+02]
+    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [-13.380430, -7.476945, -7.025285, -0.009675, 99.879812, 4.1920e+02]
```
TobiKattmann, line 1431: also noise? if yes, why would the mentioned commit create such noise? Variable initialization in the initializer list or ctor should to my humble knowledge not change any computation, nor should the const-ing do so.

Please understand that I am particularly careful when it comes to these tests

TobiKattmann (changes requested):

TestCases/parallel_regression.py
pcarruscag: That is a very good way to return the help I've given to your PR's, good job mate

TobiKattmann: I do have an actual interest in why the regression test values here change. And I am unable to explain why that is here. I've failed a ton of reg tests because I unknowingly interfered with that code, so I want to avoid a simple problem here by "forcing" nothing more than a second thought. If your verdict is that this change is safe then I am more than fine with that. 

Non-obvious reg-test value changes require at least a comment if not an explanation. That, in my opinion, is their very purpose.

TobiKattmann (approved):

TestCases/parallel_regression.py
TobiKattmann: See https://github.com/su2code/SU2/pull/1413#discussion_r818738434 for the noise creator :) 

Btw: 
> [](https://github.com/su2code/SU2/pull/1413/commits/0f4977f4b761df87737ac283518870fb7cbbd466)hmm, [0f4977f](https://github.com/su2code/SU2/pull/1413/commits/0f4977f4b761df87737ac283518870fb7cbbd466) must have introduced sth
my comment from above was wrong. In fact [e9421f5](https://github.com/su2code/SU2/pull/1413/commits/e9421f56886e067f654b6e82c9a35927d96259f5) introduced the change. This is because the ordering of commits is commit date and not the date when it was pushed. So naturally one would think the commits here on github represents a timeline of changes on github itself, but that is not the case. Not that this is a crazy insight, but maybe to be kept in mind when searching a bad commit (one should use git bisect for that anyway)

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -28,407 +26,762 @@
...
+
+      su2double StrainMag = StrainMag_i;
+
+      if (using_uq) {
+        ComputePerturbedRSM(nDim, Eig_Val_Comp, uq_permute, uq_delta_b, uq_urlx, PrimVar_Grad_i + idx.Velocity(),
+                            Density_i, Eddy_Viscosity_i, ScalarVar_i[0], MeanPerturbedRSM);
+        StrainMag = PerturbedStrainMag(ScalarVar_i[0]);
+      }
+
+      su2double pk = Eddy_Viscosity_i * pow(StrainMag, 2) - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
```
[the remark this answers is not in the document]
pcarruscag: Thank you for tracing that.

PR #1388, opened 2021-10-02
Species transport solver (passive for now)

TobiKattmann:
## Proposed Changes
Hi all,

This PR implements a passive species transport solver. The passive means, that the mass fraction at each location does not influence any fluid-properties and thereby the mean flow (which reminds me that this is a good sanity check to compare mean flow residuals).

A new `CSpeciesSolver` and `CSpeciesVariable` were added based on the `CScalar*` things done in #1330 . Additionally `species_convection/diffusion/sources` were added just as it is the case for the turbulence solver (both based on the general scalar* framework)

Please note that prob ~1000 lines are cfg's, README's, helping scripts. 

## Related Work
Preliminary work was done in #1330 with creating a base class for scalar transport equations.

#1223 and #1332 already mostly implement the here shown species solver and big chunks were copied over from there, but they are not based on the new CScalarSolver. In order to chunk the work for these branches this PR will not contain any species mixing properties or TableLookUps. 

After this PR #1332 will be tackled next. Whether we try to merge develop then into those branches or copy stuff over in a new branch will be seen when we get there.

## PR Checklist

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #516, opened 2018-03-20
Feature fileprofile

economon:
*(Note: I am prototyping a new format for a PR template with this submission. We can set a default template in the repo to help guide new contributors and to make future code reviews easier. Please also comment on this general format/idea or share any suggestions to improve.)*

## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
 
This PR is a joint effort with @clarkpede for specifying inlet profiles from a file within the C++ code. For now, it is nearest neighbor, but interpolation can be added in the future. If no file is present, a template file will be generated for your modification (ASCII file, one row per DOF on each inlet marker). It works in parallel and for any number of inlet markers. It operates similarly to restarts: there is a single file reader in CSolver, a single file writer in COutput, and solver-specific load functions in the child solver classes. Therefore, this structure can be expanded in the future for data on any marker type in any solver.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

Resolves #467 
Interacts with work by @tobadavid @oleburghardt 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [x] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution.

PR #1427, opened 2021-11-07
Fixes to inlet files

TobiKattmann:
Hi all,

I noticed problems when using multiple inlet files: At the template values were mixed between the inlets because there is a (in my eyes) superfluous `sort` done. So the problem is just with the template-writing and would not happen if alphabetical and nMarker_all order between the inlet marker_tags were identical. This PR fixes this, at least for the tests I did over at #1388 where I will also add a inc+turb+2species-with-2-inlets  reg test (and hopefully also non-dimensional).

Inlet file non-dimensionalization: The inlet file writes dimensional values for the mean flow, and the values that are read are non-dimensionalized in the BC_Inlet routine itself -> I personally find this approach quite good, but I dont believe the SA+SST are adhering to that. See also [this cfd-online issue](https://www.cfd-online.com/Forums/su2/239313-inlet-profile-sst-nondimensional-treatment.html). I will try to fix this here as well.
- nu tilde is non-dimensionalized by multiplying with densityRef/viscosityRef (which I guess from the code)
- TKE by multiplying with 1/VelocityRef^2  (which I guess from the code, and would fit with Feng Liu Non-Dim paper i have)
- Omega ~~by multiplying with length_Ref/VelocityRef (which would fit with Feng Liu) or~~ by multiplying with ViscRef/(rhoRef*VelocityRef^2)
Any info on this is appreciated.
My approach would be non-dimensionalize in the BC_Inlet routine for Turb as well to stay consistent with the mean flow, For the mean flow it has to be done like  that because the meaning of the values is determined in the inlet routine. Otherwise one could do all that consistently a bit earlier.

I have to remember to add a little hint to the laminar step tutorial to state that the inlet file should be dimensional

## Related Work
https://github.com/su2code/SU2/pull/516 initial implementation

## PR Checklist


SU2_CFD/src/CMarkerProfileReaderFVM.cpp
```diff
@@ -357,9 +357,8 @@ void CMarkerProfileReaderFVM::MergeProfileMarkers() {
       }
     }
 
-    /*--- Sort and remove the duplicate profile marker strings. ---*/
+    /*--- Remove the duplicate profile marker strings. From 1 per point to 1 per marker. ---*/
 
-    sort(profileTags.begin(), profileTags.end());
```
TobiKattmann, line 362: so this destroys the order that was always given implicitly by the loop:
```c++
for (unsigned short iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
  if (config->GetMarker_All_KindBC(iMarker) == markerType) {
```
That loop is consequently used during the template writing as well

pcarruscag:
:+1: for doing it in BC_Inlet

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -628,12 +628,20 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
...
 
+      /*--- Non-dimensionalize Inlet_TurbVars if Inlet-Files are used. ---*/
+      su2double Inlet_Vars[MAXNVAR];
+      Inlet_Vars[0] = Inlet_TurbVars[val_marker][iVertex][0];
+      Inlet_Vars[1] = Inlet_TurbVars[val_marker][iVertex][1];
+      if (config->GetInlet_Profile_From_File()) {
+        Inlet_Vars[0] *= 1 / (config->GetVelocity_Ref() * config->GetVelocity_Ref());
+        Inlet_Vars[1] *= config->GetViscosity_Ref() / (config->GetDensity_Ref() * config->GetVelocity_Ref() * config->GetVelocity_Ref());
+      }
+
```
TobiKattmann, line 639: Ok now this works but maybe it is not the most performant or elegant solution.

The test I made is to write the example_inlet (by not having an inlet in the first place) and run a simulation with that. Then take the same config without inlet file and run another simulation --> These two simulation should be identical. For SA, the diff between the history files is empty and for SST I see differences in the 9th digit  (the last in that case) popping up after 150 iterations which then accumulate in the residual. I would guess these are floating point errors but the analytical computation is the same.

TobiKattmann:
From my perspective this is finished. As mentioned above I will add a testcase with Mean-flow+SST+2Species for 2 inlets over at #1388 once this is merged.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -628,12 +628,20 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
 
       conv_numerics->SetPrimitive(V_domain, V_inlet);
 
+      /*--- Non-dimensionalize Inlet_TurbVars if Inlet-Files are used. ---*/
+      su2double Inlet_Vars[MAXNVAR];
+      Inlet_Vars[0] = Inlet_TurbVars[val_marker][iVertex][0];
+      Inlet_Vars[1] = Inlet_TurbVars[val_marker][iVertex][1];
+      if (config->GetInlet_Profile_From_File()) {
+        Inlet_Vars[0] *= 1 / (config->GetVelocity_Ref() * config->GetVelocity_Ref());
```
pcarruscag, line 636: Trust issues?
```suggestion
        Inlet_Vars[0] /= config->GetVelocity_Ref() * config->GetVelocity_Ref();
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -628,12 +628,20 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
...
+        Inlet_Vars[1] *= config->GetViscosity_Ref() / (config->GetDensity_Ref() * config->GetVelocity_Ref() * config->GetVelocity_Ref());
```
pcarruscag, line 637:
```suggestion
        Inlet_Vars[1] *= config->GetViscosity_Ref() / (config->GetDensity_Ref() * pow(config->GetVelocity_Ref(), 2));
```

pcarruscag (approved):

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
TobiKattmann: Well, I already got confused quite a bit so I thought I'll just think in multiplication of dimensional values to non-dimensionalize .. I'll edit your suggestion to use the pow(,2) here as well

PR #1438, opened 2021-11-13
fix mistake in axisymmetric turbulent source production term

FlorianDm:
This fixes a small mistake I discovered in code that I added with https://github.com/su2code/SU2/pull/1195 (I just learned the hard way that it is better to use common/general forms of equations instead of trying to simplify/factorise them as much as possible). 

In case anyone wants to verify, the sources used are in https://github.com/su2code/SU2/pull/1195, the terms are found by subtracting the cartesian formulation from the cylindrical one and here is the simplification/factorisation bit:

IMG_20211113_110505

Sorry!

### DIFF
--- a/TestCases/hybrid_regression.py
+++ b/TestCases/hybrid_regression.py
@@ -227,7 +227,7 @@ def main():
     axi_rans_air_nozzle.cfg_dir   = "axisymmetric_rans/air_nozzle"
     axi_rans_air_nozzle.cfg_file  = "air_nozzle.cfg"
     axi_rans_air_nozzle.test_iter = 10
-    axi_rans_air_nozzle.test_vals = [-12.094009, -6.643017, -8.900656, -2.393294]
+    axi_rans_air_nozzle.test_vals = [-12.094331, -6.628118, -8.781813, -2.395015]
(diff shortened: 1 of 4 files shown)


FlorianDm:
restart file update https://github.com/su2code/TestCases/pull/85

pcarruscag (approved):

TestCases/hybrid_regression.py
```diff
@@ -227,7 +227,7 @@ def main():
     axi_rans_air_nozzle.cfg_dir   = "axisymmetric_rans/air_nozzle"
     axi_rans_air_nozzle.cfg_file  = "air_nozzle.cfg"
     axi_rans_air_nozzle.test_iter = 10
-    axi_rans_air_nozzle.test_vals = [-12.094009, -6.643017, -8.900656, -2.393294]
+    axi_rans_air_nozzle.test_vals = [-12.094331, -6.628118, -8.781813, -2.395015]
```
pcarruscag, line 230:
```suggestion
    axi_rans_air_nozzle.test_vals = [-12.093575, -6.630426, -8.798725, -2.399130]
```

PR #1445, opened 2021-11-17
Fix compilation with GCC 4.8.5

pcarruscag:
So this:
```c++
template <class T> \\ <--- This...
class A {
  su2double a[4] = {0.0}; \\ <--- ...with = {0.0};
};
```
seems to be the problem.

### DIFF
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -334,9 +334,9 @@ private:
 
   su2double kAmb, omegaAmb;
 
-  su2double Residual[2],
-  *Jacobian_i[2] = {nullptr},
-  Jacobian_Buffer[4] = {0.0}; /// Static storage for the Jacobian (which needs to be pointer for return type).
+  su2double Residual[2];
+  su2double* Jacobian_i[2];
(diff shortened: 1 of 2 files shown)


TobiKattmann:
Nice job 👍 , I guess you checked with compiler Explorer :) 

I am confident that this will also solve [this persons cfd online issue](https://www.cfd-online.com/Forums/su2/239313-inlet-profile-sst-nondimensional-treatment.html)

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -334,9 +334,9 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
 
   su2double kAmb, omegaAmb;
 
-  su2double Residual[2],
-  *Jacobian_i[2] = {nullptr},
-  Jacobian_Buffer[4] = {0.0}; /// Static storage for the Jacobian (which needs to be pointer for return type).
+  su2double Residual[2];
+  su2double* Jacobian_i[2];
+  su2double Jacobian_Buffer[4]; /// Static storage for the Jacobian (which needs to be pointer for return type).
```
TobiKattmann, line 339: Maybe do the initialization by hand in the constructor then? Or is it not necessary

PR #1449, opened 2021-11-19
Collect Turb output from Comp/Inc/Nemo in CFlowOutput

TobiKattmann:
Hi,

Comp/Inc/Nemo output all had the same RMS_TKE etc output -> now this is done once in CFlowOutput  and Comp/Inc/NemoOutput only call sth like `SetHistoryOutput_Turb` in their respective `SetHistoryOutput`.

this is just a cleanup and shouldnt change too much.

I do this in anticipation of addition of Species related output which I dont want to repeat. So #1388 will use the same concept

## PR Checklist


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -580,6 +580,153 @@ void CFlowOutput::SetAnalyzeSurface(const CSolver *solver, const CGeometry *geom
...
+      if (config->GetKind_SlopeLimit_Turb() != NO_LIMITER) {
+        AddVolumeOutput("LIMITER_TKE", "Limiter_TKE", "LIMITER", "Limiter value of turb. kinetic energy");
+        AddVolumeOutput("LIMITER_DISSIPATION", "Limiter_Omega", "LIMITER", "Limiter value of dissipation rate");
+      }
+      break;
+    case TURB_MODEL::NONE:
+      break;
+  }
+
+  if (config->GetKind_Turb_Model() != TURB_MODEL::NONE) {
```
TobiKattmann, line 671:
```suggestion
  if (config->GetKind_Solver() == RANS || config->GetKind_Solver() == INC_RANS) {
```

TobiKattmann: NEMO did not have EddyVisc. Therefore I wont include it here

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -580,6 +580,153 @@ void CFlowOutput::SetAnalyzeSurface(const CSolver *solver, const CGeometry *geom
...
+      }
+      break;
+    case TURB_MODEL::SST: case TURB_MODEL::SST_SUST:
+      AddVolumeOutput("TKE", "Turb_Kin_Energy", "SOLUTION", "Turbulent kinetic energy");
+      AddVolumeOutput("DISSIPATION", "Omega", "SOLUTION", "Rate of dissipation");
+      AddVolumeOutput("RES_TKE", "Residual_TKE", "RESIDUAL", "Residual of turbulent kinetic energy");
+      AddVolumeOutput("RES_DISSIPATION", "Residual_Omega", "RESIDUAL", "Residual of the rate of dissipation");
+      if (config->GetKind_SlopeLimit_Turb() != NO_LIMITER) {
+        AddVolumeOutput("LIMITER_TKE", "Limiter_TKE", "LIMITER", "Limiter value of turb. kinetic energy");
+        AddVolumeOutput("LIMITER_DISSIPATION", "Limiter_Omega", "LIMITER", "Limiter value of dissipation rate");
```
pcarruscag, line 664: This is great but we probably need a finer-grained division because of the ordering problem, otherwise the restart files are not going to work

TobiKattmann: Limiters are not relevant for restarts, right 🥟

pcarruscag: They are not, but I admit I like the residuals close to the residuals and the limiters close to the limiters :grimacing:

TobiKattmann: Intermittency for transition model is also the solution to a transport equation (right?) ... so that should adhere to a restart order as well ... but I would need to check whether that is even written.
Man do I have no clue of transition modelling 🐔

TobiKattmann: Transition would use CTurbSolver::LoadRestart and would therefore read crap anyway atm .. maybe catch Trans-restarts in the cfg-setpost

pcarruscag: No, intermittency is a derived quantity

TobiKattmann: c050e57ce3c384e366d590239fb2c1d697192f0d adds the distinction for the volume TurbSolution and the rest (res+limiters)

TobiKattmann: oc I always knew that the SA-BC model uses intermittency and does not require (to my knowledge) any more than the SA's nu_tilde. And as a certified transition expert I hope we have the revised SA-BCM implemented and not the original SA-BC ... the NASA-dudes even added a warning in red color(!) on [TMR](https://turbmodels.larc.nasa.gov/sa-bc_1eqn.html)

For the Langter-Mentry we would need this gamma an Re_theta value for restart but we have `SU2_MPI::Error("The LM transition model is under maintenance.", CURRENT_FUNCTION);` anyway

So I think transition model restarts are not an issue for now

TobiKattmann: > They are not, but I admit I like the residuals close to the residuals and the limiters close to the limiters 😬

But... all those deletd lines 😢 ... fine I'll do the family reunion (tomorrow)

SU2_CFD/src/output/CNEMOCompOutput.cpp
```diff
@@ -271,6 +229,9 @@ void CNEMOCompOutput::SetHistoryOutputFields(CConfig *config){
     AddHistoryOutput("DEFORM_RESIDUAL", "DeformRes", ScreenOutputFormat::FIXED, "DEFORM", "Residual of the linear solver for the mesh deformation");
   }
 
+  AddHistoryOutputFields_Turb(config);
+
```
TobiKattmann, line 233:
```suggestion
```

TobiKattmann:
> They are not, but I admit I like the residuals close to the residuals and the limiters close to the limiters 

I fear, that also applies to the History output

pcarruscag:
But that is processed in groups, it should not have the same problem :thinking:

TobiKattmann:
> But that is processed in groups, it should not have the same problem 🤔

I mean... it can still stay like that, or should I revert 😬

pcarruscag:
It's ok, I'll have a look today or tomorrow to see if there is some way to recover those precious deleted lines.

TobiKattmann:
Or I revert the History split-up and we call it a weekend

pcarruscag (approved):
Registering the outputs defined the output order, but data can be loaded in any order.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -598,101 +598,78 @@ void CFlowOutput::AddHistoryOutputFields_TurbRMS_RES(const CConfig* config) {
...
 
-void CFlowOutput::LoadHistoryData_TurbLinsol(const CConfig* config, const CSolver* const* solver) {
-  if (config->GetKind_Turb_Model() != TURB_MODEL::NONE) {
-    SetHistoryOutputValue("LINSOL_ITER_TURB", solver[TURB_SOL]->GetIterLinSolver());
-    SetHistoryOutputValue("LINSOL_RESIDUAL_TURB", log10(solver[TURB_SOL]->GetResLinSolver()));
+void CFlowOutput::LoadHistoryData_Turb(const CConfig* config, const CSolver* const* solver) {
+  switch (config->GetKind_Turb_Model()) {
+    case TURB_MODEL::NONE:
+      /*--- Early return if there is no turbulence model in use. ---*/
+      return;
```
TobiKattmann, line 649: nice 👍

TobiKattmann (commented):
Big Thanks for the extensive update and the research what actually matters for the order! And of course kudos for the additional 🔴 ;)

I have no more comments or updates to make, so in case you're good I guess we can merge

PR #1461, opened 2021-12-02
Feature species - more adjoint parts

TobiKattmann:
Hi,

Some necessary changes for the species transport solver to work nicely with the Discrete adjoint solver. So it is the same structure that was Introduced for CFlowOutput but now it is CAdjFlowOutput.
- generalize turb+species output for Comp/Inc/NEMO away into a parent class to avoid duplication (Done)
- Make More OF available such that one can actually use the DA solver
- Add a Gradient validation + Optimization FADO template plus add DA regression tests (or tutorials 🤔 I think I will make a tutorial thing for this)
-  Some smaller fixes to the previous PR

## Related Work
directly following #1388 where the majority of adjoint work for the species solver was already done

## PR Checklist


TobiKattmann:
> generalize turb+species output for Comp/Inc/NEMO away into a parent class to avoid duplication

Actually I just noticed  there is no `CAdjNEMOOutput` or similar. Not sure @WallyMaier , but is the NEMO-tribe not doing in adjoints?

SU2_CFD/src/output/CAdjFlowOutput.cpp
```diff
@@ -0,0 +1,315 @@
...
+  if (!frozen_visc) {
+    switch (turb_model) {
+      case TURB_MODEL::SA:
+      case TURB_MODEL::SA_NEG:
+      case TURB_MODEL::SA_E:
+      case TURB_MODEL::SA_COMP:
+      case TURB_MODEL::SA_E_COMP:
+        /// DESCRIPTION: Root-mean square residual of the adjoint nu tilde.
+        AddHistoryOutput("RMS_ADJ_NU_TILDE", "rms[A_nu]", ScreenOutputFormat::FIXED, "RMS_RES",
+                         "Root-mean square residual of the adjoint nu tilde.", HistoryFieldType::RESIDUAL);
```
TobiKattmann, line 48: this is the clang-format style btw

WallyMaier:
@TobiKattmann I am actually developing the adjoint stuff for nemo right now! Though there are some recording issues....

TobiKattmann:
> @TobiKattmann I am actually developing the adjoint stuff for nemo right now! Though there are some recording issues....

Ok if this PR gets merged then there will be a `CAdjFlowOutput` class between `COutput` and e.g. `CNEMOAdjOutput` (or to be more precise: it is for inc and comp and it should be done for NEMO as well). That class provides output for the turbulence solvers and the separate species solver. So looking at `CAdjFLowIncOutput` then will show the methods called and that should be a straight forward change in  `CNEMOAdjOutput

SU2_CFD/src/output/CAdjFlowOutput.cpp
```diff
@@ -0,0 +1,317 @@
...
+ */
+
+#include "../../include/output/CAdjFlowOutput.hpp"
+
+#include "../../include/solvers/CSolver.hpp"
+
+CAdjFlowOutput::CAdjFlowOutput(CConfig* config, unsigned short nDim) : COutput(config, nDim, false) {
+  turb_model = config->GetKind_Turb_Model();
+  cont_adj = config->GetContinuous_Adjoint();
+  frozen_visc = (config->GetFrozen_Visc_Disc() && !cont_adj) || (config->GetFrozen_Visc_Cont() && cont_adj);
```
pcarruscag, line 35: Initializer list please and make them const

SU2_CFD/src/output/CAdjFlowOutput.cpp
```diff
@@ -0,0 +1,317 @@
...
+}
+
+void CAdjFlowOutput::AddHistoryOutputFields_AdjScalarRMS_RES(const CConfig* config) {
+  if (!frozen_visc) {
+    switch (turb_model) {
+      case TURB_MODEL::SA:
+      case TURB_MODEL::SA_NEG:
+      case TURB_MODEL::SA_E:
+      case TURB_MODEL::SA_COMP:
+      case TURB_MODEL::SA_E_COMP:
```
pcarruscag, line 45: This shall not pass :mage_man: 
I'll help clean this up when I have a minute, I'm thinking:
```c++
inline TURB_MODEL TurbModelFamily(TURB_MODEL model) {
  switch (model) {
    case TURB_MODEL::SA:
    case TURB_MODEL::SA_NEG:
    case TURB_MODEL::SA_E:
    case TURB_MODEL::SA_COMP:
    case TURB_MODEL::SA_E_COMP:
      return TURB_MODEL::SA;
    case TURB_MODEL::SST:
    case TURB_MODEL::SST_SUST:
      return TURB_MODEL::SST;
    case TURB_MODEL::NONE:
      return TURB_MODEL::NONE;
  }
}
```
in option_structure close to the enum.

TobiKattmann: clever

SU2_CFD/src/output/CAdjFlowOutput.cpp
TobiKattmann: yep, had that already on my list

TobiKattmann: And here c2e652a it is

Common/include/option_structure.hpp
```diff
@@ -929,6 +929,34 @@ static const MapType<std::string, TURB_MODEL> Turb_Model_Map = {
...
+    case TURB_MODEL::SA_NEG:
+    case TURB_MODEL::SA_E:
+    case TURB_MODEL::SA_COMP:
+    case TURB_MODEL::SA_E_COMP:
+      return TURB_FAMILY::SA;
+    case TURB_MODEL::SST:
+    case TURB_MODEL::SST_SUST:
+      return TURB_FAMILY::KW;
+  }
+  return TURB_FAMILY::NONE;
```
TobiKattmann, line 957: What do you need this additional return for? In case a new TURB_MODEL (e.g. k-eps) is added the `enum class` forces the person who writes the turb model to handle this ... of course this extra catch helps against a lazy `default: break;` fix. In case there is another reason let me know

TobiKattmann (commented):
Really nice 👍 puts this logic into one place and avoids a bunch of work for future changes

Common/include/option_structure.hpp
pcarruscag: Compiler warning otherwise "control may reach the end of function with return type" (loosely paraphrasing)

SU2_CFD/src/output/CAdjFlowOutput.cpp
TobiKattmann: 5b3f7d3

Common/include/option_structure.hpp
TobiKattmann: Just to reassure my understanding here: The compiler warning is "wrong" in this special case with what i state above - without a `default break` the switch of an `enum class`, where each option is followed by a return, cannot end without a return. But I guess that is why its a just warning ...

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -130,8 +130,17 @@ void CScalarSolver<VariableType>::Upwind_Residual(CGeometry* geometry, CSolver**
                                                   CNumerics** numerics_container, CConfig* config,
                                                   unsigned short iMesh) {
   const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
-  const bool muscl = config->GetMUSCL_Turb();
-  const bool limiter = (config->GetKind_SlopeLimit_Turb() != NO_LIMITER);
+
+  bool muscl;
+  bool limiter;
```
pcarruscag: Might be better to store during construction.

TobiKattmann: 07b1a8c like so?

TobiKattmann:
I will add surface avg species_0 as another Objective function but that would be it from my side concerning this PR

TobiKattmann:
So the merging process would of course be:
1. merge tutorials (+website)
2. revert regression.yml changes
3. merge this PR 
... in case people are happy with it

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -79,22 +79,6 @@ CScalarSolver<VariableType>::~CScalarSolver() {
...
-    muscl = config->GetMUSCL_Turb();
-    limiter = (config->GetKind_SlopeLimit_Turb() != NO_LIMITER) &&
-              (config->GetInnerIter() <= config->GetLimiterIter());
-  }
-  else if (SolverName == "SPECIES") {
-    implicit = (config->GetKind_TimeIntScheme_Species() == EULER_IMPLICIT);
-    muscl = config->GetMUSCL_Species();
-    limiter = (config->GetKind_SlopeLimit_Species() != NO_LIMITER) &&
-              (config->GetInnerIter() <= config->GetLimiterIter());
-  }
```
pcarruscag, line 82:
```suggestion
  const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);

```

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -129,10 +113,6 @@ template <class VariableType>
 void CScalarSolver<VariableType>::Upwind_Residual(CGeometry* geometry, CSolver** solver_container,
                                                   CNumerics** numerics_container, CConfig* config,
                                                   unsigned short iMesh) {
-  const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
-  const bool muscl = config->GetMUSCL_Turb();
-  const bool limiter = (config->GetKind_SlopeLimit_Turb() != NO_LIMITER);
-
   /*--- Only reconstruct flow variables if MUSCL is on for flow (requires upwind) and turbulence. ---*/
```
pcarruscag, line 116:
```suggestion
  const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);

  /*--- Only reconstruct flow variables if MUSCL is on for flow (requires upwind) and turbulence. ---*/
```

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -475,7 +455,6 @@ template <class VariableType>
 void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSolver** solver_container, CConfig* config,
                                                        unsigned short iRKStep, unsigned short iMesh,
                                                        unsigned short RunTime_EqSystem) {
-  const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
   const bool first_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_1ST);
```
pcarruscag, line 458:
```suggestion
  const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
  const bool first_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_1ST);
```

SU2_CFD/src/solvers/CSpeciesSolver.cpp
```diff
@@ -390,7 +396,6 @@ void CSpeciesSolver::BC_Inlet(CGeometry* geometry, CSolver** solver_container, C
       LinSysRes.AddBlock(iPoint, residual);
 
       /*--- Jacobian contribution for implicit integration ---*/
-      const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
```
pcarruscag, line 393: revert all of these please, we want to be able to switch between explicit and implicit dynamically.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -37,6 +37,11 @@ CTurbSolver::CTurbSolver(CGeometry* geometry, CConfig *config, bool conservative
   : CScalarSolver<CTurbVariable>(geometry, config, conservative) {
   /*--- Store if an implicit scheme is used, for use during periodic boundary conditions. ---*/
   SetImplicitPeriodic(config->GetKind_TimeIntScheme_Turb() == EULER_IMPLICIT);
+
+  implicit = (config->GetKind_TimeIntScheme_Turb() == EULER_IMPLICIT);
```
pcarruscag, line 41: having "implicit" has member is starting to sound error prone.
Keep muscl and limiter but get implicit where it is needed via config->GetKind_TimeIntScheme() as before

TobiKattmann: Alright. But a differentiation between `GetKind_TimeIntScheme_Turb/Species()` has to be made, right? I currently do not see a major reason why the Turb and Species system should not be solvable with a different time integration scheme. Any thoughts

pcarruscag: That is why we have things like:
```
config[val_iZone]->SetGlobalParam(config[val_iZone]->GetKind_Solver(), RUNTIME_SPECIES_SYS);
```
To then be able to use the generic `GetKind_` without Turb/Species in "runtime function", it is only during construction that the specifics are needed.
So in fact we should also be able to remove the muscl and limiter member vars.

TobiKattmann: I apologize if I do not understand it, but I am not sure whether you suggestion solves the problem.

**The Problem**
We have functions for e.g. the time integration scheme:
```c++
GetKind_TimeIntScheme_Turb/Species
GetKind_TimeIntScheme_Turb/Species
GetKind_TimeIntScheme_Turb/Species
```

TobiKattmann: 5fc376e implements your suggestion, Now limiter and muscl are local const bools which use the generic function 👍

SU2_CFD/include/solvers/CScalarSolver.inl
TobiKattmann: 54a0c69

SU2_CFD/include/solvers/CScalarSolver.inl
TobiKattmann: 54a0c69

SU2_CFD/include/solvers/CScalarSolver.inl
TobiKattmann: 54a0c69

SU2_CFD/src/solvers/CSpeciesSolver.cpp
TobiKattmann: 54a0c69 Again I apologize to not use your suggestion but making it locally, gives me at least the chance to check compilation before pushing. I just checked how can add you as a [commit contributer](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) locally

TobiKattmann: just gave it  a test 04533a5 there. So now I can add you as a co author whenever I dont feel like using the github suggestion. Credit where Credit's due 💵

SU2_CFD/src/solvers/CTurbSolver.cpp
TobiKattmann: the `Preprocessing` routines actually get a RUNTIME_SYS as input so there it is no prob because one could call the SetGolbalParams it in the Preprocessing routine itself ... but the Turbulence solvers compute muT in Postprocessing which do not get  a `RUNTIME_SYS ` as input... so I guess right now limiter, turb and implicit are not used in the Postprocess routines of the turb models but that might become a nasty bug at some point

SU2_CFD/src/solvers/CSpeciesSolver.cpp
pcarruscag: lol, thanks you don't have to do that, well I guess it's useful if it turns out to be a bug xD

Common/include/option_structure.hpp
```diff
@@ -1588,6 +1618,8 @@ static const MapType<std::string, ENUM_OBJECTIVE> Objective_Map = {
   MakePair("SURFACE_MOM_DISTORTION", SURFACE_MOM_DISTORTION)
   MakePair("SURFACE_SECOND_OVER_UNIFORM", SURFACE_SECOND_OVER_UNIFORM)
   MakePair("SURFACE_PRESSURE_DROP", SURFACE_PRESSURE_DROP)
+  MakePair("SURFACE_SPECIES_0", SURFACE_SPECIES_0)
+  MakePair("SURFACE_SPECIES_VARIANCE", SURFACE_SPECIES_VARIANCE)
```
pcarruscag, line 1622: Something entirely different, historyMap.py
You know, to update shape_optimization in its inoperability

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -459,7 +460,7 @@ void CHeatSolver::Viscous_Residual(CGeometry *geometry, CSolver **solver_contain
   su2double laminar_viscosity, Prandtl_Lam, Prandtl_Turb, eddy_viscosity_i, eddy_viscosity_j,
             thermal_diffusivity_i, thermal_diffusivity_j, Temp_i, Temp_j;
 
-  const bool turb = ((config->GetKind_Solver() == MAIN_SOLVER::INC_RANS) || (config->GetKind_Solver() == MAIN_SOLVER::DISC_ADJ_INC_RANS));
```
TobiKattmann, line 462: For some reason my IDE decided to change these without me knowing (I must have overlooked it in QC `git diff --staged` ) ... why

Common/include/option_structure.hpp
TobiKattmann: I will take a look after lunch how adding CFlowOutput to that list changes things

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1565,6 +1565,7 @@ void CEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_con
 
 void CEulerSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iMesh,
                                  unsigned short iRKStep, unsigned short RunTime_EqSystem, bool Output) {
+  config->SetGlobalParam(config->GetKind_Solver(), RunTime_EqSystem);
```
pcarruscag, line 1568: This breaks the "switching between explicit and implicit dynamically" I was telling you about.
Because we are setting a "custom" explicit mode before calling the solver in CNewtonIntegration::ComputeResiduals and in CMultiGridIntegration::MultiGrid_Cycle, but then the solver discards it.
But at the same time I like your solution as it allows base solver classes to do the right thing using the general MUSCL, Limiter, etc., and opens room for us to simplify the FluidIteration...

Maybe let's remove these SetGlobalParam from flow solvers (that may use MG or NK) for now while I think of something, but keep them in scalar solvers to allow the restarts to work properly.

pcarruscag (approved):

TestCases/parallel_regression.py
```diff
@@ -1628,7 +1616,7 @@ def main():
     species_passive_val.cfg_dir   = "species_transport/passive_transport_validation"
     species_passive_val.cfg_file  = "passive_transport.cfg"
     species_passive_val.test_iter = 50
-    species_passive_val.test_vals = [-16.559189, -16.315116, -16.908670, -4.316833, 10.000000, -4.523292, 8.000000, -5.173152]
+    species_passive_val.test_vals = [-16.559189, -16.315116, -16.908670, -4.257599, 10.000000, -4.523292, 8.000000, -5.193350]
```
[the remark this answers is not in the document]
TobiKattmann: Yes, before the MUSCL_SPECIES had no effect and this was the only case that had MUSCL_SPECIES on.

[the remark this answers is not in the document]
TobiKattmann: And with MUSCL the results look reasonable and sometimes even visually better

SU2_CFD/src/solvers/CEulerSolver.cpp
TobiKattmann: 👍 57a87e5 removes the flow solver instances. What about the SetGlobalParam CDiscAdjSolver, do you foresee any issues there

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -572,6 +572,7 @@ void CDiscAdjSolver::SetSurface_Sensitivity(CGeometry *geometry, CConfig *config
 
 void CDiscAdjSolver::Preprocessing(CGeometry *geometry, CSolver **solver_container, CConfig *config, unsigned short iMesh,
                                    unsigned short iRKStep, unsigned short RunTime_EqSystem, bool Output) {
+  config->SetGlobalParam(config->GetKind_Solver(), RunTime_EqSystem);
```
pcarruscag, line 575: Should be fine I think

TobiKattmann: Awesome, then I'll start merging the PR's

Issue #1463, opened 2021-12-04
Writing solution data in steady state cases without overwriting them


thunderbird-98:
I am running an FSI steady-state case with CFD and FEA coupled. I want to write every outer iteration without the file getting overwritten.

"OUTPUT_WRT_FREQ" is overwriting the output files

pcarruscag:
Hi, we don't have an option to "keep all files", the intended behaviour in steady state is to have only the last solution.
You could emulate what you want via the python wrapper by running some steps and them saving the files manually.

bigfooted:
We have introduced something like this in one of the feature branches (feature_flamelet).
We have introduced an additional keyword WRT_SOL_OVERWRITE= NO
Instead of overwriting the restart and paraview files, it will append the iteration number to the filename each time it writes the files. Is this what you need? 

Maybe we can accelerate this and put it into develop independent of this feature branch.

thunderbird-98:
Yes that's exactly what I need. Thanks for your help

pcarruscag:
Will be available in the next release thanks to @bigfooted

PR #1465, opened 2021-12-06
Options for appending iteration number to filenames (keep solution snapshots)

bigfooted:
## Proposed Changes
 WRT_RESTART_OVERWRITE= NO
WRT_VOLUME_OVERWRITE= NO
WRT_SURFACE_OVERWRITE= NO
will create, every WRT_FREQ iterations, output files with the iteration number appended. This is additional to the normal output, so the restart.dat, etc files will always contain the most recent results.

Other features that we can add:
WRT_RESTART_MAX_FILES_TO_KEEP= 5
will keep only the 5 most recently written restart_xxxx.dat files, throwing away older ones.

## Related Work
related to the issue #1463

## PR Checklist

- [] I have added a test case that demonstrates my contribution, if necessary.
- [] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.


pcarruscag (commented):
Like this, this will probably not be compatible with any other feature of SU2, what about:
- Never do it for unsteady;
- Instead of outputting with the iteration appended, output with the normal name. However, if the file already exists, rename (move) the existing file with the "previous" iteration number, and then continue with the normal output.

This way none of the existing feature should break, and we get the desired feature of having checkpointed files we can go back to if the solver blows up?

bigfooted:
Thanks for the comments!

>     * Never do it for unsteady;
I saw that too, good idea!
 
>     * Instead of outputting with the iteration appended, output with the normal name. However, if the file already exists, rename (move) the existing file with the "previous" iteration number, and then continue with the normal output.
This approach does mean that the file sequence is always behind by OUTPUT_WRT_FREQ. We could also create a symbolic link where restart.dat points to the latest restart_xxx.dat if the existence of restart.dat is necessary for existing workflows.

pcarruscag:
You can store the iteration number for the next renaming so that the sequence is in sync.
I would rather not rely on symbolic links to ensure we don't run into issues with OS or file system quirks. For maximum peace of mind :)

bigfooted:
OK, I focus on the binary restart for now. What I do now is at iteration 20 I rename the restart.dat to restart_0010.dat and then write the new restart.dat. at the final iteration 100, I check if we have reached the final iteration and I write the final restart file twice, once as restart.dat and once as restart_0100.dat A problem that occurs is that the iterations start at 0 and not 1, so the last 2 files are called restart_0089.dat and restart_0099.dat

bigfooted:
OK, code is now functional and cleaned up. If the approach for binary restart is fixed, I can will extend it to other outputs. I think the most interesting are the outputs that are used for visualization.

Common/src/CConfig.cpp
```diff
@@ -8092,7 +8104,7 @@ string CConfig::GetFilename(string filename, string ext, int Iter) const {
   /*--- Add the extension --- */
 
   filename = filename + string(ext);
-
+
```
pcarruscag, line 8107:
```suggestion

```

Common/src/CConfig.cpp
```diff
@@ -8101,10 +8113,11 @@ string CConfig::GetFilename(string filename, string ext, int Iter) const {
   if (GetnTimeInstances() > 1)
     filename = GetMultiInstance_FileName(filename, GetiInst(), ext);
 
+  /*--- Append the iteration number for unsteady problems ---*/
   if (GetTime_Domain()){
-    filename = GetUnsteady_FileName(filename, Iter, ext);
+    filename = GetUnsteady_FileName(filename, timeIter, ext);
   }
-
+
```
pcarruscag, line 8120:
```suggestion

```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -308,9 +308,18 @@ void COutput::Load_Data(CGeometry *geometry, CConfig *config, CSolver** solver_c
 void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short format, string fileName){
 
   CFileWriter *fileWriter = nullptr;
+  CFileWriter *fileWriterFinalIteration = nullptr;
```
pcarruscag, line 311: This seems unnecessary.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -308,9 +308,18 @@ void COutput::Load_Data(CGeometry *geometry, CConfig *config, CSolver** solver_c
...
 
-  unsigned short lastindex = fileName.find_last_of(".");
-  fileName = fileName.substr(0, lastindex);
+  string filename_iter;
+
+  /*--- filename with iteration number ---*/
+  unsigned long previous_inner_iteration = curInnerIter - config->GetVolume_Wrt_Freq(); 
+  unsigned long previous_outer_iteration = curOuterIter - config->GetVolume_Wrt_Freq(); 
+
+
```
pcarruscag, line 319: Simply store the iteration used for the last output as a member variable.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -347,15 +356,57 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
     case RESTART_BINARY:
 
-      if (fileName.empty())
+      if (fileName.empty()){
         fileName = config->GetFilename(restartFilename, "", curTimeIter);
 
+        if (!config->GetWrt_Restart_Overwrite()){
+          if (config->GetMultizone_Problem()){
+            /* --- do nothing --- */  
+          }
```
pcarruscag, line 365: Can you append numbers also to multizone problems?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -347,15 +356,57 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
+            // for the final iteration we need to add 1
+            std::stringstream inner_iter_ss;
+            if (config->GetnIter() == curInnerIter+1){ 
+              inner_iter_ss << "_" << std::setw(8) << std::setfill('0') << previous_inner_iteration+1;
+              filename_iter.append((inner_iter_ss).str());
+            }
+            else {   
+              inner_iter_ss << "_" << std::setw(8) << std::setfill('0') << previous_inner_iteration;
+              filename_iter.append(inner_iter_ss.str());
+            }
```
pcarruscag, line 379: If you store the previous number you don't need this logic.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -347,15 +356,57 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
+            filename_iter.append(CSU2BinaryFileWriter::fileExt);
+
+             /*--- rename previous restart.dat to restart_xxxx.dat ---*/
+            rename(restartFilename.c_str(),filename_iter.c_str());
```
pcarruscag, line 383: Should this be done by all ranks or only master?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -347,15 +356,57 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
 
       fileWriter = new CSU2BinaryFileWriter(fileName, volumeDataSorter);
 
+      if ((!config->GetWrt_Restart_Overwrite()) && (config->GetnIter() == curInnerIter+1)) {
+        std::stringstream inner_iter_ss;
+        inner_iter_ss << "_" << std::setw(8) << std::setfill('0') << curInnerIter;
+        filename_iter = fileName;
+        filename_iter.append(inner_iter_ss.str());
+        fileWriterFinalIteration = new CSU2BinaryFileWriter(filename_iter, volumeDataSorter);
+
```
pcarruscag, line 403: I don't get why this is needed.

SU2_CFD/src/output/COutput.cpp
bigfooted: yes, it looks like the only difference is that there is already a zone number added.

SU2_CFD/src/output/COutput.cpp
bigfooted: yep, master only.

SU2_CFD/src/output/COutput.cpp
bigfooted: Because I want the file restart_0099.dat and else we will just have restart.dat at iteration 99. It will mainly be important for the paraview files to get a nice sequence, but I would rather have SU2 write an additional file than do the copy/rename myself all the time. 

We could also simply write a restart.dat, and also write a restart_xxxx.dat every WRT_FREQ iterations, so we write it twice. There is additional writing, but it will make the implementation a bit cleaner. This is how certain commercial flow solvers do it as well. 
On the one hand I like that we can skip a complete file write, on the other I do not like that we are WRT_FREQ behind in writing the sequence.

pcarruscag: Ok but this logic fails if the solver converges before reaching the maximum number of iterations (so why bother? if someone wants to make a steady state movie they can rename one file, or symlink, etc.).
It looks too clunky having to create another file writer just to write the same thing with a different name, maybe add a method to the base file writer to modify the name and then call it again.

We do not have enough people maintaining the code, hence my insistence in making this as simple as possible.

bigfooted: Easiest is to just write (and not rename) 2 files all the time: the restart.dat (or other files) and the same file with the iteration number appended. Writing is slower than renaming, but it will make the code simpler.

bigfooted:
Simplest is to just copy the saved file again, with an iteration number appended. No keeping track of what the iteration number at the previous write was, no copying of the final saved file with manually appending the final iteration number, no lag of WRT_FREQ in the availability of the restart_xxx.dat file.

It does mean that we do a copy instead of a rename. But I think this is the better (because the simplest) solution. I will extend it to the other file options.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -347,11 +350,28 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
+          }
+        }
+      }
+      
       if (rank == MASTER_NODE) {
-          (*fileWritingTable) << "SU2 restart" << fileName + CSU2BinaryFileWriter::fileExt;
+        if (!config->GetWrt_Restart_Overwrite())
+          (*fileWritingTable) << "SU2 restart with iteration" << filename_iter ;
+
+        (*fileWritingTable) << "SU2 restart" << fileName + CSU2BinaryFileWriter::fileExt;
```
pcarruscag, line 374: Before you apply this to the other types, put this block in a private function, to which you pass part of the description ("SU2 restart") and the file extension.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -741,6 +761,16 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, unsigned short f
...
 
   }
+
+  if (!config->GetWrt_Restart_Overwrite() && (rank == MASTER_NODE)){
+    /*--- copy the file with iteration number appended ---*/
+    std::ifstream  src(restartFilename.c_str(), std::ios::binary);
+    std::ofstream  dst(filename_iter.c_str(),   std::ios::binary);
+    dst << src.rdbuf();
+
+  }
```
pcarruscag, line 771: Change the file writer so that you can pass the filename to Write_Data(), and call it twice.
Most of the writers are parallel, here you are using a single core to do the copy.

bigfooted: Yes, it should be parallel, but it's less intrusive to just instantiate a new CFileWriter then.

pcarruscag: I will not approve solution that involve duplicating patterns.
If you delete the "fileName" member variable from the writer, you immediately see where it is being used, and so the chance of that introducing a bug is very low.

bigfooted: OK, I changed the file writer, and also switched to enum class for the enum_output while we're at it. The filename is not part of the constructor but an argument of Write_Data.

Common/src/grid_movement/CSurfaceMovement.cpp
```diff
@@ -499,18 +499,18 @@ vector<vector<su2double> > CSurfaceMovement::SetSurface_Deformation(CGeometry *g
...
             auto FileFormat = config->GetVolumeOutputFiles();
 
-            if (isParaview(static_cast<ENUM_OUTPUT>(FileFormat[iFile]))) {
+            if (isParaview(FileFormat[iFile])) {
               cout << "Writing a Paraview file of the FFD boxes." << endl;
               for (iFFDBox = 0; iFFDBox < GetnFFDBox(); iFFDBox++) {
                 FFDBox[iFFDBox]->SetParaview(geometry, iFFDBox, false);
               }
-            } else if (isTecplot(static_cast<ENUM_OUTPUT>(FileFormat[iFile]))) {
+            } else if (isTecplot(FileFormat[iFile])) {
```
pcarruscag, line 507: :+1:

SU2_CFD/src/output/filewriter/CSU2MeshFileWriter.cpp
```diff
@@ -30,23 +30,26 @@
...
-   CFileWriter(std::move(valFileName), valDataSorter, fileExt), iZone(valiZone), nZone(valnZone) {}
+   CFileWriter(valDataSorter, fileExt), iZone(valiZone), nZone(valnZone) {}
 
-void CSU2MeshFileWriter::Write_Data() {
+void CSU2MeshFileWriter::Write_Data(string val_filename) {
 
   ofstream output_file;
 
+  /*--- We append the pre-defined suffix (extension) to the filename (prefix) ---*/
+  val_filename.append(fileExt);
```
pcarruscag, line 42: good idea

Common/include/option_structure.hpp
```diff
@@ -1693,65 +1693,67 @@ static const MapType<std::string, ENUM_INPUT> Input_Map = {
...
+  MakePair("SURFACE_PARAVIEW_LEGACY", OUTPUT_TYPE::SURFACE_PARAVIEW_LEGACY_BINARY)
+  MakePair("PARAVIEW", OUTPUT_TYPE::PARAVIEW_XML)
+  MakePair("SURFACE_PARAVIEW", OUTPUT_TYPE::SURFACE_PARAVIEW_XML)
+  MakePair("PARAVIEW_MULTIBLOCK", OUTPUT_TYPE::PARAVIEW_MULTIBLOCK)
+  MakePair("MESH", OUTPUT_TYPE::MESH)
+  MakePair("RESTART_ASCII", OUTPUT_TYPE::RESTART_ASCII)
+  MakePair("RESTART", OUTPUT_TYPE::RESTART_BINARY)
+  MakePair("CGNS", OUTPUT_TYPE::CGNS)
+  MakePair("SURFACE_CGNS", OUTPUT_TYPE::SURFACE_CGNS)
+  MakePair("STL_ASCII", OUTPUT_TYPE::STL_ASCII)
```
pcarruscag, line 1741:
```suggestion
  MakePair("STL", OUTPUT_TYPE::STL_ASCII)
```

bigfooted: Yes, I spotted the mistake. Unfortunately, the syntax is not consistent: STL should be coupled to the binary to be consistent with the other naming. I propose to make the naming consistent at this point (couple STL to STL_BINARY) and change the occurence of "STL" in the config files to "STL_ASCII". Or keep "STL" in the config files and let them write out binary STL instead of ASCII. What do you think?

pcarruscag: Ideally we should not break backward compatibility, but I get the consistency argument and this option is probably not used a lot... Keep STL_ASCII instead of changing the behavior silently.

bigfooted: Do you mean make naming consistent and change "STL" to "STL_ASCII"  in the config files?

pcarruscag: yep

bigfooted:
The testcases contain many comments for OUTPUT_FILES and TABULAR_FORMAT that were outdated descriptions. I have removed all comments for these keywords in the testcases. Actually all testcases need to be stripped from descriptive comments and keywords that are not used or use the default value. This will be a cleanup chore for another PR.

bigfooted:
paraview multiblock is a special case, since it writes a collection of files to a folder. I have now placed the code that was in COutput::WriteToFile into a separate function in CParaviewVTMFileWriter::WriteFolderData. Because the multiblock files need to know abound the mesh and boundary conditions, it needs CConfig and CGeometry as input. But I think all this multiblock code is best placed in this file instead of in COutput.

Common/include/option_structure.hpp
```diff
@@ -1681,7 +1681,7 @@ static const MapType<std::string, OUTPUT_TYPE> Output_Map = {
   MakePair("CGNS", OUTPUT_TYPE::CGNS)
   MakePair("SURFACE_CGNS", OUTPUT_TYPE::SURFACE_CGNS)
   MakePair("STL_ASCII", OUTPUT_TYPE::STL_ASCII)
-  MakePair("STL_BINARY", OUTPUT_TYPE::STL_BINARY)
+  MakePair("STL", OUTPUT_TYPE::STL_BINARY)
```
pcarruscag, line 1684: This is not what we talked about. No "silent changes" please, if someone had STL in their config, this changes the behaviour silently from ASCII to binary.
```suggestion
  MakePair("STL_BINARY", OUTPUT_TYPE::STL_BINARY)
```

Common/include/option_structure.hpp
```diff
@@ -1635,65 +1635,67 @@ static const MapType<std::string, ENUM_INPUT> Input_Map = {
...
+  MakePair("PARAVIEW", OUTPUT_TYPE::PARAVIEW_XML)
+  MakePair("SURFACE_PARAVIEW", OUTPUT_TYPE::SURFACE_PARAVIEW_XML)
+  MakePair("PARAVIEW_MULTIBLOCK", OUTPUT_TYPE::PARAVIEW_MULTIBLOCK)
+  MakePair("MESH", OUTPUT_TYPE::MESH)
+  MakePair("RESTART_ASCII", OUTPUT_TYPE::RESTART_ASCII)
+  MakePair("RESTART", OUTPUT_TYPE::RESTART_BINARY)
+  MakePair("CGNS", OUTPUT_TYPE::CGNS)
+  MakePair("SURFACE_CGNS", OUTPUT_TYPE::SURFACE_CGNS)
+  MakePair("STL_ASCII", OUTPUT_TYPE::STL_ASCII)
+  MakePair("STL", OUTPUT_TYPE::STL_BINARY)
```
pcarruscag, line 1684: Damn it I got confused, but I think I was right, STL by itself disappears, it does not change meaning
```suggestion
  MakePair("STL_BINARY", OUTPUT_TYPE::STL_BINARY)
```

Common/src/CConfig.cpp
```diff
@@ -8075,13 +8092,28 @@ string CConfig::GetFilename(string filename, string ext, int Iter) const {
...
+  std::stringstream inner_iter_ss;
+  inner_iter_ss << "_" << std::setw(8) << std::setfill('0') << curInnerIter;
+  std::stringstream outer_iter_ss;
+  outer_iter_ss << "_" << std::setw(8) << std::setfill('0') << curOuterIter;
+
+  if (GetMultizone_Problem())
+    filename_iter.append(outer_iter_ss.str());
+  else
+    filename_iter.append(inner_iter_ss.str());
+  return filename_iter;
```
pcarruscag, line 8114:
```suggestion
  const auto iter = GetMultizone_Problem() ? curOuterIter : curInnerIter;
  std::stringstream iter_ss;
  iter_ss << filename_iter << "_" << std::setw(6) << std::setfill('0') << iter;
  return iter_ss.str();
```

Common/src/CConfig.cpp
```diff
@@ -8075,13 +8092,28 @@ string CConfig::GetFilename(string filename, string ext, int Iter) const {
...
   if (GetTime_Domain()){
-    filename = GetUnsteady_FileName(filename, Iter, ext);
+    filename = GetUnsteady_FileName(filename, timeIter, ext);
   }
-
+  
   return filename;
 }
 
+string CConfig::GetFilename_Iter(string filename_iter, unsigned long curInnerIter, unsigned long curOuterIter) const {
```
pcarruscag, line 8103:
```suggestion
string CConfig::GetFilename_Iter(const string& filename_iter, unsigned long curInnerIter, unsigned long curOuterIter) const {
```

bigfooted: also added in function definition

Common/include/option_structure.hpp
bigfooted: This option is not used in any of the testcases so changing the name does not break anything.

pcarruscag (approved):
LGTM the feature and the other config fixes.
In the future remember to use the su2code repo for your branches, otherwise it's a pain if someone wants to checkout the branch...

PR #1470, opened 2021-12-10
Treat SU2_CFD library similar to Common in meson build process

thomasdick:
## Proposed Changes

In the last developer meeting, there were some mentions of changing the meson build process for some of the executables. 

The idea is to use the static libraries su2_cfd_lib and su2_cfd_lib_ad for linking when building SU2_DEF, SU2_DOT, SU2_SOL. This way it is no longer necessary to extract the needed sources for building the executables by hand in the meson.build files of each subdirectory. The linking process should ensure that only the necessary sources are linked and the size of the executables does not change too much.

Please feel free to comment on this pull request, whether you think it is a good change to the build process? 

### DIFF
--- a/SU2_DEF/src/meson.build
+++ b/SU2_DEF/src/meson.build
@@ -1,39 +1,9 @@
 su2_def_src = ['SU2_DEF.cpp']
 if get_option('enable-normal')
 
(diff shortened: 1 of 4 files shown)


pcarruscag (approved):
As suspected it does not change the executable size.
Thanks for the cleanup :+1:

PR #1392, opened 2021-10-06
Fix bug in SA-neg diffusion term  (and generalize indices of flow variables for use by scalar solvers and numerics)

WallyMaier:
## Proposed Changes
This draft PR aims to generalize and template the turbulence solver/ scalar solvers to reduce code duplication and allow NEMO access to the turbulence solvers.  The ideas were discussed with @pcarruscag, @TobiKattmann and @suargi, 
I am opening up the draft PR very early, so this is very much a work in process.

The main idea is use templates to introduce differing variable index values into the Scalar/Turb solvers.

## Related Work
This idea follows PR #1375 and PR #1330

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1476, opened 2021-12-24
Using the primitive variable index classes in more places (+ minor chores)

pcarruscag:
Cleanup related with #1392


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -4752,78 +4752,80 @@ void CPhysicalGeometry::SetPoint_Connectivity() {
...
-    auto Degree = nodes->GetnPoint(iPoint);
-    if (Degree < MinDegree) { MinDegree = Degree; AddPoint = iPoint; }
+  /*--- Exclude halo nodes from the ordering process. ---*/
+  for (auto iPoint = nPointDomain; iPoint < nPoint; iPoint++) {
+    InQueue[iPoint] = true;
   }
 
-  /*--- Add the node in the first free position. ---*/
+  /*--- Repeat as many times as necessary to handle disconnected graphs. ---*/
+  while (Result.size() < nPointDomain) {
```
pcarruscag, line 4771: Reason for the residual changes.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2559,15 +2550,9 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
-          else       Grad_Vel[iDim][jDim] = nodes->GetGradient_Primitive(iPoint, iDim + VEL_INDEX, jDim);
+          Grad_Vel[iDim][jDim] = nodes->GetGradient_Primitive(iPoint, prim_idx.Velocity() + iDim, jDim);
         }
-
-        /// TODO: Move the temperature index logic to a function.
-
-        if (FlowRegime == ENUM_REGIME::COMPRESSIBLE) Grad_Temp[iDim] = nodes->GetGradient_Primitive(iPoint, 0, iDim);
-
-        if (FlowRegime == ENUM_REGIME::INCOMPRESSIBLE) Grad_Temp[iDim] = nodes->GetGradient_Primitive(iPoint, nDim + 1, iDim);
+        Grad_Temp[iDim] = nodes->GetGradient_Primitive(iPoint, prim_idx.Temperature(), iDim);
```
WallyMaier, line 2555: Nice! 👍

WallyMaier (commented):
LGTM! Thanks!

Common/src/geometry/CPhysicalGeometry.cpp
TobiKattmann: -> (from dev meeting) Halo points should be at the end of the point list as a convention.
The disconnected domain parts were the change (as they happen in the CHT pin case where disconnected pin-pieces are treated as 1 zone)

TestCases/species_transport/passive_transport_validation/passive_transport.cfg
```diff
@@ -82,7 +82,6 @@ TIME_DISCRE_FLOW= EULER_IMPLICIT
 %
 % --------------------------- CONVERGENCE PARAMETERS --------------------------%
 %
-CONV_CRITERIA= RESIDUAL
```
TobiKattmann, line 85: Wasn't me - Shaggy, 2000

TobiKattmann (commented):
:bouquet: many good cleanups where hardcoded indexes were used. And good that CONV_CRITERIA is now deprecated

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -55,6 +55,9 @@ class CIncEulerVariable : public CFlowVariable {
     inline IndexType ThermalConductivity() const { return nDim+6; }
     inline IndexType CpTotal() const { return nDim+7; }
     inline IndexType CvTotal() const { return nDim+8; }
+
+    /*--- For compatible interface with NEMO. ---*/
+    inline IndexType Temperature_ve() const { return std::numeric_limits<IndexType>::max(); }
```
TobiKattmann, line 60: I need to 
```c++
#include <limits>
```
for this to work using gcc/11.1.0 + openmpi/4.1.2. as well as 11.2.0 + 4.0.5. The test-builds pass(ed) so it cannot be a major issue. 
[This stackoverflow post](https://stackoverflow.com/questions/4798936/numeric-limits-was-not-declared-in-this-scope-no-matching-function-for-call-t) gives resource that the limits header is used much less widely in gcc11 standard library headers -> requiring to explicitly include it.

And I guess we want to assure foward-compatibility ;)

PR #1482, opened 2022-01-07
Fix implicit call for NEMO cases

WallyMaier:
## Proposed Changes
This changes a boolean check in CNEMONumerics.cpp which always seems to be true.  This has been causing some segfaults in 10-11 species viscous cases.   I suspect this is causes because the Kind_TimeNumScheme is not updated before the numerics constructor.

Some work that still needs to be done is confirm/test 10 and 11 species gases do work when the problem is supposed implicit.   
 
## Related Work
PR#1422

## PR Checklist

### DIFF
--- a/SU2_CFD/src/numerics/NEMO/CNEMONumerics.cpp
+++ b/SU2_CFD/src/numerics/NEMO/CNEMONumerics.cpp
@@ -53,7 +53,7 @@ CNEMONumerics::CNEMONumerics(unsigned short val_nDim, unsigned short val_nVar,
     EDDY_VISC_INDEX = nSpecies+nDim+9;
 
     /*--- Read from CConfig ---*/
-    implicit   = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
+    implicit   = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
 
     ionization = config->GetIonization();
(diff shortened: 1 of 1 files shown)


SU2_CFD/src/numerics/NEMO/CNEMONumerics.cpp
```diff
@@ -53,7 +53,7 @@ CNEMONumerics::CNEMONumerics(unsigned short val_nDim, unsigned short val_nVar,
     EDDY_VISC_INDEX = nSpecies+nDim+9;
 
     /*--- Read from CConfig ---*/
-    implicit   = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
+    implicit   = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
```
WallyMaier, line 56: This is always true for at least the first iteration.

WallyMaier:
@CatarinaGarbacz please let me know if this fix resolves your issues.

SU2_CFD/src/numerics/NEMO/CNEMONumerics.cpp
pcarruscag: Yep you need the physics specific versions (_Flow, _Turb, etc.) in constructors.

pcarruscag (approved):

TobiKattmann (approved):

PR #1485, opened 2022-01-11
Updating remain version numbers

WallyMaier:
Small update to version numbers.


configure.ac
```diff
@@ -338,7 +338,7 @@ AC_MSG_RESULT([
 
 -------------------------------------------------------------------------
 |    ___ _   _ ___                                                      |
-|   / __| | | |_  )   Release 7.2.0 'Blackbird'                         |
```
bigfooted, line 341: Nice catch. Is there not an automatic procedure to update the version numbers?

bigfooted (approved):
LGTM

configure.ac
WallyMaier: There is a procedure, but these files may get overlooked.

PR #1492, opened 2022-01-13
Changed the registration of AD input variables in CDiscAdjFluidIteration

thomasdick:
This is a small issue I noticed when I worked on some features for the discrete adjoint driver.
In CDiscAdjFluidIteration::RegisterInput the inputs for the AD tape are registered.
At the moment if kind_recording == RECORDING::SOLUTION_AND_MESH is set it only registers the flow variables and has no difference to RECORDING::SOLUTION_VARIABLES. If I understand the logic for this enum option correctly, it should register the flow variables and the mesh coordinates from geometry. I.e. if you want to add unctions to the driver, which can evaluate derivatives w.r.t. flow and geometry in one tape evaluation, you should use this option.

In addition, I change an out-of-date comment.

## Proposed Changes
Register the geometry variables as inputs in the discrete adjoint flow iteration if 'kind_recording == RECORDING::SOLUTION_AND_MESH'.

## PR Checklist

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -4246,7 +4246,7 @@ public:
   unsigned short GetRef_Inc_NonDim(void) const { return Ref_Inc_NonDim; }
 
   /*!
-   * \brief Get the kind of SU2 software component.
+   * \brief Set the kind of SU2 software component.
    * \return Kind of the SU2 software component.
    */
   void SetKind_SU2(SU2_COMPONENT val_kind_su2) { Kind_SU2 = val_kind_su2 ; }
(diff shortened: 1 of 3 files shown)


pcarruscag (approved):
LGTM I don't think we are using that recording mode anywhere, might be a good candidate for removing / simplifying

thomasdick:
I checked and I could not find a place in the code where this option is used explicitly. Since all the drivers split between primary recording for state variables and secondary recording for geometry variables for efficiency.

My own use case was for research I did, where I needed to record a tape w.r.t. to both. Since this is not ready to become a pull request any time soon, I do not really need this option.

If you suggest I can create a commit to remove the enum option for simplicity.

pcarruscag:
If you plan to use it in the future let's keep it.

PR #1500, opened 2022-01-18
Add HDF5 backend support for CGNS in SU2

MicK7:
## Proposed Changes
 The proposed changes consist in providing an hdf5 enabled cgns library inside SU2.
This is done by embedding the needed hdf5 sources inside the cgns directory.
A basic meson.build file helps with compilation (instead of original cmake).

## Related Work
None

## PR Checklist


MicK7:
The branch can be tested by converting an existing testcase cgns file to hdf5.

MicK7:
It seems that LGTM analysis is using a very old version of meson. Can someone update it ?

pcarruscag:
Thanks for the update @MicK7, this only affects code in externals right? if so there is no need to worry about LGTM, I think it was using the old build system.

MicK7:
@pcarruscag yes all the modification are in external. It is just that the LGTM C++ if failing because it is too old and can't cope with cmake syntax. Since I mainly copy pasted the source of hdf5 without modification, I don't really want to modify them for maintainability.

MicK7:
My request too update the meson used by LGTM C/C++ is to not create issue for other PR is this one goes in.

pcarruscag (approved):
Yep I got that, I was asking to know if I should look at any of the 531 files :smile: 
LGTM seems fixed.

pcarruscag:
Does this support a type of CGNS that we didn't before? If so can you add a small test?

MicK7:
@pcarruscag I can change one of the existing testcases to have a validation

jtlau:
I'm having an issue compiling on macOS. I get an error after "./meson.py build":

`Configuring H5pubconf.h using configuration

externals/cgns/hdf5/meson.build:662:0: ERROR: #mesondefine does not contain exactly two tokens: #cmakedefine H5_HAVE_WINDOWS @H5_HAVE_WINDOWS@`

I'm not sure where to go from there... since I don't need cgns I workaround this by disabling it...
LMK if you need the full configure error

PR #1530, opened 2022-01-31
Integrated Heatflux for flow problems

TobiKattmann:
Hi,

Possibility to give the Heatflux over a boundary in [W] instead of [W/m^2]. This was already possible in the HeatSolver but now also in Inc and Comp flow solvers.
Prescribing a fixed Watt is especially useful in optimization scenarios with Temperature OF where otherwise the optimizer can exploit that a smaller boundary almost certainly yields smaller Temperatures.

The necessary SurfaceAreas are computed and stored in  CGeometry instances which is computed once at program start by a call from the Driver.

I also added NEMO for this feature.

There is currently no Testcase which uses `INTEGRATED_HEATFLUX= YES` so one might want to change that :grimacing: :shushing_face: 

For the HeatSolver this was originally introduced at some point by @oleburghardt 

## PR Checklist

### DIFF
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -199,7 +199,7 @@ public:
   unsigned long *nElem_Bound{nullptr};   /*!< \brief Number of elements of the boundary. */
   string *Tag_to_Marker{nullptr};        /*!< \brief Names of boundary markers. */
   vector<bool> bound_is_straight;        /*!< \brief Bool if boundary-marker is straight(2D)/plane(3D) for each local marker. */
-  vector<su2double> SurfaceAreaCfgFile;         /*!< \brief Total Surface area for all markers. */
+  vector<su2double> SurfaceAreaCfgFile;  /*!< \brief Total Surface area for all markers. */
 
   /*--- Partitioning-specific variables ---*/
 
@@ -917,8 +917,8 @@ public:
    */
   void ComputeSurfaceAreaCfgFile(const CConfig *config);
 
-  /*! Get global Surface Area to a local marker
-  * \brief Adapt Heatflux value for integrated heatflux.
+  /*!
+  * \brief Get global Surface Area to a local marker.
   * \param[in] config - Definition of the particular problem.
   * \param[in] val_marker - Local surface marker.
   * \return Global Surface Area to the local marker
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -199,7 +199,7 @@ public:
   unsigned long *nElem_Bound{nullptr};   /*!< \brief Number of elements of the boundary. */
   string *Tag_to_Marker{nullptr};        /*!< \brief Names of boundary markers. */
   vector<bool> bound_is_straight;        /*!< \brief Bool if boundary-marker is straight(2D)/plane(3D) for each local marker. */
-  vector<su2double> SurfaceArea;         /*!< \brief Total Surface area for all markers. */
+  vector<su2double> SurfaceAreaCfgFile;         /*!< \brief Total Surface area for all markers. */
 
   /*--- Partitioning-specific variables ---*/
 
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -199,6 +199,7 @@ public:
   unsigned long *nElem_Bound{nullptr};   /*!< \brief Number of elements of the boundary. */
   string *Tag_to_Marker{nullptr};        /*!< \brief Names of boundary markers. */
   vector<bool> bound_is_straight;        /*!< \brief Bool if boundary-marker is straight(2D)/plane(3D) for each local marker. */
+  vector<su2double> SurfaceArea;         /*!< \brief Total Surface area for all markers. */
 
   /*--- Partitioning-specific variables ---*/
 
@@ -909,6 +910,12 @@ public:
    * \param[in] fine_grid - Geometry of the fine mesh.
    */
   inline virtual void SetRestricted_GridVelocity(const CGeometry *fine_grid) {}
+
+  /*!
+   * \brief Compute the surface area of all global markers.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void ComputeSurfaceArea(const CConfig *config);
 
   /*!
    * \brief Check if a boundary is straight(2D) / plane(3D) for EULER_WALL and SYMMETRY_PLANE
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -918,6 +918,14 @@ public:
   void ComputeSurfaceArea(const CConfig *config);
 
   /*!
+  * \brief Adapt Heatflux value for integrated heatflux.
+  * \param[in,out] Wall_HeatFlux - Heatflux in [W] which is to be adapted to [W/m^2].
+  * \param[in] config - Definition of the particular problem.
+  * \param[in] val_marker - Surface marker where the boundary condition is applied.
+  */
+  void UpdateIntegrated_Heatflux(su2double* Wall_HeatFlux, unsigned short val_marker, const CConfig *config) const;
+
+  /*!
    * \brief Check if a boundary is straight(2D) / plane(3D) for EULER_WALL and SYMMETRY_PLANE
    *        only and store the information in bound_is_straight. For all other boundary types
    *        this will return false and could therfore be wrong. Used ultimately for BC_Slip_Wall.
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -915,7 +915,7 @@ public:
    * \brief Compute the surface area of all global markers.
    * \param[in] config - Definition of the particular problem.
    */
-  void ComputeSurfaceArea(const CConfig *config);
+  void ComputeSurfaceAreaCfgFile(const CConfig *config);
 
   /*!
   * \brief Adapt Heatflux value for integrated heatflux.
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -917,13 +917,13 @@ public:
    */
   void ComputeSurfaceAreaCfgFile(const CConfig *config);
 
-  /*!
+  /*! Get global Surface Area to a local marker
   * \brief Adapt Heatflux value for integrated heatflux.
-  * \param[in,out] Wall_HeatFlux - Heatflux in [W] which is to be adapted to [W/m^2].
   * \param[in] config - Definition of the particular problem.
-  * \param[in] val_marker - Surface marker where the boundary condition is applied.
+  * \param[in] val_marker - Local surface marker.
+  * \return Global Surface Area to the local marker
   */
-  void UpdateIntegrated_Heatflux(su2double* Wall_HeatFlux, unsigned short val_marker, const CConfig *config) const;
+  su2double GetSurfaceArea(const CConfig *config, unsigned short val_marker) const;
 
   /*!
    * \brief Check if a boundary is straight(2D) / plane(3D) for EULER_WALL and SYMMETRY_PLANE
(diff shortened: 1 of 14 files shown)


Common/include/geometry/CGeometry.hpp
```diff
@@ -199,6 +199,7 @@ class CGeometry {
   unsigned long *nElem_Bound{nullptr};   /*!< \brief Number of elements of the boundary. */
   string *Tag_to_Marker{nullptr};        /*!< \brief Names of boundary markers. */
   vector<bool> bound_is_straight;        /*!< \brief Bool if boundary-marker is straight(2D)/plane(3D) for each local marker. */
+  vector<su2double> SurfaceArea;         /*!< \brief Total Surface area for all markers. */
```
pcarruscag, line 202:
```suggestion
  vector<su2double> SurfaceAreaCfgFile;         /*!< \brief Total Surface area for all markers. */
```

Common/include/geometry/CGeometry.hpp
```diff
@@ -910,6 +911,12 @@ class CGeometry {
    */
   inline virtual void SetRestricted_GridVelocity(const CGeometry *fine_grid) {}
 
+  /*!
+   * \brief Compute the surface area of all global markers.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void ComputeSurfaceArea(const CConfig *config);
```
pcarruscag, line 918: Call it in CGeometry::UpdateGeometry or the sensitivities will be incorrect.
Might also be needed for the OF in multizone adjoints?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -387,6 +387,20 @@ void CIncNSSolver::BC_Wall_Generic(const CGeometry *geometry, const CConfig *con
...
+
+      /*---For integrated Heatflux in [W] instead [W/m^2] find the precomputed marker Surface Area by local-global string-matching. ---*/
+      if(config->GetIntegrated_HeatFlux()) {
+
+        for (unsigned short iMarker_Global = 0; iMarker_Global < config->GetnMarker_CfgFile(); iMarker_Global++) {
+
+          const auto Global_TagBound = config->GetMarker_CfgFile_TagBound(iMarker_Global);
+
+          if (Marker_Tag == Global_TagBound) {
+            Wall_HeatFlux /= geometry->SurfaceArea[iMarker_Global];
```
pcarruscag, line 399: Either store it for iMarker(local) or move this logic to a method of CGeometry, do not duplicate code

TobiKattmann: Yep, will move it to CGeometry instead of CSolver as is 853f5d6

Common/include/geometry/CGeometry.hpp
TobiKattmann: done in 6f91db0

Common/include/geometry/CGeometry.hpp
TobiKattmann: added the call in updateGeometry in 288a2eb . `UpdateGeometry` is only ever called for deforming meshes. When I moved the ComputeSurfaceArea routine only to `UpdateGeometry` my case segfaulted because ComputeSurfaceArea was never called and the vector never resized and filled. The call to `ComputeSurfaceArea` coming from the CDriver ctor (via `Geometrical_Preprocessing`) is always evaluated at startup (except for DG-FEM cases)

Common/include/geometry/CGeometry.hpp
```diff
@@ -910,6 +911,20 @@ class CGeometry {
...
+   */
+  void ComputeSurfaceAreaCfgFile(const CConfig *config);
+
+  /*!
+  * \brief Adapt Heatflux value for integrated heatflux.
+  * \param[in,out] Wall_HeatFlux - Heatflux in [W] which is to be adapted to [W/m^2].
+  * \param[in] config - Definition of the particular problem.
+  * \param[in] val_marker - Surface marker where the boundary condition is applied.
+  */
+  void UpdateIntegrated_Heatflux(su2double* Wall_HeatFlux, unsigned short val_marker, const CConfig *config) const;
```
pcarruscag, line 926: This would have made sense in the solver, maybe... but in geometry it is better to deal in geometry only.
```suggestion
  su2double GetSurfaceArea(const CConfig *config, unsigned short val_marker) const;
```
Also don't start passing scalars by pointer for the love of god :smile:

Common/include/geometry/CGeometry.hpp
TobiKattmann: I am unsure where you think this would go wrong for multizone adjoints :thinking:

Common/include/geometry/CGeometry.hpp
TobiKattmann: ... gotta save those 8 bytes :floppy_disk:

Common/include/geometry/CGeometry.hpp
pcarruscag: Yes I said to call it, not to move it :wink:
Just thinking of the places where `Set_Heatflux_Areas` was used.

Common/include/geometry/CGeometry.hpp
TobiKattmann: 8794f14  here you go

Common/include/geometry/CGeometry.hpp
TobiKattmann: I wanted to tackle `Set_Heatflux_Areas` at another point anyway as it is somewhat connected to `Heat_Fluxes` which computes OF's and hist-output for the HeatSolver. And in there I dont like 2 things:
1. On isothermal walls, the cfg T is used for HF computation. The Isothermal wall is weakly enforced so this could create quite a different value compared to a purely postprocessed (eg via paraview) result
2. AvgTemp is only reported for Heatflux-walls (not even for CHT-interfaces). Imo it should be users choice where he wants which quantities even if they are trivial

I reverted the cut-down of `Set_Heatflux_Areas` here e198acf  because I would need to care about above points and that might be better done in a separate PR

Common/src/geometry/CGeometry.cpp
```diff
@@ -2503,6 +2505,57 @@ void CGeometry::UpdateCustomBoundaryConditions(CGeometry **geometry_container, C
...
+        }// for iVertex
+      }//if Local == Global
+    }//for iMarker_Global
+  }//for iMarker
+
+  SU2_MPI::Allreduce(LocalSurfaceArea.data(), SurfaceAreaCfgFile.data(), SurfaceAreaCfgFile.size(), MPI_DOUBLE, MPI_SUM, SU2_MPI::GetComm());
+}
+
+su2double CGeometry::GetSurfaceArea(const CConfig *config, unsigned short val_marker) const {
+  /*---For integrated Heatflux in [W] instead [W/m^2] find the precomputed marker Surface Area by local-global string-matching. ---*/
```
pcarruscag, line 2544: update

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -436,6 +436,9 @@ void CNSSolver::BC_HeatFlux_Wall_Generic(const CGeometry *geometry, const CConfi
 
   if (kind_boundary == HEAT_FLUX) {
     Wall_HeatFlux = config->GetWall_HeatFlux(Marker_Tag)/config->GetHeat_Flux_Ref();
+    if(config->GetIntegrated_HeatFlux()) {
```
pcarruscag, line 439: clang format

Common/src/geometry/CGeometry.cpp
TobiKattmann: yep f951a07

SU2_CFD/src/solvers/CNSSolver.cpp
TobiKattmann: yes a57f4a3

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -405,7 +408,10 @@ void CNEMONSSolver::BC_HeatFluxCatalytic_Wall(CGeometry *geometry,
   string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
 
   /*--- Get the specified wall heat flux from config ---*/
-  su2double Wall_HeatFlux = config->GetWall_HeatFlux(Marker_Tag);
+  su2double Wall_HeatFlux = config->GetWall_HeatFlux(Marker_Tag) / config->GetHeat_Flux_Ref();
+  if (config->GetIntegrated_HeatFlux()) {
+    Wall_HeatFlux /= geometry->GetSurfaceArea(config, val_marker);
+  }
```
TobiKattmann, line 414: @WallyMaier to let you know: Here in `BC_HeatFluxCatalytic_Wall` the nondim with Heatflux_Red was forgotten. And the NEMO-tribe can (upon merge of this PR) prescribe HeatFlux in [W] instead of [W/m^2] in case someone wants to.

TobiKattmann:
Thanks already @pcarruscag for the quick suggestions  :+1: Let me know if there are more :)

2 linked PR's go with this:
1. The testases PR needs to be merged before this to make sure the reg tests works
2. The website update (...has more time ;)

Common/include/geometry/CGeometry.hpp
```diff
@@ -910,6 +911,20 @@ class CGeometry {
...
   inline virtual void SetRestricted_GridVelocity(const CGeometry *fine_grid) {}
 
+  /*!
+   * \brief Compute the surface area of all global markers.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void ComputeSurfaceAreaCfgFile(const CConfig *config);
+
+  /*! Get global Surface Area to a local marker
+  * \brief Adapt Heatflux value for integrated heatflux.
```
TobiKattmann, line 921:
```suggestion
  /*!
  * \brief Get global Surface Area to a local marker.
```

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_3d/configSolid.cfg
```diff
@@ -25,7 +25,9 @@ THERMAL_CONDUCTIVITY_CONSTANT= 200
 %
 MARKER_SYM= ( solid_sym_sides)
 %
-MARKER_HEATFLUX= ( solid_bottom_heater, 5e5, \
+% 5e5[W/m^2]@solid_bottom_heat * 3.59172[m^2]@solid_bottom_heater = 17.9586 [W]
```
TobiKattmann, line 28:
```suggestion
% 5e5[W/m^2]@solid_bottom_heat * 3.59172e-5[m^2]@solid_bottom_heater = 17.9586 [W]
```

TestCases/parallel_regression.py
```diff
@@ -1428,7 +1428,7 @@ def main():
     sp_pinArray_3d_cht_mf_hf_tp.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_3d"
     sp_pinArray_3d_cht_mf_hf_tp.cfg_file  = "configMaster.cfg"
     sp_pinArray_3d_cht_mf_hf_tp.test_iter = 30
-    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [0.511984, -3.063453, -0.451962, -0.008477, 214.707868, 365.670000] #last 7 lines
+    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [0.511984, -0.894867, -0.451962, -0.008477, 214.707868, 365.670000] #last 7 lines
```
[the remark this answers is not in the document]
TobiKattmann: Done and the parallel_regression workflow has a :heavy_check_mark:

SU2_CFD/src/solvers/CNEMONSSolver.cpp
WallyMaier: Thank you! Good catch!

pcarruscag (approved):

TobiKattmann:
Thanks for the reviewing @pcarruscag :bouquet: I am merging this in then. And then afterwards the website PR https://github.com/su2code/su2code.github.io/pull/69

PR #1527, opened 2022-01-25
Streamwise Periodic restarts using flow.meta + Multizone PerSurface output

TobiKattmann:
Hi,
 
- main thing of this PR is supposed to be the working restarts for prescribed massflow in streamwise periodic simulations (primal and adjoint). The massflow is achieved by dialing a pressure drop which is written and read to/from the `flow.meta` file. That file was used mainly for fixed CL stuff as far as I am aware of. The story is not done though really with this one change. There is still the restart-should-contain-everything-in-the-first-iteration-for-the-adjoint-taping issue which is currently not the case ... and not super straightforward to solve -> see https://github.com/su2code/SU2/commit/c11354115757d673a771e04562e6920024e013a2 description for more info. (EDIT: I will not tackle the massflow adjoint part in this PR, as with the PerSurf output it contains enough stuff which I like to have merged)
- I added a `weight` to heat-solid zone OF's
- when `pressure_drop` is used, the first two entries of `MARKER_ANALYZE` are used as out-inlet but more than two markers are allowed. This comes in handy with `FLOW_COEFF_SURF` imo
- Adding per surface output for multizone cases. Was not working before in the way that `_SURF[0]` fields were reported to be ignored.  A regression test is updated for that.
- Little fix to the regression system where Testcase.py had unsuitable default values such that failing reg tests were reported passing

So I will continue fiddling a little bit with massflow restarts but i already successfully optimized with that setup so it is not completely useless.

## Related Work
#773 

## PR Checklist

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1536, opened 2022-02-04
Fix adjoint for streamwise periodic massflow + General handling of adjoints of additional solution variables

TobiKattmann:
Dear reader, 

Let me give a little bit background first:

When a massflow is prescribed for periodic flow (`KIND_STREAMWISE_PERIODIC= MASSFLOW`), then a pressure drop is iterated(/adapted in each iteration) to fulfill that massflow requirement. In the Volume Source term only this pressure drop is used. The initial value at solver start is the value given in `STREAMWISE_PERIODIC_PRESSURE_DROP= ??` and for restarted/adjoint simulations the `STREAMWISE_PERIODIC_PRESSURE_DROP` from `flow.meta` is taken.

Now this PR:

Until now, the update of the pressure drop, based on the difference between current and prescribed massflow, was done in the Flow solver Preprocessing. That causes some trouble as that routine is called multiple times during normal starts, restarts and even twice(!) for every normal iteration. So do not put any iterated quantities in those like I did. I now did put the update right in front of where it is used -> in the Source_Residual. Like so there shoulnd't be that much trouble of correct taping in that one iteration for the adjoint and it makes the logic of `Inner_iter/outer_iter==0` etc unnecessary.

I am not sure the changes up until now are necessary but I am not sure whether they are enough tbh. I will update this description accordingly:

Update: So to give you an idea of the current state of gradient validation. My setup is the good'ol 2D periodic Pin case (fluid-only) with the Design variables being FFD-parameters only around the middle pin (so no Periodic Interface shenanigans up to now) and I have 4 OF: Avg_Temp and Drag on the middle-Pin Surface as well as the Pressure-drop between In/outlet plus massflow through in- OR outlet (which is the same in value but maybe not for the gradient :thinking: ):
1. For prescribed pressure drop I get a very good agreement for Avg_Temp and Drag (below 0.1% relative diff across the board).
image
 For Massflow and pressure drop the gradients are off by quite a bit but not too shabby (a good optimizer certainly could make good use of those gradients). Of course the Pressure-drop gradient should be 0 but there are prob some numerical gains to be made. The Pressure-drop gradient is O(1e-5) where the others are O(1 - 1e4) so I consider that to be ok
2. For prescribed massflow (and I differentiate here between the code before this PR and ...well this PR): The gradient for Avg_Temp and drag are okay-ish but far from being validated (just like for massflow and dp for prescribed dp)
image
For massflow OF I get wrongly reported gradients by the adjoints, both before and after this PR
image
and pressure drop OF
image

Two other aspects.
1. the PRESSURE_DROP OF is computed in streamwise periodic flow using the Recovered_Pressure (as the "standard" Pressure is the same on both faces). There I need to make sure that the recovered values are computed as well after the iteration (but with the Flow Preprocessing after the MultiGridCycle that seems to be the case)
2. As pressure drop is iterated to get the correct flow solution, is it necessary to treat is specially in the adjoint like a solution variable? <- turns out, that was the case and also is the majority of work done in this PR

Le fin

## Related Work
#773 initial implementation
#1527 first step in getting a consistent restart for massflow prescribed flow by storing the pressure drop in massflow-meta file

## PR Checklist


SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1614,6 +1614,24 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+      /*--- Compute update to Delta p based on massflow-difference ---*/
+      const su2double Average_Density_Global = SPvals.Streamwise_Periodic_AvgDensity;
+      const su2double Area_Global = SPvals.Streamwise_Periodic_BoundaryArea;
+      const su2double TargetMassFlow = config->GetStreamwise_Periodic_TargetMassFlow() / (config->GetDensity_Ref() * config->GetVelocity_Ref());
+      const su2double MassFlow_Global = SPvals.Streamwise_Periodic_MassFlow;
+      const su2double ddP = 0.5 / ( Average_Density_Global * pow(Area_Global, 2)) * (pow(TargetMassFlow, 2) - pow(MassFlow_Global, 2));
+
+      /*--- Store updated pressure difference ---*/
+      const su2double damping_factor = config->GetInc_Outlet_Damping();
+      SPvals.Streamwise_Periodic_PressureDrop = SPvals.Streamwise_Periodic_PressureDrop + damping_factor*ddP;
```
pcarruscag, line 1632:
```suggestion
      SPvals.Streamwise_Periodic_PressureDrop += damping_factor*ddP;
```

pcarruscag (approved):
Grabs :popcorn: 

The fixed-Cl mode is not active during adjoints, the AOA is just taken from the metadata.
If something is determined iteratively, then I think it is a solution variable, but I never validated this. I always made my life simpler by constraining Cl instead of using fixed Cl :blush: 
In theory mass flow outlets and your periodic mass flow fall in the same category of fixed-Cl.

TobiKattmann:
> The fixed-Cl mode is not active during adjoints, the AOA is just taken from the metadata.
I.e. you would propose to try to disable the delta-pressure udpdate during the adjoint for prescribed massflow? In my understanding that would result in an inconsistent gradient -> which might work good enough for optimizations but in a gradient validation I would expect significant deviations.

> If something is determined iteratively, then I think it is a solution variable, [...]
So it is settled then -> I will try to give pressure-drop the same treatment as regular solution variables as mentioned in point 2. of the PR description.

TobiKattmann:
> In theory mass flow outlets and your periodic mass flow fall in the same category of fixed-Cl

I thought about that as well. If I get the streamwise periodic massflow to work that should be equivalent for massflow outlets. After all, quite a few ideas were taken from that implementation.

Maybe a question to @economon in that regard (I surely asked that already but cannot remember :brain: :hammer:  ): Did you make a gradient validation for massflow-outlets back in the day? If yes, do you remember whether that went well, or did you see some imperfection that wasn't there for pressure-outlets?

pcarruscag:
Yes it is inconsistent that's why it would need validation, the worst-case scenario is when your design variable is the thing the primal solver is manipulating to implement the fixed mode.
For example I'm pitching the wing (geometric AOA let's call it) the adjoint gives me a value for dCL / dalpha, but in practice this value is 0 because the primal solver does not allow the CL to change.
But doing the above would be dumb right, the validation question is what happens to the derivatives of other outputs? Is dCD / dalpha correct? I imagine this depends on the relation between CD and CL in this case.

pcarruscag:
This can probably be tested with a dummy fixed point iteration...

TobiKattmann:
With commit [1836f11](https://github.com/su2code/SU2/pull/1536/commits/1836f111c0759eb50946ba6b68ac281ef4b1b717)  the gradients (for massflow prescribed in the primal)  can now be validated really well (tested for AvgTemp and PRESSURE_DROP OF's). 

BUT only for serial execution ... for multiple processes the adjoint diverges after a few iterations. I need to give that parallel treatment another thought.

TobiKattmann:
So for CHT the situation is this:

For prescribed massflow I do get a good gradient agreement (DA vs FD). Below a picture with avgTemp and massflow OF
image

But if I switch to prescribed massflow and look at  the avgTemp and pressureDrop OF then the gradient is again bad
image

TobiKattmann:
Now I am happy with the results of the code but the question now is: How to code this up nicely!
This concept of an additional iterated variable next to the solution variable is "new" to SU2.
1. Add the Adjoint pressure-drop variables always to the existing arrays (`Solution`, `External`, `AD_InputIndex`, etc.) which would make for ugly allocation routines and one would have to mess with `nPoint` as one would prob extend the containers by one element. So I consider this direction to be a bit too intrusive as this concept is not needed for most applications anyway.
2. Add an `Extended` part to all relevant copide methods and variables and keep that general for the different applications

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,101 @@ class CVariable {
...
+ public: // Make public for easy debug access
+
+  // Add an su2double to the origin addresses for the extra solution variables
+  // only called by direct solver
+  inline void AddSolutionExtra_OriginAdress(su2double* val) {
+    SolutionExtra_OriginAdresses.resize(1); // here a push_back would be premium
+    SolutionExtra_OriginAdresses[0] = val;
+
+    AD_InputIndex_Extra.resize(SolutionExtra_OriginAdresses.size()) = -1;
+    AD_OutputIndex_Extra.resize(SolutionExtra_OriginAdresses.size()) = -1;
```
TobiKattmann, line 117: This is still a hack for just 1 additional variable. With a push_back() this would be really nice -> The `In/Output_Index` allocation of course could move somewhere or I always resize to `size()+1` .... wait I could do that to `SolutionExtra_OriginAdresses` (does the resize keep the existing values though? ... no its not ) anyway once this would be solved the implementation is necessarily nice :grimacing: but it is easily extendible (well of course the primal handling of that value has to be considered as well, proper restart + calling just once during the simulation in the first iteration  )

I am unsure if there are applications which need these containers to become huge... so maybe an additional thought on not making it unnecessarily slow should be spent.

There was the idea to put this whole logic in a separate class (as far as I remember and understood correctly). But this instance still works the initial hacked version and that is a good thing :) .... I might have made all possible mistakes along the way and permuted a few &* in the process :cry:

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -112,6 +112,21 @@ void CVariable::SetExternalZero() { parallelSet(External.size(), 0.0, External.d
...
+      AD::SetIndex(AD_InputIndex_Extra[iEntry], *SolutionExtra_OriginAdresses[iEntry]);
+    }
+  }
+  else {
+    assert(SolutionExtra_OriginAdresses.size() == AD_OutputIndex_Extra.size());
+    for (auto iEntry = 0ul; iEntry < SolutionExtra_OriginAdresses.size(); iEntry++) {
+      AD::RegisterOutput(*SolutionExtra_OriginAdresses[iEntry]);
+      AD::SetIndex(AD_OutputIndex_Extra[iEntry], *SolutionExtra_OriginAdresses[iEntry]);
+    }
+  }
```
TobiKattmann, line 129: I can prob  use the RegisterContainer as well

SU2_CFD/include/variables/CVariable.hpp
TobiKattmann: @pcarruscag  Any  comments to the implementation would be much appreciated :) When it comes to commenting all the things and cleaning some things -> I'll do it at the end

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -123,6 +123,8 @@ void CDiscAdjSolver::SetRecording(CGeometry* geometry, CConfig *config){
   }
   END_SU2_OMP_FOR
 
+  direct_solver->GetNodes()->SetSolutionExtra(nodes->GetSolutionExtra_Direct());
```
pcarruscag, line 126: Instead of keeping and passing around vectors of pointers, make SetSolutionExtra a method of the solver, then it can set the values of its extra variables directly.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,101 @@ class CVariable {
...
+
+  VectorType SolutionExtra_Direct; // Stores values of the original direct solution, for later resetting
+  su2vector<su2double*> SolutionExtra_OriginAdresses; // Stores pointers to the original vars, these vars are resetted via this pointer vector
+
+ public: // Make public for easy debug access
+
+  // Add an su2double to the origin addresses for the extra solution variables
+  // only called by direct solver
+  inline void AddSolutionExtra_OriginAdress(su2double* val) {
+    SolutionExtra_OriginAdresses.resize(1); // here a push_back would be premium
```
pcarruscag, line 113: Just give it an interface that takes the vector of initial values, these give you the size and the direct values for you to store and then callback the direct solver to do the reset.

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -876,8 +875,10 @@ void CDiscAdjMultizoneDriver::Add_Solution_To_External(unsigned short iZone) {
 
   for (unsigned short iSol=0; iSol < MAX_SOLS; iSol++) {
     auto solver = solver_container[iZone][INST_0][MESH_0][iSol];
-    if (solver && solver->GetAdjoint())
+    if (solver && solver->GetAdjoint()) {
       solver->Add_Solution_To_External();
+      solver->GetNodes()->Add_SolutionExtra_To_ExternalExtra();
```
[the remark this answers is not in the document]
TobiKattmann: Sure np

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,101 @@ class CVariable {
...
+  }
+
+  void Add_ExternalExtra_To_SolutionExtra() {
+    assert(SolutionExtra.size() == ExternalExtra.size());
+    for (auto iEntry = 0ul; iEntry < SolutionExtra.size(); iEntry++)
+      SolutionExtra[iEntry] += ExternalExtra[iEntry];
+  }
+
+  // seed, pass by value ok here
+  inline void SetAdjointSolutionExtra(const VectorType adj_sol) {
```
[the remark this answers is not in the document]
TobiKattmann: [d745e13](https://github.com/su2code/SU2/pull/1536/commits/d745e13349197db26feb2b258e61777edd949199)

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,101 @@ class CVariable {
...
+      SolutionExtra_Direct[iEntry] = *sol_extra_adresses[iEntry];
+  }
+
+  // Get the Addresses of the Extra primal vars in order to reset their value
+  inline const su2vector<su2double*>& GetSolutionExtra_OriginAdresses() const {
+    return SolutionExtra_OriginAdresses;
+  }
+
+  // Reset the primal extra solution to its original value
+  inline void SetSolutionExtra(VectorType direct_sol) {
```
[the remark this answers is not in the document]
TobiKattmann: [d745e13](https://github.com/su2code/SU2/pull/1536/commits/d745e13349197db26feb2b258e61777edd949199)

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,101 @@ class CVariable {
...
+  inline void AddSolutionExtra_OriginAdress(su2double* val) {
+    SolutionExtra_OriginAdresses.resize(1); // here a push_back would be premium
+    SolutionExtra_OriginAdresses[0] = val;
+
+    AD_InputIndex_Extra.resize(SolutionExtra_OriginAdresses.size()) = -1;
+    AD_OutputIndex_Extra.resize(SolutionExtra_OriginAdresses.size()) = -1;
+  }
+
+  // Store a hardcopy of the original extra solution vars
+  inline void SetSolutionExtra_Direct(su2vector<su2double*> sol_extra_adresses, bool multizone) {
```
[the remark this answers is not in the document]
TobiKattmann: [9dd5d6f](https://github.com/su2code/SU2/pull/1536/commits/9dd5d6ff28dc98df7fa834cbaf5a5a7548225d94)

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
TobiKattmann: I am unsure how it helps if the the method moves to the solver because the fundamental problem is:

`SolutionExtra_OriginAdresses, AD_InputIndex_Extra, AD_OutputIndex_Extra` are owned by the direct_solver

`SolutionExtra, ExternalExtra, SolutionExtra_BGS_k, SolutionExtra_Direct` are owned by the DiscAdj solver.

So I need to pass `SolutionExtra_Direct` from the DiscAdj solver to the direct_solver such that the direct_solver can reset the primal values at the `SolutionExtra_OriginAdresses`. And I need calls to a Variable method anyway as the values are at least protected, meaning: a solver method would add code without making something else more concise?

TobiKattmann: By the way -> This distinction of container-ownership between direct and adjoint solver is already present for the current DA solver. 

With the only slight difference being `SolutionExtra_Direct` where the "normal" counterpart is part of `CDiscAdjVariable` and not `CVariable`... which saves me some virtuals

SU2_CFD/include/variables/CVariable.hpp
TobiKattmann: For now I made the (small & convenient) change to std::vector  here: [](https://github.com/su2code/SU2/pull/1536/commits/fab7a4d42e9e66967c9013cd1c0434e8af15f4a5)[fab7a4d](https://github.com/su2code/SU2/pull/1536/commits/fab7a4d42e9e66967c9013cd1c0434e8af15f4a5)

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -941,7 +940,9 @@ void CDiscAdjMultizoneDriver::SetResidual_BGS(unsigned short iZone) {
 
   for (unsigned short iSol=0; iSol < MAX_SOLS; iSol++) {
     auto solver = solver_container[iZone][INST_0][MESH_0][iSol];
-    if (solver && solver->GetAdjoint())
+    if (solver && solver->GetAdjoint()) {
       solver->ComputeResidual_Multizone(geometry_container[iZone][INST_0][MESH_0], config_container[iZone]);
+      solver->GetNodes()->Set_BGSSolutionExtra_k();
+    }
```
TobiKattmann, line 946: This is not super straight forward as `CDiscAdjMultizoneDriver::Set_BGSSolution_k_To_Solution`  is a subset of what `ComputeResidual_Multizone` does. `ComputeResidual_Multizone` is used in the primal and adjoint so if the preferred way is to cut the `Set_BGSSolution_k_To_Solution` out, then it needs to be taken care of for the primal as well

TobiKattmann: lets see whether [ef6bbc4](https://github.com/su2code/SU2/pull/1536/commits/ef6bbc43bb29586df2c0a11a95a931b05693b997) works

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -923,25 +922,31 @@ void CDiscAdjMultizoneDriver::Set_Solution_To_BGSSolution_k(unsigned short iZone
 
   for (unsigned short iSol=0; iSol < MAX_SOLS; iSol++) {
     auto solver = solver_container[iZone][INST_0][MESH_0][iSol];
-    if (solver && solver->GetAdjoint())
+    if (solver && solver->GetAdjoint()) {
       solver->GetNodes()->Restore_BGSSolution_k();
+      solver->GetNodes()->Restore_BGSSolutionExtra_k();
```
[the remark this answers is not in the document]
TobiKattmann: [7ff0655](https://github.com/su2code/SU2/pull/1536/commits/7ff06552da1a5e3cbfc53e2c62ab3c9290de1e71) and [](https://github.com/su2code/SU2/pull/1536/commits/ef6bbc43bb29586df2c0a11a95a931b05693b997)[ef6bbc4](https://github.com/su2code/SU2/pull/1536/commits/ef6bbc43bb29586df2c0a11a95a931b05693b997) do that

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
TobiKattmann: > I told you in the meeting to not put this in CVariable but here we are...

I suggest to apply [Hanlon's razor](https://en.wikipedia.org/wiki/Hanlon%27s_razor) when it comes to my work... and this post

> Put these in the direct solver:
> SolutionExtra_OriginAdresses, AD_InputIndex_Extra, AD_OutputIndex_Extra

I guess by direct solver you mean FVMFlowSolverBase, which would make the implementation not generic for HeatSolver and ScalarSolver(s) which are a direct children of CSolver. So in case someone needs to do this there as well, code has to be copied, which I am unsure makes anything better.

> Which means that SolutionExtra_OriginAdresses is not needed anymore since the original variables belong to the solver already.

That means that if I have a variable I want to use as an additional adjoint one needs to manually handle it for storing and restoring the saved value plus at RegisterIn/Output. Storing the address happens once (preferably while in the constructor) and everything else is done automatically. 

> Then what is left in CVariable belongs to the DA solver, which makes the separation clearer.

Fully agree on the clearer thing. Currently it is not obvious from looking at CVariable where each container belongs to. The DAVariable or solver is for all solvers so moving things there does not have the same problems with generality as the direct_solver has.


> You don't need to make this look like the rest of the code.
> For example, the DA solver never needed to know about "SolutionDirect", it should have simply asked the primal to reset its solution, whatever that solution is.

I agree, makes more sense to have SolutionDirect stored in the primal solver. But keeping things consistent is always nice, big fan.

pcarruscag: You don't need to have those actual fields in csolver, what you need is an interface in csolver that's allows you to implement the actions you need.
Registering extra variables, restoring values, getting the adjoints, setting the adjoints etc.
You can even dispense with the index bookkeeping and reset if you don't modify the starting value, and instead write the updated value to another variable (in adjoint mode) that way you have one input and one output and no need to store indices.

pcarruscag: RegisterExtraVariables (gives you a count for you to resize the extra stuff in the DA solver cvariable)
SetAdjointExtraVariable(vector)
GetDerivativeExtraVariable -> vector

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -94,6 +94,34 @@ class CVariable {
...
+  VectorType ExternalExtra; // External storage for the adjoint value (i.e. for the OF mainly)
+
+  VectorType SolutionExtra_BGS_k; // Intermediate storage, enables cross term extraction as that is also pushed to Solution
+
+  su2vector<int> AD_InputIndex_Extra; // Stores input index of the Extra vars
+  su2vector<int> AD_OutputIndex_Extra; // Stores output index of the Extra vars
+
+  VectorType SolutionExtra_Direct; // Stores values of the original direct solution, for later resetting
+  vector<su2double*> SolutionExtra_OriginAdresses; // Stores pointers to the original vars, these vars are resetted via this pointer vector
+
```
TobiKattmann, line 107:
```suggestion
 ```

File: SU2_CFD/include/solvers/CIncEulerSolver.hpp [hunk 13]
```diff
@@ -395,8 +395,30 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
-  StreamwisePeriodicValues GetStreamwisePeriodicValues() const final { return SPvals; }
+  StreamwisePeriodicValues GetStreamwisePeriodicValues() final { return SPvals; }
+  StreamwisePeriodicValues GetStreamwisePeriodicValuesUpdated() final { return SPvalsUpdated; }
+
+  /*!
+   * \brief Register In- or Output.
+   * \param[in] input - Boolean whether In- or Output should be registered.
+   * \param[in] config - The particular config.
+   */
+  void RegisterSolutionExtra(bool input, const CConfig* config);
```
pcarruscag, line 409:
```suggestion
  void RegisterSolutionExtra(bool input, const CConfig* config) final;
```

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -395,8 +395,30 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
+
+  /*!
+   * \brief Seed the adjoint of the extra solution at the output.
+   * \param[in] adj_sol - Vector containing the adjoint solution to seed.
+   * \param[in] config - The particular config.
+   */
+  void SetAdjoint_SolutionExtra(const VectorType& adj_sol, const CConfig* config);
```
pcarruscag, line 416:
```suggestion
  void SetAdjoint_SolutionExtra(const VectorType& adj_sol, const CConfig* config) final;
```

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -395,8 +395,30 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
+
+  /*!
+   * \brief Extract the adjoint of the extra solution at the input.
+   * \param[out] adj_sol - Vector to store the adjoint into.
+   * \param[in] config - The particular config.
+   */
+  void ExtractAdjoint_SolutionExtra(VectorType& adj_sol, const CConfig* config);
```
pcarruscag, line 423:
```suggestion
  void ExtractAdjoint_SolutionExtra(VectorType& adj_sol, const CConfig* config) final;
```

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -72,6 +72,14 @@ CVariable::CVariable(unsigned long npoint, unsigned long ndim, unsigned long nva
...
       AD_OutputIndex.resize(nPoint,nVar) = -1;
     }
+
+    if (config->GetKind_Streamwise_Periodic() == ENUM_STREAMWISE_PERIODIC::MASSFLOW) {
+      SolutionExtra.resize(1) = su2double(1e-16);
+      if (config->GetMultizone_Problem()) {
+        SolutionExtra_BGS_k.resize(1) = su2double(1e-16);
+        ExternalExtra.resize(1) = su2double(0.0);
+      }
+    }
```
TobiKattmann, line 82: This is hardcoded for now, or for longer as I currently am aware of more possible extra solutions (aoa in fixed cl and pressure-drop for massflow outlets or streamwise periodic). I wonder whether it makes sense to make this already more flexible at the moment?

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -395,8 +395,30 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
 
   /*!
-   * \brief Get values for streamwise periodic flow: delta P, m_dot, inlet T, integrated heat.
-   * \return Struct holding 4 su2doubles.
+   * \brief Get values for streamwise periodic flow: delta P, m_dot, inlet T, integrated heat, etc.
+   * \return Struct holding streamwise periodic values.
    */
-  StreamwisePeriodicValues GetStreamwisePeriodicValues() const final { return SPvals; }
+  StreamwisePeriodicValues GetStreamwisePeriodicValues() final { return SPvals; }
+  StreamwisePeriodicValues GetStreamwisePeriodicValuesUpdated() final { return SPvalsUpdated; }
```
pcarruscag, line 402: Do you also need the updated ones? Or can that be an internal detail since SPvals is set to Updated in normal conditions?

SU2_CFD/src/variables/CVariable.cpp
TobiKattmann: Otherwise this PR should be complete in terms of added things

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -64,6 +64,8 @@ using namespace std;
 
 class CSolver {
 protected:
+  using VectorType = C2DContainer<unsigned long, su2double, StorageType::ColumnMajor, 64, DynamicSize, 1>;
+
```
pcarruscag, line 68:
```suggestion
```

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -3543,6 +3545,27 @@ class CSolver {
...
+   * \param[in] config - The particular config.
+   */
+  virtual void RegisterSolutionExtra(bool input, const CConfig* config) {}
+
+  /*!
+   * \brief Seed the adjoint of the extra solution at the output.
+   * \param[in] adj_sol - Vector containing the adjoint solution to seed.
+   * \param[in] config - The particular config.
+   */
+  virtual void SetAdjoint_SolutionExtra(const VectorType& adj_sol, const CConfig* config) {}
```
pcarruscag, line 3560:
```suggestion
  virtual void SetAdjoint_SolutionExtra(const su2activevector& adj_sol, const CConfig* config) {}
```

pcarruscag (commented):
:+1: Looks quite simple to me now, what do you think?

SU2_CFD/include/solvers/CIncEulerSolver.hpp
TobiKattmann: That method is not necessary anymore indeed :+1:

.github/workflows/regression.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         uses: docker://su2code/test-su2:20200303
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c feature_StreamPer_massflow_itWorksDisTime=D -s ${{matrix.testscript}}
```
TobiKattmann, line 95:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```

TestCases/parallel_regression_AD.py
```diff
@@ -346,6 +346,18 @@ def main():
     da_sp_pinArray_cht_2d_dp_hf.multizone = True
     test_list.append(da_sp_pinArray_cht_2d_dp_hf)
 
+    # 2D DA cht streamwise periodic case, 2 zones, PressureDrop objective, additional pressure drop adjoint equation
+    da_sp_pinArray_cht_2d_mf           = TestCase('da_sp_pinArray_cht_2d_mf')
+    da_sp_pinArray_cht_2d_mf.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_2d_new"
```
TobiKattmann, line 351:
```suggestion
    da_sp_pinArray_cht_2d_mf.cfg_dir   = "incomp_navierstokes/streamwise_periodic/dp-adjoint_chtPinArray_2d"
```

TestCases/incomp_navierstokes/streamwise_periodic/dp-adjoint_chtPinArray_2d/configFluid.cfg
```diff
@@ -0,0 +1,141 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: Unit Cell flow around pin array (fluid)                    %
+% Author: T. Kattmann                                                          %
+% Date: 2020.12.15                                                             %
```
TobiKattmann, line 6:
```suggestion
% Date: 2022.02.10                                                             %
```

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -406,6 +413,31 @@ class CVariable {
     for(unsigned long iVar = 0; iVar < nVar; iVar++) External(iPoint,iVar) += val_sol[iVar];
   }
 
+  /*!
+   * \brief Store the adjoint solution of the extra adjoint into the external container.
+   */
+  void Add_SolutionExtra_To_ExternalExtra() {
```
pcarruscag, line 419:
```suggestion
  void Set_ExternalExtra_To_SolutionExtra() {
```

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2198,12 +2198,16 @@ void CSolver::Add_External_To_Solution() {
...
+
+  base_nodes->Add_ExternalExtra_To_SolutionExtra();
 }
 
 void CSolver::Add_Solution_To_External() {
   for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++) {
     base_nodes->Add_External(iPoint, base_nodes->GetSolution(iPoint));
   }
+
+  base_nodes->Add_SolutionExtra_To_ExternalExtra();
```
pcarruscag, line 2210:
```suggestion
  base_nodes->Set_ExternalExtra_To_SolutionExtra();
```

pcarruscag (approved):

SU2_CFD/include/variables/CVariable.hpp
TobiKattmann: :eagle: :eye: 

:+1:

Issue #1487, opened 2022-01-12
MAJOR ISSUES / DEVELOPMENT AREAS


pcarruscag:
Because we can only pin up to 3 issues in GitHub, I'm creating this one to track major aspects we need to fix or implement in SU2.
This issue shall only contain the list and a brief description, with links to associated issues and pull requests.
To propose major issues / development areas please use discussion #1486

The purpose of this list is two fold:
- Direct the efforts of the SU2 Foundation towards addressing these issues, by efforts I mean funds (as they become available).
- Try to enlist help from the community to start addressing these issues, as much as possible we will try to break these down into "bite sized" chunks.

# List
Coming soon.

pcarruscag:
**V&V**
Now we have a regression script dedicated to the V&V examples see #1538
I have added a high-lift case to the website ([here](https://github.com/su2code/su2code.github.io/pull/71), and [here](https://su2code.github.io/vandv/30p30n/)).
I think that to avoid the [issue of outdated files](https://github.com/su2code/VandV/pull/3) in V&V repo, the best strategy is to add only meshes and documentation to the V&V repo (without duplicating the website). Then anything needed to run the case should be put under regression (via the TestCases and SU2 repos as usual).
If the meshes are too large for GitHub, and are not publicly available elsewhere, we can use the SU2 Foundation Google drive. The config files can be forced to remain updated by running them in --dry_run mode in the V&V regressions.

So I think the next steps would be:
- Revive the configs of the [other V&V cases](https://github.com/su2code/VandV) and put them under regression.
- Add V&V for other areas.

So if you have ever done any V&V of an SU2 feature not covered yet please consider **adding the data to the V&V repo, the results to the website, and creating a minimal regression test** for it. If you have never done any V&V, please consider reviving one of the existing ones.

cvnh:
Hi I'm interested in contributing to the development of turbulence models, parallelisation and adaptation. Is anyone working on these topics currently?

PR #1538, opened 2022-02-07
Regression script for the V&V repo

pcarruscag:


TestCases/vandv.py
```diff
@@ -0,0 +1,80 @@
...
+    channel.test_iter = 20
+    channel.test_vals = [-10.731407, -10.312481, -10.517059, -10.348701, -13.517110,  0.050962, 2.828564, 1.317850, -0.240861]
+    test_list.append(channel)
+
+    #################
+    ### RUN TESTS ###
+    #################
+
+    for test in test_list:
+        test.su2_exec = "mpirun -n 2"
```
pcarruscag, line 56:
```suggestion
        test.su2_exec = "mpirun -n 2 SU2_CFD"
```

TestCases/vandv.py
```diff
@@ -0,0 +1,80 @@
...
+    ##########################
+    ### Compressible RANS ###
+    ##########################
+
+    # 30P30N
+    channel           = TestCase('30P30N')
+    channel.cfg_dir   = "vandv/rans/30p30n"
+    channel.cfg_file  = "config.cfg"
+    channel.test_iter = 20
+    channel.test_vals = [-10.731407, -10.312481, -10.517059, -10.348701, -13.517110,  0.050962, 2.828564, 1.317850, -0.240861]
```
pcarruscag, line 48:
```suggestion
    channel.test_vals = [-10.628370, -10.299097, -10.485327, -10.238060, -13.517229, 0.050962, 2.828563, 1.317849, -0.227233]
```

TestCases/vandv.py
```diff
@@ -0,0 +1,80 @@
...
+
+    #################
+    ### RUN TESTS ###
+    #################
+
+    for test in test_list:
+        test.su2_exec = "mpirun -n 2 SU2_CFD"
+        test.timeout = 300
+        test.tol = 1e-5
+    #end
```
TobiKattmann, line 59: We could do that style for parallel regression too :+1: good idea

TestCases/vandv/rans/30p30n/config.cfg
```diff
@@ -0,0 +1,94 @@
...
+NEWTON_KRYLOV_DPARAM= ( 0.0, 1e-20, -3, 1e-5 ) % r0, tp, rf, e
+%
+% ------------------------ CONVERGENCE CRITERIA ------------------------- %
+%
+ITER= 10000
+CONV_RESIDUAL_MINVAL= -11.5
+%
+% --------------------------- INPUT / OUTPUT ---------------------------- %
+%
+MESH_FILENAME= 2D_L1_coarse_r1.su2
```
TobiKattmann, line 85: Concerning https://github.com/su2code/VandV/commit/f5a89277dbf61616c9c0d11524939e522a61934f ... where do you unzip the mesh files?

TestCases/vandv.py
```diff
@@ -0,0 +1,80 @@
...
+    ### Compressible RANS ###
+    ##########################
+
+    # 30P30N
+    channel           = TestCase('30P30N')
+    channel.cfg_dir   = "vandv/rans/30p30n"
+    channel.cfg_file  = "config.cfg"
+    channel.test_iter = 20
+    channel.test_vals = [-10.628370, -10.299097, -10.485327, -10.238060, -13.517229, 0.050962, 2.828563, 1.317849, -0.227233]
+    test_list.append(channel)
```
TobiKattmann, line 49: Why `channel` for 30p30n?

TestCases/vandv/rans/30p30n/config.cfg
```diff
@@ -0,0 +1,94 @@
+% ------------------------------- SOLVER -------------------------------- %
```
TobiKattmann, line 1: Why no header? I personally dont mind having a infos about the origin of a case

TobiKattmann (commented):
Thanks for adding this! much needed 

It would have appreciated a few words though as the title does not give away the added case for example. What is the short-term plan to do with the V&V cases? Are you verifying them all and add them to the reg-test? Maybe link https://github.com/su2code/VandV/pull/3 as an example why we need this.

And is the 30p30n case now a v&v case? It certainly could, but then it currently completely lacks any easy accessible postprocessing results that would make it easy to verify. Especially for V&V cases imo it should be known what result we are aiming for and not just some relative comparison with residuals. But I dont need to tell you that and I am sure you did make some thoughts on all of that... so no essay needed, but a little more `pedro -v`

pcarruscag:
Talk is cheap.
I'll announce the plan in #1487 this is one of the areas in the list.

TestCases/vandv/rans/30p30n/config.cfg
pcarruscag: https://su2code.github.io/vandv/30p30n/

TestCases/vandv/rans/30p30n/config.cfg
pcarruscag: I don't. I cannot change the machines running the tests, therefore I cannot make them checkout the vandv repo, thus I must add what I need to run the tests in Testcases.

PR #1550, opened 2022-02-18
changed limiter to enum class to fix adjoint species bug 

bigfooted:
## Proposed Changes
change limiter to enum class. 

Note: For some reason this fixes a problem for the species transport, where the discrete adjoint solver would stop in computelimiters.hpp with the message "unknown limiter type". This only happens when a computation is run over more than one node (more than one core is OK though).  This is because in CSolver::SetSolution_Limiter, GetKind_SlopeLimit() does not return a valid value for an existing slope limiter. Kind_SlopeLimit is set in CConfig::SetKind_ConvNumScheme, called in CConfig::SetGlobalParam, but this looks OK to me. 

I did just notice that in CConfig:: line 3454 we check for MUSCL_Flow, MUSCL_Turb and MUSCL_AdjFlow and MUSCL_AdjTurb, but not for MUSCL_Species or MUSCL_Adj_Species(does not even exist). It looks like there is still something missing here.

I do not know why the error "unknown limiter type" occured and why it is now fixed by only changing to enum class.

## Related Work
continuing effort to change all enums to enum class

## PR Checklist


Common/include/option_structure.hpp
```diff
@@ -887,23 +887,23 @@ static const MapType<std::string, ENUM_MATRIX_COLORING> MatrixColoring_Map = {
...
  */
-enum ENUM_LIMITER {
-  NO_LIMITER           = 0, /*!< \brief No limiter. */
-  VENKATAKRISHNAN      = 1, /*!< \brief Slope limiter using Venkatakrisnan method (stencil formulation). */
-  VENKATAKRISHNAN_WANG = 2, /*!< \brief Slope limiter using Venkatakrisnan method, eps based on solution (stencil formulation). */
-  BARTH_JESPERSEN      = 3, /*!< \brief Slope limiter using Barth-Jespersen method (stencil formulation). */
-  VAN_ALBADA_EDGE      = 4, /*!< \brief Slope limiter using Van Albada method (edge formulation). */
-  SHARP_EDGES          = 5, /*!< \brief Slope limiter using sharp edges. */
-  WALL_DISTANCE        = 6  /*!< \brief Slope limiter using wall distance. */
+enum class ENUM_LIMITER {
```
pcarruscag, line 890: Remove the ENUM_ prefix please, to keep things shorter.
Sounds like some time of undefined behaviour, have you tried running valgrind?
The AdjLimiter are only for continuous adjoints, one for species is not needed.

bigfooted: thanks for the comment. Haven't tried valgrind yet, I'm not looking forward to doing parallel debugging. But it looks like it might be necessary.

pcarruscag: If it is due to an uninitialized variable it will show up also in serial.

TobiKattmann (commented):
Thanks @bigfooted for enum class-ing the limiters :bouquet: 

I also dont know how and why this fixes the problem you saw but I would also suspect that this doesnt solve the root-cause. So valgrinding a DA case with species limiters would be nice. But in CConfig.cpp::3454 I would certainly add KindSLopeLimit_Species=NONE if !MUSCL_Species. On that note: Did your case had MUSCL_SPECIES with limiters on or off?

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1567,11 +1567,13 @@ su2double CTurbSASolver::GetInletAtVertex(su2double *val_inlet,
...
-
-  for(unsigned long iVertex=0; iVertex < nVertex[iMarker]; iVertex++){
-    Inlet_TurbVars[iMarker][iVertex][0] = GetNuTilde_Inf();
+  /*--- Find BC string to the numeric-identifier. ---*/
+  if (config->GetMarker_All_KindBC(iMarker) == INLET_FLOW) {
+    const string Marker_Tag = config->GetMarker_All_TagBound(iMarker);
+    for (unsigned long iVertex = 0; iVertex < nVertex[iMarker]; iVertex++) {
+      Inlet_TurbVars[iMarker][iVertex][0] = GetNuTilde_Inf();
+    }
   }
```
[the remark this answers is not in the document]
bigfooted: Thanks, I also noticed this.

Common/include/basic_types/ad_structure.hpp
```diff
@@ -538,7 +538,8 @@ namespace AD{
       }
     }
   }
-
+  
+  // Nijso says: This delete_handler needs to be called but I do not know where
```
pcarruscag, line 542: Why does it need to be called?

Common/src/CConfig.cpp
```diff
@@ -3448,10 +3448,10 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
   /*--- Set limiter for no MUSCL reconstructions ---*/
 
-  if ((!MUSCL_Flow) || (Kind_ConvNumScheme_Flow == SPACE_CENTERED)) Kind_SlopeLimit_Flow = NO_LIMITER;
-  if ((!MUSCL_Turb) || (Kind_ConvNumScheme_Turb == SPACE_CENTERED)) Kind_SlopeLimit_Turb = NO_LIMITER;
-  if ((!MUSCL_AdjFlow) || (Kind_ConvNumScheme_AdjFlow == SPACE_CENTERED)) Kind_SlopeLimit_AdjFlow = NO_LIMITER;
-  if ((!MUSCL_AdjTurb) || (Kind_ConvNumScheme_AdjTurb == SPACE_CENTERED)) Kind_SlopeLimit_AdjTurb = NO_LIMITER;
+  if ((!MUSCL_Flow) || (Kind_ConvNumScheme_Flow == SPACE_CENTERED)) Kind_SlopeLimit_Flow = LIMITER::NONE;
+  if ((!MUSCL_Turb) || (Kind_ConvNumScheme_Turb == SPACE_CENTERED)) Kind_SlopeLimit_Turb = LIMITER::NONE;
+  if ((!MUSCL_AdjFlow) || (Kind_ConvNumScheme_AdjFlow == SPACE_CENTERED)) Kind_SlopeLimit_AdjFlow = LIMITER::NONE;
+  if ((!MUSCL_AdjTurb) || (Kind_ConvNumScheme_AdjTurb == SPACE_CENTERED)) Kind_SlopeLimit_AdjTurb = LIMITER::NONE;
```
pcarruscag, line 3454: Same for MUSCL species?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -2250,7 +2250,7 @@ void CSolver::SetGridVel_Gradient(CGeometry *geometry, const CConfig *config) {
 
 void CSolver::SetSolution_Limiter(CGeometry *geometry, const CConfig *config) {
 
-  auto kindLimiter = static_cast<ENUM_LIMITER>(config->GetKind_SlopeLimit());
+  LIMITER kindLimiter = config->GetKind_SlopeLimit();
```
pcarruscag, line 2253:
```suggestion
  const auto kindLimiter = config->GetKind_SlopeLimit();
```

bigfooted:
@pcarruscag  When running SU2_CFD_AD, the solver is DISC_ADJ_INC_RANS, and a number of initializations are skipped when we do not include this check. Most notably setting kind_limiter, which then leads to the 'conditional depends on uninitialized value' message from valgrind.

Common/src/CConfig.cpp
```diff
@@ -8397,11 +8408,11 @@ void CConfig::SetGlobalParam(MAIN_SOLVER val_solver,
         SetKind_TimeIntScheme(Kind_TimeIntScheme_Species);
       }
       if (val_system == RUNTIME_HEAT_SYS) {
-        SetKind_ConvNumScheme(Kind_ConvNumScheme_Heat, NONE, NONE, NONE, NONE, NONE);
+        SetKind_ConvNumScheme(Kind_ConvNumScheme_Heat, NONE, NONE, LIMITER::NONE, NONE, NONE);
         SetKind_TimeIntScheme(Kind_TimeIntScheme_Heat);
       }
       break;
-    case MAIN_SOLVER::RANS: case MAIN_SOLVER::INC_RANS:
+    case MAIN_SOLVER::RANS: case MAIN_SOLVER::INC_RANS: case MAIN_SOLVER::DISC_ADJ_INC_RANS: case MAIN_SOLVER::DISC_ADJ_RANS:
```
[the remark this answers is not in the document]
pcarruscag: I understand but I think this is a red herring, because this problem has not showed up with anything else.
Can you post a config that reproduces the problem?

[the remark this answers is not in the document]
pcarruscag: Hmmm I can kinda see how this would happen, it should be reproducible with turbulence...

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -107,26 +109,24 @@ void CFluidIteration::Iterate(COutput* output, CIntegration**** integration, CGe
...
-      config[val_iZone]->SetGlobalParam(MAIN_SOLVER::RANS, RUNTIME_TRANS_SYS);
+      config[val_iZone]->SetGlobalParam(main_solver, RUNTIME_TRANS_SYS);
       integration[val_iZone][val_iInst][TRANS_SOL]->SingleGrid_Iteration(geometry, solver, numerics, config,
                                                                          RUNTIME_TRANS_SYS, val_iZone, val_iInst);
     }
   }
 
   if (config[val_iZone]->GetKind_Species_Model() != SPECIES_MODEL::NONE){
-    config[val_iZone]->SetGlobalParam(config[val_iZone]->GetKind_Solver(), RUNTIME_SPECIES_SYS);
+    config[val_iZone]->SetGlobalParam(main_solver, RUNTIME_SPECIES_SYS);
```
pcarruscag, line 129: This should be why it was not working, turbulence did not suffer from the same problem.

pcarruscag (approved):

PR #1554, opened 2022-03-04
Fix supersonic inflow/outflow for turbulence and passive scalar solvers

pcarruscag:
I have a v&v test baking for this which I'll add in another PR


TobiKattmann (approved):
lgtm, looking forward to the testcase :)

Issue #1551, opened 2022-02-28
SST(m) implementation is inconsistent with literature


pcarruscag:
TLDR:
There is [SST and SSTm](https://turbmodels.larc.nasa.gov/sst.html), the latter drops the tke term from the viscous stress tensor AND from the production term. Currently we do the former (because it was found beneficial here #797) but not the latter, which leaves the implementation in some kind of limbo state.

### Discussed in https://github.com/su2code/SU2/discussions/1483

<div type='discussions-op-text'>

<sup>Originally posted by **sun5k** January 10, 2022</sup>
Question about Menter's SST model on SU2

Hello! This post is my first Q&A for SU2; happy new Year everyone.
I have two questions about the SST model for the SU2.

The abbreviation is referred from the [SST Turbulence model of NASA TMR](https://turbmodels.larc.nasa.gov/sst.html), [Compressible RANS of NASA TMR](https://turbmodels.larc.nasa.gov/sst.html) , and [Running cases with CFD of NASA TMR](https://turbmodels.larc.nasa.gov/noteonrunning.html). 

Briefly 
1. In SU2, The production term of the SST model uses incompressible assumptions. Is there some special reason? 

2. Always ignore the turbulence kinetic energy to compute the stress tensor for the momentum and energy equation, is it a bug?

Here is the detail!

1.  

The Boussinesq relation of RANS is  

![image](https://user-images.githubusercontent.com/50565314/148731113-5d141cd0-bea0-4bc4-8f2a-e40932645844.png)

Form [Running cases with CFD of NASA TMR](https://turbmodels.larc.nasa.gov/noteonrunning.html): "In the equation above, the![image](https://user-images.githubusercontent.com/50565314/148731540-ace2cac5-4168-48d8-9125-c053806ba463.png) term is sometimes ignored for non-supersonic speed flows, and the **second term in parentheses is identically zero for incompressible flows**."

The production term of SST model is :

![image](https://user-images.githubusercontent.com/50565314/148771167-b96d6d8f-b638-4efd-a6a3-6731ddb4f4a0.png)

Rewrite the first term of the above equation :

![image](https://user-images.githubusercontent.com/50565314/148731731-23cdd717-28d1-453c-a867-77c99494f86f.png)

![image](https://user-images.githubusercontent.com/50565314/148771466-f1613617-0502-4809-9861-a64f4ce0dee7.png)

Finally, the production term becomes :
 
[...]

PR #1557, opened 2022-03-06
[WIP] Feature sst2003m

bigfooted:
## Proposed Changes
edit:
- Implement the SST2004,SST2004m, SST2003 and  SST2003m model according to the NASA website https://turbmodels.larc.nasa.gov/sst.html
- implementation compatible with ongoing work on SA model

## Related Work
#1551 
#1364 

## PR Checklist

### DIFF
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -574,7 +574,7 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
   su2double Jacobian_Buffer[4];  /// Static storage for the Jacobian (which needs to be pointer for return type).
 
   /*!
-   * \brief Get strain magnitude based on perturbed reynolds stress matrix.
+   * \brief Get strain magnitude based on perturbed Reynolds stress matrix.
    * \param[in] turb_ke: turbulent kinetic energy of the node.
    */
   inline su2double PerturbedStrainMag(su2double turb_ke) const {
@@ -732,12 +732,23 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
         StrainMag = PerturbedStrainMag(ScalarVar_i[0]);
       }
 
-      su2double pk = Eddy_Viscosity_i * pow(StrainMag, 2) - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
-      pk = max(0.0, min(pk, 20.0 * beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0]));
+      /*--- Definition of production P according to SSTm approximation P = mu_t S^2 ---*/
 
-      const su2double VorticityMag = GeometryToolbox::Norm(3, Vorticity_i);
-      const su2double zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
-      su2double pw = alfa_blended * Density_i * max(pow(StrainMag, 2) - 2.0 / 3.0 * zeta * diverg, 0.0);
+      /*--- original SST2003 model as reference (including divergence terms) ---*/
+      // su2double pk = Eddy_Viscosity_i * (pow(StrainMag, 2) - 2.0 / 3.0 * diverg * diverg) - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
+      /*--- SST2003m model, neglecting divergence terms ---*/
+      su2double pk = Eddy_Viscosity_i * pow(StrainMag, 2);
+      su2double pw = (alfa_blended * Density_i) * pk;  
+
+      /*--- Production limiter for k (and not for w) according to SST2003m ---*/
+      pk = min(pk, 10.0 * beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0]);
+
+      // const su2double VorticityMag = GeometryToolbox::Norm(3, Vorticity_i);
+
+      /*--- denominator of turbulent eddy viscosity computation according to SST2003m: mu_t = rho*k/(max(w,SF_2/a1)) ---*/
+      const su2double zeta = max(ScalarVar_i[1], StrainMag * F2_i / a1);
+      //su2double pw = alfa_blended * Density_i * max(pow(StrainMag, 2) - 2.0 / 3.0 * zeta * diverg, 0.0);
+      //su2double pw = (alfa_blended * Density_i) * pk;  
 
       /*--- Sustaining terms, if desired. Note that if the production terms are
             larger equal than the sustaining terms, the original formulation is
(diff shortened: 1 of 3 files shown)


bigfooted:
I looked at the transport equations and implemented the SST2003m model (as far as I can tell)
I'm not entirely sure about what happens in CTurbSSTSolver::Postprocessing
what's the difference between:
flownodes->
nodes->
geometry->nodes

I also have to check the implementation of the boundary conditions

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -205,7 +209,7 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
   }
 
   AD::StartNoSharedReading();
-
+  // nijso asks: why do we do this?
   auto* flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
```
pcarruscag, line 213: Because it allows de-virtualizing the access to the flow primitives and gradients.
```suggestion
  auto* flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -216,10 +220,15 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
     su2double rho = flowNodes->GetDensity(iPoint);
     su2double mu  = flowNodes->GetLaminarViscosity(iPoint);
 
+    // nijso asks: difference between geometry->nodes->GetWall_Distance and nodes->GetWall_Distance ???
     su2double dist = geometry->nodes->GetWall_Distance(iPoint);
 
-    su2double VorticityMag = GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint));
-    VorticityMag = max(VorticityMag, 1e-12); // safety against division by zero
+    // nijso asks: difference between nodes and flownodes ???
```
pcarruscag, line 226: nodes are the turbulent variables, flowNodes are the flow variables.

SU2_CFD/src/variables/CTurbSSTVariable.cpp
```diff
@@ -66,7 +66,7 @@ void CTurbSSTVariable::SetBlendingFunc(unsigned long iPoint, su2double val_visco
   for (unsigned long iDim = 0; iDim < nDim; iDim++)
     CDkw(iPoint) += Gradient(iPoint,0,iDim)*Gradient(iPoint,1,iDim);
   CDkw(iPoint) *= 2.0*val_density*sigma_om2/Solution(iPoint,1);
-  CDkw(iPoint) = max(CDkw(iPoint), pow(10.0, -20.0));
+  CDkw(iPoint) = max(CDkw(iPoint), pow(10.0, -10.0));
```
pcarruscag, line 69:
```suggestion
  CDkw(iPoint) = max(CDkw(iPoint), 1e-10);
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -763,17 +774,17 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      /*--- Cross diffusion for w ---*/
 
       Residual[1] += (1.0 - F1_i) * CDkw_i * Volume;
 
       /*--- Contribution due to 2D axisymmetric formulation ---*/
 
       if (axisymmetric) ResidualAxisymmetric(alfa_blended, zeta);
 
-      /*--- Implicit part ---*/
+      /*--- Implicit part (includes contributions from P ---*/
```
pcarruscag, line 785:
```suggestion
      /*--- Implicit part (includes contributions from P) ---*/
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -732,12 +732,23 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
-      const su2double zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
-      su2double pw = alfa_blended * Density_i * max(pow(StrainMag, 2) - 2.0 / 3.0 * zeta * diverg, 0.0);
+      /*--- original SST2003 model as reference (including divergence terms) ---*/
+      // su2double pk = Eddy_Viscosity_i * (pow(StrainMag, 2) - 2.0 / 3.0 * diverg * diverg) - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
+      /*--- SST2003m model, neglecting divergence terms ---*/
+      su2double pk = Eddy_Viscosity_i * pow(StrainMag, 2);
+      su2double pw = (alfa_blended * Density_i) * pk;  
+
+      /*--- Production limiter for k (and not for w) according to SST2003m ---*/
+      pk = min(pk, 10.0 * beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0]);
```
pcarruscag, line 744: The limiter applies to both terms. See https://turbmodels.larc.nasa.gov/sst.html second paragraph in Menter SST Two-Equation Model from 2003

pcarruscag (commented):
Good start :+1: 
Of course this means we'll have to go to SU2 v8 :grimacing: 
@jayantmukho anything we need to be aware here to avoid damaging UQ? Do we need to remove the `tke` term also from the perturbed Reynolds stress tensor?

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
bigfooted: Yes, I now found a direct reference from Menter where he corrected this: http://dx.doi.org/10.1080/10618560902773387

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
bigfooted: ah of course, thanks!

TobiKattmann:
> make the SST model exactly the SST2003m model according to the NASA website

Does it make sense to change the existing SST model rather than just adding the 2003m-version alongside of the 1994-version? Maybe rename to `SST-1994` and `SST-2003m`, that makes it transparent. One could even keep `SST` while adding a warning that `SST` will be deprecated in the future and that `SST-1994` is the equivalent successor.

The upside of doing it like that is that it wouldn't interfere that much with peoples cases (only little cfg change necessary to produce the same results) and it would be much less work for you to change all the regression tests.

And although the 2003-version is supposed to be an update to the 1994 version, I would bet there are people who prefer the 1994 version by choice. After all, when the Ancients first implemented SU2 they chose 1994 despite the 2003 version being around for 10 years already.

I would be ok with switching to 2003m completely but I dont see really see the benefit as it means more work for everybody (right?).

bigfooted:
Hi Tobi, Well the thing is that currently we do not have sst1994 nor sst2003. We can implement both version and make one of them the default but in any case it means that the regression tests have to be recalibrated. And if you want to update 'everything' then we also need to redo the V&V cases and make the plots again - preferably with something in place that will auto-create these plots when something changes in the future.

PR #1559, opened 2022-03-10
Improve the configuration file for the SA Neg test case

suargi:
## Proposed Changes
Improve the configuration file for the turbulence_models SA Neg test case. With this new config file we should be better able to detect bugs in the code. For instance, with this new config file and number of iterations for the regression test, we will be able to detect bd4021020147e8e03d519cad2ddbaeef10c0bf3d.

## PR Checklist


TestCases/parallel_regression.py
```diff
@@ -673,8 +673,8 @@ def main():
     turbmod_sa_neg_rae2822           = TestCase('turbmod_sa_neg_rae2822')
     turbmod_sa_neg_rae2822.cfg_dir   = "turbulence_models/sa/rae2822"
     turbmod_sa_neg_rae2822.cfg_file  = "turb_SA_NEG_RAE2822.cfg"
-    turbmod_sa_neg_rae2822.test_iter = 20
-    turbmod_sa_neg_rae2822.test_vals = [-2.004689, 0.742306, 0.497308, -5.265793, 0.809463, 0.062016]
+    turbmod_sa_neg_rae2822.test_iter = 10
```
TobiKattmann, line 676: I am curious why a reduced number of iterations would be beneficial? I would always tend to do more as small errors might accumulate enough only after a certain amount of iterations, to break the regression test. That scenario might not happen to frequently but as you also remove multigrid there should be some time freed-up, right :)

So feel free (from my point of view) to revert to 20 or higher if it doesnt cost minutes.

TestCases/turbulence_models/sa/rae2822/turb_SA_NEG_RAE2822.cfg
```diff
@@ -164,4 +168,4 @@ SURFACE_FILENAME= surface_flow
 OUTPUT_WRT_FREQ= 250
 %
 % Screen output fields 
-SCREEN_OUTPUT= (INNER_ITER, RMS_DENSITY, RMS_MOMENTUM-X, RMS_MOMENTUM-Y, RMS_NU_TILDE, LIFT, DRAG)
+SCREEN_OUTPUT= (INNER_ITER, RMS_DENSITY, RMS_MOMENTUM-X, RMS_MOMENTUM-Y, RMS_NU_TILDE, LIFT, DRAG)
```
TobiKattmann, line 171:
```suggestion
SCREEN_OUTPUT= (INNER_ITER, RMS_DENSITY, RMS_MOMENTUM-X, RMS_MOMENTUM-Y, RMS_NU_TILDE, LIFT, DRAG)

```

TobiKattmann: `No newline at end of file` warning fix

TestCases/turbulence_models/sa/rae2822/turb_SA_NEG_RAE2822.cfg
```diff
@@ -14,7 +14,7 @@ RESTART_SOL= NO
 MACH_NUMBER= 0.729
 %
 % Angle of attack (degrees, only for compressible flows)
-AOA= 2.31
+AOA= 20.31
```
TobiKattmann, line 17: I recon, the case still converges to a reasonable solution with that more aggressive AoA?

TobiKattmann (approved):

suargi:
Thank you @TobiKattmann for your feedback.

The idea behind this new regression test config file is as follows:
We should have a test case that triggers the negative part of the SA model. Depending on the flow conditions, geometry and CFD parameters it might be triggered or not. With the previous config file, it was not. 
I think there is already a regression test in SU2 that triggers the negative SA, the `turb_oneram6_nk`. However, I decided to stick to rae2822 airfoil as it is a simpler (faster) case. In order to force the negative part of the SA for the rae2822 in a reasonable amount of iterations for a regression test, I increased the angle of attack. I have removed the multigrid as it might not be stable, but I have not tested though. Anyway, with the current configuration, e.g., convective scheme, CFL number, etc, the solution is not stable and diverges after some iterations, around 15. In that sense, I reduced the number of iterations from 20 to 10.

In my opinion, a diverging regression test is not a problem at all as it might not be used as a tutorial, only to verify the integrity of the commit. "The solution should always diverge to the same results". If the regression test should converge, let me know and I will update the config file :)

pcarruscag:
That is a bit of a necessary evil, I would leave a warning in the config explaining exactly what you just did.

TobiKattmann:
Thanks for the explanation @suargi . I would personally advocate for that the testcases should converge to some reasonable solution people might use it as a starting point (copy the cfg and doing mild adaptions) for their own stuff. And the Testcases show off the capabilities to some degree, to do so, convergence is beneficial.

But as we have a bunch of working  2D airfoils in regression already I recon that adding a clear explanation and warning to the cfg as suggested by Pedro is fine. Otherwise you might try to bisect the AoA ... maybe there is a value that triggers negative SA and does not diverge :thinking: 

Knowingly adding a diverging test without a clear warning is not good imo :)

suargi:
Header with a warning included in commit 017616f30e0dbe070447dd704d83a00e71ed2342.

suargi:
Is it ready to merge?

TestCases/turbulence_models/sa/rae2822/turb_SA_NEG_RAE2822.cfg
```diff
@@ -1,3 +1,19 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+%                                  WARNING                                     %
+%                      This regression test will diverge!                      %
+%									       %
+% This test case is only meant to trigger the negative part of the SA model,   %
+% to check the integrity of the commits. It is inadvisable to use it as an     %
+% starting point for tutorials.						       %
+% The rae2822 airfoil is a simple (fast) case. A high angle of attack is being %
```
TobiKattmann, line 9: For some reason github `Files changed` tab shows me a messed up indentation ... but looking at the actual file the indentation is on point. Not sure why github does that :shrug:

TobiKattmann: hmm ... in chromium that looks good :thinking:

TobiKattmann: image

TobiKattmann:
> Is it ready to merge?

From my side yes :) Can you though make sure the reg test run and report here in the PR correctly. I am not really what happened there. I started a re-run of all the jobs so if that runs like usual feel free to merge :+1:

TestCases/turbulence_models/sa/rae2822/turb_SA_NEG_RAE2822.cfg
```diff
@@ -61,7 +77,11 @@ NUM_METHOD_GRAD= WEIGHTED_LEAST_SQUARES
 CFL_NUMBER= 2.5
 %
 % Adaptive CFL number (NO, YES)
-CFL_ADAPT= NO
+CFL_ADAPT= YES
+%
+% Parameters of the adaptive CFL number (factor down, factor up, CFL min value,
```
TobiKattmann, line 82:
```suggestion
% Parameters of the adaptive CFL number ( factor down, factor up, CFL min value,
```

suargi:
I do not know why a regression test from hybrid_regression_AD.py has failed. But it has nothing to do with this pull request. Therefore I think we can merge it.

TobiKattmann:
> I do not know why a regression test from hybrid_regression_AD.py has failed. But it has nothing to do with this pull request. Therefore I think we can merge it.

Yep sometimes certain cases of hybrid-reg-ad fail ... sometimes they don't :) 
That is why they are not a hard requirement for a merge. Usually I take a look nonetheless to not overlook sth.
So feel free to merge 👍

PR #1560, opened 2022-03-11
Refactor SST model implementation and addition of variants/corrections

WallyMaier:
## Proposed Changes
This PR refactors the SST model implementation to include (hopefully) the KL, RC, V, m, SUST models.  
Mainly for the models SST-V(m)  and SST-Vsust(-m).  These tend to be used in higher speed flows. https://turbmodels.larc.nasa.gov/sst.html

A quick summary of some of the changes are:

Adding the"V" term simply changes the P-term from
`Eddy_Viscosity_i * pow(StrainMag,2)`  to
`Eddy_Viscosity_i * pow(VorticityMag,2)`

Adding the "KL" source term changes the P-term to
`Eddy_Viscosity_i *StrainMag*VorticityMag`

Adding the "RC" term:
Multiplying the P-term by a function, f. 

Many of the Options can be used together -> SST-V-sust-m.  The NASA TMR doesn't explicitly mention something like SST-2003-V-RC-sust-m,so I don't know if that is a valid permutation.  That being said, the amount of options is fairly so I'm not sure the best way to approach this formulation, without unnecessarily bloating the code.  Hence this isn't operational yet.  Work will need to be done to validate the addition.

## Related Work
This is part of the larger turbulence fixes/additions/modernization efforts seen in #1557.  It also addresses some comments in #1551 and #1413

Also as @pcarruscag pointed out below, this falls out in the larger discussion mentioned in #1364.

## PR Checklist


SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -732,12 +733,16 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
         StrainMag = PerturbedStrainMag(ScalarVar_i[0]);
       }
```
pcarruscag, line 734:
```suggestion
      } else if (vorticity_source) {
        StrainMag = VorticityMag;
      }
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -732,12 +733,16 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
       const su2double zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
-      su2double pw = alfa_blended * Density_i * max(pow(StrainMag, 2) - 2.0 / 3.0 * zeta * diverg, 0.0);
+
+      if (vorticity_source) {su2double p_base = Eddy_Viscosity_i * pow(VorticityMag,2); }
+      else                  {su2double p_base = Eddy_Viscosity_i * pow(StrainMag,2); }
+
+      su2double pk = p_base - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
+      pk = max(0.0, min(pk, 20.0 * beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0]));
+
+      su2double pw = (alfa_blended * Density_i) * pk;
```
pcarruscag, line 745: This is not consistent with SST-1994, there is no production limiter for `w`, in SST-2003 both are limited.

pcarruscag (commented):
The direction we are going is to have turbulence model variants/corrections specified according to the strategy in #1364 
```
TURB_MODEL= SST
SST_OPTIONS= 1994/2003, m, V, KL, RC, sust, UQ
```
The code change for V is the 3 lines below, if you want to start the implementation of "SST_OPTIONS" that would be awesome.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
WallyMaier: Good point.  I'm still catching up to speed on the current implementation and work.

WallyMaier:
Ill go ahead and start the SST_OPTIONS option.  I imagine with the amount of variants/corrections, we will need to follow the SA implementation style from @suargi ?

Common/include/option_structure.hpp
```diff
@@ -912,51 +912,52 @@ static const MapType<std::string, ENUM_LIMITER> Limiter_Map = {
...
   MakePair("SA", TURB_MODEL::SA)
-  MakePair("SA_NEG", TURB_MODEL::SA_NEG)
-  MakePair("SA_E", TURB_MODEL::SA_E)
-  MakePair("SA_COMP", TURB_MODEL::SA_COMP)
-  MakePair("SA_E_COMP", TURB_MODEL::SA_E_COMP)
   MakePair("SST", TURB_MODEL::SST)
-  MakePair("SST_SUST", TURB_MODEL::SST_SUST)
 };
 
-/*!
```
WallyMaier, line 933: This definitely wont work as written.  Just trying to gauge if this is the direction we want to go in

Common/include/option_structure.hpp
```diff
@@ -912,51 +912,52 @@ static const MapType<std::string, ENUM_LIMITER> Limiter_Map = {
...
- */
-enum class TURB_FAMILY {
-  NONE,   /*!< \brief No turbulence model. */
-  SA,     /*!< \brief Spalart-Allmaras variants. */
-  KW,     /*!< \brief k-w models. */
+enum class SA_OPTIONS {
+  NONE,     /*!< \brief Stadard Spalart-Allmaras (SA_ Model. */
+  NEG,      /*!< \brief Negative SA model. */
+  E,        /*!< \brief SA with Edwards modificiation. */
+  COMP,     /*!< \brief SA model with no tau_ij in momentum and energy eqns. */
```
WallyMaier, line 927: wrong brief.

Common/include/option_structure.hpp
```diff
@@ -912,51 +912,52 @@ static const MapType<std::string, ENUM_LIMITER> Limiter_Map = {
...
+  MakePair("NEG", SA_OPTION::NEG)
+  MakePair("E", SA_OPTION::E)
+  MakePair("COMP", SA_OPTION::COMP)
+  MakePair("E_COMP", SA_OPTION::E_COMP)
+};
+
+/*!
+ * \brief Variants and corrections of SST k-w models
+ */
+enum class SST_OPTIONS {
```
WallyMaier, line 941: This causes some problems since (at least as Im aware) the SST options can be used in almost set of permutations.  This is slightly different to SA.

If we want to allow for these types of modifications, In config.cpp, I propose to something similiar to #1413, but the opposite.  i.e We take in a list of the options and create/name the desired model from there.  This would make a large decision tree however.

Common/src/CConfig.cpp
```diff
@@ -5165,12 +5177,12 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
   else {monoatomic = false;}
 
   /*--- Set number of Turbulence Variables. ---*/
-  switch (TurbModelFamily(Kind_Turb_Model)) {
-    case TURB_FAMILY::NONE:
+  switch (Kind_Turb_Model) {
```
WallyMaier, line 5180: in the proposed change, TURB_FAMILY would be unnecessary since we will only have NONE, SA or SST turb models.

Common/include/option_structure.hpp
```diff
@@ -912,51 +912,52 @@ static const MapType<std::string, ENUM_LIMITER> Limiter_Map = {
 enum class TURB_MODEL {
   NONE,      /*!< \brief No turbulence model. */
   SA,        /*!< \brief Kind of Turbulent model (Spalart-Allmaras). */
-  SA_NEG,    /*!< \brief Kind of Turbulent model (Spalart-Allmaras). */
-  SA_E,      /*!< \brief Kind of Turbulent model (Spalart-Allmaras Edwards). */
-  SA_COMP,   /*!< \brief Kind of Turbulent model (Spalart-Allmaras Compressibility Correction). */
```
pcarruscag, line 917: Let's not get ahead of ourselves here :smile: we cannot simply remove these options from users, see the backwards compatibility discussion in #1364 
Start with only SST_OPTIONS for now please, we can even start with just V.

Common/include/option_structure.hpp
pcarruscag: We don't want to unfold these options into turbulence models. What we can do is read an enum list from the config, as you have started defining, then we convert that list into a small struct with sst boolean options and we validate that they are compatible.
```
Version: enum - 1994 or 2003
Modified: bool - m
Production: enum - V or KL or UQ
Sustaining terms: bool - sust
Curvature correction: bool - RC
```
Then we use that struct in the code instead of deciding what to do based on kind turb model.

Common/include/option_structure.hpp
```diff
@@ -958,6 +956,20 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
   return TURB_FAMILY::NONE;
 }
 
+/*!
+ * \brief SST Base Model
+ */
+enum class SST_BASE {
```
WallyMaier, line 962: Im not sure how the actual code/ enum list would look like in EnumList for the config.  Would we need an option for each option

Common/src/CConfig.cpp
```diff
@@ -1093,8 +1093,12 @@ void CConfig::SetConfig_Options() {
 #endif
   /*!\brief MATH_PROBLEM  \n DESCRIPTION: Mathematical problem \n  Options: DIRECT, ADJOINT \ingroup Config*/
   addMathProblemOption("MATH_PROBLEM", ContinuousAdjoint, false, DiscreteAdjoint, discAdjDefault, Restart_Flow, discAdjDefault);
+
   /*!\brief KIND_TURB_MODEL \n DESCRIPTION: Specify turbulence model \n Options: see \link Turb_Model_Map \endlink \n DEFAULT: NONE \ingroup Config*/
   addEnumOption("KIND_TURB_MODEL", Kind_Turb_Model, Turb_Model_Map, TURB_MODEL::NONE);
+  /*!\brief SST_OPTIONS \n DESCRIPTION: Specify SST turbulence model options/corrections. \n Options: see \link SST_Options_Map \endlink \n DEFAULT: NONE \ingroup Config*/
+  addEnumListOption("SST_OPTIONS", nSST_Options, SST_Options_Map, SST_BASE::V1994, SST_MODIFIED::YES, SST_PROUDCTIONS::NONE, SST_SUST::NONE, SST_CURVE::NONE );
```
WallyMaier, line 1100: Would this necessitate a new config option base?

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -722,22 +722,28 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      /*--- Modifications of SST production terms (uq, V, KL) --*/
       if (using_uq) {
         ComputePerturbedRSM(nDim, Eig_Val_Comp, uq_permute, uq_delta_b, uq_urlx, PrimVar_Grad_i + idx.Velocity(),
                             Density_i, Eddy_Viscosity_i, ScalarVar_i[0], MeanPerturbedRSM);
         StrainMag = PerturbedStrainMag(ScalarVar_i[0]);
-      }
 
-      su2double pk = Eddy_Viscosity_i * pow(StrainMag, 2) - 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
+      } (vorticity_source) {
+        StrainMag = VorticityMag;
```
WallyMaier, line 736: This may be better as "p_base" to be more clear of what is happening.

Common/include/option_structure.hpp
```diff
@@ -958,6 +956,20 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  NONE,   /*!< \brief No SST Turb model. */
+  V1994,  /*!< \brief 1994 Menter k-w SST model. */
+  V2003,  /*!< \brief 2003 Menter k-w SST model. */
+};
+static const MapType<std::string, SST_BASE> SST_Base_Map = {
+  MakePair("NONE", SST_BASE::NONE)
+  MakePair("v1994", SST_BASE::V1994)
+  MakePair("v2003", SST_BASE::V2003)
```
bigfooted, line 970:
```suggestion
  MakePair("NONE", SST_BASE::NONE)
  MakePair("v1994", SST_BASE::V1994)
  MakePair("v2003", SST_BASE::V2003)
  MakePair("modified", SST_BASE::MODIFIED)
  MakePair("sust", SST_BASE::SUST)
  MakePair("vorticity", SST_BASE::VORTICITY)
  MakePair("curvature", SST_BASE::RC)
```

Common/src/CConfig.cpp
bigfooted: the enum class list can act as the list of boolean switches for all the models, so we would not need additional SST_PRODUCTIONS::, SST_SUST:: etc. 
Since they are usually one-line modifiers I would do the implementation directly in turb_sources

Common/include/CConfig.hpp
```diff
@@ -705,6 +705,8 @@ class CConfig {
 
   unsigned short  nConfig_Files;          /*!< \brief Number of config files for multiphysics problems. */
   string *Config_Filenames;               /*!< \brief List of names for configuration files. */
+  string *SST_Options;                    /*!< \brief List of modifications/corrections/versions of SST turbulence model.*/     
+  string *SA_Options;                     /*!< \brief List of modifications/corrections/versions of SA turbulence model. */
```
pcarruscag, line 709: This need to be `SST_OPTION*` not string.

Common/src/CConfig.cpp
WallyMaier: So after trying to implement this Im finding a peculiar issue:
`SST_Options_Map.find("OPTION")`  yields a type mismatch/conversion issue.

Common/src/CConfig.cpp
```diff
@@ -5175,6 +5183,35 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
+
+    //PRODUCTION INCOMPABILITY
+    unsigned short counter = 0;
+    if (SST_Options.Map_find("VORTICITY")) counter += 1;
+    if (SST_Options.Map_find("KATO-LAUNDER")) counter += 1;
+    if (SST_Options.Map_find("UQ")) counter += 1;
+    if (counter > 1) SU2_MPI::Error(string("Please select only one SST production term modifier (V, KL, UQ)."), CURRENT_FUNCTION);
+
+  };
+  //if (SST_Options_Map.find("VORTICITY")) cout <<"IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII"<<endl;
```
WallyMaier, line 5213: This doesnt seem to work.

Common/src/CConfig.cpp
```diff
@@ -5175,6 +5183,35 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
       nTurbVar = 2; break;
   }
 
+  /*--- Postprocess SST_OPTIONS into structure. ---*/
+  //TODO this wont work.
+  string sst_string = "SST";
+  struct SST_Struct {
```
WallyMaier, line 5189: Im not entirely sure how create this structure.  But I imagine I can pass this structure into the turb solver like how we currently pass the sustaining_terms bool?

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -636,7 +639,7 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
                            su2double val_omega_Inf, const CConfig* config)
       : CNumerics(val_nDim, 2, config),
         idx(val_nDim, config->GetnSpecies()),
-        sustaining_terms(config->GetKind_Turb_Model() == TURB_MODEL::SST_SUST),
+        sustaining_terms(false),      //TODO this is temporary
```
WallyMaier, line 642: could this be replaced with the structure? or would we template the source terms?

Common/src/CConfig.cpp
pcarruscag: The map is a static "look-up" to convert user strings into enum options, the options (enums) are read into SST_Options, not into SST_Options_Map.
So what you need is:
```c++
const auto sst_options_end = SST_Options + nSST_Options;
const bool sst_v = std::find(SST_Options, sst_options_end, SST_OPTIONS::VORTICITY) != sst_options_end;
```

Common/src/CConfig.cpp
pcarruscag, line 5189: declare and define the struct in option_structure, close to SST_OPTIONS.
Then create a free function in there that populates the struct based on the SST_Options array (you pass the pointer and size to the function and it returns the struct.
Here in CConfig you put a member variable of the type of that struct, e.g. SST_ParsedOptions, and then do:
SST_ParsedOptions = ParseSSTOptions(SST_Options, nSST_Options);
Finally you have a method to return this SST_ParsedOptions to use it in sources, etc.

bigfooted:
I moved some things around to make it run. I also added sst1994, sst1994m,sst2003,sst2003m.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -637,7 +640,7 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
```
WallyMaier, line 643: Should we pass the created structure here in a similar manner?

WallyMaier:
With the pushes from @bigfooted,  this is mostly complete.  I believe all options are now implemented, except for the rotational correction.

What should the plan for V&V/testing be?
I think we need to also address all the regressions....since no SST case should work now.

bigfooted:
For the V&V, I had a look at the turbulent flat plate case.I think we should create a couple of python scripts to run the cases and produce the figures from the result automatically. It should then be easy to create different figures, comparing SA and SST, SST1994 vs SST2003, etc. 

Maybe somebody already has some scripts to produce the V&V figures? 

We can split the flat plate V&V into compressible/incompressible and SA/SST. When this looks good, we can modify the regression test results.

pcarruscag:
Guys, we cannot simply change defaults like that, update regressions, and call it a day... Even fixing #1551 is a major change that should warrant a major version update. We want SU2 users to be able to rely and trust the code we release...
That is why I suggested that this PR should be used only to change the way of specifying SST options, and introduce simple ones like the V and KL modifications. Then the validation work for SST 2003 (with and w/o modification) would be done in #1557.
But ok, let me look at this and propose a way forward that gets in develop ASAP, **please don't start updating regressions**.

Common/src/CConfig.cpp
```diff
@@ -7734,7 +7737,25 @@ CConfig::~CConfig() {
...
+    delete[] IntInfo_WallFunctions;
+  }
+
+  if (DoubleInfo_WallFunctions != nullptr) {
+    for (iMarker = 0; iMarker < nMarker_WallFunctions; ++iMarker) {
+      if (DoubleInfo_WallFunctions[iMarker] != nullptr)
+        delete[] DoubleInfo_WallFunctions[iMarker];
+    }
+    delete[] DoubleInfo_WallFunctions;
+  }
```
pcarruscag, line 7758: Bad merge, also not need to delete list-type options anymore #1579 
```suggestion
  delete[] Marker_All_SendRecv;
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -711,34 +709,82 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      /* SST-1994 */
+      if (sstParsedOptions.version == SST_OPTIONS::V1994){
+        ProdLimConstant = 20.0; 
+        P_Base = VorticityMag;
+        /* mut = rho*k/max(w,Omega*F_2/a1) = rho*k/zeta */
+      }
+      else {
+        /* SST-2003 */  
+        ProdLimConstant = 10.0;
+        P_Base = StrainMag_i;
```
pcarruscag, line 760: This implementation is incorrect, the production modifications take precedence over the base version.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -747,13 +793,19 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
         const su2double sust_k = beta_star * Density_i * kAmb * omegaAmb;
         const su2double sust_w = beta_blended * Density_i * omegaAmb * omegaAmb;
         pk = max(pk, sust_k);
         pw = max(pw, sust_w);
       }
 
+      /*--- Rotation/ Curvature Corrections ---*/
+      if (sstParsedOptions.rc){
+        //TODO multiply the Ps by some fun stuff that I dont want to implement at the moment.
+
```
pcarruscag, line 806: That is fine, it does not look trivial, remove the RC from the options.

Common/include/option_structure.hpp
```diff
@@ -952,19 +950,59 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  MakePair("KATO-LAUNDER", SST_OPTIONS::KL)
+  MakePair("UNCERTAINTY", SST_OPTIONS::UNCERTAINTY)
+  MakePair("CURVATURE", SST_OPTIONS::RC)
+};
+
+/*!
+ * \brief Structure containing parsed SST options.
+ */
+struct SST_ParsedOptions {
+  std::string sst_string = "Menter's k-w SST"; /*!< \brief Initial string for SST problems. */
```
pcarruscag, line 988: remove this string member please, the objective for the struct is to be light weight, you can still produce the string while parsing the options, it only needs to be printed by CConfig right?

Common/include/option_structure.hpp
```diff
@@ -952,19 +950,59 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  VORTICITY,   /*!< \brief Menter k-w SST model with vorticity production terms. */
+  KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
+  UNCERTAINTY, /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+  RC,          /*!< \brief Menter k-w SST model with rotation/curvature corrections. */
+};
+static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
+  MakePair("NONE", SST_OPTIONS::NONE)
+  MakePair("V1994", SST_OPTIONS::V1994)
+  MakePair("V2003", SST_OPTIONS::V2003)
+  MakePair("MODIFIED", SST_OPTIONS::MODIFIED)
```
pcarruscag, line 976: For fast integration I think the way forward is:
- Keep the inconsistent 1994 as the default.
- Allow 2003 as option but do not expose MODIFIED, we will use modified by default, and in the future we can add `NOT_MODIFIED` to the options, this saves you the trouble of implementing the changes needed to the stress tensor. Moreover we should have robust default options, and the -m versions are better for that, probably.
- The 2003m would of course not feature the inconsistency from #1551, and we can issue a warning when defaulting to 1994 that users should prefer 2003 which will eventually become the default.

If having "NOT_MODIFIED" in the future does not sound good and alternative is to have V1994m V2003m now, and then add V1994 and V2003.
```suggestion
  MakePair("V1994", SST_OPTIONS::V1994)
  MakePair("V2003", SST_OPTIONS::V2003)
```

bigfooted:
> Guys, we cannot simply change defaults like that, update regressions, and call it a day... Even fixing #1551 is a major change that should warrant a major version update. We want SU2 users to be able to rely and trust the code we release... That is why I suggested that this PR should be used only to change the way of specifying SST options, and introduce simple ones like the V and KL modifications. Then the validation work for SST 2003 (with and w/o modification) would be done in #1557. But ok, let me look at this and propose a way forward that gets in develop ASAP, **please don't start updating regressions**.

I think we should treat the current SST model implementation as a bug that must be fixed. The original model is closest to the original SST-1994 model, so we could keep that as the default model. For now, we can also introduce 'legacy SST' as an option.
Almost all SST variants are one-liner modifications, so I don't think it is a big issue to add this, but they all need validation to make sure that they are working as intended.

Common/include/option_structure.hpp
pcarruscag: Only inconsistent 1994m and proper 2003m, for now.
We warn that users should change to 2003m and that we will make it the default (release notes, CFD-online, we can tweet it as well...).
Then once we have the opportunity to make breaking changes we **fix** 1994m and we make 2003m the default.
If/when proper 1994 / 2003 are implemented does not matter since they'll be new, and for that reason they also don't need to be tackled now.

bigfooted: sounds good, and a reasonable work package

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -711,34 +709,82 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
 
+    su2double ProdLimConstant;
+    su2double zeta;
+
     /*--- Computation of blended constants for the source terms ---*/
 
     const su2double alfa_blended = F1_i * alfa_1 + (1.0 - F1_i) * alfa_2;
     const su2double beta_blended = F1_i * beta_1 + (1.0 - F1_i) * beta_2;
 
+    SST_ParsedOptions sstParsedOptions = config->GetSSTParsedOptions();
```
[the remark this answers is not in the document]
bigfooted: I now have it inCTurbSSTVariables as well as CTurbSSTSolver, I did not find a nice way to define it in only one header.

Common/src/CConfig.cpp
```diff
@@ -5875,7 +5889,7 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
           case SA_ZDES:  cout << "Delayed Detached Eddy Simulation (DDES) with Vorticity-based SGS" << endl; break;
           case SA_EDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
         }
-        if (using_uq){
+        if (sstParsedOptions.production == SST_OPTIONS::UNCERTAINTY){
```
WallyMaier, line 5892: Why did these change?  Is using_uq not set properly before hand?

bigfooted: I'm trying to get all regression tests running again, but the results are different between my machine and github, so that's why I pushed this intermediate solution to see what the differences are. I will fix it this weekend and make sure that the uq+sust+sst regressions are working with the sstparsedoptions implementation.

Common/include/option_structure.hpp
```diff
@@ -958,13 +958,55 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  SUST,        /*!< \brief Menter k-w SST model with sustaining terms. */
+  VORTICITY,   /*!< \brief Menter k-w SST model with vorticity production terms. */
+  KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
+  UNCERTAINTY, /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+  RC,          /*!< \brief Menter k-w SST model with rotation/curvature corrections. */
+};
+static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
+  MakePair("NONE", SST_OPTIONS::NONE)
+  MakePair("V1994m", SST_OPTIONS::V1994)
+  MakePair("V2003m", SST_OPTIONS::V2003)
```
pcarruscag, line 978: If you add "m" also to the enum you can then convert internally to 1994/2003 + the "m" boolean.
```suggestion
  MakePair("V1994m", SST_OPTIONS::V1994m)
  MakePair("V2003m", SST_OPTIONS::V2003m)
```

Common/include/CConfig.hpp
```diff
@@ -9617,10 +9620,73 @@ class CConfig {
...
+    const bool sst_1994  = std::find(SST_Options, sst_options_end, SST_OPTIONS::V1994) != sst_options_end;
+    const bool sst_2003 = std::find(SST_Options, sst_options_end, SST_OPTIONS::V2003) != sst_options_end;
+
+    /* when V2003 is selected, we automatically select sst_m as well */
+    const bool sst_m     = std::find(SST_Options, sst_options_end, SST_OPTIONS::MODIFIED) != sst_options_end || sst_2003;
+
+    const bool sst_sust  = std::find(SST_Options, sst_options_end, SST_OPTIONS::SUST) != sst_options_end || Kind_Turb_Model==TURB_MODEL::SST_SUST;
+    const bool sst_v     = std::find(SST_Options, sst_options_end, SST_OPTIONS::VORTICITY) != sst_options_end;
+    const bool sst_kl    = std::find(SST_Options, sst_options_end, SST_OPTIONS::KL) != sst_options_end;
+    const bool sst_uq    = (std::find(SST_Options, sst_options_end, SST_OPTIONS::UNCERTAINTY) != sst_options_end || using_uq);
```
pcarruscag, line 9653: If you want to support the old way of specifying things, I would modify the SST_Options list upstream.

Common/src/CConfig.cpp
```diff
@@ -5197,6 +5194,20 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
       nTurbVar = 2; break;
   }
 
+  /*--- Postprocess SST_OPTIONS into structure. ---*/
+  if (Kind_Turb_Model==TURB_MODEL::SST || Kind_Turb_Model==TURB_MODEL::SST_SUST){
+    sstParsedOptions = ParseSSTOptions(SST_Options, nSST_Options);
```
pcarruscag, line 5199: So, here I would copy SST_Options into a std vector and push_back SUST (if SST_SUST) and UQ (if using_uq), then you pass the vector for parsing, duplicated don't matter.

Common/include/CConfig.hpp
```diff
@@ -9617,10 +9620,73 @@ class CConfig {
...
+   * \return SST option data structure.
+   */
+  SST_ParsedOptions GetSSTParsedOptions(void) const {return sstParsedOptions; }
+
+  /*!
+   * \brief function to parse SST options.
+   * \param[in] SST_Options - Selected SST option from config.
+   * \param[in] nSST_Options - Number of options selected.
+   */
+  SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS* SST_Options, unsigned short nSST_Options){
```
pcarruscag, line 9640: You also need to pass either the rank or a "is_master" boolean, otherwise the messages will be printed too many times.

pcarruscag, line 9640: I just saw you moved this to the config, please put it back in option structure, the config is already too big.

bigfooted:
The regression tests pass locally at my pc, but not on github. Not sure what is going on here...

bigfooted:
still don't know exactly why it sometimes fails all sst tests, I probably made some stupid error that I couldn't see. But at least the current situation is that it passes the regression tests...

Common/src/CConfig.cpp
bigfooted: Do you mean create a vector<SST_OPTIONS> and pass that as argument to ParseSSTOptions (instead of SST_OPTIONS itself)?

Common/include/CConfig.hpp
bigfooted: I've added the rank==MASTER_NODE. This routine is still called 3 times, so we see the messages 3 times (because setpostprocessing is called 3 times?)

pcarruscag: Probably.
Ok, but move this function back to option_structure please...
And remove the files that were committed by mistake, it makes this difficult to navigate.

TestCases/rans/naca0012/turb_NACA0012_sst_sust.cfg
```diff
@@ -18,7 +18,8 @@
 SOLVER= RANS
 %
 % Specify turbulent model (NONE, SA, SA_NEG, SST)
-KIND_TURB_MODEL= SST_SUST
+KIND_TURB_MODEL= SST
+SST_OPTIONS=(SUSTAINING)
```
pcarruscag, line 22: So what is the plan for the SUST and UQ options? From the code it looks like you want to support both, from the testcases it looks like you want to drop the old stuff?
I don't mind it either way, just wan't to know where we're going before looking at the code.

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_2d/FINDIFF/configFluid.cfg
```diff
@@ -0,0 +1,127 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
```
pcarruscag, line 2: Two files left to remove in this folder.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -725,20 +729,68 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+        if (sstParsedOptions.version == SST_OPTIONS::V1994) {
+          /*--- INTRODUCE THE SST-V1994 BUG WHERE DIVERGENCE TERM IS MISSING ---*/
+          /* --- note that mu_t = rho*a1*k/zeta and we can make this term independent of k ---*/
+          // the correct production with the additional divergence term
+          // pw -=  (alfa_blended*Density_i * (2.0 / 3.0 * diverg*diverg) +  2.0 / 3.0 * zeta * diverg);
+          /*--- the tke term only ---*/
+          pw = pw - 2.0 / 3.0 * zeta * diverg;
+        } else {
+         /* Note that we have to make the stress tensor tau_ij consistent everywhere when we neglect (or not) the divergence */
+         pw = pw - (2.0 / 3.0 * diverg * diverg + 2.0 / 3.0 * zeta * diverg);
```
pcarruscag, line 787: Wasn't the plan to leave this out for now?

TestCases/rans/naca0012/turb_NACA0012_sst_sust.cfg
bigfooted: I was first of the opinion that we should have a grace period for keywords that will become obsolete in the future. On the other hand, it is usually not a big deal to change a single keyword in a config file. 
If @WallyMaier agrees, we can ditch the old keywords in its entirety.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
bigfooted: Yes, we can comment it and make this option unavailable. However, we do need the term for v1994 at this moment since it is currently used.

TestCases/rans/naca0012/turb_NACA0012_sst_sust.cfg
pcarruscag: SUST and UQ are probably not used very often out in the wild, and we would go to 7.4 when this is released, so I think it is fine to drop the old ones.

Common/include/CConfig.hpp
```diff
@@ -9617,10 +9620,72 @@ class CConfig {
...
+    const bool sst_1994 = std::find(SST_Options, sst_options_end, SST_OPTIONS::V1994) != sst_options_end;
+    const bool sst_2003 = std::find(SST_Options, sst_options_end, SST_OPTIONS::V2003) != sst_options_end;
+ 
+    /* when V2003 is selected, we automatically select sst_m as well */
+    const bool sst_m     = std::find(SST_Options, sst_options_end, SST_OPTIONS::MODIFIED) != sst_options_end || sst_2003;
+
+    const bool sst_sust  = std::find(SST_Options, sst_options_end, SST_OPTIONS::SUST) != sst_options_end || Kind_Turb_Model==TURB_MODEL::SST_SUST;
+    const bool sst_v     = std::find(SST_Options, sst_options_end, SST_OPTIONS::VORTICITY) != sst_options_end;
+    const bool sst_kl    = std::find(SST_Options, sst_options_end, SST_OPTIONS::KL) != sst_options_end;
+    const bool sst_uq    = (std::find(SST_Options, sst_options_end, SST_OPTIONS::UNCERTAINTY) != sst_options_end || using_uq);
```
bigfooted, line 9653: @pcarruscag I'm confused about how to properly move this to option_structure. We move parseSSTOptions to option_structure, but keep GetSSTParsedOptions in CConfig? option_structure doesn't know about Kind_Turb_Model and rank, so we definitely have to pass rank, and deal with turb_model upstream then.

Also, ParseSSTOptions is a definition, so I made it inline to get it to compile.  Let me know if this is what you had in mind. I still have to do the 'modification upstream' to get everything to compile again.

pcarruscag: Since we are dropping the old options, you only need to pass the rank, or a `print_messages` boolean, the latter option could eventually be used to not print the messages 3 times.

TestCases/rans/naca0012/turb_NACA0012_sst_sust.cfg
WallyMaier: Im on board with completely moving the SUST and UQ over to SST_OPTIONS.  If a big deal, we could put an error saying "SST_SUST no longer an option"

pcarruscag: We will put that error.

Common/include/option_structure.hpp
```diff
@@ -951,20 +951,60 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
+  UNCERTAINTY, /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+};
+static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
+  MakePair("NONE", SST_OPTIONS::NONE)
+  MakePair("V1994m", SST_OPTIONS::V1994m)
+  MakePair("V2003m", SST_OPTIONS::V2003m)
+  MakePair("V1994", SST_OPTIONS::V1994)
+  MakePair("V2003", SST_OPTIONS::V2003)
+  //MakePair("MODIFIED", SST_OPTIONS::MODIFIED)
```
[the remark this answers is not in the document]
bigfooted: Just to be clear about this: the currently implemented 1994 model still has a 'divergence source term', which should be present only when the model is NOT modified. 
So, for now I can add this term to the V1994 model, and then we remove it when we switch to the literature V1994 model.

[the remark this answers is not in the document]
bigfooted: I also think that we should not stick with the name "V1994m" and V2003m". I think we should have only 1994 and 2003 as base versions, and "Modified" as an additional option. For now I am making "V1994m" as much as possible an alias of "V1994" + "modified"

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
bigfooted: OK, I am removing everything in turb_sources that we do not need right now, and we can get back to this commit when introducing the V1994 and V2003 models where we keep all divergence terms. We can then focus on making sure that V2003 is correct, and having the new options list in place.

Common/include/option_structure.hpp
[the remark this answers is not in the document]
pcarruscag: I know, but we are dropping terms with `k` which I guess makes it closer to the modified.

V?m would only be the ways for the user to specify, internally we would always convert to V? + m.
The problem with the user specifying V? + m is how to default m. Because we want to default to modified but then a user would also need to explicitly say `UNMODIFIED`.
So this can either be specified as (V1994 V1994m V2003 V2003m) or as (V1994 V2003 MODIFIED UNMODIFIED) or as (V1994 V2003 UNMODIFIED).
In the last one the absence of UNMODIFIED implies m, and we get the desired defaulting behavior.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
pcarruscag: Perfect.

bigfooted:
@WallyMaier, @pcarruscag I will double check the 2003m model and run a couple of cases. At this moment (for this PR) we can add a regression test for 2003m for the flat plate. I can add some more testcases later for the tutorials and V&V in separate PRs (I have some additional cases with measurements like the Stanford diffuser from Eaton, the 90 degree bend from Sudo and the axisymmetric jet from Hussein).

Common/include/option_structure.hpp
```diff
@@ -2331,6 +2368,70 @@ class COptionBase {
...
+    } else if (sst_kl) {
+      SSTParsedOptions.production = SST_OPTIONS::KL;
+    } else if (sst_uq) {
+      SSTParsedOptions.production = SST_OPTIONS::UQ;
+    }
+
+    // Parse boolean options
+    SSTParsedOptions.sust = sst_sust;
+    SSTParsedOptions.modified = sst_m;
+    SSTParsedOptions.uq = sst_uq;
```
WallyMaier, line 2431: is this boolean unneeded since we already set the production enum above?

Common/src/CConfig.cpp
```diff
@@ -3395,9 +3395,15 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
     SU2_MPI::Error("A turbulence model must be specified with KIND_TURB_MODEL if SOLVER= INC_RANS", CURRENT_FUNCTION);
   }
 
+  /*--- Postprocess SST_OPTIONS into structure. ---*/
+  if (Kind_Turb_Model==TURB_MODEL::SST){
+    sstParsedOptions = ParseSSTOptions(SST_Options, nSST_Options, rank);
+    using_uq = sstParsedOptions.uq;
```
WallyMaier, line 3401: I guess this is where the uq bool comes into play.  Would it be better to do:
`using_uq = sstParsedOption.prouduction == SST_OPTIONS::UQ` to avoid multiple paths of data?

pcarruscag: It would be better to rely only on the struct throughout the code, no extra booleans.

bigfooted: ok, so we remove using_uq

bigfooted: I have kept config::getUsingUq() but it now returns sstParsedOptions.uq, it is only used in viscous_fluxes.hpp.

Common/include/option_structure.hpp
bigfooted: I like the shorthand if(sstParsedOptions.uq) instead of if(sstParsedOptions.production == SST_OPTIONS::UQ).
But I'm fine with removing this boolean if you think it is unnecessary.

SU2_CFD/src/solvers/CAdjNSSolver.cpp
```diff
@@ -773,8 +773,8 @@ void CAdjNSSolver::Viscous_Sensitivity(CGeometry *geometry, CSolver **solver_con
             Enthalpy = solver_container[FLOW_SOL]->GetNodes()->GetEnthalpy(iPoint);
 
             /*--- Turbulent kinetic energy ---*/
-
-            if ((config->GetKind_Turb_Model() == TURB_MODEL::SST) || (config->GetKind_Turb_Model() == TURB_MODEL::SST_SUST))
+            /* Nijso says: This does not seem to be consistent with the primal treatment. */
+            if (config->GetKind_Turb_Model() == TURB_MODEL::SST) 
               val_turb_ke = solver_container[TURB_SOL]->GetNodes()->GetSolution(iPoint,0);
             else
               val_turb_ke = 0.0;
```
bigfooted, line 780: This has an impact on if we take into account the divergence term of the tke,  so I think we only should set a nonzero val_turb_ke in case of the original 1994, 2003, or 1994v and 2003v model

bigfooted: so this seems to be consistent with the current sst implementation, but we have to modify this condition when we fix the sst1994 model. 
I think for V2003m (current implementation) we should put val_turb_ke=0.0 since we drop all tke divergence terms?

bigfooted:
@pcarruscag, @WallyMaier What is left for this PR to do?

pcarruscag:
Regressions for other SST options, and an update of the documentation, including the warning we talked about.

Common/include/option_structure.hpp
```diff
@@ -952,19 +952,57 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+/*!
+ * \brief SST Options
+ */
+enum class SST_OPTIONS {
+  NONE,        /*!< \brief No SST Turb model. */
+  V1994,       /*!< \brief 1994 Menter k-w SST model. */
+  V2003,       /*!< \brief 2003 Menter k-w SST model. */
+  V1994m,       /*!< \brief 1994m Menter k-w SST model. */
+  V2003m,       /*!< \brief 2003m Menter k-w SST model. */
+  MODIFIED,    /*!< \brief Modified k-w SST model. */
```
pcarruscag, line 969:
```suggestion
```

Common/include/option_structure.hpp
```diff
@@ -952,19 +952,57 @@ inline TURB_FAMILY TurbModelFamily(TURB_MODEL model) {
...
+  KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
+  UQ,          /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+};
+static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
+  MakePair("NONE", SST_OPTIONS::NONE)
+  MakePair("V1994m", SST_OPTIONS::V1994m)
+  MakePair("V2003m", SST_OPTIONS::V2003m)
+  //MakePair("V1994", SST_OPTIONS::V1994)
+  //MakePair("V2003", SST_OPTIONS::V2003)
+  //MakePair("MODIFIED", SST_OPTIONS::MODIFIED)
```
pcarruscag, line 981:
```suggestion
```

Common/include/option_structure.hpp
```diff
@@ -38,6 +38,8 @@
 #include <algorithm>
 #include <cassert>
 
+using namespace std;
+
```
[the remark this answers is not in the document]
pcarruscag: Never do using std, never do using anything in a header file.

bigfooted:
> Regressions for other SST options, and an update of the documentation, including the warning we talked about.
Which other SST options do you mean? or do you mean V2003m compressible/incompressible and serial/parallel?
for sustaining and uq we already have regression tests

bigfooted:
@WallyMaier can you check Kato-Launder and vorticity and add a regression test?

WallyMaier:
@bigfooted Yup, I can make a few regressions here on my end.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -2098,10 +2098,11 @@ void CFlowOutput::WriteForcesBreakdown(const CConfig* config, const CSolver* flo
           break;
         case TURB_MODEL::SST:
           file << "Menter's SST\n";
-          break;
-        case TURB_MODEL::SST_SUST:
-          file << "Menter's SST with sustaining terms\n";
-          break;
+          // add the submodels here
+          //file << "Menter's SST with sustaining terms\n";
+          if (config->GetSSTParsedOptions().sust)
```
WallyMaier, line 2103: This is where I think a string in the options would be useful.
@pcarruscag you mentioned we want to keep the structure light.  Do you see a better way to output these messages without a large if/else structure?

pcarruscag: hmm ok we need it in more than one place... I'll fix that somehow as I review top to bottom.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -220,16 +229,26 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
...
 
     /*--- Compute the eddy viscosity ---*/
 
     su2double kine  = nodes->GetSolution(iPoint,0);
     su2double omega = nodes->GetSolution(iPoint,1);
-    su2double zeta  = min(1.0/omega, a1/(VorticityMag*F2));
+
+    su2double P_Base;
+
+    if (sstParsedOptions.version == SST_OPTIONS::V1994){
```
WallyMaier, line 244: This is a bit confusing to me.  
Where does this value come from?  In  turb_sources, P_base for both version is set using the StrainMag.

Further, shouldn't we have to include the altered P_Base for the UQ terms?  (I plan to introduce the proper vorticity and kato-launder terms here as well)

bigfooted: at this location, we use P_Base (defined in the if statement) to define the turbulent viscosity, which only changes depending on sst1994 or sst2003, and not on the other submodels

WallyMaier: Ah, so no need to include the sub-variants.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -717,28 +721,81 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+      su2double zeta;
+
+      if (sstParsedOptions.version == SST_OPTIONS::V1994){
+        zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
+        /*--- ---*/
+      } else {
+        zeta = max(ScalarVar_i[1], StrainMag_i * F2_i / a1);
+      }
+
+      su2double pw = pow(StrainMag, 2);
```
pcarruscag, line 780: I think this is wrong, the term used in the production is the same for both equations.

bigfooted: Yes, production should contain strain in the zeta term for both models, this is another difference between the literature V1994m and the su2 implementation. su2 mixed with the vorticity model here.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -717,28 +721,81 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+    //  } 
+
+      /*--- production limiter ---*/
+      pk = min(pk, ProdLimConstant * beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0]);
+      pk = max(0.0, pk);
+
+      su2double zeta;
+
+      if (sstParsedOptions.version == SST_OPTIONS::V1994){
+        zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
```
bigfooted, line 767:
```suggestion
      if (sstParsedOptions.version == SST_OPTIONS::V1994){
        /*--- FIXME: Note that the literature version of V1994m uses strain rate ---*/
        zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -746,64 +741,39 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
-      if (sstParsedOptions.version == SST_OPTIONS::V1994){
-        zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
-        /*--- ---*/
-      } else {
-        zeta = max(ScalarVar_i[1], StrainMag_i * F2_i / a1);
+      if (sstParsedOptions.version == SST_OPTIONS::V1994) {
+        /*--- INTRODUCE THE SST-V1994m BUG WHERE DIVERGENCE TERM IS MISSING ---*/
+        P -= 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
       }
+      su2double pk = max(0.0, min(pk, prod_limit));
```
pcarruscag, line 762:
```suggestion
      P = max(0, P);
      su2double pk = min(P, prod_limit);
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -717,37 +718,64 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
-      const su2double VorticityMag = GeometryToolbox::Norm(3, Vorticity_i);
-      const su2double zeta = max(ScalarVar_i[1], VorticityMag * F2_i / a1);
-      su2double pw = alfa_blended * Density_i * max(pow(StrainMag, 2) - 2.0 / 3.0 * zeta * diverg, 0.0);
+      su2double P = Eddy_Viscosity_i * pow(P_Base, 2);
+
+      if (sstParsedOptions.version == SST_OPTIONS::V1994) {
+        /*--- INTRODUCE THE SST-V1994m BUG WHERE DIVERGENCE TERM IS MISSING ---*/
+        P -= 2.0 / 3.0 * Density_i * ScalarVar_i[0] * diverg;
+      }
+      P = max(0, P);
```
pcarruscag, line 762:
```suggestion
      P = max(0.0, P);
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -209,30 +218,30 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
...
 
-    su2double kine  = nodes->GetSolution(iPoint,0);
-    su2double omega = nodes->GetSolution(iPoint,1);
-    su2double zeta  = min(1.0/omega, a1/(VorticityMag*F2));
-    su2double muT   = max(rho*kine*zeta,0.0);
+    const su2double kine = nodes->GetSolution(iPoint,0);
+    const su2double omega = nodes->GetSolution(iPoint,1);
+
+    const auto& eddy_visc_var = sstParsedOptions.version == SST_OPTIONS::V1994 ? VorticityMag : StrainMag;
+    const su2double muT = max(0.0, rho * a1 * kine / max(a1 * omega, eddy_visc_var * F2));
```
pcarruscag, line 242:
```suggestion
    // const su2double muT = max(0.0, rho * a1 * kine / max(a1 * omega, eddy_visc_var * F2));
    su2double zeta  = min(1.0 / omega, a1 / (eddy_visc_var * F2));
    su2double muT   = max(rho * kine * zeta, 0.0);
```

WallyMaier:
What else is left TODO on this PR?
Have reviewers?

pcarruscag:
Reviews (whoever reviews this please look also at https://turbmodels.larc.nasa.gov/sst.html), and:
- Update the docs on the website with new options and the warning about SST 1994
- Update the V&V results, flatplate with SST
- Some other cases? I made meshes for https://www.grc.nasa.gov/www/wind/valid/m5swbli/m5swbli.html maybe a good candidate for SST-V?

TobiKattmann (approved):
Hi y'all, thanks for this next chapter in what feels basically like a full refactoring the Turb-models in SU2. It is a lot of tedious work for just a handful of physics additions / corrections (which makes it even more important) :bouquet: 

The code looks tidy altogether and I looked at the physical differences just on the surface following TMR. The only real question I have is the reg-test one because I guess none of the values should change (but computers are involved which makes everything a lot more tricky ;)

TestCases/parallel_regression.py
```diff
@@ -1440,7 +1463,7 @@ def main():
     sp_pinArray_3d_cht_mf_hf_tp.cfg_dir   = "incomp_navierstokes/streamwise_periodic/chtPinArray_3d"
     sp_pinArray_3d_cht_mf_hf_tp.cfg_file  = "configMaster.cfg"
     sp_pinArray_3d_cht_mf_hf_tp.test_iter = 30
-    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [-13.380430, -7.476945, -7.025285, -0.009675, 99.879812, 4.1920e+02]
+    sp_pinArray_3d_cht_mf_hf_tp.test_vals = [-13.374306, -7.476945, -7.025285, -0.009675, 99.879812, 419.200000]
```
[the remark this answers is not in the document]
pcarruscag: I traced it down to https://github.com/su2code/SU2/pull/1560/commits/f61026be64f28a45c8b996f9b2324e3d0650cf6c so just a line of code which is now written exactly like tmr

bigfooted:
>     * Update the V&V results, flatplate with SST
I've started with the flatplate, rerunning the large meshes right now (original SST and 2003m model currently).

config_template.cfg
```diff
@@ -18,9 +18,12 @@
 %              HEAT_EQUATION_FVM, ELASTICITY)
 SOLVER= EULER
 %
-% Specify turbulence model (NONE, SA, SA_NEG, SST, SA_E, SA_COMP, SA_E_COMP, SST_SUST)
+% Specify turbulence model (NONE, SA, SA_NEG, SST, SA_E, SA_COMP, SA_E_COMP)
 KIND_TURB_MODEL= NONE
 %
+% Specify versions/corrections to SST model (NONE, V2003m, V1994, VORTICITY, SUSTAINING, UQ)
```
pcarruscag, line 24:
```suggestion
% Specify versions/corrections to SST model (V2003m, V1994m, VORTICITY, KATO_LAUNDER, UQ, SUSTAINING)
```

TestCases/rans/naca0012/turb_NACA0012_sst_1994-KLm.cfg
```diff
@@ -0,0 +1,118 @@
...
+
+% -------------------------- MULTIGRID PARAMETERS -----------------------------%
+%
+MGLEVEL= 0
+MG_PRE_SMOOTH= ( 1, 1, 1, 1, 1, 1 )
+MG_POST_SMOOTH= ( 0, 0, 0, 0, 0, 0 )
+MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0, 0, 0 )
+MG_DAMP_RESTRICTION= 0.75
+MG_DAMP_PROLONGATION= 0.75
+
```
pcarruscag, line 73:
```suggestion
```

TestCases/rans/naca0012/turb_NACA0012_sst_2003-Vm.cfg
```diff
@@ -0,0 +1,118 @@
...
+
+% -------------------------- MULTIGRID PARAMETERS -----------------------------%
+%
+MGLEVEL= 0
+MG_PRE_SMOOTH= ( 1, 1, 1, 1, 1, 1 )
+MG_POST_SMOOTH= ( 0, 0, 0, 0, 0, 0 )
+MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0, 0, 0 )
+MG_DAMP_RESTRICTION= 0.75
+MG_DAMP_PROLONGATION= 0.75
+
```
pcarruscag, line 73:
```suggestion
```

TestCases/rans/naca0012/turb_NACA0012_sst_2003m.cfg
```diff
@@ -0,0 +1,96 @@
...
+VENKAT_LIMITER_COEFF= 0.03
+LIMITER_ITER= 99999
+
+% ------------------------ LINEAR SOLVER DEFINITION ---------------------------%
+%
+LINEAR_SOLVER= FGMRES
+LINEAR_SOLVER_PREC= ILU
+LINEAR_SOLVER_ERROR= 1E-10
+LINEAR_SOLVER_ITER= 20
+MGLEVEL= 0
```
pcarruscag, line 61:
```suggestion
```

bigfooted:
flatplate_cf_0p97_gridconv_sst
typical result of comparison of original SU2 (1994m) vs SU2-V2003m. The difference (< 2%) is due entirely to the change from vorticity to strain in the computation of muT (CturbSSTSolver.cpp)

bigfooted:
bump_cl_gridconv_sst
NOTE: for this result, the finest grid was not computed yet
For bump in channel: same trend. For this case, V2003m results are available of FUN3D and CFL3D, but they are almost the same for this test case

pcarruscag:
It would be nice to match their results, but I think ours look better, i.e. seem to converge with grid size.

WallyMaier:
The presented 1994m were results from the "bad" 1994 version in SU2?
Also, I am trying to find some cluster space to run the SWBLI case to check vorticity model.

pcarruscag:
That case has some problems, probably because of #1631 (supersonic inflow is also affected)

WallyMaier:
I can just run the flat plate with the V and KL options, if that works.

pcarruscag:
Up to you but given that it's a one line change... I would give more priority to resurrect the V&V pages for the plate and bump cases, and adding something to the vandv.py regressions.

TestCases/vandv.py
```diff
@@ -48,6 +48,22 @@ def main():
...
+    flatplate_sst1994m           = TestCase('flatplate_sst1994m')
+    flatplate_sst1994m.cfg_dir   = "vandv/rans/flatplate"
+    flatplate_sst1994m.cfg_file  = "turb_flatplate_sst.cfg"
+    flatplate_sst1994m.test_iter = 20
+    flatplate_sst1994m.test_vals = [-3.929386, -3.061217, 3.958147, 0.001080, 0.003899]
+    test_list.append(flatplate_sst1994m)
+
+    # bump in channel - sst-v1994m
+    bump_sst1994m           = TestCase('bump_sst1994m')
+    bump_sst1994m.cfg_dir   = "vandv/rans/bump_in_channel"
```
pcarruscag, line 61: Nice :+1:
Please cleanup the configs from comments and for V&V regressions all cases should be restarted from a converged solution, we want to track the values at convergence.

I'm working on having the SA options in the same format as SST btw.

bigfooted: ok, great, I've modified the python scripts for the postprocessing, I'll update them as well (upgraded to use pandas framework)

aeroamit:
> There is a shock near the start of the plate, and it's vertical position changes 1:1 with mesh size image

I had also tried this case sometime back with SU2. Faced the same issues as reported by Pedro. Also, SST was not tallying with the results presented in the reference (https://www.grc.nasa.gov/WWW/wind/valid/m5swbli/study02/m5swbli2.html).  There seems to be a shift of 5mm. It will be useful to sort out this case...
See some details in the attached pdf file... 
[SWBLI_case_issue.pdf](https://github.com/su2code/SU2/files/8710653/SWBLI_case_issue.pdf)

TestCases/vandv.py
```diff
@@ -48,6 +48,22 @@ def main():
...
+    bump_sst1994m.cfg_file  = "turb_bump_sst.cfg"
+    bump_sst1994m.test_iter = 20
+    bump_sst1994m.test_vals = [-4.295705, -1.571766, 2.662294, 0.024496, 0.004525]
```
pcarruscag, line 64: These residuals are too high for a restarted case, what's wrong here?

bigfooted: It's wrong, that's for sure. I must have been sleeping that I didn't see this. I updated the residuals and the converged solution files.

bigfooted: It passes on my workstation, but not on github. The residuals stored in vandv.py are OK, but the restart on github still shows initial residuals that are too low.

bigfooted: OK, I had to manually rerun the V&V regression by pressing the button... But it looks like the residuals on github are off by the last digit due to some difference between my local machine and the remote machine.

bigfooted: OK, looks good now...

pcarruscag:
Do you want to do the honors @bigfooted ?

SU2_CFD/src/solvers/CAdjNSSolver.cpp
```diff
@@ -773,8 +773,8 @@ void CAdjNSSolver::Viscous_Sensitivity(CGeometry *geometry, CSolver **solver_con
             Enthalpy = solver_container[FLOW_SOL]->GetNodes()->GetEnthalpy(iPoint);
 
             /*--- Turbulent kinetic energy ---*/
-
-            if ((config->GetKind_Turb_Model() == TURB_MODEL::SST) || (config->GetKind_Turb_Model() == TURB_MODEL::SST_SUST))
+            /* Nijso says: This does not seem to be consistent with the primal treatment. */
```
WallyMaier, line 776:
```suggestion
            /* Nijso says: This does not seem to be consistent with the primal treatment. */
            //TODO or FIXME (just so we know where to look
```

WallyMaier (commented):
This looks good to me!  Thanks for taking the lead @bigfooted and @pcarruscag!

PR #1566, opened 2022-03-19
Fix CGNS mesh reader for multizone problems (one CGNS mesh per zone)

### LINKED ISSUE #1564: CGNS import fails in multizone simulations
**Describe the bug**
I'm attempting to run coupled CHT simulations with a single fluid and a single solid zone. Both zones are read as seperate CGNS files.
In Dry-run mode everything works fine. However, when starting an actual run no matter in which order the zones are specified in the configuration list, the 2nd zone is always considered invalid:

> ------------------- Geometry Preprocessing ( Zone 1 ) -------------------
> Reading the CGNS file: MeshSolid.cgns.
> WARNING: The CGNS file version (3.31)  is old and may cause high memory usage issues, consider updating the file with the cgnsupdate tool.
> CGNS file contains 1 database(s).
> Database 1, BASE#1:  cell dimension of 2, physical dimension of 3.
> 1 total zone(s).
> Zone number 2 invalid

It seems this error is thrown during the CGNS import.
https://github.com/su2code/SU2/blob/72b2fa977b64b9683a388920f05298a40d39e5c5/externals/cgns/cgns_internals.c#L10186-L10198

Both zones run fine individually (when adding the missing, coupled BC of course).
When the second zone is converted to SU2 format and read in the native format, the case runs normally.

**To Reproduce**
*Please provide a config file and mesh for a minimal example that recreates the problem.*
[CoupledSim.zip](https://github.com/su2code/SU2/files/8303540/CoupledSim.zip)

 - OS: SUSE Linux Enterprise Server 15 SP1
 - C++ compiler and version: intel 19.0.5.281
 - MPI implementation and version: intel MPI 2019
 - SU2 Version: v7.3.0 (develop)

pcarruscag:
Fixes #1564

### DIFF
--- a/Common/include/geometry/meshreader/CCGNSMeshReaderFVM.hpp
+++ b/Common/include/geometry/meshreader/CCGNSMeshReaderFVM.hpp
@@ -45,10 +45,9 @@ private:
 
 #ifdef HAVE_CGNS
   int cgnsFileID; /*!< \brief CGNS file identifier. */
-  int cgnsBase;   /*!< \brief CGNS database index. */
-  int cgnsZone;   /*!< \brief CGNS zone index. */
+  const int cgnsBase = 1; /*!< \brief CGNS database index (the CGNS reader currently assumes a single database). */
+  const int cgnsZone = 1; /*!< \brief CGNS zone index (and 1 zone in that database). */
 
-  int nZones;     /*!< \brief Total number of zones in the CGNS file. */
   int nSections;  /*!< \brief Total number of sections in the CGNS file. */
 
   vector<bool> isInterior;             /*!< \brief Vector of booleans to store whether each section in the CGNS file is an interior or boundary section. */
--- a/Common/src/geometry/meshreader/CCGNSMeshReaderFVM.cpp
+++ b/Common/src/geometry/meshreader/CCGNSMeshReaderFVM.cpp
@@ -35,12 +35,6 @@ CCGNSMeshReaderFVM::CCGNSMeshReaderFVM(CConfig        *val_config,
 : CMeshReaderFVM(val_config, val_iZone, val_nZone) {
 
 #ifdef HAVE_CGNS
-  /*--- We use val_iZone with +1 for the 1-based indexing in CGNS. ---*/
-  cgnsZone = val_iZone + 1;
-  nZones   = val_nZone;
-
-  /*--- The CGNS reader currently assumes a single database. ---*/
-  cgnsBase = 1;
   OpenCGNSFile(config->GetMesh_FileName());
 
   /*--- Read the basic information about the database and zone(s). ---*/


MicK7:
I don't think the fix is as simple as it seems.
Indeed looking back at how we developed the CGNS reader, It was originally meant to read multiple zone in a single file. But during development, someone decided to restrict the reader to only one zone per file (and I don't know if it was validated). So now we are seating in the middle.

If we replace line 169 of CGNSFVMMeshReader :
<pre>
if ( nzones > 1 ) {
    SU2_MPI::Error(string("CGNS reader currently expects only 1 zone per CGNS file.") +
                   string("Multizone problems can be run with separate CGNS files for each zone."), CURRENT_FUNCTION);
}
</pre>
by
<pre>
 if ( cgnsZone > nzones) {
  cgnsZone = 1;
}
</pre>

we can easily support multiple zone in one file.

To support one CGNS zone per file, I guess that user should provide either the index in the cgns file of the zone we want to read or even better its name and not rely on SU2 numbering of zones.

I think that supporting multiple mesh zones in the same file at the same time as one zone per mesh file should be possible as long as enough information is provided by the user.

In this case, I am wondering how the option MULTIZONE_MESH and MULTIZONE option are interacting in the related issue.

When MULTIZONE_MESH is set to NO do we expect one mesh file per zone ?
And in this case we can force CGNS Reader to read only the first Zone.

In a more generic way something like this should be possible:
MULTIZONE=YES
CONFIG_LIST= (zone_1.cfg, zone_2.cfg, zone_3.cfg)
CGNSZONENAMES = ("FluidRotor", "Solid", "FluidStator") # To let CGNS pick the right zone in the file and if it not found the first zone can be used (current SU2 behavior).

CGNSZONENAMES could also be set in each config file.

pcarruscag:
Thank you for having a look @MicK7.
Allowing multizone cgns based on the MULTIZONE_MESH option sounds reasonable but I did not want to remove the error without testing. Do you (or @ChristianBauerEng) have a multizone cgns we can use to test this?
I don't think we should add more config options.

MicK7:
I can provide a multizone mesh.

ChristianBauerEng (approved):
Unfortunately I don't have a multizone-cgns mesh at the moment.
But I've tested this fix with my single-zone cgns meshes and I can confirm, that this fix works just as the documentation specifies.
MicK7's suggestion to make the reader also work with multi-zone cgns files would also be nice, but I understand that this requires a lot more work.

pcarruscag:
Thanks for testing 👍 it's only a bit more work, just needs a quick test.

pcarruscag:
Thanks for testing 👍 it's only a bit more work, just needs a quick test.

MicK7:
@pcarruscag I will create a separate PR for multizone cgns file.

pcarruscag:
Sounds good, thanks.

MicK7:
@pcarruscag I have a small file with two zones here: https://github.com/MicK7/TestcaseStage37
It is the classical NASA stage 37 (https://www.turbostream-cfd.com/case-studies/nasa-stage-37-stall-inception) but the file is 149 Mo. Do you think the size would not be too big to have it as a testcase ?
Sorry for replying on a closed PR ...

pcarruscag:
Great, thanks @MicK7. It is a bit big to have as a regression test just for this, but we can certainly use it to test the 1 or 2 lines we need to change to enable multizone cgns, are you planing to make those changes?

PR #1579, opened 2022-03-26
Fix some "resource not released in destructor" warnings

pcarruscag:
The COption classes should delete what they allocate


TobiKattmann (approved):
Thanks for thinking into that structure and fixing and cleaning this :bouquet: 

option_structure.inl` is not exactly my favorite file and tedious to understand what's really going on imo, so extra :+1:  for doing that :bow:

Issue #1467, opened 2021-12-08
Instability in BL with Periodic Boundary Conditions


jf-thomas:
**Describe the bug**
In a RANS computation of a 2.5D section of a wing, the periodic plane exhibits oscillations of pressure in the boundary layer. The issue is only present on the periodic face, not in the volume. Using a CFL of 1 helps but the instability does not disappear. A CFL of 10 leads to a crash of the computation with multiple spots like on the screenshot. Would it be linked to issue #1090  ?

![pres_per_face](https://user-images.githubusercontent.com/91184648/145236024-fc713cbd-66d8-40b3-9036-7c1925f10229.png)
![pres_cut](https://user-images.githubusercontent.com/91184648/145236048-a91ca27a-c1c9-4d6c-9639-8b34ea5de307.png)

Here is the config file:
[hl_3d_landing.txt](https://github.com/su2code/SU2/files/7677092/hl_3d_landing.txt)

**Hardware and software:**
 - OS:  Linux Red Hat ELS 6.5
 - Nodes: Intel Xeon Ivy bridge
 - C++ compiler and version: intel/2018.0.1
 - MPI implementation and version: intelmpi/2018.0.1
 - SU2 Version: v7.2.1

Thank you for looking at this issue and for any help.

jf-thomas:
I made a light test case to reproduce the issue. It's a cylinder in a flowfield. It's only 112k cells.

![mesh](https://user-images.githubusercontent.com/91184648/145433376-ec7d4309-acc6-43d2-a172-39502e823147.png)
![swpeptCylinder](https://user-images.githubusercontent.com/91184648/145433398-97123613-824f-4f81-a143-aaae790ce252.png)
![residual_energy](https://user-images.githubusercontent.com/91184648/145433426-e6fc38f8-8482-47be-bd52-dca747b3e3e4.png)

The residuals never converge:
![residual_convergence](https://user-images.githubusercontent.com/91184648/145432900-0c2d7432-807b-4fe4-b24d-abf01cdc7bc4.png)
Here is the archive containing the case:
[cyl.zip](https://github.com/su2code/SU2/files/7686715/cyl.zip)

pcarruscag:
Hi,
Can you try the develop branch or SA instead of SA_NEG. We fixed an issue in SA_NEG that could cause this type of oscillation.
However, I suspect it is due to issues found in #763 and that @TobiKattmann has also run into when implementing periodic BC's for heat-transfer problems.

[...]

PR #1585, opened 2022-03-29
[WIP] Improve implementation of periodicity for implicit integration

pcarruscag:
## Proposed Changes
This is a simple re-shuffling to have matrix-vector products (and therefore linear systems) that are aware of periodic connections.
It uses something like:
- A "`PERIODIC_RESIDUAL`" communication (sum contributions) applied to the result of matrix-vector products.
- A "`PERIODIC_IMPLICIT`" communication (enforce periodicity) applied to the result of preconditioners.

It seems to avoid the performance degradation that @TobiKattmann saw when implementing periodicity for heat transfer, and also the issues in issue #1467.

However, making the matrix-vector product "periodicity-aware" through comms, without modifying the matrix, means that any preconditioner derived from the matrix is not consistent with the product, thus it is a worse preconditioner and more linear iterations are needed for the same reduction in residuals.

## Related Work
Should help #1467
Should avoid #763

## PR Checklist


Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -387,96 +316,169 @@ void CSysMatrixComms::Complete(CSysVector<T>& x, CGeometry *geometry,
...
+      const auto buf_offset = (msg_offset + iSend)*COUNT_PER_POINT;
+
+      for (auto iVar = 0ul; iVar < x.GetNVar(); iVar++)
+        bufDSend[buf_offset+iVar] = x(iPoint,iVar);
+
+      if (vec_idx < 0) continue;
+
+      const auto iPeriodic = geometry->Local_Marker_PeriodicSend[msg_offset + iSend];
+      const auto tag = config->GetMarker_All_TagBound(iPeriodic);
+      const auto* angles = config->GetPeriodicRotAngles(tag);
```
TobiKattmann, line 410:
```suggestion
      const su2double* angles = config->GetPeriodicRotAngles(tag);
```
I think auto* for su2double* is fine :thinking:  Just resolve in case auto* works :+1:

TobiKattmann:
I'll take a more in-depth look my cases tomorrow (Saturday) to adapt the regression values, and options if necessary. In some cases I saw a massive improvement in convergence speed so I'll decrease the prescribed iterations to reflect that :+1: I hope it is ok to push to this PR directly (won't be that many lines ;)

TobiKattmann:
Whilst the primal computations work perfectly (greatly increased convergence speed in some cases, whilst not/or very very minorly altering the final solution), I see some issues with the discrete adjoint: The cases I have diverge after ~20-40 iterations. I tried to increase Linear solver iterations and lower the CFL-number quite a bit, without success. 

Whether I use the primal solution-file created with the old version or with this branch, did not matter.

da_sp_pinArray_cht_2d_dp_hf from parallel_regression_AD is maybe a good case to start but I'll try to make a fluid-only case from that which might be easier to debug (in case that would be necessary) and post it here

TobiKattmann:
Here a fluid-only case that somewhat quickly converges. Contains the mesh, 1 cfg for primal and adjoint and a little bash-script where the path to the SU2 binaries has to be adapted to fit your system. The idea was to first run primal and adjoint with vanilla develop and then the same with the new code.

I left the log-files of my run in there as well for a quick look 

If there is something else I can do let me know. 

[new-per-bc-test.tar.gz](https://github.com/su2code/SU2/files/8408958/new-per-bc-test.tar.gz)

pcarruscag:
Yeah... I had not tested discrete adjoint but my spider senses were tingling.

pcarruscag:
WIP

PR #1552, opened 2022-03-03
Allow different OUTPUT_WRT_FREQ for each output file

bigfooted:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*

 - add frequency for each filename in OUTPUT_FILES: 
` OUTPUT_WRT_FREQ= 10, 100, 50`

 - for dual timestepping, write only the restart files twice (for timestep n and n-1) and not the visualisation files.
 -  history output accepts fields
 - marker_inlet does not need normalized vectors (removed the error message)

```diff
- DONE
The suggestion was to just extend the OUTPUT_WRT_FREQ instead of introducing a new keyword VOLUME_OUTPUT_FREQUENCIES. 

```

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*
#1495 
#1539 
#1493 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

Issue #1493, opened 2022-01-14
HISTORY_OUTPUT only accepts groups, not fields


bigfooted:
**Describe the bug**

Maybe "it's not a bug, it's a feature", but:
I can for example select a field for output
SCREEN_OUTPUT= INNER_ITER 
or use the group,
SCREEN_OUTPUT= ITER
which is equivalent to choosing the fields INNER_ITER, OUTER_ITER, TIME_ITER

But for HISTORY_OUTPUT, I can only choose the group. This leads to a lot of data in the history file if I want to know what happens at a number of markers. So I would like to do this:

HISTORY_OUTPUT= (INNER_ITER, SURFACE_MASSFLOW)

to get 2 columns in my history file. Instead I have to do this:

HISTORY_OUTPUT= (ITER, SURFACE_MASSFLOW)

which leads to a lot of columns. Is there a reason that only groups are allowed or can this limitation be lifted safely?

pcarruscag:
In some parts of the code it looks like the intention was to allow both modes, see
COutput::SetHistoryFile_Header `if (requestedField == field.outputGroup || (requestedField == fieldIdentifier)){`

In principle you "just" need to apply that logic to every use of `requestedHistoryFields`.

If would be more difficult to make the screen output know about groups, because it uses the names you give to lookup values directly, whereas the history loops over all available values and checks their group against all groups you give it.

bigfooted:
Thanks! Yeah I noticed that in COutput::CheckHistoryOutput, there is some preliminary check to filter out invalid keywords, where for history the group keywords are checked.  I will then look into making history accept fields. Not sure how useful an extension of screen output will be to accept groups.

PR #1571, opened 2022-03-22
Fix for discrete adjoint: axisymmetry + SST turbulence model

lkusch:
## Proposed Changes
The velocity component used in the axisymmetric residual is now considered as an input for preaccumulation in the residual computation. When not considered, the use of axisymmetry and the SST model will result in wrong sensitivities when using the discrete adjoint method.

I also wanted to note that there does not seem to be any treatment of axisymmetric cases for the SA model. Is this correct or does it make sense to include a warning? We could, for example, do that in the context of this pull request. 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x ] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [ x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1587, opened 2022-03-31
Allow field groups in SCREEN_OUTPUT (e.g. RMS_RES like for HISTORY_OUTPUT)

pcarruscag:
Small change for consistency with work done in #1552 to address #1493 
Useful at least to get all the residuals with little work...


WallyMaier (approved):

TobiKattmann (approved):
Awesome! Love it :bouquet: I tend to write every RMS, so this will ease things a bit. I guess this also works for adjoint cases which is one thing less to care about in a setup :+1: 

https://github.com/su2code/SU2/pull/1571/files#diff-7d139cfb7a735e5c085a26f115306f13b8209745edf69dc56281269478784972 from #1571 would be a nice application as it would cover a primal and an adjoint case (momentum residuals are not present in the current setup though)

(I guess the order of RMS is changed for your adapted testcase: Now it is rho, rhoE, rhoU, rhoV instead of rho, rhoU, rhoV, rhoE ... but no values changed)

bigfooted (approved):
great work, thanks for the follow up!

PR #1496, opened 2022-01-16
[WIP] Develop for transition model.

sun5k:
Hi, all.

This PR is for implementing various transition models, which are composed of 1~2 additional transport equations.

List of wishes to implement transition model  : 

- [Langtry and Menter model](https://doi.org/10.2514/1.42362) 
- [k-w-gamma](https://doi.org/10.1016/j.paerosci.2012.08.004)
- [SA-AFT](https://doi.org/10.2514/6.2019-0039)

Now working on : 
- Langtry and Menter model

## Related Work

LM feature -> development

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1592, opened 2022-04-05
[WIP] Develop for Langtry and Menter transition model.

sun5k:
## Proposed Changes
 
Hi, all.

This PR is resuming the #1496. I'm not familiar with the git, so the branch is unfortunately disconnected.

## Now working on 

[Langtry and Menter model](https://doi.org/10.2514/1.42362)

## Current State 

Code is well running. But, the result is not satisfactory. 😂
The LM(Langtry and Menter) model implemented to Su2 is under debugging compared to the results of Fluent's LM model.

Transition onset is located upward then Fluent result. But, the good news is that when the correlations are set fixed value, the transition onset location is similar to the Fluent result.

## PR Checklist


pcarruscag:
Hi @sun5k the SST version implemented in fluent is not the same as we have, maybe that's where the difference comes from.

sun5k:
Hello. 

Here is sharing the status of debugging the LM model. 


T3A flat plate case : 
image

NASA 2DZP Mach number 0.2 case :

image


The after transition process, I think the difference in the skin friction coefficient comes from using a different k-w SST model.

pcarruscag:
Hi @sun5k, that looks promising :+1: 
It would be nice if you could attend the developer meetings and tell us about your plans regarding the integration of this feature.
Every Wednesday at 3pm UK, 4pm CET, https://meet.jit.si/SU2_DevMeeting
There are folks working on implementing SST-2003m, this could be a better comparison with fluent.

pcarruscag:
Now that you have running examples it would be extremely important to add them as regressions tests, so that you can ensure that changes coming from develop do not break the code.
Please turn on the GitHub notifications because it seems you have missed quite a few of the comments on your pull requests...

sun5k:
Hi @pcarruscag 

Thank you for the invitation to the meeting. Unfortunately, this week I have a domestic conference. I think I'll have to promise next(not a next week).

rois1995:
Hi @sun5k ,

I am also working on the implementation of this transition model within SU2. 

In order to verify the code, I am running simulations on the E387 profile (taken from DOI:10.1177/0954406217743537) and on the Flat Plate test cases (taken from DOI: 10.2514/1.42362). 

While for the E387 profile the results are similar (although not really convincing yet), for the flat plate (NASA 2DZP) the transition occurs right at the leading edge. Could you please share the config files that you have used?

I also noted a couple of inaccuracies within your implementation (by inspection of the "Files changed" section), maybe we can have a talk to discuss about it.

Thank you in advance.
Best regards,
Andrea

Common/src/CConfig.cpp
```diff
@@ -5877,7 +5877,11 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
           case TURB_MODEL::SST_SUST:  cout << "Menter's SST with sustaining terms" << endl; break;
         }
         if (QCR) cout << "Using Quadratic Constitutive Relation, 2000 version (QCR2000)" << endl;
-        if (Kind_Trans_Model == TURB_TRANS_MODEL::BC) cout << "Using the revised BC transition model (2020)" << endl;
+        switch (Kind_Trans_Model) {
+        case TURB_TRANS_MODEL::NONE:  break;
+        case TURB_TRANS_MODEL::LM:    cout << "Trnasition model: Langtry and Menter's 4 euqation model (2009)" << endl; break;
+        case TURB_TRANS_MODEL::BC:    cout << "Trnasition model: Using the revised BC transition model (2020)" << endl; break;
```
WallyMaier, line 5883:
```suggestion
        case TURB_TRANS_MODEL::BC:    cout << "Transition model: Using the revised BC transition model (2020)" << endl; break;
```

SU2_CFD/src/solvers/CTransLMSolver.cpp
```diff
@@ -37,480 +40,461 @@
...
-    } else {
-      REth_Inf = 331.5*pow(tu_Inf-0.5658,-0.671);
+
+  /*--- Momentum thickness Reynolds number, initialized from freestream turbulent intensity*/
+  if (Intensity <= 1.3) {
+    if(Intensity >=0.027) {
+      ReThetaT_Inf = (1173.51-589.428*Intensity+0.2196/(Intensity*Intensity)) * f_lambda;
+    }
+    else {
+      ReThetaT_Inf = (1173.51-589.428*Intensity+0.2196/(0.27*0.27))*f_lambda;
```
RichRoos, line 128: For consistency: does this have to be 0.027*0.027 as this forces the minimum value of the Intensity of 0.027?

SU2_CFD/src/solvers/CSolverFactory.cpp
```diff
@@ -92,7 +92,9 @@ CSolver** CSolverFactory::CreateSolverContainer(MAIN_SOLVER kindMainSolver, CCon
     case MAIN_SOLVER::RANS:
       solver[FLOW_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::NAVIER_STOKES, solver, geometry, config, iMGLevel);
       solver[TURB_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::TURB, solver, geometry, config, iMGLevel);
+      solver[TRANS_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::TRANSITION, solver, geometry, config, iMGLevel);
       solver[SPECIES_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::SPECIES, solver, geometry, config, iMGLevel);
+
       break;
     case MAIN_SOLVER::INC_RANS:
       solver[FLOW_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::INC_NAVIER_STOKES, solver, geometry, config, iMGLevel);
```
RichRoos, line 100: Adding "solver[TRANS_SOL] = CreateSubSolver(SUB_SOLVER_TYPE::TRANSITION, solver, geometry, config, iMGLevel);" here for incompressible simulations.

RichRoos (commented):
First review comment :). These are a few points I noticed while looking through the code. Let me know whether this is the correct method or not and whether these are helpfull.

SU2_CFD/src/solvers/CTransLMSolver.cpp
rois1995: Also, there shouldn't be any f_lambda multiplication for the freestream conditions.

SU2_CFD/src/solvers/CSolverFactory.cpp
sun5k: I know this but have not applicated the LM yet for an incompressible solver. I want to finish the LM model for the compressible solver first and start later for the incompressible solver.

rois1995:
Hi @sun5k,

I think that the restart of the solution is missing. Since for CTransLMSolver you have inherited the functions from CTurbSolver class, the restart is only applied to the turbulent solver. I solved this problem by creating another class called CTransSolver which is a copy of the CTurbSolver class but with a modified LoadRestart function. I guess that the function in the CTurbSolver class can be modified to take into account the transition solver. It depends if we want to divide the transition from the turbulence solver or not.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -82,6 +82,11 @@ class CNumerics {
   turb_ke_i,  /*!< \brief Turbulent kinetic energy at point i. */
   turb_ke_j;  /*!< \brief Turbulent kinetic energy at point j. */
   su2double
+  intermittency_sep_i,  /*!< \brief separation intermittency at point i. */
+  intermittency_sep_j,  /*!< \brief separation intermittency at point j. */
+  intermittency_eff_i,  /*!< \brief separation intermittency at point i. */
+  intermittency_eff_j;  /*!< \brief separation intermittency at point j. */
```
WallyMaier, line 88:
```suggestion
  intermittency_eff_i,  /*!< \brief effective intermittency at point i. */
  intermittency_eff_j;  /*!< \brief effective intermittency at point j. */ 
```

SU2_CFD/include/numerics/turbulent/transition/trans_diffusion.hpp
```diff
@@ -0,0 +1,106 @@
+/*!
+ * \file turb_diffusion.hpp
```
WallyMaier, line 2: This can be updated

SU2_CFD/include/numerics/turbulent/transition/trans_sources.hpp
```diff
@@ -0,0 +1,303 @@
...
+   * \param[in] val_kine_Inf - Freestream k, for SST with sustaining terms.
+   * \param[in] val_omega_Inf - Freestream w, for SST with sustaining terms.
+   * \param[in] config - Definition of the particular problem.
+   */
+  CSourcePieceWise_TransLM(unsigned short val_nDim, unsigned short val_nVar, const CConfig* config) 
+      : CNumerics(val_nDim, 2, config),
+        idx(val_nDim, config->GetnSpecies()) {
+
+    /*--- LM Closure constants ---*/
+    c_e1 = 1.0;
```
WallyMaier, line 72: const these

SU2_CFD/src/variables/CTurbSSTVariable.cpp
```diff
@@ -81,6 +81,13 @@ void CTurbSSTVariable::SetBlendingFunc(unsigned long iPoint, su2double val_visco
   arg2 = max(2.0*arg2A, arg2B);
   F2(iPoint) = tanh(pow(arg2, 2.0));
 
+  /*--- LM model for F1 ---*/
+  if(trans_model == TURB_TRANS_MODEL::LM) {
```
WallyMaier, line 85:
```suggestion
  if (trans_model == TURB_TRANS_MODEL::LM) {
```

sun5k:
Hi guys~ 

Here is sharing the status of the LM model.

V&V cases: T3A, T3A-, NLF

Flow conditions are the reference from:  https://doi.org/10.2514/6.2022-3679.

The grids of T3A and NLF cases are provided by [TMW](https://transitionmodeling.larc.nasa.gov/workshop_i/)(Transition Model Workshop). And I made the grid of T3A-  myself.


To validate the LM model, the simulation results of SU2 are compared with the results of  Fluent19.0 with a similar numerical setting.

Here is the numerical scheme : 

<html xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">



</head>

<body>




  | Fluent | SU2
-- | -- | --
Flux | Roe-FDS | L2ROE
Gradient | Least   Squares Cell Based | WEIGHTED_LEAST_SQUARES
Spatial   Discretization Flow | Third-order   MUSCL | MUSCL_FLOW
Spatial   Discretization Turbulence | First-order   Upwind | MUSCL_NO




</body>

</html>
The CFL number is set constant value.


Some high-resolution grid-level results are missed because I don't have enough computational resources.
Here is Result

T3A : 

image
image
image
image


T3A-: 
The gird test result is not converged, but It's enough to compare the model.

image
image
image
image

NLF : 
AoA 0 case is shown. In this case, the flow separation occurred on the upper surface. So, not fully converged, and I show only Fine level grid results according to the number of iterations. 
image
Zoom to the separation region : 
image


Now I'm testing on effect of the numerical schemes and other options like CONV_NUM_METHOD_FLOW, MUSCL_TURB, CFL, and CFL_ADAPT.

pcarruscag:
Results look good :+1: 
You should present this work at the SU2 conference https://su2code.github.io/ please submit a short abstract if you want to present.

rois1995:
Hi @sun5k, could you please add also the configuration files to reproduce these results? I would try to run these simulations too with my implementation of the transition model.

Thank you in advance!

sun5k:
> Results look good 👍 You should present this work at the SU2 conference https://su2code.github.io/ please submit a short abstract if you want to present.
@pcarruscag  Thank you for the suggestion. 
I want to really participate in the SU2 conference, but I can't afford to make a presentation at the SU2 conference because I'm busy with my wedding and preparing for domestic conferences. Guess I'll have to look forward to participating in next year's SU2 conference.

sun5k:
> Hi @sun5k, could you please add also the configuration files to reproduce these results? I would try to run these simulations too with my implementation of the transition model.
> 
> Thank you in advance!

Hi @rois1995  Of course! 
However, I'm still under testing, so I can't guarantee the result if the numerical scheme or the other options are changed.
Which configuration file would you want?


By any chance, are you implementing or developing a transition model? 
Can you tell me what model you are willing to use if it's a model implementation? My next goal is to implement the AFT model or the k-w-gamma model, but I'm trying to implement a k-w-gamma model because it seems like the AFT model is being implemented by the other developer.

rois1995:
> > Hi @sun5k, could you please add also the configuration files to reproduce these results? I would try to run these simulations too with my implementation of the transition model.
> > Thank you in advance!
> 
> Hi @rois1995 Of course! However, I'm still under testing, so I can't guarantee the result if the numerical scheme or the other options are changed. Which configuration file would you want?
> 
> By any chance, are you implementing or developing a transition model? Can you tell me what model you are willing to use if it's a model implementation? My next goal is to implement the AFT model or the k-w-gamma model, but I'm trying to implement a k-w-gamma model because it seems like the AFT model is being implemented by the other developer.

If you could send me the config files for the T3A, T3A- and NLF test cases it would be awesome!

I have implemented the same model as yours, the LM model, in parallel to you 'cause I needed it for some simulations. I also tried implementing the same version but related to the SA turbulence model, however it still is bothering me that I cannot seem to get good results on it. 

If you want, I also tried the prolate spheroid test-case with good results. If you need I can give the configs and the meshes. The only thing about this test-case is that it is more related to the cross-flow transition, thus it needs to be implemented, which I think I'll do in these days.

I think that the k-w-gamma model is the way to go right after the implementation of the LM model since it has lots of equations in common. Do you think that you need another class for it or you will write on the same classes that you have already built (like the CTransLMSolver)?

sun5k:
> > > Hi @sun5k, could you please add also the configuration files to reproduce these results? I would try to run these simulations too with my implementation of the transition model.
> > > Thank you in advance!
> > 
> > 
> > Hi @rois1995 Of course! However, I'm still under testing, so I can't guarantee the result if the numerical scheme or the other options are changed. Which configuration file would you want?
> > By any chance, are you implementing or developing a transition model? Can you tell me what model you are willing to use if it's a model implementation? My next goal is to implement the AFT model or the k-w-gamma model, but I'm trying to implement a k-w-gamma model because it seems like the AFT model is being implemented by the other developer.
> 
> If you could send me the config files for the T3A, T3A- and NLF test cases it would be awesome!
> 
> I have implemented the same model as yours, the LM model, in parallel to you 'cause I needed it for some simulations. I also tried implementing the same version but related to the SA turbulence model, however it still is bothering me that I cannot seem to get good results on it.
> 
> If you want, I also tried the prolate spheroid test-case with good results. If you need I can give the configs and the meshes. The only thing about this test-case is that it is more related to the cross-flow transition, thus it needs to be implemented, which I think I'll do in these days.
> 
> I think that the k-w-gamma model is the way to go right after the implementation of the LM model since it has lots of equations in common. Do you think that you need another class for it or you will write on the same classes that you have already built (like the CTransLMSolver)?


The test case(T3A, T3A-, and NLF-0416) configuration files were uploaded to my git repository. Here is the [link](https://github.com/sun5k/SU2/tree/develop/TestCases/transition  ).

Suggestions for the prolate spheroid test-case is appreciated. However, my transition model is the LM-2009, and I have no plans for the 2015-LM model. Therefore, the cross-flow transition is left for future developers.


I think the CTransGammaSolver, CTransGSolver, or the other name of classes will be added for the k-w-gamma transition model. It's a vague idea, I haven't really thought about detail yet.

SU2_CFD/include/numerics/turbulent/transition/trans_sources.hpp
```diff
@@ -0,0 +1,290 @@
...
+  AD::SetPreaccIn(Vorticity_i, 3);
+  
+
+  unsigned short iDim;
+  
+  su2double Corr_Rec, Corr_F_length, Corr_Ret, F_length;
+  su2double F_onset1, F_onset2, F_onset3, F_onset;
+  su2double f_turb, Re_v, f_sub, r_omega;
+  su2double Pg, Dg, Pthetat;
+  su2double diverg;
```
WallyMaier, line 108: Introduce these where they are used please.

sun5k: This is NASA turbulence modeling resources of LM model (TMR) [SST-2003-LM2009](https://turbmodels.larc.nasa.gov/langtrymenter_4eqn.html) 

All variables are almost the same in Langtry's [paper](https://doi.org/10.2514/1.42362) and [dissertation.](https://elib.uni-stuttgart.de/handle/11682/1722)


F_onset1, F_onset2, Fonset3, and F_onset of line 105 is the transition onset function that triggers the transition.


f_turb is used to disable the destruction/relaminarization source outside of a laminar boundary layer or in the viscous sublayer and is defined as follows :  
image

Re_v  is vorticity Reynolds number : 
image

f_sub is forcing F_length to always be equal to its maximum value (in this case 40.0) in the viscous sublayer : 
image

r_omega is Reynolds number base on omega : 
image


 Pg, Dg, and Pthetat are production and destruction terms of each transport equation.

diverg is divergence of velocity.

WallyMaier: Sorry, I mean instead of mass declaring the variables at the top of a function, declare the variables when they are used ->

i.e. "su2double F_onset = ......"

SU2_CFD/include/numerics/turbulent/transition/trans_sources.hpp
```diff
@@ -0,0 +1,290 @@
...
+
+  Velocity_Mag = sqrt(vel_u*vel_u + vel_v*vel_v + vel_w*vel_w);
+
+  AD::SetPreaccIn(V_i[idx.Density()], V_i[idx.LaminarViscosity()], V_i[idx.EddyViscosity()]);
+
+  Density_i = V_i[idx.Density()];
+  Laminar_Viscosity_i = V_i[idx.LaminarViscosity()];
+  Eddy_Viscosity_i = V_i[idx.EddyViscosity()];
+
+  Tu = 100.0*sqrt( 2.0 * ScalarVar_i[0] / 3.0 ) / Velocity_Mag;
```
WallyMaier, line 143: For example
```suggestion
  su2double Tu = 100.0*sqrt( 2.0 * ScalarVar_i[0] / 3.0 ) / Velocity_Mag;
```

sun5k: Ok, now I understand what you mean. There is nothing special reason. It just habitually declared all the variables above. So I'll fix all the variables declared that same way.

sun5k:
"[WIP] Develop for Langtry and Menter transition model. #1592" was merged while trying to update my remote repository to v7.40. I'm not good at using GitHub, so I'm sorry for this situation. 

My remote repository(forked) is not synchronized with v7.40, and I don't know what the problem is. So, I will delete the forked repository and fork it again to PR with the same name and content. Fortunately, the LM model updated to v7.40 obtained the same residuals and results as the previous v7.3.1.

PR #1595, opened 2022-04-06
Fix reference pressure for Cp with moving frames (e.g. ROTATING_FRAME)

### LINKED ISSUE #1588: pysu2: NaNs in GetFlowLoad() in combination with dynamic / moving mesh
**Describe the bug**
*A clear and concise description of what the bug is and what you expect the behavior to be instead. If applicable, add screenshots to help explain your problem.*

This issue is a follow-up of a discussion here: https://www.cfd-online.com/Forums/su2/241889-rotating_frame-grid-velocity-correct-flow-physics-wrong.html

Summary: To calculate an aircraft during pitch-up as a steady Euler simulation, I use the dynamic / moving mesh module with GRID_MOVEMENT= STEADY_TRANSLATION and/or ROTATING_FRAME and set the farfield mach number = 0.0. The results look fine, but when reading the nodal forces f_xyz on the surface via pysu2 and GetFlowLoad(), I now receive NaNs... My first idea was that maybe the forces are calculated from the pressure coefficients, which show "strange" values in Tecplot, probably due to the division by V_inf=0.0.

**To Reproduce**
*Please provide a config file and mesh for a minimal example that recreates the problem.*
I derived a test case from the Naca0012 example and added a short python script that should reproduce the NaNs in line 66.
[testcase_dynamic_mesh.zip](https://github.com/su2code/SU2/files/8396025/testcase_dynamic_mesh.zip)

 - OS: Red Hat Enterprise Linux Workstation 7.9
 - C++ compiler and version: mpicc (gcc 4.8.5 "gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)")
 - MPI implementation and version: MPICH 3.4.2
 - SU2 Version:  7.3.0 "Blackbird"

pcarruscag:
Fixes #1588
cc @ArneVoss 
I have not fully tested the python wrapper forces, but they should be ok for dimensional simulations.
This is WIP until I add a regression test.


pcarruscag:
python forces are also ok for non-dimensional simulations

ArneVoss:
Pedro, 
I cherry-picked your your commit [c2018c8](https://github.com/su2code/SU2/pull/1595/commits/c2018c836791944383c13d32dcfec903d5657072) and tested it with my configuration. Looks good so far, I haven't see any issues / problems on my side. Thank you very much!

PR #1548, opened 2022-02-16
Changed time-averaging algorithm to reduce performance impact.

ChristianBauerEng:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
 - During time-averaging, the weighted sum over previous timesteps is re-used to quickly update the time-average during inner iterations.
 - Square-windowing now does not save the entire sample history, which should eliminate performance degradation when time-averaging is performed over many timesteps.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*
#1545 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [ ] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary. (UnitTest)
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #836, opened 2019-12-11
Added Windowing Regularizers for time averaged outputs.

SteffenMath:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
Based on the observations of Albring et al. in "Challenges in Sensitivity Computations for (D)DES and URANS" we implemented the windowing regularizers proposed in the work of  Krakos et al. "Sensitivity analysis of limit cycle oscillations" and adapted it to the AD framework of SU2. 
We call the feature in the following "windowing". 
Windowing can be applied if one takes the (windowed) time average of a periodic function, i.e. in the case of an instationary flow field that exhibits periodic behaviour. 
 - We changed the functionality of the output-field-prefix TAVG and D_TAVG from a simple time average to a time averaged weighted with a windowing function. For the direct run this is just a post-processing step and changes happen only in signal_processing_toolbox and Coutput. 
- We adapted the seeding for the adjoint run from a the classic time average to a windowed time average. 
-  Tests have shown, that higher order windowing regularizers, tend to converge quickly with respect to the averaged time span. We have implemented a "timeConvergenceMonitor", that acts similarly to the convergence monitoring of the inner convergence. It applies a cauchy convergence criterion to the chosen windowed time average output coefficients. 
- The cauchy convergence criterion works for direct run, adjoint run and shape optimization scripts
- In case of shape optimization, the script remembers the final time iteration of the direct run. If the final time step is smaller than the "original" final time step stated in the configuration file, it adapts the starting time of the adjoint run; and the time to seed the objective function, such that the seeding does not exceed the original time to end the seeding.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags).
- [X] My contribution is commented and consistent with SU2 style.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [x] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1601, opened 2022-04-11
Fix running averages for restarted simulations

TobiKattmann:
Hi,

When restarted at e.g. Time-Iter 2 the first value would be divided by 3 (number of timesteps counting from zero).
So if your first value in that restarted sim is a 3, the running average would be 1.

This is fixed with this commit by setting WINDOW_START_ITER = RESTART_ITER in the Config Postprocessing,
if the option is not set. If the option is set, but incorrectly, an error is raised.

## Related Work
Original implementation #836 
Change in previous behavior introduced in #1548 

## PR Checklist


TestCases/hybrid_regression.py
```diff
@@ -436,7 +436,7 @@ def main():
     square_cylinder.cfg_dir   = "unsteady/square_cylinder"
     square_cylinder.cfg_file  = "turb_square.cfg"
     square_cylinder.test_iter = 3
-    square_cylinder.test_vals = [-1.162564, 0.066401, 1.399788, 2.220402]
+    square_cylinder.test_vals = [-1.162564, 0.066401, 1.399788, 2.220402, 1.399743, 2.218605]
```
TobiKattmann, line 439: I'll prob have to adapt this according to what the workflow says, as the available values are already slightly different to the MPI-only values and I am too lazy to set up a compilation with OMP

pcarruscag (approved):

TestCases/hybrid_regression.py
TobiKattmann:
```suggestion
    square_cylinder.test_vals = [-1.162564, 0.066401, 1.399788, 2.220402, 1.399743, 2.218603]
```
Tiny change to have exactly the workflow values. The Test didn't fail as it was the difference was below the threshold.

TobiKattmann:
Thanks for the review :) Once the tests pass with this tiny commited regression value change for the hybrid case - I'll merge this in

pcarruscag:
Sounds good thanks for the fix.

WallyMaier (approved):

PR #1615, opened 2022-04-22
Try to detect wrong meson being used to configure SU2

pcarruscag:
Many issues caused by users doing `meson build` instead of `./meson.py build`, maybe this will make it less likely... In hindsight we should not have called it `meson.py`


bigfooted (approved):
good idea!

PR #1627, opened 2022-05-08
Newton-Raphson for NEMO temperature computations

WallyMaier:
## Proposed Changes
This replaces the bisection method in NEMO to compute the Tve values.  The NR solver should be faster and offer some speed up, I believe the ComputeTemperatures function is the slowest in NEMO.

So I have some cases running to highlight the speedup (hopefully not a slowdown).  From a quick glance, it seems the iteration time is very sensitive to the scaling factor.  

Lastly, this will probably destroy the regression tests for NEMO, since our temperatures will be slightly different now. 

## Related Work

## PR Checklist


SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,78 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
   /* Determine if the temperature lies within the acceptable range */
   if      (T < Tmin) T = Tmin;
   else if (T > Tmax) T = Tmax;
 
   /*--- Set vibrational temperature algorithm parameters ---*/
+  su2double NRtol         = 1.0E-6;    // Tolerance for the Newton-Raphson method
   su2double Btol          = 1.0E-6;    // Tolerance for the Bisection method
   unsigned short maxBIter = 50;        // Maximum Bisection method iterations
+  unsigned short maxNIter = 50;        // Maximum Newton-Raphson iterations
+  su2double scale         = 0.5;       // Scaling factor for Newton-Raphson step
```
WallyMaier, line 2019: This scaling factor of 0.5 would result in a slow down (~25%).  But 0.9 seems to be a 20% speed increase.....

WallyMaier, line 2019:
```suggestion
  su2double scale         = 0.9;       // Scaling factor for Newton-Raphson step
```

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,78 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
-      break;
+  bool NRconvg = false;
+  su2double rhoEve_t = 0.0, rhoCvve = 0.0;
+  vector<su2double> val_eves;
+  vector<su2double> val_cvves;
+
+  /*--- Newton-Raphson Method --*/
+  for (unsigned short iIter = 0; iIter < maxNIter; iIter++) {
+    rhoEve_t = rhoCvve = 0.0;
+    val_eves  = ComputeSpeciesEve(Tve);
```
pcarruscag, line 2034: If these functions return references you should use const auto& "X" = "f" to avoid copies in this loop.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,78 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
   /*--- Set vibrational temperature algorithm parameters ---*/
+  su2double NRtol         = 1.0E-6;    // Tolerance for the Newton-Raphson method
   su2double Btol          = 1.0E-6;    // Tolerance for the Bisection method
   unsigned short maxBIter = 50;        // Maximum Bisection method iterations
+  unsigned short maxNIter = 50;        // Maximum Newton-Raphson iterations
+  su2double scale         = 0.9;       // Scaling factor for Newton-Raphson step
 
+  /*--- Execute a Newton-Raphson root-finding method for Tve ---*/
   //Initialize solution
   Tve = T;
```
pcarruscag, line 2023: Is this the best initialization possible? Could you pass in the value of Tve found in the previous iteration.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,78 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
 
   /* Determine if the temperature lies within the acceptable range */
   if      (T < Tmin) T = Tmin;
   else if (T > Tmax) T = Tmax;
 
   /*--- Set vibrational temperature algorithm parameters ---*/
+  su2double NRtol         = 1.0E-6;    // Tolerance for the Newton-Raphson method
   su2double Btol          = 1.0E-6;    // Tolerance for the Bisection method
   unsigned short maxBIter = 50;        // Maximum Bisection method iterations
+  unsigned short maxNIter = 50;        // Maximum Newton-Raphson iterations
```
pcarruscag, line 2018: Const everything you can please.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,78 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
+  bool NRconvg = false;
+  su2double rhoEve_t = 0.0, rhoCvve = 0.0;
+  vector<su2double> val_eves;
+  vector<su2double> val_cvves;
+
+  /*--- Newton-Raphson Method --*/
+  for (unsigned short iIter = 0; iIter < maxNIter; iIter++) {
+    rhoEve_t = rhoCvve = 0.0;
+    val_eves  = ComputeSpeciesEve(Tve);
+    val_cvves = ComputeSpeciesCvVibEle(Tve);
```
pcarruscag, line 2035: What is the relative cost of computing f and df for the Newton step? If df is much more expensive you can switch to the secant method after converging a bit.

pcarruscag (approved):

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: Probably not.
I wasnt sure if it would be worth it try and pass a Tve value along, but I can try it.

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: Thats a good question.  I tried running WRT_PERFORMANCE=YES option, but I didnt get too much granularity. 
Is there a formal way to check? or is using tic/toc fine?

pcarruscag: You'd need to either use a profiler, or manually instrument like you're saying.

WallyMaier: Using the secant method would require the previous Tve value, and as of right now, theres no good way of getting that value.  Ill try to figure out a way to get.  Im not sure if its worth creating a new variable vector to store data.

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: Update: I have made the method use a value of Tve_old.  As expected, it is much faster (~2times as fast with the case I've been running).

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: After looking into it, there doesn't seem to be much performance gain using the secant method.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,76 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
     } else {
-      if (rhoEve_t > rhoEve) Tve2 = Tve;
-      else                  Tve_o = Tve;
+      Tve = Tve2;  
     }
   }
+
+  // If the Newton-Raphson method has converged, assign the value of Tve.
+  // Otherwise, execute a bisection root-finding method
+  Tve_o = Tvemin; Tve2 = Tvemax;
```
jtneedels, line 2057: Why is having the bisection as a backup still necessary? Is this how it was implemented in feature_TNE2? Have any of the test cases you tried required the use of the bisection to solve for value? Are there specific circumstances where bisection will work where NR will fail? The only examples I can think of are bad initialization or if there is some weirdness/non-smoothness in the function.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,76 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
+  if (Tve_old<Tvemin) Tve_old = Tvemin; else if (Tve_old>Tvemax) Tve_old = Tvemax;
 
   /*--- Set vibrational temperature algorithm parameters ---*/
-  su2double Btol          = 1.0E-6;    // Tolerance for the Bisection method
-  unsigned short maxBIter = 50;        // Maximum Bisection method iterations
+  const su2double NRtol         = 1.0E-6;    // Tolerance for the Newton-Raphson method
+  const su2double Btol          = 1.0E-6;    // Tolerance for the Bisection method
+  const unsigned short maxBIter = 50;        // Maximum Bisection method iterations
+  const unsigned short maxNIter = 50;        // Maximum Newton-Raphson iterations
+  const su2double scale         = 0.9;       // Scaling factor for Newton-Raphson step
```
jtneedels, line 2019: Do you have a sense of the stability of the method wrt to learning rate? 0.9 ~ 1, seems like you could get away without this parameter. Are there cases for which the method fails to converge for large learning rates, but converges for small ones? Would it make sense to have this be an adaptive parameter? (i.e. scale = 1, if NR does not converge, scale = 0.2, break?)

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,76 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
+
+    for (iSpecies = 0; iSpecies < nSpecies; iSpecies++){
+      rhoEve_t += rhos[iSpecies] * val_eves[iSpecies];
+      rhoCvve += rhos[iSpecies] * val_cvves[iSpecies];
+    } 
+
+    /*--- Find the roots ---*/
+    su2double f  = rhoEve - rhoEve_t;
+    su2double df = -rhoCvve;
+    Tve2 = Tve - (f/df)*scale;
```
jtneedels, line 2043: Similar to the question on learning rate above, do we know any properties about the minimization problem we are solving here that help us guarantee convergence, or maybe can help us demonstrate robustness?

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,76 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
+        else                  Tve_o = Tve;
+      }
+    }
+  }
+
   // If absolutely no convergence, then assign to the TR temperature
-  if (!Bconvg) {
+  if (!NRconvg && !Bconvg ) {
     Tve = T;
     cout <<"Warning: temperatures did not converge, error= "<< fabs(rhoEve_t-rhoEve)<<endl;
```
jtneedels, line 2077: Have you tried this on any cases where we previously didn't get convergence? Does it help with that, or does it just speed up the solver?

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: It was how the original TNE2 was implemented, and generally doesn't seem to get activated except the first iteration.

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: From my limited test cases, I haven't seen any issue with stability.  I think I can test once I find a case where we seen temperature non-convergence.

I think having the parameter adaptive could be interesting, though may not be necessary if 0.9 (or 1 is sufficient).

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: I have a case with the Fire2 waiting to be run, I will report back on that.

SU2_CFD/src/fluid/CSU2TCLib.cpp
jtneedels: So at the first iteration, NR fails and it switches to bisection, then it switches back to NR? That sounds like a problematic initialization.

WallyMaier: Yeah, the issue is the first iteration where the primitive vector isn't set  ie:
Tve_old = V[TVE_INDEX] in the solver is actually just doing Tve_old = 0.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -2004,39 +2003,77 @@ vector<su2double>& CSU2TCLib::ComputeTemperatures(vector<su2double>& val_rhos, s
...
-  su2double Tmin  = 50.0; su2double Tmax = 8E4;
-  su2double Tve_o = 50.0; su2double Tve2 = 8E4;
+  const su2double Tmin   = 50.0; const su2double Tmax   = 8E4;
+  const su2double Tvemin = 50.0; const su2double Tvemax = 8E4;
+  su2double Tve_o  = 50.0; su2double Tve2  = 8E4;
 
   /* Determine if the temperature lies within the acceptable range */
-  if      (T < Tmin) T = Tmin;
-  else if (T > Tmax) T = Tmax;
+  if (Tve_old < 1) Tve_old = T;                           //For first fluid iteration
```
WallyMaier, line 2011: During the first fluid iteration, the Tve_old is zero.  This causes the newton solver to fail.

This line fixes that issue, and seemed to be least intrusive fix, though open to other ideas.

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: I added the boolean above.

SU2_CFD/src/fluid/CSU2TCLib.cpp
oleburghardt: Is it a backtracking method like e.g. the Armijo rule used for gradient descent/line search what you have in mind? Such a thing could at least prevent you from divergence.
Maybe adding a solver class for Newton-like methods to somewhere in Common/ could make sense at some point :)

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: The new temperature routine avoid the "did not converge" errors, but did not improve the solution quality.

jtneedels (approved):

PR #1638, opened 2022-05-18
Bugfix for a call to LoadUnsteady_Solution in CDiscAdjFluidIteration::Preprocess

MaxSagebaum:
The parameters in iInst and iZone are swapped which gives a segmentation fault for adjoint HB computations.


pcarruscag (approved):
Thanks Max, seems that we have that mistake also in lines 150 and 152, can you update them please?

TobiKattmann (approved):
Thanks for this fix Max :bouquet:

PR #1661, opened 2022-06-08
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -10,8 +10,13 @@ on:
       - 'develop'
       - 'master'
 
+permissions:
+  contents: read
+
 jobs:
   build:
+    permissions:
+      contents: none
     name: Build SU2
     strategy:
       fail-fast: false
@@ -54,6 +59,8 @@ jobs:
           name: ${{ matrix.config_set }}
           path: install/bin
   regression_tests:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
     name: Regression Tests
     needs: build
@@ -94,6 +101,8 @@ jobs:
           # -t <Tutorials-branch> -c <Testcases-branch>
           args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
   unit_tests:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
     name: Unit Tests
     needs: build


martin-g:
> even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

But if an attacker compromises the workflow then (s)he will be able to change the permissions too, no ?

.github/workflows/regression.yml
```diff
@@ -10,8 +10,13 @@ on:
       - 'develop'
       - 'master'
 
+permissions:
+  contents: read
```
martin-g, line 14: Here you set the workflow permission to `read` and then for each job to `none`.
Why not set it to `none` on workflow level directly ?

pcarruscag:
@naveensrinivasan Thank you for the contribution, please reply to Martin's questions so we can get this moving.

.github/workflows/regression.yml
naveensrinivasan: The contents read is required for cloning the repo. Now I notice the code isn't cloning the repo.

martin-g:
@naveensrinivasan Could you please comment on https://github.com/su2code/SU2/pull/1661#issuecomment-1149557138 too? Thanks!

naveensrinivasan:
> > even if the attackers will succeed in compromising your workflow, they won’t be able to do much.
> 
> But if an attacker compromises the workflow then (s)he will be able to change the permissions too, no ?

Yes, AFAIK. That is why it is recommended to Pin by SHA

.github/workflows/regression.yml
martin-g: So, should it be `none` at workflow level ?

martin-g:
> Yes, AFAIK. That is why it is recommended to Pin by SHA

OK! I see your point here! Indeed the reduced permissions would help if a malicious action is executed!

naveensrinivasan:
> > Yes, AFAIK. That is why it is recommended to Pin by SHA
> 
> OK! I see your point here! Indeed the reduced permissions would help if a malicious action is executed!

👍

PR #1664, opened 2022-06-09
Small misc fixes

pcarruscag:
Leftovers from other PR's


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6456,7 +6456,7 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
   unsigned short iDim;
   unsigned long iVertex, iPoint;
   su2double P_Total, T_Total, Velocity[MAXNDIM], Velocity2, H_Total, Temperature, Riemann,
-  Pressure, Density, Energy, *Flow_Dir, Mach2, SoundSpeed2, SoundSpeed_Total2, Vel_Mag,
+  Pressure, Density, Energy, Flow_Dir[MAXNDIM] = {0.0}, Mach2, SoundSpeed2, SoundSpeed_Total2, Vel_Mag,
```
bigfooted, line 6459: Nice catch. But why initialize Flow_Dir and not Velocity (same for UnitNormal and Normal)?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6510,13 +6510,17 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
 
         /*--- Total properties have been specified at the inlet. ---*/
 
-        case INLET_TYPE::TOTAL_CONDITIONS:
+        case INLET_TYPE::TOTAL_CONDITIONS: {
```
bigfooted, line 6513:
```suggestion
        case INLET_TYPE::TOTAL_CONDITIONS:
```

bigfooted: not sure if you forgot something here or if this is a typo?

bigfooted:
Poiseuille_profile is now failing with very small changes in residuals. Is this due to the PR? We sometimes have these failures due to very strict comparison of residuals, so could be that.

bigfooted (approved):
LGTM, Thank you for your service!

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: Sometimes the compiler will complain that some of the values in the array may be used uninitialised. For example if we mix loops over nDim with loops over MAXNDIM.

pcarruscag:
It's probably just the extra division or extra brackets but I'll make sure it's not the initialization of the arrays.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6614,16 +6618,20 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
           V_inlet[nDim+3] = Energy + Pressure/Density;
 
           break;
+        }
+        /*--- Mass flow has been specified at the inlet. ---*/
 
-          /*--- Mass flow has been specified at the inlet. ---*/
-
-        case INLET_TYPE::MASS_FLOW:
+        case INLET_TYPE::MASS_FLOW: {
```
TobiKattmann, line 6624: What is the intention behind wrapping each `case` in curly braces? I can only think of some variable-scope implications, but I guess the `switch` already introduces that variable scope "barrier" :thinking:

TobiKattmann (approved):
:bouquet: lgtm

SU2_CFD/src/solvers/CEulerSolver.cpp
pcarruscag: It's the only way to introduce local variables inside switch cases.

PR #1417, opened 2021-11-02
Addition of catalytic isothermal wall boundary conditions in SU2-NEMO

jtneedels:
## Proposed Changes
Addition and cleanup of isothermal catalytic wall boundary, integration with rest of code. Code optimization of catalytic wall routines, removal of array allocations.

Submitting this draft PR primarily to get comments on approach, style, not looking to PR into develop quite yet.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1698, opened 2022-07-04
Add Heat Flux values to some regressions

WallyMaier:
## Proposed Changes
This PR will add heat flux to the values that are checked in the regression testing.  This should help capture more of the code performance. 

## Related Work
@jtneedels noticed a possible issue of changes made in #1417.  This will help avoid any potential unwanted changes.


pcarruscag (approved):

TestCases/ddes/flatplate/ddes_flatplate.cfg
```diff
@@ -226,4 +227,4 @@ SURFACE_ADJ_FILENAME= surface_adjoint
 OUTPUT_WRT_FREQ= 1000
 %
 % Screen output
-SCREEN_OUTPUT= (TIME_ITER, INNER_ITER, RMS_DENSITY, RMS_NU_TILDE, LIFT, DRAG)
```
WallyMaier, line 229: I removed the TIME_ITER to help with the regression scripts.   Though the "SIM_VALUES" that are used in the regression don't seem to be the final simulation residual/coefficient values.

WallyMaier:
@pcarruscag @martin-g I'm not sure how the regressions should run using the AARCH64 tests. I made some changes, but everything is still passing.

TestCases/ddes/flatplate/ddes_flatplate.cfg
pcarruscag: That may not be ideal, without time iter the script will look at time step 0, we should test more than that.
With time iter it will look at the first inner iteration for the timestep you want.

pcarruscag:
Looks like the stator was just the intermittent hybrid AD problem that is being worked on.
As for the nozzle... it did not have an objective function specified, therefore it must have been defaulting to drag, which was zero because MARKER_MONITORING was empty, now you added something to monitor and puff, results changed.

So before the case was not testing anything :shrug: you have my blessing to do whatever you want, check if the new adjoints make some sense and update (good samaritan option), or blindly update, or don't modify the case, or delete it, up to you :)

PR #1703, opened 2022-07-09
[WIP] Radiative equilibrium wall bounday condition

jtneedels:
## Proposed Changes
*This PR adds the capability of a radiative equilibrium wall boundary condition to the standard air and NEMO solvers. 
 

## PR Checklist


SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1077,6 +1078,102 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+    // Note: Contributions from qtr and qve are used for proportional control
+    //       to drive the solution toward the specified heatflux more quickly.
+    const auto V      = nodes->GetPrimitive(iPoint);
+    const auto GradV  = nodes->GetGradient_Primitive(iPoint);
+    su2double dTdn   = 0.0;
+    su2double dTvedn = 0.0;
+    for (auto iDim = 0u; iDim < nDim; iDim++) {
+      dTdn   += GradV[T_INDEX][iDim]*Normal[iDim];
+      dTvedn += GradV[TVE_INDEX][iDim]*Normal[iDim];
+    }
```
WallyMaier, line 1126: Theres a geometry tool box function for this I think

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1077,6 +1078,102 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+    su2double eddy_viscosity = nodes->GetEddyViscosity(iPoint);
+    for (unsigned short iSpecies=0; iSpecies<nSpecies; iSpecies++)
+      Mass += V[iSpecies]/V[RHO_INDEX]*Ms[iSpecies];
+    Cptr = V[RHOCVTR_INDEX]/V[RHO_INDEX]+Ru/Mass;
+    tmp1 = Cptr*(eddy_viscosity/Prandtl_Turb);
+    scl  = tmp1/ktr;
+    ktr += Cptr*(eddy_viscosity/Prandtl_Turb);
+    kve  = kve*(1.0+scl);
+    //Cpve = V[RHOCVVE_INDEX]+Ru/Mass;
+    //kve += Cpve*(val_eddy_viscosity/Prandtl_Turb);
```
WallyMaier, line 1146: Im pretty sure this can be removed from all case BC's.   Eddy viscosity in NEMO is always zero at the wall.

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1077,6 +1078,102 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+    su2double Twall = V[T_INDEX];
+    /*--- Compute residual ---*/
+      Res_Visc[nSpecies+nDim]   += (ktr*dTdn + kve*dTvedn - emissivity*STEFAN_BOLTZMANN*pow(Twall,4))*Area;
+      Res_Visc[nSpecies+nDim+1] += (kve*dTvedn)*Area; //TODO: check addition of radiation for Tve?
```
WallyMaier, line 1150:
```suggestion
    Res_Visc[nSpecies+nDim]   += (ktr*dTdn + kve*dTvedn - emissivity*STEFAN_BOLTZMANN*pow(Twall,4))*Area;
    Res_Visc[nSpecies+nDim+1] += (kve*dTvedn)*Area; //TODO: check addition of radiation for Tve?
```

WallyMaier (changes requested):

WallyMaier:
This looks pretty good!  I think you should add a regression.

Should an incompressible version be added as well?

SU2_CFD/src/solvers/CNEMONSSolver.cpp
jtneedels: Yeah good point, I'll change this in both the isothermal and rad wall BCs.

SU2_CFD/src/solvers/CNEMONSSolver.cpp
jtneedels: So remove lines 1130 - 1146?

jtneedels:
> This looks pretty good! I think you should add a regression.
Thanks! Yeah good point, I'll work on making one.

> Should an incompressible version be added as well?
Mmmm, that's a good question. I think physically for aerodynamic heating loads to be sufficiently high, you'd need to have a high enough Ma to ned the compressible solver. But maybe there are cases in the incompressible solver where you're dealing with high temperature flows and want to model the wall radiation out? Maybe this is already implemented in a more complex way?

SU2_CFD/include/solvers/CNEMONSSolver.hpp
```diff
@@ -215,4 +215,20 @@ class CNEMONSSolver final : public CNEMOEulerSolver {
...
+    
+   /*!
+    * \brief Impose the Navier-Stokes boundary condition (strong).
+    * \param[in] geometry - Geometrical definition of the problem.
+    * \param[in] solver_container - Container vector with all the solutions.
+    * \param[in] conv_numerics - Description of the numerical method for convective terms.
+    * \param[in] visc_numerics - Description of the numerical method for viscous terms.
+    * \param[in] config - Definition of the particular problem.
+    * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
```
pcarruscag, line 227:
```suggestion
    */
```

Common/include/CConfig.hpp
```diff
@@ -6833,6 +6836,13 @@ class CConfig {
    */
   su2double GetWall_HeatFlux(string val_index) const;
 
+  /*!
+   * \brief Get the emissivity of radiative equilibrium wall boundary.
+   * \param[in] val_index - Index corresponding to the radiative equilibrium wall boundary.
+   * \return The emissivity.
+   */
+  su2double Get_Emissivity(string val_index) const;
```
pcarruscag, line 6844:
```suggestion
  su2double Get_Emissivity(const string& val_index) const;
```

Common/src/CConfig.cpp
```diff
@@ -8952,6 +8971,15 @@ su2double CConfig::GetWall_HeatFlux(string val_marker) const {
   return Heat_Flux[0];
 }
 
+su2double CConfig::Get_Emissivity(string val_marker) const {
```
pcarruscag, line 8974:
```suggestion
su2double CConfig::Get_Emissivity(const string& val_marker) const {
```

SU2_CFD/include/solvers/CNEMONSSolver.hpp
```diff
@@ -215,4 +215,20 @@ class CNEMONSSolver final : public CNEMOEulerSolver {
                                CNumerics *visc_numerics,
                                CConfig *config,
                                unsigned short val_marker) override;
+    
+   /*!
+    * \brief Impose the Navier-Stokes boundary condition (strong).
```
pcarruscag, line 220: Out of date?

SU2_CFD/include/solvers/CNSSolver.hpp
```diff
@@ -203,6 +203,19 @@ class CNSSolver final : public CEulerSolver {
   void BC_HeatTransfer_Wall(const CGeometry *geometry,
                             const CConfig *config,
                             const unsigned short val_marker) override;
+    
+    /*!
+     * \brief Impose a heat flux by prescribing a heat transfer coefficient and a temperature at infinity.
```
pcarruscag, line 208: Copy pasted from the wrong place.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -1258,6 +1258,23 @@ class CSolver {
...
                                               unsigned short val_marker) { }
+  
+   /*!
+    * \brief A virtual member.
+    * \param[in] geometry - Geometrical definition of the problem.
+    * \param[in] solver_container - Container vector with all the solutions.
+    * \param[in] conv_numerics - Description of the numerical method.
+    * \param[in] visc_numerics - Description of the numerical method.
+    * \param[in] config - Definition of the particular problem.
+    * \param[in] val_marker - Surface marker where the boundary condition is applied.
```
pcarruscag, line 1269: I know we have this for most boundaries, but let me give you something to think about. This "a virtual member" is useless from the point of view of documenting the interface.
The place where functions are declared virtual defines the interface of the class, the basic purpose of this function should be stated here, then, when the function is overridden by derived classes they should only document what is specific of that class, towards implementing what the interface exposes... but anyway SU2 being SU2.

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1077,6 +1078,102 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+
+    /*--- Impose the value of the velocity as a strong boundary
+    condition (Dirichlet). Fix the velocity and remove any
+    contribution to the residual at this node. ---*/
+    su2double zero[MAXNDIM] = {0.0};
+    nodes->SetVelocity_Old(iPoint, zero);
+
+    for (auto iDim = 0u; iDim < nDim; iDim++)
+      LinSysRes(iPoint, nSpecies+iDim) = 0.0;
```
pcarruscag, line 1159: use the indices

SU2_CFD/src/solvers/CNEMONSSolver.cpp
```diff
@@ -1077,6 +1078,102 @@ void CNEMONSSolver::BC_Smoluchowski_Maxwell(CGeometry *geometry,
...
+
+    /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
+    if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+    /*--- Compute dual-grid area and boundary normal ---*/
+    const auto Normal = geometry->vertex[val_marker][iVertex]->GetNormal();
+    su2double Area = GeometryToolbox::Norm(nDim, Normal);
+
+    /*--- Initialize the convective & viscous residuals to zero ---*/
+    for (auto iVar = 0u; iVar < nVar; iVar++) {Res_Visc[iVar] = 0.0;}
```
pcarruscag, line 1113: Try not to use these "helper" member variables of the solver `Res_Visc` they make everything worse.
```suggestion
    su2double Res_Visc[MAXNVAR] = {0.0};
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -613,6 +613,144 @@ su2double CNSSolver::GetCHTWallTemperature(const CConfig* config, unsigned short
...
+
+      /*--- Compute the normal gradient in temperature using Twall ---*/
+
+      su2double dTdn = -(There - Twall)/dist_ij;
+
+      /*--- Apply a weak boundary condition for the energy equation.
+       Compute the residual due to the prescribed heat flux. ---*/
+
+      su2double Res_Conv = 0.0;
+      su2double Res_Visc = Area * (thermal_conductivity * dTdn - emissivity*STEFAN_BOLTZMANN*pow(Twall,4));
```
pcarruscag, line 700: Too much copy pasting from somewhere else :smile: 

What is the theory behind this formulation?
To what extent would it be possible and physical to combine this with the other types of wall boundary?

SU2_CFD/src/solvers/CNEMONSSolver.cpp
WallyMaier: Yes exactly.

SU2_CFD/src/solvers/CNSSolver.cpp
jtneedels: Screen Shot 2022-07-11 at 12 24 46 AM

jtneedels: So the theory behind this BC is that as your wall temperature increases, physically it will start to radiate according to the Stefan-Boltzmann Law ~ q_rad = emissivity*SB_constant*(T_w^4 - T_surrounding^4). T_surrounding is generally taken as 0, unless there is a more complex treatment of the shock layer radiation (which we don't have in SU2) so I've dropped it here. Then, for an insulated wall, you'll have the convective and radiative heat fluxes in equilibrium. Adding this to the viscous flux residual and driving to 0, you impose a radiative equilibrium wall temperature at the wall. Here's a comparison of the temperature you get with an adiabatic wall vs radiative equilibrium over a range of Mach. Basically, by incorporating the large radiative heat flux you get at high temperatures, the radiative wall gives you a much more accurate prediction of wall temp.

wall_temp_comp

pcarruscag: I see, thank you for the explanation.
In that case I guess it would not be appropriate to combine this with the heat flux boundary, and certainly not with isothermal... Which is sad because there is more boiler plate you'll have to add.

Common/src/CConfig.cpp
```diff
@@ -8952,6 +8971,15 @@ su2double CConfig::GetWall_HeatFlux(string val_marker) const {
   return Heat_Flux[0];
 }
 
+su2double CConfig::Get_Emissivity(const string& val_marker) const {
```
pcarruscag, line 8974: Here in CConfig.cpp look for `Roughness_Height` and `Kind_Wall`, it's used to specify roughness on no-slip boundaries for the SA model. Bonus points if you make the existing code a tiny bit less copy-pastie (you'll see what I mean)

SU2_CFD/src/solvers/CNSSolver.cpp
jtneedels: Yeah, I'm wondering how sustainable the current structure of adding a new function for each BC is. @WallyMaier floated the idea of a generic viscous wall BC, where most of the code is shared but the residual/Jacobian terms are edited based on the BC type (isothermal, HF, etc.). Is this something we might use to combat code bloat?

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -613,6 +613,144 @@ su2double CNSSolver::GetCHTWallTemperature(const CConfig* config, unsigned short
   return Twall;
 }
 
+void CNSSolver::BC_RadiativeEquilibrium_Wall(CGeometry *geometry,
+                                             CSolver **solver_container,
+                                             CNumerics *conv_numerics,
+                                             CNumerics *visc_numerics,
+                                             CConfig *config,
+                                             unsigned short val_marker){
+
```
pcarruscag, line 622: It can still have it's own function but that would simply call either BC_Isothermal_Wall_Generic or BC_HeatFlux_Wall_Generic (whichever looks closer to what you want, heat flux I suppose).

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -613,6 +613,144 @@ su2double CNSSolver::GetCHTWallTemperature(const CConfig* config, unsigned short
...
+      if (implicit) {
+
+        /*--- Add contributions to the Jacobian from the weak enforcement of the energy equations. ---*/
+
+        su2double Density = nodes->GetDensity(iPoint);
+        su2double Vel2 = GeometryToolbox::SquaredNorm(nDim, &nodes->GetPrimitive(iPoint)[prim_idx.Velocity()]);
+        su2double dTdrho = 1.0/Density * ( -Twall + (Gamma-1.0)/Gas_Constant*(Vel2/2.0) );
+
+        Jacobian_i[nDim+1][0] = thermal_conductivity/dist_ij * dTdrho * Area;
```
pcarruscag, line 712: You need the radiation part here. Or did it make things less stable? If so leave a comment please.

SU2_CFD/src/solvers/CNSSolver.cpp
pcarruscag: It seems like you are describing exactly what we have in CNSSolver :rofl:

Those pictures are looking nice btw, they would look even better here https://su2code.github.io/tutorials/home/ or here https://su2code.github.io/vandv/home/ :smile:

WallyMaier: ^^*wink wink*

But I think its worth discussing at the dev meeting to figure out the best way forward with additional boundaries.

jtneedels: @pcarruscag  Good point, you mean in that way that the isothermal wall BC calls isothermal wall generic, etc.? Yeah I think I see what you are getting at in your comment below. And yes! Will definitely add to tutorials, currently looking for a V&V case, but before all of that putting together an SU2 conference abstract 😆 

@WallyMaier Yeah good point, I'll definitely join next Tuesday, it would be great to talk through how to do this most efficiently going forward.

WallyMaier:
Trying to unstale this.

PR #1704, opened 2022-07-11
Check if compilers support flags before using them

### LINKED ISSUE #1701: develop does not compile anymore
**Describe the bug**
*A clear and concise description of what the bug is and what you expect the behavior to be instead. If applicable, add screenshots to help explain your problem.*

some recent changes causes SU2/develop to fail to compile on older compilers. Specifically:
I cannot compile SU2 with gnu compilers older than 7.4.0. My next most recent compiler gnu/10.2.1 works.
I cannot compile SU2 with intel compilers < 17.0, which is the most recent compiler installed on our system
I cannot compile SU2 with openmpi support, even for the gnu compiler gnu/10.2.1 that works. 
EDIT: The HPC guys added openmpi/4.0.4 which, together with gnu/10.2.1, compiles
I have mpich on my local machine, together with gnu 10.3.0, which still works.
What I understood from @Cristopher-Morales is that even the most recent openmpi/4.1.3 does not work (@Cristopher-Morales  what was the gnu compiler version you used?). 

So this issue affects the RB HPC as well as the TU/e HPC.

pcarruscag:
Fixes #1701

### DIFF
--- a/externals/cgns/hdf5/meson.build
+++ b/externals/cgns/hdf5/meson.build
@@ -1,21 +1,25 @@
-if target_machine.system() == 'windows' or meson.get_compiler('c').get_id() == 'intel'
-  hdf5_default_warnings = []
-else
-  hdf5_default_warnings = ['-Wno-unused-result',
-                           '-Wno-unused-parameter',
-                           '-Wno-unused-variable',
-                           '-Wno-unused-but-set-variable',
-                           '-Wno-sign-compare',
-                           '-Wno-stringop-overflow',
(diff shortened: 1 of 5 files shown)


TobiKattmann (approved):
Thanks for this additional check and caring about the grey-haired, long-serving compilers :D

martin-g:
I was going to work today on adding gcc-7 to the new Ubuntu 20.04 based build images but these changes look better!

martin-g (approved):

PR #1726, opened 2022-07-28
Update catch.hpp to v2.13.9

bigfooted:
## Proposed Changes
updated catch.hpp from the catchorg github repository

## PR Checklist


bigfooted:
license did not change, so no need to update that file.

pcarruscag (approved):
LGTM

eli-schwartz:
I wonder if maybe it would be a better idea to delete this (and other stuff) from externals/ and add it to subrojects/ instead. You can use `meson wrap install catch2`, and it's a simple ini file to vet instead of an 18k line file. git diffs become a lot simpler too.

pcarruscag:
It's a bit more involved because we want to support "offline" mode. But that's what we do with dependencies that we cannot copy due to license restrictions.

With that said Nijso has probably fought enough with unit tests for the rest of the year, so if you want to send a pull request our way you are most welcome :)
Cheers

eli-schwartz:
Offline mode could be done by creating source code tarballs as release artifacts instead of using github's autogenerated archives.

Meson has a built-in tool for this, which automatically includes offline copies of used subprojects: `meson dist --include-subprojects`
See https://mesonbuild.com/Creating-releases.html#include-subprojects-in-your-release

PR #1731, opened 2022-08-09
Fix macos build

pcarruscag:
https://github.com/su2code/SU2/runs/7753910763?check_suite_focus=true


pcarruscag:
Fixed https://github.com/su2code/SU2/actions/runs/2828047564

PR #1752, opened 2022-09-11
Always use vectorization when the numerical scheme supports it

pcarruscag:
Recent changes for hybrid parallel AD made it simpler to use the physical reconstruction checks in vectorized schemes (an old TODO).
Consequently, the vectorized Roe implementation is now equivalent to the scalar version, and thus the user no longer needs to "opt-in".


WallyMaier:
This looks good to me!

SU2_CFD/include/numerics_simd/flow/convection/common.hpp
```diff
@@ -132,10 +132,12 @@ FORCEINLINE CPair<ReconVarType> reconstructPrimitives(Int iEdge, Int iPoint, Int
       break;
     }
     /*--- Detect a non-physical reconstruction based on negative pressure or density. ---*/
-    const Double neg_p_or_rho = fmax(fmin(V.i.pressure(), V.j.pressure()) < 0.0,
-                                     fmin(V.i.density(), V.j.density()) < 0.0);
+    /*--- Some weird issues with forward AD if this is all done in one line. ---*/
```
jblueh, line 135: What happened exactly?

pcarruscag: The result of the one line version was 1e-310, but I could not reproduce in a unit test, and also didnt get anything from valgrind. So i'm not sure, might have something to do with the expression types interfeering with each other?

pcarruscag: Instead of 0 or 1.

jblueh: I think I managed to reproduce it, I'll look into it a little further.

pcarruscag: Awesome thanks

SU2_CFD/include/numerics_simd/flow/convection/common.hpp
```diff
@@ -93,15 +93,15 @@ FORCEINLINE void musclEdgeLimited(Int iPoint,
 
 /*!
  * \brief Retrieve primitive variables for points i/j, reconstructing them if needed.
- * \param[in] iPoint, jPoint - Nodes of the edge.
+ * \param[in] iEdge, iPoint, jPoint - Edge and its nodes.
  * \param[in] muscl - If true, reconstruct, else simply fetch.
  * \param[in] vector_ij - Distance vector from i to j.
```
bigfooted, line 98:
```suggestion
 * \param[in] muscl - If true, reconstruct, else simply fetch.
 * \param[in] V1st - pair of compressible flow primitives for nodes i,j
 * \param[in] vector_ij - Distance vector from i to j.
```

SU2_CFD/include/numerics_simd/flow/diffusion/common.hpp
```diff
@@ -171,7 +171,8 @@ FORCEINLINE void addTauWall(Int iPoint, Int jPoint,
   Double tauWall_ij = (tauWall_i+tauWall_j) * isNormalEdge;
 
   /*--- Scale is 1 for those edges, i.e. tau is not changed. ---*/
-  Double scale = tauWall_ij / norm(tangentProjection(tau,unitNormal)) + (1.0-isNormalEdge);
+  Double scale =
+      tauWall_ij / fmax(norm(tangentProjection(tau,unitNormal)), EPS) + (1.0-isNormalEdge);
```
bigfooted, line 175: :+1: 
These div-by-zero risks are actually a huge drawback of using y+ instead of y* as the non-dimensional length scale. I'll look into this 'soon'.

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -269,13 +269,50 @@ class CFVMFlowSolverBase : public CSolver {
...
   void SumEdgeFluxes(const CGeometry* geometry);
 
+  /*!
+   * \brief Sums edge fluxes (if required) and computes the global error counter.
+   * \param[in] pausePreacc - Whether preaccumulation was paused durin.
+   * \param[in] localCounter - Thread-local error counter.
+   * \param[in,out] config - Used to set the global error counter.
+   */
+  inline void FinalizeResidualComputation(const CGeometry *geometry, bool pausePreacc,
+                                          unsigned long localCounter, CConfig* config) {
```
bigfooted, line 286: localcounter is const?

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -66,6 +66,23 @@ class CFlowVariable : public CVariable {
...
+   * \brief Updates the non-physical counter of an edge.
+   * \param[in] iEdge - Edge index.
+   * \param[in] isNonPhys - Should be 1 (true) if a non-physical reconstruction was occurred.
+   * \return Whether the reconstruction should be limited to first order, based on the counter.
+   */
+  template <class T>
+  inline T UpdateNonPhysicalEdgeCounter(unsigned long iEdge, const T& isNonPhys) const {
+    if (isNonPhys != 0) {
+      /*--- Force 1st order for this edge for at least 20 iterations. ---*/
+      NonPhysicalEdgeCounter[iEdge] = 21;
```
bigfooted, line 80: 20 iterations ought to be enough for everyone? Or should it be an option that can be set?  In any case, it might be a good idea to move the magic number to a more proper location?

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
pcarruscag: It is reused below.

SU2_CFD/include/variables/CFlowVariable.hpp
pcarruscag: 🤷 if it is not, it is probably better to just use first 1st order to start, it has been 20 since Tom implemented this. We dont really have a place for this type of constant, at least it is not repeated now.

Common/include/parallelization/vectorization.hpp
```diff
@@ -58,7 +58,17 @@ template<class T>
...
+#ifdef CODI_REVERSE_TYPE
+  /*--- Use a SIMD size of 1 for reverse AD, larger sizes increase
+   * the pre-accumulation time with no performance benefit. ---*/
+  return 1;
+#else
+  /*--- For forward AD there is a performance benefit. This covers
+   * forward AD and primal mode (su2double == passivedouble). ---*/
+  return PREFERRED_SIZE / sizeof(passivedouble);
+#endif
+}
```
jblueh, line 71: Thanks for adding this. :+1: I rechecked also with linear index handling and observed no benefits either.

pcarruscag: Yep maybe I confused it between non-vectorized and vectorized (which lumps more in a single preacc region).

SU2_CFD/include/numerics_simd/flow/convection/common.hpp
jblueh: The issue is indeed buried deep inside expression templates. It boils down to a reference to a local variable.

We have
```
  VecExpr::max_::operator[] (size_t i) -> Scalar const& { return std::max(u[i], v[i]); }
```
with the reference-returning template `std::max`. Both `u[i]` and `v[i]` are returned by
```
  VecExpr::lt_::operator[] (size_t i) const -> Scalar { return Scalar(u[i] < v[i]); }
```
by value, so that `std::max` returns a reference to one of these local variables, which becomes dangling when returned by `max_::operator[]`.

pcarruscag: I'm using codi::max for the implementation of max_, is this falling through to std::max somehow?

jblueh: Yes. There is `using std::max` in the `codi` namespace, and the `std` template prevails against the overloads of `max` for CoDiPack expressions in this case.

pcarruscag: And is that also the case for reverse mode?

jblueh: The overall expression template structure is the same for forward and reverse, so I guess that it affects the reverse mode too.

Working towards a solution, using the version of `max` that returns a CoDiPack expression is not sufficient either, because the CoDiPack expression would capture the local `su2double` returned by
```
VecExpr::lt_::operator[] (size_t i) const -> Scalar { return Scalar(u[i] < v[i]); }
```
by an eventually dangling reference. Since everything is passive anyway,
```
VecExpr::lt_::operator[] (size_t i) const -> passivedouble { return passivedouble(u[i] < v[i]); }
```
is a better solution and CoDiPack expressions capture `passivedouble` correctly by value. However, this reintroduces the return-by-reference issue in the shape of `std::max` instantiated for `passivedouble`.

I think the best way out is using `codi::fmin` and `codi::fmax` together with the `passivedouble` change. As soon as `su2double` or other CoDiPack expressions are involved, they will always produce a CoDiPack expression, and the `cmath` versions of `fmin` and `fmax` return by value.

jblueh: I updated CoDiPack and made `std::fmin` and `std::fmax` available in `namespace codi` (next to overloads for them that were already there).

Mind if I add the fix suggested above?

pcarruscag: Fmin and fmax are not in std why was that needed? Im not sure I like that fix.

pcarruscag: Because fmin and fmax are not suitable implementations of min max for ints

jblueh: The `std` is not the point, just that writing 
```
VecExpr::max_::operator[] (size_t i) -> /*auto deduced*/ { return codi::fmax(u[i], v[i]); }
```
requires the `cmath` version of `fmax` to be accessible like this, too. [https://en.cppreference.com/w/cpp/numeric/math/fmax](https://en.cppreference.com/w/cpp/numeric/math/fmax) looks to me as if they are in `std`, though?

jblueh: True, the suggested solution with `passivedouble` and `fmin`/`fmax` is specific to the `su2double` instantiation of the vector expressions. But maybe we could introduce some small traits that choose what I suggested for `su2double` and default to the previous solution for everything else?

pcarruscag: Usually fmin and fmax work without namespace specification.
We can use fmin/max as the implementation for now and leave a note that they will not be efficient for integers.
I dont think we use that at the moment, Ill double check when I have some time and make the changes, thanks.

jblueh: Argument-dependent lookup should select the CoDiPack overloads automatically. Alright.

pcarruscag: I think the fix covers everything, active to passive, and suitable overloads.
Thank you for getting to the root cause.

PR #1764, opened 2022-09-16
Cleanup/fix regression tests and other small fixes

jblueh:
## Proposed Changes
This PR started as a collection of small fixes for issues I came across during recent work with SU2, mostly in the build/test system, but then the focus shifted towards general cleanup of the regression tests.

## PR Checklist


meson.build
```diff
@@ -298,7 +298,8 @@ message('''---------------------------------------------------------------------
          Use './ninja -C @12@ install' to compile and install SU2
 '''.format(get_option('prefix')+'/bin', meson.source_root(), get_option('enable-tecio'), get_option('enable-cgns'),
            get_option('enable-autodiff'), get_option('enable-directdiff'), get_option('enable-pywrapper'), get_option('enable-mkl'),
-           get_option('enable-openblas'), get_option('enable-pastix'), get_option('enable-mixedprec'), get_option('enable-librom'), meson.build_root().split('/')[-1]))
+           get_option('enable-openblas'), get_option('enable-pastix'), get_option('enable-mixedprec'), get_option('enable-librom'),
+           meson.build_root().startswith(meson.source_root()) ? meson.build_root().split('/')[-1] : meson.build_root()))
```
jblueh, line 302: The purpose of this is to output a ninja build command suitable for copy-paste if the build directory is located outside the SU2 main directory.

TestCases/TestCase.py
```diff
@@ -68,8 +68,10 @@ def __init__(self,tag_in):
         self.cpu_arch = platform.processor()
         self.enabled_on_cpu_arch = ["x86_64", "aarch64"]
 
-        # These can be optionally varied 
-        self.su2_exec    = "SU2_CFD" 
+        # These can be optionally varied
+        self.su2_exec_prefix = ""
+        self.su2_exec        = "SU2_CFD"
+        self.su2_exec_suffix = ""
```
jblueh, line 74: `self.su2_exec` is usually more complicated than `SU2_CFD`, e.g.

- `mpirun -n 2 SU2_CFD`,
- `SU2_CFD -t 2`,
- some python script.

Those strings are not suitable as arguments for the `killall` command that should cleanup timed out tests, it would need `SU2_CFD` as a process name in the first two above examples, hence the distinction between the executable itself, and additional commands and parameters that go in front and after it, respectively.

TestCases/parallel_regression.py
```diff
@@ -1517,7 +1627,9 @@ def main():
     pywrapper_square_cylinder.cfg_file  = "turb_square.cfg"
     pywrapper_square_cylinder.test_iter = 3
     pywrapper_square_cylinder.test_vals = [-1.162664, 0.066378, 1.399789, 2.220404, 1.399743, 2.218605, -0.453170]
-    pywrapper_square_cylinder.su2_exec  = "mpirun -np 2 SU2_CFD.py --parallel -f"
+    pywrapper_square_cylinder.su2_exec_prefix = "mpirun -np 2"
+    pywrapper_square_cylinder.su2_exec  = "SU2_CFD.py"
+    pywrapper_square_cylinder.su2_exec_suffix = "--parallel -f"
```
pcarruscag, line 1632: All this extra test boilerplate is a bit cumbersome. Instead can the executable be written as some tuple or list?
And while this is being changed would it be too difficult to have a function that sets some suitable default like we do for the hybrid parallel tests?

jblueh: The hybrid AD regression tests have the advantage that they all use `su2_exec = "SU2_CFD_AD -t 2"`.  A loop over all tests takes care of setting the common exec. Also, all use the same test call, `test.run_test()`.

For the parallel regression tests, both vary, although the majority of tests still uses `test.run_test()`. Among those, we have
| # | command |
| --- | --- |
| 1 | compute_polar.py -i 11 |
| 37 | mpirun -n 2 SU2_CFD |
| 1 | mpirun -np 2 python launch_flatPlate_rigidMotion.py --parallel -f |
| 1 | mpirun -np 2 python launch_unsteady_CHT_FlatPlate.py --parallel -f |
| 2 | mpirun -np 2 SU2_CFD |
| 1 | mpirun -np 2 SU2_CFD.py --nZone 2 --fsi True --parallel -f |
| 4 | mpirun -np 2 SU2_CFD.py --parallel -f |
| 2 | parallel_computation_fsi.py -f |
| 72 | parallel_computation.py -f |
| 12 | SU2_CFD |

so that we could at least maintain lists for the three major groups, analogously to what is done in the hybrid AD regression tests. Anyone adding new tests would have to ensure that they are placed in the correct list.

Regarding the prefix/exec/suffix boilerplate, a tuple would have to always consist of three components (possibly ""), to ensure we can tell apart the exec from the rest.
```
  my_test.su2_exec = ("", "SU2_CFD_AD", "-t 2")
```
This is nice and short, but not self-explanatory and the "" components are a bit awkward.

Do you think something more involved like
```
  my_test.command = SU2Command(launch = "mpirun -n 2", exec = "SU2_CFD_AD", param = "-t 2")
```
where omitted keyword arguments would default to empty strings is worth the effort?

pcarruscag: We should phase out parallel_computation.py in favor of just calling SU2, it used to convert output files but that is redundant now.
So we could do something like a loop that sets the default unless something has been specified already. So long as 95% is covered by the default loop Im fine with anything.

jblueh:
Some test cases in `parallel_regression.py` use the command `"SU2_CFD"`, as far as I can see there is nothing parallel for these cases. The only difference to putting them in `serial_regression.py` seems to be that they are run by an MPI build. Is that what is supposed to be tested?

pcarruscag (approved):
Really nice cleanup :+1:

PR #1771, opened 2022-09-25
Improve doxygen documentation

pcarruscag:
I had some offline time at airports and on planes, so I made the documentation easier to build (see README) and navigate by placing classes and functions in groups (which show up as "modules").
It is not complete, but I am out of offline time now :)


WallyMaier (approved):
Thanks for doing this! LGTM!!

PR #1784, opened 2022-10-07
Fix axisymmetric divergence for SST production

bigfooted:
## Proposed Changes
 the axisymmetric terms in the production term of the SST model are not correct and the current implementation is prone to error. 

## Related Work
#1783 

## PR Checklist


SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -716,6 +712,8 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
     AD::SetPreaccIn(PrimVar_Grad_i, nDim + idx.Velocity(), nDim);
     AD::SetPreaccIn(Vorticity_i, 3);
     AD::SetPreaccIn(V_i[idx.Density()], V_i[idx.LaminarViscosity()], V_i[idx.EddyViscosity()]);
+    AD::SetPreaccIn(Coord_i[1]);
+    AD::SetPreaccIn(V_i[idx.Velocity() + 1]);
```
pcarruscag, line 716: You can remove these 2 lines from ResidualAxisymmetric

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -738,6 +736,8 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
       su2double diverg = 0.0;
       for (unsigned short iDim = 0; iDim < nDim; iDim++)
         diverg += PrimVar_Grad_i[iDim + idx.Velocity()][iDim];
+      if (axisymmetric && Coord_i[1]>EPS) diverg += V_i[idx.Velocity() + 1]/Coord_i[1];
```
pcarruscag, line 739:
```suggestion
      if (axisymmetric && Coord_i[1] > EPS) diverg += V_i[idx.Velocity() + 1] / Coord_i[1];
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -627,18 +627,14 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
     const su2double sigma_w_i = F1_i * sigma_w_1 + (1.0 - F1_i) * sigma_w_2;
 
-    /*--- Production ---*/
-    const su2double pk_axi = max(
-        0.0, 2.0 / 3.0 * rhov * k * ((2.0 * yinv * V_i[idx.Velocity() + 1] - PrimVar_Grad_i[idx.Velocity()+1][1] - PrimVar_Grad_i[idx.Velocity()][0]) / zeta - 1.0));
-    const su2double pw_axi = alfa_blended * zeta / k * pk_axi;
-
     /*--- Convection-Diffusion ---*/
     const su2double cdk_axi = rhov * k - (Laminar_Viscosity_i + sigma_k_i * Eddy_Viscosity_i) * ScalarVar_Grad_i[0][1];
     const su2double cdw_axi = rhov * w - (Laminar_Viscosity_i + sigma_w_i * Eddy_Viscosity_i) * ScalarVar_Grad_i[1][1];
```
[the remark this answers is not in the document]
bigfooted: ok, I've added the rhov-terms.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
bigfooted: ok, done!

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -627,18 +627,14 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
     const su2double sigma_k_i = F1_i * sigma_k_1 + (1.0 - F1_i) * sigma_k_2;
     const su2double sigma_w_i = F1_i * sigma_w_1 + (1.0 - F1_i) * sigma_w_2;
 
-    /*--- Production ---*/
-    const su2double pk_axi = max(
-        0.0, 2.0 / 3.0 * rhov * k * ((2.0 * yinv * V_i[idx.Velocity() + 1] - PrimVar_Grad_i[idx.Velocity()+1][1] - PrimVar_Grad_i[idx.Velocity()][0]) / zeta - 1.0));
-    const su2double pw_axi = alfa_blended * zeta / k * pk_axi;
-
     /*--- Convection-Diffusion ---*/
```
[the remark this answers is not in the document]
bigfooted: ok, done!

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -627,18 +624,20 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
     const su2double cdw_axi = rhov * w - (Laminar_Viscosity_i + sigma_w_i * Eddy_Viscosity_i) * ScalarVar_Grad_i[1][1];
 
     /*--- Add terms to the residuals ---*/
-    Residual[0] += yinv * Volume * (pk_axi - cdk_axi);
-    Residual[1] += yinv * Volume * (pw_axi - cdw_axi);
+ 
+    Residual[0] -= yinv * Volume * cdk_axi;
+    Residual[1] -= yinv * Volume * cdw_axi;
+
+    Jacobian_i[0][0] = yinv * Volume * rhov;
```
pcarruscag, line 636: I think the signs are wrong?
And this is being overwritten after the call to the function, you need to change the normal jacobian to increment instead of setting.

bigfooted: You mean for the Jacobian? Isn't it just d/dk (rhov * k) = rhov?

pcarruscag: You are subtracting from the residual, so the jacobian is missing a minus sign.

bigfooted: Ah, of course! Thanks!

pcarruscag:
Actually, can you elaborate on why the strain magnitude doesn't need adjustments?
Contrary to cartesian coordinates, a constant velocity in the radial direction causes some compression/expansion of the fluid and thus there should be a `(v / y)^2` missing.
Disclaimer: I am not a maths person and my [reference](https://eng.libretexts.org/Bookshelves/Mechanical_Engineering/Structural_Mechanics_(Wierzbicki)/01%3A_The_Concept_of_Strain/1.03%3A_Description_of_Strain_in_the_Cylindrical_Coordinate_System) is for solid mechanics :laughing:

bigfooted:
> Actually, can you elaborate on why the strain magnitude doesn't need adjustments? Contrary to cartesian coordinates, a constant velocity in the radial direction causes some compression/expansion of the fluid and thus there should be a `(v / y)^2` missing. Disclaimer: I am not a maths person and my [reference](https://eng.libretexts.org/Bookshelves/Mechanical_Engineering/Structural_Mechanics_(Wierzbicki)/01%3A_The_Concept_of_Strain/1.03%3A_Description_of_Strain_in_the_Cylindrical_Coordinate_System) is for solid mechanics 😆

Ah yes, I erroneously thought: 2D means only considering the r and z-directions and considered the 2D strain tensor S_ij.  But you are right, the stress tensor component $S_{\theta \theta} = \frac{u_r}{r}$ contributes to the total strain magnitude.

bigfooted:
OK, should this term be added to CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag? There is a call to GetStrainMag in SST and in SA, and in SetRoe_dissipation_NTS.

pcarruscag:
Good point, I think we can include the axisymmetric term there. And there is something else... That strain magnitude is including the divergence and so it's closer to the norm of the stress tensor than to the norm of the stress tensor, and thus it is not consistent with the literature, again... :facepalm:

bigfooted:
yes, I see it. Do you know why this is done? I think we should remove the Div term to make it compatible with literature. This will then probably change all turbulence regression tests.

bigfooted:
The divergence term comes from the idea to write strain rate as [part without volume change] + [part with volume change]: 
https://www.whoi.edu/cms/files/12.800_Chapter_3_%2706_25330.pdf (p 13)
Not sure what the advantage is of writing it like that in the code since the divergence terms should cancel again. I'm also not sure why we only add the divergence term when nDim==2. It looks wrong to me. 

In any case,magnitude of strain rate is $S=\sqrt{2S_{ij}S_{ij}}$, which is, written out for 3D Cartesian coordinates,

$\frac{1}{2}S^2 = S^2_{11} + S^2_{22} + S^2_{33} + S^2_{12} + S^2_{21} + S^2_{13} + S^2_{31} + s^2_{23} + S^2_{32} $

It's symmetric, so:

$\frac{1}{2}S^2 = S^2_{11} + S^2_{22} + S^2_{33} + 2S^2_{12} + 2S^2_{13} + 2s^2_{23}$

and using $S_{ij}=\frac{1}{2}(\frac{\partial u_i}{\partial x_j} + \frac{\partial u_j}{\partial x_i})$, and $\frac{\partial u}{\partial x} = u_x$ etc. : 

$\frac{1}{2}S^2 = u_x^2 + v_y^2 + w_z^2  + 2(\frac{1}{2}(u_y+v_x))^2 + 2(\frac{1}{2}(u_z+w_x))^2 + 2(\frac{1}{2}(v_z+w_y))^2$

$\frac{1}{2}S^2 = u_x^2 + v_y^2 + w_z^2  + \frac{1}{2}(u_y+v_x)^2 + \frac{1}{2}(u_z+w_x)^2 + \frac{1}{2}(v_z+w_y)^2$

For axisymmetry, all the terms with velocity component _w_ and the gradients in _z_ are zero, except that $w_z=\frac{v}{r}$

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -614,20 +614,16 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
     AD::StartPreacc();
```
pcarruscag, line 614:
```suggestion
    const su2double vy = nodes->GetVelocity(iPoint, 1);
    const su2double y = geometry->nodes->GetCoord(iPoint, 1);
    AD::StartPreacc();
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -614,20 +614,16 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
...
     AD::SetPreaccIn(VelocityGradient, nDim, nDim);
```
pcarruscag, line 615:
```suggestion
    AD::SetPreaccIn(VelocityGradient, nDim, nDim);
    AD::SetPreaccIn(vy, y);
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -614,20 +614,16 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
...
 
     for (unsigned long iDim = 0; iDim < nDim; iDim++) {
-      StrainMag(iPoint) += pow(VelocityGradient(iDim, iDim) - Div, 2);
+      StrainMag(iPoint) += pow(VelocityGradient(iDim, iDim), 2);
     }
-    if (nDim == 2) {
-      StrainMag(iPoint) += pow(Div, 2);
+    if (config.GetAxisymmetric()){
+      auto Coord = geometry->nodes->GetCoord(iPoint);
+      if ((Coord[1])>EPS) StrainMag(iPoint) += pow(nodes->GetVelocity(iPoint,1)/Coord[1], 2);
```
pcarruscag, line 626:
```suggestion
    if (config.GetAxisymmetric() && y > EPS) {
      StrainMag(iPoint) += pow(vy / y, 2);
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -639,6 +635,7 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
       StrainMag(iPoint) += 2.0*pow(0.5*(VelocityGradient(1,2) + VelocityGradient(2,1)), 2);
     }
 
+
```
pcarruscag, line 638:
```suggestion
```

pcarruscag:
yeah the sst test are not happy now of course, oh well, I can help re-running and updating.
Maybe we should go over the equations once more while we are at it :shushing_face:

bigfooted:
I think the sst-rewrite team did a pretty good job going over the equations for the restructuring, and this one went unnoticed because it happened outside of the turbulence modeling. We have to go through the equations anyway as part of the sst-v1994 bugfix PR...

bigfooted:
I have started updating the regression tests. 
One thing I would like to do is this: when a test uses RESTART=YES, then the restart file comes from the previous test, which is exactly the same except for RESTART=NO. In this way, restart files will always be up to date. Question is if it is ever important for a regression test to have restart files that are more fully converged.

pcarruscag:
I would say it is fundamental to have some cases that start from converged solutions.
IMO all the cases we have that run for 10 iterations are trash from the point of view of testing correctness, they only test that we don't have severe integration bugs.
Even the cases that start from a good restart are NOT covering whether the solver is able to reproduce the correct behavior when starting from scratch...

bigfooted:
Yes, I'd say that the V&V regression tests should definitely start from the converged solutions that were used to produce the results on the V&V website. 

But for cases that do not have V&V documentation, the regression tests are imo merely a way to detect and prevent unwanted changes/side-effects (so, what you said). If they need to be more than that, they should be upgraded to a V&V with proper documentation. For the cases that are not VandV I'd rather have a 'start' + 'restart' regression test, with restarts from the 'start' results. In that way the restart files are always generated from the current code.

bigfooted:
Also: I see that discrete adjoint cases fail with a segmentation fault. But this does not happen on my local machine if I run the testcase. Not sure what is going on here.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -745,6 +745,10 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
       su2double diverg = 0.0;
       for (unsigned short iDim = 0; iDim < nDim; iDim++)
         diverg += PrimVar_Grad_i[iDim + idx.Velocity()][iDim];
+      if (axisymmetric && Coord_i[1] > EPS) {
+        AD::SetPreaccIn(Coord_i[1]);
+        diverg += V_i[idx.Velocity() + 1] / Coord_i[1];
+      }
```
bigfooted, line 751: @pcarruscag looks like it was this Coord_i[1] that caused the segfault.

TestCases/hybrid_regression.py
```diff
@@ -253,11 +253,20 @@ def main():
...
     test_list.append(axi_rans_air_nozzle)
 
+    # Axisymmetric air nozzle (transonic)
+    axi_rans_air_nozzle_restart           = TestCase('axi_rans_air_nozzle_restart')
+    axi_rans_air_nozzle_restart.cfg_dir   = "axisymmetric_rans/air_nozzle"
+    axi_rans_air_nozzle_restart.cfg_file  = "air_nozzle_restart.cfg"
+    axi_rans_air_nozzle_restart.test_iter = 10
+    axi_rans_air_nozzle_restart.test_vals         = [-4.333049, 1.244961, -0.175781, 4.228753, -1515.600000]
+    axi_rans_air_nozzle_restart.test_vals_aarch64 = [-12.093539, -6.630357, -8.798732, -2.399130, -1938.200000]
+    test_list.append(axi_rans_air_nozzle_restart)
```
pcarruscag, line 268: We are not going to update restart cases like this

bigfooted: Clear. For the restart, I suggest we always converge the cases to the convergence criterion that is in the config file, so it is consistent and easy to update the restart.

pcarruscag: Makes sense.
For the record I think your idea for the restart is good for testing _the restart_ but we would need to run N + M in the first run, restart at N and measure at M (with reference values obtained automatically from N+M).
It's just that this doesn't cover the correctness of the solution, unless the first run is all the way to convergence, which we should do for small-enough cases.

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop_tmp -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```
pcarruscag, line 131: It would have been less work for both if we had gone with the division of labor I suggested initially

TestCases/incomp_navierstokes/streamwise_periodic/chtPinArray_2d/of_grad_findiff.csv.ref
```diff
@@ -1,2 +1,2 @@
...
++0               , 0.0             , -100000.01639127731, -1.1101999999881826e-08, 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , -0.0399999991462785, 0.0                , -2.1099999997220564      , 0.0                     , 2.1799999999627673  , 3.8000000007754053            , 319.9999980552093         , -39.999997625272954          , 318.0000007318995        , -39.999997625272954         , -1.400000004814217   , -139.99999737279722, 0.0                , -540.0000077315781, 1e-08
```
pcarruscag, line 2:
```suggestion
0               , 0.0             , -100000.01639127731, -1.1101999999881826e-08, 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , 0.0             , -0.0399999991462785, 0.0                , -2.1099999997220564      , 0.0                     , 2.1799999999627673  , 3.8000000007754053            , 319.9999980552093         , -39.999997625272954          , 318.0000007318995        , -39.999997625272954         , -1.400000004814217   , -139.99999737279722, 0.0                , -540.0000077315781, 1e-08           
```

TestCases/vandv.py
```diff
@@ -75,13 +84,13 @@ def main():
...
-    # SWBLI SST
-    swbli_sst           = TestCase('swbli_sst')
-    swbli_sst.cfg_dir   = "vandv/rans/swbli"
-    swbli_sst.cfg_file  = "config_sst.cfg"
-    swbli_sst.test_iter = 20
-    swbli_sst.test_vals = [-11.039738, -10.619752, -11.577767, -10.276345, -11.117432, -3.84103, 0.002315, -2.316445, -5.392453, 2.786]
-    test_list.append(swbli_sst)
+    # SWBLI - sst-v2003m restart
+    swbli_sst_restart           = TestCase('swbli_sst')
+    swbli_sst_restart.cfg_dir   = "vandv/rans/swbli"
```
pcarruscag, line 89: @bigfooted I'm running the V&V cases to make new restarts

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop_tmp -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c update_vandv -s ${{matrix.testscript}}
```
pcarruscag, line 131:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```

TestCases/vandv.py
```diff
@@ -50,21 +50,21 @@ def main():
...
-    test_list.append(flatplate_sst1994m)
-
-    # bump in channel - sst-v1994m
+    flatplate_sst1994m_restart           = TestCase('flatplate_sst1994m')
+    flatplate_sst1994m_restart.cfg_dir   = "vandv/rans/flatplate"
+    flatplate_sst1994m_restart.cfg_file  = "turb_flatplate_sst.cfg"
+    flatplate_sst1994m_restart.test_iter = 5
+    flatplate_sst1994m_restart.test_vals         = [-13.000480, -9.646253, -10.645471, -7.477528, -9.684807, -5.353455, 0.002809]
+    flatplate_sst1994m_restart.test_vals_aarch64 = [-13.000480, -9.646253, -10.645471, -7.477528, -9.684807, -5.353455, 0.002809]
+    test_list.append(flatplate_sst1994m_restart)
```
pcarruscag, line 59:
```suggestion
    flatplate_sst1994m           = TestCase('flatplate_sst1994m')
    flatplate_sst1994m.cfg_dir   = "vandv/rans/flatplate"
    flatplate_sst1994m.cfg_file  = "turb_flatplate_sst.cfg"
    flatplate_sst1994m.test_iter = 5
    flatplate_sst1994m.test_vals         = [-13.044282, -9.682503, -10.712657, -7.519027, -9.742878, -5.374663, 0.002809]
    flatplate_sst1994m.test_vals_aarch64 = [-13.044282, -9.682503, -10.712657, -7.519027, -9.742878, -5.374663, 0.002809]
    test_list.append(flatplate_sst1994m)
```

TestCases/vandv.py
```diff
@@ -75,12 +75,12 @@ def main():
...
     test_list.append(swbli_sa)
 
-    # SWBLI SST
+    # SWBLI - sst-v2003m
     swbli_sst           = TestCase('swbli_sst')
     swbli_sst.cfg_dir   = "vandv/rans/swbli"
     swbli_sst.cfg_file  = "config_sst.cfg"
     swbli_sst.test_iter = 20
-    swbli_sst.test_vals = [-11.039738, -10.619752, -11.577767, -10.276345, -11.117432, -3.84103, 0.002315, -2.316445, -5.392453, 2.786]
+    swbli_sst.test_vals = [-11.020298, -10.520189, -11.572148, -10.164444, -11.112109, -3.678967, 0.002339, -2.979149, -5.343362, 1.3401]
```
pcarruscag, line 83:
```suggestion
    swbli_sst.test_vals = [-11.030236, -10.542428, -11.668592, -10.210996, -11.121572, -3.875880, 0.002339, -2.472912, -5.464633, 2.786]
```

TestCases/vandv/rans/bump_in_channel/turb_bump_sst.cfg
```diff
@@ -56,13 +53,13 @@ CONV_RESIDUAL_MINVAL= -13
 MESH_FILENAME= mesh_bump_2d_0089x041.cgns
 MESH_FORMAT= CGNS
 MESH_OUT_FILENAME= mesh_out.su2
-SOLUTION_FILENAME= solution_flow.dat
+SOLUTION_FILENAME= restart_flow.dat
```
pcarruscag, line 56:
```suggestion
SOLUTION_FILENAME= solution_flow.dat

```

TestCases/vandv.py
```diff
@@ -54,17 +54,17 @@ def main():
...
 
-    # bump in channel - sst-v1994m
+    # bump in channel - sst-v1994m 
     bump_sst1994m           = TestCase('bump_sst1994m')
     bump_sst1994m.cfg_dir   = "vandv/rans/bump_in_channel"
     bump_sst1994m.cfg_file  = "turb_bump_sst.cfg"
     bump_sst1994m.test_iter = 5
-    bump_sst1994m.test_vals         = [-13.033896, -10.325323, -5.558745, 0.024576, 0.004967]
-    bump_sst1994m.test_vals_aarch64 = [-13.085614, -10.325319, -5.559183, 0.024576, 0.004967]
+    bump_sst1994m.test_vals         = [-13.005103, -9.889383, -10.555402, -7.584122, -10.312935, -5.470352, 0.004967]
```
pcarruscag, line 66:
```suggestion
    bump_sst1994m.test_vals         = [-13.033125, -9.924352, -10.574465, -7.579818, -10.322321, -5.537195, 0.004967]
```

PR #1796, opened 2022-10-26
BC inlet for SST solver based on local conditions at the Inlet

Cristopher-Morales:
## Proposed Changes

The Boundary conditions in the SST turbulence model are based on the free stream conditions. However if we mix two stream with different fluid properties at the inlets, the Turbulent Kinetic Energy, omega and eddy viscosity at the inlets should be based on these local fluid properties. So in this pull request, we aim to use the known fluid properties at the inlets in order to set the BC for the SST model. 
I can add a test case if it is needed.

## Related Work

This pull request is related to #1784. 

## PR Checklist


bigfooted:
It's what I feared: all SST tests that have inlet BCs are failing. Can it be set up in such a way that we end up with the same inlet boundary condition?

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -602,15 +602,35 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
...
+      Intensity = config->GetTurbulenceIntensity_FreeStream();
+      viscRatio = config->GetTurb2LamViscRatio_FreeStream();
+      VelMag2 = GeometryToolbox::SquaredNorm(nDim, Velocity_Inlet);
+
       /*--- Non-dimensionalize Inlet_TurbVars if Inlet-Files are used. ---*/
       su2double Inlet_Vars[MAXNVAR];
-      Inlet_Vars[0] = Inlet_TurbVars[val_marker][iVertex][0];
-      Inlet_Vars[1] = Inlet_TurbVars[val_marker][iVertex][1];
+      Inlet_Vars[0] = 3.0 / 2.0 * (VelMag2 * Intensity * Intensity);
+      Inlet_Vars[1] = density_inlet * Inlet_Vars[0] / (laminar_viscosity_inlet * viscRatio);
```
pcarruscag, line 622: You should check bounds to avoid too low or too high values

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -602,15 +602,35 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
...
       if (config->GetInlet_Profile_From_File()) {
         Inlet_Vars[0] /= pow(config->GetVelocity_Ref(), 2);
         Inlet_Vars[1] *= config->GetViscosity_Ref() / (config->GetDensity_Ref() * pow(config->GetVelocity_Ref(), 2));
       }
 
+      /*--- Obtain eddy viscosity at inlet boundary node ---*/
+      muT = density_inlet * Inlet_Vars[0] / Inlet_Vars[1];
+
+      /*-- Set eddy viscosity at inlet boundary node*/
+      nodes->SetmuT(iPoint, muT);
```
pcarruscag, line 632: This is not a good idea. It creates an inconsistency with the flow solver because a different viscosity will be computed in the post-processing of this solver, and does not work with MPI because this is only being done for domain points.

bigfooted: > This is not a good idea. It creates an inconsistency with the flow solver because a different viscosity will be computed in the post-processing of this solver, and does not work with MPI because this is only being done for domain points.

You mean it is sufficient to remove muT here and let CTurbSSTSolver::Postprocessing deal with it (postprocessing is called for the boundary)?

pcarruscag: It's a week BC so probably muT will not be exactly equal to this, also because the limits of the SST model may be active.
But that is probably ok.

bigfooted:
After a discussion with @Cristopher-Morales, the inlet boundary condition modification was moved to the UniformInlet function. 
At the moment the modification is restricted to species transport.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -1034,10 +1039,29 @@ su2double CTurbSSTSolver::GetInletAtVertex(su2double *val_inlet,
...
   if (config->GetMarker_All_KindBC(iMarker) == INLET_FLOW) {
+    const string Marker_Tag = config->GetMarker_All_TagBound(iMarker);
     for (unsigned long iVertex = 0; iVertex < nVertex[iMarker]; iVertex++) {
-      Inlet_TurbVars[iMarker][iVertex][0] = GetTke_Inf();
-      Inlet_TurbVars[iMarker][iVertex][1] = GetOmega_Inf();
+      if (config->GetKind_Species_Model() != SPECIES_MODEL::NONE) {
+        CFluidModel* FluidModel= nullptr;
+        FluidModel = new CFluidScalar(config->GetSpecific_Heat_Cp(), config->GetGas_Constant(), config->GetPressure_Thermodynamic(), config);
+        FluidModel->SetTDState_T(config->GetInlet_Ttotal(Marker_Tag), config->GetInlet_SpeciesVal(Marker_Tag));
+        const su2double VelMag2_inlet = pow(config->GetInlet_Ptotal(Marker_Tag),2);
```
pcarruscag, line 1048: What you had before was good... Now this is wrong for everything except velocity inlets

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -602,15 +602,35 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
 
       conv_numerics->SetPrimitive(V_domain, V_inlet);
 
+      /*--- Obtain flow velocity vector at inlet boundary node ---*/
+
+      su2double Velocity_Inlet[MAXNDIM] = {0.0}, density_inlet, laminar_viscosity_inlet, Intensity, viscRatio, VelMag2, muT;
```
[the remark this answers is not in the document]
Cristopher-Morales: Done, Thank you so much!!

Cristopher-Morales:
Hi @pcarruscag , Thank you so much for your comments. I went to some of the test cases that are failing and I noticed that some residuals are very different because they are being compared at 5 - 10 or 20 iterations but if I compare them after 100 iterations they are very similar. I have run the NICFD_nozzle in the tutorial test cases and I got the following results:
residuals
solid-line: develop
dash-line: proposed BC_INLET
Also I got the same results in both cases:
results_NICFD

I also run the turb_flatplate_sst.cfg test case in the vandv folder and I got the same residuals and the same solution. I read in the pull request of @bigfooted that he experienced the same issues with the test cases with the restart option. So I am not sure if the other test cases are failing for similar reasons. Do you have some insight about this? 
Thank you so much in advance :) .

pcarruscag (commented):
Let's wait for #1784 and then you can start updating what needs to be updated

TestCases/hybrid_regression.py
```diff
@@ -254,7 +254,7 @@ def main():
     axi_rans_air_nozzle_restart.cfg_dir   = "axisymmetric_rans/air_nozzle"
     axi_rans_air_nozzle_restart.cfg_file  = "air_nozzle_restart.cfg"
     axi_rans_air_nozzle_restart.test_iter = 10
-    axi_rans_air_nozzle_restart.test_vals         = [-12.094405, -6.649212, -8.899548, -2.491859, -1924.900000]
+    axi_rans_air_nozzle_restart.test_vals         = [-5.942825, -0.377882, -2.220841, 0.802589, -1924.900000]
```
pcarruscag, line 257: This is now how we update cases with restarts.
You need to run the case to convergence and update the restart file.

Cristopher-Morales: Hi Pedro, thank you so much for your comment. I have rerun the test cases that use the restart option and I have created a pull request for the testcases. Please let me know if this is correct. Thank you so much in advance!!

bigfooted: Thanks, Cristopher! When your PR is ready, I'll update the VandV website (config, images,...)

Cristopher-Morales:
> @Cristopher-Morales you can change the file .github/workflows/regression.yml to point to your branch with updated restarts (look for -c develop in that file). I thought I had commented this but maybe I forgot to press comment xD

Hi Pedro, I have updated the restart test cases. Please let me know if something else is missing. Thank you so much in advance!!

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -608,13 +608,43 @@ void CTurbSSTSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container, C
...
+        /*--- Obtain flow velocity vector at inlet boundary node ---*/
+
+        su2double Velocity_Inlet[MAXNDIM] = {0.0};
+        for (auto iDim = 0u; iDim < nDim; iDim++) Velocity_Inlet[iDim] = V_inlet[iDim + 1];
+        su2double Temperature_Inlet;
+        su2double Pressure_Inlet;
+        su2double Density_Inlet;
+        if (config->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE) {
+          Pressure_Inlet = V_inlet[nDim+1];
+          Density_Inlet = V_inlet[nDim + 2];
```
pcarruscag, line 630: The hardcoded `nDim + x` will not work with NEMO, let me think a bit.

pcarruscag (approved):

PR #1817, opened 2022-11-10
Fix regression branch

WallyMaier:
## Proposed Changes
A recent change from #1417, which I merged in had the regression set to feature_cat_wall  in the TestCases. This will update and fix that call.  I'm sure I will need to check additional fixes associated with this change.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c feature_cat_wall -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
       - name: Cleanup
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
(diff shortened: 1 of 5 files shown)


pcarruscag (changes requested):
Ok wait. @bigfooted also merged something to testcases with new restarts for the work in axisymmetry fix.

WallyMaier:
@pcarruscag @bigfooted I can see the changes. It looks like a restart filename changed.  But I can hold off on fixing the regression value and merging the changes.

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:220614-1237
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c feature_cat_wall -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c develop_tmp -s ${{matrix.testscript}}
```
pcarruscag, line 131: Update what you need to update but do not change this branch to develop for now

pcarruscag (approved):

WallyMaier:
@pcarruscag I saw this got merged.  What do you think the next steps are with develop_tmp?

pcarruscag:
We remove it after #1784 is merged

PR #1825, opened 2022-11-21
Generalized, quasi 3D look-up table

EvertBunschoten:
## Proposed Changes
Generalize the variables in the CLookUpTable class for use in general look-up cases (not only progress variable-enthalpy).
Add dimension to CLookUpTable class to allow for quasi-3D interpolation (linear interpolation between two trapezoidal maps stacked in the third dimension).
 
I added a unit test as a test case with the quasi-3D table to show the correct use of syntax for 3D look-up operations.

## Related Work
Can be used in feature_multilayer_perceptron in the CDataDrivenFluid fluid model for look-up operations.

## PR Checklist


Common/include/containers/CLookUpTable.hpp
```diff
@@ -47,16 +47,19 @@ class CLookUpTable {
...
+  unsigned long *n_triangles;
   unsigned long n_variables;
-  unsigned long n_hull_points;
+  unsigned long *n_hull_points;
+  unsigned long n_table_levels = 1;
+  su2double *z_values_levels;
 
   /*! 
-   * \brief the lower and upper limits of the enthalpy and progress variable.
+   * \brief the lower and upper limits of the z, y and x variable.
```
bigfooted, line 58: In the combustion/manifold literature, the generalized formulation uses the terms "control variable". 
https://www.researchgate.net/publication/309147831_State-of-the-art_in_premixed_combustion_modeling_using_flamelet_generated_manifolds

EvertBunschoten: I replaced name_x and name_y in the CLookUpTable class by name_CV1 and name_CV2. I think it would be nice to keep the LUT class as general as possible so it's also convenient to use for other applications like data-driven fluid models.

Common/src/containers/CFileReaderLUT.cpp
```diff
@@ -128,13 +172,19 @@ void CFileReaderLUT::ReadRawLUT(const string& file_name) {
 
   /*--- now that n_variables, n_points, n_hull_points and n_variables are available, allocate memory ---*/
   if (rank == MASTER_NODE) cout << "allocating memory for the data, size = ( " << GetNVariables() << " , " << GetNPoints() << " )" << endl;
-  table_data.resize(GetNVariables(), GetNPoints());
+  table_data = new su2activematrix[n_levels];
```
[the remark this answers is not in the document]
EvertBunschoten: I replaced the std::vector with su2vector where possible.

Common/include/containers/CLookUpTable.hpp
bigfooted: Matteo also uses it now for NE thermo problems, right? Makes sense to not couple the independent variables to specific names. For the work with Kaiserslautern, the progress variable is actually a mixture fraction and the source term is set to 0 to simulate species mixing with a lookup table.

EvertBunschoten: Yes, we are using the tables for NICFD problems in other research with the control variables being density and static energy.

Common/src/containers/CLookUpTable.cpp
```diff
@@ -395,16 +572,20 @@
 
         /* first, copy the single triangle from the large triangle list*/
         for (int p = 0; p < 3; p++) 
-          triangle[p] = triangles[id_triangle][p]; 
+          triangle[p] = triangles[i_level][id_triangle][p];
```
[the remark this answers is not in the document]
EvertBunschoten: .

Common/src/containers/CFileReaderLUT.cpp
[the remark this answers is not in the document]
EvertBunschoten: Do you see other instances where ND container classes can be applied? Or can we resolve this conversation?

Common/src/containers/CFileReaderLUT.cpp
```diff
@@ -59,33 +59,77 @@ void CFileReaderLUT::ReadRawLUT(const string& file_name) {
...
       GetNextNonEmptyLine(file_stream, line);
       version_lut = line;
+
+      if(version_lut.compare("1.0.1") == 0){
+        table_dim = 2;
+        n_levels = 1;
+        found_level_count = true;
+        n_points      = new unsigned long[n_levels];
+        n_triangles   = new unsigned long[n_levels];
+        n_hull_points = new unsigned long[n_levels];
```
pcarruscag, line 77: yes, in any place where you are using `new`, replace by std::vector or su2vector

Common/src/containers/CFileReaderLUT.cpp
```diff
@@ -128,13 +172,19 @@ void CFileReaderLUT::ReadRawLUT(const string& file_name) {
...
-  triangles.resize(GetNTriangles(), 3);
+  triangles = new su2matrix<unsigned long>[n_levels];
+  for(unsigned long i_level=0; i_level<n_levels; i_level++)
+    triangles[i_level].resize(GetNTriangles(i_level), 3);
 
   if (rank == MASTER_NODE) cout << "allocating memory for the hull points, size = " << GetNHullPoints() << endl;
-  hull.resize(GetNHullPoints());
+  hull = new su2vector<unsigned long>[n_levels];
+  for(unsigned long i_level=0; i_level<n_levels; i_level++)
+    hull[i_level].resize(GetNHullPoints(i_level));
```
pcarruscag, line 187: If these cannot be a matrix then use std::vector<std::vector<>>, just don't use new basically.

EvertBunschoten: I see, thanks! Is this more memory efficient than the array? Or just a way to save some potential trouble with destructors?

Common/include/containers/CLookUpTable.hpp
```diff
@@ -44,48 +44,66 @@ class CLookUpTable {
...
+  su2vector<su2matrix<unsigned long>>triangles;
 
-  /* we do not know this size in advance until we go through the entire lookup table */
-  std::vector<std::vector<unsigned long> > edges;
-  std::vector<std::vector<unsigned long> > edge_to_triangle;
+  /*! \brief
+   * Edge information for each table level.
+   */
+  su2vector<std::vector<su2vector<unsigned long>>>edges;
+  su2vector<su2vector<std::vector<unsigned long>>>edge_to_triangle;
```
pcarruscag, line 91: Is it necessary to mix vector types?

Common/include/containers/CLookUpTable.hpp
```diff
@@ -167,27 +183,46 @@ class CLookUpTable {
...
+   * \param[in] val_z - z-coordinate of the query point.
+   * \param[in] lower_level - Table level index of the table level directly below the query point.
+   * \param[in] upper_level - Table level index of the table level directly above the query point.
+   * \param[in] lower_value - Result from x-y interpolation on the lower table level.
+   * \param[in] upper_value - Result from x-y interpolation on the upper table level.
+   * \param[in] val_vars - Pointer to interpolation result.
+  */
+  void Linear_Interpolation(const su2double val_z, const unsigned long lower_level, const unsigned long upper_level, su2double&lower_value,su2double&upper_value, su2double*&var_vals);
+
+
```
pcarruscag, line 198: clang-format the files please

Common/src/containers/CFileReaderLUT.cpp
```diff
@@ -184,33 +248,45 @@ void CFileReaderLUT::ReadRawLUT(const string& file_name) {
...
+        eoLevel = false;
+        triCounter = 0;
+        GetNextNonEmptyLine(file_stream, line);
+      }
+      if(line.compare("</Level>") == 0){
+        eoLevel = true;
+        if (n_triangles[levelCounter] != triCounter)
+        SU2_MPI::Error(
+            "Number of read points on level " + std::to_string(levelCounter) + " does not match number of points "
+            "specified in lookup table library header.",
```
pcarruscag, line 261: points or triangles?

Common/src/containers/CLookUpTable.cpp
```diff
@@ -53,20 +56,58 @@ CLookUpTable::CLookUpTable(const string& var_file_name_lut, const string& name_p
...
+    if(rank == MASTER_NODE){
+      su2double progress = su2double(i_level) / n_table_levels;
+      unsigned short barwidth = 65;
+      auto pos = barwidth * progress;
+      cout << "[";
+      for(int iBar=0; iBar<barwidth; ++iBar){
+          if(iBar < pos) cout << "=";
+          else cout << " ";
+      }
+      cout << "] "<< floor(100*progress) << " %\r";
```
pcarruscag, line 89: https://cplusplus.com/reference/iomanip/setfill/

Common/src/containers/CLookUpTable.cpp
```diff
@@ -79,32 +120,67 @@ void CLookUpTable::LoadTableRaw(const string& var_file_name_lut) {
...
+  limits_table_y.resize(n_table_levels);
+
+  /* we find the lowest and highest value of y and x in the table */
+  for(auto i_level=0u; i_level<n_table_levels; i_level++)
+  {
+    su2double min_x, max_x, min_y, max_y;
+    min_y = *min_element(&table_data[i_level][idx_Y][0], &table_data[i_level][idx_Y][0]+table_data[i_level].cols());
+    min_x = *min_element(&table_data[i_level][idx_X][0], &table_data[i_level][idx_X][0]+table_data[i_level].cols());
+    max_y = *max_element(&table_data[i_level][idx_Y][0], &table_data[i_level][idx_Y][0]+table_data[i_level].cols());
+    max_x = *max_element(&table_data[i_level][idx_X][0], &table_data[i_level][idx_X][0]+table_data[i_level].cols());
```
pcarruscag, line 173: https://en.cppreference.com/w/cpp/algorithm/minmax_element

Common/src/containers/CLookUpTable.cpp
```diff
@@ -117,41 +193,84 @@ void CLookUpTable::PrintTableInfo() {
...
+      cout << "| Average number of triangles:" << setw(36) << right << floor(n_tria_av) << " |" << endl;
+      cout << "| Average number of edges:" << setw(40) << right << floor(n_edges_av) << " |" << endl;
+      cout << "+------------------------------------------------------------------+" << endl;
+      cout << "| Minimum Z:" << setw(54) << right << limits_table_z.first << " |" << endl;
+      cout << "| Maximum Z:" << setw(54) << right << limits_table_z.second << " |" << endl;
+      cout << "| Minimum Y:" << setw(54) << right << min_y << " |" << endl;
+      cout << "| Maximum Y:" << setw(54) << right << max_y << " |" << endl;
+      cout << "| Minimum X:" << setw(54) << right << min_x << " |" << endl;
+      cout << "| Maximum X:" << setw(54) << right << max_x << " |" << endl;
+      cout << "+------------------------------------------------------------------+" << endl;
```
pcarruscag, line 244: Would it be too verbose if you printed the information per level? It would re-use the code for 1 level and maybe be more helpful for debugging.

Common/src/containers/CLookUpTable.cpp
```diff
@@ -291,8 +415,131 @@ void CLookUpTable::GetInterpMatInv(const su2double* vec_x, const su2double* vec_
...
 
-unsigned long CLookUpTable::LookUp_ProgEnth(const string& val_name_var, su2double *val_var, su2double val_prog,
-                                            su2double val_enth, const string& name_prog, const string& name_enth) {
+std::pair<unsigned long, unsigned long> CLookUpTable::FindInclusionLevels(const su2double val_z, bool*within_table_limits)
+{
+  /*--- Find the table levels with constant z-values directly below and above the query value val_z ---*/
+
+  /* Check if val_z lies outside table bounds */
+  if(val_z >= limits_table_z.second){
+    *within_table_limits = false;
```
pcarruscag, line 424: Why is this passed by pointer?

pcarruscag (commented):
Thanks for the changes, other than the comments below this LGTM, but the original authors should also review.

Common/include/containers/CLookUpTable.hpp
EvertBunschoten: Unfortunately yes. For 'edges', the outer vector can be an su2vector since the length is the number of table levels. The vector level below that is of inconsistent size, so it has to be an std::vector. The most inner vector is of length 2, so it can be an su2vector. For 'edge_to_triangle', it's similar.

Common/src/containers/CLookUpTable.cpp
EvertBunschoten: For small tables, maybe. I'm using the CLookUpTable class with 100+ levels. In such cases it would get too verbose in my opinion. I could make it a conditional statement where if the number of levels is like below 10, it will display all level info.

Common/include/containers/CLookUpTable.hpp
pcarruscag: If something has a known constant size it should be a std::array which will avoid heap allocation.

Common/src/containers/CLookUpTable.cpp
pcarruscag: I think it's ok to do just statistics then.

Common/src/containers/CLookUpTable.cpp
```diff
@@ -79,32 +120,67 @@ void CLookUpTable::LoadTableRaw(const string& var_file_name_lut) {
...
-  int ixProg = GetIndexOfVar(name_prog);
-
-  /* we find the lowest and highest value of enthalpy and progress variable in the table */
-  limits_table_enth[0] = *min_element(&table_data[ixEnth][0], &table_data[ixEnth][0]+table_data.cols());
-  limits_table_enth[1] = *max_element(&table_data[ixEnth][0], &table_data[ixEnth][0]+table_data.cols());
-  limits_table_prog[0] = *min_element(&table_data[ixProg][0], &table_data[ixProg][0]+table_data.cols());
-  limits_table_prog[1] = *max_element(&table_data[ixProg][0], &table_data[ixProg][0]+table_data.cols());
+void CLookUpTable::FindTableLimits(const string& name_cv1, const string& name_cv2) {
+  int idx_X = GetIndexOfVar(name_cv1);
+  int idx_Y = GetIndexOfVar(name_cv2);
```
bigfooted, line 162: Where possible, I think we should use the naming convention "cv1,cv2,cv3" for the controlling variables and avoid additional names for the manifold dimensions like X,Y,Z. At the level of the LUT we can indeed avoid the specific words  "enthalpy, progress variable, mixture fraction"
```suggestion
  int idx_cv1 = GetIndexOfVar(name_cv1);
  int idx_cv2 = GetIndexOfVar(name_cv2);
```

Common/src/containers/CLookUpTable.cpp
```diff
@@ -31,14 +31,17 @@
 
 using namespace std;
 
-CLookUpTable::CLookUpTable(const string& var_file_name_lut, const string& name_prog, const string& name_enth) {
+CLookUpTable::CLookUpTable(const string& var_file_name_lut, const string& name_x, const string& name_y) {
```
bigfooted, line 34:
```suggestion
CLookUpTable::CLookUpTable(const string& var_file_name_lut, const string& name_cv1, const string& name_cv2) {
```

Common/src/containers/CLookUpTable.cpp
```diff
@@ -31,14 +31,17 @@
...
 
   rank = SU2_MPI::GetRank();
 
+  name_CV1 = name_x;
+  name_CV2 = name_y;
+
   LoadTableRaw(var_file_name_lut);
 
-  FindTableLimits(name_prog, name_enth);
+  FindTableLimits(name_x, name_y);
```
bigfooted, line 44:
```suggestion
  FindTableLimits(name_cv1, name_cv2);
```

bigfooted, line 44: why introduce name_CV1 for name_x?

pcarruscag (commented):
Anything else on your side @bigfooted ?

bigfooted:
> Anything else on your side @bigfooted ?

@pcarruscag 
I had some small comments, besides that it looks good. 
@EvertBunschoten 
We can look into generalizing and merging the 2D and 3D LUT approach once the flamelet model is also in develop, but this also means changing the LUT file format. There are some open issues like how to best treat lookups that fall outside of the table, and the extention from the slab approach to the trapezoidal map approach would definitely help with the memory consumption. Not a show stopper right now, but would be nice to do this as part of a student project this year.

pcarruscag (approved):
Ok you can figure out the order you want to merge things.
LGTM

bigfooted (approved):
LGTM

Common/src/containers/CLookUpTable.cpp
```diff
@@ -275,140 +379,307 @@ void CLookUpTable::GetInterpMatInv(const su2double* vec_x, const su2double* vec_
...
+                                            std::vector<su2double*>& val_vars, su2double val_prog, su2double val_enth,
+                                            std::string name_prog, std::string name_enth) {
+  return LookUp_XY(val_names_var, val_vars, val_prog, val_enth);
+}
+
+unsigned long CLookUpTable::LookUp_ProgEnth(const std::vector<std::string>& val_names_var,
+                                            std::vector<su2double>& val_vars, su2double val_prog, su2double val_enth,
+                                            std::string name_prog, std::string name_enth) {
+  return LookUp_XY(val_names_var, val_vars, val_prog, val_enth);
+}
```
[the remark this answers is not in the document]
EvertBunschoten: I left them in to be compatible with feature_flamelet. We can remove them now

Common/src/containers/CLookUpTable.cpp
```diff
@@ -275,140 +379,307 @@ void CLookUpTable::GetInterpMatInv(const su2double* vec_x, const su2double* vec_
...
+  /* Interpolate the table limits to the query point coordinates. */
+  su2double x_min_q = x_min_lower + (x_min_upper - x_min_lower) * (val_CV3 - z_values_levels[lower_level]) /
+                                        (z_values_levels[upper_level] - z_values_levels[lower_level] + EPS);
+  su2double x_max_q = x_max_lower + (x_max_upper - x_max_lower) * (val_CV3 - z_values_levels[lower_level]) /
+                                        (z_values_levels[upper_level] - z_values_levels[lower_level] + EPS);
+  su2double y_min_q = y_min_lower + (y_min_upper - y_min_lower) * (val_CV3 - z_values_levels[lower_level]) /
+                                        (z_values_levels[upper_level] - z_values_levels[lower_level] + EPS);
+  su2double y_max_q = y_max_lower + (y_max_upper - y_max_lower) * (val_CV3 - z_values_levels[lower_level]) /
+                                        (z_values_levels[upper_level] - z_values_levels[lower_level] + EPS);
+
```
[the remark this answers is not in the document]
EvertBunschoten: Ok. I'll implement a routine which checks for duplicate z-values when loading the table.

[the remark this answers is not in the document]
bigfooted: I think we already check the validity in the table generator. I also think it is OK not to try and catch everything that can go wrong, especially when it involves checking the entire content of the table, loading it is already slow. Although with the recent rewrite, it is much faster now.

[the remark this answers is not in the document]
EvertBunschoten: Duplicate table levels were not checked for in the table generator or reader. The check I implemented is not computationally intensive; it only checks for duplicate z-values of the table levels.

PR #1826, opened 2022-11-22
[WIP] Flamelet approach for premixed flames

bigfooted:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
 This PR introduces a tabulated chemistry approach. The focus for this PR is on laminar and premixed flames in the incompressible framework, but it is generalizable to non-premixed and partially premixed flames, turbulent flames and compressible flames. 
In this approach, 2 transport equations are solved for progress variable and total enthalpy. The lookup table retrieves quantities (temperature, density, viscosity, diffusivity, source terms, etc) as a function of these independent control variables. 2 additional independent transport equations are solved for the species NO and CO.

## Related Work
*Resolve any issues (bug fix or feature request), note any related PRs, or mention interactions with the work of others, if any.*

## Nijso Checklist
- [ ] Replace bloated initialization (burnt and unburnt region) with python wrapper
- [ ] Check Conjugate Heat Transfer
- [ ] Check adjoint optimization workflow
- [x] Check axisymmetric setup
- [x] Check diffusivity
- [x] Check primal restart
- [x] Check adjoint restart
- [x] Adjoint comparison with FD

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [ ] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

PR #1839, opened 2022-12-01
Adding thermodynamics pressure as a user value for FLUID_MIXTURE model

Cristopher-Morales:
## Proposed Changes
 
The thermodynamic pressure influences the density at the inlet boundaries, especially when the mixture consists of more than 2 gases. So, with this pull request, we aim to obtain the pressure_thermodynamic as a user value or a default value, as 101325 [pa]. and based on this value compute the density. This is consistent with the low-mach approximation used in combustion.
I can add a test case if it is needed.

## Related Work
This is pull request is related to  #1826.

## PR Checklist


SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -48,14 +48,13 @@ CFluidScalar::CFluidScalar(su2double val_Cp, su2double val_gas_constant, su2doub
       n_species_mixture(config->GetnSpecies() + 1),
       Gas_Constant(val_gas_constant),
       Gamma(config->GetGamma()),
-      Pressure_Thermodynamic(value_pressure_operating),
```
pcarruscag, line 51: Is the value not constant throughout the simulation anymore?

Common/src/CConfig.cpp
```diff
@@ -1161,6 +1161,8 @@ void CConfig::SetConfig_Options() {
   addDoubleOption("GAS_CONSTANT", Gas_Constant, 287.058);
   /*!\brief GAMMA_VALUE  \n DESCRIPTION: Ratio of specific heats (1.4 (air), only for compressible flows) \ingroup Config*/
   addDoubleOption("GAMMA_VALUE", Gamma, 1.4);
+  /*!\brief THERMODYNAMIC_PRESSURE  \n DESCRIPTION: Thermodynamics(operating) Pressure (101325 Pa), only for incompressible flows) \ingroup Config*/
+  addDoubleOption("THERMODYNAMIC_PRESSURE", Pressure_Thermodynamic, 101325.0);
```
pcarruscag, line 1165: Sounds like having the right value is very important? Does it make sense to make this not optional?

SU2_CFD/src/fluid/CFluidScalar.cpp
Cristopher-Morales: Hi Pedro, thank you so much for your comments. It is still constant but now it is a user input. I have deleted this constant variable because after comparing with the CIncIdealGas.cpp I realized that the Pressure there has the same meaning as the Pressure used in the CFluidScalar. So I think this Pressure_Thermodynamic is redundant.

Common/src/CConfig.cpp
Cristopher-Morales: Hi Pedro, Yes, it is important because currently the thermodynamic pressure is computed in the CIncEulerSolver as

$p_{inc}=\rho_{inc}*R*T_{inc}$

where Inc means the INC_DENSITY_INIT and INC_TEMPERATUR_INIT and the R is the gas constant defined as $R_{u}/M$ , where $R_{u}$ is the universal gas constant and $M$ molecular weight of the first species (first gas). For a mixture of 2 gases, the $\rho_{inc}, T_{inc}, M$ can be set in the cfg file in order to obtain the desired operating pressure (usually 101325 pa for combustion) , but when the mixture is more than 2 gases, setting $\rho_{inc}, T_{inc}, R$ is more difficult because they depend on the composition at the inlets. So we could have two possibilities: 1) before setting the operating pressure , we could do two loops for computing the density and M in order to obtain the operating pressure or 2) set the operating pressure as a user input and  avoid these two additional loops, which is the way that we think is better and also allows using be default the use of the operating pressure which is usually considering in combustion problems (101325 pa) if the operating pressure is not provided in the cfg file. Please let me know f you have some suggestions/ideas about how this could be improved. Thank you so much in advance :) .

SU2_CFD/src/fluid/CFluidScalar.cpp
pcarruscag: If the value is constant please keep using a `const su2double` variable, the other models should have done the same.

Common/src/CConfig.cpp
pcarruscag: Understood, thank you for the explanation.
Then I would say it is important to document and make sure this is included in the examples.
I leave it up to you and @bigfooted to decide if you want to enforce this to have been set by the user.

SU2_CFD/src/fluid/CFluidScalar.cpp
Cristopher-Morales: Understood, I will keep the current implementation. Thank you so much for your comments!!

Common/src/CConfig.cpp
Cristopher-Morales: Thanks for the suggestions. We are working in examples, we plan soon to upload a tutorial explaining the composition-dependent model. Thank you so much!!

Cristopher-Morales: Hi Pedro, yesterday we decided with @bigfooted to enforce the thermodynamic pressure to be set by the user as it is in this pull request. Please let me know if there is something else to be done for this pull request. Thank you so much in advance!!

Cristopher-Morales: Hi Pedro, thanks for your comment. Yes, we want to enforce it, but we also want that if the thermodynamic pressure is not provided in the .cfg file, the default value will be used. Similar to what happens with gamma, when it is not provided, SU2 considers the default value as 1.4.

pcarruscag: That's not enforcing anything, that's doing nothing...
Update the config_template and put this options explicitly in the testcases that now use it, and document what it does there.

Cristopher-Morales: Understood. I will do it. Thanks for your comment!!

PR #1844, opened 2022-12-06
Heat solver using scalar framework - Part 1

pcarruscag:


bigfooted:
Nice work, looks good as always, so what is left for the WIP?

SU2_CFD/src/solvers/CHeatSolver.cpp
```diff
@@ -26,13 +26,16 @@
  */
 
 #include "../../include/solvers/CHeatSolver.hpp"
+#include <cstddef>
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
+#include "../../include/solvers/CScalarSolver.inl"
 
-CHeatSolver::CHeatSolver(CGeometry *geometry, CConfig *config, unsigned short iMesh) : CSolver(),
-             flow(config->GetFluidProblem()), heat_equation(config->GetHeatProblem()) {
+/*--- Explicit instantiation of the parent class of CTurbSolver. ---*/
```
WallyMaier, line 33: Should this comment have CScalarSolver?

pcarruscag: yes

pcarruscag, line 33:
```suggestion
/*--- Explicit instantiation of the parent class of CHeatSolver. ---*/
```

WallyMaier (approved):
LGTM!

PR #1856, opened 2022-12-25
Cleanup Linelets and create output to visualize them

pcarruscag:
I was playing with variants of this preconditioner (Linelet - GS, Linelet - LU-SGS), they don't seem to be any better than LU-SGS or ILU, or even Linelet - Jacobi :shrug: so I kept the Linelet - Jacobi we had but with cleaner implementation.


Common/include/geometry/CGeometry.hpp
```diff
@@ -186,7 +187,29 @@ class CGeometry {
 
   ColMajorMatrix<uint8_t> CoarseGridColor_;  /*!< \brief Coarse grid levels, colorized. */
 
-public:
+  /*--- Linelets (mesh lines perpendicular to stretching direction). ---*/
+ public:
```
pcarruscag, line 191:
```suggestion
 public:
  /*!< \brief Linelets (mesh lines perpendicular to stretching direction). */
```

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -29,8 +29,11 @@
 
 #include "../../include/geometry/CGeometry.hpp"
 #include "../../include/toolboxes/allocation_toolbox.hpp"
+#include "../../include/toolboxes/geometry_toolbox.hpp"
 
+#include <algorithm>
 #include <cmath>
+#include <limits>
```
pcarruscag, line 36:
```suggestion

#include <cmath>
```

SU2_CFD/include/numerics_simd/flow/convection/roe.hpp
```diff
@@ -168,7 +168,7 @@ class CRoeBase : public Base {
 
     VectorDbl<nVar> flux;
     for (size_t iVar = 0; iVar < nVar; ++iVar) {
-      flux(iVar) = kappa * (flux_i(iVar) + flux_j(iVar));
```
WallyMaier, line 171: Can kappa ever be changed?

WallyMaier (approved):
LGTM! Thanks!

SU2_CFD/include/numerics_simd/flow/convection/roe.hpp
pcarruscag: This was a bug that was fixed in the non-simd Roe scheme but was still here somehow.
kappa should only appear in the dissipation term, here we want the average of the fluxes, and anything different from 0.5 would be wrong.

pcarruscag: So yes kappa can be changed and a higher value results in lower dissipation (someone wrote it like that :shrug:)

PR #1862, opened 2022-12-30
Fix NEMO Supersonic Inlet BC  & BC Cleanup

jtneedels:

## Proposed Changes

Fixes supersonic inlet BC in CNEMOEulerSolver so that it becomes operational.
Clean up NEMO supersonic outlet BC.
Clean up NEMO farfield BC.
 
## Related Work
See associated PR to update documentation: https://github.com/su2code/su2code.github.io/pull/112

## PR Checklist


SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2148,6 +2148,7 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
 
 }
 
+// NECESSARY?
```
jtneedels, line 2151: Is BC_SUPERSONIC_INLET redundant with BC farfield, or should it remain it's own BC? If so, I can work on making it operational, but wanted to check before investing the time.

jtneedels:
Here's what I see:

- Viscous components of BC_INLET and BC_OUTLET commented out, can these be ucnommented?
- BC_INLET is non-operational.
- BC_SUPERSONIC_INLET is commented out and non-functional. Is this redundant in function to BC_FARFIELD, can it be removed, or should it be fixed?

jtneedels:
Here's my proposed changes:

- BC_supersonic_outlet seems to be covered by the automatic check for Ma > 1 in BC_outlet, so BC_supersonic_outlet can be removed instead of fixed. Is that true or should be keep supersonic_outlet?
- BC_supersonic_inlet seems useful to have, but I'm not sure on the difference between it and BC_farfield for a supersonic case. 
- BC_inlet: is there a strong case for keeping it? Or could it be removed? Is there a case where we want a subsonic inlet?

pcarruscag:
What NEMO has for farfield is a supersonic inlet, it's not handling the outflow case correctly.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
WallyMaier: I think supersonic inlet should remain its own BC.  Particularly for the use of inflow conditions not necessarily at freestream conditions. In my mind, scramjet engine outlet conditions could be modeled as a supersonic inlet.

WallyMaier:
I believe the viscous components of a lot of bcs are commented out and can probably be removed at this point.  Im not even sure if they are formulated properly for NEMO problems.

I think BC_Inlet can be removed.  Or all the commented code can be removed with just an error message.


Ragarding the outlet boundaries, supersonic outlet does seem to be redundant.

jtneedels:
> What NEMO has for farfield is a supersonic inlet, it's not handling the outflow case correctly.

Because it’s not considering incoming characteristics for subsonic flow?

jtneedels:
Good point on scramjets, I’ll start by working on fixing supersonic inlet, then we can make a call on what to do with the normal inlet.

pcarruscag:
Supersonic outflow only has outgoing characteristics, so forcing the farfield state may not be ideal. If the flow is going out it should work as an outflow boundary.

jtneedels:
@pcarruscag  @WallyMaier  Working on the supersonic inlet. I brought things mostly up to date with fluid model. There is a comment saying a root-finding method needs to be implemented to find rhoCvve. But we already have the full conservative state, could we just call cons2primvar?

Any other comments?

WallyMaier:
The root finding method was to find Tve.  I can think of 2/3 solutions:
1.) If we already have the conservative state, then using cons2primvar seems like the best bet in my mind.
2.) Try to just use ComputeTemperatures
3.) Follow the method of initializing node_infty.  (This may actually be 1+2)

jtneedels:
Yeah, so there was a comment left here from someone saying a root-finding method was needed to compute rhocvve in this case. Closer look, it isn't since you assume you already have vib temp and can compute Eve. I used the setTDStatePTTv which seems to give all values needed to initialize. There are likely some bugs there but I can work through that. 

Now I am confused why at line 2238 there is a call to conv_numerics to set values based on node_infty. Should that not be based on node_inlet, or am I missing something?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,211 +2151,141 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+
+ su2double *U_inlet = new su2double[nVar];     su2double *U_domain = new su2double[nVar];
+ su2double *V_inlet = new su2double[nPrimVar]; su2double *V_domain = new su2double[nPrimVar];
+ su2double *Normal = new su2double[nDim];
+
+ /*--- Supersonic inlet flow: there are no outgoing characteristics,
+  so all flow variables can be imposed at the inlet.
+  First, retrieve the specified values for the primitive variables. ---*/
+
+ auto Mass_Frac      = config->GetInlet_MassFrac(Marker_Tag); //TODO: add get inlet mass_frac function
```
WallyMaier, line 2170: can this be const?

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,211 +2151,141 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+       Area += Normal[iDim]*Normal[iDim];
+     Area = sqrt (Area);
+
+     /*--- Set various quantities in the solver class ---*/
+     conv_numerics->SetNormal(Normal);
+     conv_numerics->SetConservative(U_domain, U_inlet);
+     conv_numerics->SetPrimitive(V_domain, V_inlet);
+
+     /*--- Pass supplementary info to CNumerics ---*/ //TODO: why pass node infty?
+     conv_numerics->SetdPdU(nodes->GetdPdU(iPoint), node_infty->GetdPdU(0));
```
WallyMaier, line 2236: This would need to be updated for the specific inlet condition.

I wonder if a temporary variable structure could be initialized in this function or in CNEMOEulerSolver if we have a supersonic inlet.  Then all this math would be done once early, but just a thought.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
jtneedels: Yes, will address.

jtneedels:
Added a node_inlet variable to hold the inlet flow properties and allow us to get dPdU, etc., for implicit solver with nemo convective schemes. We could get away with a single allocation if we put this with the node_infty allocation, but for now will leave in the BC function.

jtneedels:
I think we're going to need to add a separate config option NEMO_SUPERSONIC_INLET since we need mass fracs and Tve, or is there a way to overload the existing option?

pcarruscag:
No, currently there is no way of doing that. The "inlet option" takes a fixed number of arguments.

jtneedels:
> No, currently there is no way of doing that. The "inlet option" takes a fixed number of arguments.
Ok, thanks Pedro. So either I need to create a new supersonic inlet for NEMO, or an additional config option that allows the user to specify the mass fractions and T_ve. Which of these two approaches would be better?

pcarruscag:
I think generalizing the inlet option to read more values would be good

jtneedels:
> I think generalizing the inlet option to read more values would be good

I worked on trying to create a new inlet and it got complicated pretty quickly. The main issue I run into is in COptionInlet in SetValues. To add mass fractions, we need to know the number of species which can vary, so this would need access to config. Just adding an "INLET_GAS_COMPOSITION" field lets us use the same supersonic inlet bc (no additional repeated code) but with just one extra get function to pull from config. It feels a little hacky... but seems to save a lot of additional repeated code.

jtneedels:
Hi @pcarruscag @WallyMaier, it looks like things are working. The big issue I'm having now is computing values for dPdU, etc., for the AUSM Jacobian. Since the inlet solution isn't stored, these values aren't computed in variable. And the functions in fluid model for computing them don't return values, so we would have to allocate in the BC. Are there thoughts on a way to compute these that don't involve all this?

jtneedels:
I tested this implementation on the inviscid wedge regression case. Setting the supresonic inlet to have the same conditions as farfield appears to give the correct solution. As noted in a commit above, there are some non-physical points when using this with AUSM with MUSCL turned on, but turning MUSCL off removes these.

image

jtneedels:
And here is a simple test case demonstrating potential use. A major benefit of the supersonic inlet is being able to model supersonic propulsion systems, such as a scramjet, where the propulsion system exit boundary is modeled as a supersonic inlet, with distinct properties from the farfield flow conditions. Here we have a two dimensional test case demonstrating the interaction of exhaust flow with free-stream flow at the exit plane of a 2D scramjet system. Flow conditions adapted from [A Design Method for Three-Dimensional Scramjet Nozzles with Shape Transition](https://arc-aiaa-org.stanford.idm.oclc.org/doi/abs/10.2514/1.B38293)
Jens Kunze, Michael K. Smart, and Rowan Gollan
Journal of Propulsion and Power 2022 38:1, 3-17

jtneedels:
Hi Everyone, 

I think this PR is pretty much ready to pull in. In review, if someone could double check for potential memory mgmt issues, I didn't see any with valgrind and test cases, but want to verify. Would it be worth adding a test case to the regression suite? If so I'll probably open a separate PR.

Common/include/CConfig.hpp
```diff
@@ -6686,6 +6688,20 @@ class CConfig {
...
+   * \return The inlet velocity vector.
+   */
+  const su2double* GetInlet_MassFrac(void) const { return Inlet_MassFrac; };
+
+    /*!
+   * \brief Get the Tve value for a NEMO supersonic inlet boundary.
+   * \param[in] val_index - Index corresponding to the inlet boundary.
+   * \return The inlet velocity vector.
+   */
+  const su2double GetInlet_Temperature_ve(void) const { return Inlet_Temperature_ve; };
```
pcarruscag, line 6703:
```suggestion
  const su2double* GetInlet_MassFrac() const { return Inlet_MassFrac; }

  /*!
   * \brief Get the Tve value for a NEMO supersonic inlet boundary.
   * \param[in] val_index - Index corresponding to the inlet boundary.
   * \return The inlet velocity vector.
   */
  const su2double GetInlet_Temperature_ve() const { return Inlet_Temperature_ve; }
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,146 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
-//  delete [] Normal;
+ unsigned short iDim;
+ unsigned long iVertex, iPoint, Point_Normal;
+
+ bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
+ bool dynamic_grid  = config->GetGrid_Movement();
+ bool viscous       = config->GetViscous();
+ string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
+
+ su2double *Normal = new su2double[nDim];
```
pcarruscag, line 2162:
```suggestion
 su2double Normal[MAXNDIM] = {0.0};
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,146 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
-//  delete [] V_domain;
-//  delete [] V_inlet;
-//  delete [] Normal;
+ unsigned short iDim;
+ unsigned long iVertex, iPoint, Point_Normal;
+
+ bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
+ bool dynamic_grid  = config->GetGrid_Movement();
+ bool viscous       = config->GetViscous();
+ string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
```
pcarruscag, line 2160: Declare things as needed not at the top, const everything
```suggestion
 const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
 const bool dynamic_grid = config->GetGrid_Movement();
 const bool viscous = config->GetViscous();
 const auto Marker_Tag = config->GetMarker_All_TagBound(val_marker);
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,146 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+
+ /*--- Supersonic inlet flow: there are no outgoing characteristics,
+  so all flow variables can be imposed at the inlet.
+  First, retrieve the specified values for the primitive variables. ---*/
+
+ const su2double* Mass_Frac       = config->GetInlet_MassFrac(); 
+ const su2double Temperature      = config->GetInlet_Temperature(Marker_Tag);
+ const su2double Pressure         = config->GetInlet_Pressure(Marker_Tag);
+ const su2double* Velocity        = config->GetInlet_Velocity(Marker_Tag);
+ su2double Temperature_ve   = config->GetInlet_Temperature_ve();
```
pcarruscag, line 2172: No pretty formatting please.
```suggestion
 const su2double* Mass_Frac = config->GetInlet_MassFrac(); 
 const su2double Temperature = config->GetInlet_Temperature(Marker_Tag);
 const su2double Pressure = config->GetInlet_Pressure(Marker_Tag);
 const su2double* Velocity = config->GetInlet_Velocity(Marker_Tag);
 const su2double Temperature_ve   = config->GetInlet_Temperature_ve();
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,146 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ su2double* Mvec = new su2double[nDim];
+
+ for (iDim = 0; iDim < nDim; iDim++)
+   Mvec[iDim] = Velocity[iDim] / soundspeed;
+
+ /*--- Allocate inlet node to compute gradients for numerics ---*/
+ node_inlet = new CNEMOEulerVariable(Pressure, Mass_Frac, Mvec,
+                                        Temperature, Temperature_ve,
+                                        1, nDim, nVar, nPrimVar, nPrimVarGrad,
+                                        config, FluidModel);
```
pcarruscag, line 2192: This is not very good because it will allocate a bunch of stuff everytime, either persist node_inlet, only allocating it once, or make it a local variables as shown below.
```suggestion
 CNEMOEulerVariable node_inlet(Pressure, Mass_Frac, Mvec,
                                        Temperature, Temperature_ve,
                                        1, nDim, nVar, nPrimVar, nPrimVarGrad,
                                        config, FluidModel);
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,146 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+
+ node_inlet->SetPrimVar(0, FluidModel);
+
+ /*--- Loop over all the vertices on this boundary marker ---*/
+ for(iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+   iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
+
+   /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
+   if (geometry->nodes->GetDomain(iPoint)) {
```
pcarruscag, line 2201:
```suggestion
   if (!geometry->nodes->GetDomain(iPoint)) continue;
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,141 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ for (iDim = 0; iDim < nDim; iDim++)
+   Mvec[iDim] = Velocity[iDim] / soundspeed;
+
+ /*--- Allocate inlet node to compute gradients for numerics ---*/
+ CNEMOEulerVariable node_inlet(Pressure, Mass_Frac, Mvec,
+                                        Temperature, Temperature_ve,
+                                        1, nDim, nVar, nPrimVar, nPrimVarGrad,
+                                        config, FluidModel);
+
+ node_inlet.SetPrimVar(0, FluidModel);
```
WallyMaier, line 2191: does the ".SetPrimVar"  function the same as "->"

jtneedels:
I think I addressed all changes, except for Temperature_ve can't be constant since that value is set to Temperature if the default input is given, so needs to be mutable.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,140 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+   iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
+
+   /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
+   if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+     /*--- Index of the closest interior node ---*/
+     Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
+ 
+     /*--- Normal vector for this vertex (negate for outward convention) ---*/
+     geometry->vertex[val_marker][iVertex]->GetNormal(Normal);
```
pcarruscag, line 2204: Too much indentation now

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,140 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
-//  /*--- Free locally allocated memory ---*/
-//  delete [] U_domain;
-//  delete [] U_inlet;
-//  delete [] V_domain;
-//  delete [] V_inlet;
-//  delete [] Normal;
+ unsigned short iDim;
+ unsigned long iVertex, iPoint, Point_Normal;
+
+ string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
```
pcarruscag, line 2157: Please read my review again

Common/include/CConfig.hpp
```diff
@@ -6686,6 +6688,20 @@ class CConfig {
...
+   * \return The inlet velocity vector.
+   */
+  const su2double* GetInlet_MassFrac() const { return Inlet_MassFrac; }
+
+  /*!
+   * \brief Get the Tve value for a NEMO supersonic inlet boundary.
+   * \param[in] val_index - Index corresponding to the inlet boundary.
+   * \return The inlet velocity vector.
+   */
+  const su2double GetInlet_Temperature_ve() const { return Inlet_Temperature_ve; }
```
pcarruscag, line 6703:
```suggestion
  su2double GetInlet_Temperature_ve() const { return Inlet_Temperature_ve; }
```

SU2_CFD/include/solvers/CNEMOEulerSolver.hpp
```diff
@@ -62,6 +62,7 @@ class CNEMOEulerSolver : public CFVMFlowSolverBase<CNEMOEulerVariable, ENUM_REGI
   CNEMOGas  *FluidModel;          /*!< \brief fluid model used in the solver */
 
   CNEMOEulerVariable* node_infty = nullptr;
+  CNEMOEulerVariable* node_inlet = nullptr;
```
pcarruscag, line 65: Unused ATM
```suggestion
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ /*--- If no Tve value specified (left as 0 K default), set to Ttr value ---*/
+ if (Temperature_ve == 0.0) {Temperature_ve = Temperature;}
+
+ /*--- Set mixture state ---*/
+ FluidModel->SetTDStatePTTv(Pressure, Mass_Frac, Temperature, Temperature_ve);
+
+ /*--- Compute Ma vector for flow direction ---*/
+ const su2double soundspeed = FluidModel->ComputeSoundSpeed();
+
+ su2double* Mvec = new su2double[nDim];
```
pcarruscag, line 2175: MAXNDIM

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+  so all flow variables can be imposed at the inlet.
+  First, retrieve the specified values for the primitive variables. ---*/
 
+ const string Marker_Tag = config->GetMarker_All_TagBound(val_marker);
+
+ const su2double Temperature = config->GetInlet_Temperature(Marker_Tag);
+ const su2double Pressure = config->GetInlet_Pressure(Marker_Tag);
+ const su2double* Velocity = config->GetInlet_Velocity(Marker_Tag);
+ const su2double* Mass_Frac = config->GetInlet_MassFrac(); 
+ su2double Temperature_ve   = config->GetInlet_Temperature_ve();
```
pcarruscag, line 2164:
```suggestion
 su2double Temperature_ve = config->GetInlet_Temperature_ve();
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ /*--- Set mixture state ---*/
+ FluidModel->SetTDStatePTTv(Pressure, Mass_Frac, Temperature, Temperature_ve);
+
+ /*--- Compute Ma vector for flow direction ---*/
+ const su2double soundspeed = FluidModel->ComputeSoundSpeed();
+
+ su2double* Mvec = new su2double[nDim];
+
+ unsigned short iDim;
+ for (iDim = 0; iDim < nDim; iDim++)
```
pcarruscag, line 2178:
```suggestion
 for (unsigned short iDim = 0; iDim < nDim; iDim++)
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
pcarruscag: yes

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ 
+ /*--- Loop over all the vertices on this boundary marker ---*/
+ for(unsigned long iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+   unsigned long iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
+
+   /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
+   if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+   /*--- Index of the closest interior node ---*/
+   unsigned long Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
```
pcarruscag, line 2199:
```suggestion
   const auto Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+                                        1, nDim, nVar, nPrimVar, nPrimVarGrad,
+                                        config, FluidModel);
+
+ node_inlet.SetPrimVar(0, FluidModel);
+ 
+ su2double Normal[MAXNDIM] = {0.0};
+ 
+ /*--- Loop over all the vertices on this boundary marker ---*/
+ for(unsigned long iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
+   unsigned long iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
```
pcarruscag, line 2193:
```suggestion
   const auto iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
```

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,133 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+   conv_numerics->SetPrimitive(nodes->GetPrimitive(iPoint), node_inlet.GetPrimitive(0));
+
+   /*--- Pass supplementary info to CNumerics ---*/
+   conv_numerics->SetdPdU(nodes->GetdPdU(iPoint), node_inlet.GetdPdU(0));
+   conv_numerics->SetdTdU(nodes->GetdTdU(iPoint), node_inlet.GetdTdU(0));
+   conv_numerics->SetdTvedU(nodes->GetdTvedU(iPoint), node_inlet.GetdTvedU(0));
+   conv_numerics->SetEve(nodes->GetEve(iPoint), node_inlet.GetEve(0));
+   conv_numerics->SetCvve(nodes->GetCvve(iPoint), node_inlet.GetCvve(0));
+
+   bool dynamic_grid  = config->GetGrid_Movement();
```
pcarruscag, line 2217: everything that can be const should be const
```suggestion
   const bool dynamic_grid  = config->GetGrid_Movement();
```

Common/include/CConfig.hpp
```diff
@@ -6686,6 +6688,20 @@ class CConfig {
    */
   const su2double* GetInlet_Velocity(string val_index) const;
 
+  /*!
+   * \brief Get the mass fraction vector for a NEMO supersonic inlet boundary.
```
WallyMaier, line 6692:
```suggestion
   * \brief Get the mass fraction vector for a NEMO inlet boundary.
```

Common/include/CConfig.hpp
```diff
@@ -6686,6 +6688,20 @@ class CConfig {
...
+   * \param[in] val_index - Index corresponding to the inlet boundary.
+   * \return The inlet velocity vector.
+   */
+  const su2double* GetInlet_MassFrac() const { return Inlet_MassFrac; }
+
+  /*!
+   * \brief Get the Tve value for a NEMO supersonic inlet boundary.
```
WallyMaier, line 6699:
```suggestion
   * \brief Get the Tve value for a NEMO inlet boundary.
```

WallyMaier:
```suggestion
   * \brief Get the Tve value for a NEMO supersonic inlet boundary.
```

Just in case we use these for subsonic inlets.

Common/src/CConfig.cpp
```diff
@@ -1195,6 +1195,10 @@ void CConfig::SetConfig_Options() {
   addEnumOption("TRANSPORT_COEFF_MODEL", Kind_TransCoeffModel, TransCoeffModel_Map, TRANSCOEFFMODEL::WILKE);
   /* DESCRIPTION: Specify mass fraction of each species */
   addDoubleListOption("GAS_COMPOSITION", nSpecies, Gas_Composition);
+  /* DESCRIPTION: Specify mass fraction of each species for NEMO supersonic inlet*/
+  addDoubleListOption("INLET_GAS_COMPOSITION", nSpecies_inlet, Inlet_MassFrac);
+  /*!\brief INLET_TEMPERATURE_VE \n DESCRIPTION: NEMO supersonic inlet temperature_ve (K), if left 0 K, set to Ttr value \ingroup Config*/
```
WallyMaier, line 1200: Same here, if it gets used for subsonic inlets.

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,129 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+   /*--- Check if the node belongs to the domain (i.e, not a halo node) ---*/
+   if (!geometry->nodes->GetDomain(iPoint)) continue;
+
+   /*--- Index of the closest interior node ---*/
+   const auto Point_Normal = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
+
+   /*--- Normal vector for this vertex (negate for outward convention) ---*/
+   geometry->vertex[val_marker][iVertex]->GetNormal(Normal);
+   for (unsigned short iDim = 0; iDim < nDim; iDim++) Normal[iDim] = -Normal[iDim];
+
```
WallyMaier, line 2203: @pcarruscag if we add the CharacPrimVar, would the below be alright?
`    auto* V_inlet = GetCharacPrimVar(val_marker, iVertex);
     V_inlet=node_inlet.GetPrimitive(0);
`

TestCases/nonequilibrium/invwedge/invwedge_ss_inlet.cfg
```diff
@@ -0,0 +1,79 @@
...
+% File Version 7.5.0 "Blackbird"                                               %
+%                                                                              %
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+% ------------- DIRECT, ADJOINT, AND LINEARIZED PROBLEM DEFINITION ------------%
+%
+SOLVER= NEMO_EULER
+GAS_MODEL= AIR-5
+GAS_COMPOSITION= (0.77, 0.23, 0.0, 0.0, 0.0)
+INLET_GAS_COMPOSITION = (0.77, 0.23, 0.0, 0.0, 0.0)
```
WallyMaier, line 17: Can you move this by the supersonic inlet?

WallyMaier:
This looks good to me. Thanks for the work.
Two last questions:
1.) Would it better to include a viscous case versus an euler?
2.) How to you fix code factor issues?

WallyMaier (approved):

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
pcarruscag: No, that would change the pointer but not the data in CharacPrimVar, turbulence solvers downstream would not be reading the data used by the flow solver, which is the objective of CharacPrimVar

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,129 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+   auto residual = conv_numerics->ComputeResidual(config);
+   LinSysRes.AddBlock(iPoint, residual);
+
+   /*--- Jacobian contribution for implicit integration ---*/
+   bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
+   if (implicit)
+     Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
+
+   /*--- Viscous contribution ---*/
+   bool viscous = config->GetViscous();
```
pcarruscag, line 2230: These can be const too...

SU2_CFD/src/solvers/CNEMOEulerSolver.cpp
```diff
@@ -2151,212 +2151,129 @@ void CNEMOEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container
...
+ su2double Mvec[MAXNDIM] = {0.0};
+
+ for (unsigned short iDim = 0; iDim < nDim; iDim++)
+   Mvec[iDim] = Velocity[iDim] / soundspeed;
+
+ /*--- Allocate inlet node to compute gradients for numerics ---*/
+ CNEMOEulerVariable node_inlet(Pressure, Mass_Frac, Mvec,
+                                        Temperature, Temperature_ve,
+                                        1, nDim, nVar, nPrimVar, nPrimVarGrad,
+                                        config, FluidModel);
```
pcarruscag, line 2184: weird indentation, I recommend copying this function to a file and running clang-format on it, then copy it back.

TestCases/serial_regression.py
```diff
@@ -67,6 +67,16 @@ def main():
...
+    
+    # Inviscid single wedge, implicit, NEMO supersonic inlet
+    invwedge_ss_inlet = TestCase('invwedge_ss_inlet')
+    invwedge_ss_inlet.cfg_dir = "nonequilibrium/invwedge"
+    invwedge_ss_inlet.cfg_file = "invwedge_ss_inlet.cfg"
+    invwedge_ss_inlet.test_iter = 10
+    invwedge_ss_inlet.test_vals = [-1.046196, -1.570959, -18.250158, -18.579516, -18.523248, 2.271892, 1.875990, 5.315899, 0.870132]
+    invwedge_ss_inlet.test_vals_aarch64 = [-1.046196, -1.570959, -18.250158, -18.579516, -18.523248, 2.271892, 1.875990, 5.315899, 0.870132]
+    invwedge_ss_inlet.new_output = True
+    test_list.append(invwedge_ss_inlet)
```
pcarruscag, line 79: No more tests in serial regression please, it's slow
```suggestion
```

pcarruscag (approved):
Nice cleanup :+1:

PR #1773, opened 2022-09-26
Introduction of AUSM+M and AUSM scheme refactoring in NEMO

fmpmorgado:
## Proposed Changes
This PR makes several changes to the NEMO code:
1.)  @fmpmorgado Introduced the AUSM+M convective scheme that avoids the carbuncle phenomena in hypersonic blunt bodies (https://doi.org/10.1016/j.apm.2019.09.005)
2.) Wally - A full refactoring of the AUSM schemes within NEMO
3.) Removal of AUSMPW+ scheme
 
## Related Work
This work is a follow-up and a cleaner version of the work in the branch **feature_NEMO_AUSMPLUSM**. I will delete both branches after merge.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [X] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [X] I have added a test case that demonstrates my contribution, if necessary.
- [X] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

### Benchmark

I have conducted a comparison between the AUSM and AUSM+M for a 2D axisymmetric non-equilibrium flow over a cylinder with a  0.1 meter diameter and a freestream of M=24, P = 10.23 N and T = 224.5 K.

image
The left image represents the converged 2nd order solution using AUSM+M and the right image using the standard AUSM scheme. The new AUSM+M scheme enables to get rid of the pressure oscillations near the stagnation area.

image
Temperature at the stagnation line (solid line is AUSM+M and dashed line is AUSM)

image
image
Pressure at the stagnation line and close-up at the stagnation point (solid line is AUSM+M and dashed line is AUSM)

PR #1885, opened 2023-01-12
Add regressions for all convective numerical schemes for NEMO

WallyMaier:
I am adding regressions for all the convective numerical schemes in NEMO using the implicit solver.  All cases run on an inviscid wedge at the moment, but I am happy to make a new folder in the regression.  

## Related Work
This will allow us to better track our changes and is the start of a more comprehensive set of regression tests for the Nemo solver coming in the future (boundary conditions, etc.) 
This directly relates to #1773 to eventually refactor the AUSM schemes.

## PR Checklist
- [] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.


WallyMaier:
As mentioned, these all use the same geometry, and I happy to discuss inclusion of different test cases.

Also, I noticed that mutation++ does have a regression case, do we not compile M++ for regression builds?

jtneedels (commented):
This looks good to me, definitely important to get test coverage on these schemes! 

Sounds like all these tested cases run, do you have a comparison of convergence history? I'd be curious if they have similar performance on this test case, and would be good to confirm correct behavior past 10 iters.

WallyMaier:
I do not have a good grasp on the performance of each scheme.  I fear that may be a much time-intensive task, particularly since these all are implicit at the moment.

jtneedels:
> I do not have a good grasp on the performance of each scheme. I fear that may be a much time-intensive task, particularly since these all are implicit at the moment.

I'm thinking just run this test case out to convergence for each scheme then comparing convergence histories. That way if any don't converge, we know we're not grandfathering in broken cases/schemes.

pcarruscag (approved):

jtneedels (approved):

WallyMaier:
Here is a quick summary for the inviscid wedge case using different NEMO schemes.  There is clearly work to be done in the convergence/robustness world.
[nemo_scheme_regressions.pdf](https://github.com/su2code/SU2/files/10441554/nemo_scheme_regressions.pdf)

PR #1555, opened 2022-03-04
NEMO:   Native Air-7 gas model implementation for inviscid flows

WallyMaier:
## Proposed Changes
@jtneedels and I have worked on implementing the Air-7 gas model to the NEMO native fluid model.  This allows for simulations of inviscid weakly ionized flows with the addition of NO+ and e-.   

I am open to adding a test case for this, similar to the thermal bath case, since they run extremely fast, though I am open to idea.

Work is ongoing to extend this viscous simulations.
Updates from #1533 are added here as well.

## PR Checklist
- [X] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [X] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1497, opened 2022-01-17
NEMO: protection for viscous flow with ionization

CatarinaGarbacz:
## Proposed Changes
adding basic protection for the case of a nonequilibrium viscous flow with ionization
 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x ] I am submitting my contribution to the develop branch.
- [x ] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [ x] My contribution is commented and consistent with SU2 style.
- [x ] I have added a test case that demonstrates my contribution, if necessary.
- [x ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #1887, opened 2023-01-13
Adding ionization capabilities for viscous simulations (NEMO)

WallyMaier:
## Proposed Changes
This PR adds the capability to simulate the AIR-7 gas model using SU2TCLib for viscous problems. Specific contributions include:
- Fix Gupta-Yos transport models for all flows (including ionized)
- inclusion of Coulomb interaction cross sections for ion-ion collisions

Following the implementation detailed in Scalabrin's thesis.

## Related Work
PRs #1497 #1555 

Update to NEMO documentation includes description of GY model, see: https://github.com/su2code/su2code.github.io/pull/116

## PR Checklist


SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1008,72 +1017,127 @@ CSU2TCLib::CSU2TCLib(const CConfig* config, unsigned short val_nDim, bool viscou
...
-    //TODO: Implement Collision Integral Data for AIR-7 (JN)
     /*--- Collision integral data ---*/
-    //TODO: ELECTRONS HAvE NO OMEGAS....NEED TO ALTER LOOPS? or add dummy vector.....
+    // Index 1: collider
+    // Index 2: partner
+    // Index 3: A1, A2, A3
+
     // Omega(0,0) ----------------------
+    //e-  (TODO: NO+ and e-,  NO+ and NO+, e- and e- ??????)
+    Omega00(0,0,0) = 0.0000000E+00;  Omega00(0,0,1) = 0.0000000E+00;  Omega00(0,0,2) = 0.0000000E+00;  Omega00(0,0,3) = 0.0000000E+00;
```
WallyMaier, line 1027: @jtneedels these values are undefined in Scalabrin's thesis.  Something to look into

jtneedels: I believe these are detailed in Table A.10 where attractive/repulsive Coulomb potential is discussed. Since these are all ion-ion pairs, I think the collision cross section is based on Coulomb potential since all are either attractive/repulsive.

jtneedels:
Reading Scalabrin, he says the WBE transport model breaks down at Ttr > 10,000 K, and he uses the GY model for incorporating ionization in LeMANS. So I think the roadmap to for viscous/ionized flows is:

1. Implement Coulomb potential for ion-ion collisions
2. Implement Gupta-Yos model according to Scalabrin
3. Add some flag to make sure transport model for viscous-ionized is always GY

WallyMaier:
The path forward sounds good.  Gupta-Yos needs to be reworked/fixed :(

jtneedels:
> The path forward sounds good. Gupta-Yos needs to be reworked/fixed :(

Yeah, I'm working on that now. I think one way to do this would be to code in whether to use the Coulomb potential using the omega00 tables for each collision pair. For example, if omega00(ispecies, jspecies, 0) == 1.0, then use Coulomb. We could use 1.0, or some other non-physical value.

jtneedels:
An update: I re-worked this so the computation of the delta_ij values (used in each of the GY functions) is itself a function to remove repeated code. It is now also set up to handle ionized flows. Use of the Coulombic potential will signaled by the omega00 and omega11 values. Forward work is ensuring the correct collision integral curve fit values are used, modifying the omega00 and omega11 tables.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,174 +1927,83 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
+    for (jSpecies = 0; jSpecies < nSpecies; jSpecies++) {
+      if (jSpecies != iSpecies) { 
+
+        const su2double Mj    = MolarMass[jSpecies];
+        const su2double gam_j = rhos[iSpecies] / (Density*Mj);
+
+        const su2double pi = PI_NUMBER;
+        const su2double kb = BOLTZMANN_CONSTANT;
+
+        if (iSpecies == 0) {
```
WallyMaier, line 1949: So this would mean that Gupta-Yos must be used with ionization.
Would there be any benefit of allowing neutral simulations with Gupta-Yos?

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,174 +1927,83 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
+          const su2double d1_ij = ComputeDelta(iSpecies, jSpecies, Mi, Mj, Tve, true);
+          const su2double D_ij = kb*Tve/(Pressure*d1_ij);
+          denom += gam_j/D_ij;
+        } else {
+          const su2double d1_ij = ComputeDelta(iSpecies, jSpecies, Mi, Mj, T, true);
```
WallyMaier, line 1954: Good use of these functions.  Maybe we could be more specific with the name "ComputeCollisionalDeltas"?   Just brain storming.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,174 +1927,83 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
-
-  if (ionization) {
-    SU2_MPI::Error("NEEDS REVISION w/ IONIZATION",CURRENT_FUNCTION);
-  }
+  const su2double Na   = AVOGAD_CONSTANT;
+  const su2double kb   = BOLTZMANN_CONSTANT;
 
   /*--- Mixture vibrational-electronic specific heat ---*/
-  Cvves = ComputeSpeciesCvVibEle(Tve);
+  const auto Cvves = ComputeSpeciesCvVibEle(Tve);
```
WallyMaier, line 2002: 👍

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: I think Gupta-Yos should be computationally efficient, though I don't know which model is more accurate at "low" temperatures

jtneedels: We can make this more general by saying if (ionization && iSpecies == 0), so that if there is not ionization it uses Ttr vs Tve. That would give the option of using GY for neural sims.

SU2_CFD/src/fluid/CSU2TCLib.cpp
jtneedels: Sure, yes that sounds good!

WallyMaier:
I dont believe there is any other places to implement changes for viscous cases.

Common/src/CConfig.cpp
```diff
@@ -6155,14 +6155,14 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
         if (Kind_Regime == ENUM_REGIME::COMPRESSIBLE) cout << "Compressible two-temperature thermochemical non-equilibrium Euler equations." << endl;
         if (Kind_FluidModel == SU2_NONEQ){
           if ((GasModel != "N2") && (GasModel != "AIR-5") && (GasModel != "AIR-7") && (GasModel != "ARGON"))
-          SU2_MPI::Error("The GAS_MODEL given as input is not valid. Choose one of the options: N2, AIR-5, AIR-7, ARGON.", CURRENT_FUNCTION);
+            SU2_MPI::Error("The GAS_MODEL given is unavailble using CSU2TCLIB. Choose one of the options: N2, AIR-5, AIR-7, or ARGON.", CURRENT_FUNCTION);
```
WallyMaier, line 6158:
```suggestion
            SU2_MPI::Error("The GAS_MODEL given is unavailable using CSU2TCLIB. Choose one of the options: N2, AIR-5, AIR-7, or ARGON.", CURRENT_FUNCTION);
```

Common/src/CConfig.cpp
```diff
@@ -6155,14 +6155,14 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
...
         }
         break;
       case MAIN_SOLVER::NEMO_NAVIER_STOKES:
         if (Kind_Regime == ENUM_REGIME::COMPRESSIBLE) cout << "Compressible two-temperature thermochemical non-equilibrium Navier-Stokes equations." << endl;
         if (Kind_FluidModel == SU2_NONEQ){
-          if ((GasModel != "N2") && (GasModel != "AIR-5") && (GasModel != "ARGON"))
-          SU2_MPI::Error("The GAS_MODEL given as input is not valid. Choose one of the options: N2, AIR-5, ARGON.", CURRENT_FUNCTION);
+          if ((GasModel != "N2") && (GasModel != "AIR-5") && (GasModel != "AIR-7") && (GasModel != "ARGON"))
+          SU2_MPI::Error("The GAS_MODEL given is unavailble using CSU2TCLIB. Choose one of the options: N2, AIR-5, AIR-7, or ARGON.", CURRENT_FUNCTION);
```
WallyMaier, line 6165:
```suggestion
          SU2_MPI::Error("The GAS_MODEL given is unavailable using CSU2TCLIB. Choose one of the options: N2, AIR-5, AIR-7, or ARGON.", CURRENT_FUNCTION);
```

SU2_CFD/include/fluid/CSU2TCLib.hpp
```diff
@@ -249,6 +265,18 @@ class CSU2TCLib : public CNEMOGas {
...
 
+  /*!
+   *\brief Compute the collision terms (deltas)
+   *\param[in] iSpecies - Species of gas
+   *\param[in] jSpecies - Collision partner of species i
+   *\param[in] Mi - Molar mass of species i
+   *\param[in] Mj - Molar mass of species j
+   *\param[in] T - Temperature of gas
+   *\param[in] d1 - Whether delta_1 or delta_2 is computed
+   *\param[out] Delta_1 or Delta_2
```
WallyMaier, line 276:
```suggestion
   *\param[out] Delta_1 or Delta_2 - Collision term between species i and j
```

jtneedels:
Update: model runs with air-5 and air-7 but diverges in ~10 iterations. Need to check implementation, look for bugs.

jtneedels:
For any folks following this saga, one of the issues was thermal conductivity. I think I figured it out and it’s a weird one… In short: it was the thermal conductivities, but not for the reason we think. They are divided by a denom term that depends on a value a that indeed goes to nan if any free-stream concentrations are zero. However, the reason it doesn’t diverge immediately is if that denom term is less than eps, the thermal conductivity is set to zero, so thermal conductivity was always zero which is why it blew up after a few iters rather than the first one.
Fix: FS concentrations > 0, but also have to remove the if-statement that sets k_tr to zero. The denom term is order 1E-21 and always less than eps, so with that in the code it is always 0.

Diffusion coefficient is now the only parameter giving wrong values, so need to check that next.

jtneedels:
Update: Wally found an index issue with the Ds calculation. Now model works for neutral and ionized flows, as long as free-stream concentrations are non-zero. Forward work is to fix it so that free-stream concentrations can be zero and not cause a problem.

jtneedels:
Hi @pcarruscag I think we're at a place in this PR where we can ask for a code review. In particular, I'm wondering if there is a way to reduce the code complexity in the ComputeCollisionDelta() function. Moving this to its own function reduced repeated code (which was called for each GY transport model) but now I've got some cumbersome logic statements and I'm not sure how to simplify.

jtneedels:
Both Gupta-Yos and WBE are now capable of running with ionization for the viscous simulations using explicit time integration. Viscous numerics needs to be modified to permit implicit. Initial testing at Mach 10 and Mach 15 by @WallyMaier show promising results for GY.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1817,10 +1877,58 @@ void CSU2TCLib::ThermalConductivitiesWBE(){
...
+  if (coulomb) {
+    
+    const su2double e_cgs = FUND_ELEC_CHARGE_CGS; // CGS unit of fundamental electric charge 
+    const su2double kb_cgs = BOLTZMANN_CONSTANT * 1E7; // CGS unit of Boltzmann Constant 
+    const su2double ne_cgs = Na * rhos[0] / MolarMass[0] * 1E-6; // CGS unit of electron number density
+        
+    const su2double debyeLength = sqrt(kb_cgs * T / 4 / pi / ne_cgs / pow(e_cgs,2));
+    const su2double T_star = debyeLength / (pow(e_cgs,2) / (kb_cgs * T));
+
+    if (Omega11(iSpecies, jSpecies, 0) == 1.0 && d1) {
```
WallyMaier, line 1894: Can we add some comments to this block just for clarity?

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:221224-1158
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c feature_NEMO_visc_ion -s ${{matrix.testscript}}
```
WallyMaier, line 131: Just a note to not forget this.

pcarruscag (approved):
LGTM code-wise

jtneedels:
Status update: Wally ran a Mach 20 cylinder test case with AIR-7 model using SU2_NONEQ and results looked good. I need to check everything works with MPP air-7, and regressions need to be updated. Once confirmed/updated I think we can merge.

SU2_CFD/src/fluid/CSU2TCLib.cpp
jtneedels: Sure, but I think this is best explained by the PR to documentation which gives the actual equations.

WallyMaier: I think something like  'Compute collision term'  and something to help distinguish the loops a little bit.  Nothing crazy.

.github/workflows/regression.yml
jtneedels: @WallyMaier Anything we still need to do here?

SU2_CFD/include/fluid/CSU2TCLib.hpp
```diff
@@ -71,14 +71,30 @@ class CSU2TCLib : public CNEMOGas {
 
   const su2double T_ref_suth = 273.15; /*!<\brief Reference temperature for Sutherland's model [K] */
 
+  /*--- Attractive Coulombic potential constants ---*/
```
jtneedels, line 74: Should I add a comment with a reference for these values, or okay just to keep track of that in documentation?

jtneedels (approved):

.github/workflows/regression.yml
WallyMaier: Not until we finally push

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,227 +1946,134 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
-      d2_ij = 16.0/5.0 * sqrt((2.0*Mi*Mj) / (pi*Ru*T*(Mi+Mj))) * Omega_ij;
-      denom_t += a_ij*gam_j*d2_ij;
       denom_r += gam_j*d1_ij;
+      denom_re += gam_j*d2_ij;
     }
 
-    /*--- Translational contribution to thermal conductivity ---*/
-    ThermalCond_tr += (denom_t > EPS) ? ((15.0/4.0)*kb*gam_i/denom_t) : su2double(0.0);
+    /*--- Prevent divide by 0 ---*/
+    if (denom_t == 0.0) denom_t = EPS;
```
WallyMaier, line 2063: @pcarruscag @jtneedels This helps prevent the solution going NaN and seems to have little impact on solution quality/convergence behavior.   Is there a better/safer way to do this?

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,227 +1946,134 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
-          * pow(T, Omega00(iSpecies,jSpecies,0)*log(T)*log(T)
-          + Omega00(iSpecies,jSpecies,1)*log(T)
-          + Omega00(iSpecies,jSpecies,2));
+      const su2double T_col = ((iSpecies == 0 && ionization) || (jSpecies == 0 && ionization)) ? Tve : T; 
 
-      /*--- Calculate "delta1_ij" ---*/
-      d1_ij = 8.0/3.0 * sqrt((2.0*Mi*Mj) / (pi*Ru*T*(Mi+Mj))) * Omega_ij;
+      su2double d1_ij = ComputeCollisionDelta(iSpecies, jSpecies, Mi, Mj, T_col, true);
+      su2double d2_ij = ComputeCollisionDelta(iSpecies, jSpecies, Mi, Mj, T_col, false);
+      if (d1_ij >1E16) d1_ij = 1E16;
```
WallyMaier, line 2052: Is there a better way to check for Inf?   In some cases d1 or d2 can be infinity but this causes NaNs.

SU2_CFD/include/fluid/CSU2TCLib.hpp
WallyMaier: I think that would be helpful!

jtneedels (approved):

SU2_CFD/src/fluid/CSU2TCLib.cpp
jtneedels: LGTM - we can't say if denom < EPS because these values are so small (that was causing issues with the original implementation) so I think this is good.

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,227 +1946,134 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
-                                   + Omega00(iSpecies,jSpecies,1)*log(T)
-                                   + Omega00(iSpecies,jSpecies,2));
-        /*--- Calculate "delta1_ij" ---*/
-        su2double d1_ij = 8.0/3.0 * sqrt((2.0*Mi*Mj) / (pi*Ru*T*(Mi+Mj))) * Omega_ij;
+        const su2double T_col = (iSpecies == 0 && ionization) ? Tve : T; 
 
-        /*--- Calculate heavy-particle binary diffusion coefficient ---*/
-        su2double D_ij = kb*T/(Pressure*d1_ij);
+        su2double d1_ij = ComputeCollisionDelta(iSpecies, jSpecies, Mi, Mj, T_col, true);
+        if (d1_ij > 1E16) d1_ij = 1E16;
```
pcarruscag, line 1970:
```suggestion
        su2double d1_ij = fmin(ComputeCollisionDelta(iSpecies, jSpecies, Mi, Mj, T_col, true), 1E16);
```

SU2_CFD/src/fluid/CSU2TCLib.cpp
```diff
@@ -1829,227 +1946,134 @@ void CSU2TCLib::DiffusionCoeffGY(){
...
+      const su2double T_col = (iSpecies == 0 && ionization) ? Tve : T; 
 
-      /*--- Calculate "delta" quantities ---*/
-      su2double Omega_ij = 1E-20 * Omega11(iSpecies,jSpecies,3)
-          * pow(T, Omega11(iSpecies,jSpecies,0)*log(T)*log(T)
-          + Omega11(iSpecies,jSpecies,1)*log(T)
-          + Omega11(iSpecies,jSpecies,2));
-      su2double d2_ij = 16.0/5.0 * sqrt((2.0*Mi*Mj) / (pi*Ru*T*(Mi+Mj))) * Omega_ij;
+      su2double d2_ij = ComputeCollisionDelta(iSpecies, jSpecies, Mi, Mj, T_col, false);
+      if (d2_ij > 1E16) d2_ij = 1E16;
```
pcarruscag, line 2002: Repeated, why don't you put the limit inside the function?
Or define a constant for the limit, otherwise this repetition is bound to cause problems at some point

SU2_CFD/src/fluid/CSU2TCLib.cpp
pcarruscag: Is denom always positive? If so consider using fmax(denom, EPS) comparing floats with == can cause surprises.

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: I moved it into ComputeCollisionDelta function

SU2_CFD/src/fluid/CSU2TCLib.cpp
WallyMaier: Unfortunately, adding in the fmax causes issues down the line in the simulation.  I added a check for denom<= 0 to avoid any extra weirdness.

PR #1892, opened 2023-01-16
Heat solver using scalar transport framework - Part 2

pcarruscag:
As discussed in the last dev meeting, this deprecates the centered scheme for weakly coupled heat transfer, because it had a bug in the speed of sound, and because this made it much easier to adapt the rest of the heat solver to the scalar framework.

Bonus:
- Add SLOPE_LIMITER_HEAT option.
- OpenMP of the heat solver.
- Correct issues with Inlet and Outlet boundaries and gain farfield (not implemented before) for free.
- Regression for weakly coupled heat equation (primal was not covered before).
- PyCustom markers for heat solver.

ToDo:
- Bounded scalar mode.
- Docs.
- Fix sensitivity wrt inlet velocity.


WallyMaier (approved):
This looks good to me.

PR #1897, opened 2023-01-20
Wrap MPI_Allgatherv for NdFlattener

### LINKED ISSUE #1893: Undefined behavior in NDFlatner causing spurious values of wall roughness
**Describe the bug**
I faced some non determinism in #1892 which I traced back to spurious values of walll roughness.
I tried running UB sanitizer and valgrind, the latter finds one memory error when accessing the flatner, but theres nothing while it's built.
The ndflatner unit test also fails on my machine, both in release and debug builds.

Given the complexity of that class, and it only being used in one place, I just wrote alternative code.
@maxaehle can you investigate the issue?

 - ubuntu 22 + gcc + mpich (standard packages)

maxaehle:
## Proposed Changes
Let NdFlattener use a wrapped [MPI_Allgatherv](https://www.mpich.org/static/docs/v3.2/www3/MPI_Allgatherv.html) function which checks the number of participating MPI ranks. If there is only one, perform a simple copy operation instead.

I hope that this fixes #1893, which I believe to be the result of a bug in MPICH. In my Ubuntu MPICH container, the `[NdFlattener]` test succeeds with these changes. 

## Related Work
#1893 

## PR Checklist

### DIFF
--- a/Common/include/toolboxes/ndflattener.hpp
+++ b/Common/include/toolboxes/ndflattener.hpp
@@ -210,12 +210,12 @@ class NdFlattener;
 /*! Allgatherv wrapper defaulting to a copy operation if there is only a single MPI rank.
  *
  *  Introducing this was necessary because MPICH's Allgatherv behaved unexpectedly if there
- *  is only one MPI rank.
+ *  is only one MPI rank (seemingly ignoring displs[0] != 0).
  */
 static inline void SU2_MPI_Allgatherv_safe(const void* sendbuf, int sendcount, SU2_MPI::Datatype sendtype, void* recvbuf,
                               const int* recvcounts, const int* displs, SU2_MPI::Datatype recvtype, SU2_MPI::Comm comm) {
-  if(SU2_MPI::GetSize()==1){
-    SU2_MPI::CopyData(sendbuf,recvbuf,sendcount,sendtype,displs[0]);
+  if (SU2_MPI::GetSize() == 1) {
+    SU2_MPI::CopyData(sendbuf, recvbuf, sendcount, sendtype, displs[0]);
   } else {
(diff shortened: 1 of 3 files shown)


Common/include/toolboxes/ndflattener.hpp
```diff
@@ -207,6 +207,20 @@
...
 
+/*! Allgatherv wrapper defaulting to a copy operation if there is only a single MPI rank.
+ *
+ *  Introducing this was necessary because MPICH's Allgatherv behaved unexpectedly if there
+ *  is only one MPI rank.
+ */
+static inline void SU2_MPI_Allgatherv_safe(const void* sendbuf, int sendcount, SU2_MPI::Datatype sendtype, void* recvbuf,
+                              const int* recvcounts, const int* displs, SU2_MPI::Datatype recvtype, SU2_MPI::Comm comm) {
+  if(SU2_MPI::GetSize()==1){
+    SU2_MPI::CopyData(sendbuf,recvbuf,sendcount,sendtype,displs[0]);
```
pcarruscag, line 218:
```suggestion
 *  is only one MPI rank (seemingly ignoring displs[0] != 0).
 */
static inline void SU2_MPI_Allgatherv_safe(const void* sendbuf, int sendcount, SU2_MPI::Datatype sendtype, void* recvbuf,
                              const int* recvcounts, const int* displs, SU2_MPI::Datatype recvtype, SU2_MPI::Comm comm) {
  if (SU2_MPI::GetSize() == 1) {
    SU2_MPI::CopyData(sendbuf, recvbuf, sendcount, sendtype, displs[0]);
```

pcarruscag (approved):
Thanks

PR #1900, opened 2023-01-26
Remove git extension in coolprop download link

davidscn:
## Proposed Changes

While trying to install SU2 7.5.0 on ubuntu, the CoolProp auto download was failing (although CoolProp was disabled in the meson options, why is the download triggered at all?). The reason, however, was the `.git` extension in the download link. Removing the link fixed the problem for me.

## PR Checklist


pcarruscag (approved):
LGTM Thanks

PR #1751, opened 2022-09-05
Langtry and Menter transition model

sun5k:
Hi, all.

This PR is resuming the #1592 I'm not familiar with the git, so the branch is unfortunately merged.

"[WIP] Develop for Langtry and Menter transition model. https://github.com/su2code/SU2/pull/1592" was merged while trying to update my remote repository to v7.40. I'm not good at using GitHub, so I'm sorry for this situation.

My remote repository(forked) is not synchronized with v7.40, and I don't know what the problem is. So, I will delete the forked repository and fork it again to PR with the same name and content. Fortunately, the LM model updated to v7.40 obtained the same residuals and results as the previous v7.3.1.

## Current State 

Code is well running and, validation is almost done.

V&V cases: T3A, T3A-, NLF

Flow conditions are the reference from:  https://doi.org/10.2514/6.2022-3679.

The grids of T3A and NLF cases are provided by [TMW](https://transitionmodeling.larc.nasa.gov/workshop_i/)(Transition Model Workshop). And I made the grid of T3A-  myself.

To validate the LM model, the simulation results of SU2 are compared with the results of  Fluent19.0 with a similar numerical setting.

Here is the numerical scheme : 

<html xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">

</head>

<body>

  | Fluent | SU2
-- | -- | --
Flux | Roe-FDS | L2ROE
Gradient | Least   Squares Cell Based | WEIGHTED_LEAST_SQUARES
Spatial   Discretization Flow | Third-order   MUSCL | MUSCL_FLOW
Spatial   Discretization Turbulence | First-order   Upwind | MUSCL_NO

</body>

</html>
The CFL number is set constant value.

Some high-resolution grid-level results are missed because I don't have enough computational resources.
Here is Result

T3A : 

image
image
image
image

T3A-: 
The gird test result is not converged, but It's enough to compare the model.

image
image
image
image

NLF : 
AoA 0 case is shown. In this case, the flow separation occurred on the upper surface. So, not fully converged, and I show only Fine level grid results according to the number of iterations. 
image
Zoom to the separation region : 
image

## PR Checklist
[...]

PR #1901, opened 2023-01-27
Implementation of Simplified LM Transition model [WIP]

rois1995:
## Proposed Changes
Hi everyone! I am starting to implement the Simplified version of the Langtry-Menter transition model following the work in https://doi.org/10.2514/6.2012-672.  More updates will follow.

## Related Work
This follows the implementation of the LM model as #1751.


rois1995:
Technically, this model has only one equation, the one for intermittency. Most of the functions are the same as the original LM model. Should I consider the Simplified model (SLM) as an option for the LM model to avoid duplicates? This is how I started, but I am open to discussions and suggestions.

pcarruscag (commented):
If the changes to the solver don't become too intrusive this approach sounds good to me

rois1995:
I have a question: for each point in the mesh I am trying to compute the dot product between the velocity vector and the normal to the wall of the nearest point on the wall. How do I access such information? I've found that in the CPoint class I have the ClosestWall_Elem variable which stores the index of the closest element on a wall. However, when I try to assess the information with a number of cores greater than 2, it crashes. Moreover, to recover the normal of the element I perform a mean of the normals on the nodes of that element. Is there a structure that has the normals saved for each element of the primal grid? The part that I am referring to is from line 208 in CTransLMSolver.cpp .

pcarruscag:
See what is done at the bottom of CGeometry.cpp in CGeometry::ComputeWallDistance.
You need to communicate the normals you need first.

Common/include/geometry/dual_grid/CPoint.hpp
```diff
@@ -456,6 +456,18 @@ class CPoint {
   }
   inline void SetWall_Distance(unsigned long iPoint, su2double distance) { Wall_Distance(iPoint) = distance; }
 
+  /*!
+   * \brief Get the index of the closest wall element.
+   * \param[in] iPoint - Index of the point.
+   */
+  inline unsigned long GetClosestWall_Elem(unsigned long iPoint) {return ClosestWall_Elem(iPoint);}
```
WallyMaier, line 463: Can you add the param[out] for these please

Common/include/option_structure.hpp
```diff
@@ -1265,10 +1291,25 @@ inline LM_ParsedOptions ParseLMOptions(const LM_OPTIONS *LM_Options, unsigned sh
...
+  if (IsPresent(LM_OPTIONS::MOD_EPPLER_SLM)) {
+    LMParsedOptions.Correlation_SLM = TURB_TRANS_CORRELATION_SLM::MOD_EPPLER_SLM;
+    NFoundCorrelations_SLM++;
+  }
 
   if (NFoundCorrelations > 1) {
     SU2_MPI::Error("Two correlations selected for LM_OPTIONS. Please choose only one.", CURRENT_FUNCTION);
   }
+  if (NFoundCorrelations_SLM > 1) {
+    SU2_MPI::Error("Two correlations selected for Simplified model into LM_OPTIONS. Please choose only one.", CURRENT_FUNCTION);
```
WallyMaier, line 1311:
```suggestion
    SU2_MPI::Error("Two correlations selected for simplified LM_OPTIONS. Please choose only one.", CURRENT_FUNCTION);
```

Common/src/CConfig.cpp
```diff
@@ -3471,9 +3471,9 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
   if (Kind_Trans_Model == TURB_TRANS_MODEL::LM) {
     lmParsedOptions = ParseLMOptions(LM_Options, nLM_Options, rank, Kind_Turb_Model);
 
-    /*--- Check if problem is 2D and LM2015 has been selected ---*/
-    if (lmParsedOptions.LM2015 && val_nDim == 2) {
-      SU2_MPI::Error("LM2015 is available only for 3D problems", CURRENT_FUNCTION);
```
WallyMaier, line 3476: is LM2015 gone?  or is crossflow the same?

SU2_CFD/include/numerics/turbulent/transition/trans_correlations.hpp
```diff
@@ -189,4 +190,107 @@ class TransLMCorrelations {
...
+          const su2double SecondTerm = -123.66 * pow(lambda_theta_local, 2.0);
+          const su2double ThirdTerm = -405.689 * pow(lambda_theta_local, 3.0);
+          FPG = 1 - (FirstTerm + SecondTerm + ThirdTerm) * exp(-pow(Tu_L/1.5,1.5));
+        } else {
+          FPG = 1 + 0.275 * (1 - exp(-35.0 * lambda_theta_local)) * exp(-Tu_L/0.5);
+        }
+
+        // This is not reported in the paper
+        //FPG = max(FPG, 0.0);
+
```
WallyMaier, line 268:
```suggestion
```

Common/src/CConfig.cpp
rois1995: I have changed the option to CROSSFLOW, since I will use it also for the Simplified model.

SU2_CFD/include/numerics/turbulent/transition/trans_correlations.hpp
rois1995: I have left it there since I do not know if I have to keep it or not.

Common/include/option_structure.hpp
rois1995: I'd stick with my change since the options are for the simplified model. They are not simplified options.

Common/include/geometry/dual_grid/CPoint.hpp
rois1995: Sure, I'll do it now.

bigfooted:
> I have a question: for each point in the mesh I am trying to compute the dot product between the velocity vector and the normal to the wall of the nearest point on the wall. How do I access such information? I've found that in the CPoint class I have the ClosestWall_Elem variable which stores the index of the closest element on a wall. However, when I try to assess the information with a number of cores greater than 2, it crashes. Moreover, to recover the normal of the element I perform a mean of the normals on the nodes of that element. Is there a structure that has the normals saved for each element of the primal grid? The part that I am referring to is from line 208 in CTransLMSolver.cpp .

There is the geometry toolbox for dot product and normal:
https://github.com/su2code/SU2/blob/master/Common/include/toolboxes/geometry_toolbox.hpp
just look in the code for examples, in vscode you can search for GeometryToolbox::DotProduct or GeometryToolbox::Norm

rois1995:
> > I have a question: for each point in the mesh I am trying to compute the dot product between the velocity vector and the normal to the wall of the nearest point on the wall. How do I access such information? I've found that in the CPoint class I have the ClosestWall_Elem variable which stores the index of the closest element on a wall. However, when I try to assess the information with a number of cores greater than 2, it crashes. Moreover, to recover the normal of the element I perform a mean of the normals on the nodes of that element. Is there a structure that has the normals saved for each element of the primal grid? The part that I am referring to is from line 208 in CTransLMSolver.cpp .
> 
> There is the geometry toolbox for dot product and normal: https://github.com/su2code/SU2/blob/master/Common/include/toolboxes/geometry_toolbox.hpp just look in the code for examples, in vscode you can search for GeometryToolbox::DotProduct or GeometryToolbox::Norm

The problem is more related to the finding of the wall-normal for a point within the volume mesh, not to the computations that it will be involved in.

pcarruscag:
The solution I suggested didn’t work?

rois1995:
> The solution I suggested didn’t work?

I still have to check if the implementation is correct but with more than two cores the code breaks. However, I found out that the wall-normal of a volume point can be computed as the normalized gradient of the wall-distance. Does this sound correct to you? 

However, there is a problem: I am using the aux variables to compute these gradients, but to compute dot(n, U) I first need n, thus I cannot compute them simultaneously. Since these computations are performed in the Preprocessing of the solvers, I was thinking to compute the normal within the FLOW_SOL preprocessing and the dot(n, U) in the TRANS_SOL preprocessing since the flow solver comes before the trans solver. Is this right?

pcarruscag:
It's not correct. You need to follow the pattern from CGeometry::ComputeWallDistance
First you collect all the normals that are local to the rank
Then you collect the normals across all ranks (using the NDFlatner), then you access this information by (ClosestRank, ClosestMarker, ClosestVertex)
You cannot access the local information directly because ClosestMarker and ClosestVertex may refer to a different rank.

pcarruscag:
Do what we do with wall roughness and do it in the same place (computewalldistance and store in CPoint)

rois1995:
Ok, I think I managed to add it to the ComputeWallDistance function. I tested it on a 2D cylinder and it works just fine. However, when I use it on a profile with a sharp trailing edge it gives me strange results on the region behind the TE.

Cylinder:
Normal_x
Normal_y

Profile:
Normal_x
Normal_y

I guess it is intrinsic within the computation of the wall distances.

pcarruscag:
My guess is that the closest element for those regions is somewhere on the trailing edge and the normals there change very quickly.

rois1995:
Yeah, I do not think that it is solvable. Plus, I do not think that it will affect the solution either.

pcarruscag:
Any updates @rois1995? Is this almost ready?

bigfooted:
@rois1995 I'd just merge it with develop and have a look at the conflicts in vscode. It's probably just some lines that have moved around.

rois1995:
The problem is that I am behind more than 1100 commits. I have already created another branch starting from the current develop and I am running the test cases from the paper in the description. I may be able to merge the new branch into this one and then I'll be on par with the develop, right?

rois1995:
I have been working on the validation of the implemented models. It has been quite difficult for the flatplate cases since results change a bit between all the reference papers, either in cases of mesh generation or in presentation of the same results. The situation is better when the 2D Eppler E387 airfoil case is considered. In the end, I consider it validated if the results are in line with the more complex $\gamma-Re_\theta$ model. 

## Notation
SLM stands for Simplified Langtry-Menter model
CODER refers to the first set of correlation functions in Ref. [Coder and Maughmer](https://doi.org/10.2514/6.2012-672) (Eqs 11-14)
MODEPPLER refers to the second set of correlation functions in Ref. [Coder and Maughmer](https://doi.org/10.2514/6.2012-672) (Eqs 15-16)
MENTER refers to the set of correlation functions in Ref. [Menter et al.](https://doi.org/10.1007/s10494-015-9622-4) (Eqs 15-16)

## ERCOFTAC Flatplates
The flow conditions are visible here (taken from [Menter et al., 2015](https://doi.org/10.1007/s10494-015-9622-4), however there are different values in literature, as well as results for the same model ex. [Menter et al., 2022](https://doi.org/10.1007/s10494-022-00336-8)):

FlowConditions

### Mesh as described in [Menter et al., 2022](https://doi.org/10.1007/s10494-022-00336-8)
Menter_SFC_Meters

### Very Fine mesh (much finer than Menter's one)
VeryFine_SFC_Meters

### Other Results
OtherResults_1
OtherResults_Algebraic

It is interesting to see how the Mod-Eppler correlations predict fully-laminar flow for every flatplate case (and some hint of transition for the T3B case which has the highest $TI_\infty$). On the other hand, Coder and Menter correlations predict transition at reasonable locations. From the other results, $\gamma$ model refers to the MENTER correlation. My suspect is that no grid convergence has been performed in the chordwise direction in the other papers.

The Eppler simulations are still on going and will be shown in the next comment.

rois1995:
All of the AD regression tests are failing, do you have any idea on why is it happening? This is common to other PRs too.

rois1995:
# Results for the Eppler E367

## Aerodynamic coefficients
Coefficients

Here you can see a comparison on the aerodynamic coefficients between the SU2 implementation and the one coming from reference papers. Results are a little bit different. In particular, for the Menter implementation the mesh topology is different as they have simulated a wind-tunnel like domain whereas mine is in free air (O-grid). I will try to compare the SU2 implementation with their meshes in the next days. 

## $c_P$ and $c_F$ distributions
### $\alpha = 5^\circ$
CPPlot_5

Here, the main comparison is between the Coder implementation and the reference paper. The laminar bubble is slightly smaller in the reference paper but the shape of the pressure coefficient curve is very similar. 

### $\alpha = 6^\circ$
CPPlot_6
CFPlot_6

Here, the main comparison is between the Menter implementation and the reference paper. As said before, the mesh topologies are completely different. However, the same differences in the skin friction coefficient between the SLM and the LM model can be seen in both the reference and the SU2 implementation. On the other hand, the $c_P$ distribution is very different, with the SLM reference implementation predicting a much smoother effect coming from the LSB. The same smoothing is visible in the SU2 implementation, although it is much weaker. The peak of the $c_P$ at the leading edge is not visible in the reference paper as the image is cropped.

rois1995:
Every testcase that is failing is due to a bug that I fixed in the CIncFlowOutput where the value of MAX_RMS _W was  used for the RMS_W output. I should fix in the tutorials.

bigfooted:
> Every testcase that is failing is due to a bug that I fixed in the CFlowIncOutput where the value of MAX_RMS_W was used for the RMS_W output. I should fix in the tutorials. Should I open another PR for that?

The overhead in this PR is not so large, so I'd say keep it in this PR.

rois1995:
> > Every testcase that is failing is due to a bug that I fixed in the CFlowIncOutput where the value of MAX_RMS_W was used for the RMS_W output. I should fix in the tutorials. Should I open another PR for that?
> 
> 
> 
> The overhead in this PR is not so large, so I'd say keep it in this PR. 

Guess it's too late now! Sorry 😐

pcarruscag:
You did well to make a small PR for it.

pcarruscag:
If you are happy with the validation does that mean this PR is ready for review? Can you remove [WIP] if that is the case?

rois1995:
I think that the Modified Eppler correlations can be removed as they do not perform very well. We can leave there just the Coder and the Menter correlations.

rois1995:
I guess I am still missing the cross-flow validation, I will do it in the next days.

rois1995:
Hi @pcarruscag, yes, it is ready. I actually performed the 3D simulations on the prolate ellipsoid case but I did not share those here. 

Mesh_Fine_AoA_5
Mesh_Fine_AoA_10
Mesh_Fine_AoA_15

Comparing simulations with the reference paper (https://doi.org/10.2514/6.2017-3159) shows really poor results. In the paper, only the Menter correlations have been investigated. The cross-flow effects work in promoting transition to turbulence only for the Modified Eppler correlations, whereas the Coder one shows no influence (the onset function related to the cross-flow transition is always smaller than the standard one).

rois1995:
That is so true! There is some variability, especially with the turbulence intensity at freestream and the turbulent-to-laminar viscosity ratio at freestream (although it is lower for the latter).

bigfooted:
@rois1995 put a student on this one to validate further? Else this work might disappear into obscurity again.

rois1995:
Sorry for the long pause on this PR. I have now brought it up to date and reviewed it against the literature:

- Merged `develop`. The one-equation model now uses the new edge-flux numerics of `develop` (`CScalarFlux_TransLM` with one variable), which also makes the convective flux of γ conservative (it was missing the density).
- Checked the implementation against Menter et al. (2015), Vallinayagam Pillai & Lardeau (2017) and, for SA, Lee & Baeder (2021). The SA coupling previously reused the constants of the two-equation SA-LM coupling (Medida & Baeder); it now follows Lee & Baeder, who coupled exactly this one-equation model to SA.
- Combinations not found in the literature (e.g. `CODER_SLM`, `MOD_EPPLER_SLM` with the one-equation model) now stop with a config error.
- Fixed some bugs, also in the existing LM model: restart consistency, the free-stream Re_θt limit, missing AD inputs, and a crash with the discrete adjoint (now a clear error).
- Added regression tests on the 6:1 prolate spheroid (TestCases#207).

I updated the description with the details. Validation runs are next.

PR #1938, opened 2023-02-26
Access solution and primitives via python wrapper

pcarruscag:
## Proposed Changes
@patelha57 @aa-g @Nicola-Fonzi @bigfooted 
I'm proposing this as the way to interact with volume fields via the python wrapper.
We create a matrix view (does not copy data) into the coordinates, solution, primitives, etc. which then allows point-wise or row-wise access to the data (read or write). We can extend it to also expose derivative information.
For example:
```python
coords = driver.Coordinates()
print(coords(iPoint, 1)) # read a value
coords.Set(iPoint, (x, y, z)) # write a row
coords.Set(iPoint, 0, coords.Get(iPoint, 0)) # set a value and different way of reading a value

solversIndices = driver.GetSolverIndices() # maps solver names to our solver integers, similar to what we do with markers
solution = driver.Solution(solverIndices["C.FLOW"]) # same object type as coords

primitiveIndices = driver.GetPrimitiveIndices() # maps primitive names to their indices.
temperatureIndex = primitiveIndices["TEMPERATURE"]
primitives = driver.Primitives()
print(primitives(iPoint, temperatureIndex))
```

## PR Checklist


bigfooted:
Oh, that's great! I like this way to get the solver index. This would also clean up the flamelet PR, that has a hacky pywrapper implementation for creating an initial solution for the flamelet solver.

SU2_CFD/include/variables/CPrimitiveIndices.hpp
```diff
@@ -89,4 +96,37 @@ struct CPrimitiveIndices {
...
+  nameToIndex["VELOCITY_X"] = idx.Velocity();
+  nameToIndex["VELOCITY_Y"] = idx.Velocity() + 1;
+  if (idx.NDim() == 3) {
+    nameToIndex["VELOCITY_Z"] = idx.Velocity() + 2;
+  }
+  for (IndexType iSpecies = 0; iSpecies < idx.NSpecies(); ++iSpecies) {
+    nameToIndex["DENSITY_" + std::to_string(iSpecies)] = idx.SpeciesDensities() + iSpecies;
+  }
+  return nameToIndex;
+}
```
bigfooted, line 132:
```suggestion
}

```

bigfooted (approved):
LGTM, if no more changes are required for this PR, let's merge it.

bigfooted:
maybe @patelha57 @aa-g have additional comments? Else let's merge, LGTM.

pcarruscag:
I talked with @Nicola-Fonzi and I'll try to use the same "matrix view" strategy for data at markers. But if the current solution unblocks the flamelet work go ahead and merge this.

bigfooted:
Ok, thanks. I will merge this now, then clean up the flamelet branch, so it will be ready for review next week.

PR #1806, opened 2022-11-07
CoolProp fluid model

PENGYAN777:
## Proposed Changes
This project aims to couple the free and open-source thermodynamics library CoolProp with SU2.

There are **two advantages:**

1. More accurate numerical results. 
CoolProp can provide accurate thermodynamics properties (P, T, rho, ect) of commone fluids by either extracting experiment data from published papers, or computing based on the state-of-the-art Helmholtz energy formulations. 
So, thermodynamics properties obtained from CoolProp should be more accurate than those obtained from Van der Waals or peng robinson gas model embedded in SU2. Also, the numerical results should be more accurate.
2. More user-friendly
To use CoolProp feature, the ony input from the user is the name of fluid. Then all other thermodynamics can be provided by CoolProp.
%FLUID_MODEL = COOLPROP
%FLUID_NAME = nitrogen
 
## Related Work
This project references the following work:
1.the work of coupling 'Mutationpp' with SU2.
4. the pull request of 'Feature nppr gas #1718'

## PR Checklist

- [x] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings.
- [x] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [x] Simulations with the following test fluids succeed: nitrogen, oxygen, carbon dioxide, MM.
- [x] Currently, this new feature is applied for only Euler flow. For viscous flow, further work need to be done, including viscosity, conductivity from CoolProp. 
- [ ] CoolProp does not support non-dimension.

PR #1947, opened 2023-03-04
Bug fix of CoolProp fluid model

PENGYAN777:
## Proposed Changes
Two bugs related to CoolProp fluid model are found and solved:
1.Since there is no solver for two phase flow, then I ask CoolProp to assume the flow is single gas phase if the thermodynamics states are located on the liquid-vapor saturation curve.
 2. CoolProp has some issues to update the thermodynamics states if the input pressure is very close critical pressure Pc, see [here](https://github.com/CoolProp/CoolProp/issues/1678). So I let pressure equal to 0.99 Pc if pressure is very close to Pc.

## Related Work
This work is related to PR #1806.

## PR Checklist


SU2_CFD/src/fluid/CCoolProp.cpp
```diff
@@ -56,28 +56,51 @@ void CCoolProp::SetTDState_rhoe(su2double rho, su2double e) {
   dTdrho_e = fluid_entity->first_partial_deriv(CoolProp::iT, CoolProp::iDmass, CoolProp::iUmass);
   dTde_rho = fluid_entity->first_partial_deriv(CoolProp::iT, CoolProp::iUmass, CoolProp::iDmass);
   if (fluid_entity->phase() == 6) {
```
pcarruscag, line 58: Is there a constant like CoolProp::iphase_gas that you can use instead of "6"?
```suggestion
  if (fluid_entity->phase() == CoolProp::iphase_?) {
```

SU2_CFD/src/fluid/CCoolProp.cpp
```diff
@@ -56,28 +56,51 @@ void CCoolProp::SetTDState_rhoe(su2double rho, su2double e) {
...
+    // assume it is pure gas
+    fluid_entity->specify_phase(CoolProp::iphase_gas);
+    if (abs(Pressure / Pressure_Critical - 1) < 0.01) {
+      // if P is very cloase to Pc, the CoolProp has issue
+      Pressure = Pressure_Critical * 0.99;
+    }
```
pcarruscag, line 64: Can the pressure be higher than the critical pressure? Is the objective to put an upper bound or just to avoid P = Pc?

SU2_CFD/src/fluid/CCoolProp.cpp
```diff
@@ -56,28 +56,51 @@ void CCoolProp::SetTDState_rhoe(su2double rho, su2double e) {
...
+    fluid_entity->update(CoolProp::PT_INPUTS, Pressure, Temperature);
+    if (abs(fluid_entity->rhomass() / Density - 1) * 100 < 1) {
```
pcarruscag, line 66: It is good practice to define these thresholds (of 1%) as a constant of the class to allow tunning via a single parameter.

SU2_CFD/src/fluid/CCoolProp.cpp
PENGYAN777: Yes, the pressure can be higher or lower than critical pressure. 
The point is to avoid P=Pc or P is very close to Pc.

SU2_CFD/src/fluid/CCoolProp.cpp
PENGYAN777: Thanks for you advice.

I change "if (fluid_entity->phase() == 6)" to "if (fluid_entity->phase() == CoolProp::iphase_twophase)".

Now the code should be more meaningful.

SU2_CFD/src/fluid/CCoolProp.cpp
PENGYAN777: Good idea.

I add a consant su2double **epsilon{0.01}** in _CoolProp.h_ to control the threshold for pressure.

SU2_CFD/src/fluid/CCoolProp.cpp
WallyMaier: Would it make more sense to have a switch here?  if Pressure is close to but above Pc, would be better to set to Pressure = 1.01* Pc?
  Will there be problems crossing through the critical pressure?

PENGYAN777: You are right, it is better to add a switch here.

I added this switch to set P = 0.99 Pc or 1.01 Pc, depending P is lower or larger than Pc.

SU2_CFD/src/fluid/CCoolProp.cpp
```diff
@@ -55,29 +55,71 @@ void CCoolProp::SetTDState_rhoe(su2double rho, su2double e) {
   dPde_rho = fluid_entity->first_partial_deriv(CoolProp::iP, CoolProp::iUmass, CoolProp::iDmass);
   dTdrho_e = fluid_entity->first_partial_deriv(CoolProp::iT, CoolProp::iDmass, CoolProp::iUmass);
   dTde_rho = fluid_entity->first_partial_deriv(CoolProp::iT, CoolProp::iUmass, CoolProp::iDmass);
-  if (fluid_entity->phase() == 6) {
+  if (fluid_entity->phase() == CoolProp::iphase_twophase) {
+    // assume it is pure gas
+    fluid_entity->specify_phase(CoolProp::iphase_gas);
+    if (abs(Pressure / Pressure_Critical - 1) < epsilon) {
```
pcarruscag, line 61: Please put this modification of the pressure in a private function of the class so that you dont need to duplicate the code.

PENGYAN777: Thanks for your advice.

I added **check_pressure(P)** as a private function in **CCoolProp.h** and reuse it in **CCoolProp.cpp**. 

The code is more concise.

SU2_CFD/include/fluid/CCoolProp.hpp
```diff
@@ -30,27 +30,41 @@
...
-  su2double Gamma{1.4};           /*!< \brief Ratio of Specific Heats. */
-  su2double Gas_Constant{297};    /*!< \brief specific Gas Constant. */
-  su2double Pressure_Critical{0.0};   /*!< \brief critical pressure */
-  su2double Temperature_Critical{0.0};    /*!< \brief critical temperature */
+  su2double Gamma{1.4};                /*!< \brief Ratio of Specific Heats. */
+  su2double Gas_Constant{297};         /*!< \brief specific Gas Constant. */
+  su2double Pressure_Critical{0.0};    /*!< \brief critical pressure */
+  su2double Temperature_Critical{0.0}; /*!< \brief critical temperature */
   su2double acentric_factor{0.0};      /*!< \brief acentric factor */
+  su2double dp{0.01};             /*!< threshold for pressure */
```
pcarruscag, line 50:
```suggestion
  const su2double dp{0.01};             /*!< threshold for pressure */
```

SU2_CFD/include/fluid/CCoolProp.hpp
```diff
@@ -30,27 +30,41 @@
...
 #ifdef USE_COOLPROP
-  std::unique_ptr<CoolProp::AbstractState> fluid_entity;   /*!< \brief fluid entity */
+  std::unique_ptr<CoolProp::AbstractState> fluid_entity; /*!< \brief fluid entity */
 #endif
+  /*!
+   * \brief Avoid critical pressure
+   * \return .
```
pcarruscag, line 56:
```suggestion
   * \param[in,out] Pressure: Modified so that it is not too close to critical pressure to avoid issues in CoolProp.
```

SU2_CFD/include/fluid/CCoolProp.hpp
```diff
@@ -30,27 +30,41 @@
...
+  void Check_Pressure(su2double Pressure) {
+    if (abs(Pressure / Pressure_Critical - 1) < dp) {
+      if (Pressure >= Pressure_Critical) {
+        Pressure = Pressure_Critical * (1 + dp);
+      }
+      else {
+        Pressure = Pressure_Critical * (1 - dp);
+      }
+    }
+  }
```
pcarruscag, line 67: This function was not modifying pressure.
```suggestion
  void CheckPressure(su2double& Pressure) const {
    if (Pressure > Pressure_Critical) Pressure = fmax(Pressure, (1 + dp) * Pressure_Critical);
    else Pressure = fmin(Pressure, (1 - dp) * Pressure_Critical);
  }
```

SU2_CFD/include/fluid/CCoolProp.hpp
PENGYAN777: 'const' added

SU2_CFD/include/fluid/CCoolProp.hpp
PENGYAN777: new description added

SU2_CFD/include/fluid/CCoolProp.hpp
PENGYAN777: this function is modified following your advice.

pcarruscag (approved):

PR #1949, opened 2023-03-05
Access solution and primitives **at markers** via python wrapper

pcarruscag:


Nicola-Fonzi:
Hi @pcarruscag,

Thank you very much for taking care of this! Looks great to me the new scheme for interacting with the solver

PR #1960, opened 2023-03-10
[WIP] Pywrapper modification for steady computations

bigfooted:
## Proposed Changes
With steady computations, the pywrapper call to `SU2Driver.Run()` will run it for the number of iterations in the config file. The first change this PR makes is that a call to `SU2Driver.Preprocess(nIter)` will overwrite the number of inner iterations when the case is not time dependent. We can now do `SU2Driver.Run()` for a single iteration by setting `SU2Driver.Preprocess(1)`.

A call to `SU2Driver.Monitor(Iter)` will check the stopping criteria, including the maximum number of iterations. When you have set this to 1 using `SU2Driver.Preprocess(1)`, a call to `SU2Driver.Monitor(0)` will still stop the calculations because we are checking if (iter+1)>=nIter. The second change is that we now check if (iter >= nIter). I also change this behavior for time dependent problems to make this behavior consistent.

## PR Checklist


pcarruscag (commented):
This is too hacky, mixing the iteration counts and adding +1 here and there. Doesn't look safe.

bigfooted:
> This is too hacky, mixing the iteration counts and adding +1 here and there. Doesn't look safe.

We should of course first update the iteration count, and then check the convergence criteria for this iteration. 
Now the +1 is inside the monitor() function and monitor(iter) actually checks if iter+1 fulfills the convergence criterion, which is not what the stopping criterion should be. If we correct this, then all python files that use this have to be re-ordered a bit.  I still have to check if there are other places where we make use of the fact that the current iteration is actually Iter+1.

bigfooted:
So just to be clear: In my opinion, after calling SU2Driver.Run() the iteration count should be increased immediately. Now we postpone it in the python scripts, which means we have to add +1 in monitor to get the correct iteration count.

bigfooted:
I think it looks better now by just rearranging things. You could also decide to put Monitor at the start of the while-loop, before calling Run(), and keep the Iter+=1 at the end, but this is just nitpicking.  
 I have added an example pywrapper case that shows how to initialize velocity and species for the primitive Venturi case.

pcarruscag:
Nijso this type of change is too fragile I'm not going to approve it... There are initialization actions inside Run itself, it was not designed or tested to do single steps, and there's no chance this is going to work with all the features.

Is this required to be able to initialize the solution? Convert this PR to adding the wrapper example and I'll thing about a better way of doing single steps.

bigfooted:
It's not required fo initialization. Ok, if it's not going to work in this way let's start a discussion about a proper way to do this. I will create a new PR for just the initialization example and put it in a regression test. Thanks for the feedback.

PR #1966, opened 2023-03-15
Fix SU2 initialization when using the python wrapper

jblueh:
## Proposed Changes

The python wrapper instantiates drivers directly. To ensure that SU2 is initialized properly, also if used from python, some initialization code is already placed in the driver constructor. This PR moves additional initialization code to the driver constructor.

## Related Work
This came up during #1903.

## PR Checklist

### DIFF
--- a/TestCases/hybrid_regression_AD.py
+++ b/TestCases/hybrid_regression_AD.py
@@ -222,6 +222,37 @@ def main():
 
     pass_list = [ test.run_test() for test in test_list ]
 
+    ###################################
+    ### Python Wrapper              ###
+    ###################################
+
+    # FEA AD Flow Load Sensitivity
+    pywrapper_FEA_AD_FlowLoad               = TestCase('pywrapper_FEA_AD_FlowLoad')
+    pywrapper_FEA_AD_FlowLoad.cfg_dir       = "py_wrapper/disc_adj_fea/flow_load_sens"
+    pywrapper_FEA_AD_FlowLoad.cfg_file      = "configAD_fem.cfg"
+    pywrapper_FEA_AD_FlowLoad.test_iter     = 100
+    pywrapper_FEA_AD_FlowLoad.test_vals     = [-0.13945587401682674, -0.5859858866082689, -0.00036377840086080753, -0.00310056701747564] #last 4 columns
+    pywrapper_FEA_AD_FlowLoad.command       = TestCase.Command(exec = "python", param = "run_adjoint.py --parallel -f")
+    pywrapper_FEA_AD_FlowLoad.timeout       = 1600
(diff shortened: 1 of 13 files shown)


jblueh:
I have a couple of question and suggestions about things I noticed during moving the initialization.
1. `SU2_DOT` did not call `omp_initialize()`, it does now after moving `omp_initialize()` to the driver constructor. Does it actually run any hybrid AD stuff?
2. The number of OpenMP threads can be specified for `SU2_CFD` via the `-t` parameter, but for `SU2_DOT` and the python wrapper I saw no way to configure it. Making the number of threads a parameter of the driver constructors and calling `omp_set_num_threads` in the constructor of `CDriverBase` seems like a natural way to enable this for all flavours of SU2. Would that be a nice addition?
3. I couldn't find any python wrapper tests among the parallel AD and hybrid AD tests. Could any of the serial AD python wrapper tests serve as a basis for a parallel AD or hybrid AD one?

pcarruscag (approved):
Thanks :+1: LGTM

jblueh:
Thanks for your approval @pcarruscag. Could I have your opinion on the questions? E.g., I could add 2. as part of this PR, now that I already looked into it.

jblueh:
Thanks for your feedback! I agree, `OMP_NUM_THREADS` is better than having the number of threads in the constructor. I adapted `disc_adj_flow` and `disc_adj_fea` for MPI and added both to parallel AD and hybrid AD tests.

TestCases/py_wrapper/disc_adj_fea/flow_load_sens/run_adjoint.py
```diff
@@ -77,15 +77,18 @@ def main():
   if MarkerName in MarkerList and MarkerName in allMarkerIDs.keys():
     MarkerID = allMarkerIDs[MarkerName]
 
+  # Only print on the rank to which the marker belongs.
```
pcarruscag, line 80:
```suggestion
  # Only print on the rank to which the marker belongs.
  # WARNING: We only do this for the regression test, there is no guarantee that a marker will only belong to one rank.
```

TestCases/hybrid_regression_AD.py
```diff
@@ -222,6 +222,37 @@ def main():
 
     pass_list = [ test.run_test() for test in test_list ]
 
+    ###################################
+    ### Python Wrapper              ###
+    ###################################
+
+    # FEA AD Flow Load Sensitivity
+    pywrapper_FEA_AD_FlowLoad               = TestCase('pywrapper_FEA_AD_FlowLoad')
```
WallyMaier, line 230: Nice!

WallyMaier (approved):

PR #1978, opened 2023-03-21
Consistent python wrapper function names

pcarruscag:
I've tried to follow some convention for naming the functions. We were using "SetMarkerCustomXX" for CHT-type boundary conditions, so I extended that to other applications.
"Custom" means that the values set this way act as boundary conditions. Whereas the other accessors allow manipulating the state (e.g. for initialization) but do not change the final solution.
I also tried to disambiguate the terms for Flow loads of the flow solver vs FEA solver and mesh deformation.


SU2_CFD/include/drivers/CDriver.hpp
```diff
@@ -477,54 +478,6 @@ class CDriver : public CDriverBase {
...
-   */
-  void SetVertexNormalHeatFlux(unsigned short iMarker, unsigned long iVertex, passivedouble val_WallHeatFlux);
-
-  /*!
-   * \brief Get the thermal conductivity at a vertex on a specified marker.
-   * \param[in] iMarker - Marker identifier.
-   * \param[in] iVertex - Vertex identifier.
-   * \return Thermal conductivity at the vertex.
-   */
-  passivedouble GetThermalConductivity(unsigned short iMarker, unsigned long iVertex) const;
```
pcarruscag, line 526: This type of nodal value can be read directly from the solution accessors now, so I've removed them.

jtneedels (approved):
This LGTM.

WallyMaier (approved):

PR #1981, opened 2023-03-23
[WIP] Refactor  Code Base 

kursatyurt:
## Proposed Changes

Some method names make no sense like `GetNodeDomain` the user/coder has to look explanation and it is also not clear `Get the halo flag of a mesh node.` give no information and also provides no information, whether true indicates halo node or domain node.

Python wrapper functions are almost complete in the camel case just a few of them in the snake case. Now all of them are in camel case, consistently.

Similarly in the codebase, even inside same class some functions are the snake and others are in the camel case. In terms of the SU2 style guide, it is okay but pretty annoying for a developer. I am changing method names to at least provide some consistency inside the classes.

## PR Checklist


bigfooted:
Nice! By the way, did you use some scripts for going through all the files and returning 'bad' naming conventions?

WallyMaier:
Can you explain the snake/camel terminology? :)

pcarruscag (changes requested):
Please revert the python accessor changes.

kursatyurt:
> Nice! By the way, did you use some scripts for going through all the files and returning 'bad' naming conventions?

No, now I am just going through it by hand probably some linters are available to at least detect them. Actually, I am a bit confused as it is not so clear in the guide. It looks like the beginning of the project was just abbreviated from the Google style guide and was not enforced. Enforcing function names in UpperCamelCase and leaving variable names to the developer seems reasonable.

> Can you explain the snake/camel terminology? :)

Let's say we have a variable we want to name as  `number of nodes per mesh` there are 2 common choices as 

```cpp
number_of_nodes_per_mesh = 42;
numberOfNodesPerMesh = 42;
```
The first case is called the [snake case](https://en.wikipedia.org/wiki/Snake_case) and the second one is the [camel case ](https://en.wikipedia.org/wiki/Camel_case). Snake case is generally used in Python as [PEP8](https://peps.python.org/pep-0008/) suggests. 


> Please revert the python accessor changes.

I merged develop many of them vanished, should I revert the remaining ones?

kursatyurt:
PR merge commits are not squashed it makes a lot of trouble to rebase, I will start with a clean branch. 

Let's keep here as an open discussion about code style.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -885,7 +885,7 @@ void CSysMatrix<ScalarType>::ComputeLineletPreconditioner(const CSysVector<Scala
   /*--- Solve the tridiagonal systems for the linelets. ---*/
 
   SU2_OMP_FOR_DYN(1)
-  for (auto iLinelet = 0ul; iLinelet < li.linelets.size(); iLinelet++) {
+  for (const auto& linelet : li.linelets) {
```
pcarruscag, line 888: Doesn't work with OpenMP

SU2_CFD/include/drivers/CDriverBase.hpp
```diff
@@ -312,7 +312,7 @@ class CDriverBase {
    * \param[in] iVertex - Marker vertex index.
    * \return Marker vertex.
    */
-  unsigned long GetMarkerNode(unsigned short iMarker, unsigned long iVertex) const;
+  unsigned long GetMarkerNodeIndex(unsigned short iMarker, unsigned long iVertex) const;
```
pcarruscag, line 315: Yes this one for example.
I'm using GetNode<Property>, GetElement<Property>, etc. so adding Index doesn't really clarify much (I wanted to use Point instead of Node, but tried to respect some of the original author's personal preferences).
If you want to make GetNodeDomain clearer then call it GetNodeIsDomain to keep the convention.

You can remove underscores as you please, our use of `Worm_Case` is an eyesore. Even worse is the `What_theHeckcase`

Please try to not mix everything in a single PR, this title makes me uneasy, tackle the modernization, the proper case styles, and proposed changes to the python interface separately.

pcarruscag:
Don't rebase, just use normal merge.

SU2_CFD/include/drivers/CDriverBase.hpp
kursatyurt: I see the logic now, the index is dropped as one can suggest from the return type `unsigned int` and is x in/at/on y  
from return type bool. 

Thanks.

PR #1989, opened 2023-03-26
Add clang-tidy checks and apply them

kursatyurt:
## Proposed Changes

[clang-tidy ](https://clang.llvm.org/extra/clang-tidy/) is a well-known tool. This PR adds clang-tidy file with common checks and applies to the code base.

## PR Checklist


Common/src/adt/CADTElemClass.cpp
```diff
@@ -583,11 +585,11 @@ void CADTElemClass::Dist2ToElement(const unsigned long elemID, const su2double*
             function returns true. If the projection is not inside the triangle,
             false is returned and the distance to each of the lines of the
             triangle is computed and the minimum is taken. ---*/
-      su2double r, s;
+      su2double r = NAN, s = NAN;
```
pcarruscag, line 588: Don't initialize anything to NAN please.

kursatyurt: This comes from `cppcoreguidelines-init-variables` check/fixes. 

It is not adjustable I will change them manually. Leaving them uninitialized just leads to random numbers. What is a better choice `0.0`? 

In my view, `NaN` is not problematic, even helpful. At any point of code, we are not expecting to use created and untouched float variables initializing them to `NaN` also makes debugging easier. Where the error occurs the variable would be `NaN` if one trace back the variable it will be always `NaN` since the declaration, means something wrong with the function those variable used.

pcarruscag: It's ugly and not very useful because it is a quiet NaN, i.e. it won't be caught by the FP environment. So 0 is indeed better because at least you can catch the FP exception if you divide something.
What we should avoid is declaring variables without initializing them in the first place (which we've been requesting on new PRs). And folks are following it nicely and making things const in the process :) Is there a check for "variable could be const?"

kursatyurt: Unfortunately, there is no check for if a variable can be const.

There are checks about whether arguments can be const or not, whether const ref is better etc. 

I want to later add an [action](https://github.com/platisd/clang-tidy-pr-comments) which I think is quite nice for whom don't use clang-tidy locally. So I will change `NaN` to `0.0` rather than removing them.

kursatyurt:
@pcarruscag  I would like to get your opinion about following checks since they are related to code style and changes many things

https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html -> This can be disabled for short ones. Needs to decide short is how many characters.

https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html 

If chosen those ones can be applied just before release v8. to avoid merge conflicts

Common/include/grid_movement/CFreeFormDefBox.hpp
```diff
@@ -502,49 +502,49 @@ class CFreeFormDefBox : public CGridMovement {
...
    */
-  void SetCyl2Cart_ControlPoints(CConfig* config);
+  void SetCyl2Cart_ControlPoints(CConfig* config) const;
 
   /*!
    * \brief Set Cartesians to Cylindrical ControlPoints.
    * \param[in] config - Definition of the particular problem.
    */
-  void SetCart2Cyl_ControlPoints(CConfig* config);
+  void SetCart2Cyl_ControlPoints(CConfig* config) const;
```
pcarruscag, line 511: Many of these const changes are incorrect. These functions modify the class, but they do so through a member variable that is a pointer, and thus C++ allows const qualification since the pointer doesn't change, but the data does.
Note that the functions are called SetSomething, therefore there is clear intent in modifying something.

Don't get me wrong but I would suggest that you avoid automatic changes of this type until you know the code base better.

pcarruscag:
Else after return is fine but I don't like the others.
Isolate declarations is not addressing the true problem. Variables should be declared when needed and made const if possible.
We cannot use auto for the return type of operations with su2double

Common/include/grid_movement/CFreeFormDefBox.hpp
kursatyurt: I reverted the false ones and removed the check, almost all of them from Geometry class.

It should be fine now. Please check and merge.

Common/src/CConfig.cpp
```diff
@@ -8756,187 +8750,186 @@ unsigned short CConfig::GetMarker_SobolevBC(string val_marker) const {
...
   for (auto iMarker = 0u; iMarker < nMarker_Inlet_Turb; iMarker++) {
     if (Marker_Inlet_Turb[iMarker] == val_marker) return Inlet_TurbVal[iMarker];
   }
   if (Kind_Turb_Model == TURB_MODEL::SST) {
     return TurbIntensityAndViscRatioFreeStream;
-  } else {
-    return &NuFactor_FreeStream;
-  }
+  }     return &NuFactor_FreeStream;
+
```
pcarruscag, line 8845:
```suggestion
  }
  return &NuFactor_FreeStream;
```

pcarruscag (approved):

pcarruscag:
I went over the changes very quickly, if someone else wants to take a look it would be nice.

kursatyurt:
@bigfooted, @tbellosta,  @WallyMaier if any of you had time I would be great.

bigfooted:
> @bigfooted, @tbellosta, @WallyMaier if any of you had time I would be great.

I've had a look at the changes. I do not have strong opinions about the checks that were chosen (or not chosen), but I am glad that we have this now.

bigfooted (approved):
Great work!

PR #1990, opened 2023-03-26
Deprecate CUSP scheme and make it an error to use MUSCL_FLOW=YES with centered schemes

pcarruscag:
## Proposed Changes
I never found this scheme to be useful (not an improvement in either accuracy or robustness) and it is not used in any tests.

MUSCL_FLOW= YES is currently ignored for centered schemes, if we ever want to change that it would lead to backward compatibility issues, since we now have a chance to break it, I'm disallowing setting MUSCL_FLOW= YES with centered schemes. 

## PR Checklist


WallyMaier (approved):
LGTM.  I have had terrible success with CUSP as well.

PR #2003, opened 2023-04-05
Remove the legacy output classes

pcarruscag:
## Proposed Changes
Remove the legacy output and run turbomachinery cases with the single and multizone drivers.
The special turbomachinery outputs have been lost in the process but folks in the Turbo group are working on re-introducing them under the "new" (3 years old already :smile:) output classes.

The harmonic balance driver loses very little, it only had one small special output file in the legacy output, we do not have one output per time instance and so the output from all instances is shown on screen and piped to the same history file.
I'm leaving it like this hoping that the vacuum sucks in some new development into HB in SU2.
 
## PR Checklist


Common/include/CConfig.hpp
```diff
@@ -1147,7 +1147,7 @@ class CConfig {
   unsigned long *VolumeOutputFrequencies; /*!< \brief list containing the writing frequencies */
 
   bool Multizone_Mesh;            /*!< \brief Determines if the mesh contains multiple zones. */
-  bool SinglezoneDriver;          /*!< \brief Determines if the single-zone driver is used. (TEMPORARY) */
+  bool SinglezoneDriver = true;   /*!< \brief Determines if the single-zone driver is used. (TEMPORARY) */
```
WallyMaier, line 1150: Is this still temporary?

pcarruscag: Good point, the boolean can be removed and deduced from "not multizone" but some of the logic is still required.

PR #2014, opened 2023-04-20
Corrections to SA-NEG-R and SA model

rois1995:
## Proposed Changes
Hi, I found some problems with the implementation of the SA-NEG and SA-NEG-R models wrt to what is described on TMR. Plus, I found a suggestion on the computation of the SHat variable to avoid clipping for negative values (also reported on TMR).
 

## Related Work
- Implementation of new computation of SHat (Jacobian was computed by hand and is somewhat complex. Probably I could derive a simplified formulation)
- Added CRot constant for SA-R model
- Modified computation of var.S for the combination SA-NEG-R
- Slightly change the computation of muT from nu_tilde: now muT = 0 if nu_tilde <0. Before it was muT = max(muT, 0.0), which should not always coincide with nu_tilde > 0.
- Not related to SA: added viscous drag as output in the history file. 

I should also try to add the maximum value of nu_tilde within the mesh to the history file, although I still have to implement it.

## PR Checklist


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1371,6 +1371,8 @@ void CFlowOutput::AddAerodynamicCoefficients(const CConfig* config) {
   AddHistoryOutput("REFERENCE_FORCE", "RefForce", ScreenOutputFormat::FIXED, "AERO_COEFF", "Reference force used to compute aerodynamic coefficients", HistoryFieldType::COEFFICIENT);
   /// DESCRIPTION: Drag coefficient
   AddHistoryOutput("DRAG",       "CD",   ScreenOutputFormat::FIXED, "AERO_COEFF", "Total drag coefficient on all surfaces set with MARKER_MONITORING", HistoryFieldType::COEFFICIENT);
+  /// DESCRIPTION: viscous Drag coefficient
+  AddHistoryOutput("DRAG_VISC",       "CD_visc",   ScreenOutputFormat::FIXED, "AERO_COEFF", "Viscous drag coefficient on all surfaces set with MARKER_MONITORING", HistoryFieldType::COEFFICIENT);
```
pcarruscag, line 1375: Can we keep this one for another day? It's not consistent to have viscous contributions only for one coefficient.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -48,6 +48,7 @@ struct CSAVariables {
   const su2double cb2_sigma = cb2 / sigma;
   const su2double cw1 = cb1 / k2 + (1 + cb2) / sigma;
   const su2double cr1 = 0.5;
+  const su2double CRot = 1.0; // It should be 2.0, but the HLPW5 wants 1.0
```
pcarruscag, line 51: We want consistency with the literature.
> For example, when using a value of 1 instead of 2, the (SA-noft2-R) would instead have the naming convention (SA-noft2-R(Crot=1)).
```suggestion
  const su2double CRot = 2.0;
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -303,12 +305,36 @@ struct ModVort {
...
+	      const su2double d_Sbar = (var.fv2 + nue * var.d_fv2);
+        const su2double k2_d2 = var.k2 * var.dist_i_2;
+        const su2double num1 = c2*c2 * var.S*var.S * k2_d2;
+        const su2double den1 = pow(k2_d2 * (c3-2*c2) * var.S - nue * var.fv2, 2.0);
+        const su2double num2_1 = c3*var.S*(k2_d2 * (c3-2*c2)*var.S - nue*var.fv2);
+        const su2double num2_2 = c3*var.S*nue*var.fv2;
+        const su2double den2 = pow(k2_d2* (c3-2*c2) * var.S - nue*var.fv2, 2.0);
+
+        var.d_Shat = num1*d_Sbar/den1 + (num2_1+num2_2)*d_Sbar/den2 ;
+      }
```
pcarruscag, line 337: It's better to keep the Jacobian simple.
```suggestion
      const su2double Num = var.S * (c2 * c2 * var.S + c3 * Sbar);
      const su2double Den = (c3 - 2 * c2) * var.S - Sbar;
      var.Shat = var.S + Num / Den;
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -303,12 +305,36 @@ struct ModVort {
...
     }
```
pcarruscag, line 338:
```suggestion
    }
    if (var.Shat <= 1e-10) {
      var.Shat = 1e-10;
      var.d_Shat = 0.0;
    } else {
      var.d_Shat = (var.fv2 + nue * var.d_fv2) * var.inv_k2_d2;
    }
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
rois1995, line 1: Do you think I should remove the computation of the derivatives of SHat and keep only the simple one?

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -307,34 +308,20 @@ struct Bsl {
...
-        const su2double den2 = pow(k2_d2* (c3-2*c2) * var.S - nue*var.fv2, 2.0);
-
-        var.d_Shat = num1*d_Sbar/den1 + (num2_1+num2_2)*d_Sbar/den2 ;
-      }
+    }
+    if (var.Shat <= 1e-10) {
+      var.Shat = 1e-10;
+      var.d_Shat = 0.0;
+    } else {
+      var.d_Shat = (var.fv2 + nue * var.d_fv2) * var.inv_k2_d2;
```
rois1995, line 324: Do you think I should remove the computation of the derivatives of SHat and keep only the simple one?

SU2_CFD/src/output/CFlowOutput.cpp
rois1995: Yeah sure, I added it here just because HLPW5 requires the history of the viscous drag coefficient. We could leave it here for those who want to use this branch for HLPW5 computations and then remove it just before the merge. You tell me.

pcarruscag: Yep we can do it that way. If you output the forces breakdown file you should have the different contributions for each coefficient.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -235,7 +235,7 @@ void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_contain
 
     su2double muT = rho*fv1*nu_hat;
 
-    if (neg_spalart_allmaras) muT = max(muT,0.0);
+    if (neg_spalart_allmaras && nu_hat < 0) muT = 0.0;
```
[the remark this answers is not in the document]
pcarruscag: Ah yes the sign can flip again if nu_hat is very negative

SU2_CFD/src/output/CFlowOutput.cpp
rois1995: Yeah but I do not have it in the history file, right?

pcarruscag: Not at the moment

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
bigfooted: Another keyword CROT1 then? Or make CROT=2.0 a config option?

pcarruscag:
Can you guys wrap this up and update the required restart files?
I want to make the fixes to the SST model we want for v8.

pcarruscag:
Some of the testcases that show as failed need to be updated, if the update is significant for cases with restart files we need a new restart file to match the previous convergence level.

bigfooted:
I just updated the serial and parallel regression tests. There is one SA test that has a restart file, I will also update that one.
@rois1995 can you have a look at the failed cases for the tutorial.py and vandv.py regression tests? I will have a look at the *_AD.py regressions this Friday.

rois1995:
@bigfooted  sure, I'll look into it right now.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -122,7 +124,9 @@ class CSourceBase_TurbSA : public CNumerics {
 
     /*--- Dacles-Mariani et. al. rotation correction ("-R"). ---*/
     if (options.rot) {
-      var.Omega += 2.0 * min(0.0, StrainMag_i - var.Omega);
+      var.Omega += var.CRot * min(0.0, StrainMag_i - var.Omega);
+      /*--- Do not allow negative production for SA-neg. ---*/
+      if(ScalarVar_i[0] < 0) var.Omega = abs(var.Omega);
```
pcarruscag, line 129:
```suggestion
      if (ScalarVar_i[0] < 0) var.Omega = abs(var.Omega);
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1386,6 +1386,8 @@ void CFlowOutput::AddAerodynamicCoefficients(const CConfig* config) {
...
```
pcarruscag, line 1390:
```suggestion
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1443,6 +1445,7 @@ void CFlowOutput::SetAerodynamicCoefficients(const CConfig* config, const CSolve
 
   SetHistoryOutputValue("REFERENCE_FORCE", flow_solver->GetAeroCoeffsReferenceForce());
   SetHistoryOutputValue("DRAG", flow_solver->GetTotal_CD());
+  SetHistoryOutputValue("DRAG_VISC", flow_solver->GetAllBound_CD_Visc());
```
pcarruscag, line 1448:
```suggestion
```

pcarruscag (approved):

PR #2031, opened 2023-05-13
Add word-by-word comparison with tolerance and threshold to file diff.

danielmayer:
## Proposed Changes
The current file comparison that is used for regression tests, does not feature a possibility to specify a tolerance or a threshold. This PR implements such a feature and does not affect the existing regression tests.
 

## Related Work
Currently, there are no regression tests that make use of this feature and the PR does not include a new test that would use it. However, the feature_new_flamelet branch already uses this feature. This branch here is to split off the feature as it is not related to flamelet models but it is a general feature that can used for a variety of regression tests that use a file comparison. 

## PR Checklist

- [O] I have added a test case that demonstrates my contribution, if necessary.
- [O] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.


TestCases/TestCase.py
```diff
@@ -325,7 +334,90 @@ def run_filediff(self):
...
+                                # Compare floats
+                                if from_isfloat:
+                                    try:
+                                        # Only do a relative comparison when the threshold is met.
+                                        # This is to prevent large relative differences for very small numbers.
+                                        if (abs(float(from_word)) > self.comp_threshold):
+                                            delta = abs( (float(from_word) - float(to_word)) / float(from_word) ) * 100
+                                            compare_counter += 1
+                                        else:
+                                            delta = 0.0
```
pcarruscag, line 397: Should we compare absolute values in this case? If all values happen to be small this could pass everything.

pcarruscag (approved):
Thank you Daniel, LGTM after taking care of any CodeQL warnings

TestCases/TestCase.py
danielmayer: Thanks for your review and comments Pedro. The main reason to compare relative values is that we want to run this comparison on restart files that hold quantities that vary on very different ranges. E.g., enthalpy might range on O(1e6) while progress variable typically ranges on O(1). I am not sure if specifying the same tolerance for these quantities for an absolute comparison would make much sense as a fall back solution. But we can definitely add it as an alternative for files that are better suited for absolute comparisons. What do you think?

Regarding your comment on CodeQL warnings: where do I find those? Do you mean the issues found by CodeFactor?

pcarruscag: I meant when the values are below the tolerance, instead of assuming 0, doing an absolute comparison. But I understand the motivation.
Yes CodeFactor, if it is only complaining about white spaces it's fine.

bigfooted (approved):
Thanks, Daniel, looks good.

TestCases/TestCase.py
danielmayer: The value of "delta" will later be used to check if the compared value is within the specified relative tolerance. If we assign the absolute difference here it gets inconsistent. My line of thinking was that it's most straightforward to ignore values below the threshold and make the developer, who hooks up his test case, responsible for specifying reasonable values for tolerance and threshold. As a guide, we report the number of ignored and compared file entries, as well as the maximum difference at the end of the test. Do you think that's ok to ask of the other developers? 

I removed some white space and a semicolon which CodeFactor didn't like.

pcarruscag: Sounds good

PR #2036, opened 2023-05-17
Fix functionality of 'WRT_*_OVERWRITE' flags.

danielmayer:
## Proposed Changes
The flags 'WRT_VOLUME_OVERWRITE', 'WRT_SURFACE_OVERWRITE', and  'WRT_RESTART_OVERWRITE' are not working correctly. This commit fixes this.
 
The expected behavior is this:
The user specifies what kind of output files should be written using the OUTPUT_FILES flag. These output files can be volume or surface files for visualization like Tecplot or Paraview files, volumetric restart files to be used as initial conditions for another SU2 run, or mesh format files like CGNS or STL files.
Furthermore, the user can choose to not overwrite these files, but rather add the current iteration number to the file name and keep dumping such files during the calculation. This is done using the three WRT_*_OVERWRITE flags mentioned above.

However, the current implementation in develop only works correctly for PARAVIEW_MULTIBLOCK and for RESTART type output files. This is because in the case of PARAVIEW_MULTIBLOCK, the additional file output using the iteration number in the filename, is performed in the switch case block directly. 

For all other output files, the switch case block only generates the correct filename with the iteration number appended. After the switch block, it is (wrongly) checked if (and only if) the WRT_RESTART_OVERWRITE flag is set to NO. If so, the additional file is written. Therefore, this logic only works correctly for RESTART files (and PARAVIEW_MULTIBLOCK files, because for that format, the file dump is independent of the file dump after the switch block. 

The proposed fix here introduces a local boolean (doWriteIterFile) that acts as a switch for the additional file dump. It is set in the switch case block and afterwards it is checked if this variable is true or false. Based on this value, the file dump is executed or not. Furthermore, the fix introduces and corrects some comments to make the function easier to understand.

## Related Work
The flags are already available in the develop branch as config options but they don't work as expected.

## PR Checklist


danielmayer:
> Can you describe the bug this is fixing a bit more

I edited the PR description. Hope it's clear now. Please let know if you need more info.

pcarruscag:
I see, I get it now thanks (it sounded like everything was not working :) )

pcarruscag (approved):

SU2_CFD/src/output/COutput.cpp
```diff
@@ -853,29 +711,27 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, OUTPUT_TYPE form
...
 
     fileWriter->WriteData(fileName);
 
     su2double BandWidth = fileWriter->GetBandwidth();
 
-    /*--- Write data with iteration number to file ---*/
+    /*--- Write data with iteration number to file if required ---*/
 
-    if (!filename_iter.empty() && !config->GetWrt_Restart_Overwrite()){
+    if (!filename_iter.empty()) {
```
pcarruscag, line 724: It was enough to base it just on the filename :) but in the process I simplified a bit more

pcarruscag (approved):

danielmayer:
Sweet! Thank you Pedro!

PR #2038, opened 2023-05-21
Periodic BC: add dummy BCs.

ShiheJia:
## Proposed Changes
Add periodic BCs with dummy points.  With internal markers, we could probe data from any inner surfaces, and here, we choose to construct periodic boundaries using dummy points which is preset on the input domain just like the internal markers. Four layers in both side is needed for MUSCL: the inner two layers send and the outer layers recieve.
With this way, we can use a larger CFL number similar to the cases using symmetry BCs and avoid passing too many variables with mpi.
 
## Related Work
#1467 or other works about periodic BCs

## PR Checklist

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -224,6 +224,7 @@ private:
   string Inlet_Filename;        /*!< \brief Filename specifying an inlet profile. */
   su2double Inlet_Matching_Tol; /*!< \brief Tolerance used when matching a point to a point from the inlet file. */
   string ActDisk_FileName;      /*!< \brief Filename specifying an actuator disk. */
+  bool Periodic_DummyBC;        /*!< \brief True if achieve periodic BCs with dummy points. */
 
   string *Marker_Euler,           /*!< \brief Euler wall markers. */
   *Marker_FarField,               /*!< \brief Far field markers. */
@@ -4866,6 +4867,12 @@ public:
    * \return <code>FALSE</code> means that the sharp edges will be removed from the sensitivity.
    */
   bool GetSens_Remove_Sharp(void) const { return Sens_Remove_Sharp; }
+
+  /*!
+   * \brief Check if achieve periodic BCs with dummy points.
+   * \return True if achieve periodic BCs with dummy points.
+   */
+  bool GetPeriodic_DummyBC(void) const { return Periodic_DummyBC; }
 
   /*!
    * \brief Get the kind of inlet boundary condition treatment (total conditions or mass flow).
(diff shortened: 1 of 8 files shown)


pcarruscag (commented):
I think you are mixing concepts from structured and unstructured world. In unstructured vertex based we have half of the dual volume on each side of the periodicity, its the same variables on each side but rotated, that's why we combine the residuals.
Our issue is that this is hard to represent in the Jacobian.
MUSCL in unstructured is done with gradients, if you use inconsistent treatment you lose the equivalence between the periodic case and one with repeated mesh sections.

ShiheJia:
Thank you for your comment. No need to create dummy points for periodic BC is really an ingenious design and when I run RANS/URANS cases, I get good results. But there are still some problems for me, the first is convergence when I run big 3D cases and the second is the periodic BC cannot cooperate with hybrid RANS/LES method.

So I thought a dummy layer is sometimes necessary just as in Blazek's book(Computational Fluid Dynamics: Principles and Applications, Second Edition, Page 293). 
book

One dummy layer need to communicate solution, gradients or maybe Laplacian, sensors and so on. It's still complex so I choose two dummy layers to only communicate solution. As in the picture, the original use the send-recieve pair1 as periodic BCs and this feature need four pairs  with a restriction that the outermost four layers should be structured grids.  Three more repeated mesh sections are indeed added but it is acceptable in terms of robustness and reducing data communication.
pairs

The above diagram can be set with the following settings:
MARKER_PERIODIC= ( y0.48, y0.04, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.44, 0.0, y0, y0.44, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.44, 0.0 )
PERIODIC_DUMMY= YES

This functionality also requires a pre-processing program for the grid to add the appropriate boundary conditions to the .su2 file. I am currently doing this using a fortran code and trying to replicate it using python scripts. But the changes to the c++ code section have been verified with various RANS/URANS/DES cases and a result of NACA0012 swept wings is listed below.

pcarruscag:
The dummy layer is what we used before version 7 and moved away from it for simplicity.
Two layers doesn't sound possible for unstructured meshes.

ShiheJia:
Thanks for your comments, commonality does suffer.

pcarruscag:
Can you describe the issue you see for hybrid rans /LES? Maybe something is not implemented correctly

ShiheJia:
It takes some time to prepare. 
It seems that in some case, periodic BCs  run with hybrid rans /LES method well, such as a tandem cylinder case that I give similar settings and mesh with Edward Molina's paper. But when I run some case, such as the following A-Airfoil case, the hybrid rans /LES method cannot run well. The reduction of the CFL number from 100 to 5 does not help, and just change hybrid rans /LES to URANS get a very normal iteration with CFL=100.
Mesh:




Setting and output:
[set.cfg.txt](https://github.com/su2code/SU2/files/11543866/set.cfg.txt)
[cfl100.output.txt](https://github.com/su2code/
[cfl5.output.txt](https://github.com/su2code/SU2/files/11543872/cfl5.output.txt)
SU2/files/11543871/cfl100.output.txt)

PR #2060, opened 2023-06-24
Add local timestep + CFL as volume output

j-signorelli:
## Proposed Changes
Added local timestep and local CFL as volume output for both the flow and turbulence variables
 

## Related Work
In relation to #2056

## PR Checklist

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -92,6 +92,7 @@ Jessie Lauzon
 João Loureiro
 Johannes Blühdorn
 JonathanSmith1936
+Joseph Signorelli
 Josy P. Pullockara
 Kedar Naik
 Kürşat Yurt
--- a/SU2_CFD/src/output/CFlowOutput.cpp
+++ b/SU2_CFD/src/output/CFlowOutput.cpp
@@ -1451,6 +1451,15 @@ void CFlowOutput::SetVolumeOutputFieldsScalarMisc(const CConfig* config) {
     }
     AddVolumeOutput("Q_CRITERION", "Q_Criterion", "VORTEX_IDENTIFICATION", "Value of the Q-Criterion");
   }
+
+  // Timestep info
+  AddVolumeOutput("DELTA_TIME", "Delta_Time", "TIMESTEP", "Value of the local timestep for the flow variables");
+  AddVolumeOutput("CFL", "CFL", "TIMESTEP", "Value of the local CFL for the flow variables");
+  if (config->GetKind_Turb_Model() != TURB_MODEL::NONE)
+  {
+    AddVolumeOutput("TURB_DELTA_TIME", "Turb_Delta_Time", "TIMESTEP", "Value of the local timestep for the turbulence variables");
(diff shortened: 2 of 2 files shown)


pcarruscag:
Thanks, please add yourself to AUTHORS.md

pcarruscag (approved):

j-signorelli:
Hi @pcarruscag, just wanted to follow up with you on something about this: it _looks_ like to me that the volume output $\Delta t$'s are nondimensional. Would it be worthwhile to mention this in the `SU2_CFD -d` call? Or even just calculate the actual physical time as the volume output alternatively? This may be confusing.

pcarruscag:
Our reference for time is always 1 so they should match physical time even if they appear non-dimensional. What area of the code made it look nondimensional?

j-signorelli:
Oh interesting, just to be fully complete here's what made me think that:

1.I have a well-converged steady-state solution and outputted the minimum DT's for each cell when running with a CFL=1.0, no adaptive. I was unable to restart this for an unsteady one with no changes in BCs without using a DT orders-of-magnitude smaller.

2. When I run /Tutorials/compressible_flow/Turbulent_OneraM6 with `MIN_DELTA_TIME` set as a screen output, I get orders-of-magnitude different final timestep when I run with `REF_DIMENSIONALIZATION= DIMENSIONAL` instead of `FREESTREAM_VEL_EQ_ONE`. 

3. When I run `SU2_CFD -d` on the /Tutorials/compressible_flow/Unsteady_NACA0012 example **using `FREESTREAM_VEL_EQ_ONE`** I see the attached image below, which appears to nondimensionalize using length/velocity.

Unsteady_NACA0012_SU2CFDd

4. I also looked briefly at the code it _appeared_ that the calculations were using nondimensional values. Just to be fully complete with you, here's what I looked at:

Setting of local timestep for inviscid:
https://github.com/su2code/SU2/blob/b80e4ce0e4fd05afa136af0990a17dd619cdce42/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp#L506-L522

Calculation of Lambda's:
https://github.com/su2code/SU2/blob/b80e4ce0e4fd05afa136af0990a17dd619cdce42/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp#L413-L444

GetProjVel:
https://github.com/su2code/SU2/blob/b80e4ce0e4fd05afa136af0990a17dd619cdce42/SU2_CFD/include/variables/CEulerVariable.hpp#L244-L254

Then at this point admittedly I just assumed that the `MatrixType Primitive` of `CEulerVariable` was indeed nondimensional. I am not familiar with where/when data is stored/updated throughout SU2 as nondimensional v dimensional.

pcarruscag:
You are right, but in nondim mode all outputs are the nondim ones, so this is consistent.

j-signorelli:
Would it be worthwhile to output the time step reference line for steady simulations in the "Solver Preprocessing" section as well then? This is only done for unsteady simulations.

pcarruscag:
Sure sounds good

PR #2062, opened 2023-06-26
Make SST-1994m consistent with literature and use 2003m by default

### LINKED ISSUE #1551: SST(m) implementation is inconsistent with literature

pcarruscag:
## Proposed Changes
Remove divergence terms from the source term, use 2003m by default.

## Related Work
Closes #1551

## PR Checklist


TestCases/hybrid_regression.py
```diff
@@ -234,7 +234,7 @@ def main():
     turb_naca0012_sst_expliciteuler.cfg_dir   = "rans/naca0012"
     turb_naca0012_sst_expliciteuler.cfg_file  = "turb_NACA0012_sst_expliciteuler.cfg"
     turb_naca0012_sst_expliciteuler.test_iter = 10
-    turb_naca0012_sst_expliciteuler.test_vals = [-3.532228, -3.157766, 3.364025, 1.124824, 0.501717, -float("inf")]
+    turb_naca0012_sst_expliciteuler.test_vals = [-3.532289, -3.157766, 3.364024, 1.122901, 0.500798, -inf]
```
pcarruscag, line 237:
```suggestion
    turb_naca0012_sst_expliciteuler.test_vals = [-3.532289, -3.157766, 3.364024, 1.122901, 0.500798, -float("inf")]
```

bigfooted (approved):
Great work, looks like we are following the nasa website description of the model now, and the residuals did not change by a lot so that's comforting.

PR #2067, opened 2023-07-03
Bug fix cht flamelet solver

EvertBunschoten:
## Proposed Changes
The conjugate heat variables were not communicated properly between the flow and flamelet solver, resulting in the interface temperature between the fluid and solid to be always maintained at 300 kelvin for flamelet simulations. The changes mainly regard the recognition of the inclusion of energy for flamelet problems for the CHT methods, while in the flamelet solver, the conjugate variables are obtained from the flow solver instead of the flamelet solver itself. 
 

## Related Work
Related work includes the pull requests regarding the flamelet solver, as well as the related regression tests. The latter were obtained with this bug included, so the solution files have to be updated and the residual values in the regression tests have to be updated. 

## PR Checklist


SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -427,7 +427,7 @@ void CSpeciesFlameletSolver::BC_Isothermal_Wall_Generic(CGeometry* geometry, CSo
   for (unsigned long iVertex = 0; iVertex < geometry->nVertex[val_marker]; iVertex++) {
     unsigned long iPoint = geometry->vertex[val_marker][iVertex]->GetNode();
 
-    if (cht_mode) temp_wall = GetConjugateHeatVariable(val_marker, iVertex, 0);
+    if (cht_mode) temp_wall = solver_container[FLOW_SOL]->GetConjugateHeatVariable(val_marker, iVertex, 0);
```
pcarruscag, line 430: In this case does the flamelet solver need conjugate heat variables at all?
Also I don't think this would work for weakly coupled energy, where the temperature is stored in the heat solver instead of the flow solver.

EvertBunschoten: The flamelet solver indeed doesn't need conjugate heat variables anymore. I just removed those lines. Regarding the weakly coupled energy, I don't know for sure. When running flamelet simulations, the energy transport equation is solved by the flamelet solver, not the flow solver, so my best guess is that weakly coupled heat should be incompatible with flamelet simulations.

EvertBunschoten: I discussed with @bigfooted regarding the compatibility of the flamelet solver with the weakly coupled heat solver. We concluded that it does not make much sense to use both weakly coupled heat and the flamelet fluid model in the same zone, as the flamelet fluid model computes the temperature based on the transport of controlling variables. This prevents the temperature from being an independent, transported scalar variable, as is the case for the weakly coupled heat option.

.github/workflows/regression.yml
```diff
@@ -128,7 +128,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:230225-2136
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c bug_fix_cht_flamelet -s ${{matrix.testscript}}
```
pcarruscag, line 131: makes sense, PR for the test case update?

EvertBunschoten: I just opened a pull request for the highlighted test case branch.

pcarruscag, line 131:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```

PR #2078, opened 2023-07-12
Custom Python wrapper marker BC values appropriately nondimensionalized

j-signorelli:
## Proposed Changes
In current release of SU2, having, for example, a `MARKER_ISOTHERMAL=(marker, 300)` and `MARKER_PYTHON_CUSTOM=(marker)` causes the BC temperature to be set to a _nondimensional_ value of 300, which is inconsistent with what that marker's value is supposed to mean. This can be seen being set here:

https://github.com/su2code/SU2/blob/b80e4ce0e4fd05afa136af0990a17dd619cdce42/Common/src/geometry/CGeometry.cpp#L2441-L2461

and called for example here:

https://github.com/su2code/SU2/blob/b80e4ce0e4fd05afa136af0990a17dd619cdce42/SU2_CFD/src/solvers/CNSSolver.cpp#L712-L714

This PR nondimensionalizes any custom boundary temperatures or heat fluxes appropriately so that any setting of them does not need to be nondimensionalized manually before, both in `MARKER_ISOTHERMAL` and `MARKER_HEATFLUX` **and** in any Python code setting values of them.

## Related Work
None that I am aware of.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CHeatSolver.cpp
+++ b/SU2_CFD/src/solvers/CHeatSolver.cpp
@@ -377,7 +377,7 @@ void CHeatSolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_conta
   SU2_OMP_FOR_STAT(OMP_MIN_SIZE)
   for (auto iVertex = 0ul; iVertex < geometry->nVertex[val_marker]; iVertex++) {
     if (IsPyCustom) {
-      Twall = geometry->GetCustomBoundaryTemperature(val_marker, iVertex);
+      Twall = geometry->GetCustomBoundaryTemperature(val_marker, iVertex) / config->GetTemperature_Ref();
     }
     IsothermalBoundaryCondition(geometry, solver_container[FLOW_SOL], config, val_marker, iVertex, Twall);
   }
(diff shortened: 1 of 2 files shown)


pcarruscag:
Thanks, can you look for other places where the values are used, I think the incompressible solver might need the same fix.

j-signorelli:
A `grep -R CustomBoundaryTemperature .` and `grep -R CustomBoundaryHeatFlux .` doesn't seem to show anything done for the incompressible solver, and also doesn't appear to be an option here at least:
https://github.com/su2code/SU2/blob/db21a33291a6ac17cdf85c61625c3b87441149c6/SU2_CFD/src/solvers/CIncNSSolver.cpp#L350-L372

However, I _might_ have missed a spot here:

https://github.com/su2code/SU2/blob/db21a33291a6ac17cdf85c61625c3b87441149c6/Common/src/geometry/CMultiGridGeometry.cpp#L960-L994

but my _guess_ is that it's fine since values from here are re-nondimensionalized from the above commits. Please let me know if not however - I am not familiar with the specifics of multigrid and its implementation.

Note that the commits above presume that values specified in `CustomBoundaryTemperature` and `CustomBoundaryHeatFlux` in `CGeometry` are dimensional, irregardless if SU2 is in nondimensional mode. 

I will look into why one of the regression tests failed and get back to you.

pcarruscag:
You are right, custom temperature is not implemented in the incompressible solver.
@bigfooted the inlet profiles are always dimensional now right?

bigfooted:
> You are right, custom temperature is not implemented in the incompressible solver. @bigfooted the inlet profiles are always dimensional now right?

Yes, the inlet profiles are dimensional.

pcarruscag (approved):

Issue #2074, opened 2023-07-09
"PARAVIEW_MULTIBLOCK" issue with path that contains period


idocruhman:
I have enabled the writing of the individual surfaces using the "PARAVIEW_MULTIBLOCK" directive.
The folder was created on the wrong place because the path contained periods.
for a simulation run in /foo/bar/polar_M0.6/aoa0/config.cfg
the folder was created in /foo/bar/polar_M0/zone_0
I assume this is due to the foder name being generated by cutting the path at the first period to remove the .cfg from the name.
While I understand that chaging the name will fix the issue, a period is a legal character for a folder name.

 - OS: Centos 7.5
 - C++ compiler and version: not sure
 - MPI implementation and version: not sure
 - SU2 Version: 7.5.0

bigfooted:
We actually do a find_last_of('.'), which strips the filename extension, if it exists. But it leads to the problem that you describe if the filename does not have an extension but the folder name does.
The documentation says that VOLUME_FILENAME should be provided without the extension/suffix, so we could also remove the find_last_of('.') to prevent unwanted chopping of the folders and filenames.

pcarruscag:
We can check for the last "." that is after the last / (if one exists).

bigfooted:
Maybe a user wants filenames like flow.a, flow.b, and I think we should allow this.
I think we should not be too smart about it and not strip anything. Then the code does what the documentation says it does.

pcarruscag:
Ok.

PR #2079, opened 2023-07-12
Fix volume names

bigfooted:
## Proposed Changes
For volume names, we assume that the filename is filename.ext and we strip the extension using find_last_of('.')
But:
1. This can lead to unwanted file stripping when you want the filename or folder to be volume.a
2. We already mention in the documentation that the volume filename should be given without an extension

What we do in the PR: 
filenames that have multiple possible extensions (csv/dat or cgns/su2) are not required to have their extension in the config file. However, if the correct extension is provided, this will be stripped internally so it is easy to append iteration numbers, zone numbers etc. This means that when the expected filetype is solution.csv but the user provides solution.dat, then the .dat extension will not be stripped but seen as part of the 'base' or 'prefix'. When iterations are added, the new filename will be solution.dat_001.csv.

## Related Work
Issue: #2074 

## PR Checklist


bigfooted:
There are some other find_last_of lines as well, if all filenames are supposed to be given without any file extension, then these should all be removed.

pcarruscag:
In principle sounds good, it will probably break lots of cases, optimization scripts and what not. We'll see, we are not leaving this inconsistent though, all or nothing.

bigfooted:
> In principle sounds good, it will probably break lots of cases, optimization scripts and what not. We'll see, we are not leaving this inconsistent though, all or nothing.

I agree.

SU2_CFD/src/output/CMeshOutput.cpp
```diff
@@ -38,7 +38,7 @@ CMeshOutput::CMeshOutput(CConfig *config, unsigned short nDim) : COutput(config,
 
   /*--- Set the volume filename --- */
 
-  volumeFilename = config->GetMesh_Out_FileName();
+  volumeFilename = config->GetMesh_Out_FileName() + ".su2";
```
pcarruscag: I don't like the hard-coded extensions everywhere.
We should encapsulate this name giving and extensions into the function to retrieve the name.
This encapsulation should also make it possible to handle names both with and without extensions, forcing everyone to remove extensions from their config is a large break of backwards compatibility and I don't want to release version 9 for something as trivial as one user complaining that the named without extension didn'e work...

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -5688,8 +5688,6 @@ void CPhysicalGeometry::SetTurboVertex(CConfig* config, unsigned short val_iZone
     char buffer[50];
 
     if (GetnZone() > 1) {
-      unsigned short lastindex = multizone_filename.find_last_of('.');
-      multizone_filename = multizone_filename.substr(0, lastindex);
       SPRINTF(buffer, "_%d.dat", SU2_TYPE::Int(val_iZone));
       multizone_filename.append(string(buffer));
     }
```
pcarruscag, line 5696: This looks wrong, were either missing the extension when we arrive here or we are adding it twice.

Common/src/grid_movement/CSurfaceMovement.cpp
```diff
@@ -3514,8 +3514,6 @@ void CSurfaceMovement::SetExternal_Deformation(CGeometry* geometry, CConfig* con
        physical time, so perform mesh motion in reverse. ---*/
       unsigned long nFlowIter = config->GetnTime_Iter() - 1;
       flowIter = nFlowIter - iter;
-      unsigned short lastindex = DV_Filename.find_last_of('.');
-      DV_Filename = DV_Filename.substr(0, lastindex);
       if ((SU2_TYPE::Int(flowIter) >= 0) && (SU2_TYPE::Int(flowIter) < 10))
         SPRINTF(buffer, "_0000%d.dat", SU2_TYPE::Int(flowIter));
```
pcarruscag, line 3523: same

Common/src/grid_movement/CSurfaceMovement.cpp
```diff
@@ -3530,8 +3528,6 @@ void CSurfaceMovement::SetExternal_Deformation(CGeometry* geometry, CConfig* con
     } else {
       /*--- Forward time for the direct problem ---*/
       flowIter = iter;
-      unsigned short lastindex = DV_Filename.find_last_of('.');
-      DV_Filename = DV_Filename.substr(0, lastindex);
       if ((SU2_TYPE::Int(flowIter) >= 0) && (SU2_TYPE::Int(flowIter) < 10))
         SPRINTF(buffer, "_0000%d.dat", SU2_TYPE::Int(flowIter));
```
pcarruscag, line 3537: same... this logic needs to be contained in helper functions, it will be a pain to get rid of the bugs this way... we'll have to reason about extensions in every area which is the opposite of what you are trying to achieve here I think.

Common/src/grid_movement/CSurfaceMovement.cpp
```diff
@@ -4922,8 +4918,8 @@ void CSurfaceMovement::WriteFFDInfo(CSurfaceMovement** surface_movement, CGeomet
     /*--- Read the name of the output file ---*/
 
     auto str = config[ZONE_0]->GetMesh_Out_FileName();
-    unsigned short lastindex = str.find_last_of('.');
-    str = str.substr(0, lastindex) + ".su2";
+    //unsigned short lastindex = str.find_last_of('.');
+    str += ".su2";
```
pcarruscag: The function should return the name with extension and the callers of the function should not need to make further ad-hoc modifications

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2313,9 +2313,17 @@ void CDriver::InitializeNumerics(CConfig *config, CGeometry **geometry, CSolver
...
+
+      unsigned short lastindex = filename.find_last_of('.');
+      filename = filename.substr(0, lastindex);
+
       if (nZone > 1)
-        filename = config->GetMultizone_FileName(filename, iZone, ".dat");
+        filename = config->GetMultizone_FileName(filename, iZone, "");
+
+      /*--- At the moment we stay consistent with the default extension .dat for all input files. ---*/
+      filename += ".dat";
```
pcarruscag: the function should handle the extension, not the client

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -2433,12 +2433,6 @@ void CFlowOutput::WriteForcesBreakdown(const CConfig* config, const CSolver* flo
   const auto nMonitoring = config->GetnMarker_Monitoring();
 
   auto fileName = config->GetBreakdown_FileName();
-  if (unsteady) {
-    const auto lastindex = fileName.find_last_of('.');
-    const auto ext = fileName.substr(lastindex, fileName.size());
-    fileName = fileName.substr(0, lastindex);
-    fileName = config->GetFilename(fileName, ext, curTimeIter);
-  }
```
pcarruscag: seems wrong

SU2_CFD/src/output/COutput.cpp
```diff
@@ -82,25 +82,28 @@ COutput::COutput(const CConfig *config, unsigned short ndim, bool fem_output):
...
   restartFilename = "restart";
 
-  /*--- Retrieve the history filename ---*/
 
-  historyFilename = config->GetConv_FileName();
 
   /*--- Add the correct file extension depending on the file format ---*/
 
   string hist_ext = ".csv";
   if (config->GetTabular_FileFormat() == TAB_OUTPUT::TAB_TECPLOT) hist_ext = ".dat";
```
pcarruscag: same, move the entire name logic into a helper function

SU2_CFD/src/output/CMeshOutput.cpp
bigfooted: So we accept "restart" as well as "restart.dat" etc...? 
But we check if the extension is actually the correct one, so restart.csv will not be stripped, but seen as having 'no extension'?

SU2_PY/SU2/io/state.py
```diff
@@ -270,11 +270,11 @@ def find_files(self, config):
         adjoint_name = config.SOLUTION_ADJ_FILENAME
 
         if "RESTART_ASCII" in config.get("OUTPUT_FILES", ["RESTART"]):
-            direct_name = direct_name.split(".")[0] + ".csv"
-            adjoint_name = adjoint_name.split(".")[0] + ".csv"
+            direct_name = direct_name + ".csv"
+            adjoint_name = adjoint_name + ".csv"
```
pcarruscag, line 274: the idea to keep things backwards compatible is:
```suggestion
            direct_name = direct_name + ("" if direct_name.endswith(".csv") else ".csv")
            adjoint_name = adjoint_name + ".csv"
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -82,25 +82,28 @@ COutput::COutput(const CConfig *config, unsigned short ndim, bool fem_output):
...
 
-  /*--- Append the zone ID ---*/
+  /*--- Retrieve the history filename ---*/
+
+  historyFilename = config->GetConv_FileName();
```
bigfooted: @pcarruscag so we add the hist_ext stuff to GetConv_Filename?

pcarruscag: Yeah that and any logic to suffix with number of iterations.

pcarruscag: Whenever a given file name (like the history or breakdown) is only used for a particular type of file, the accessor to get the file name can return it with extension.
The for output files whose extension depends on the type of format being output, we can add an accessor per file type, or make it clear in the name accessor that the name is without extension. However, it should contain everything else in the name that we need, zone idx, time step, etc.
I think this is the safest way to make this change while providing backward compatibility easily in one place.

SU2_CFD/src/output/CMeshOutput.cpp
pcarruscag: Yes but that would be an user error so I think it's fine.

SU2_CFD/src/output/COutput.cpp
bigfooted: ok,
I move the entire filename construction into GetConv_FileName. 
I strip ".dat" and ".csv" only and keep other extensions.  Users can still use "history.dat", but using "history"  or "history.1.2.3.old" will now work.

pcarruscag:
I'll take a look when I can

Common/include/CConfig.hpp
```diff
@@ -2276,7 +2275,23 @@ class CConfig {
...
+    /*--- If multizone, append zone name ---*/
+    if (Multizone_Problem)
+      FEAFilename = GetMultizone_FileName(FEAFilename, GetiZone(), "");
+
+    /*--- Add the extension again ---*/
+    FEAFilename += ".dat";
+
+    /*--- return the stripped filename base, without extension. ---*/
+    return FEAFilename;
+    }
```
```suggestion
  }
```

Common/include/CConfig.hpp
```diff
@@ -5021,11 +5036,7 @@ class CConfig {
    */
   bool GetInlet_Profile_From_File(void) const { return Inlet_From_File; }
 
-  /*!
-   * \brief Get name of the input file for the specified inlet profile.
-   * \return Name of the input file for the specified inlet profile.
-   */
-  string GetInlet_FileName(void) const { return Inlet_Filename; }
+
```
```suggestion
```

Common/src/CConfig.cpp
```diff
@@ -7467,7 +7465,7 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
   }
 
   if (val_software == SU2_COMPONENT::SU2_DEF) {
-    cout << "Output mesh file name: " << Mesh_Out_FileName << ". " << endl;
+    cout << "Output mesh file name: " << Mesh_Out_FileName << ".su2 . " << endl;
```
pcarruscag:
```suggestion
    cout << "Output mesh file name: " << GetMesh_Out_FileName() << ".su2. " << endl;
```

SU2_CFD/src/output/CFlowOutput.cpp
pcarruscag: Still seems wrong, where is the current iteration being appended now?

SU2_CFD/src/output/COutput.cpp
```diff
@@ -383,18 +368,11 @@ void COutput::LoadData(CGeometry *geometry, CConfig *config, CSolver** solver_co
...
 }
 
-void COutput::WriteToFile(CConfig *config, CGeometry *geometry, OUTPUT_TYPE format, string fileName){
+void COutput::WriteToFile(CConfig *config, CGeometry *geometry, OUTPUT_TYPE format, string fileName) {
 
   /*--- File writer that will later be used to write the file to disk. Created below in the "switch" ---*/
   CFileWriter *fileWriter = nullptr;
 
-  /*--- Set current time iter even if history file is not written ---*/
-  curTimeIter = config->GetTimeIter();
```
pcarruscag: I believe this was a fix for something, double check with git-blame please.

Common/include/CConfig.hpp
```diff
@@ -2279,8 +2280,7 @@ class CConfig {
     string FEAFilename = FEA_FileName;
 
     /*--- strip the extension if it is present, only if it is .dat ---*/
-    auto extIndex = FEAFilename.rfind(".dat");
-    if (extIndex != std::string::npos) FEAFilename.resize(extIndex);
+    PrintingToolbox::TrimExtension(".dat", FEAFilename);
```
pcarruscag, line 2283: Use this helper please, it checks that the extension is actually at the end of the file

pcarruscag (approved):

pcarruscag:
Can you do a diff on the entire screen output for the failing case? develop vs your branch? I think this case uses the metadata file, it seems to be ok based on this:
```
WARNING: SU2 will use the STREAMWISE_PERIODIC_PRESSURE_DROP provided in the direct solution file: 241.878867596797
```
But worth checking, maybe it got broken at some point.

SU2_CFD/src/output/COutput.cpp
bigfooted: yes, fix_UnstNames, I'll revert it.

bigfooted: current iteration is now also appended to breakdown file.

Common/include/CConfig.hpp
```diff
@@ -5549,20 +5558,70 @@ class CConfig {
...
+  string GetMesh_FileName(void) const {
+
+    /*--- we keep the original Mesh_FileName  ---*/
+    string meshFilename = Mesh_FileName;
+
+    /*--- strip the extension, only if it is .su2 or .cgns ---*/
+    auto extIndex = meshFilename.rfind(".su2");
+    if (extIndex != std::string::npos) meshFilename.resize(extIndex);
+    extIndex = meshFilename.rfind(".cgns");
+    if (extIndex != std::string::npos) meshFilename.resize(extIndex);
```
pcarruscag, line 5570: I added the TrimExtension helper so it could be used everywhere you trim extensions

PR #2083, opened 2023-07-17
Fix wall time for discrete adjoint MZ driver + OpenMP the GMRES orthogonalization when used by MZ driver

pcarruscag:


Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -452,7 +452,9 @@ unsigned long CSysSolve<ScalarType>::FGMRES_LinSolver(const CSysVector<ScalarTyp
 
     /*---  Modified Gram-Schmidt orthogonalization ---*/
 
+    SU2_OMP_PARALLEL_(if (!omp_in_parallel()))
```
bigfooted, line 455: How do I interpret this?
omp_in_parallel returns nr of active levels, so ModGramSchmidt is not allowed to use nesting?

bigfooted (approved):
looks good to me.

Common/src/linear_algebra/CSysSolve.cpp
pcarruscag: You can interpret as "make it parallel if not parallel yet", but it's broken because of shared access to the arguments of the function.
I need to think a bit more on this.

PR #2090, opened 2023-07-25
Fix composite surface motions

pcarruscag:
fixes this issue: https://www.cfd-online.com/Forums/su2/250979-problem-oscillating-airfoil.html


WallyMaier (approved):
LGTM

PR #2109, opened 2023-08-14
Fix HeatFlux non-dimensionalization

TobiKattmann:
## Proposed Changes
Hi all,

volume heatflux output is not non-dimensional which is inconsistent with how the rest of volume output is handled.
Note that the History outputs (like TOTAL_HEATFLUX over a bunch of surfaces) is re-dimensionalized -> Apparently that hasn't bothered too many people until now but generally that might be sth we/I want to change (?) Let me know

I fixed it in a way that follows the current convention.

In CConjugateHeatInteface.cpp the required heat_flux is (re)computed on its own so I do not expect a problem there but I will check manually as well
- [ ] Reminder Checkmark to actually do so
If there is other places which might be affected let me know.

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
+++ b/SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
@@ -2430,7 +2430,6 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
   const su2double Alpha = config->GetAoA() * PI_NUMBER / 180.0;
   const su2double Beta = config->GetAoS() * PI_NUMBER / 180.0;
   const su2double RefLength = config->GetRefLength();
-  const su2double RefHeatFlux = config->GetHeat_Flux_Ref();
   const su2double Gas_Constant = config->GetGas_ConstantND();
   auto Origin = config->GetRefOriginMoment(0);
 
@@ -2587,7 +2586,7 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
           if (!energy) dTdn = 0.0;
           thermal_conductivity = nodes->GetThermalConductivity(iPoint);
         }
-        HeatFlux[iMarker][iVertex] = -thermal_conductivity * dTdn * RefHeatFlux;
+        HeatFlux[iMarker][iVertex] = -thermal_conductivity * dTdn;
 
       } else {
 
@@ -2658,8 +2657,11 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
         MomentZ_Force[0] += (-Force[0] * Coord[1]);
         MomentZ_Force[1] += (Force[1] * Coord[0]);
 
-        HF_Visc[iMarker] += HeatFlux[iMarker][iVertex] * Area;
-        MaxHF_Visc[iMarker] += pow(HeatFlux[iMarker][iVertex], MaxNorm);
+        /*--- Note that Screen Outputs are re-dimensionalzed but volume outputs are not.
+         *    Since the HeatFlux-container is used for volume output as well, introduce the re-dim here. ---*/
+        const su2double dimensional_HeatFlux = HeatFlux[iMarker][iVertex] * config->GetHeat_Flux_Ref();
+        HF_Visc[iMarker] += dimensional_HeatFlux * Area;
+        MaxHF_Visc[iMarker] += pow(dimensional_HeatFlux, MaxNorm);
       }
     }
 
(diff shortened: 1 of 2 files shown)


j-signorelli:
I also found out the hard way that `TOTAL_HEATFLUX` was dimensional, but one additional thing to consider here is that if I call `CDriverBase::GetMarkerNormalHeatFlux` in a Python-wrapped code then I think this PR would cause the returned heat flux to be nondimensional, which can be a nuisance if using it for CHT while running in nondimensional mode. If for FSI, `CDriverBase::GetMarkerFlowLoad` is the dimensional flow load, then this PR would also be inconsistent with that; but if that function returns nondim. loads then this update makes much more sense (I am not sure if it's nondim or not).

TobiKattmann:
@j-signorelli afai see this PR would indeed cause `CDriverBase::GetMarkerNormalHeatFlux` to return nondimensional HeatFlux (compared to dimensional HeatFlux in current master) and could trouble people there ... but in doubt we can still dimensionalize in that function to keep the python wrapper as-is. (But there might be a cleaner solution to this)

As I do not know much about the structural solver in SU2, I cannot say with confidence -> but from what I saw from skimming over CFEA(Base)Solver, config_template, and a handful of Elasticity cfg's there is no nondimensionalization for the structural solver (but somebody with better knowledge of that should educate me here).

pcarruscag:
Let's keep the dimensional outputs, we are already trying to have dimensional inputs to keep things simple.
In the longer term we should aim to have dimensional inputs and outputs and make the non-dimensionalization an internal detail of the solvers that users don't have to worry with.

pcarruscag:
Including volume outputs, which right now are a pain for restarting with different types of non-dim

jaspe55:
Hello Tobi, by the way, is there a good explanation as to why even for an
adiabatic wall, and REF_DIMENSIONALIZATION= DIMENSIONAL, I get  an abs
(heat flux) > 200 at the conclusion of the flow solver, when
log10(Residual) ~ -9?

TobiKattmann:
Hi jaspe55, I would have to check -> a possible explanation I can think of rn (without having checkde) is that we compute the heatflux on the wall based on the temperature gradient of wall temperature with an interior Temperature and that can lead to a non-zero heatflux while there is non at the boundary. Of course that should tend to zero upon refinement and for an adiabatic case. 

Maybe I am on the wrong track here but I'll try to check soon

pcarruscag:
you are correct, there is a discussion on CFD-online about it. we impose 0 heat flux, but report an "apparent heat flux" most codes will simply give you back the imposed heat flux value you specify, nevertheless there will probably be a temperature gradient close to the wall

jaspe55:
Thank you so much, Tobi and Pedro. Yes, there is a temperature gradient
close to the wall.  So  SU2 gives me, in this case, a 'virtual heat flux'!
Kind regards.

jaspe55:
Yet another doubt on the same subject: I just realized that SU2V7.31 (I
have not tested it on later SU2 releases) accepts imposing both adiabatic
walls plus isothermal boundary condition on some walls (when running Menter
model), but when I simply impose only isothermal conditions (leaving
commented out the adiabatic markers), I get the message: SU2 process
returned error '1'. Is there a workaround for this? Kind regards.

pcarruscag:
Is there a simple example to reproduce the issue?

jaspe55:
Unfortunately, my example involves a swirler and nozzle for the
Navier-Stokes solver with Menter model, so the dataset is relatively large.
Let me check whether some older small Euler mesh exhibits the same message!
I will let you know! Thank you very much.

jaspe55:
Hello Pedro. While searching a small sample dataset, I have run some
variations on the jobs, and my findings were:
(Please note that *all* "markers" I have mentioned in the next items refer
to solid walls in the mesh, using the RANS solver)
1- When I impose HEATFLUX=0 *and* ISOTHERMAL=290 Kelvin, for the same
markers, there are no complaints from SU2. It shows that SU2 ignores the
imposition of HEATFLUX=0, in this case.
2- Just setting ISOTHERMAL=290 Kelvin (using all of the same markers as in
(item 1), above), without any  HEATFLUX setting, then SU2 produces the same
results as in (item1).
3- When some markers are set with HEATFLUX=0 and the rest of them are set
with ISOTHERMAL=290 Kelvin (so that each wall marker in the mesh is
referenced), there are no complaints from SU2, and the job goes as expected.
4-The error message mentioned earlier, appears when one or more solid wall
markers *is not marked* concerning HEATFLUX or ISOTHERMAL
.
My conclusion then is that the error message appeared when I mistakenly
have not included a wall marker (assigning it as a HEATFLUX or an
ISOTHERMAL marker!)
Does it make sense?
Kind regards,

pcarruscag:
Yes that makes sense, we are probably missing some validation on markers not appearing on incompatible boundary conditions.

jaspe55:
Thank you very much, Pedro!
Kind regards

PR #2110, opened 2023-08-15
Signal handling - SIGTERM will stop calculation and force file saving

bigfooted:
## Proposed Changes
Special signal handling for SIGTERM. When SIGTERM is sent to the solver, we catch it and simply exit the for-loop for the number of iterations. Files will then automatically be saved.

## Related Work
Related to Github discussion #2094 

## PR Checklist


bigfooted:
Here is a simple proof-of-concept for the fluid iteration where we catch SIGTERM. If you run it and hit ctrl-c, we catch the signal and we set stopcalc to true. This also forces a saving of the files.
I think implementing the signal handler like this is the best, because we let the solver handle all the cleanup.

bigfooted:
We probably also have to do something like this at other places where we have external iterations.

pcarruscag (commented):
This should be at the driver level, not inside iteration classes otherwise it won't work for multizone.
How are signals handled by MPI? Can this result in a race condition?

bigfooted:
OK, so I am moving the signal handling to COutput::ConvergenceMonitoring as discussed in the dev-meeting.

bigfooted:
@pcarruscag This works for me. Do you think this is OK or we need to do it different/more, maybe a config keyword to activate/deactivate, or other options?

pcarruscag (commented):
I have not had time to test it for single/multi zone, steady/unsteady, and primal/adjoint.

pcarruscag (approved):

PR #2039, opened 2023-05-22
Hybrid Parallel AD Performance Improvements

jblueh:
## Proposed Changes
When accessing the adjoints, bounds checking acquires an internal lock in CoDiPack. This turned out to be a bottleneck in `InitializeAdjoint` and `IterateDiscreteAdjoint`. Therefore, we ensure sufficient size up front and skip the bounds checking.

## Related Work
Previous work on hybrid parallel AD, like #1214, #1284, #1294.

## PR Checklist

- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [X] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [X] I have added a test case that demonstrates my contribution, if necessary.
- [X] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

PR #2112, opened 2023-08-22
Fix adjoint vector size

jblueh:
## Proposed Changes
This PR fixes an issue in the discrete adjoint multizone driver where adjoints are seeded before sufficient adjoint vector size is guaranteed, resulting in an objective function gradient of zero.

The PR also contains other small fixes.
 
## PR Checklist

### DIFF
--- a/Common/include/basic_types/ad_structure.hpp
+++ b/Common/include/basic_types/ad_structure.hpp
@@ -66,7 +66,7 @@ inline void PrintStatistics() {}
  * graph. \param[in] data - The variable to be registered as input. \param[in] push_index - boolean whether we also want
  * to push the index.
  */
-inline void RegisterInput(su2double& data, bool push_index = true) {}
+inline void RegisterInput(su2double& data) {}
 
 /*!
  * \brief Registers the variable as an output. I.e. as the root of the computational graph.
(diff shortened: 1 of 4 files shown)


Common/include/basic_types/ad_structure.hpp
```diff
@@ -66,7 +66,7 @@ inline void PrintStatistics() {}
  * graph. \param[in] data - The variable to be registered as input. \param[in] push_index - boolean whether we also want
  * to push the index.
```
pcarruscag, line 67: Nice stuff :+1: is there any way to get warned about this in the future?
Also looks like we can update this comment.

jblueh: Thanks for noting, I updated the comment. `AD::SetDerivative` used to have an implicit size check and automatic resizing if needed. This was removed in #2039 because it harms the hybrid AD performance if called in parallel, like from the `InitializeAdjoint` functions. I will add assertions to CoDiPack though, so that there are warnings about this when debugging with `-DCODI_EnableAssert`.

jblueh: Makes sense, I updated CoDiPack and added the flag

pcarruscag (approved):

PR #2137, opened 2023-09-20
[WIP] Standard initialization for Incompressible flows using thermodynamic_pressure and the Inc_Temperature_init

Cristopher-Morales:
## Proposed Changes

With this pull request, we aim to initialize the fluid models based on the Ideal gas laws for incompressible flows in a standard way using the THERMODYNAMIC_PRESSURE($P_{op}$) and the INC_TEMPERATURE_INIT ($T_{init}$), the initial density is computed based on the ideal gas law :

$\rho_{init}= P_{op} / R T_{init}$

When the FLUID_MODEL is CONSTANT_DENSITY, the constant density must be given in the config file using the option DENSITY_CONSTANT, similar as it is done when viscosity, thermal conductivity and diffusivity models are chosen as constant.

## PR Checklist


Common/src/CConfig.cpp
```diff
@@ -3022,6 +3023,11 @@ void CConfig::SetConfig_Parsing(istream& config_buffer){
             newString.append("DYNAMIC_ANALYSIS is deprecated. Use TIME_DOMAIN instead.\n\n");
           else if (!option_name.compare("SPECIES_USE_STRONG_BC"))
             newString.append("SPECIES_USE_STRONG_BC is deprecated. Use MARKER_SPECIES_STRONG_BC= (marker1, ...) instead.\n\n");
+          else if (!option_name.compare("INC_DENSITY_INIT"))
+            newString.append(
+                "INC_DENSITY_INIT is deprecated. The initial density is computed using the ideal gas law based on the "
+                "THERMODYNAMICS_PRESSURE and INC_TEMPERATURE_INIT. For Using the FLUID_MODEL=CONSTANT_DENSITY, the "
+                "constant density must be given using the following option DENSITY_CONSTANT.\n\n");
```
pcarruscag, line 3030: We just released a major version, we are not breaking backward compatibility for an important setting like this.

bigfooted: Sure, we can wait for a major version to implement a change in config option. But now, sometimes we use thermodynamic pressure and sometimes we use density (to compute thermodynamic pressure) and I think the inconsistency should be fixed at some point, and I think it should be fixed in this way.

pcarruscag: I'm not saying we cannot fix things, but the option needs to be deprecated more gracefully than this.
As for the inconsistency between density init and pressure init we can use a similar strategy to the compressible solver.
```
/*!
 * \brief Types of initialization option
 */
enum class FREESTREAM_OPTION {
  TEMPERATURE_FS, /*!< \brief Temperature initialization. */
  DENSITY_FS, /*!< \brief Density initalization. */
};
```

TobiKattmann:
I am trying to find some time to think about how to do a bit of option-consolidation on the non-dimensionalization side and initialization is quite a bit linked to that process -> I hope to open a discussion with some more information and proposal soon -> And since this PR is looking to change the INC_DENSITY_INIT option  which is part of that my request would be to maybe spare that option up to the conference (the init change to use thermodynamic pressure is  a little  more independent of that I think via pedros suggestion to mimic what we do for compressible flow atm)

Common/include/CConfig.hpp
```diff
@@ -821,6 +821,7 @@ class CConfig {
   su2double Gamma,      /*!< \brief Ratio of specific heats of the gas. */
   Bulk_Modulus,         /*!< \brief Value of the bulk modulus for incompressible flows. */
   Beta_Factor,          /*!< \brief Value of the epsilon^2 multiplier for Beta for the incompressible preconditioner. */
+  Density_Constant,     /*!< \brief Constant Density for ConstantDensity model (only for incompressible flows). */
```
TobiKattmann, line 824: this is a test

PR #2150, opened 2023-10-22
[WIP] Implementation of DDES formulations for SST

rois1995:
## Proposed Changes
Hi everyone,

I have been working on the implementation of the DDES formulations for the SST model, following the work in "Development of DDES and IDDES Formulations for the k-ω Shear Stress Transport Model" (DOI:10.1007/s10494-011-9378-4). The implementation is easy, whereas the validation may take some time. I am currently working on the backward facing step and I will work on the flatplate too. The choice of the validation test cases is key since the computational cost is quite large, thus if you have any suggestions that may speed up this part then feel free to write them down. 

## PR Checklist

### DIFF
--- a/Common/include/option_structure.hpp
+++ b/Common/include/option_structure.hpp
@@ -1380,7 +1380,10 @@ enum ENUM_HYBRIDRANSLES {
   SA_DES   = 1,          /*!< \brief Kind of Hybrid RANS/LES (SA - Detached Eddy Simulation (DES)). */
   SA_DDES  = 2,          /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Delta_max SGS ). */
   SA_ZDES  = 3,          /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Vorticity based SGS like Zonal DES). */
-  SA_EDDES = 4           /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Shear Layer Adapted SGS: Enhanced DDES). */
+  SA_EDDES = 4,           /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Shear Layer Adapted SGS: Enhanced DDES). */
+  SST_DDES = 5,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): DDES). */
+  SST_IDDES = 6,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): Improved DDES). */
+  SST_SIDDES = 7           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): Simplified Improved DDES). */
 };
 static const MapType<std::string, ENUM_HYBRIDRANSLES> HybridRANSLES_Map = {
   MakePair("NONE", NO_HYBRIDRANSLES)
@@ -1388,6 +1391,9 @@ static const MapType<std::string, ENUM_HYBRIDRANSLES> HybridRANSLES_Map = {
   MakePair("SA_DDES", SA_DDES)
   MakePair("SA_ZDES", SA_ZDES)
   MakePair("SA_EDDES", SA_EDDES)
+  MakePair("SST_DDES", SST_DDES)
+  MakePair("SST_IDDES", SST_IDDES)
+  MakePair("SST_SIDDES", SST_SIDDES)
 };
 
 /*!
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -6157,6 +6157,9 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
           case SA_DDES:  cout << "Delayed Detached Eddy Simulation (DDES) with Standard SGS" << endl; break;
           case SA_ZDES:  cout << "Delayed Detached Eddy Simulation (DDES) with Vorticity-based SGS" << endl; break;
           case SA_EDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_DDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_IDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_SIDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
         }
         break;
       case MAIN_SOLVER::NEMO_EULER:
--- a/SU2_CFD/include/numerics/CNumerics.hpp
+++ b/SU2_CFD/include/numerics/CNumerics.hpp
@@ -190,6 +190,8 @@ protected:
 
   bool bounded_scalar = false;    /*!< \brief Flag for bounded scalar problem */
 
+  su2double lengthScale_i, lengthScale_j;
+
 public:
   /*!
    * \brief Return type used in some "ComputeResidual" overloads to give a
@@ -826,6 +828,16 @@ public:
   void SetDistance(su2double val_dist_i, su2double val_dist_j) {
     dist_i = val_dist_i;
     dist_j = val_dist_j;
+  }
+
+  /*!
+   * \brief Set the value of the length scale for SST.
+   * \param[in] val_lengthScale_i - Value of of the length scale for SST from point i.
+   * \param[in] val_lengthScale_j - Value of of the length scale for SST from point j.
+   */
+  void SetLengthScale(su2double val_lengthScale_i, su2double val_lengthScale_j) {
+    lengthScale_i = val_lengthScale_i;
+    lengthScale_j = val_lengthScale_j;
   }
 
   /*!
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -834,6 +834,9 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
       /*--- Dissipation ---*/
 
       su2double dk = beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0];
+      if (config->GetKind_HybridRANSLES() != NO_HYBRIDRANSLES)
+        dk = Density_i * sqrt(ScalarVar_i[0]*ScalarVar_i[0]*ScalarVar_i[0]) / lengthScale_i;
+        
       su2double dw = beta_blended * Density_i * ScalarVar_i[1] * ScalarVar_i[1];
 
       /*--- LM model coupling with production and dissipation term for k transport equation---*/
--- a/SU2_CFD/include/variables/CTurbSAVariable.hpp
+++ b/SU2_CFD/include/variables/CTurbSAVariable.hpp
@@ -39,7 +39,6 @@
 class CTurbSAVariable final : public CTurbVariable {
 
 private:
-  VectorType DES_LengthScale;
   VectorType Vortex_Tilting;
 
 public:
@@ -59,19 +58,6 @@ public:
    * \brief Destructor of the class.
    */
   ~CTurbSAVariable() override = default;
-
-  /*!
-   * \brief Get the DES length scale
-   * \param[in] iPoint - Point index.
-   * \return Value of the DES length Scale.
-   */
-  inline su2double GetDES_LengthScale(unsigned long iPoint) const override { return DES_LengthScale(iPoint); }
-
-  /*!
-   * \brief Set the DES Length Scale.
-   * \param[in] iPoint - Point index.
-   */
-  inline void SetDES_LengthScale(unsigned long iPoint, su2double val_des_lengthscale) override { DES_LengthScale(iPoint) = val_des_lengthscale; }
 
   /*!
    * \brief Set the vortex tilting measure for computation of the EDDES length scale
--- a/SU2_CFD/include/variables/CVariable.hpp
+++ b/SU2_CFD/include/variables/CVariable.hpp
@@ -385,6 +385,25 @@ public:
    * \brief A virtual member.
    * \param[in] iPoint - Point index.
    */
+  inline virtual void SetDebug_Quantities(CConfig *config, unsigned long iPoint, su2double val_ftilda_d, su2double val_l_RANS, su2double val_l_LES, su2double val_r_d) {}
+  
+  inline virtual su2double Get_L_RANS(unsigned long iPoint) const { return 0.0; }
+  inline virtual su2double Get_L_LES(unsigned long iPoint) const { return 0.0; }
+  inline virtual su2double Get_ftilda_d(unsigned long iPoint) const { return 0.0; }
+  inline virtual su2double Get_r_dt(unsigned long iPoint) const { return 0.0; }
+  inline virtual su2double Get_r_dl(unsigned long iPoint) const { return 0.0; }
+  inline virtual su2double Get_r_d(unsigned long iPoint) const { return 0.0; }
+
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   */
+  inline virtual void SetDebug_Quantities(CConfig *config, unsigned long iPoint, su2double val_ftilda_d, su2double val_l_RANS, su2double val_l_LES, su2double val_r_dl, su2double val_r_dt) {}
+
+  /*!
+   * \brief A virtual member.
+   * \param[in] iPoint - Point index.
+   */
   virtual void SetSolution_New() {}
 
   /*!
--- a/SU2_CFD/src/output/CFlowOutput.cpp
+++ b/SU2_CFD/src/output/CFlowOutput.cpp
@@ -1481,6 +1481,7 @@ void CFlowOutput::SetVolumeOutputFieldsScalarMisc(const CConfig* config) {
       AddVolumeOutput("L_LES", "l_LES", "DDES", "DES length scale value");
       AddVolumeOutput("R_DT", "r_dt", "DDES", "DES length scale value");
     }
+    AddVolumeOutput("LESIQ", "LESIQ", "DDES", "LESIQ index for SRS simulations");
   }
 
   if (config->GetViscous()) {
@@ -1597,6 +1598,10 @@ void CFlowOutput::LoadVolumeDataScalar(const CConfig* config, const CSolver* con
       SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
       SetVolumeOutputValue("R_DT", iPoint, Node_Turb->Get_r_dt(iPoint));
     }
+    const su2double mut = Node_Flow->GetEddyViscosity(iPoint);
+    const su2double mu = Node_Flow->GetLaminarViscosity(iPoint); 
+    const su2double LESIQ = 1.0/(1.0+0.05*pow((mut+mu)/mu, 0.53));
+    SetVolumeOutputValue("LESIQ", iPoint, LESIQ);
   }
 
   switch (config->GetKind_Species_Model()) {
--- a/SU2_CFD/src/output/CFlowOutput.cpp
+++ b/SU2_CFD/src/output/CFlowOutput.cpp
@@ -1464,6 +1464,23 @@ void CFlowOutput::SetVolumeOutputFieldsScalarMisc(const CConfig* config) {
   if (config->GetKind_HybridRANSLES() != NO_HYBRIDRANSLES) {
     AddVolumeOutput("DES_LENGTHSCALE", "DES_LengthScale", "DDES", "DES length scale value");
     AddVolumeOutput("WALL_DISTANCE", "Wall_Distance", "DDES", "Wall distance value");
+    if ( config->GetKind_HybridRANSLES() == SST_DDES){
+      AddVolumeOutput("F_D", "f_d", "DDES", "DES length scale value");
+      AddVolumeOutput("L_RANS", "l_RANS", "DDES", "DES length scale value");
+      AddVolumeOutput("L_LES", "l_LES", "DDES", "DES length scale value");
+      AddVolumeOutput("R_D", "r_d", "DDES", "DES length scale value");
+    } else if ( config->GetKind_HybridRANSLES() == SST_IDDES){
+      AddVolumeOutput("F_D", "f_d", "DDES", "DES length scale value");
+      AddVolumeOutput("L_RANS", "l_RANS", "DDES", "DES length scale value");
+      AddVolumeOutput("L_LES", "l_LES", "DDES", "DES length scale value");
+      AddVolumeOutput("R_DT", "r_dt", "DDES", "DES length scale value");
+      AddVolumeOutput("R_DL", "r_dl", "DDES", "DES length scale value");
+    } else if ( config->GetKind_HybridRANSLES() == SST_SIDDES){
+      AddVolumeOutput("F_D", "f_d", "DDES", "DES length scale value");
+      AddVolumeOutput("L_RANS", "l_RANS", "DDES", "DES length scale value");
+      AddVolumeOutput("L_LES", "l_LES", "DDES", "DES length scale value");
+      AddVolumeOutput("R_DT", "r_dt", "DDES", "DES length scale value");
+    }
   }
 
   if (config->GetViscous()) {
@@ -1563,6 +1580,23 @@ void CFlowOutput::LoadVolumeDataScalar(const CConfig* config, const CSolver* con
   if (config->GetKind_HybridRANSLES() != NO_HYBRIDRANSLES) {
     SetVolumeOutputValue("DES_LENGTHSCALE", iPoint, Node_Flow->GetDES_LengthScale(iPoint));
     SetVolumeOutputValue("WALL_DISTANCE", iPoint, Node_Geo->GetWall_Distance(iPoint));
+    if ( config->GetKind_HybridRANSLES() == SST_DDES){
+      SetVolumeOutputValue("F_D", iPoint, Node_Turb->Get_ftilda_d(iPoint));
+      SetVolumeOutputValue("L_RANS", iPoint, Node_Turb->Get_L_RANS(iPoint));
+      SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
+      SetVolumeOutputValue("R_D", iPoint, Node_Turb->Get_r_d(iPoint));
+    } else if ( config->GetKind_HybridRANSLES() == SST_IDDES){
+      SetVolumeOutputValue("F_D", iPoint, Node_Turb->Get_ftilda_d(iPoint));
+      SetVolumeOutputValue("L_RANS", iPoint, Node_Turb->Get_L_RANS(iPoint));
+      SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
+      SetVolumeOutputValue("R_DT", iPoint, Node_Turb->Get_r_dt(iPoint));
+      SetVolumeOutputValue("R_DL", iPoint, Node_Turb->Get_r_dl(iPoint));
+    } else if ( config->GetKind_HybridRANSLES() == SST_SIDDES){
+      SetVolumeOutputValue("F_D", iPoint, Node_Turb->Get_ftilda_d(iPoint));
+      SetVolumeOutputValue("L_RANS", iPoint, Node_Turb->Get_L_RANS(iPoint));
+      SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
+      SetVolumeOutputValue("R_DT", iPoint, Node_Turb->Get_r_dt(iPoint));
+    }
   }
 
   switch (config->GetKind_Species_Model()) {
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -1107,7 +1107,7 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
     switch(kind_hybridRANSLES){
       case SST_DDES: {
 
-        const su2double r_d = (eddyVisc + lamVisc) / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double r_d = (eddyVisc + lamVisc) / max((KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag))), 1e-10);
         const su2double C_d1 = 20.0;
         const su2double C_d2 = 3.0;
 
@@ -1132,11 +1132,11 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
 
         const su2double alpha = 0.25 - sqrt(wallDist2) / h_max;
         const su2double f_b = min(2.0 * exp(-9.0 * alpha*alpha), 1.0);
-        const su2double r_dt = eddyVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double r_dt = eddyVisc / max((KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag))), 1e-10);
         const su2double f_dt = 1 - tanh(pow(C_dt1 * r_dt, C_dt2));
         const su2double ftilda_d = max(1.0 - f_dt, f_b);
 
-        const su2double r_dl = lamVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double r_dl = lamVisc / max((KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag))), 1e-10);
         const su2double f_l = tanh(pow(C_l*C_l*r_dl, 10.0));
         const su2double f_t = tanh(pow(C_t*C_t*r_dt, 3.0));
         const su2double f_e2 = 1.0 - max(f_t, f_l);
@@ -1164,7 +1164,7 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
 
         const su2double alpha = 0.25 - sqrt(wallDist2) / h_max;
         const su2double f_b = min(2.0 * exp(-9.0 * alpha*alpha), 1.0);
-        const su2double r_dt = eddyVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double r_dt = eddyVisc / max((KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag))), 1e-10);
         const su2double f_dt = 1 - tanh(pow(C_dt1 * r_dt, C_dt2));
         const su2double ftilda_d = max(1.0 - f_dt, f_b);
 
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -1117,6 +1117,8 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
         
         DES_lengthScale = l_RANS - f_d * max(0.0, l_RANS - l_LES);
 
+        nodes->SetDebug_Quantities(config, iPoint, f_d, l_RANS, l_LES, r_d);
+
         break;
       }
       case SST_IDDES: {
@@ -1145,6 +1147,8 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
         const su2double Delta = min(C_w * max(sqrt(wallDist2), h_max), h_max);
         const su2double l_LES = C_DES * Delta;
 
+        nodes->SetDebug_Quantities(config, iPoint, ftilda_d, l_RANS, l_LES, r_dl, r_dt);
+
         DES_lengthScale = ftilda_d *(1.0+f_e)*l_RANS + (1.0 - ftilda_d) * l_LES;
 
         break;
@@ -1166,6 +1170,8 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
 
         const su2double Delta = min(C_w * max(sqrt(wallDist2), h_max), h_max);
         const su2double l_LES = C_DES * Delta;
+
+        nodes->SetDebug_Quantities(config, iPoint, ftilda_d, l_RANS, l_LES, r_dt);
 
         DES_lengthScale = ftilda_d*l_RANS + (1.0 - ftilda_d) * l_LES;
 
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -188,8 +188,19 @@ void CTurbSSTSolver::Preprocessing(CGeometry *geometry, CSolver **solver_contain
          unsigned short iMesh, unsigned short iRKStep, unsigned short RunTime_EqSystem, bool Output) {
   SU2_OMP_SAFE_GLOBAL_ACCESS(config->SetGlobalParam(config->GetKind_Solver(), RunTime_EqSystem);)
 
+  const auto kind_hybridRANSLES = config->GetKind_HybridRANSLES();
+
   /*--- Upwind second order reconstruction and gradients ---*/
   CommonPreprocessing(geometry, config, Output);
+
+  if (kind_hybridRANSLES != NO_HYBRIDRANSLES) {
+
+    /*--- Compute the DES length scale ---*/
+
+    SetDES_LengthScale(solver_container, geometry, config);
+
+  }
+
 }
 
 void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_container,
@@ -356,6 +367,10 @@ void CTurbSSTSolver::Source_Residual(CGeometry *geometry, CSolver **solver_conta
     if (axisymmetric){
       /*--- Set y coordinate ---*/
       numerics->SetCoord(geometry->nodes->GetCoord(iPoint), geometry->nodes->GetCoord(iPoint));
+    }
+
+    if (config->GetKind_HybridRANSLES() != NO_HYBRIDRANSLES) {
+      numerics->SetLengthScale(nodes->GetDES_LengthScale(iPoint), 0.0);
     }
 
     /*--- Compute the source term ---*/
@@ -1059,4 +1074,107 @@ void CTurbSSTSolver::SetUniformInlet(const CConfig* config, unsigned short iMark
     }
   }
 
+}
+
+
+void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CConfig *config){
+
+  const auto kind_hybridRANSLES = config->GetKind_HybridRANSLES();
+
+
+  auto* flowNodes = su2staticcast_p<CFlowVariable*>(solver[FLOW_SOL]->GetNodes());
+
+  SU2_OMP_FOR_STAT(omp_chunk_size)
+  for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
+    const su2double StrainMag = max(flowNodes->GetStrainMag(iPoint), 1e-12);
+    const su2double VortMag = max(GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint)), 1e-12);
+
+    const su2double KolmConst2 = 0.41*0.41;
+    const su2double wallDist2 = geometry->nodes->GetWall_Distance(iPoint)*geometry->nodes->GetWall_Distance(iPoint); 
+    
+    const su2double eddyVisc = nodes->GetmuT(iPoint)/flowNodes->GetDensity(iPoint);
+    const su2double lamVisc = nodes->GetLaminarViscosity(iPoint)/flowNodes->GetDensity(iPoint);
+
+    const su2double C_DES1 = 0.78;
+    const su2double C_DES2 = 0.61;
+
+    const su2double h_max = geometry->nodes->GetMaxLength(iPoint);
+    const su2double C_DES = C_DES1 * nodes->GetF1blending(iPoint) + C_DES2 * (1-nodes->GetF1blending(iPoint));
+    const su2double l_RANS = sqrt(nodes->GetSolution(iPoint, 0)) / (constants[6] * nodes->GetSolution(iPoint, 1));
+
+    su2double DES_lengthScale = 0.0;
+
+    switch(kind_hybridRANSLES){
+      case SST_DDES: {
+
+        const su2double r_d = (eddyVisc + lamVisc) / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double C_d1 = 20.0;
+        const su2double C_d2 = 3.0;
+
+        const su2double f_d = 1 - tanh(pow(C_d1 * r_d, C_d2));
+
+        const su2double l_LES = C_DES * h_max;
+        
+        DES_lengthScale = l_RANS - f_d * max(0.0, l_RANS - l_LES);
+
+        break;
+      }
+      case SST_IDDES: {
+        
+        // Constants
+        const su2double C_w = 0.15;
+        const su2double C_dt1 = 20.0;
+        const su2double C_dt2 = 3.0;
+        const su2double C_l = 5.0;
+        const su2double C_t = 1.87;
+
+        const su2double alpha = 0.25 - sqrt(wallDist2) / h_max;
+        const su2double f_b = min(2.0 * exp(-9.0 * alpha*alpha), 1.0);
+        const su2double r_dt = eddyVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double f_dt = 1 - tanh(pow(C_dt1 * r_dt, C_dt2));
+        const su2double ftilda_d = max(1.0 - f_dt, f_b);
+
+        const su2double r_dl = lamVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double f_l = tanh(pow(C_l*C_l*r_dl, 10.0));
+        const su2double f_t = tanh(pow(C_t*C_t*r_dt, 3.0));
+        const su2double f_e2 = 1.0 - max(f_t, f_l);
+        const su2double f_e1 = alpha >= 0 ? 2.0 * exp(-11.09*alpha*alpha) : 2.0 * exp(-9.0*alpha*alpha);
+        const su2double f_e = f_e2 * max((f_e1 - 1.0), 0.0);
+          
+
+        const su2double Delta = min(C_w * max(sqrt(wallDist2), h_max), h_max);
+        const su2double l_LES = C_DES * Delta;
+
+        DES_lengthScale = ftilda_d *(1.0+f_e)*l_RANS + (1.0 - ftilda_d) * l_LES;
+
+        break;
+      }
+      case SST_SIDDES: {
+        
+        // Constants
+        const su2double C_w = 0.15;
+        const su2double C_dt1 = 20.0;
+        const su2double C_dt2 = 3.0;
+        const su2double C_l = 5.0;
+        const su2double C_t = 1.87;
+
+        const su2double alpha = 0.25 - sqrt(wallDist2) / h_max;
+        const su2double f_b = min(2.0 * exp(-9.0 * alpha*alpha), 1.0);
+        const su2double r_dt = eddyVisc / (KolmConst2*wallDist2 * sqrt(0.5 * (StrainMag*StrainMag + VortMag*VortMag)));
+        const su2double f_dt = 1 - tanh(pow(C_dt1 * r_dt, C_dt2));
+        const su2double ftilda_d = max(1.0 - f_dt, f_b);
+
+        const su2double Delta = min(C_w * max(sqrt(wallDist2), h_max), h_max);
+        const su2double l_LES = C_DES * Delta;
+
+        DES_lengthScale = ftilda_d*l_RANS + (1.0 - ftilda_d) * l_LES;
+
+        break;
+      }
+    }
+
+    nodes->SetDES_LengthScale(iPoint, DES_lengthScale);
+
+  }
+  END_SU2_OMP_FOR
 }
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -1081,11 +1081,11 @@ void CTurbSSTSolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, C
 
   const auto kind_hybridRANSLES = config->GetKind_HybridRANSLES();
 
-
   auto* flowNodes = su2staticcast_p<CFlowVariable*>(solver[FLOW_SOL]->GetNodes());
 
   SU2_OMP_FOR_STAT(omp_chunk_size)
   for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
+
     const su2double StrainMag = max(flowNodes->GetStrainMag(iPoint), 1e-12);
     const su2double VortMag = max(GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint)), 1e-12);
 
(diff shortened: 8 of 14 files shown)


SU2_CFD/include/variables/CTurbSAVariable.hpp
```diff
@@ -87,4 +75,26 @@ class CTurbSAVariable final : public CTurbVariable {
...
+   */
+  inline su2double GetSSTVariables_omega(unsigned long iPoint) const { return Omega(iPoint); }
+
+  /*!
+   * \brief Set the value of the SST variables computed with SA solution.
+   * \param[in] val_k
+   * \param[in] val_Omega
+   */
+  void SetSSTVariables(unsigned long iPoint, su2double val_k, su2double val_Omega) {
+    k(iPoint) = val_k;
```
pcarruscag, line 96: Are these changes related to the main theme? or were just debugging? If they can be separated to another PR let's do that please.

externals/codi
pcarruscag, line 1: You need to undo this change to the codi version.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1563,6 +1597,27 @@ void CFlowOutput::LoadVolumeDataScalar(const CConfig* config, const CSolver* con
...
+      SetVolumeOutputValue("L_RANS", iPoint, Node_Turb->Get_L_RANS(iPoint));
+      SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
+      SetVolumeOutputValue("R_DT", iPoint, Node_Turb->Get_r_dt(iPoint));
+      SetVolumeOutputValue("R_DL", iPoint, Node_Turb->Get_r_dl(iPoint));
+    } else if ( config->GetKind_HybridRANSLES() == SST_SIDDES){
+      SetVolumeOutputValue("F_D", iPoint, Node_Turb->Get_ftilda_d(iPoint));
+      SetVolumeOutputValue("L_RANS", iPoint, Node_Turb->Get_L_RANS(iPoint));
+      SetVolumeOutputValue("L_LES", iPoint, Node_Turb->Get_L_LES(iPoint));
+      SetVolumeOutputValue("R_DT", iPoint, Node_Turb->Get_r_dt(iPoint));
+    }
```
pcarruscag, line 1616: Cleanup the repetition here, most statement are shared across the 3 cases

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1563,6 +1597,27 @@ void CFlowOutput::LoadVolumeDataScalar(const CConfig* config, const CSolver* con
...
+    const su2double mut = Node_Flow->GetEddyViscosity(iPoint);
+    const su2double mu = Node_Flow->GetLaminarViscosity(iPoint); 
+    const su2double LESIQ = 1.0/(1.0+0.05*pow((mut+mu)/mu, 0.53));
```
pcarruscag, line 1619: Are the constant here also used somewhere else? It would be good to avoid repeating hard-coded constants

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1647,6 +1666,46 @@ class CVariable {
...
+   */
+  inline virtual void AddVelLapl(unsigned long iPoint, su2double val_VelLapl_X, su2double val_VelLapl_Y) {}
+  /*!
+   * \brief Set the value of the value of FTrans.
+   */
+  inline virtual void AddVelLapl_Z(unsigned long iPoint, su2double val_VelLapl_Z) {}
+  /*!
+   * \brief Set the value of the value of FTrans.
+   */
+  inline virtual void SetVelLapl(unsigned long iPoint, su2double val_VelLapl_X, su2double val_VelLapl_Y) {}
```
pcarruscag, line 1703: Use a pointer for velocity and then you only need one of each function instead of handling xyz separately.

Common/include/option_structure.hpp
```diff
@@ -990,6 +990,8 @@ enum class SST_OPTIONS {
   V,           /*!< \brief Menter k-w SST model with vorticity production terms. */
   KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
   UQ,          /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+  SAS_SIMPLE,         /*!< \brief Menter k-w SST model with Scale Adaptive Simulations modifications. */
+  SAS_COMPLICATED,         /*!< \brief Menter k-w SST model with Scale Adaptive Simulations modifications. */
```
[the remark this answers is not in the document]
rois1995: No, these are just placeholders. There are no names in literature, that is why I named them according to the complexity of the model

Common/include/option_structure.hpp
```diff
@@ -1010,8 +1014,10 @@ static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
 struct SST_ParsedOptions {
   SST_OPTIONS version = SST_OPTIONS::V1994;   /*!< \brief Enum SST base model. */
   SST_OPTIONS production = SST_OPTIONS::NONE; /*!< \brief Enum for production corrections/modifiers for SST model. */
+  SST_OPTIONS sasModel = SST_OPTIONS::SAS_SIMPLE;   /*!< \brief Enum SST base model. */
   bool sust = false;                          /*!< \brief Bool for SST model with sustaining terms. */
   bool uq = false;                            /*!< \brief Bool for using uncertainty quantification. */
+  bool sas = false;                           /*!< \brief Bool for using Scale Adaptive Simulations. */
```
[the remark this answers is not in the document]
rois1995: Sure, I'll modify it.

Common/src/CConfig.cpp
```diff
@@ -6157,6 +6157,9 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
           case SA_DDES:  cout << "Delayed Detached Eddy Simulation (DDES) with Standard SGS" << endl; break;
           case SA_ZDES:  cout << "Delayed Detached Eddy Simulation (DDES) with Vorticity-based SGS" << endl; break;
           case SA_EDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_DDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_IDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
+          case SST_SIDDES: cout << "Delayed Detached Eddy Simulation (DDES) with Shear-layer Adapted SGS" << endl; break;
```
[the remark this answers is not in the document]
rois1995: Yes, I planned to modify the details on the final implementation. I'll do it now

Common/include/option_structure.hpp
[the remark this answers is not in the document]
pcarruscag: Maybe something after the authors since the models come from different papers.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -847,9 +852,53 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+        const su2double sigma_phi = 2.0/3.0;
+        const su2double C = 2.0;
+        const su2double C_s = 0.5; // Honestly I do not know if it is the right one.
+        const su2double gridSize = pow(Volume, 1.0/3.0);
+        // Scale of the modeled turbulence
+        const su2double L = sqrt(ScalarVar_i[0]) / (pow(beta_star, 0.25) * ScalarVar_i[1]);
+        // Von Karman Length Scale
+        su2double VelLaplMag = VelLapl_X*VelLapl_X + VelLapl_Y*VelLapl_Y;
+        if (nDim == 3) VelLaplMag += VelLapl_Z*VelLapl_Z;
+        const su2double L_vK_1 = KolmConst * StrainMag_i / sqrt(VelLaplMag);
```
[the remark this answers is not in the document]
rois1995: It is just for the source term

SU2_CFD/include/variables/CTurbSAVariable.hpp
rois1995: These are for computing the grid size for scale resolving simulations from an SA solution. Probably it has to be revised a little bit, since I found out that the formulation for the TKE is different if the QCR mod is used, thus I think I'll remove them for now.

SU2_CFD/src/output/CFlowOutput.cpp
rois1995: They are used just here

SU2_CFD/include/variables/CVariable.hpp
rois1995: Yeah, I'll try using the matrix type

Common/include/option_structure.hpp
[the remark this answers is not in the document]
rois1995: Sure, why not

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -136,6 +136,9 @@ class CNumerics {
   const su2double
   *TurbPsi_i,  /*!< \brief Vector of adjoint turbulent variables at point i. */
   *TurbPsi_j;  /*!< \brief Vector of adjoint turbulent variables at point j. */
+  su2double lengthScale_i, lengthScale_j; /*!< \brief length scale for SST */
+  su2double FTrans;   /*!< \brief SAS function */
+  su2double *VelLapl; /*!< \brief Laplacian of the velocity */
```
pcarruscag, line 141:
```suggestion
  const su2double *VelLapl; /*!< \brief Laplacian of the velocity */
```

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -707,6 +710,18 @@ class CNumerics {
...
 
+  /*!
+   * \brief Get the value of the value of FTrans.
+   */
+  inline virtual su2double GetFTrans() const { return 0.0; }
+
+  /*!
+   * \brief Get the value of the value of FTrans.
+   */
+  inline void SetVelLapl(su2double* val_VelLapl) {
```
pcarruscag, line 721:
```suggestion
  inline void SetVelLapl(const su2double* val_VelLapl) {
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1243,11 +1243,19 @@ void CFlowOutput::SetVolumeOutputFieldsScalarSolution(const CConfig* config){
...
 
     case TURB_FAMILY::KW:
       AddVolumeOutput("TKE", "Turb_Kin_Energy", "SOLUTION", "Turbulent kinetic energy");
       AddVolumeOutput("DISSIPATION", "Omega", "SOLUTION", "Rate of dissipation");
+      AddVolumeOutput("SRS_GRID_SIZE", "Srs_grid_size", "SOLUTION", "desired grid size for Scale Resolving Simulations");
+      if (config->GetSSTParsedOptions().sasModel == SST_OPTIONS::SAS_TRAVIS) AddVolumeOutput("FTRANS", "FTrans", "SOLUTION", "value of FTrans for SAS simulation");
+      if (config->GetSSTParsedOptions().sasModel == SST_OPTIONS::SAS_BABU){
+        AddVolumeOutput("VEL-LAPLACIAN_X", "Vel Laplacian x", "SOLUTION", "value of laplacian of x-velocity for SAS simulation");
+        AddVolumeOutput("VEL-LAPLACIAN_Y", "Vel Laplacian y", "SOLUTION", "value of laplacian of y-velocity for SAS simulation");
+        if (nDim == 3) AddVolumeOutput("VEL-LAPLACIAN_Z", "Vel Laplacian z", "SOLUTION", "value of laplacian of z-velocity for SAS simulation");
```
pcarruscag, line 1257: These outputs need to go somewhere else or they will break the restarts.
Solution is exclusively for solution variables, these are derived, put them where we have similar things (e.g. Q criterion)

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -188,8 +189,71 @@ void CTurbSSTSolver::Preprocessing(CGeometry *geometry, CSolver **solver_contain
...
+
+      /*--- Loop over the neighbors of point i. ---*/
+      for (auto jPoint : geometry->nodes->GetPoints(iPoint)) {
+
+        bool boundary_j = geometry->nodes->GetPhysicalBoundary(jPoint);
+
+        /*--- If iPoint is boundary it only takes contributions from other boundary points. ---*/
+        if (boundary_i && !boundary_j) continue;
+
+        /*--- Add solution differences, with correction for compressible flows which use the enthalpy. ---*/
```
pcarruscag, line 225: Check if comments need updates please

bigfooted:
Hi @rois1995  any progress? Would be very interesting to add this to develop, looking forward to using it actually.

rois1995:
> Hi @rois1995 any progress? Would be very interesting to add this to develop, looking forward to using it actually.

Hi @bigfooted, unfortunately I do not have enough time/computational power to perform a proper validation. I only tried something with a NACA0021 at 17 deg and 60 deg and they were promising, at least for the DDES implementation. The SAS ones are not that good. I'll upload some post-processing as soon as I have the time to do so.

Linnnnnn23:
I find your work really interesting. I've been studying the internal flow field in compressors and have had good results using SU2's SA_EDDES for calculating the cantilevered stator with a tip clearance. If you need help with code verification, I'd be glad to assist.

rois1995:
> I find your work really interesting. I've been studying the internal flow field in compressors and have had good results using SU2's SA_EDDES for calculating the cantilevered stator with a tip clearance. If you need help with code verification, I'd be glad to assist.

Hi @Linnnnnn23, every help on the validation/verification is gladly accepted! Let me know if you need anything by my side.

BerkeCan97:
Hi @rois1995 , I'd like to help with validation/verification if I can. Let me know if you are interested. You can contact me at berkecancfd@gmail.com.

BerkeCan97:
> > Hi @rois1995 any progress? Would be very interesting to add this to develop, looking forward to using it actually.
> 
> Hi @bigfooted, unfortunately I do not have enough time/computational power to perform a proper validation. I only tried something with a NACA0021 at 17 deg and 60 deg and they were promising, at least for the DDES implementation. The SAS ones are not that good. I'll upload some post-processing as soon as I have the time to do so.

Hi @rois1995, I believe there is an error or a typo in the Babu's paper in the Q_SAS source term and in turn in your implementation, which may be the reason for the poor performance in SAS results. 

image

This term changes units depending on the output of the max function, which doesn't make any sense. I believe the terms in the max function should be divided by omega^2 and k^2 instead of omega and k, respectively. Which is exactly how it is done in [https://resolver.tudelft.nl/uuid:5d23e2a6-5675-450d-bf3d-1dd40d736cae.](url)

I will try some of the benchmark cases in these papers when I have the time. Let me know what you think.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -918,9 +923,43 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+        // Scale of the modeled turbulence
+        const su2double L = sqrt(ScalarVar_i[0]) / (pow(beta_star, 0.25) * ScalarVar_i[1]);
+        // Von Karman Length Scale
+        const su2double VelLaplMag = GeometryToolbox::SquaredNorm(nDim, VelLapl);
+        const su2double L_vK_1 = KolmConst * StrainMag_i / sqrt(VelLaplMag);
+        const su2double L_vK_2 = C_s * sqrt(KolmConst * csi2 / (beta_blended/beta_star - alfa_blended)) * gridSize;
+        const su2double L_vK = max(L_vK_1, L_vK_2);
+        
+        const su2double gradTKE = GeometryToolbox::SquaredNorm(nDim, ScalarVar_Grad_i[0])/ScalarVar_i[0];
+        const su2double gradOmega = GeometryToolbox::SquaredNorm(nDim, ScalarVar_Grad_i[1])/ScalarVar_i[1];
```
BerkeCan97, line 946:
```suggestion
        const su2double gradOmega = GeometryToolbox::SquaredNorm(nDim, ScalarVar_Grad_i[1])/(ScalarVar_i[1]*ScalarVar_i[1]);
```
See https://resolver.tudelft.nl/uuid:5d23e2a6-5675-450d-bf3d-1dd40d736cae

BerkeCan97:
> hi @BerkeCan97, thank you for your correction and for helping with the verification part. Indeed, that term seems really wrong, thus I expect that your correction would improve the results. However, the link that you have provided for the corrected version is broken, can you fix it please?
> 
> Regarding the test-cases, we were trying to run the NACA0021 at 60deg AoA, but with relatively poor results. Maybe we should start with the flatplate and the mixing layer proposed originally in the implementation of the DDES formulation coupled to the SA model in https://www.researchgate.net/profile/Cleber-Spode-2/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2/links/59c07bfba6fdcca8e56fe3c3/Hybrid-RANS-LES-Calculations-in-SU2.pdf (or any other test-case in the paper you proposed, but I am not able to read that).

Fixed it! I think we can start with flat plate and maybe BFS. It might take a week or so for me to get anything done though.

BerkeCan97:
> > Fixed it! I think we can start with flat plate and maybe BFS. It might take a week or so for me to get anything done though.
> 
> Sorry but it still says "Error 404 page not found". Maybe I cannot access it. Can you share the title please?
> 
> A colleague of mine is working on the verification too on the flatplate, we will soon share the results and we will build on top of these ones.

Evaluation of the SST-SAS Model, L. Davidson

rois1995:
I think I got the flatplate covered, it should be easy (if the equations have been implemented correctly). The backward facing step and the diffuser I think are the most natural continuation as test cases, what do you think about it?

rguglielmi99:
I am currently working with @rois1995 on the flat plate case, taking as a reference result the skin friction coefficient computed using the coarse grid and the SA_DDES model (∆=∆max) from figure 3a in [https://www.researchgate.net/profile/Cleber-Spode-2/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2/links/59c07bfba6fdcca8e56fe3c3/Hybrid-RANS-LES-Calculations-in-SU2.pdf ](https://www.researchgate.net/profile/Cleber-Spode-2/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2/links/59c07bfba6fdcca8e56fe3c3/Hybrid-RANS-LES-Calculations-in-SU2.pd]. I ran the simulations using the already available DDES model for SA (SA_DDES, SA_ZDES,SA_EDDES) and the models under development for SST (SST_DDES,SST_IDDES,SST_SIDDES). The reference data were extracted using a plot digitalizer, so they may be not so accurate, particularly in the zoom region. I attach also the density residuals for all the simulations and the TKE residuals for the simulations conducted with SST. The options SST_IDDES and SST_DDES do not allow a complete convergence and this issue is associated to a region close to the inlet and the symmetry boundary conditions. Although seemingly the implementation of the models does not have a direct dependence on the boundary conditions, this issue is absent for SST_DDES. As evidence of this undesirable behavior, I attach the plot of the vertical velocity computed along a line parallel to the symmetry boundary (y=1e-6 , x between -0.33 and -0.05, with the flat plate beginning at x=0). I am currently working for understanding its cause.

Cf
Cf_zoom
Residual_rho
Residual_TKE
Issue region
Vertical_velocity

rguglielmi99:
I am currently working with @rois1995 on the flat plate case, taking as a reference result the skin friction coefficient computed using the coarse grid and the SA_DDES model (∆=∆max) from figure 3a in [Hybrid RANS-LES Calculations in SU2](https://www.researchgate.net/profile/Cleber-Spode-2/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2/links/59c07bfba6fdcca8e56fe3c3/Hybrid-RANS-LES-Calculations-in-SU2.pdf). I ran the simulations using the already available DDES model for SA (SA_DDES, SA_ZDES,SA_EDDES) and the models under development for SST (SST_DDES,SST_IDDES,SST_SIDDES). The reference data were extracted using a plot digitalizer, so they may be not so accurate, particularly in the zoom region. I attach also the density residuals for all the simulations and the TKE residuals for the simulations conducted with SST. The options SST_IDDES and SST_DDES do not allow a complete convergence and this issue is associated to a region close to the inlet and the symmetry boundary conditions. Although seemingly the implementation of the models does not have a direct dependence on the boundary conditions, this issue is absent for SST_DDES. As evidence of this undesirable behavior, I attach the plot of the vertical velocity computed along a line parallel to the symmetry boundary (y=1e-6 , x between -0.33 and -0.05, with the flat plate beginning at x=0). I am currently working for understanding its cause.

Cf
Cf_zoom
Residual_rho
Residual_TKE
Issue region
Vertical_velocity

rguglielmi99:
I am currently working with @rois1995 on the flat plate case, taking as a reference result the skin friction coefficient computed using the coarse grid and the SA_DDES model (∆=∆max) from figure 3a in [Hybrid RANS-LES Calculations in SU2](https://www.researchgate.net/profile/Cleber-Spode-2/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2/links/59c07bfba6fdcca8e56fe3c3/Hybrid-RANS-LES-Calculations-in-SU2.pdf). I ran the simulations using the already available DDES model for SA (SA_DDES, SA_ZDES,SA_EDDES) and the models under development for SST (SST_DDES,SST_IDDES,SST_SIDDES). The reference data were extracted using a plot digitalizer, so they may be not so accurate, particularly in zoom region. I attach also the density residuals for all the simulations and the TKE residuals for the simulations conducted with SST. The options SST_IDDES and SST_DDES do not allow a complete convergence and this issue is associated to a region close to the inlet and the symmetry boundary conditions. Although seemingly the implementation of the models does not have a direct dependence on the boundary conditions, this issue is absent for SST_DDES. As evidence of this undesirable and unexpected behavior, I attach the plot of the vertical velocity computed along a line parallel to the symmetry boundary (y=1e-6 , x between -0.33 and -0.05, with the flat plate beginning at x=0). I am currently working for understanding its cause.

Cf
Cf_zoom
Residual_rho
Residual_TKE
Issue region
Vertical_velocity

BerkeCan97:
That's strange indeed. Could you share your cfg and mesh files so I can also take a look at it?
Meanwhile I am waiting for the IT guys to compile it on my cluster but it is taking a while I guess...

rguglielmi99:
The reference config file for the simulations with SST model is [ddes_flatplate.txt](https://github.com/user-attachments/files/17720115/ddes_flatplate.txt). To switch between SST_DDES, SST_IDDES, SST_SIDDES I only modify the option HYBRID_RANSLES. The mesh I am using can be downloaded at [mesh](https://github.com/su2code/Tutorials/blob/master/compressible_flow/Turbulent_Flat_Plate/mesh_flatplate_turb_137x97.su2) .

BerkeCan97:
Hi @rguglielmi99 , I noticed that the mesh you are using is 2D. I don't think it makes sense to use a 2D grid in DES solutions because the Eddies are in itself 3D flow features. I do not think that's what causing a problem but it is worth a try.

rguglielmi99:
Hi @BerkeCan97,  thank you for the suggestion. It was an aspect I had already considered. However, I made the same choice as in the reference paper [Hybrid RANS-LES Calculations in SU2](https://www.researchgate.net/publication/318143234_Hybrid_RANSLES_Calculations_in_SU2
) where DDES with SA model are performed using 2D grids. There, it is assumed that the boundary layer is in quasi-2D mode, and therefore, 3D effects are considered negligible.

rois1995:
Are there any updates on the validation? On my side, I think that the SAS Travis implementation is wrong as the reference paper implements it for the k-w model, not for the SST one. Thus its formulation should be revised for that and it is consistently wrong as it is.

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -190,6 +190,10 @@ class CNumerics {
 
   bool bounded_scalar = false;    /*!< \brief Flag for bounded scalar problem */
 
+  su2double lengthScale_i, lengthScale_j;
+  su2double FTrans;   /*!< \brief SAS function */
+  su2double VelLapl_X, VelLapl_Y, VelLapl_Z;
```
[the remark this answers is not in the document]
rois1995: I think the main problem with my implementation is that the norm of the velocity laplacian stated in the paper is $||U''||=\sqrt{\sum_i \left( \frac{\partial^2 U_i}{\partial x_j \partial x_j}\right)^2}$ and I am not sure that this is what I am computing. Do I need to store a matrix for each velocity component, compute it as gradient of the gradient and then sum the traces of each matrix?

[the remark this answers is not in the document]
pcarruscag: That looks like the norm of the Laplacian of the velocity components, the way you are computing it is not very accurate but maybe it is good enough. You could do it the finite volume way by reusing the numerics classes we use for scalar viscous fluxes.

BerkeCan97:
I couldn't find find the time to do it yet but I think I can find some free time in a few weeks. I'll post an update if when I can.

rois1995:
I have been working on the validation on the NACA0021 at $\alpha = 60^\circ$ case. I ran incompressible simulations with the JST scheme as  $Ma = 0.012$. I tried 2 meshes, a coarse and a medium one (I also have a fine mesh but it will take a lot to get results on that one and I do not think that it will be necessary). I tried just the SST-DDES model as the SAS model of Babu is not giving good results. All simulations have been run for 505 CUs.

The two grids have the following setups:

|Mesh | Coarse|Medium|
|----|------|--|
|Num. of points chordwise |100|200|
|Num. of points spanwise |71|142|
|Num. of points normal |62|121|
|dt|0.0137s|0.00685s|

The spanwise dimension is 4 chords. The farfield domain is set at a distance of 20c from the wing.

# Results

FFT
Cp

|Case | Coarse | Coarse - dt/2|Medium| Exp.|
|----|------|--|--|--|
|mean $C_L$ | 1.47 | 1.49|0.9|0.931|
|mean $C_D$ | 2.64 | 2.73|1.48| 1.517|
|$C_L$ peaks| [(0.07), 0.144, 0.211] | [(0.07), 0.144, 0.211] |[0.187, 0.393]|[0.197, 0.396]|
|$C_D$ peaks |[(0.07), 0.144, 0.211] | [(0.07), 0.144, 0.211] |[0.187, 0.394]|[0.195, 0.388]|

The medium grid seems to be necessary to correctly capture the flow-physics of the problem, both in terms of signal spectra and mean $c_P$ distribution over the wing.

However, in the [article where the DDES where introduced in SU2](https://doi.org/10.2514/6.2017-4284), it is stated that the Roe scheme with low-dissipation correction was used, thus the case was run with the compressible solver. I have tried with the Roe and the L2Roe scheme but neither of them worked as I had problems at the periodic boundaries. I was able to obtain some results switching to the JST scheme. Unfortunately I wrongly set the 4-th coefficient to 0, whereas the other simulations where with the standard value of 0.02. I am now re-running with the standard value. Nevertheless, the results are the following:

FFT_Comp-Incomp
Cp_Comp-Incomp

Slightly better results on the coarse grid have been achieved with a compressible simulation. The FFT peaks are only two but, again, are too low in frequency compared to the experimental results. Adding to the previous table:

|Case | Coarse | Coarse - dt/2|Coarse - Comp.| Medium|Exp.|
|----|------|--|--|--|--|
|mean $C_L$ | 1.47 | 1.49|1.14|0.9|0.931|
|mean $C_D$ | 2.64 | 2.73|1.82|1.48| 1.517|
|$C_L$ peaks| [(0.07), 0.144, 0.211] | [(0.07), 0.144, 0.211]| [(0.07), 0.144, 0.211] |[0.187, 0.393]|[0.197, 0.396]|
|$C_D$ peaks |[(0.07), 0.144, 0.211] | [(0.07), 0.144, 0.211]| [(0.07), 0.144, 0.211] |[0.187, 0.394]|[0.195, 0.388]|
  
Also, the $c_P$ distribution is much better with the compressible simulation for the same grid level. 

It is not clear to me how in the aforementioned paper very good results were achieved even on the coarser grid. If we compare their results, it seems that my incompressible simulations on the coarse grid essentially revert to a URANS simulation. Further investigations are necessary, but a first guess is that their coarse mesh is actually like my medium one but with a smaller spanwise extension. I will try with their mesh also.

Next steps: 
- perform compressible simulations on the coarse and medium grid with 4-th order coefficient of JST scheme set to std. value.
- revert to the BFS test case which may be easier to run.

pcarruscag:
What are the issues with symmetry planes? Have you tried with periodic boundaries instead?

rois1995:
Yes, I have used periodic boundaries. However, with the Roe scheme I end up with large values at the wall, which may be related either to the grid or to the periodic boundaries. You can see a snapshot here. The simulation diverges after 6 timesteps.

Periodic

Whereas here I am plotting the pressure coefficient distribution along the leading edge line. The line for t=1 refers to the steady computation peformed with JST scheme (otherwise I would've had divergence problems). Other timesteps have been computed with the Roe scheme.

CpGraph

We can see how the $C_P$ is not constant in the spanwise direction and changes a lot when reaching the periodic boundaries (y = 0 and y = 4). In the same graph we can see the steady incompressible $C_P$ distribution and we can see how it still has problems at the periodic boundaries but the deviation ins much smaller.

pcarruscag:
Maybe there are terms in the DES formulation that don't do well on symmetries, we have a recent example of a mesh of this type with RANS and it seemed fine.

rois1995:
Unfortunately also the steady RANS have that problem, which worsen when switching from the Incompressible solver to the Compressible one.

pcarruscag:
Have you made that case with the symmetry issues available somewhere?

rois1995:
> Have you made that case with the symmetry issues available somewhere?

Are you referring to the one in #2390?

pcarruscag:
That's not a naca0012

rois1995:
Ok but why should I do that? I have periodic boundaries here.

pcarruscag:
You mentioned problems at the periodic boundaries
> I have tried with the Roe and the L2Roe scheme but neither of them worked as I had problems at the periodic boundaries.

rois1995:
I am actually quite unsure of the implementation of the EDDES for SA. I think that the correct implementation  is the one that I have used in the SST model (they are equivalent).

SA Implementation
```
su2double ln_max = 0.0;
for (const auto jPoint : geometry->nodes->GetPoints(iPoint)) {
  const auto coord_j = geometry->nodes->GetCoord(jPoint);
  su2double delta[MAXNDIM] = {};
  for (auto iDim = 0u; iDim < nDim; iDim++){
    delta[iDim] = fabs(coord_j[iDim] - coord_i[iDim]);
  }
  su2double ln[3];
  ln[0] = delta[1]*ratioOmega[2] - delta[2]*ratioOmega[1];
  ln[1] = delta[2]*ratioOmega[0] - delta[0]*ratioOmega[2];
  ln[2] = delta[0]*ratioOmega[1] - delta[1]*ratioOmega[0];
  const su2double aux_ln = sqrt(ln[0]*ln[0] + ln[1]*ln[1] + ln[2]*ln[2]);
  ln_max = max(ln_max, aux_ln);
  vortexTiltingMeasure += nodes->GetVortex_Tilting(jPoint);
}
vortexTiltingMeasure /= (nNeigh + 1);
```

SST Implementation
```
for (const auto jPoint : geometry->nodes->GetPoints(iPoint)){
  const auto coord_j = geometry->nodes->GetCoord(jPoint);

  for (const auto kPoint : geometry->nodes->GetPoints(iPoint)){
    const auto coord_k = geometry->nodes->GetCoord(kPoint);

    su2double delta[MAXNDIM] = {};
    // This should only be performed on 3D cases anyway
    for (auto iDim = 0u; iDim < 3; iDim++){
      delta[iDim] = (coord_j[iDim] - coord_k[iDim])/2.0; // Should I divide by 2 as I am interested in the dual volume?
    }
    su2double l_n_minus_m[3];
    GeometryToolbox::CrossProduct(delta, vorticityDir, l_n_minus_m);
    deltaOmega = max(deltaOmega, GeometryToolbox::Norm(nDim, l_n_minus_m));
  }

  // Add to VTM(iPoint) to perform the average
  vortexTiltingMeasure += nodes->GetVortex_Tilting(jPoint);
}
deltaOmega /= (1/sqrt(3));
vortexTiltingMeasure /= (nNeigh+1);
```

Proposed [here](https://link.springer.com/article/10.1007/S10494-015-9618-0) 
Computation

rois1995:
# Backward Facing Step

The test case is the one in the [NASA TMR page](https://turbmodels.larc.nasa.gov/backstep_val.html). 

$Re_H = 36000$
$Ma = 0.128$
$T_\infty= 537 \text{R}$

The grid is taken from the Level 3 among the 2D CGNS grids given [here](https://turbmodels.larc.nasa.gov/backstep_grids.html). Then it has been extended in the spanwise direction for 4H and with 40 points. The timestep is $\Delta t = 0.02 H / U_\infty$, the total simulation time is $t_{tot} = 300 H/U_\infty$ and the results have been averaged on the last 270 CUs.  The JST scheme has been used. 

## Velocity profiles in the channel

FirstLinePlots
SecondLinePlots

The SST-DDES model is lacking accuracy across all the velocity profiles, not really clear why since the results for the NACA0021 airfoil were satisfactory. The SST-EDDES and SA-EDDES solutions are similar, but the SST implementation is even better when the velocity fluctuations are considered.

## Reynolds stresses in the channel

Reference from [here](https://doi.org/10.1115/1.4052019):
ReynoldsStressesPaper

Results:
ReynoldStresses_3

The Reynolds stresses are consistent between the SA-EDDES and SST-EDDES formulations and are quite in line with the reference solutions (although they have used a finer grid). The SST-DDES results are the ones that deviates a lot, as also seen in the velocity profiles above.

rois1995:
## BFS Update

I have run the simulations with the SST-IDDES and SST-SIDDES also for the coarse mesh of the BFS case. Results are quite good compared to SST-EDDES simulations. Only the SST-DDES case seems to have the worst results among all of them.

# Velocity Profiles in the channel
FirstLinePlots
SecondLinePlots

# Reynolds Stresses
ReynoldStresses_3

I am currently running the Medium grid only for the SST-EDDES case, but it will take a long time before having meaningful results.

Up to know I haven't found a way of making the SAS model to work. There is something wrong with the computation of the Von-Karman length scale, which can be directly related to a wrong computation of the laplacian of the velocity. If anyone is interested in investigating this I think it will be material for another PR.

I will remove the WIP from this PR.

rois1995:
## Backward Facing Step

I have re-run the SST-DDES and SST-EDDES simulations with the bug fixed from #2484 and a minor bug solved for SST-EDDES.

FirstLinePlots
SecondLinePlots

The fix seems to be slightly improving results for SST-DDES and  worsening the results for SST-EDDES. However, what is happening for the EDDES model is that it is much more aligned with the SA-EDDES results with the fix included, which I think is desired as the LES formulation is essentially the same.

FirstLinePlots
SecondLinePlots


I have also tried changing the computation of the vorticity-aligned cell surface for the SA-EDDES model as it is supposed to be implemented (formulation shown in previous comments). The results are not much different for the BFS case but I think that the new (correct) formulation should be preferred.

FirstLinePlots
SecondLinePlots

I will try some more with the SAS formulation, if nothing works then I'll remove it from this PR.

bigfooted:
are there simulation results from other codes for this case available for comparison? So we know if the current performance is as good as it gets or not?

rois1995:
It seems that the problem is encountered on skew elements. I have tried running the SA-EDDES simulation for 30 timesteps with only one inner iteration (the problem remains when running for one temporal iteration with 30 inner iterations).

Periodic boundary
periodic

Center plane
center

The same problem remains with SA-URANS simulation.

Periodic boundary SA-URANS
periodic

rois1995:
Hi @pcarruscag, I am not planning on running other tests for now, I think that the results are good enough. For me, it is ready for review. Should I remove the debug variables/outputs before reviewing?

Common/include/option_structure.hpp
```diff
@@ -995,6 +995,8 @@ enum class SST_OPTIONS {
   COMP_Wilcox, /*!< \brief Menter k-w SST model with Compressibility correction of Wilcox. */
   COMP_Sarkar, /*!< \brief Menter k-w SST model with Compressibility correction of Sarkar. */
   DLL,         /*!< \brief Menter k-w SST model with dimensionless lower limit clipping of turbulence variables. */
+  SAS_TRAVIS,         /*!< \brief Menter k-w SST model with Scale Adaptive Simulations modifications. */
+  SAS_BABU,         /*!< \brief Menter k-w SST model with Scale Adaptive Simulations modifications. */
```
pcarruscag: Formatting

Common/include/option_structure.hpp
```diff
@@ -1062,6 +1067,15 @@ inline SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS *SST_Options, unsigne
   const bool sst_compWilcox = IsPresent(SST_OPTIONS::COMP_Wilcox);
   const bool sst_compSarkar = IsPresent(SST_OPTIONS::COMP_Sarkar);
   const bool sst_dll = IsPresent(SST_OPTIONS::DLL);
+  const bool sst_sas_travis = IsPresent(SST_OPTIONS::SAS_TRAVIS);
+  const bool sst_sas_babu = IsPresent(SST_OPTIONS::SAS_BABU);
+  if (sst_sas_travis && sst_sas_babu) {
+    SU2_MPI::Error("Two versions (Simple and Complicated) selected for SAS under SST_OPTIONS. Please choose only one.", CURRENT_FUNCTION);
```
pcarruscag: Update simple and complicated?

Common/include/option_structure.hpp
```diff
@@ -1476,14 +1490,24 @@ enum ENUM_HYBRIDRANSLES {
   SA_DES   = 1,          /*!< \brief Kind of Hybrid RANS/LES (SA - Detached Eddy Simulation (DES)). */
   SA_DDES  = 2,          /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Delta_max SGS ). */
   SA_ZDES  = 3,          /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Vorticity based SGS like Zonal DES). */
-  SA_EDDES = 4           /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Shear Layer Adapted SGS: Enhanced DDES). */
+  SA_EDDES = 4,           /*!< \brief Kind of Hybrid RANS/LES (SA - Delayed DES (DDES) with Shear Layer Adapted SGS: Enhanced DDES). */
+  SST_DDES = 5,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): DDES). */
+  SST_IDDES = 6,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): Improved DDES). */
+  SST_SIDDES = 7,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): Simplified Improved DDES). */
+  SST_EDDES = 8,           /*!< \brief Kind of Hybrid RANS/LES (SST - Delayed DES (DDES): Enhanced (SLA) DDES). */
```
pcarruscag: formatting

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -136,6 +136,14 @@ class CNumerics {
   const su2double
   *TurbPsi_i,  /*!< \brief Vector of adjoint turbulent variables at point i. */
   *TurbPsi_j;  /*!< \brief Vector of adjoint turbulent variables at point j. */
+  su2double lengthScale_i, lengthScale_j; /*!< \brief length scale for SST */
+  su2double FTrans;   /*!< \brief SAS function */
+  su2double Q_SAS_1;   /*!< \brief SAS function */
+  su2double Q_SAS_2;   /*!< \brief SAS function */
+  su2double L;   /*!< \brief SAS function */
+  su2double L_vK_1;   /*!< \brief SAS function */
+  su2double L_vK_2;   /*!< \brief SAS function */
```
pcarruscag: If these were here for debugging, please use local variables in the function where these quantities are used.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -893,6 +895,9 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
       /*--- Dissipation ---*/
 
       su2double dk = beta_star * Density_i * ScalarVar_i[1] * ScalarVar_i[0] * (1.0 + zetaFMt);
+      if (config->GetKind_HybridRANSLES() != NO_HYBRIDRANSLES)
+        dk = Density_i * sqrt(ScalarVar_i[0]*ScalarVar_i[0]*ScalarVar_i[0]) / lengthScale_i;
```
pcarruscag:
```suggestion
        dk = Density_i * sqrt(pow(ScalarVar_i[0], 3)) / lengthScale_i;
```

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -906,9 +911,46 @@ class CSourcePieceWise_TurbSST final : public CNumerics {
...
+        // Scale of the modeled turbulence
+        L = sqrt(ScalarVar_i[0]) / (pow(beta_star, 0.25) * ScalarVar_i[1]);
+        // Von Karman Length Scale
+        const su2double VelLaplMag = GeometryToolbox::Norm(nDim, VelLapl);
+        L_vK_1 = KolmConst * StrainMag_i / VelLaplMag;
+        L_vK_2 = C_DES * sqrt(KolmConst * csi2 / (beta_blended/beta_star - alfa_blended)) * gridSize;
+        const su2double L_vK = max(L_vK_1, L_vK_2);
+        
+        const su2double gradTKE = GeometryToolbox::SquaredNorm(nDim, ScalarVar_Grad_i[0])/(ScalarVar_i[0]*ScalarVar_i[0]);
+        const su2double gradOmega = GeometryToolbox::SquaredNorm(nDim, ScalarVar_Grad_i[1])/(ScalarVar_i[1]*ScalarVar_i[1]);
```
pcarruscag: pow(x, 2)

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1664,6 +1683,93 @@ class CVariable {
...
+
+  /*!
+   * \brief Incrementally add the velocity laplacian vector.
+   */
+  inline virtual void AddVelLapl(unsigned long iPoint, su2double val_VelLapl_X, su2double val_VelLapl_Y, su2double val_VelLapl_Z) {}
+  
+  /*!
+   * \brief Set the value of the velocity laplacian.
+   */
+  inline virtual void SetVelLapl(unsigned long iPoint, su2double* val_VelLapl) {}
```
pcarruscag: Can you use fewer modes of accessing the Laplacian? I.e. just return pointer or just use (iPoint, iDim)?

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -1712,6 +1818,18 @@ class CVariable {
...
+   * \brief Get the value of the desired grid size for Scale Resolving Simulations.
+    * \return the value of the desired grid size for Scale Resolving Simulations.
+  */
+  inline virtual su2double GetSRSGridSize(unsigned long iPoint) const { return 0.0; }
+
+  /*!
+   * \brief Set the value of the desired grid size for Scale Resolving Simulations.
+   * \param[in] val_gridSize
+   */
+  inline virtual void SetSRSGridSize(unsigned long iPoint, su2double val_gridSize) {}
```
pcarruscag: If this is one of the debug quantities you are removing please move these accessors to the turbulence variable (thus not virtual anymore).

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -968,7 +974,75 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
+                  }
+                }
+              }
+            }
+
+            /*--- Store the components to be communicated in the buffer. ---*/
+
+            for (iDim = 0; iDim < nDim; iDim++)
+              bufDSend[buf_offset+iDim] = Vel_Lapl[iDim];
+
```
pcarruscag: This should be the "else" to "if rotate_periodic", it changes the same values.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -968,7 +974,75 @@ void CSolver::InitiatePeriodicComms(CGeometry *geometry,
...
+                size_t iEdge = geometry->nodes->GetEdge(iPoint, iNeigh);
+                weight = halfOnVol;
+
+                const auto area = geometry->edges->GetNormal(iEdge);
+                AD::SetPreaccIn(area, nDim);
+
+                const auto coordsJPoint = geometry->nodes->GetCoord(jPoint);
+
+                su2double I2JVec[3] = {0.0};
+                for (iDim = 0; iDim < nDim; ++iDim) I2JVec[iDim] = coordsJPoint[iDim] - coordsIPoint[iDim];
```
pcarruscag: GeoToolbox::Distance

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1574,6 +1575,62 @@ void CTurbSASolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CC
...
+        for (const auto jPoint : geometry->nodes->GetPoints(iPoint)){
+          const auto coord_j = geometry->nodes->GetCoord(jPoint);
+
+          for (const auto kPoint : geometry->nodes->GetPoints(iPoint)){
+            const auto coord_k = geometry->nodes->GetCoord(kPoint);
+
+            su2double delta[MAXNDIM] = {};
+            // This should only be performed on 3D cases anyway
+            for (auto iDim = 0u; iDim < 3; iDim++){
+              delta[iDim] = (coord_j[iDim] - coord_k[iDim])/2.0; // Should I divide by 2 as I am interested in the dual volume?
```
pcarruscag: Please turn some of these questions into `TODO: ` for things you are not sure, or comments stating a conclusion.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1574,6 +1575,62 @@ void CTurbSASolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CC
...
+         Flow Turbulence Combust - 2015
+         ---*/
+
+        su2double vortexTiltingMeasure = nodes->GetVortex_Tilting(iPoint);
+
+        const su2double omega = GeometryToolbox::Norm(3, vorticity);
+
+        su2double ratioOmega[MAXNDIM] = {};
+
+        for (auto iDim = 0; iDim < 3; iDim++){
```
pcarruscag: replace the 3's by MAXNDIM

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -1574,6 +1575,62 @@ void CTurbSASolver::SetDES_LengthScale(CSolver **solver, CGeometry *geometry, CC
...
+        }
+        
+        vortexTiltingMeasure /= (nNeigh + 1);
+
+        const su2double f_kh = max(f_min,
+                                   min(f_max,
+                                       f_min + ((f_max - f_min)/(a2 - a1)) * (vortexTiltingMeasure - a1)));
+
+        const su2double r_d = (kinematicViscosityTurb+kinematicViscosity)/(uijuij*k2*pow(wallDistance, 2.0));
+        const su2double f_d = 1.0-tanh(pow(8.0*r_d,3.0));
```
pcarruscag:
```suggestion
        const su2double r_d = (kinematicViscosityTurb+kinematicViscosity)/(uijuij*k2*pow(wallDistance, 2));
        const su2double f_d = 1.0-tanh(pow(8.0*r_d, 3));
```

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -28,6 +28,7 @@
 #include "../../include/solvers/CTurbSSTSolver.hpp"
 #include "../../include/variables/CTurbSSTVariable.hpp"
 #include "../../include/variables/CFlowVariable.hpp"
+#include "../../include/variables/CMeshVariable.hpp"
```
pcarruscag: Is this used?

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -197,8 +198,103 @@ void CTurbSSTSolver::Preprocessing(CGeometry *geometry, CSolver **solver_contain
...
+        size_t iEdge = geometry->nodes->GetEdge(iPoint, iNeigh);
+        size_t jPoint = geometry->nodes->GetPoint(iPoint, iNeigh);
+
+        /*--- Determine if edge points inwards or outwards of iPoint.
+        *    If inwards we need to flip the area vector. ---*/
+
+        const su2double weight = halfOnVol;
+
+        const auto area = geometry->edges->GetNormal(iEdge);
+        AD::SetPreaccIn(area, nDim);
```
pcarruscag: This is not doing anything because you are not starting/ending the preaccumulation.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -1019,3 +1157,172 @@ void CTurbSSTSolver::SetUniformInlet(const CConfig* config, unsigned short iMark
...
+
+    const su2double h_max = geometry->nodes->GetMaxLength(iPoint);
+    const su2double C_DES = C_DES1 * nodes->GetF1blending(iPoint) + C_DES2 * (1-nodes->GetF1blending(iPoint));
+    const su2double l_RANS = sqrt(nodes->GetSolution(iPoint, 0)) / (constants[6] * nodes->GetSolution(iPoint, 1));
+
+    su2double DES_lengthScale = 0.0;
+
+    switch(kind_hybridRANSLES){
+      case SST_DDES: {
+
```
pcarruscag, line 943: Can you add some links to the references you used for each formulation? (here or somewhere you think is more appropriate)

SU2_CFD/src/variables/CTurbVariable.cpp
```diff
@@ -35,8 +35,61 @@ CTurbVariable::CTurbVariable(unsigned long npoint, unsigned long ndim, unsigned
...
+  StrainDotVort[2] = Strain[2][0]*Vorticity[0]+Strain[2][1]*Vorticity[1]+Strain[2][2]*Vorticity[2];
+
+  numVecVort[0] = StrainDotVort[1]*Vorticity[2] - StrainDotVort[2]*Vorticity[1];
+  numVecVort[1] = StrainDotVort[2]*Vorticity[0] - StrainDotVort[0]*Vorticity[2];
+  numVecVort[2] = StrainDotVort[0]*Vorticity[1] - StrainDotVort[1]*Vorticity[0];
+
+  numerator = sqrt(6.0) * sqrt(numVecVort[0]*numVecVort[0] + numVecVort[1]*numVecVort[1] + numVecVort[2]*numVecVort[2]);
+  trace0 = 3.0*(pow(Strain[0][0],2.0) + pow(Strain[1][1],2.0) + pow(Strain[2][2],2.0));
+  trace1 = pow(Strain[0][0] + Strain[1][1] + Strain[2][2],2.0);
+  denominator = pow(Omega, 2.0) * sqrt(trace0-trace1);
```
pcarruscag: Geometry toolbox where possible please

SU2_CFD/src/variables/CTurbVariable.cpp
```diff
@@ -35,8 +35,61 @@ CTurbVariable::CTurbVariable(unsigned long npoint, unsigned long ndim, unsigned
...
+  Strain[0][0] = PrimGrad_Flow[1][0];
+  Strain[1][0] = 0.5*(PrimGrad_Flow[2][0] + PrimGrad_Flow[1][1]);
+  Strain[0][1] = 0.5*(PrimGrad_Flow[1][1] + PrimGrad_Flow[2][0]);
+  Strain[1][1] = PrimGrad_Flow[2][1];
+  if (nDim == 3){
+    Strain[0][2] = 0.5*(PrimGrad_Flow[3][0] + PrimGrad_Flow[1][2]);
+    Strain[1][2] = 0.5*(PrimGrad_Flow[3][1] + PrimGrad_Flow[2][2]);
+    Strain[2][0] = 0.5*(PrimGrad_Flow[1][2] + PrimGrad_Flow[3][0]);
+    Strain[2][1] = 0.5*(PrimGrad_Flow[2][2] + PrimGrad_Flow[3][1]);
+    Strain[2][2] = PrimGrad_Flow[3][2];
```
pcarruscag, line 68: Don't we have a helper function for this somewhere? Maybe in the viscous fluxes?

SU2_CFD/src/variables/CTurbVariable.cpp
```diff
@@ -35,8 +35,61 @@ CTurbVariable::CTurbVariable(unsigned long npoint, unsigned long ndim, unsigned
...
+
+  AD::StartPreacc();
+  AD::SetPreaccIn(PrimGrad_Flow, nDim+1, nDim);
+  AD::SetPreaccIn(Vorticity, 3);
+  /*--- Eddy viscosity ---*/
+  AD::SetPreaccIn(muT(iPoint));
+  /*--- Laminar viscosity --- */
+  AD::SetPreaccIn(LaminarViscosity);
+
+  Strain[0][0] = PrimGrad_Flow[1][0];
```
pcarruscag, line 59: We need to abstract the index for velocity somehow, otherwise this is not compatible with the NEMO solver.
Maybe change PrimGrad_Flow to VelGrad and whoever call this function is responsible by passing the correct part of the gradients.

pcarruscag (commented):
Nice additions :+1:

rois1995:
I will definitively add a V&V case, although I do not have time at the moment. I will also review the suggestions as soon as possible.

SU2_CFD/include/variables/CVariable.hpp
rois1995: I cannot remove it from CVariable otherwise it is not recognized in the CFlowOutput. Is there a way to fix this?

SU2_CFD/include/variables/CVariable.hpp
rois1995: I am just removing everything related to SAS implementation, as it was not working.

SU2_CFD/src/variables/CTurbVariable.cpp
rois1995: I see that there is a ComputeMeanRateOfStrainMatrix in CNumerics, I will use that.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: If you want it for output you can keep the Get virtual but remove the Set for example.
If the accessor is only used in the TurbSolver you can save all the virtual boilerplate and have it only in TurbVariable.

rois1995:
Should we allow DDES to be run on 2D meshes? The DDES flatplate testcase has a 2D mesh, but it does not really make any sense to run 2D cases with DDES.

pcarruscag:
If someone is determined enough they'll make a 2.5D mesh and still run it :smile: 
Maybe you can just print a big warning but allow it to run? It could be useful for regression tests.
Is there anything in the implementation that only works for 3D?

rois1995:
I am pretty sure that the EDDES models have something that does not work in 2D, like the vortex-tilting measure (depends on the scalar product of the gradient of the velocity and the vorticity, thus in 2D it should be zero).

bigfooted:
@rois1995 @pcarruscag should we finalize this before the conflicts get out of control? If time is the issue, I propose we only add a regression test for this PR. 
I can support with the V&V if you can share a testcase, or we can find a student to do this.

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -186,15 +186,24 @@ void CTurbSASolver::Preprocessing(CGeometry *geometry, CSolver **solver_containe
...
       for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
+        su2double **Grad_Vel = new su2double* [nDim];
+        su2double **StrainMat = new su2double* [nDim];
         auto Vorticity = flowNodes->GetVorticity(iPoint);
-        auto PrimGrad_Flow = flowNodes->GetGradient_Primitive(iPoint);
+        for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+          Grad_Vel[iDim] = new su2double [nDim];
+          StrainMat[iDim] = new su2double [nDim];
+          for (unsigned short jDim = 0; jDim < nDim; jDim++) {
+            Grad_Vel[iDim][jDim] = nodes->GetGradient_Primitive(iPoint, prim_idx.Velocity() + iDim, jDim);
```
pcarruscag, line 201: Memory leak.
Use something with static sizes instead of allocating with `new`

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -199,8 +199,43 @@ void CTurbSSTSolver::Preprocessing(CGeometry *geometry, CSolver **solver_contain
...
+      for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++){
+        su2double **Grad_Vel = new su2double* [nDim];
+        su2double **StrainMat = new su2double* [nDim];
+        auto Vorticity = flowNodes->GetVorticity(iPoint);
+        for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+          Grad_Vel[iDim] = new su2double [nDim];
+          StrainMat[iDim] = new su2double [nDim];
+          for (unsigned short jDim = 0; jDim < nDim; jDim++) {
+            Grad_Vel[iDim][jDim] = nodes->GetGradient_Primitive(iPoint, prim_idx.Velocity() + iDim, jDim);
+          }
```
pcarruscag, line 224: same

rois1995:
Sorry for the long silence on this PR, I could not work on it for a while. I have now brought it up to date:

- Merged `develop` and resolved the conflicts with the stochastic backscatter model: the DES length scale and the vortex tilting measure are in `CTurbVariable`, shared by SA and SST.
- Fixed a memory leak in the EDDES preprocessing: the strain matrices were allocated with `new` for every point at every iteration and never freed, so long runs ran out of memory. The velocity gradient was also read from the turbulence variables instead of the flow variables.
- Fixed `SST_EDDES`, which fell through to `SST_EDDES_UNSTR` (missing `break`) and diverged from a uniform initial flow (no vorticity, so Δ̃_ω = 0).
- Checked the implementation against Gritskevich et al. (2012) and Shur et al. (2015). SST DDES/IDDES/SIDDES match. For the SLA length scale I fixed the vortex tilting measure, the F_KH limiter and the structured Δ̃_ω, also for the existing `SA_EDDES`, so the `ddes_flatplate` reference values changed.
- Hybrid RANS/LES now requires `TIME_DOMAIN= YES`.
- Removed the SAS variants, which were not working.

I updated the title and the description. Next I will add regression tests for the new options and set up the validation cases.

bigfooted:
Great! Can you add a small tutorial for the website as well? We have a lot of these implementations that are not exposed to the public.
A couple of pictures with an explanation of how to set up a testcase would be sufficient.

PR #2152, opened 2023-10-27
Solid to solid conjugate heat transfer

EvertBunschoten:
## Proposed Changes
Allow for conjugate heat transfer between solid domains and the modeling of contact resistance. 

## PR Checklist

- [x] Configuration option for contact resistance at interfaces.
- [x] Communicate contact resistance values between config and CInterface.
- [x] Test grid independence of contact resistance model.
- [x] Test case with both solid-solid and solid-fluid.
- [x] Add test case to regression tests.
- [x] Update documentation.


pcarruscag (commented):
I get the idea but from the numerical standpoint this is an even worse idea than the way we do fluid-fluid interfaces.
Additive decomposition of elliptic equations is not effective at all.
Are you looking to model contact between multiple solid materials?

EvertBunschoten:
> I get the idea but from the numerical standpoint this is an even worse idea than the way we do fluid-fluid interfaces. Additive decomposition of elliptic equations is not effective at all. Are you looking to model contact between multiple solid materials?

Yes, I'm looking to model the transfer of heat between multiple solid zones with different thermal properties. Eventually I'd like to model contact resistance of such interfaces as well, where the user could specify the contact resistance value of each solid-to-solid interface. This is something that me and Praharsh will be working on.

pcarruscag:
Okay, then allow me to save you guys some time, implement something similar to what we use for the FEA solver.
See the use of `element_properties` in CFEASolver.cpp.

EvertBunschoten:
> Okay, then allow me to save you guys some time, implement something similar to what we use for the FEA solver. See the use of `element_properties` in CFEASolver.cpp.

Thanks for the suggestion! We'll have a look.

NAnand-TUD:
Good job @EvertBunschoten. If you want we can schedule a call together and plan the development a bit.

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -133,7 +133,10 @@
         (donor_config->GetKind_CHT_Coupling() == CHT_COUPLING::AVERAGED_TEMPERATURE_ROBIN_HEATFLUX)) {
 
       const su2double rho_cp_solid = donor_config->GetSpecific_Heat_Cp()*donor_config->GetMaterialDensity(0);
-      conductivity_over_dist  = thermal_diffusivity*rho_cp_solid/dist;
+      const su2double thermal_conductivity = thermal_diffusivity * rho_cp_solid;
```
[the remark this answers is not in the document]
NAnand-TUD: Have you fixed this?

SU2_CFD/src/interfaces/cht/CConjugateHeatInterface.cpp
```diff
@@ -133,7 +133,10 @@ void CConjugateHeatInterface::GetDonor_Variable(CSolver *donor_solution, CGeomet
...
+      // TODO: add proper contact resistance here.
+      const su2double R_c = 0.0;
```
NAnand-TUD, line 138: Also, are you adding this to the config/option structure?

EvertBunschoten: Of course, it's in the check-list.

EvertBunschoten:
CHT_Fluid_Solid

Was able to run a case with fluid and two solids with contact resistance between the solid domains. By setting a non-zero contact resistance between the solid domains there appears a temperature discontinuity between the solids, but by setting the contact resistance value to zero on the fluid-to-solid interfaces, the temperature trends are continuous. 
Further testing still required on more complicated cases, but the core functionality seems to work so far.

EvertBunschoten:
> I get the idea but from the numerical standpoint this is an even worse idea than the way we do fluid-fluid interfaces. Additive decomposition of elliptic equations is not effective at all. Are you looking to model contact between multiple solid materials?

Hi Pedro. What do you mean by additive decomposition? And what do you mean by "effective" in this context? In terms of stability, the model converges rapidly, even when initializing one zone at a temperature far above that of the neighboring zone.

pcarruscag:
https://en.wikipedia.org/wiki/Additive_Schwarz_method
How large are the meshes in your examples? I expect convergence will slow down as you go to bigger problems.

EvertBunschoten:
> https://en.wikipedia.org/wiki/Additive_Schwarz_method How large are the meshes in your examples? I expect convergence will slow down as you go to bigger problems.

I ran the earlier problem with increasing node count in the solid domains. I scaled the CFL in the solid domains proportional to the grid in order to keep the same time step as the unscaled grid. I ran 5000 iterations for each grid and plotted the average iteration time scaling against the node count scaling. 

ScalingStudy

The iteration time scaling does not seem to change with increasing grid size. Additionally, the various simulations converged at the same rate. Is this a good test to test the scaling or did you have a better suggestion?

pcarruscag:
You had I comment but I don't see it here.
In any case, it's the total time to solution (same residual drop) that I expect will get worse. At constant CFL.

EvertBunschoten:
> You had I comment but I don't see it here. In any case, it's the total time to solution (same residual drop) that I expect will get worse. At constant CFL.

Sorry for replying so late, I had to put this project on the back burner for a bit. I made a mistake in my previous analysis, which is why I removed the message. I now ran five simulations with only solid to solid CHT heat transfer with contact resistance. With each simulation I quadrupled the number of mesh elements and ran with a constant CFL.  
iterations
The number of iterations required until convergence seems to scale linearly with the number of elements. However, this was not the case for total simulation time, which scales quadratically. 
simtime

pcarruscag:
@EvertBunschoten even if the performance isn't great this is a pretty small change so I think we can merge it.
Do you have time to wrap it up and add a simple regression test?

.github/workflows/regression.yml
```diff
@@ -209,7 +209,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t feature_solid-solid_cht -c feature_solid-solid_cht -s ${{matrix.testscript}}
```
pcarruscag, line 212: Don't forget to update

Common/src/CConfig.cpp
```diff
@@ -3560,6 +3562,17 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
       SU2_MPI::Error(string("You probably want to set INC_ENERGY_EQUATION= YES for the fluid solver. \n"), CURRENT_FUNCTION);
   }
 
+  /*--- Check correctness and consistency of contact resistance options. ---*/
+  if (nMarker_ContactResistance > 0) {
+    ContactResistance = true;
+    if ((nMarker_CHTInterface/2) != nMarker_ContactResistance)
+      SU2_MPI::Error(string("Number of CHT interfaces does not match number of contact resistances. \n"), CURRENT_FUNCTION);
```
pcarruscag, line 3569: You don't need to cast the error messages to string.
```suggestion
      SU2_MPI::Error("Number of CHT interfaces does not match number of contact resistances.", CURRENT_FUNCTION);
```

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2494,19 +2494,29 @@ void CDriver::InitializeInterface(CConfig **config, CSolver***** solver, CGeomet
...
+              interface_type = NO_TRANSFER;
 
+            if (config[fluidZone]->GetEnergy_Equation() || (config[fluidZone]->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE) 
+                || (config[fluidZone]->GetKind_FluidModel() == ENUM_FLUIDMODEL::FLUID_FLAMELET))
+              interface_type = heat_target? CONJUGATE_HEAT_FS : CONJUGATE_HEAT_SF;
+            else if (config[fluidZone]->GetWeakly_Coupled_Heat())
+              interface_type = heat_target? CONJUGATE_HEAT_WEAKLY_FS : CONJUGATE_HEAT_WEAKLY_SF;
+            else
+              interface_type = NO_TRANSFER;
+          }
```
pcarruscag, line 2518: Are these chunks of code the same?

TestCases/coupled_cht/solid_solid_contact_resistance/fluid.cfg
```diff
@@ -0,0 +1,55 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: solid-to-solid and solid-to-fluid conjugate heat transfer  %
```
pcarruscag, line 4: Didn't you add the configs to the tutorials repo? Or just the meshes?

Common/src/CConfig.cpp
```diff
@@ -3560,6 +3562,17 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
       SU2_MPI::Error(string("You probably want to set INC_ENERGY_EQUATION= YES for the fluid solver. \n"), CURRENT_FUNCTION);
   }
 
+  /*--- Check correctness and consistency of contact resistance options. ---*/
+  if (nMarker_ContactResistance > 0) {
+    ContactResistance = true;
```
pcarruscag, line 3567: You don't need the bool, you can just use nMarker_ContactResistance > 0, or even better allocate an array with zeros.

Common/src/CConfig.cpp
```diff
@@ -3560,6 +3562,17 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
+    if ((nMarker_CHTInterface/2) != nMarker_ContactResistance)
```
pcarruscag, line 3568: If you want to be nice you could also allow a single value and fill the array with the constant.

TestCases/parallel_regression.py
pcarruscag, line 1: Since you added a tutorial (kudos) use tutorials.py for the test instead of parallel regression

TestCases/coupled_cht/solid_solid_contact_resistance/fluid.cfg
EvertBunschoten: The configs, meshes, and gmesh .geo scripts are in the tutorials repo. I'll remove the contact resistance test case from TestCases, as the tutorial contains all that's required for running the regression test.

SU2_CFD/src/drivers/CDriver.cpp
EvertBunschoten: They are! Thanks for spotting

TestCases/parallel_regression.py
EvertBunschoten: I moved the solid-solid CHT regression from the TestCases to Tutorials

pcarruscag (approved):

.github/workflows/regression.yml
```diff
@@ -209,7 +209,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t feature_solid-solid_cht -c develop -s ${{matrix.testscript}}
```
pcarruscag, line 212: Bloody hell mate...

EvertBunschoten: doh!

PR #2167, opened 2023-11-15
Adaptive Edge Color Group Size

jblueh:
## Proposed Changes
Edge coloring tends to perform better than reductions for the hybrid parallel discrete adjoint because it uses less memory and enables the shared reading optimization. This PR puts two additional strategies in place for the discrete adjoint that help with obtaining a coloring before resorting to reductions, making successful coloring more likely.

1. If the requested edge color group size does not produce an efficient coloring, the maximum edge color group size that still produces an efficient coloring is chosen instead. This decision is made per rank.
2. The discrete adjoint is allowed to use more colors.

## Related Work
work on hybrid parallel AD, like #2039

## PR Checklist


Common/include/CConfig.hpp
```diff
@@ -9738,6 +9739,11 @@ class CConfig {
    */
   unsigned long GetEdgeColoringGroupSize(void) const { return edgeColorGroupSize; }
 
+  /*!
+   * \brief Check if the discrete adjoint is allowed to relax the coloring, that is, allow smaller edge color group sizes and allow more colors.
+   */
+  bool GetEdgeColoringRelaxDiscAdj(void) const { return edgeColoringRelaxDiscAdj; }
```
kursatyurt, line 9745:
```suggestion
  bool GetEdgeColoringRelaxDiscAdj() const { return edgeColoringRelaxDiscAdj; }
```

jblueh: Thanks for the remark. Is there general agreement in SU2 that we do it one way or the other? I saw that functions in `CConfig` have both variants, I decided to conform with the other edge coloring related function `GetEdgeColoringGroupSize(void)`.

kursatyurt: The formatting and clang tools are not applied to `CConfig` yet. Where the formatting is applied,t is always `function()`  the `void` argument comes from the C inheritance, not used in any modern C++ code. 

The rest of the code looks very clean, but since I don't have enough knowledge of this part, I cannot give feedback 👍🏼

jblueh: Thanks for the feedback!

pcarruscag: Yep it's in the plan to auto-apply clang-format to more files, we need to check if there are big PRs with changes to CConfig, and if not we can start formatting it.

Common/src/geometry/CGeometry.cpp
```diff
@@ -3637,7 +3638,56 @@ const CCompressedSparsePatternUL& CGeometry::GetEdgeColoring(su2double* efficien
 
     /*--- Color the edges. ---*/
     constexpr bool balanceColors = true;
-    edgeColoring = colorSparsePattern(pattern, edgeColorGroupSize, balanceColors);
+
+    /*--- lambda to account for different numbers of colors ---*/
```
pcarruscag, line 3642: Just for consistency with the comment style in this function.
```suggestion
    /*--- Lambda to account for different numbers of colors. ---*/
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -3637,7 +3638,56 @@ const CCompressedSparsePatternUL& CGeometry::GetEdgeColoring(su2double* efficien
...
+      } else {
+        return colorSparsePattern(pattern, edgeColorGroupSize, balanceColors);
+      }
+    };
+
+    /*--- if requested, find an efficient coloring with maximum color group size (up to edgeColorGroupSize) ---*/
+    if (maximizeEdgeColorGroupSize) {
+      auto upperEdgeColorGroupSize = edgeColorGroupSize + 1; /* upper bound that is deemed too large */
+      auto nextEdgeColorGroupSize = edgeColorGroupSize;      /* next value that we are going to try */
+      auto lowerEdgeColorGroupSize = 1ul;                    /* lower bound that is known to work */
```
pcarruscag, line 3655: const where possible please

jblueh: `upperEdgeColorGroupSize`, `nextEdgeColorGroupSize`, `lowerEdgeColorGroupSize` are all overwritten in the course of the loop that adapts the edge color group size. I found some other places to add `const` though, thanks for the remark :+1:

Common/include/geometry/CGeometry.hpp
```diff
@@ -1720,10 +1720,16 @@ class CGeometry {
   /*!
    * \brief Get the edge coloring.
    * \note This method computes the coloring if that has not been done yet.
+   * \note Can be instructed to determine and use the maximum edge color group size between 1 and
+   * CGeometry::edgeColorGroupSize that yields a coloring that is at least as efficient as #COLORING_EFF_THRESH.
    * \param[out] efficiency - optional output of the coloring efficiency.
+   * \param[in] maximizeEdgeColorGroupSize - use the maximum edge color group size that gives an efficient coloring.
+   * \param[in] largeNumberOfColors - allow up to 255 colors instead of 64 (default)
```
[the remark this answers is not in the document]
jblueh: changed in 09efb1ae7614fd988ddc83e9c8dd8c6b3fc39863

Common/src/geometry/CGeometry.cpp
```diff
@@ -3637,7 +3638,56 @@ const CCompressedSparsePatternUL& CGeometry::GetEdgeColoring(su2double* efficien
...
+
+        if (increment == 0) {
+          break;
+        }
+      }
+
+      edgeColorGroupSize = nextEdgeColorGroupSize;
+    }
+
+    edgeColoring = getColorSparsePattern(edgeColorGroupSize);
```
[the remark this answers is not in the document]
jblueh: It could be rewritten so that the last coloring evaluated in the loop is immediately used as the final coloring. So yeah, the final coloring is computed twice with adaptivity at the moment, I'll change it.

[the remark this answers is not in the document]
jblueh: The last evaluated coloring is not necessarily identical to the final coloring. If the last coloring is not admissible, the final coloring still needs to be recomputed.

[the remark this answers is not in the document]
pcarruscag: Makes sense :+1:

Issue #1625, opened 2022-05-04
NEMO symmetry boundary condition problem


otumuklu:
**Describe the bug**
*This is not a bug, rather it is a defect in the implementation of the symmetry boundary conditions. The pressure field show imperfections in the vicinity of the symmetry plane whereas the temperature field seems to be fine, as can be seen in the attached figure. Of note, although the attached result and corresponding config and mesh file belong to SU2-NEMO (i.e. non-equilibrium), this problem exists for the SU2 branch, based on private communication with @WallyMaier and @jtneedels. This case is  Argon flow over a cylinder at Mach 7.68 with a freestream pressure of 50Pa.*

**To Reproduce**
*Please find the attached config and mesh file.*

[configmesh.zip](https://github.com/su2code/SU
![SymmetryPlaneDefect](https://user-images.githubusercontent.com/104862595/166668880-2c0ba583-b276-4759-9ded-0b977e0658dc.png)
2/files/8619649/configmesh.zip)

otumuklu:
Thank you @TobiKattmann, the previous symmetry issues were very helpful for us to narrow down the problem. It is obvious that NAVIER_STOKES (SU2 EQUILIBRIUM) solver with the STANDARD_AIR model doesn't have the issue any longer, as opposed to NEMO_NAVIER_STOKES. I ran the same case with the standard air, see the attached.  We (@WallyMaier and @jtneedels) will take care of the SU2-NEMO part.


![SymmetryPlane_NS_SU2](https://user-images.githubusercontent.com/104862595/167034076-d1c627c4-2d4d-494a-9524-72ea7074b49a.png)

jtneedels:
Adding on here: we also observe a similar issue with inviscid simulation in NEMO. I've run the same case with NEMO_EULER_SOLVER and still observe a defect at the stagnation point. I implemented a mirrored state for the evaluation of the convective fluxes, following the procedure of the symmetry BC implemented in CFVMFlowSolverBase.inl, and I still observe the issue. I've zoomed in on the stagnation point here.
![image](https://user-images.githubusercontent.com/44848904/167720992-44aecdd7-2cd6-4896-a2a1-e1bd7c8e6294.png)

jtneedels:
I'll add that when I run the same case with standard air (EULER_SOLVER) with MUSCL turned off, I get the same behavior, though it goes away when MUSCL is turned on. 
[...]

PR #2174, opened 2023-11-22
Set the symmetry BC mandatorily, only for velocity.

ShiheJia:
After updating solution and before communications, the normal component is subtracted from the velocity of the sym wall.

## Related Work
https://github.com/su2code/SU2/pull/1168, https://github.com/su2code/SU2/pull/657, https://github.com/su2code/SU2/issues/1625, https://github.com/su2code/SU2/issues/1125,https://github.com/su2code/SU2/discussions/2169

The symmetry BC in SU2 is waiting for further implementation, for scalar vars and vectors. 
But first of all, we should ensure that the wall normal velocity is zero for the sym BC. 
When we run an unsteady simulation with sym BC such as the following A-airfoil cases, the spanwise velocity on the spanwise sym BC has a distribution that we donot want to see. (Free stream velocity is 0.15 and the spanwise velocity can exceed 0.01 (the ±0.001 in the figure is to show more clearly) where it should be zero.)
And this PR is to ensure the zero normal velocity.

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
+++ b/SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
@@ -991,6 +991,31 @@ class CFVMFlowSolverBase : public CSolver {
       END_SU2_OMP_FOR
     }
 
+    /*--- Set the symmetry BC mandatorily, only for velocity. ---*/
+    for (unsigned short iMarker = 0; iMarker < config->GetnMarker_All(); iMarker++) {
+      if (config->GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE) {
+        SU2_OMP_FOR_DYN(OMP_MIN_SIZE)
+        for (unsigned long iVertex = 0; iVertex < geometry->nVertex[iMarker]; iVertex++) {
+          unsigned long iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
+          if (geometry->nodes->GetDomain(iPoint)) {
+            su2double Normal[MAXNDIM] = {0.0}, UnitNormal[MAXNDIM] = {0.0},
+                      Velocity[MAXNDIM] = {0.0}, Area, ProjVel = 0.0;
+            geometry->vertex[iMarker][iVertex]->GetNormal(Normal);
+            Area = GeometryToolbox::Norm(nDim, Normal);
+            for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+              UnitNormal[iDim] = Normal[iDim] / Area;
+              Velocity[iDim] = nodes->GetSolution(iPoint, iDim+1);
+              ProjVel += UnitNormal[iDim] * Velocity[iDim];
+            }
+            for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+              nodes->SetSolution(iPoint, iDim+1, Velocity[iDim]-ProjVel*UnitNormal[iDim]);
+            }
+          }
+        }
+        END_SU2_OMP_FOR
+      }
+    }
+
     for (unsigned short iPeriodic = 1; iPeriodic <= config->GetnMarker_Periodic()/2; iPeriodic++) {
       InitiatePeriodicComms(geometry, config, iPeriodic, PERIODIC_IMPLICIT);
       CompletePeriodicComms(geometry, config, iPeriodic, PERIODIC_IMPLICIT);


pcarruscag (commented):
Can you show the residual convergence in each time step before and after this change please.

ShiheJia:
Ok, here is an example. A steady simulation is used and the relevant files are attached below.

The geometry is A-airfoil case with AoA = 13.3 deg, and the spanwise boundaries are set to MARKER_SYM.

Two cases shown below use the same configuration and mesh file, the only difference comes from the submitted code and the original code.

For the distribution on the sym bc, the spanwise speed keeps zero.


For the history of drag and lift (CSF here for the spanwise direction is z-direction), submitted code shows little differences after convergence.



For the history of the residual, increasing instabilities is not observed.

pcarruscag:
It looks like your example is only converging a couple orders of magnitude, have you tried running it for more iterations or tweaking the settings?
Have you tested this fix on the cases of the related issues? Does this change improve those results?
Have you tested this strategy for curved Euler (slip) walls, which share the implementation with the symmetry plane boundary.

bigfooted:
This is inconsistent with the current implementation. The symmetry boundary condition that is implemented is using the reflection tensor $$\bf{R} = \bf{I} - 2n_f \bigotimes n_f$$ and can be found in CFVMFlowSolverBase.inl in CFVMFlowSolverBase<V, R>::BC_Sym_Plane 

https://github.com/su2code/SU2/blob/master/SU2_CFD/include/solvers/CFVMFlowSolverBase.inl

You can also see there the implementation of the symmetry boundary condition for the velocity. You are now basically partially overwriting the weakly imposed boundary condition with a strongly imposed boundary condition. 

If you want to see a failing testcase you can try your approach with the 90 degree bend tutorial testcase and you will see that the residuals will stall after 100 iterations. With the current implementation in develop you can see that the symmetry boundary is only weakly imposed and does not fulfill the symmetry condition exactly, but the same is true for the inlet boundary condition.

ShiheJia:
Sorry for the late reply, here are some personal opinions.

(1) I have also test the turb flatplate and Onera M6 case (Fig1.Original; Fig2.Submit) in the tutorials. The convengence and results also shows just slightly difference. 

(2) This PR is not suitable for the euler wall, for the wall is not always a straight plane and singular points do not statisfy the condition for the normal direction is hard to define.

(3) Well, I have noticed the BC_Sym_Plane before, but it is very puzzling to me that only the closed point to the vertex on the sym wall is considered to add additional negative normal flux contribution. Maybe the mesh is unstructured and more than one point out of wall is linked to the vortex. Maybe the closed point is not normal to the wall while it always occurs on the 3D sym plane such as the ONERA M6 wing case.

(4) This PR is to add a limitation of the velocity on the sym wall, just like what we do to the turb walls in the CTurbSSTSolver::CTurbSSTSolver(CGeometry *geometry, CConfig *config, unsigned short iMesh). The fix is only available for the SYMMETRY wall and in the implicit iteration, just to avoid uncertainty and potential instability.
`
CTurbSSTSolver::CTurbSSTSolver(CGeometry *geometry, CConfig *config, unsigned short iMesh)
    : CTurbSolver(geometry, config, true) {
  ......

  /*--- Initialize lower and upper limits---*/
  lowerlimit[0] = 1.0e-10;
  upperlimit[0] = 1.0e10;

  lowerlimit[1] = 1.0e-4;
  upperlimit[1] = 1.0e15;
`
untitled
untitled

ShiheJia:
Here is the files of the ONERA M6 case.
[turb_ONERAM6.txt](https://github.com/su2code/SU2/files/13488523/turb_ONERAM6.txt)
[Original.txt](https://github.com/su2code/SU2/files/13488524/Original.txt)
[Submit.txt](https://github.com/su2code/SU2/files/13488527/Submit.txt)

bigfooted:
Hi,
Any progress on this PR? As I said, this implementation is not correct. A car analogy would be that you hear a strange sound coming from your car engine and to fix the problem, you turn up the volume of the radio. 
A start would be to read the relevant chapters on boundary conditions in e.g. Blazeks book and then go back to bc_sym_plane and try to understand what is going on there. The reflection condition is reflecting everything, including nonzero normal components of velocity. So the next questions should be where these normal components come from and where to fix them.

ShiheJia:
Hello, 
You are right, this pr cannot solve the problem but just make the solution look reasonable.
For some complicated case, it is hard to converge, not like in the simple cases.
I have read the relevant code and book, and I think maybe a good way is to rewrite the bc code and move the nonzero normal components limitation to where the flux are calculated. Or we can store the flux of the points on the sym bc, and in CFVMFlowSolverBase<V, R>::BC_Sym_Plane we just use the stored flux without recalculation.

PR #2190, opened 2023-12-27
[WIP] Change unsteady timestep using Python wrapper

j-signorelli:
## Proposed Changes
I have found it useful to be able to adaptively change deltaT using the Python wrapper -- this PR implements that functionality.

Instead of `CDriver::GetUnsteadyTimeStep` returning `CConfig::GetTime_Step`, it was updated to return the actual dimensional timestep used in iterations. A new `CDriver::SetUnsteadyTimeStep` function was implemented that updates both `Delta_UnstTimeND` and `Delta_UnstTime` in `CConfig`. A new setter function for `Delta_UnstTime` had to be implemented to do this as well.

A quick grep showed me that the only usage of `CConfig::GetTime_Step` appears to be in SU2_SOL, initialization, and outputting in the NonDimTable. However, I did see one usage of it here, which appears to be called every iteration:

https://github.com/su2code/SU2/blob/2c9fbb6cfc0f9dfbb4a1f6afb9cf5004c33b1c25/SU2_CFD/src/integration/CFEM_DG_Integration.cpp#L75

I can update that line if necessary to instead call `CConfig::GetDelta_UnstTime`. I don't _believe_ that there are any additional issues that may be arisen beyond this.

## Related Work
N/A

## PR Checklist

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -1140,7 +1140,8 @@ private:
   Restart_Iter;                  /*!< \brief Determines the restart iteration in the multizone problem */
   su2double Time_Step;           /*!< \brief Determines the time step for the multizone problem */
   su2double Max_Time;            /*!< \brief Determines the maximum time for the time-domain problems */
-
+  su2double Start_Time;          /*!< \brief Determines the starting time for the time-domain problems */
+  
   unsigned long HistoryWrtFreq[3],    /*!< \brief Array containing history writing frequencies for timer iter, outer iter, inner iter */
                 ScreenWrtFreq[3];     /*!< \brief Array containing screen writing frequencies for timer iter, outer iter, inner iter */
   OUTPUT_TYPE* VolumeOutputFiles;     /*!< \brief File formats to output */
@@ -9378,6 +9379,13 @@ public:
    * \return Simulation time for multizone problems, it is set on all the zones
    */
   su2double GetMax_Time(void) const { return Max_Time; }
+
+  /*!
+   * \brief Get the start simulation time for time-domain problems
+   * \return Simulation start time for multizone problems, it is set on all the zones
+   */
+  su2double GetStart_Time(void) const { return Start_Time; }
+
 
   /*!
    * \brief Get the level of MPI communications to be performed.
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -5738,6 +5738,13 @@ public:
   su2double GetDelta_UnstTime(void) const { return Delta_UnstTime; }
 
   /*!
+   * \brief Set Delta_UnstTime (dimensional timestep)
+   * \param[in] val_delta_unsttime - Value of the unsteady dimensional timestep
+   */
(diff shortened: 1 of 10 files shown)


j-signorelli:
Changed to WIP, found several different other areas that presume a constant timestep in calculation of time, like:

https://github.com/su2code/SU2/blob/2c9fbb6cfc0f9dfbb4a1f6afb9cf5004c33b1c25/SU2_CFD/src/drivers/CSinglezoneDriver.cpp#L117-L118

and

https://github.com/su2code/SU2/blob/2c9fbb6cfc0f9dfbb4a1f6afb9cf5004c33b1c25/SU2_CFD/src/solvers/CMeshSolver.cpp#L984-L991.

I am sure there are many others, so this probably is not as straightforward as I had thought.

j-signorelli:
@bigfooted Has there ever been any discussion regarding implementation of a `START_TIME=` capability into the config file? After going through the code a bit, it looks like doing this would be a good starting point for implementing ability to adjust deltaT mid-runs, as in general the code presumes `TimeIter*deltaT` as being the current time. This would be problematic for unsteady restarts. Plus, this would allow one to, without using the Python wrapper, use a different timestep in the config file for an unsteady restart. It also may clear up some confusion to have an explicit option for this, as in #2071.


It can maybe look something like:

    % Start time for restarting unsteady simulations
    % = -1 for default calculation (START_TIME=RESTART_ITER*TIME_STEP)
    START_TIME=-1

Then `CConfig::GetPhysicalTime` could be appropriately updated and used in-place of all locations in the code where a physical time is manually calculated.

pcarruscag:
Sounds good to implement START_TIME for the restart issue. But for what you want to do it would be better to just accumulate the dts to compute physical time

j-signorelli:
Just want to leave some notes here, open to any thoughts or suggestions.

I think it's best to have it work something like this for 1st-order dual timestepping:

    TIME_STEP=0.5
    RESTART_ITER=0
    ----------------------------------------------------------------------
    Time_Iter    Cur_Time
    0                 0.0
    1                 0.5
    2                 1.0
    ----------------------------------------------------------------------
    Output of restart_flow_00002.dat associated with t=1.0

Then if I want to restart that simulation but with a different timestep, I would have to make sure `START_TIME` represents the time at which the first iteration in the restart is completed to, NOT the time from the previous restart file:

    TIME_STEP=0.25
    RESTART_ITER=3
    START_TIME=1.25
    ----------------------------------------------------------------------
    Time_Iter    Cur_Time
    3                 1.25

I think it may make more sense to just call it `RESTART_TIME` so it is consistent with `RESTART_ITER`.

However I don't think that this adaptive timestepping functionality / changing of timesteps for unsteady restarts would work for 2nd-order dual timestepping since information on the previous dt is lost. It wouldn't be too challenging to have a check for this in `CDriver::SetUnsteadyTimestep` and also a warning if `START_TIME` is used on an unsteady restart with it, but it may be getting a bit clunky at that point.

j-signorelli:
Closing this for now unfortunately - I will not have the time in the immediate future to work through intricacies of this. There are too many other features of SU2 that presume a dt*time_iter as the present physical time, so implementing this in a general way that accounts for these is a bit more tedious than I had expected.

PR #2194, opened 2024-01-03
Correction of symmetry plane implementation

bigfooted:
## Proposed Changes
The implementation of the symmetry plane is incomplete. We follow here the book of Blazek, _Computational Fluid Dynamics: Principles and Applications_ . According to Blazek, (chapter 8.6) 4 conditions have to be met on a symmetry plane:

1. No flux across the boundary
2. $\overline n \cdot \nabla \phi = 0$ (gradient of scalars are zero)
3. $\overline n \cdot \nabla (\overline U \cdot \overline t) = 0$ (gradient of tangential velocity normal to boundary)
4. $\overline t \cdot \nabla (\overline U \cdot \overline n) = 0$ (gradient of normal velocity along the boundary)

Points 2-4 all deal with gradients and can be dealt with in the gradient computation, i.e. Green-Gauss or Least Squares. According to Blazek, 2 approaches can be followed. "One possibility is to construct the missing half of the control volume by mirroring the grid  on the boundary. The fluxes and gradients are then evaluated like in the interior using reflected flow variables." This approach can be implemented in an easy way when computing the  Green-Gauss gradients. In SU2, routines are already in place that deal with GG gradients on boundaries. Here, we just have to identify the symmetry planes and mirror the flux through the faces. 
Blazek continues: "The second methodology computes the fluxes for the halved control volume (but not accross the boundary). The components of **the residual normal to the symmetry plane** are then zeroed out. It is also necessary to correct normal vectors of those faces of the control volume, which touch the boundary. The modification consists of removing all components of the face vector, which are normal to the symmetry plane. The gradients also have to be corrected according to Eq. (8.40) [points 2,3,4 above]"

1. Modify Green-Gauss and Least-Squares gradient computation for symmetry planes
-> This seems to work fine
2. Fix normal components of edges connected to symmetry plane
-> This seems to work fine
3. Fix residual components of velocity on the symmetry plane
-> This seems to work fine
4. Multiple symmetry planes, including Euler walls
-> This seems to work fine
6. multigrid
-> Works fine now that we use the agglomeration as in the paper of Diskin.
7. grid movement
8. -> Works fine with the new fix to the residuals (energy update)

## Related Work
#1168 

## PR Checklist


Common/include/CConfig.hpp
```diff
@@ -6396,6 +6396,13 @@ class CConfig {
    */
   bool GetSolid_Wall(unsigned short iMarker) const;
 
+  /*!
+   * \brief Determines whether a marker with index iMarker is an inlet, outlet or far-field.
+   * \param iMarker
+   * \return <TRUE> it marker with index iMarker is a solid boundary.
+   */
+  bool Getinoutfar(unsigned short iMarker) const;
+
```
bigfooted, line 6405: We use large arrays to determine if iPoint is on a wall. I copied this procedure. This results in a boolean array of size NPoints. Maybe better to use a smarter data structure for this.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -70,81 +83,148 @@ void computeGradientsGreenGauss(CSolver* solver,
...
+
+  for (size_t iMarker = 0; iMarker < geometry.GetnMarker(); ++iMarker) {
+    if (config.GetMarker_All_KindBC(iMarker) == SYMMETRY_PLANE) {
+      for (size_t iVertex = 0; iVertex < geometry.GetnVertex(iMarker); ++iVertex) {
+
+        size_t iPoint = geometry.vertex[iMarker][iVertex]->GetNode();
+        cout << "iPoint = " << iPoint << endl;
+        auto nodes = geometry.nodes;
+        // we need to set the gradient to zero for the entire marker to prevent double-counting
+        // points that are shared by other markers
```
pcarruscag, line 156: Did you have issues without this? I expected we could modify multiple times since the modification is based on removing the projection of the gradient in certain direction, so once it's removed there's no issue in trying to remove again since the projection will be 0.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -153,29 +233,100 @@ void computeGradientsGreenGauss(CSolver* solver,
...
+          }
+
+          // find the point iPoint on the symmetry plane and get the vertex normal at ipoint wrt the symmetry plane
+          const su2double* VertexNormal = getVertexNormalfromPoint(config, geometry,iPoint);
+          cout << "  vertex normal = " << VertexNormal[0] <<", " << VertexNormal[1] << endl;
+          // get the normal on the vertex
+
+          // now reflect in the mirror
+          // reflected normal V=U - 2U_t
+          const auto NormArea = GeometryToolbox::Norm(nDim, VertexNormal);
```
pcarruscag, line 274: My understanding of the book is that we take out normal components, reflection should not be necessary

pcarruscag (commented):
So are things exploding? If you feed in a solution and look at the gradients after the first iteration does it look like the modifications are doing the job?

bigfooted:
> So are things exploding? If you feed in a solution and look at the gradients after the first iteration does it look like the modifications are doing the job?

I am using a very coarse 2D channel flow with the top a wall and bottom a symmetry, left inlet, right outlet (laminar incompressible navier stokes). The line figure is the normal velocity components at the symmetry axis, which should be zero. The black line develop. The red line is only the Green-Gauss correction. The peak at 0.5 is due to the mixed-mesh problem that the GG method has. The blue line is the current PR. 
screenshot_velocity


screenshot_axis

bigfooted:
The edge correction in PhysicalGeometry is correctly modifying the normals, all edges  between ipoint and jpoint that lie on the symmetry plane are identified, and the normal of the dual edge for those points is then corrected:
```
...
iPoint 329 and jPoint 338 are in the symmetry plane 
normal=-0.0284601 0
...
```
But only activating this edge correction is immediately causing issues. I would expect that this correction would have as an effect that the fluxes through those edges would not have components normal to the symmetry plane anymore.

pcarruscag:
So modifying the gradient + residual is an improvement and it's just the normals that cause issues? Or the change to the residual is also problematic?

bigfooted:
> So modifying the gradient + residual is an improvement and it's just the normals that cause issues? Or the change to the residual is also problematic?

The modification of the Green-Gauss gradients only is working correctly. If I take the normal velocity on the symmetry plane as a performance indicator, then I observe that there is a mean (or maximum) discrepancy $\Delta V$ for develop, which reduces to less than $0.5 \Delta V$ for the new GG.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
bigfooted: It is necessary for the stand-alone GG gradient correction. It might not be necessary when the normals and fluxes are corrected.

bigfooted: Consider a quadrilateral. We only modify the side edges. We still have contributions to the gradient in both directions from the top edge, so we still need to mirror the cell or do something else to cancel contributions normal to the symmetry plane.

pcarruscag: If you do g' = g - g.n n you ensure that g'.n = 0. The modification can be done to the entire gradient, you don’t have to go face by face.

bigfooted:
Below is a new testcase, laminar flow around a cylinder. In the line figure I show the normal velocity on the axis y=0, with the solid red line the results from develop. The dashed line is with only the new Green-Gauss correction. It always leads to an improvement. These 2 results were obtained with an undisturbed grid. I then distort the grid a bit (as seen in the contour plot) and rerun the GG correction and I get the green line. Still an improvement, but a bit more bumps on the line.
I then correct the edge normals as done in CPhysicalGeometry. I then get the jagged line shown.
The correction using the residuals in BC_Sym_Plane seems to be ill-defined because the final converged solution depends on the initial solution. 
Screenshot_cylinder

bigfooted:
Symmetry plane is correct now.
Below the figure of the super coarse diverging channel flow.
Top figure: corrected symmetry, Bottom figure: original.
screenshot_corrected

bigfooted:
Here is a picture of the diverging channel flow with the pressure contours. This is not related to the current implementation, but notice that at the corner where the outlet and the wall meet, there is a small pressure peak.

screenshot_corrected_pressure

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
bigfooted, line 1: No correction of timestep for symmetry planes.

bigfooted:
Update: Everything seems to work correctly now.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -153,29 +155,205 @@ void computeGradientsGreenGauss(CSolver* solver,
...
+          UnitNormal[iDim] = VertexNormal[iDim] / NormArea;
+
+        /*--- Normal of the primary symmetry plane ---*/
+        su2double NormalPrim[MAXNDIM] = {0.0}, UnitNormalPrim[MAXNDIM] = {0.0};
+
+        /*--- At this point we can find out if the node is shared with another symmetry.
+         * Step 1: do we have other symmetries? ---*/
+        if (nSym>1) {
+          /*--- Step 2: are we on a shared node? ---*/
+          for (auto jMarker=0;jMarker<nSym;jMarker++) {
```
bigfooted, line 214: This is duplicated in FVMFlowSolverBase, I think this can be a function here, callable from CFVMFlowSolverBase?

bigfooted:
Results for laminar flow around a cylinder. right-top figure shows normal velocity at the symmetry axis, which should be zero. old (develop) can show very high values for normal velocity. new implementation: normal velocity < 1e-8
right-bottom figure: normal velocity along a vertical line in front of the cylinder. The normal velocity should be zero at the symmetry plane.
image

bigfooted:
Flow around a sphere, taking a 5 degree slice with 2 symmetry planes. Top:new implementation. bottom: old (develop). This picture shows axial velocity, for both converged to R=-10.
image

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1089,248 +1089,186 @@
...
+    for (unsigned short iDim = 0; iDim < nDim; iDim++) {
+      LinSysRes(iPoint, iVel + iDim) -= NormalProduct * UnitNormal[iDim];
+    }
+
+    /*--- Jacobian contribution for implicit integration. ---*/
+    if (implicit) {
+
+      Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
+      /*--- Add Jacobian again for more diagonal dominance. ---*/
+      //Jacobian.AddBlock2Diag(iPoint, residual.jacobian_i);
```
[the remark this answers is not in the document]
bigfooted: Checking if the additional Jacobian update is still necessary or if it is fine to leave it out now...

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
```diff
@@ -0,0 +1,319 @@
...
+      /*--- project to symmetry aligned base ---*/
+      for (auto iDim = 0u; iDim < nDim; iDim++) {
+        gradPhi[iDim] = Gradients_iPoint[iVar][iDim];
+        gradPhiReflected[iDim] = 0.0;
+      }
+
+      for (auto jDim = 0u; jDim < nDim; jDim++) {
+        for (auto iDim = 0u; iDim < nDim; iDim++) {
+          /*--- map transpose T' * grad(phi) ---*/
+          gradPhiReflected[jDim] += TensorMap[jDim][iDim] * Gradients_iPoint[iVar][iDim];
```
pcarruscag, line 129: This is not the map transposed right? You are adding over iDim

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+      /*--- gradient in direction normal to symmetry is cancelled ---*/
+      gradPhiReflected[0] = 0.0;
+
+      /*--- Now transform back ---*/
+      for (auto jDim = 0u; jDim < nDim; jDim++) {
+        for (auto iDim = 0u; iDim < nDim; iDim++) {
+          gradPhi[jDim] += TensorMap[iDim][jDim] * gradPhiReflected[iDim];
+        }
+      }
```
pcarruscag, line 143: For example, for scalars these operations are equivalent to:
```
g = T' * [0 0 0; 0 1 0; 0 0 1] * T * g
```
(setting the first component to 0 in matrix form).
This is equivalent to
```
g = (I - T(1,:)' * T(1,:)) * g
```
(which is the same as taking out the normal component)
The tangential and orthogonal directions cancel out (as desired since they are arbitrary).

After you review the tests we can look into these simplifications.

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -49,19 +54,13 @@ namespace detail {
...
-                                const CConfig& config,
-                                const FieldType& field,
-                                size_t varBegin,
-                                size_t varEnd,
-                                GradientType& gradient)
-{
+template <size_t nDim, class FieldType, class GradientType>
+void computeGradientsGreenGauss(CSolver* solver, MPI_QUANTITIES kindMpiComm, PERIODIC_QUANTITIES kindPeriodicComm,
+                                CGeometry& geometry, const CConfig& config, const FieldType& field, size_t varBegin,
+                                size_t varEnd, GradientType& gradient, int idx_vel) {
```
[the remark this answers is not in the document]
bigfooted: idxVel = -1 when no velocity present, so just keep it at int then?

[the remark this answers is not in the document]
bigfooted: ... short int

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+#include <vector>
+#include <algorithm>
+
+#include "../../../Common/include/parallelization/omp_structure.hpp"
+#include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
+namespace detail {
+
+/*!
+ * \brief Reflect a gradient using a tensor mapping. Used for symmetry reflection.
```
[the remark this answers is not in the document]
bigfooted: It is removing normal component, I updated the description...

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
```diff
@@ -0,0 +1,319 @@
...
+        /*--- Compute a new base for TensorMap aligned with the unit normal ---*/
+        detail::BaseFromNormal(nDim,UnitNormal,TensorMap);
+
+        su2activematrix Gradients_iPoint(varEnd-varBegin,nDim);
+
+        for (auto iVar = varBegin; iVar < varEnd; iVar++) {
+          for (auto iDim = 0u; iDim < nDim; iDim++) {
+            Gradients_iPoint[iVar][iDim] = gradient(iPoint, iVar, iDim);
+          }
+        }
```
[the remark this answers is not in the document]
bigfooted: What do you mean with the gradient iterator?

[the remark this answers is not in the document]
pcarruscag: The 3D matrix returns an object that looks like a matrix but does not require copying the data. I can help with that sort of thing once all the tests are green.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1159,7 +1159,7 @@ void CFVMFlowSolverBase<V, FlowRegime>::BC_Sym_Plane(CGeometry* geometry, CSolve
 
   /*--- Loop over all the vertices on this boundary marker. ---*/
 
-  SU2_OMP_FOR_DYN(OMP_MIN_SIZE)
+  //SU2_OMP_FOR_DYN(OMP_MIN_SIZE)
```
jblueh: What was the intent behind this change?

bigfooted: I wanted to see what the effect was on the online regression tests. Sorry for the temporary confusion, I already changed it back :-)

jblueh: Word of cuation: If you remove a worksharing directive (instead of replacing, say, the dynamic schedule by a static one), the work is no longer shared, in the sense that each thread runs the full loop. This is usually not intended.

Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -7288,12 +7288,11 @@ void CPhysicalGeometry::SetBoundControlVolume(const CConfig* config, unsigned sh
      All nodes that are shared by multiple symmetries have to get a corrected normal. */
   static constexpr size_t MAXNSYMS = 100;
   unsigned short Syms[MAXNSYMS] = {0};
-
+  symmetryNormals.resize(nMarker);
```
jblueh, line 7291: `symmetryNormals` is shared, you'll either want to resize it outside OpenMP parallel code or, if it has to be done here, safeguard it with
```
SU2_OMP_SAFE_GLOBAL_ACCESS(symmetryNormals.resize(nMarker);)
```

bigfooted: Thanks Johannes, I was already wondering what I should do with it. I will move it outside of openmp code.

bigfooted: btw does the above have a performance impact? I see for instance that for a similar routine in CMultiGridGeometry.cpp, the entire function is inside the SAFE_GLOBAL_ACCESS guard.

jblueh: The safeguarding involves two barriers and serialization (only the master thread executes the code). In this specific case, where serialization is a must, you basically save the synchronization overhead due to the barriers if you move the code out of the parallel region. In general, you'll also lose performance if you place something inside the safeguards that can be parallelized efficiently, since all threads except the master idle during the safeguarded block. `CMultiGridGeometry::SetBoundControlVolume` could probably be changed along the lines of `CPhysicalGeometry::SetBoundControlVolume`. How large is `nMarker` usually?

bigfooted: nMarker here is just the number of marker/boundaries in the mesh, so that would be of O(10) usually. 
For every marker we store a vector with the ipoint and corrected normals when they are on nodes with shared symmetry or euler wall. This vector is zero when the marker is not symmetry/euler and also 0 when the symmetry/euler does not have nodes that are shared with other symmetry/eulerwalls.

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
bigfooted: OK, I get it I think, so replace like this:
```
      for (auto iDim = 0u; iDim < nDim; iDim++) {
        for (auto jDim = 0u; jDim < nDim; jDim++) {
          gradients_iPoint[iVar][iDim] -= TensorMap[0][iDim] * TensorMap[0][jDim] * Gradients_iPoint[iVar][jDim];
        }
      }
```

Common/src/geometry/CPhysicalGeometry.cpp
bigfooted: Hi @jblueh I removed the SAFE_GLOBAL_ACCESS from CMultiGridGeometry::SetBoundControlVolume and replaced it with SU2_OMP_FOR_DYN(1) for the individual loops over markers.

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
bigfooted: implemented

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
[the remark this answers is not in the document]
bigfooted: Can you have a look at the current implementation? Everything seems to work as expected.

bigfooted:
TO DO: add testcase with 2 symmetry planes (laminar flow around sphere, slice)

SU2_CFD/include/gradients/computeGradientsSymmetry.hpp
```diff
@@ -0,0 +1,319 @@
...
+        Gradients_Velocity[iVar][iDim] = Gradients_iPoint[idx_vel + iVar][iDim];
+        Gradients_Velocity_Reflected[iVar][iDim] = 0.0;
+      }
+    }
+
+    /*--- Q' = L^T*Q*T ---*/
+    for (auto iDim = 0u; iDim < nDim; iDim++) {
+      for (auto jDim = 0u; jDim < nDim; jDim++) {
+        for (auto kDim = 0u; kDim < nDim; kDim++) {
+          for (auto mDim = 0u; mDim < nDim; mDim++) {
```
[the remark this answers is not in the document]
bigfooted: modified the 4 loops...

SU2_CFD/include/gradients/computeGradientsGreenGauss.hpp
```diff
@@ -70,81 +69,75 @@ void computeGradientsGreenGauss(CSolver* solver,
...
+  /*--- Add edges of markers that contribute to the gradients ---*/
+  for (size_t iMarker = 0; iMarker < geometry.GetnMarker(); ++iMarker) {
     if ((config.GetMarker_All_KindBC(iMarker) != INTERNAL_BOUNDARY) &&
         (config.GetMarker_All_KindBC(iMarker) != NEARFIELD_BOUNDARY) &&
-        (config.GetMarker_All_KindBC(iMarker) != PERIODIC_BOUNDARY))
-    {
+        (config.GetMarker_All_KindBC(iMarker) != SYMMETRY_PLANE) &&
+        (config.GetMarker_All_KindBC(iMarker) != EULER_WALL) &&
+        (config.GetMarker_All_KindBC(iMarker) != PERIODIC_BOUNDARY)) {
+
```
pcarruscag, line 135:
```suggestion
        (config.GetMarker_All_KindBC(iMarker) != PERIODIC_BOUNDARY)) {
```

UnitTests/SU2_CFD/gradients.cpp
```diff
@@ -133,7 +133,7 @@ void testGreenGauss() {
   C3DDoubleMatrix gradient(field.geometry->GetnPoint(), field.nVar, field.geometry->GetnDim());
 
   computeGradientsGreenGauss(nullptr, MPI_QUANTITIES::SOLUTION, PERIODIC_NONE, *field.geometry.get(),
-                             *field.config.get(), field, 0, field.nVar, gradient);
+                             *field.config.get(), field, 0, field.nVar, gradient, -1);
```
pcarruscag, line 136:
```suggestion
                             *field.config.get(), field, 0, field.nVar, -1, gradient);
```

UnitTests/SU2_CFD/gradients.cpp
```diff
@@ -145,7 +145,7 @@ void testLeastSquares(bool weighted) {
   C3DDoubleMatrix gradient(field.geometry->GetnPoint(), field.nVar, nDim);
 
   computeGradientsLeastSquares(nullptr, MPI_QUANTITIES::SOLUTION, PERIODIC_NONE, *field.geometry.get(),
-                               *field.config.get(), weighted, field, 0, field.nVar, gradient, R);
+                               *field.config.get(), weighted, field, 0, field.nVar, gradient, R, -1);
```
pcarruscag, line 148:
```suggestion
                               *field.config.get(), weighted, field, 0, field.nVar, -1, gradient, R);
```

pcarruscag (approved):

PR #2209, opened 2024-02-13
[WIP] Update turbo interfaces

joshkellyjak:
## Proposed Changes
Update to the interface initialisation to enable both mixing plane and sliding interfaces within the same turbomachinery simulation. Addition of mass flow outlet to Giles BC.

## Related Work
This issue solves a problem reported by a number of users, when simulating radial turbomachinery cases a second-order accurate solution cannot be achieved in some cases when using mixing plane interfaces. This is due to circumfrential averaging in regions where the flow exhibits large circumfrential nonuniformity and/or interspace gaps between adjacent blade-rows/zones small. This issue has been reported by @LorenzoFabris, in a presentation at the SU2 Conference 2022 "Performance Analysis and Shape Optimization of a Radial Compressor with the Open-Source Software SU2" and by myself.

Mass flow outlet included in the PR as requested by users.

TBC:
-Update APU Turbocharger test case and regression test to include this feature, add diffuser and converge 2nd order.

## PR Checklist


SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6530,6 +6535,23 @@ void CEulerSolver::BC_Giles(CGeometry *geometry, CSolver **solver_container, CNu
...
+
+
+      /* --- Compute avg characteristic jump  --- */
+      if (nDim == 2){
+        c_avg[3] = -2.0*(AveragePressure[val_marker][nSpanWiseSections]-Pressure_e);
+      }
+      else
+      {
+        c_avg[4] = -2.0*(AveragePressure[val_marker][nSpanWiseSections]-Pressure_e);
+      }
```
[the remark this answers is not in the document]
joshkellyjak: Index of c_avg is associated with the type of wave and the travel direction across the boundary (downstream index 1-3, upstream index 4)

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -138,6 +138,7 @@ class CEulerSolver : public CFVMFlowSolverBase<CEulerVariable, ENUM_REGIME::COMP
   su2activematrix ExtAverageNu;
   su2activematrix ExtAverageKine;
   su2activematrix ExtAverageOmega;
+  vector<su2double> AverageMassFlowRate;
```
EvertBunschoten, line 141: Could this be an array or su2vector?

SU2_CFD/include/solvers/CEulerSolver.hpp
```diff
@@ -1191,6 +1192,15 @@ class CEulerSolver : public CFVMFlowSolverBase<CEulerVariable, ENUM_REGIME::COMP
     return AverageDensity[valMarker][valSpan];
   }
 
+  /*!
+   * \brief Provide the integral masss flow rate at the bondary of interest
+   * \param[in] val_marker - bound marker
+   * \return Value of the mass flow rate on the surface <i>val_marker</i>
+  */
+  //inline su2double GetAverageMassFlowRate(unsigned short valMarker) const final {
```
EvertBunschoten, line 1200: Don't forget to remove the commented method

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -3799,6 +3799,13 @@ class CSolver {
    */
   inline virtual su2double GetAverageDensity(unsigned short valMarker, unsigned short valSpan) const { return 0.0; }
 
+  /*!
+   * \brief virtual member
+   * \param[in] val_marker - boundary marker
+   * \return Value of the mass flow rate on the surface <i>val_marker</i>
+  */
+  inline virtual su2double GetAverageMassFlowRate(unsigned short valMarker) const {return 0.0; }
```
EvertBunschoten, line 3807:
```suggestion
  inline virtual su2double GetTargetMassFlowRate(unsigned short valMarker) const {return 0.0; }
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6407,27 +6412,15 @@ void CEulerSolver::BC_Giles(CGeometry *geometry, CSolver **solver_container, CNu
...
 
       /* --- Compute avg characteristic jump  --- */
-      if (nDim == 2){
-        c_avg[3] = -2.0*(AveragePressure[val_marker][iSpan]-Pressure_e);
-      }
-      else
-      {
-        c_avg[4] = -2.0*(AveragePressure[val_marker][iSpan]-Pressure_e);
-      }
+      c_avg[nDim + 1] = -2.0*(AveragePressure[val_marker][iSpan]-Pressure_e);
```
EvertBunschoten, line 6415:
```suggestion
      c_avg[nDim + 1] = 2.0*(Pressure_e-AveragePressure[val_marker][iSpan]);
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6438,6 +6431,16 @@ void CEulerSolver::BC_Giles(CGeometry *geometry, CSolver **solver_container, CNu
 
       break;
 
+    case MASS_FLOW_OUTLET:
+      auto const MassFlowRate_e = config->GetGiles_Var1(Marker_Tag);
+      auto const relFacMassFlowRate = config->GetGiles_Var2(Marker_Tag);
+
+      Pressure_e = AveragePressure[val_marker][nSpanWiseSections]+relFacMassFlowRate*(AverageMassFlowRate[val_marker]-MassFlowRate_e);
```
EvertBunschoten, line 6438: Maybe try this in order to keep the units more consistent here.
```suggestion
      Pressure_e = AveragePressure[val_marker][nSpanWiseSections]+relFacMassFlowRate*GetFluidModel()->GetdPdrho_e()*(AverageMassFlowRate[val_marker]-MassFlowRate_e);
```

SU2_CFD/src/solvers/CEulerSolver.cpp
joshkellyjak: Is there a guideline regarding operations like this? I kept it this way as it is more consistent with the literature

SU2_CFD/include/solvers/CSolver.hpp
joshkellyjak: This isn't the target it is the calculated value, the average indicates that is the sum of all the spanwise values

EvertBunschoten:
Don't forget to update the config_template

joshkellyjak:
> Don't forget to update the config_template

Luckily (or foolishly) I never updated the config_template when I first added the option I have now removed

EvertBunschoten (approved):

Common/src/CConfig.cpp
```diff
@@ -6008,6 +6009,35 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
     Marker_CfgFile_MixingPlaneInterface[iMarker_CfgFile] = indexMarker;
   }
 
+  /*--- Once we have identified the MixingPlane and Turbomachinery markers
+   *    we next need to determine what type of interface between zones is
+   *    used. It is convenient to do this here as it tidies up the interface
+   *    preproccesing in CDriver ---*/
+  nTurboInterfaces = (nMarker_Turbomachinery -  1)*2; //Two markers per zone
```
jblueh, line 6016: Could you explain the logic behind calculating the sizes? Why the `- 1`? At first glance, it looks like this could be conflicting with the `nMarker_Turbomachinery*2-1` further below, but I havn't checked this in detail and it might be correct.

Common/src/CConfig.cpp
```diff
@@ -6008,6 +6009,35 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
...
+    /*--- Loop over all interface markers ---*/
+    for (iMarker_ZoneInterface = 0; iMarker_ZoneInterface < nMarker_ZoneInterface; iMarker_ZoneInterface++) {
+      /*--- Identify mixing plane markers ---*/
+      if (Marker_MixingPlaneInterface != nullptr){ // Necessary in cases where no mixing plane interfaces are defined
+        if (Marker_CfgFile_TagBound[iMarker_CfgFile] == Marker_MixingPlaneInterface[iMarker_ZoneInterface]) {
+          /*--- Find which list position this marker is in turbomachinery markers ---*/
+          const auto* target = std::find(&Marker_Turbomachinery[0], &Marker_Turbomachinery[nMarker_Turbomachinery*2-1], Marker_CfgFile_TagBound[iMarker_CfgFile]);
+          const auto target_index = target - Marker_Turbomachinery;
+          /*--- Assign the correct interface ---*/
+          SetKind_TurboInterface(target_index - 1, TURBO_INTERFACE_KIND::MIXING_PLANE);
```
jblueh, line 6028: Continuing the previous question, why the `- 1`?

Common/src/CConfig.cpp
```diff
@@ -6008,6 +6009,35 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
...
+        }
+      }
+      if (Marker_Fluid_InterfaceBound != nullptr){ // Necessary in cases where no fluid interfaces are defined
+        if (Marker_CfgFile_TagBound[iMarker_CfgFile] == Marker_Fluid_InterfaceBound[iMarker_ZoneInterface]) {
+          const auto* target = std::find(&Marker_Turbomachinery[0], &Marker_Turbomachinery[nMarker_Turbomachinery*2-1], Marker_CfgFile_TagBound[iMarker_CfgFile]);
+          const auto target_index = target - Marker_Turbomachinery;
+          SetKind_TurboInterface(target_index-1, TURBO_INTERFACE_KIND::FROZEN_ROTOR);
```
jblueh, line 6035: From investigating the `slinc_steady` test case with a debug build, there is something wrong here.

`Marker_Turbomachinery` happens to be a nullpointer, so that the results of `std::find` are nonsense and you end up with `target_index == -1` and an out-of-bounds access. Also, `nMarkerTurbomachinery == 0`, which breaks both `(nMarker_Turbomachinery - 1)*2` and `nMarker_Turbomachinery*2-1`.

Are there some checks missing? It sounds like this code should not have been executed by the test case in the first place.

jblueh (commented):
Please take another look

Common/src/CConfig.cpp
joshkellyjak: nMarker_Turbomachinery is equal to half the number of markers you specify in MARKER_TURBOMACHINERY, this is because of line 1653 in option_structure.inl

`unsigned long nVals = totalVals / mod_num;
    this->size = nVals;`

I am not sure if this is a mistake, however this is used frequently in CPhysicalGeometry for defining the TurboVertex structure and it was simpler to only modify the value here, than have to change it in many multiple more places if it is a mistake. We take off 1 and multiply by 2 as we are only interested in the markers specified in MARKER_TURBOMACHINERY which are interfaces between zones, i.e. remove the inlet and outlet. I have added a comment to make this clearer in the code.

Common/src/CConfig.cpp
joshkellyjak: Added an if to only run this code if turbomachinery markers exists, however I think the issue that is causing the address sanitizer is that a fluid interface is specified in a multizone config file for the test case `slinc_steady`, hence when it loops over the iMarker_ZoneInterface it tries to access 'Marker_Fluid_InterfaceBound[iMarker_ZoneInterface]` for a position that doesn't exist in that zone - hence why it throws the error when parsing zone 1, and not when parsing the whole config at the start of the simulation. I have replaced 'Marker_CfgFile_TagBound[iMarker_CfgFile]' with 'Marker_CfgFile_KindBC[iMarker_CfgFile]' as this may be an issue in the future.

'target_index - 1' exists as this sets the interface in the correct zone when it is read in CDriver by essentially removing the inlet from the list. It was a trade-off between doing something that looks weird here, or reformat the whole interface preprocessing. I guess you could also restrict the 'std::find' but again you would run into something that looks a bit strange, this just seems to me to be the simplest way to do it - open to any suggestions on this, I will clear it up with a comment.

There is a lot of strong assumptions here, mainly because elsewhere in the code MARKER_TURBOMACHINERY is assumed to be the flow path of the machine from inlet to outlet.

Common/src/CConfig.cpp
joshkellyjak: see below

Common/src/CConfig.cpp
joshkellyjak: Well I fixed the address sanitizer; just broke something else in doing so 😄

joshkellyjak: @jblueh Appeared to have fixed these issues, would appreciate your input

Common/include/CConfig.hpp
```diff
@@ -235,6 +235,7 @@ class CConfig {
   *Marker_MixingPlaneInterface,   /*!< \brief MixingPlane interface boundary markers. */
   *Marker_TurboBoundIn,           /*!< \brief Turbomachinery performance boundary markers. */
   *Marker_TurboBoundOut,          /*!< \brief Turbomachinery performance boundary donor markers. */
+  *Marker_Turbomachinery,         /*!< \breif Turbomachinery markers */
```
jblueh, line 238:
```suggestion
  *Marker_Turbomachinery,         /*!< \brief Turbomachinery markers */
```

Common/src/CConfig.cpp
```diff
@@ -6008,6 +6009,35 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
...
+  if (nMarker_Turbomachinery != 0) {
+    nTurboInterfaces = (nMarker_Turbomachinery -  1)*2; //Two markers per zone minus inlet & outlet
+    Kind_TurboInterface.resize(nTurboInterfaces);
+    /*--- Loop over all markers ---*/
+    for (iMarker_CfgFile = 0; iMarker_CfgFile < nMarker_CfgFile; iMarker_CfgFile++) {
+      /*--- Identify mixing plane markers ---*/
+      if (Marker_MixingPlaneInterface != nullptr){ // Necessary in cases where no mixing plane interfaces are defined
+        if (Marker_CfgFile_MixingPlaneInterface[iMarker_CfgFile] != 0) { //Is a mixing plane
+          /*--- Find which list position this marker is in turbomachinery markers ---*/
+          const auto* target = std::find(&Marker_Turbomachinery[0], &Marker_Turbomachinery[nMarker_Turbomachinery*2-1], Marker_CfgFile_TagBound[iMarker_CfgFile]);
```
jblueh, line 6025: I take it from the documentation of `nMarkerTurbomachinery` that it is the size of the array `Marker_Turbomachinery`. If this is true, then any memory access beyond `nMarker_Turbomachinery - 1` is illegal, and you'll want to specify `&MarkerTurbomachinery[nMarker_Turbomachinery]` as your _last_ item, i.e., the first item beyond the range to be searched.
```suggestion
          const auto* target = std::find(&Marker_Turbomachinery[0], &Marker_Turbomachinery[nMarker_Turbomachinery], Marker_CfgFile_TagBound[iMarker_CfgFile]);
```
Are you assuming to always find your marker among the turbomachinery markers? If yes, you could make the assumption visible by adding a comment and an assertion in the spirit of
```
assert(target != &Marker_Turbomachinery[nMarker_Turbomachinery]);
```
i.e. test whether the _last_ item was returned. If no, there should be a treatment for not finding the marker.

Common/src/CConfig.cpp
```diff
@@ -6008,6 +6009,35 @@ void CConfig::SetMarkers(SU2_COMPONENT val_software) {
...
+          const auto target_index = target - Marker_Turbomachinery;
+          /*--- Assign the correct interface ---*/
+          SetKind_TurboInterface(target_index - 1, TURBO_INTERFACE_KIND::MIXING_PLANE); // Need to subtract 1 from index as to not consider the inlet an interface
+        }
+      }
+      if (Marker_Fluid_InterfaceBound != nullptr){ // Necessary in cases where no fluid interfaces are defined
```
jblueh, line 6031: `Marker_Fluid_InterfaceBound` is no longer used, is this test still needed? Should it be adapted to test for `Marker_CfgFile_KindBC != nullptr` ?

Common/src/CConfig.cpp
jblueh: Maybe you can make some of the underlying assumptions visible by assertions? See a suggestion in another comment of mine. I can't judge whether the change to `Marker_CfgFile_KindBC` is the correct solution here, maybe someone else can comment?

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2490,14 +2490,23 @@ void CDriver::InitializeInterface(CConfig **config, CSolver***** solver, CGeomet
...
-          if (config[donor]->GetBoolMixingPlaneInterface()) {
-            interface_type = MIXING_PLANE;
-            auto nVar = solver[donor][INST_0][MESH_0][FLOW_SOL]->GetnVar();
-            interface[donor][target] = new CMixingPlaneInterface(nVar, 0);
-            if (rank == MASTER_NODE) {
-              cout << "Set mixing-plane interface from donor zone "
-                  << donor << " to target zone " << target << "." << endl;
+          /*--- Interface handling for turbomachinery applications. ---*/
+          if (config[donor]->GetBoolTurbomachinery()) {
+            auto interfaceIndex = donor+target; // Here we assume that the interfaces at each side are the same kind
```
jblueh, line 2495: Out of curiosity, why is this computation of an interface index not ambiguous? Like, donor 1 and target 4 would yield the same as donor 2 and target 3?

Common/src/CConfig.cpp
joshkellyjak: The value of `nMarkerTurbomachinery` is equal to half the size of the array `Marker_Turbomachinery`. See option_structur.inl:1652 `unsigned long nVals = totalVals / mod_num;`. I don't know why it is done like this, but this is used extensively when defining the turbovertex structure and I do not want to go through and change all this. My best guess would be when this was originally implemented it was assumed that each marker specified here will have a corresponding marker in an adjacent blade row where the turboperformance variables don't need to be evaluated again. This pops up everywhere and I think it is more convenient to follow the _convention_ here rather than undertake a much larger refactoring in the other 40 places where this value is used (at least for now). I apply the -1 as I do not want to search for the final value in the MARKER_TURBOMACHINERY array, which will be the outflow.

Common/src/CConfig.cpp
joshkellyjak: This checks to see if and fluid interfaces are specified in the config, `Marker_Fluid_IntrerfaceBound` is no longer a null pointer when specified in the config?

`addStringListOption("MARKER_FLUID_INTERFACE", nMarker_Fluid_InterfaceBound, Marker_Fluid_InterfaceBound);`

SU2_CFD/src/drivers/CDriver.cpp
joshkellyjak: Because we are making a strong assumption that MARKER_TURBOMACHINERY follows the main flow path of the machine, this is done throughout other parts of the code too. These parts as well as this index will need to be refactored to include secondary flow paths in the future.

jblueh: I see :+1:

Issue #2211, opened 2024-02-14
Wall emissivity array out of bounds


maxaehle:
**Describe the bug**

The serial regression test `p1rad`, which runs SU2_CFD on [TestCases/radiation/p1model/configp1.cfg](https://github.com/su2code/SU2/blob/f9f7a0111f7ffd5b4b3a4c7fd165878eea69a7af/TestCases/radiation/p1model/configp1.cfg#L50), has a heap-buffer-overflow in [`CConfig::GetWall_Emissivity`](https://github.com/su2code/SU2/blob/f9f7a0111f7ffd5b4b3a4c7fd165878eea69a7af/Common/src/CConfig.cpp#L9614) according to Clang's address sanitizer. 

According to my printf debugging, `nMarker_Emissivity` is `2` and `Wall_Emissivity` is `{"left", "right"}` (matching the cfg file), but `CRadP1Solver::BC_Marshak` calls `GetWall_Emissivity` with the marker string `"upper"`. 

**Bug report checklist**
- [X] Consulting the build instructions (https://su2code.github.io/docs_v7/Build-SU2-Linux-MacOS/).
- [X] Looking for similar problems on GitHub or CFD-Online (https://www.cfd-online.com/Forums/su2/).
- [X] Updating to the newest version of SU2 (either develop, master, or the pre-built executables https://su2code.github.io/download.html).

**Desktop (please complete the following information):**
Ubuntu 22.04 with Clang++ 16.0.6, MPI disabled, SU2 at current `develop` branch, commit f9f7a0111f7ffd5b4b3a4c7fd165878eea69a7af.

pcarruscag:
Looks like BC_Marshak is used for all boundary types of the radiation solver, maybe we need to return 0 emissivity for markers that are not found in the list.

PR #2212, opened 2024-02-14
Fix bugs detected by the address sanitizer

maxaehle:
## Steps to reproduce the underlying issue

Compile SU2 with Clang and `-fsanitize=address` (and possibly `-fno-omit-frame-pointer`), and run the testcases as usual. AddressSanitizer warnings appear in the output (if any).

So far, only the tests in serial_regression.py have been analyzed.

## Proposed Changes

Depending on the context of the memory leak, 
- add `delete` / `delete[]` statement
- turn `new`-allocated arrays into `std::vector`s, or
- for one bug caused by a `delete` to a base class pointer, make its destructor `virtual`. 

## Related Work

#2211 and #2213 have been detected in the same way, but I'm not sure how to properly fix them.

## PR Checklist

- [not necessary] I have added a test case that demonstrates my contribution, if necessary.
- [not necessary] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.


pcarruscag (approved):
Nice catches

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -979,10 +979,8 @@ void CFVMFlowSolverBase<V, R>::LoadRestart_impl(CGeometry **geometry, CSolver **
   {
   /*--- Delete the class memory that is used to load the restart. ---*/
 
-    delete [] Restart_Vars;
-    Restart_Vars = nullptr;
-    delete [] Restart_Data;
-    Restart_Data = nullptr;
+    Restart_Vars.clear(); Restart_Vars.shrink_to_fit();
+    Restart_Data.clear(); Restart_Data.shrink_to_fit();
```
pcarruscag, line 983:
```suggestion
    Restart_Vars = decltype(Restart_Vars){};
    Restart_Data = decltype(Restart_Data){};
```

jblueh: That's better, as `shrink_to_fit` tends to free but does not guarantee it. Thanks :+1:

Common/include/option_structure.inl
```diff
@@ -1696,7 +1696,22 @@ class COptionActDisk : public COptionBase {
...
+    }
+    delete[] press_jump;
+    delete[] temp_jump;
+    delete[] omega;
+
+    delete[] marker_inlet;
+    delete[] marker_outlet;
+
+    SetDefault();
+  };
```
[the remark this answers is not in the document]
maxaehle: Indeed, `string *Marker_ActDiskBemInlet, *Marker_ActDiskBemInlet` were accessed via both the config file options `MARKER_ACTDISK_BEM_CG`, `MARKER_ACTDISK_BEM_AXIS`, and thus `delete[]`'d by both COptionActDisk instances. This was not a bug because `SetDefault` sets the pointers to zero.

As this is still kind of a dangerous construct (i.e. if the user supplies different marker lists to both cfg file options, one of them "wins" without an error message), [fc23554](https://github.com/su2code/SU2/pull/2212/commits/fc2355455e679b15ec0b8ab028a17e1df93af112) refactors the code such that both options initialize different marker lists, and those are checked to be identical in `CConfig::SetPostprocessing`.

pcarruscag (approved):

jblueh:
The contributions I made are from investigating `serial_regression_AD.py` in the same manner. `parallel_regression.py` and `parallel_regression_AD.py` are checked to a certain extent, too.

PR #2214, opened 2024-02-16
Improved look-up table query scaling with variable count

EvertBunschoten:
## Proposed Changes
The ```CLookUpTable``` class allows for the fast interpolation of unstructured 2D data.

It was found that the query time of this class scaled significantly with the number of output variables, indicating inefficiencies in the query process.
  
These inefficiencies came from string comparison routines in the ```GetIndexOfVar``` function of the ```CLookUpTable``` class. By implementing overloaded instances of the look-up functions which retrieve data from data indices directly, the query time efficiency scaling was significantly improved. 

Also reduced duplicate code within the ```CLookUpTable``` class.

## PR Checklist


pcarruscag:
You may want to remove the string based interface since it's known to be inneficient.

EvertBunschoten:
> You may want to remove the string based interface since it's known to be inneficient.

When replacing the string-based with index-based look-up methods, the performance improves substantially. The figure below shows the average measured evaluation time (measured with chrono library) vs the number of output variables. It's clear that using index-based look-up methods results in reduced query time, as well as improved scaling.

scaling_nVars

SU2_CFD/src/fluid/CFluidFlamelet.cpp
```diff
@@ -196,6 +196,22 @@ void CFluidFlamelet::PreprocessLookUp(CConfig* config) {
...
+      if (noSource(varnames_Sources[iVar])) {
+        LUT_idx = look_up_table->GetNullIndex();
+      } else {
+        LUT_idx = look_up_table->GetIndexOfVar(varnames_Sources[iVar]);
+      }
+      LUT_idx_Sources.push_back(LUT_idx);
+    }
+    for (auto iVar=0u; iVar < varnames_LookUp.size(); iVar++){
+      LUT_idx_LookUp.push_back(look_up_table->GetIndexOfVar(varnames_LookUp[iVar]));
+    }
```
bigfooted, line 214:
```suggestion
  } else {
    for (auto iVar=0u; iVar < varnames_TD.size(); iVar++) {
      LUT_idx_TD.push_back(look_up_table->GetIndexOfVar(varnames_TD[iVar]));
    }
    for (auto iVar=0u; iVar < varnames_Sources.size(); iVar++) {
      unsigned long LUT_idx;
      if (noSource(varnames_Sources[iVar])) {
        LUT_idx = look_up_table->GetNullIndex();
      } else {
        LUT_idx = look_up_table->GetIndexOfVar(varnames_Sources[iVar]);
      }
      LUT_idx_Sources.push_back(LUT_idx);
    }
    for (auto iVar=0u; iVar < varnames_LookUp.size(); iVar++) {
      LUT_idx_LookUp.push_back(look_up_table->GetIndexOfVar(varnames_LookUp[iVar]));
    }
```

Common/include/containers/CLookUpTable.hpp
```diff
@@ -284,6 +295,27 @@ class CLookUpTable {
...
 
+  /*!
+   * \brief Find the triangle within which the query point (val_CV1, val_CV2) is located.
+   * \param[in] val_CV1 - First controlling variable value.
+   * \param[in] val_CV2 - Second controlling variable value.
+   * \param[in] id_triangle - Reference to inclusion triangle index.
+   * \param[in] iLevel - Table level index.
+   * \returns 0 if inside data set, 1 if outside.
+   */
+  unsigned long FindInclusionTriangle(const su2double val_CV1, const su2double val_CV2, unsigned long& id_triangle,
```
bigfooted, line 306: Return value can be a boolean?  (If yes, the maybe change behavior to: true=inside, false=outside)

```suggestion
   * \returns 0 if inside data set, 1 if outside.
   */
  bool FindInclusionTriangle(const su2double val_CV1, const su2double val_CV2, unsigned long& id_triangle,
```

bigfooted (approved):
Thanks Evert, it looks good to me, Also nice improvement in speed.

Common/include/containers/CLookUpTable.hpp
EvertBunschoten: I changed the query inclusion flag of all lookup operations now to booleans instead of unsigned.

PR #2239, opened 2024-03-11
Add test case for fuel paper

danielmayer:
## Proposed Changes
This pull request adds the test case that we used for the Fuel paper.

## Related Work

## PR Checklist


bigfooted:
@danielmayer where does the python script for FADO go? here or in the Testcases repository?

danielmayer:
> @danielmayer where does the python script for FADO go? here or in the Testcases repository?

Good question. I think the purpose of the TestCases repo is to separate large files that don't change much from the code base. Since those scripts are small, I would lean towards putting them in the SU2 repo. Another option is to add them as a new test case to FADO, but that adds another layer of complexity. What do you think? What does @pcarruscag think?

pcarruscag:
TestCases repo is used for large files that are used in regression tests.
Then we have the tutorials repo for tutorial files (mesh + configs) and we also exercise those in regressions.
If this is too large for either of the above it should go into V&V, I guess that is how it was used for the paper?

bigfooted:
> If this is too large for either of the above it should go into V&V, I guess that is how it was used for the paper?

I suggest we add this to the regression tests, put the mesh in the testcases repo, and add additional scripts and results to the V&V and create a V&V page pointing to the paper and showing some main results.

PR #2266, opened 2024-04-16
Fix shared wall-inlet corner node for compressible

bigfooted:
## Proposed Changes
Nodes shared by an inlet and a wall show nonphysical values in the corner node for energy, temperature, pressure, density. This fixes the issue

## Related Work

## PR Checklist


bigfooted:
@Cristopher-Morales Can you check if this fixes the issue for you? This works for the testcase that you gave me (2D compressible channel flow).
There is still a (compared to develop) very small local change in pressure but I think this is because at the inlet, the radial velocity is not zero and close to the wall the radial velocity component is not negligible, this is I think why the pressure locally is still not completely homogeneous.

bigfooted:
Pressure, old (develop) and new variables. The local pressure changes reduce by almost 2 orders of magnitude:
2024-04-16_23h37_19
2024-04-16_23h34_15

Old and new energy near the corner:
2024-04-16_23h35_55
2024-04-16_23h33_50


Near the corner, the y-momentum (radial velocity) is not zero:
2024-04-16_23h34_41

bigfooted:
the user defined function case: a flat plate. top is with the fix, bottom (mirrored): the old develop.
udf

bigfooted:
@Cristopher-Morales can you check if bounded scalar works for you for this setup with compressible flow? I activated it and it seems to work fine for the small testcase that you sent. It is much better than scalar upwind.

bigfooted:
Poiseuille flow, top is corrected, bottom is develop. There is still a small discrepancy in the energy but this might be due to the strong velocity components close to the wall.
image

SU2_CFD/include/solvers/CSpeciesSolver.hpp
```diff
@@ -153,6 +153,26 @@ class CSpeciesSolver : public CScalarSolver<CSpeciesVariable> {
...
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] conv_numerics - Description of the numerical method.
+   * \param[in] visc_numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
+  void BC_Isothermal_Wall(CGeometry* geometry, CSolver** solver_container, CNumerics* conv_numerics,
+                          CNumerics* visc_numerics, CConfig* config, unsigned short val_marker) override;
+
```
bigfooted, line 175: I had to add this because scalar upwind for species transport in compressible flows showed a hard Dirichlet boundary condition of Y=0 on isothermal solid walls.

Cristopher-Morales:
> @Cristopher-Morales can you check if bounded scalar works for you for this setup with compressible flow? I activated it and it seems to work fine for the small testcase that you sent. It is much better than scalar upwind.

species_transport
Hi, yes it works much better, specially if we refine the mesh towards the walls.

SU2_CFD/include/solvers/CSpeciesSolver.hpp
Cristopher-Morales: I have an idea why we see those results in the walls, we can discuss it this Tuesday. 
image

GiuSirianni:
I'm sorry to interject on this issue with what is probably a useless comment, but I wanted to ask if this issue affects you also in 1st order simulations?

In my experience MUSCL reconstruction at boundaries in SU2 is "incorrect" as it doesn't take into account the boundary condition applied and just assumed ```v_i = v_j```, as far as I have understood from the scouring the code. This usually is a small first order approximation locally, but with complex EoS or unlucky combinations of BCs and numerical settings can cause unphysical solutions and crashes.

In a project of mine I have had some success by either disabling MUSCL outright at boundaries (very conservative and brutal solution), or disable it only at specific kinds of corners, which requires more memory unfortunately as one needs to somehow save if the node is a corner, etc.

The "real solution" would be to either save the boundary solution in ghost nodes and use it for the gradient computation, or fix gradients in post-process with a small performance penalty.

EvertBunschoten:
It looks like the edge mass fluxes are not stored when running compressible RANS simulations, preventing any species convection when using BOUNDED_SCALAR. I'm looking into why this happens. Below is the species solution after 10 iterations for the passive_transport_validation test case for the RANS solver and using BOUNDED_SCALAR. The species values specified at the inlet do not progress downstream using the current settings.

bounded_scalar_compressible

GiuSirianni:
@bigfooted I will open an issue, thank you!

bigfooted:
@EvertBunschoten since bounded scalar was never implemented for compressible flow, I think we should create a separate PR for it. 
If the symmetry PR is merged, I will update the residuals for this PR as well.

Common/src/CConfig.cpp
```diff
@@ -5563,16 +5563,12 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
...
     if (n_control_vars != (nSpecies - n_user_scalars))
-      SU2_MPI::Error("Number of initial species incompatbile with number of controlling variables and user scalars.", CURRENT_FUNCTION);
+      SU2_MPI::Error("Number of initial species incompatible with number of controlling variables and user scalars.", CURRENT_FUNCTION);
     /*--- We can have additional user defined transported scalars ---*/
     n_scalars = n_control_vars + n_user_scalars;
   }
 
-  if (Kind_Regime == ENUM_REGIME::COMPRESSIBLE && GetBounded_Scalar()) {
-    SU2_MPI::Error("BOUNDED_SCALAR discretization can only be used for incompressible problems.", CURRENT_FUNCTION);
-  }
```
pcarruscag, line 5575: We still have this restriction

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -6863,10 +6863,11 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
                             CConfig *config, unsigned short val_marker) {
   unsigned short iDim;
   unsigned long iVertex, iPoint;
-  su2double P_Total, T_Total, Velocity[MAXNDIM], Velocity2, H_Total, Temperature, Riemann,
+  su2double P_Total, T_Total, Velocity[MAXNDIM], Velocity2, H_Total, Riemann,
   Pressure, Density, Energy, Flow_Dir[MAXNDIM], Mach2, SoundSpeed2, SoundSpeed_Total2, Vel_Mag,
   alpha, aa, bb, cc, dd, Area, UnitNormal[MAXNDIM], Normal[MAXNDIM];
   su2double *V_inlet, *V_domain;
+  su2double Temperature = 288.15;
```
pcarruscag, line 6870: This looks like a bad idea

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -7079,8 +7080,8 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
           if (tkeNeeded) Energy += GetTke_Inf();
 
           /*--- Primitive variables, using the derived quantities ---*/
-
-          V_inlet[0] = Pressure / ( Gas_Constant * Density);
+          Temperature = Pressure / ( Gas_Constant * Density);
```
pcarruscag, line 7083:
```suggestion
          const su2double Temperature = Pressure / ( Gas_Constant * Density);
```

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -7094,6 +7095,65 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
...
+            }
+
+            Pressure = nodes->GetPressure(iPoint);
+            Density = nodes->GetDensity(iPoint);
+
+            Energy = Pressure / (Density * Gamma_Minus_One)  + 0.5 * Velocity2;
+
+            if (tkeNeeded) Energy += GetTke_Inf();
+
+            V_inlet[nDim+3] = Energy + Pressure/Density;
```
pcarruscag, line 7146: Why not pressure and density too?

SU2_CFD/src/solvers/CSpeciesSolver.cpp
```diff
@@ -406,6 +406,20 @@ void CSpeciesSolver::BC_Inlet(CGeometry* geometry, CSolver** solver_container, C
...
+
+}
+
+void CSpeciesSolver::BC_Isothermal_Wall(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics,
+                                       CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
+
+  BC_Isothermal_Wall_Generic(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
+
+}
+
```
pcarruscag, line 421: Why are these needed?

bigfooted:
@pcarruscag Finally had some time to look into this again. Behavior is still as described above. If you have time, can you have another look?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -7114,6 +7114,69 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
...
+            V_inlet[0] = Temperature;
+            V_inlet[nDim+1] = Pressure;
+            V_inlet[nDim+2] = Density;
+            V_inlet[nDim+3] = Energy + Pressure / Density;
+            break;
+          }
+
+          case INLET_TYPE::MASS_FLOW: {
+
+            /*--- Impose the wall velocity from the interior wall node. ---*/
```
pcarruscag, line 7152: I don't understand why the treatment differs for total conditions and mass flow.

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -7114,6 +7114,69 @@ void CEulerSolver::BC_Inlet(CGeometry *geometry, CSolver **solver_container,
...
+            Velocity2 = 0.0;
+            for (iDim = 0; iDim < nDim; iDim++) {
+              V_inlet[iDim+1] = nodes->GetVelocity(iPoint,iDim);
+              Velocity2 += V_inlet[iDim+1] * V_inlet[iDim+1];
+            }
+
+            /*--- Match the pressure, density and energy at the wall. ---*/
+
+            Pressure = nodes->GetPressure(iPoint);
+            Density = Pressure / (Gas_Constant * Temperature);
```
pcarruscag, line 7138: This temperature comes from the normal stagnation treatment?
Total conditions already respects the velocity of the interior node, what is this changing?
And, can we instead change the PTotal and TTotal that are specified and let the normal treatment do the rest?

bigfooted:
Thanks, Pedro :-)

PR #2272, opened 2024-04-24
fix grad_val

NAnand-TUD:
## Proposed Changes
*As mentioned in the last few dev meetings, there was some error with GradVal for the streamwise solver with Massflow. This was because the take was not cleared during SetRecoding. In the current version, the register of extra solutions is moved to Register Variables and gradients look consistent.*

## Related Work
*SWP-Isothermal with mass flow also shows good gradient validation after implementing this fix. In theory, this should not fix any other part of the code.*

## PR Checklist

- [ x ] I am submitting my contribution to the develop branch.
- [ x ] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [ x ] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
CAD_GRAD_AFTER
CAD Sensitivites after making changes, gradients are symmetric for the upper and lower variables.
CAD_GRAD_BEFORE
CAD Sensitivities before making changes, gradients are not-symmetric for the upper and lower variables.
Case_Image_RTemp
Test Case!
FFD_GRAD_AFTER
FFD Sensitives also co-relate well

### DIFF
--- a/SU2_CFD/src/solvers/CDiscAdjSolver.cpp
+++ b/SU2_CFD/src/solvers/CDiscAdjSolver.cpp
@@ -162,8 +162,6 @@ void CDiscAdjSolver::RegisterSolution(CGeometry *geometry, CConfig *config) {
   /*--- Boolean true indicates that an input is registered ---*/
   direct_solver->GetNodes()->RegisterSolution(true);
 
-  direct_solver->RegisterSolutionExtra(true, config);
-
   if (time_n_needed)
     direct_solver->GetNodes()->RegisterSolution_time_n();
 
@@ -261,6 +259,8 @@ void CDiscAdjSolver::RegisterVariables(CGeometry *geometry, CConfig *config, boo
     config->SetIncPressureOut_BC(BPressure);
     config->SetIncTemperature_BC(Temperature);
 
+    direct_solver->RegisterSolutionExtra(reset, config);
+
   }
 
   /*--- Register incompressible radiation values as input ---*/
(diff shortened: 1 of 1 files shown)


SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -261,6 +259,8 @@ void CDiscAdjSolver::RegisterVariables(CGeometry *geometry, CConfig *config, boo
     config->SetIncPressureOut_BC(BPressure);
     config->SetIncTemperature_BC(Temperature);
 
+    direct_solver->RegisterSolutionExtra(reset, config);
```
pcarruscag, line 262: I don't think the boolean passed to this function means "reset". But I'll assign @TobiKattmann to this one.

NAnand-TUD: `unsigned long CIncEulerSolver::RegisterSolutionExtra(bool input, const CConfig* config) {
  if (config->GetKind_Streamwise_Periodic() == ENUM_STREAMWISE_PERIODIC::MASSFLOW) {
    if (input) AD::RegisterInput(SPvals.Streamwise_Periodic_PressureDrop);
    else AD::RegisterOutput(SPvalsUpdated.Streamwise_Periodic_PressureDrop);
    return 1;
  }
  return 0;
}`

Yes, you are right, so we are doing the opposite then. I will investigate a bit more.

PR #2281, opened 2024-05-14
Using free stream tremperature for initialization of wall temperature for pure conjugate heat transfer problems

Cristopher-Morales:
## Proposed Changes
For pure conjugate heat transfer problems, the temperature at the walls is initialized using the free stream temperature. Using the Isothermal_Temperature[0] for pure CHT problems results in a segmentation fault error.

## PR Checklist

### DIFF
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -9234,7 +9234,7 @@ su2double CConfig::GetIsothermal_Temperature(const string& val_marker) const {
     if (Marker_Isothermal[iMarker_Isothermal] == val_marker)
       return Isothermal_Temperature[iMarker_Isothermal];
 
-  return Isothermal_Temperature[0];
+  return Temperature_FreeStream;
 }
 
 su2double CConfig::GetWall_HeatFlux(const string& val_marker) const {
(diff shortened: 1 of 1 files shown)


Common/src/CConfig.cpp
```diff
@@ -9234,7 +9234,7 @@ su2double CConfig::GetIsothermal_Temperature(const string& val_marker) const {
     if (Marker_Isothermal[iMarker_Isothermal] == val_marker)
       return Isothermal_Temperature[iMarker_Isothermal];
 
-  return Isothermal_Temperature[0];
+  return Temperature_FreeStream;
```
EvertBunschoten, line 9237:
```suggestion
  // Return free-stream temperature for pure CHT cases.
  return Temperature_FreeStream;
```

EvertBunschoten (approved):

pcarruscag:
Why is this being accessed out of bounds? Instead of adding a potentially confusing default behavior we should not access a marker that doesnt exist.

bigfooted:
In CNSSolver, we do in BC_Isothermal_Wall_Generic:
```
if (!cht_mode) {
    Twall = config->GetIsothermal_Temperature(Marker_Tag) / Temperature_Ref;
  }

```

And we do not check for cht_mode in IncNSSolver. I think this might be the root cause of entering this routine? 
@Cristopher-Morales can you check this?

Issue #1079, opened 2020-09-30
time-varying boundary conditions


flybob91:
Dear SU2 developers,
Is it possible to have a time-varying boundary condition? I would interested in the possibility to apply a time-varying velocity

Thank you for your job

Best regards

clarkpede:
This is possible using the Python wrapper.  There are no tutorials for this yet, but you can check out PR #488, PR #503. PR #516.  There's currently a regression test using unsteady heat-transfer boundary conditions at: https://github.com/su2code/SU2/tree/master/TestCases/py_wrapper/flatPlate_unsteady_CHT . You can use that as an example.  Velocity BCs will be very similar.

flybob91:
Thank you Clark,
I will come back as soon as I will be able to run the case to extend your script.

Best regards

flybob91:
I met some problem in the original launch_unsteady_CHT_FlatPlate.py in [in the unsteady CHT flate plate] (https://github.com/su2code/SU2/tree/master/TestCases/py_wrapper/flatPlate_unsteady_CHT) case.

The first problem is that the function GetUnsteady_TimeStep() (line 141) returns an  object of class 'SwigPyObject' while the SU2Driver.GetTime_Iter() (line 142) returns a 'int' object. This causes the execution to fail since the two variable are multiplied at line 144. 

I tried to inset a numerical value for deltaT in order to avoid the problem but I got the following error
`TypeError: in method 'CDriver_SetVertexTemperature', argument 4 of type 'passivedouble'`

So I suppose the problem is the function  GetTime_Iter() which shouldn't return an 'int' object. Has anyone met the same problem?

aleargix:
Hello SU2 developers,

I am reopening this topic because I am studying a case of jet impingement on a wall with a time-varying flow at the inlet. I have used compressible RANS and TOTAL_CONDITIONS as inlet BC. Specifically, I have set a time-varying stagnation pressure to achieve a time-varying flow.

To do this, I used the Python script mentioned in the previous messages, modifying some commands to suit my problem. I should mention that I do not use Python frequently, so there are likely some errors. Indeed, I checked the documentation and noticed that the command I used to vary the pressure at the inlet (SetMarkerCustomPtotal) is not listed.

However, when I run the simulation, it seems to work quite well. Unfortunately, when examining the pressure field, it appears that the time-varying boundary conditions are applied  on the walls with no-slip conditions (represented by the MARKER_ISOTHERMAL command) and not at the inlet as I want.

[...]

PR #2296, opened 2024-06-08
Custom inlets via python wrapper

pcarruscag:
## Proposed Changes
```
SetMarkerCustomInletFlowVar0(marker, vertext, val)
SetMarkerCustomInletFlowVar1(marker, vertext, val)
SetMarkerCustomInletFlowDirection(marker, vertext, xyz)
``` 

## Related Work
#1079 

## PR Checklist

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -6841,7 +6841,7 @@ public:
    * \param[in] val_index - Index corresponding to the inlet boundary.
    * \return The total temperature.
    */
-  su2double GetInlet_Ttotal(const string& val_index) const;
+  su2double GetInletTtotal(const string& val_index) const;
 
   /*!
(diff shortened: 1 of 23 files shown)


bigfooted:
Thanks Pedro! For incompressible, we can use this but with different meaning at Var0, Var1, right? 

For each of the other solvers like turbulence, species, a CustomTurbVar(marker,vertex,val,index) and a CustomSpeciesVar(marker,vertex,val,index) can then be introduced in the same way?

pcarruscag:
Yep, even for compressible flow there is a different meaning for mass flow inlet, vars 0 and 1 are density and velocity magnitude instead of total pressure and temperature, that's why I went with Var0 and Var1 for names.

Yeah good idea, I'll rename these functions to have "Flow" in the name.

PR #2311, opened 2024-06-25
Change for sliding plane when relative frame is used

FabianYan2010:
## Proposed Changes
In some cases we use relative frame for rotor zone to perform URANS, e.g., in aeroelasticity analysis, we want the grid movement comes only from blade deformation. The problem is that the sliding plane in SU2 works only for absolute frame. In relative frame, the grid is not rotating, so the interpolation at sliding interface is not changed as time step is marching. The proposed changes rotate the sliding interface in accordance with physical time steps to perform interpolation. This function is activated only for relative frame. The grid itself is not rotating.
Currently, this is only implemented in CNearestneighbor class, other interpolation method should be changed as well in the near future.

## Related Work

## PR Checklist


TestCases/turbomachinery/axial_stage_full_annulus_unsteady/rotor_FA_Unsteady.cfg
```diff
@@ -0,0 +1,22 @@
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%                                                                              %
+% SU2 configuration file                                                       %
+% Case description: 2D Axial stage                                             %
+% Author: S. Vitale                                                            %
+% Institution: Delft University of Technology                                  %
+% Date: Feb 28th, 2017                                                         %
+% File Version 5.0.0"cardinal"                                                 %
```
bigfooted, line 8: Change authorship and get the credit that you deserve... Can you also add all new testcases to the regression tests, at least put it into parallel_regression.py

pcarruscag (commented):
Can you make this a config option as I mentioned before?
The large solution files need to be in the TestCases repo instead of together with the code.

pcarruscag:
Rebased this branch onto current `develop` and applied a few fixes found during review:

**Bug fixes**
- The donor-velocity rotation in `CInterface::BroadcastData` was gated only on `donor_config->GetBoolRelFrame_SlidingPlane()`, while the coordinate rotation in `CNearestNeighbor::SetTransferCoeff` uses `donor OR target`. Made the velocity gate match (OR), otherwise a config that only sets the flag on the target zone gets correctly-matched coordinates but unrotated velocities.
- `GetDonor_Velocity_RotatingFrame`'s target-side branch read `TimeIter` from `donor_config` instead of `target_config`.
- The per-target-vertex rotation matrix (and rotated target coordinate) in `CNearestNeighbor::SetTransferCoeff` was being recomputed inside the innermost per-donor-candidate loop; hoisted it out since it's invariant across all donor candidates for a given target vertex/zone pair. This matters because this feature makes the interpolation recompute every time step.
- `CMultizoneDriver` now only treats a `ROTATING_FRAME` zone as needing a per-timestep transfer-coefficient refresh when `RELATIVE_FRAME_SLIDINGPLANE` is actually enabled, so existing `ROTATING_FRAME` + sliding-interface setups that don't use this feature don't pay a new per-timestep cost.
- Fixed copy-pasted Doxygen comments (`target_config`/`target_geometry` were both labeled "the donor mesh") and cleaned up trailing whitespace.

**Test case / data files**
- `TestCases/sliding_interface/inlet_distortion_relative_frame/`'s `.dat` mesh/solution/profile files were committed directly into this repo, but `TestCases/.gitignore` reserves that pattern for the sibling `su2code/TestCases` data repo (same as every other `sliding_interface/*` case, e.g. `rotating_cylinders.su2`). Moved them there: su2code/TestCases#204 (also recovered the mesh files and initial restart solutions, which existed in this branch's history — commit 3f5d63d — but had been dropped before the PR's current state, so the case couldn't actually run at all as submitted).
- Also found the inlet-profile filenames didn't match `RESTART_ITER=1` (SU2's existing multizone/unsteady file-naming convention expects `Inlet_Distortion_<zone>_00001.dat` for `RESTART_ITER=1`, not `_00000`) — fixed by renaming in the TestCases PR.
- Wired the case into `parallel_regression.py`/`serial_regression.py` (with a reduced `TIME_ITER=5` for CI, following the existing `uniform_flow` pattern) using `test_vals` from an actual local run with the fixes applied.

**Still needed before this can pass CI as-is:** su2code/TestCases#204 needs to merge first (or `regression.yml`'s `-c develop` temporarily pointed at that branch) since the new regression entry's mesh/solution/profile data lives there, not in this repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

TestCases/serial_regression.py
```diff
@@ -985,6 +985,17 @@ def main():
...
+    inlet_distortion_relative_frame         = TestCase('inlet_distortion_relative_frame')
+    inlet_distortion_relative_frame.cfg_dir   = "sliding_interface/inlet_distortion_relative_frame"
+    inlet_distortion_relative_frame.cfg_file  = "duct_unsteady.cfg"
+    inlet_distortion_relative_frame.test_iter = 4
+    inlet_distortion_relative_frame.test_vals = [4.000000, 0.000000, -0.100233, -0.079874]
+    inlet_distortion_relative_frame.tol       = 0.000001
+    inlet_distortion_relative_frame.unsteady  = True
+    inlet_distortion_relative_frame.multizone = True
+    test_list.append(inlet_distortion_relative_frame)
+
```
pcarruscag:
```suggestion
```

pcarruscag (commented):
I removed the test because it was too large, almost 1 minute.

PR #2317, opened 2024-07-02
[WIP] Multizone discrete adjoints for turbomachinery

oleburghardt:
Continued by #2446.

## Original Message

This is, for now, a brief note on @joshkellyjak 's and my WIP on a discrete adjoint functionality for multizone (multi-stage, multi-row) turbomachinery cases. We intend to realize everything through the native multizone discrete adjoint solver, though some turbo-specific problems need to be resolved.

This will, probably, involve further additions/corrections/cleanups around the mixing plane interface so that it is recorded properly. The current recording seems to still be not fully correct.


oleburghardt:
@joshkellyjak @pcarruscag Should I make some first commits for a "tag type / debug" recording to this branch? (This problem is the main motivation but technically it's quite unrelated.)

joshkellyjak:
> @joshkellyjak @pcarruscag Should I make some first commits for a "tag type / debug" recording to this branch? (This problem is the main motivation but technically it's quite unrelated.)
> 
> We need it here, but it might be better to merge it from a dedicated branch.

Personally I would like this to selfishly enhance my understanding, but I don't think it should be merged when we complete the PR. Could be useful documentation for the website.

oleburghardt:
> Personally I would like this to selfishly enhance my understanding, but I don't think it should be merged when we complete the PR. Could be useful documentation for the website.

Well I think the functionality should be merged at some point. It might become too much for this PR(?). Let me just go ahead and create a dedicated branch for the tag tape. We could just use that branch as a helper then and prepare it for develop after this one is merged.

pcarruscag:
Ah now I got the comment. I would keep adding the debug tape separate from fixing the turbo issues

Issue #1851, opened 2022-12-21
Turbulent Kinetic Energy(TKE) on energy equation in SST model.


sun5k:
**Describe the bug**

I found a bug while testing the SST model.
I'm trying to simulate the high Mach number flat plate case. The flow condition is below.

Temperature      (K) | Unit   Re   (1/m) | Mach | Pressure | Wall   Temperature(K)
-- | -- | -- | -- | --
800 | 4.9×10^6 | 6.1 | 12100 | 300

After calculation, I check the flow field, including the boundary. But, the inlet boundary is not the same flow conditions I imposed. 
Here is the comparison of other models using the same flow conditions. 

![image](https://user-images.githubusercontent.com/50565314/208847631-a3ced0ab-7a70-45ee-b67e-41314b2ddfb1.png)

![image](https://user-images.githubusercontent.com/50565314/208847743-6f08eb96-ea5e-4ac5-a50a-0d980f33c547.png)

So, I tried to find the cause. 

When deriving the RANS equation from the energy equation, the turbulent kinetic energy(TKE) is derived from the kinetic energy of Total Energy(E) as shown below.
![image](https://user-images.githubusercontent.com/50565314/208849564-da4f1b0a-ff2f-4732-911a-b1ed2a4ca1c8.png)

In the boundary condition and initialization part, k is considered depending on whether or not the SST model is used for Total Energy.

In CEulerSolver.cpp 
`bool tkeNeeded          = (turbulent && config->GetKind_Turb_Model() == TURB_MODEL::SST);`
`if (tkeNeeded) { Energy_FreeStream += Tke_FreeStream; }; config->SetEnergy_FreeStream(Energy_FreeStream);`
`if (tkeNeeded) Energy_FreeStreamND += Tke_FreeStreamND;`

But In the flux calculation part, when calculating the speed of sound(a), using the total enthalpy(H). But, in this part consider only the kinetic energy. 

Expressed mathematically 
![image](https://user-images.githubusercontent.com/50565314/208861650-92b8cad6-5a1f-4742-952b-30d0187ae28b.png)

but in the SU2 code using below.

![image](https://user-images.githubusercontent.com/50565314/208861902-ff614d53-d07e-4237-a04d-73a49ad5a55c.png)

The speed of sound always has a slightly high value because TKE is not subtracted.

See "roe.cpp" line 119 ~ 149. 
`RoeSoundSpeed2 = (Gamma-1)*(RoeEnthalpy-0.5*sq_vel);` 
[...]

PR #2329, opened 2024-07-25
Inconsistencies and improvements to SST model

rois1995:
## Proposed Changes
Hi everyone, 

I have found some inconsistencies with respect to the literature on the implementation of the Menter's SST model. I would like to use this branch as test bench for any corrections/improvements made to the SST model.

Implementation errors found:
- Use of production limiter constant in the clipping of the cross-diffusion term for the computation of the F1 blending function in CTurbSSTVariable.cpp. 
- SST naming conventions (V in V2003m should stay for Vorticity production term.). Not yet assessed.
- Wrong cross-diffusion term included into the residual of w in turb_sources.hpp. It should not be only the positive value as considered in the computation of the F1 blending function. As stated in https://doi.org/10.2514/3.12149.  "CDkw is the
positive portion of the cross-diffusion term in Eq (A2)" pag. 1604. Moreover, a clipping has been introduced for large negative values of this term, as suggested in Peng, Shia-Hui, Peter Eliasson, and Lars Davidson. "Examination of the shear stress transport assumption with a low-Reynolds number k-omega model for aerodynamic flows." Eq 17.
- Boundary conditions errors at inlets, following the Issue #1851. A fix has been proposed following @pcarruscag  suggestions for the supersonic inlet BC.

Changes to SST model proposed:
- Inclusion of non modified versions of SST. In this case I think that more work is needed to be sure that the correct terms are taken into account everywhere in the code.

I've seen the proposed changes to the lower limits of k and w in #2323 and I tried implementing it in my branch. However, if the implementation proposed in the respective PR is preferred then I will change mine. 

## Related Work
#2323 #1851 

## PR Checklist

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -1168,6 +1168,9 @@ private:
   nHistoryOutput, nVolumeOutput;  /*!< \brief Number of variables printed to the history file. */
   bool Multizone_Residual;        /*!< \brief Determines if memory should be allocated for the multizone residual. */
   SST_ParsedOptions sstParsedOptions; /*!< \brief Additional parameters for the SST turbulence model. */
+  su2double lowerLimitTKE,
+            lowerLimitDissipation;
+  su2double prodLimConst;
   SA_ParsedOptions saParsedOptions;   /*!< \brief Additional parameters for the SA turbulence model. */
   LM_ParsedOptions lmParsedOptions;   /*!< \brief Additional parameters for the LM transition model. */
   su2double uq_delta_b;         /*!< \brief Parameter used to perturb eigenvalues of Reynolds Stress Matrix */
@@ -9852,6 +9855,10 @@ public:
    */
   SST_ParsedOptions GetSSTParsedOptions() const { return sstParsedOptions; }
 
+  su2double GetLowerLimitTKE() const { return lowerLimitTKE; }
+  su2double GetLowerLimitDissipation() const { return lowerLimitDissipation; }
+  su2double GetProdLimConst() const { return prodLimConst; }
+  
   /*!
    * \brief Get parsed SA option data structure.
    * \return SA option data structure.
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -869,8 +869,7 @@ private:
   array<su2double, N_POLY_COEFFS> MuPolyCoefficientsND{{0.0}};  /*!< \brief Definition of the non-dimensional temperature polynomial coefficients for viscosity. */
   array<su2double, N_POLY_COEFFS> KtPolyCoefficientsND{{0.0}};  /*!< \brief Definition of the non-dimensional temperature polynomial coefficients for thermal conductivity. */
   su2double TurbIntensityAndViscRatioFreeStream[2]; /*!< \brief Freestream turbulent intensity and viscosity ratio for turbulence and transition models. */
-  su2double TKE_ProductionLimiterConstant; /*!< \brief Freestream turbulent intensity and viscosity ratio for turbulence and transition models. */
-  bool Change_TKE_ProductionLimiterConstant;
+
   su2double Energy_FreeStream,     /*!< \brief Free-stream total energy of the fluid.  */
   ModVel_FreeStream,               /*!< \brief Magnitude of the free-stream velocity of the fluid.  */
   ModVel_FreeStreamND,             /*!< \brief Non-dimensional magnitude of the free-stream velocity of the fluid.  */
@@ -8819,8 +8818,6 @@ public:
    * \brief Set freestream turbonormal for initializing solution.
    */
   const su2double* GetFreeStreamTurboNormal(void) const { return FreeStreamTurboNormal; }
-  const su2double GetTKE_ProductionLimiterConstant(void) const { return TKE_ProductionLimiterConstant; }
-  const bool GetChange_TKE_ProductionLimiterConstant(void) const { return Change_TKE_ProductionLimiterConstant; }
 
   /*!
    * \brief Set multizone properties.
@@ -9858,7 +9855,7 @@ public:
   su2double GetLowerLimitTKE() const { return lowerLimitTKE; }
   su2double GetLowerLimitDissipation() const { return lowerLimitDissipation; }
   su2double GetProdLimConst() const { return prodLimConst; }
-  
+
   /*!
    * \brief Get parsed SA option data structure.
    * \return SA option data structure.
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -856,7 +856,8 @@ private:
   nMu_Temperature_Ref,           /*!< \brief Number of species reference temperature for Sutherland model. */
   nMu_S,                         /*!< \brief Number of species reference S for Sutherland model. */
   nThermal_Conductivity_Constant,/*!< \brief Number of species constant thermal conductivity. */
-  nPrandtl_Lam,                  /*!< \brief Number of species laminar Prandtl number. */
+  nPrandtl_Lam,                  /*!< \brief Number of species 
+  addDoubleOption("FREESTREAM_TURB2LAMVISCRATIO", TurbIntensityAndViscRatioFreeStream[1], 10.0); Prandtl number. */
   nPrandtl_Turb,                 /*!< \brief Number of species turbulent Prandtl number. */
   nConstant_Lewis_Number;       /*!< \brief Number of species Lewis Number. */
   su2double Diffusivity_Constant;   /*!< \brief Constant mass diffusivity for scalar transport.  */
@@ -868,6 +869,8 @@ private:
   array<su2double, N_POLY_COEFFS> MuPolyCoefficientsND{{0.0}};  /*!< \brief Definition of the non-dimensional temperature polynomial coefficients for viscosity. */
   array<su2double, N_POLY_COEFFS> KtPolyCoefficientsND{{0.0}};  /*!< \brief Definition of the non-dimensional temperature polynomial coefficients for thermal conductivity. */
   su2double TurbIntensityAndViscRatioFreeStream[2]; /*!< \brief Freestream turbulent intensity and viscosity ratio for turbulence and transition models. */
+  su2double TKE_ProductionLimiterConstant; /*!< \brief Freestream turbulent intensity and viscosity ratio for turbulence and transition models. */
+  bool Change_TKE_ProductionLimiterConstant;
   su2double Energy_FreeStream,     /*!< \brief Free-stream total energy of the fluid.  */
   ModVel_FreeStream,               /*!< \brief Magnitude of the free-stream velocity of the fluid.  */
   ModVel_FreeStreamND,             /*!< \brief Non-dimensional magnitude of the free-stream velocity of the fluid.  */
@@ -8813,6 +8816,8 @@ public:
    * \brief Set freestream turbonormal for initializing solution.
    */
   const su2double* GetFreeStreamTurboNormal(void) const { return FreeStreamTurboNormal; }
+  const su2double GetTKE_ProductionLimiterConstant(void) const { return TKE_ProductionLimiterConstant; }
+  const bool GetChange_TKE_ProductionLimiterConstant(void) const { return Change_TKE_ProductionLimiterConstant; }
 
   /*!
    * \brief Set multizone properties.
--- a/Common/include/option_structure.hpp
+++ b/Common/include/option_structure.hpp
@@ -990,18 +990,20 @@ enum class SST_OPTIONS {
   V,           /*!< \brief Menter k-w SST model with vorticity production terms. */
   KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
   UQ,          /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
+  FULLPROD,    /*!< \brief Menter k-w SST model with full production term. */
 };
 static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
   MakePair("NONE", SST_OPTIONS::NONE)
   MakePair("V1994m", SST_OPTIONS::V1994m)
   MakePair("V2003m", SST_OPTIONS::V2003m)
   /// TODO: For now we do not support "unmodified" versions of SST.
-  //MakePair("V1994", SST_OPTIONS::V1994)
-  //MakePair("V2003", SST_OPTIONS::V2003)
+  MakePair("V1994", SST_OPTIONS::V1994)
+  MakePair("V2003", SST_OPTIONS::V2003)
   MakePair("SUSTAINING", SST_OPTIONS::SUST)
   MakePair("VORTICITY", SST_OPTIONS::V)
   MakePair("KATO-LAUNDER", SST_OPTIONS::KL)
   MakePair("UQ", SST_OPTIONS::UQ)
+  MakePair("FULLPROD", SST_OPTIONS::FULLPROD)
 };
 
 /*!
@@ -1013,6 +1015,7 @@ struct SST_ParsedOptions {
   bool sust = false;                          /*!< \brief Bool for SST model with sustaining terms. */
   bool uq = false;                            /*!< \brief Bool for using uncertainty quantification. */
   bool modified = false;                      /*!< \brief Bool for modified (m) SST model. */
+  bool fullProd = false;                      /*!< \brief Bool for full production term. */
 };
 
 /*!
@@ -1029,6 +1032,8 @@ inline SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS *SST_Options, unsigne
     const auto sst_options_end = SST_Options + nSST_Options;
     return std::find(SST_Options, sst_options_end, option) != sst_options_end;
   };
+
+  const bool found_fullProd = IsPresent(SST_OPTIONS::FULLPROD);
 
   const bool found_1994 = IsPresent(SST_OPTIONS::V1994);
   const bool found_2003 = IsPresent(SST_OPTIONS::V2003);
@@ -1071,6 +1076,7 @@ inline SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS *SST_Options, unsigne
   SSTParsedOptions.sust = sst_sust;
   SSTParsedOptions.modified = sst_m;
   SSTParsedOptions.uq = sst_uq;
+  SSTParsedOptions.fullProd = found_fullProd;
   return SSTParsedOptions;
 }
 
--- a/Common/include/option_structure.hpp
+++ b/Common/include/option_structure.hpp
@@ -991,6 +991,8 @@ enum class SST_OPTIONS {
   KL,          /*!< \brief Menter k-w SST model with Kato-Launder production terms. */
   UQ,          /*!< \brief Menter k-w SST model with uncertainty quantification modifications. */
   FULLPROD,    /*!< \brief Menter k-w SST model with full production term. */
+  LLT,         /*!< \brief Menter k-w SST model with Lower Limits Treatments. */
+  PRODLIM,     /*!< \brief Menter k-w SST model with user-defined production limiter constant. */
 };
 static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
   MakePair("NONE", SST_OPTIONS::NONE)
@@ -1004,6 +1006,8 @@ static const MapType<std::string, SST_OPTIONS> SST_Options_Map = {
   MakePair("KATO-LAUNDER", SST_OPTIONS::KL)
   MakePair("UQ", SST_OPTIONS::UQ)
   MakePair("FULLPROD", SST_OPTIONS::FULLPROD)
+  MakePair("LLT", SST_OPTIONS::LLT)
+  MakePair("PRODLIM", SST_OPTIONS::PRODLIM)
 };
 
 /*!
@@ -1016,6 +1020,8 @@ struct SST_ParsedOptions {
   bool uq = false;                            /*!< \brief Bool for using uncertainty quantification. */
   bool modified = false;                      /*!< \brief Bool for modified (m) SST model. */
   bool fullProd = false;                      /*!< \brief Bool for full production term. */
+  bool llt = false;                           /*!< \brief Bool for Lower Limits Treatment. */
+  bool prodLim = false;                       /*!< \brief Bool for user-defined production limiter constant. */
 };
 
 /*!
@@ -1034,6 +1040,8 @@ inline SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS *SST_Options, unsigne
   };
 
   const bool found_fullProd = IsPresent(SST_OPTIONS::FULLPROD);
+  const bool found_llt = IsPresent(SST_OPTIONS::LLT);
+  const bool found_prodLim = IsPresent(SST_OPTIONS::PRODLIM);
 
   const bool found_1994 = IsPresent(SST_OPTIONS::V1994);
   const bool found_2003 = IsPresent(SST_OPTIONS::V2003);
@@ -1077,6 +1085,8 @@ inline SST_ParsedOptions ParseSSTOptions(const SST_OPTIONS *SST_Options, unsigne
   SSTParsedOptions.modified = sst_m;
   SSTParsedOptions.uq = sst_uq;
   SSTParsedOptions.fullProd = found_fullProd;
+  SSTParsedOptions.llt = found_llt;
+  SSTParsedOptions.prodLim = found_prodLim;
   return SSTParsedOptions;
 }
 
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -6180,7 +6180,7 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
             }
             cout << "." << endl;
 
-            if (Change_TKE_ProductionLimiterConstant) cout << "Changinf the value of the TKE production limiter constant to " << TKE_ProductionLimiterConstant << endl;
+            if (sstParsedOptions.prodLim) cout << "Changing the value of the TKE production limiter constant to " << prodLimConst << endl;
             
             break;
         }
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -6181,6 +6181,7 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
             cout << "." << endl;
 
             if (sstParsedOptions.prodLim) cout << "Changing the value of the TKE production limiter constant to " << prodLimConst << endl;
+            if (sstParsedOptions.llt) cout << "Changing the value of the lower limits of TKE and Omega " << endl;
             
             break;
         }
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1431,8 +1431,6 @@ void CConfig::SetConfig_Options() {
   addDoubleOption("INITIAL_BCTHRUST", Initial_BCThrust, 4000.0);
   /* DESCRIPTION:  */
   addDoubleOption("FREESTREAM_TURB2LAMVISCRATIO", TurbIntensityAndViscRatioFreeStream[1], 10.0);
-  addBoolOption("CHANGE_TKE_PRODUCTION_LIMITER_CONSTANT", Change_TKE_ProductionLimiterConstant, false);
-  addDoubleOption("TKE_PRODUCTION_LIMITER_CONSTANT", TKE_ProductionLimiterConstant, 10.0);
   /* DESCRIPTION: Side-slip angle (degrees, only for compressible flows) */
   addDoubleOption("SIDESLIP_ANGLE", AoS, 0.0);
   /*!\brief AOA  \n DESCRIPTION: Angle of attack (degrees, only for compressible flows) \ingroup Config*/
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1118,6 +1118,10 @@ void CConfig::SetConfig_Options() {
   /*!\brief SST_OPTIONS \n DESCRIPTION: Specify SA turbulence model options/corrections. \n Options: see \link SA_Options_Map \endlink \n DEFAULT: NONE \ingroup Config*/
   addEnumListOption("SA_OPTIONS", nSA_Options, SA_Options, SA_Options_Map);
 
+  addDoubleOption("LOWER_LIMIT_TKE", lowerLimitTKE, 1e-20);
+  addDoubleOption("LOWER_LIMIT_DISSIPATION", lowerLimitDissipation, 1e-6);
+  addDoubleOption("PROD_LIM_CONST", prodLimConst, 20.0);
+
   /*!\brief KIND_TRANS_MODEL \n DESCRIPTION: Specify transition model OPTIONS: see \link Trans_Model_Map \endlink \n DEFAULT: NONE \ingroup Config*/
   addEnumOption("KIND_TRANS_MODEL", Kind_Trans_Model, Trans_Model_Map, TURB_TRANS_MODEL::NONE);
   /*!\brief SST_OPTIONS \n DESCRIPTION: Specify LM transition model options/correlations. \n Options: see \link LM_Options_Map \endlink \n DEFAULT: NONE \ingroup Config*/
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1427,6 +1427,8 @@ void CConfig::SetConfig_Options() {
   addDoubleOption("INITIAL_BCTHRUST", Initial_BCThrust, 4000.0);
   /* DESCRIPTION:  */
   addDoubleOption("FREESTREAM_TURB2LAMVISCRATIO", TurbIntensityAndViscRatioFreeStream[1], 10.0);
+  addBoolOption("CHANGE_TKE_PRODUCTION_LIMITER_CONSTANT", Change_TKE_ProductionLimiterConstant, false);
+  addDoubleOption("TKE_PRODUCTION_LIMITER_CONSTANT", TKE_ProductionLimiterConstant, 10.0);
   /* DESCRIPTION: Side-slip angle (degrees, only for compressible flows) */
   addDoubleOption("SIDESLIP_ANGLE", AoS, 0.0);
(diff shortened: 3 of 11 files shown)


Common/include/CConfig.hpp
```diff
@@ -856,7 +856,8 @@ class CConfig {
   nMu_Temperature_Ref,           /*!< \brief Number of species reference temperature for Sutherland model. */
   nMu_S,                         /*!< \brief Number of species reference S for Sutherland model. */
   nThermal_Conductivity_Constant,/*!< \brief Number of species constant thermal conductivity. */
-  nPrandtl_Lam,                  /*!< \brief Number of species laminar Prandtl number. */
+  nPrandtl_Lam,                  /*!< \brief Number of species 
+  addDoubleOption("FREESTREAM_TURB2LAMVISCRATIO", TurbIntensityAndViscRatioFreeStream[1], 10.0); Prandtl number. */
```
bigfooted, line 860: Was this intentionally commented, or...? 
```suggestion
  nPrandtl_Lam,                  /*!< \brief Number of species laminar Prandtl number. */
  addDoubleOption("FREESTREAM_TURB2LAMVISCRATIO", TurbIntensityAndViscRatioFreeStream[1], 10.0); /*!<\brief  Freestream mu_turb to mu_lam viscosity ratio */
```

bigfooted:
Great contribution, Thanks @rois1995 !

pcarruscag:
If you are looking into robustness aspects too you should get in touch with @emaberman and @YairMo, seems like they have some good ideas and between the free time of 3 people a lot more can get done :)

YairMO:
Hi,

Regarding the cross-diffusion term (CD) that appears in Omega source term (residual). The SST model (1994/2003) is a high-Reynolds-number  model. Namely, It can not predict correctly the sub-layer region (especially the correct profile of the TKE). Therefore, only a positive contribution is required. Moreover, since the SST model was design as a k-w and k-epsilon blending, the CD term "belongs" only to the k-epsilon "branch", that is why the CD term include the factor "1-F1". However, it may happen, that the factor "1-F1" is not a 100% safe guarantee. It may happen that "1-F1" is not zero in region where the CD term is negative (this happen due numerical errors). To avoid such a situation, it is a good idea to clip the CD term with zero. Otherwise, severe numerical robustness issues may rise. 
Yes, it is different from Menter publications, but I think that clipping  the CD term with zero is completely inline with Menter original idea (that is why, I think,  he was including the factor "1-F1". But again the 1-F1 factor is not 100% percent "safe").

Common/include/CConfig.hpp
rois1995: This is just a wrong copy-paste from me. It should not be there in the first place.

YairMO:
Hi,

The use of an Omega production limiter (about the cross-diffusion term) is correct for low-Reynolds-number (LRN) models (the approach described by Peng et al. is very naive; there are other more rigorous treatments). For high-Reynolds-number (HRN) models, the clipping should be zero, keeping the cross-diffusion term positive; thus, the current implementation is correct.

Indeed, it is not exactly as it appears in Menter's original publication. The factor (1-F1) aimed to promise that the cross-diffusion term will be activated only outside the boundary layer, where it is positive (the cross-diffusion term switches its sign deep inside the boundary layer). This was also recognized by Peng et al. (first paragraph above Eq. 17). However, it may happen that the factor (1-F1)=1 where the cross-diffusion term is negative. Usually, it may happen at the wake, very near the airfoil trailing edge, where the upper and lower boundary layers merge. It is due to the imperfection of the F1 function.

To summarize, the current implementation is correct, and it is perfect for HRN models.

YairMO:
For the sake of clarity, "current implementation" refers to the current treatment of the production code

emaberman:
What YairMO is saying, is that allowing negative cross diffusion values is incorrect for high Reynolds models and should not be an option, this is a fix used for low Reynolds models only

rois1995:
Hi @YairMO, Hi @emaberman ,

thank you very much for your comments. I haven't found any suggestion in literature to clip to only positive values the cross-diffusion term in the w-equation. I understand that it might be more robust, but it is not the standard implementation of the SST model, which is the first thing that we need to achieve. Only then we can build on top of that to improve the robustness of SU2.

Nevertheless, I tried the SWBLI test case and I compared the results across 6 different combinations:

1- develop branch, no changes
2- develop branch, changes to Supersonic_inlet profile as suggested in #1851 
3- my branch, with original CDkw implementation (should give exactly the same result as develop+modified BC)
3- my branch, with original CDkw implementation and using boundary conditions from TMR
4- my branch, with original CDkw implementation and using your suggestions for lower limits for k and w.
5- my branch, allowing negative values of CDkw
6- my branch, allowing negative values of CDkw and using boundary conditions from TMR
7- my branch, allowing negative values of CDkw and using your suggestions for lower limits for k and w.
8- my branch, allowing negative values of CDkw, using boundary conditions from TMR and using your suggestions for lower limits for k and w.

When my branch is used, then the changes to the supersonic inlet BC are already in place.

I haven't achieved convergence with 1, 2 and 3. More precisely, 1 diverged right away (after 30 iterations), while 2 and 3 gave "FGMRES - Orthogonalization Failed" after 900ish iterations.

Here you can see the residuals for the different combinations.

OrigCDkw

NegCDkw

Unfortunately I will be busy with the AIAA Conference next week, thus I don't know how much I will be able to work on this. The next test case will be the multi-element airfoil.

YairMO:
Hi rois1995,

First of all, enjoy your time in Las Vegas. Any paper that you are presenting?

As for our discussion about the cross-diffusion term, I've emailed the "source" (Menter). I believe he will make it clear.
It may be that he will be able to answer only in a while ...

pcarruscag:
In my experience, by-the-book implementation of SST is very unstable in real-world meshes and applications.
For example there is a production limit, but it's proportional to k and w, and thus if those start growing for some reason there is nothing preventing the model from exploding.
CD is also tricky due to the 1/w part...

YairMO:
I totally agree. On real meshes, Omega usually drops to extremely low values. In cases where the cross-diffusion term is negative (allowed to be negative),  it behaves as a sink term, further amplifying the drop of Omega. A simple addition of this term to the implicit diagonal is insufficient (I tried this). Other more rigorous methods are required (some available in the open literature). 

My main argument is that the factor (1-F1) guarantees only the k-epsilon branch outside the boundary layer, in which the cross-diffusion term is already positive. It may not be so because of numerical errors.

YairMO:
Regarding "Fixed Supersonic inlet BC inclusion of TKE," does the energy equation include the contribution of turbulent diffusion?

YairMO:
Hi All,

We are lucky, Florian Menter just replied to me.

He agreed that the factor (1-F1) should activate the CD term only at the "k-epsilon" branch, where the CD term is already positive.
Therefore, clipping with zero is in terms of the model and is not incorrect.


Me: A simple boundary layer simulation will reveal that the CD term switch sign is inside the boundary layer and that the F1 function switches from 1 to zero outside the boundary layer. Namely, the CD term should be positive. Florian is unaware of the occasions when it may be negative (even though the model was designed so that this term will be activated when it is positive). 

Best wishes,
Yair

rois1995:
> Regarding "Fixed Supersonic inlet BC inclusion of TKE," does the energy equation include the contribution of turbulent diffusion?

Yeah, I do think so.

rois1995:
I have tested the tripping of the cross-diffusion term also in the residuals on a more complicated test case and indeed it improved convergence. Should I remove this change and use the value of cross-diffusion coming from the computation of F! blending function as was before?

YairMO:
Hi,

May I understand "tripping" as "clipping"? 

The cross-diffusion term (that appears in the residual of Omega) is better treated as follows:

2 * (1-F1) * sigma_w2 * rho / Omega * max(grad K * grad Omega, epsilon)

Where "epsilon" could be = 1.e-20 (personally, I'm using epsilon=0)

The term CDkw (that is used to calculate F1) should be implemented in a similar manner:

CDkw = 2 * sigma_w2 * rho / Omega * max(grad K * grad Omega, epsilon)

where epsilon = 1e-20 for the SST1994 and epsilon=1e-10 for the SST2003.

You mentioned that the SU2 implementation of CDkw is not as above?

rois1995:
Yeah, sorry, bad wording from my side. The standard version of SU2 used the value of CDkw as value of the cross-diffusion term in the residual computations. Thus, I think I should revert to that formulation.

Regarding the CDkw, it's implementation in SU2 was made such that the exponent of the epsilon term was equal to minus the production limiter constant of the model. Thus 20 for SST1994 and 10 for SST2003. Is it just a coincidence in the model formulation? In the original paper there is no mention of it being the same variable. This can lead to different results if someone changes the production limiter constant.

pcarruscag:
I expect that epsilon to be a simple measure to avoid division by 0, if that lower bound had physical meaning it would have  to be multiplied by some reference factors to make its dimensions appropriate, otherwise SST would not give the same results for the same Reynolds obtained with different rho and mu.

YairMO:
You are correct! I have asked Menter about this many, many years back. I don't remember his response, but the formulation given in the TMR is correct (with this fault!)

rois1995:
I have two considerations on the new  boundary conditions: 

1- should the "domain length" in the Omega farfield BC be adimensionalized by the Reynolds length? Otherwise w gets adimensionalized only by the reference velocity.
2- the farfield value of TKE does not take into account for the freestream turbulence intensity. If we want to match the tke computed via $k_{\infty}=w_{\infty} * (\mu_{\infty} * ViscRatio)/\rho_{\infty}$ and the one computed previously as $k_{\infty}=(3/2)*\sqrt((U_{\infty}*TI_{\infty})^2)$, we can either compute the viscosity ratio combining the two equations or just use the one  which explicitly contains the $TI_{\infty}$ disregarding the dependency on $w_{\infty}$. But can we do it? The model as written on the NASA TMR does not explicitly contains the turbulence intensity in the boundary conditions. Maybe it only matters when transition is taken into account.

YairMO:
Hi, 

I'm not sure what "new boundary conditions" means. Is someone replacing the present far-field BC of the TKE and Omega? 
I recognize these "new" BCs as the ones given on the TMR (proposed in the original paper).

I find the present setting in SU2 very comfortable and more "intuitive" than that given on the TMR.

rois1995:
So, the two boundary conditions currently implemented in this branch are:

Standard SU2 BCs 
$k_{\infty} = (3.0/2.0) * (U_{\infty} * TI_{\infty})^2$
$w_{\infty} = \rho_{\infty} * k_{\infty} / (\mu_{lam,\infty} * ViscRatio)$
$ViscRatio_{default} = 10.0$
$TI_{\infty, default} = 5.0$

NASA TMR BCs for SST1994 and SST 2003 model
$U_{\infty}/L \leq w_{\infty} \leq 10 U_{\infty}/L$
$10^{-5}U_{\infty}^2/Re_{L} \leq k_{\infty} \leq 10^{-1}U_{\infty}^2/Re_{L}$

where "L is the approximate length of the computational domain," and the combination of the two farfield values should yield a freestream turbulent viscosity between $10^{-5}$ and $10^{-2}$ times freestream laminar viscosity, as cited on the NASA TMR page.

They give very different values of k and w at the farfield. For example, for a the standard test case of the SWBLI with $ViscRatio = 0.01$ and $TI_{\infty} = 5.0$:

Standard SU2 BCs 
$k_{\infty} = 2574.55$ $m^2/s^2$ and $w_{\infty} = 3e10$ $s^{-1}$

NASA TMR BCs
$k_{\infty} = 0.0013$ $m^2/s^2$ and $w_{\infty} = 1563$ $s^{-1}$

I assume that in the end it doesn't matter since the farfield values drop significantly in the standard version of SST.

rois1995:
I am trying to run again the SWBLI V&V case on the SU2 page. However, as it is provided, it does not work with the SU2 standard boundary conditions. In particular, it always diverges due to NaN residuals in the X-Momentum equation. However, when I use the BCs from the NASA TMR page it works just fine (although only the v2003m model). 

Here are the results of the computations performed. 

Three configs are used:
1 - With dimensionless limits $k_{lim} = 10^{-20}$ and $w_{lim} = 10^{-10}$.
2 - Without dimensionless limiters.
3 - With dimensionless limits $k_{lim} = 10^{-30}$ and $w_{lim} = 10^{-20}$.

I am reporting the Skin Friction Coefficient on the lower wall against the Experimental results and the trend of the residual for $\rho$ during the simulations. Five different meshes have been considered.

ComparisonNewBCsResiduals

ComparisonNewBCsSFC

The solutions are pretty much identical for the same mesh level, but the cases with the config 3 achieved convergence across all of the meshes. It is interesting how, on finer meshes, the use of dimensionless limits (config 1) actually had worst convergence than the one with std limits (config 2) (up until the finest mesh, where config 2 diverges).

I am going to try to perform the simulations also with the non m variants of the SST model and the full production term. It may require some work since all of the occurrences of the "ComputeStressTensor" have to be changed. Ideally I think that it should be computed only once and  then used that value for all of the other instances. I will focus only on the V2003 model for now as it is the only one that does not diverge.

rois1995:
@YairMO  are these results something that you would expect to obtain when changing the lower limits of k and w? With the new updates to the Reynolds Stress computations I obtain mixed results:

ComparisonNewBCsResiduals

On the finest mesh the dimensionless limits diverge regardless of their value (I also tried decreasing them but it didn't work).

YairMO:
Hi, 

To begin with, thank you very much for your efforts and the tremendous work you put into improving the SST model!

It is very difficult for me to judge because I'm using a different time integration than the SU2, which may be critical in this regard. Moreover, my code is non-dimensional, so estimating the influence of the "newBC" and the "lower limits" is difficult for me. Can you tell the values of the "lower limits" relative to the freestream values? Please provide many details about the inputs in your config file (Linear solver, linear solver convergence, CFL, inviscid flux, etc')

YairMO:
I found the SWBLI config file. I noticed that it is the NK solver. I would test this case without the NK solver.
Moreover, I've noticed that the entropy fix is very low; how about raising it to 0.2 or using a different flux (JST, AUSMPLUSUP2)? It may sound irrelevant, but it is worth the try.

rois1995:
Changing the entropy fix coefficient (EFC) to 0.2 had the largest effects across the board. It actually degraded the convergence rate, but it removed divergence of the solvers for finer meshes.

ComparisonNewBCsResiduals_EF0 2

The converged solutions do not change when changing the EFC, which is desirable.

ComparisonNewBCsSol_EF0 2

YairMO:
Great! This is a non-linear example: Can we say that the convergence pattern is not necessarily directly affected by the turbulence model's lower limits? 
Did you try avoiding the NK solver?

rois1995:
I don't think we can actually say that, or, at least, we can say that it is not the only factor affecting the convergence. The problem is that it is not consistent across the meshes: for sure, using dimensionless limits help, but changing their values has different effects for finer meshes.

ComparisonNewBCsResiduals_EF0 2_Other

The effect on the residual of the turbulent variables is the same. The most problematic one is always the Omega equation where the residuals remain always large.

ComparisonNewBCsResiduals_EF0 2_Other_TKE

ComparisonNewBCsResiduals_EF0 2_Other_Omega

The problem is that Omega is too large with respect to all the other variables, especially near the wall (where the residuals remain high).

YairMO:
I agree. 
The NK solver is not well suited to solving the turbulence model. Maybe the linear solver alone will be more robust.?

pcarruscag:
The NK solver is only applied to the flow equations.

YairMO:
Thank you. How about our lower limits (committed):

K_lowLimit = 1E-15 * K_FarField
Omega_lowLimit = 1E-5 * Omega_FarField

rois1995:
It changes somehow the convergence, but again, it doesn't always have the same effect.

ComparisonNewBCsSol_EF0 2

My problem right now is that if I run the v2003m config on the finest mesh with 40 cores then it diverges after 125 iterations. If I run with 20 cores then it does not diverge. The number of cells/core is more then 10k thus I don't think that the domain is overly partitioned.

pcarruscag:
If you are using multigrid, it is known to be sensitive to partitioning. You could always use the hybrid parallel mode, that lets you use more cores with less partitioning.

rois1995:
Currently I am not using multigrid. I tried with with 20 cores and 2 threads and it almost results in the same rate of convergence of the simulation with only 20 cores. I also tried going back to a couple of commits, right after the merge of the develop branch and it does not diverge on 40 cores. Starting from that commit I only slightly changed the computation of the stress tensor, which does not involve new mpi communications, thus I don't understand why it should lead to divergence. I will look further into this.

rois1995:
I have tried running 7 consecutive and identical simulations with 4C10T and the residuals are not consistent. Moreover, Run 1 and 5 diverged respectively after 251 and 430 simulations. Here I am showing the residuals of the TKE (I know that NK does not apply to the turbulence solver but it is cleaner to visualize than the residual for the density)

NonDeterministicRMSk

I also tried going back a few commits right before changing the Reynolds stress computation and the situation is even worse.

NonDeterministicRMSk_PreviousCommit

Here i summarize the convergence of these runs, 1e4 iterations means that it has "converged":

| Run | Iterations | Iterations (old commit)| 
|  :---:  |      :---:       |                  :---:                  |
|    1   |      251       |                1193                 |
|    2   |      1e4       |                1286                 |
|    3   |      1e4       |                  95                   |
|    4   |      1e4       |                 458                  |
|    5   |      430       |                 1e4                  |
|    6   |      1e4       |                 254                  |
|    7   |      1e4       |                 1e4                  |

I might try with the develop branch, but if I recall correctly I was not able to run the test case in the first place. But maybe it was just a random divergence.

rois1995:
I am trying with the .geo file used to create the meshes in the V&V folder. However, I am trying with a finer mesh, with a mesh sizing factor of f = 3.9762. I also tried the develop branch and, even though it always diverges, also there the iterations done are different. I have updated the previous comment.

With the coarser meshes I might have the same problems, but convergence is easier thus I think that I might need more trials. I'll start running some simulation with the 3rd level grid.

pcarruscag:
Are you initializing from coarser results? I just checked that the L3 mesh still works fine in develop initialized from L2 results.
The NK solver has some non determinism when using threads especially if the convergence is rough. Using a lot of cores also makes the ILU preconditioner less effective, so you may need to tweak linear solver iterations or CFL to compensate. For reference when I ran the case originally I had a 4 core laptop xD

rois1995:
Every simulation starts from zero. I also tried the third grid level and these are the results for after just four runs

NonDeterministicRMSk_Level3

Run 2 diverges while Run 4 converges to the desired residual value (log(RMS(Res_rho)) < -11.5). Running on 4 cores and 10 threads is a problem in this case? Or has the same problems of using 40 cores?

pcarruscag:
The ILU issue depends on number of cores regardless of MPI/OpenMP split.
The non-determinism of the NK solver should be worse with more threads.
I recommend initializing from the coarse results.

rois1995:
I tried a couple of things:

- with 20C2T and the results are worse: 4 out of 7 runs diverged, one right away (84 iterations) and the others after more than 1000 iterations. 

- restarting the simulation from the level 3 mesh didn't change the results: 2 out of 7 simulations diverged. 

- running without NK improved a quite a lot, but the inconsistencies remain. Not only that, but something strange can be seen looking at the first hundreds of iterations: the even runs are on top of each other and the same happens for the odd ones.

No Newton-Krylov
NonDeterministicRMSk_NoNK

No Newton-Krylov Zoomed
NonDeterministicRMSk_NoNKZoom

The peak in the residuals that you can see around 3k iterations is different for each run, and leads to slightly different residuals after 10k iterations. 

- Running with 40C1T causes divergence of the simulation at the same iteration for every run (using NK), and in this case the residuals are on top of each other.

I will try having a look into the atomic operations.

pcarruscag:
There are two sources for those, `atomicAdd` which is used in dot products with CSysVectors (so it affects the linear solvers). And in the NK solver to compute the norm of the solution.
The other less obvious source is SU2_OMP_CRITICAL, critical sections do not have a guaranteed ordering. We use these to compute the RMS residuals, and those influence the CFL adaptation.
AFAIK the only way around this is to store the result of each thread into a vector visible by all threads, and then have a single thread compute the total.

rois1995:
I can try having a look into this, but I am not sure of how much help I can be. Just from a quick look at the code I can see that the CSYSVEC_PARFOR has a NOWAIT instuction, whereas the SU2_OMP_FOR_STAT doesn't, and these are used right before the atomicAdd call respectively in the CSysVector and in the CNewtonIntegration. But I doubt that's the main cause of large discrepancies when using NK.

pcarruscag:
I think NK is just more sensitive because the linear system uses a nastier Jacobian matrix.

rois1995:
Having a better look into the comparison between this branch and the develop one for the simulation with 10C and 1T.

When looking at the solution, the residuals actually improve with this branch everywhere except at the bottom symmetry BC. Don't really know why, I am checking all of the changes to check. 

ComparisonAbsResRho_MyBranchVsDevelop

SU2_CFD/include/numerics_simd/flow/variables.hpp
```diff
@@ -124,7 +135,8 @@ FORCEINLINE CRoeVariables<nDim> roeAveragedVariables(Double gamma,
     roeAvg.velocity(iDim) = (R*V.j.velocity(iDim) + V.i.velocity(iDim)) * D;
   }
   roeAvg.enthalpy = (R*V.j.enthalpy() + V.i.enthalpy()) * D;
-  roeAvg.speedSound = sqrt((gamma-1) * (roeAvg.enthalpy - 0.5*squaredNorm(roeAvg.velocity)));
+  roeAvg.tke = (R*V.j.tke() + V.i.tke()) * D;
+  roeAvg.speedSound = sqrt((gamma-1) * (roeAvg.enthalpy - 0.5*squaredNorm(roeAvg.velocity) - roeAvg.tke));
```
pcarruscag, line 139: Is enthalpy being computed with tke? If not and then you subtract tke here...
Also please control the scope of this PR, you have another 3 work in progress PRs that took time to review and now are just sitting.

rois1995: My bad, the enthalpy is computed just as E-P/rho, thus the kinetic energy should not be involved here. Regarding the other WIPs, they all involve the SST model (except for the recent SA one, but that should be finished), thus I think they should wait for this branch first.

rois1995:
As of now the problem is that the supersonic inlet has the correct values of Mach number (5) but on the symmetry BC this does not occur. It seems that on the symmetry BC the problem is the same as it was for the supersonic inlet before, but I still have to find a way to fix this. Moreover, this only happens to the lower symmetry BC, whereas at the top one the Mach is correct.

YairMO:
Maybe I'm mixing it up, but I remember there was an issue with the wrong treatment (I'm not sure exactly what it was, perhaps gradient treatment) at Symmetry BC.

rois1995:
I will look into this, but what I found is that the velocity component normal to the bottom symmetry BC is non-zero in my branch.

YairMO:
I think this was one of the problems mentioned

rois1995:
Yeah but this does not happen on the results with the develop branch and I have pulled the latest develop commits into this branch, thus it might not be the cause but the consequence of something else.

rois1995:
Part of the problem was that I was considering the tke in the computation of the boundary conditions but not inside the domain since I was using the modified version of the SST model. With this, the vertical velocity at the inlet is in the range of 10^-12. The problem on the bottom symmetry still remains though.

rois1995:
It seems that there is a problem also in the develop branch with the SA model. The Mach number at the lower symmetry BC is higher than the inlet mach number. It does not happen straight away but after some hundreds of iterations.

LowerSymmetrySA

YairMO:
Hi,

Could you please plot the eddy viscosity? How are the wall distance and tilde{\nu} defined at symmetry BC?

rois1995:
The wall distance is correctly computed. The nu_tilde, however, is higher on the symmetry plane, just like the Mach number. And, again, it happens only at the lower symmetry boundary.

LowerSymmetrySA_NuTilde

YairMO:
Based on your input (unexpected generation of the eddy viscosity at the lower symmetry BC), I would suspect the wall distance calculation. Can you tell how the wall distance is calculated in the cells above the symmetry section? Consider a cell above the symmetry section. Is its wall distance taken relative to the symmetry part or relative to the wall part?

rois1995:
It is taken relative to the wall part (which you can see as a straight white line. I don't know if the increase in nu_tilde causes an increase of the vertical velocity on the symmetry or it's the opposite.

LowerSymmetrySA_WD

YairMO:
Yes, the WD is as it should be; thank you. I also agree: who comes first, the egg or the chicken? I'll try to run this case using my code.

rois1995:
I think the problem is in the fix for the symmetry boundary in PR #2194. I tried with the develop at the last commit, at the commit before the implementation of the correction and at the commit of the implementation. It is clearly visible that something is happening with the symmetry BC correction. I will look further into it

CommitsComparison_RMSRho

rois1995:
I tried without the MUSCL reconstruction and with the AC commit and both Mach and nu_tilde profiles are correct at the symmetry plane. @bigfooted  can you comment on this please?

rois1995:
I think I found out the problem. In the correction of the gradients on the symmetry plane, somehow the gradient of the velocity along the normal direction was still not 0. By fixing it then the Mach number and nu_tilde profile are correct. I don't know if I am still missing the correction in the tangential direction, but at least the solution improves.

Here you can see the same graph as before but without the line of the last commit. I have added, instead, a try with the Weighted Least Square (LS) and a first order (FO) (no MUSCL). In the end I have added the line for the correction implemented (Corr Grad). We can see that the behaviour before the implementation of the gradient correction is recovered.

CommitsComparison_RMSRho_New

SU2_CFD/include/gradients/correctGradientsSymmetry.hpp
```diff
@@ -61,17 +61,25 @@ inline void correctGradient(const size_t varBegin, const size_t varEnd, const in
...
+    //  * dV = dV - (I - n.n^T) dV n.n^T - n.n^T dV (I - n.n^T) ---*/
 
+    // for (size_t iDim = 0; iDim < nDim; iDim++) {
+    //   for (size_t jDim = 0; jDim < nDim; jDim++) {
+    //     gradients[idxVel + iDim][jDim] -= normalGrad[iDim] * n[jDim] + n[iDim] * gradNormalVel[jDim];
+    //     gradients[idxVel + iDim][jDim] += 2 * n[iDim] * normalGradNormalVel * n[jDim];
+    //   }
+    // }
+
+    // Do I really need all of the previous code? Can't I just merge the two for loops? Maybe I am missing the tangential direction
```
pcarruscag, line 78: Yes what you are doing neglects that velocity is a vector quantity

rois1995: The previous implementation was actually correct, I'll revert to that one. Nevertheless something is off.

rois1995:
I've tried reducing the residuals up to -11.5 for the logarithm of the density residual. I've started on the coarsest grid as you saw in the previous plot. I've proceeded on the L2 mesh keeping either the tangential or the first cell spacing constant, using as restart the converged solution on the L1 mesh. The results are essentially the same as with L1 mesh. When using the NK solver straight away the AC branch does not converge, while the BC one easily converged. If I remove the NK solver, both converge. The same has happened on the L3 grid (restart from converged L2 results). 

Here you can see the residuals for the L2 grid (either the case, he outcome is the same)


CommitsComparison_RMSRho_Grid2

pcarruscag:
Thanks for testing, part of the NK options is a period where instead of NK we use the normal quasi-Newton approach.
I can't give too much detail on this, but we lack a good startup strategy, ideas from papers on NK approached could be useful even for the normal solver.

rois1995:
But what do you think of the changes between the two commits? It doesn't seem that the convergence is improved when using the new symmetry BC implementation.

YairMO:
Could you please share your L3 grid (together with your input file)?

pcarruscag:
The changes to the symmetry BC were to improve the accuracy, especially when there are walls perpendicular to the symmetry plane.
It's possible the Jacobians can be improved to be more consistent with the RHS and recover some of the convergence rate.

rois1995:
@YairMO in the zip file you can find the .geo file to create the mesh via gmsh, the L3 mesh and the configs file used for simulations with SA.
[MeshAndConfig.zip](https://github.com/user-attachments/files/17174513/MeshAndConfig.zip)

rois1995:
If you see the stagnation in residuals on the L1 meshes while not using the NK solver (line blue and yellow), it is due to a single point in the grid that you can see here. I checked the mesh quality in that point and it doesn't seem to be bad, or at least not worse than the other points.

StrangeResiduals

YairMO:
Hi,

I'm using SU2 8.1.0 but with some differencing in SA implicit treatment (underwork).
I've run the L3 grid with two different input files that differ only by the number of linear solver iterations. Attached is the convergence of the density 

SWBLI-SU2-Conv

rois1995:
Impressive! Can you share some info on your implementation? What about the CFL values and the computational time?

YairMO:
Thank you. 
The input files that I've used are based on your file. As shown in the figure, mine differs only by the number of linear solver iterations. 
As for the algorithm, I truly apologize; at this moment, I can't share this, but we will soon be able to (we are working on this). Nevertheless, my point in this exercise is to check in your simulation how the number of linear solver iterations affects the convergence of the non-linear problem, especially that of the 77fed34-NK run. 

As for the VFE-2 mesh, I've obtained this mesh from someone else. Please let me check with him first (I believe there will be no problem. Sorry).

rois1995:
Sure, thank you for checking! 

Regarding the SWBLI case, if I understood correctly, you have used your code with NK solver from the start and you changed the number of linear solver iterations, right?

YairMO:
Yes, your understanding is correct.
As for the SST ... checking

YairMO:
Here are the results obtained using the SST2003 model

SWBLI-Conv-SST2003

rois1995:
It seems that yoru implementation affects also the SST model, is it correct? Nevertheless, huge improvement wrt both this branch and develop branch results (density residual stagnates around -10)!

YairMO:
Yes, it also affects the SST model. Thanks

YairMO:
Hi Rois1995,

I have permission to share the VFE-2 grid; please send me your private email.

rois1995:
That's perfect, thank you! andrea.rausa@polimi.it

rois1995:
I have a few updates. I found out that most of my FGMRES problems in this branch where related to the fact that I was using single precision for the linear system. You can see it in this graphs for the residuals of Rho.

RMSRho_Mesh_3

The first two curves are in mixed-precision and they stop long before reaching the minimum residual required due to divergence of the linear solver.

Moreover, it seems that the 2003 model here implemented (which, simply put, considers the full reynolds stress tensor for the computation of the production of k and always considers k in the stress tensor and in the thermodynamic variables) has faster convergence and increases the recirculating zone (maybe due to the reduction of turbulence kinetic energy) which is a feature that has also been seen in [DOI:10.1017/aer.2020.93]. 

SFC_Mesh_3

There are some differences between the use of the TMR boundary conditions and the ones used before.

The results improve with respect to the develop in the prediction of the SFC distribution, which has results completely off from the V&V page of SU2

SFC_Mesh_3_Old

I may keep on checking if the implementation of the v2003 model is correct or not by searching for other test cases (probably coming from DOI:10.1017/aer.2020.93).

YairMO:
How can one switch between single or double precision?

rois1995:
when you execute the meson command use the flag -Denable-mixedprec=true

bigfooted:
I am trying to catch up on the developments in this PR but I've lost the overview. Are there still issues that we need to figure out? It looks like convergence is now recovered and results look OK? 
Regarding the symmetry plane, as @pcarruscag mentioned it is improving the accuracy, but not necessarily the convergence. @pcarruscag and I tried to get it to converge as well as before, but no guarantees. Do you still see issues with the new symmetry plane?
Regarding the peak residual that you observed in a single node in the mesh: This is a common observation when using Green-Gauss, which gives O(0) accuracy in mixed mesh nodes where quadrilateral and triangular meshes meet. There are some methods that try to fix this, although none of them are holy grails and the best methods are the most expensive ones.

rois1995:
The 2003 version of SST seems to be working fine. However, the 1994 version is not. I get Mach 7 values on the upper symmetry boundary even on the coarser mesh. I am trying to understand what is happening.

rois1995:
Maybe a stupid question, but I noticed that if I print the coordinates of the i-th point in the CTurbSSTSolver class and the problem is 2D, then the z-coordinate is not 0 but has random values. Does this affects anything in the code or it is not important since the problem is 2D?

pcarruscag:
The z coordinate is not allocated for 3D, for you are either reading out of bounds or reading the x or y value from somewhere else.

rois1995:
That's what I thought, thanks for confirming it.

rois1995:
One more thing: should I change the naming scheme for the model version of SST? As it is now, the naming scheme is V+model version but the NASA TMR site says to use the V only for the vorticity form of the production term. Should we remove the V then, or maybe set it as lower case? This will break all of the configs, no problem for me to go through them all and change it though.

YairMO:
Chris Rumsey made a tremendous effort to standardize popular RANS turbulence models (via the TMR website).
In my opinion, follow his work. Namely, change the present names to exactly what is given on the TMR.

pcarruscag:
No changes that break backward compatibility.

rois1995:
That's what I thought.

Common/include/CConfig.hpp
```diff
@@ -865,7 +865,7 @@ class CConfig {
   nMu_Temperature_Ref,           /*!< \brief Number of species reference temperature for Sutherland model. */
   nMu_S,                         /*!< \brief Number of species reference S for Sutherland model. */
   nThermal_Conductivity_Constant,/*!< \brief Number of species constant thermal conductivity. */
-  nPrandtl_Lam,                  /*!< \brief Number of species laminar Prandtl number. */
```
bigfooted, line 868: The original description was correct, for species transport we can have a Prandtl number for each of the species

Common/include/CConfig.hpp
```diff
@@ -865,7 +865,7 @@ class CConfig {
...
+  nPrandtl_Lam,                  /*!< \brief Prandtl number. */
```
bigfooted, line 868:
```suggestion
  nPrandtl_Lam,                  /*!< \brief Number of species laminar Prandtl number. */
```

Common/include/CConfig.hpp
rois1995: Yeah I do not recall changing that but for sure it was wrong in my branch. Corrected, thanks!

PR #2346, opened 2024-08-27
[GSOC24] Addition of CUDA and GPU Acceleration to FGMRES Linear Solver in SU2

areenraj:
## Proposed Changes

This is the modified version of SU2 code that supports CUDA usage for the FGMRES solver and the use of NVBLAS. The main focus is the offloading of the Matrix Vector Product in the FGMRES solver to the GPU using CUDA Kernels. This implementation shows promise with marginally better run times (all benchmarks were carried out with the GPU Error Checking switched off and in debug mode to check if the correct functions were being called).

The use of NVBLAS is secondary and while functionality has been added to make it usable, it is not activated as it doesn't cause an appreciative increase in performance.

## Compilation and Usage

Compile using the following MESON Flag
```
-Denable-cuda=true
```
And activate the functions using the following Config File Option
```
ENABLE_CUDA=YES
```

## NOTE ON IMPLEMENTATION

I've decided to go with a single version of the code where the CPU and GPU implementations co-exist in the same linear solver and can be disabled or switched using a combination of Meson and Config File options. This is why I have defined three classes - one over-arching class that is named CExecutionPath that has two child classes - CCpuExecution and CGpuExecution. These child classes contain the correct member function for each path - CPU or GPU functioning. 

All of this could also be easily achieved with an if statement that switches between the two - but that particular implementation will access and run the statement for each call. In our case once a Matrix Vector Product object is created, it will immediately know whether to use CPU or GPU mode of execution. 

Recommendations are most welcome to improve or make this implementation better

## PR Checklist

Warning Levels do come (only at level 3) but they are all of the following type

```
style of line directive is a GCC extension
```

The documentation for compiling with CUDA needs to be added by forking the SU2 site repo and adding the relevant changes to it? Or do I need to contact someone to change things on the site itself?

DOxygen Documentations and config template are all updated.


Common/src/linear_algebra/GPU_lin_alg.cu
```diff
@@ -0,0 +1,103 @@
...
+
+  unsigned long mat_size = nnz*nVar*nEqn;
+  unsigned long vec_size = nPointDomain*nVar;
+
+  gpuErrChk(cudaMalloc((void**)(&d_vec), (sizeof(ScalarType)*vec_size)));
+  gpuErrChk(cudaMalloc((void**)(&d_prod), (sizeof(ScalarType)*vec_size)));
+
+  gpuErrChk(cudaMemcpy((void*)(d_matrix), (void*)&matrix[0], (sizeof(ScalarType)*mat_size), cudaMemcpyHostToDevice));
+  gpuErrChk(cudaMemcpy((void*)(d_vec), (void*)&vec[0], (sizeof(ScalarType)*vec_size), cudaMemcpyHostToDevice));
+  gpuErrChk(cudaMemcpy((void*)(d_prod), (void*)&prod[0], (sizeof(ScalarType)*vec_size), cudaMemcpyHostToDevice));
```
pcarruscag, line 86: you don't need to copy the product, you just need to memset to 0

Common/src/linear_algebra/GPU_lin_alg.cu
```diff
@@ -0,0 +1,103 @@
...
+
+  double xDim = (double) 1024.0/(nVar*nEqn);
+  dim3 blockDim(floor(xDim), nVar, nEqn);
+  double gridx = (double) nPointDomain/xDim;
+  dim3 gridDim(ceil(gridx), 1, 1);
+
```
pcarruscag, line 92: Can you document the choice of work distribution between blocks and threads?

Common/src/linear_algebra/GPU_lin_alg.cu
```diff
@@ -0,0 +1,103 @@
...
+   if(i<nPointDomain)
+   {
+      for(int index = d_row_ptr[i]; index<d_row_ptr[i+1]; index++)
+      {
+        int matrix_index = index * nVar * nEqn;
+        int vec_index = d_col_ind[index] * nEqn;
+
+        res += matrix[matrix_index + (j * nEqn + k)] * vec[vec_index + k];
+      }
```
pcarruscag, line 62: Is this based on some publication? Did you experiment with other divisions of work?
For example, I see you are going for coalesced access to the matrix blocks, but this requires multiple reads of the same vector entries.

pcarruscag (commented):
Nice work

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: Good catch, will add this

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: [Full report with Benchmark Results and other Explanations](https://github.com/areenraj/GSoC-24/blob/d011b609d17f111d1918929f1f49adc158b05c4b/REPORT.md)

[Complete definition of how the Kernel is initialized](https://github.com/user-attachments/assets/e831d942-332f-408f-abf1-7f637834cb68)

[Explanation as to why the blocks are multi-directional](https://github.com/user-attachments/assets/82bb626f-d501-46e8-a31e-1f29dacd3045)

[Algorithm for Block Matrix Calculations](https://github.com/user-attachments/assets/c013ace4-991d-4cd3-86c6-56fcaffcccc3)

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: I haven't experimented with different implementations as I went with this because it seemed optimal. It does access the same vector elements repeatedly while going through an entire row. 

Haven't particularly checked out publications yet, if you do have some recommendations on improvements then let me know.

pcarruscag: You can do something like a flag in CSysMatrix that is set to true when the matrix is uploaded to GPU and set to false when the matrix changes (for example when we clear the matrix to write new blocks).
You can use pinned host memory to make the transfers faster.
You can try uploading the matrix in chunks and overlap the uploads with the CPU work of filling another chunk.
Ultimately, the issue of transferring the matrix only goes away by porting the entire code :sweat_smile:

pcarruscag: Regarding recommendations I have to be a bit cryptic because of my current job, but the general goals are coalesced access, and avoid reading or writing the same global memory location more than once.
I read this paper before my current job.
Optimization of Block Sparse Matrix-Vector Multiplication on Shared-Memory
Parallel Architectures
But like you said, there is much more to gain by porting more linear algebra operations than to micro-optimize the multiplications.

areenraj: I'll go through the paper and follow up the pinned memory lead

Will continue to work on porting the solver, lets hope we get some interesting work in time for the conference :smile:

If necessary, I'll contact you with updates either on this thread or catch you in the next dev meeting. Thanks for the help Pedro

kursatyurt:
I could not follow the work here; I just watched the conference presentation. As much work has been done, have you considered using an existing linear algebra library such as https://github.com/ginkgo-project/ginkgo?

AFAIK the only thing you need to do is copy the matrix to Gingko format on GPU. Then Gingko will provide an efficient, scalable solver that works not only on NVIDIA but also on AMD and Intel. 

GMRES and ILU preconditioners are available there, so it is pretty much ready to go for all problems.

areenraj:
@kursatyurt Hello, thank you so much for the lead.

Our initial scope mostly involved writing our own kernels and I did explore some libraries at the start - I was planning on using CUSP as well but my main concern was its lack of being updated to the newly compatible versions of the toolkit. cuSolver and cuBLAS do exist, but I chose to go ahead with a "simple" kernel implementation to have more control. I also felt that if I could keep the block size of the grid in optimal territory then they could be just as fast as those options (please do correct me if my reading of the literature or the situation was incorrect)

I was not aware of Ginkgo and I will surely give it a go and try to produce some comparative results. I am currently super busy for this month and will get to working on the code with some delay.

Again, thank you for the lead!

areenraj:
Also could you mention what you meant by "you could not follow the work here" If there is a specific doubt in the work then I would love to clarify it over slack whenever I get the time

kursatyurt:
> Also could you mention what you meant by "you could not follow the work here" If there is a specific doubt in the work then I would love to clarify it over slack whenever I get the time

I am not very familiar with the linear solver implementation in SU2 not about the work itself

kursatyurt:
> @kursatyurt Hello, thank you so much for the lead.
> 
> Our initial scope mostly involved writing our own kernels and I did explore some libraries at the start - I was planning on using CUSP as well but my main concern was its lack of being updated to the newly compatible versions of the toolkit. cuSolver and cuBLAS do exist, but I chose to go ahead with a "simple" kernel implementation to have more control. I also felt that if I could keep the block size of the grid in optimal territory then they could be just as fast as those options (please do correct me if my reading of the literature or the situation was incorrect)

To learn the basics, it's a good idea, but for large-scale projects, I prefer using existing libraries if possible.
Those libraries generally exploit state-of-the-art solution like mixed-precision computing. A gaming GPU is not way faster than a good CPU in double precision,  but way faster in single precision, most of them have 64:1 ratio, however server class GPU have 2:1 ratio. Also when available they use vendor libraries like cuBLAS or hipBLAS. It is always nice to have you only care about connection and somebody else handle the solver as performant as possible. In future probably they will provide more and more solvers and it will be automagically works.

It is kind of light-weight too, not a huge dependency like Trilinos or PETSc. 
> 
> I was not aware of Ginkgo and I will surely give it a go and try to produce some comparative results. I am currently super busy for this month and will get to working on the code with some delay.
> 
> Again, thank you for the lead!

I can test on various GPUs  (P100/V100/A100 and 4070Mobile)  on single node multi-gpu etc.

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: > Regarding recommendations I have to be a bit cryptic because of my current job, but the general goals are coalesced access, and avoid reading or writing the same global memory location more than once. I read this paper before my current job. Optimization of Block Sparse Matrix-Vector Multiplication on Shared-Memory Parallel Architectures But like you said, there is much more to gain by porting more linear algebra operations than to micro-optimize the multiplications.

This change is finished and I have updated the algorithm, I have also synced the changes from the latest SU2 update. If you try it out then let me know. I have used the second algo from the paper you mentioned - Assigning a warp to the matrix row and iterating through the block column-wise. This should increase performance (although I am I still testing it).

Now that the algorithm is down I'll start using it in the port and begin working on transferring FGMRES to GPU

pcarruscag: FGMRES does not need to be transferred to gpu

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -131,6 +147,23 @@ void CSysMatrix<ScalarType>::Initialize(unsigned long npoint, unsigned long npoi
...
 
+  /*--- Allocate data. ---*/
+  auto allocAndInit = [](ScalarType*& ptr, unsigned long num) {
+    ptr = MemoryAllocation::aligned_alloc<ScalarType, true>(64, num * sizeof(ScalarType));
+  };
+
+#if defined(HAVE_CUDA)
+  gpuErrChk(cudaMalloc((void**)(&d_row_ptr), (sizeof(row_ptr) * (nPointDomain + 1.0))));
+  gpuErrChk(cudaMalloc((void**)(&d_col_ind), (sizeof(col_ind) * nnz)));
+  gpuErrChk(cudaMallocHost((void**)(&matrix), (sizeof(ScalarType) * nnz * nVar * nEqn)));
```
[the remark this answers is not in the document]
areenraj: This prevents cuda Memory functions like Malloc to be called by the compiler when the user doesn't have CUDA. The conditional is only executed when the user specifies that they would like to compile with cuda and have the toolkit along with nvcc installed. If I leave it as is then when compiling with just gcc will throw an exception.

areenraj:
> Is your idea to return to this project as part of GSOC or just personal interest?

@pcarruscag No, I won't be applying to GSoC any time in the future. Its just that I started a work and I would like to see this through. If anyone is selected for GPU Acceleration this time then I would be more than happy to work and collaborate with them to make sure the project receives as much help as it can.

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: Hmm, I'm a bit confused here. My entire project and the previous pull request's purpose has been to slowly transfer the FGMRES solver to GPU.

Do you mean that we should just port the helper functions like Matrix Vector Product and other linear algebra functions and let the solver decide based on user given parameters?

pcarruscag: Correct, the same way the FGMRES doesn't need to know how to multiply matrices and vectors, it doesn't need to know anything about GPUs

areenraj: Got it, I will get on this and will also discuss some things I have in mind during the dev meet

Common/src/linear_algebra/CSysMatrix.cpp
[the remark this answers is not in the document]
pcarruscag: But you removed d_matrix, so now the GPU is working directly from host memory which is probably slow.
Also if you read the docs for cudaMallocHost it's not advised to use large chunks of pinned memory, and the matrix is the largest block of memory in SU2...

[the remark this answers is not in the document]
areenraj: > But you removed d_matrix, so now the GPU is working directly from host memory which is probably slow.

My reasoning was that when pinnned memory is allocated on 64 bit OSes, the memory is also simultaneously mapped to the GPU as the conditions for universal virtual addressing are met. This method proves to be faster than the regular MemCpy when the data address is only read/written to once. In our case, the Mat Vec Kernel only read the matrix values once per CudaMemCpy - so I switched it out this way. If my understanding of this was wrong then do let me know. 

But now that we will be working on porting the other linear algebra functions it will make more sense to use the CudaMemCpy approach of transferring the pinned memory to the GPU. As the same variables will be accessed multiple times over the course of a single memory transfer

> Also if you read the docs for cudaMallocHost it's not advised to use large chunks of pinned memory, and the matrix is the largest block of memory in SU2

i never ran into a problem with this as the cases I ran didn't take a performance hit. But I can understand that with increase in problem size the pinned memory could decrease the total available system memory and bottleneck the performance of the solver. Would you recommend using pinned memory for the input and output vectors then?

Common/src/linear_algebra/GPU_lin_alg.cu
areenraj: > ...This should increase performance (although I am still testing it).

NSight shows the new Kernel is twice as fast. Although this doesn't mean anything for overall performance as the bottleneck is still memory transfer.

areenraj:
## 29 Mar 2025 Commits
The commits currently made aim to provide more control over memory access and transfer so that we can make further changes to the surrounding linear algebra functions which make up the FGMRES Solver.

### Streamlined Vector Storage in the GPU
Each vector definition now also creates its corresponding analog in the GPU Memory space. This is done by allocating memory for it using cudaMalloc when the Initialize function of the CSysVector Class is called. This allows for the continuous storage of vector data in the GPU memory in between calls of the Matrix Vector Product and other linear algebra functions. The previous implementation only allowed this data to persist during a single call of the Matrix Vector Product and it had to be refreshed for each call. 

This implementation was similar to how the matrix was stored previously. Saving the matrix in pinned memory is also removed due to its huge size as pointed out by earlier feedback. The device pointer can be accessed at any point using a new dedicated public member function (GetDevicePointer) of the CsysVector Class.

### Added Memory Transfer Control
As previously discussed, we needed more control over the memory transfer in between calls so that a larger load of the computation could be carried out on the GPU without multiple memory transfers. Now these transfers are carried out by member functions with a flag built into them to decide whether the copy needs to be carried out or not. This flag is set to true by default and does not need to be specified all the time. 

Further changes are necessary to actually use this flag to decrease the frequency of memory transfer - namely a variable that allows the inner loop of FGMRES to communicate with the MatrixVectorProduct function to know when to switch the flag on or off. This will be added after I port the preconditioner.

### Minor Change - Standardized File Structure Slightly
Redundant .cuh header files are now gone. I have added a GPUComms.cuh file so that any functions that need to be accessed by all CUDA files - like error checking - can be added here for future references. I've also added GPUVector and GPUMatrix files here - each containing the cuda wrapping member functions for the CSysVector and CSysMatrix class respectively.

Please let me know if you notice any bugs or if my implementations can be improved in any way.

areenraj:
I'll now move forward and start porting the LU_SGS Preconditioner

Common/src/linear_algebra/GPUVector.cu
pcarruscag, line 1: Rename to CSysVectorGPU or something similar, this implements functions of CSysVector so it makes more sense to name it consistently with the CPU version.

Common/include/linear_algebra/CMatrixVectorProduct.hpp
```diff
@@ -60,6 +66,50 @@ class CMatrixVectorProduct {
...
+class CGpuExecution : public CExecutionPath<ScalarType> {
+ public:
+  void mat_vec_prod(const CSysVector<ScalarType>& u, CSysVector<ScalarType>& v, CGeometry* geometry,
+                    const CConfig* config, const CSysMatrix<ScalarType>& matrix) override {
+#ifdef HAVE_CUDA
+    matrix.GPUMatrixVectorProduct(u, v, geometry, config);
+#else
+    std::cerr << "\nError in launching Matrix-Vector Product Function\n";
+    std::cerr << "ENABLE_CUDA is set to YES\n";
+    std::cerr << "Please compile with CUDA options enabled in Meson to access GPU Functions" << std::endl;
```
pcarruscag, line 108: SU2_MPI::Error(...

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,21 @@ class CSysMatrix {
...
+   * \param[in] config - Definition of the particular problem.
+   * \param[out] prod - Result of the product.
+   */
+
+  void GPUMatrixVectorProduct(const CSysVector<ScalarType>& vec, CSysVector<ScalarType>& prod, CGeometry* geometry,
+                              const CConfig* config) const;
+
+  void FGMRESMainLoop(std::vector<ScalarType> W, std::vector<ScalarType> Z, su2vector<ScalarType>& g,
+                      su2vector<ScalarType>& sn, CSysVector<ScalarType>& cs, su2vector<ScalarType>& y,
+                      su2vector<ScalarType>& H, int m, CGeometry* geometry, const CConfig* config) const;
```
pcarruscag, line 864: This seems unused
```suggestion
```

Common/src/linear_algebra/CSysVector.cpp
```diff
@@ -52,6 +56,11 @@ void CSysVector<ScalarType>::Initialize(unsigned long numBlk, unsigned long numB
 
   if (vec_val == nullptr) vec_val = MemoryAllocation::aligned_alloc<ScalarType, true>(64, nElm * sizeof(ScalarType));
 
+#if defined(HAVE_CUDA)
+  gpuErrChk(cudaMalloc((void**)(&d_vec_val), (sizeof(ScalarType) * nElm)));
+  gpuErrChk(cudaMemset((void*)d_vec_val, 0.0, (sizeof(ScalarType) * nElm)));
```
pcarruscag, line 61: Similarly to how we have `MemoryAllocation::aligned_alloc` and free, please create wrappers for the cuda-specific functions so that we avoid #ifdef HAVE_CUDA in many places.

Common/src/linear_algebra/GPUMatrix.cu
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+#include "../../include/linear_algebra/CSysMatrix.hpp"
+#include "../../include/linear_algebra/GPUComms.cuh"
+
+template<typename matrixType, typename vectorType>
+__global__ void GPUMatrixVectorProductAdd(matrixType* matrix, vectorType* vec, vectorType* prod, unsigned long* d_row_ptr, unsigned long* d_col_ind, unsigned long nPointDomain, unsigned long nVar, unsigned long nEqn)
+{
+   int row = (blockIdx.x * blockDim.x + threadIdx.x)/32;
+   int threadNo = threadIdx.x%32;
```
pcarruscag, line 35: Define 32 (threads per warp) as a constant somewhere.

Common/src/linear_algebra/GPUMatrix.cu
```diff
@@ -0,0 +1,89 @@
...
+   ScalarType* d_vec = vec.GetDevicePointer();
+   ScalarType* d_prod = prod.GetDevicePointer();
+
+   HtDTransfer();
+   vec.HtDTransfer();
+   prod.GPUSetVal(0.0);
+
+  dim3 blockDim(1024,1,1);
+  double gridx = (double) nPointDomain/32.0;
+  gridx = double(ceil(gridx));
```
pcarruscag, line 79: We have functions to round up integer division in the memory allocation toolbox.
Define 1024 as the default block size somewhere too.

pcarruscag (commented):
Can you fix the merge conflicts so the regression tests can run in this PR?

Common/include/linear_algebra/CMatrixVectorProduct.hpp
areenraj: This is fixed, however in the gpuErrChk function I have left it as fprintf since the the string has some placeholders to it. I could use a temporary character array along with SPRINTF to use SU2_MPI::Error. But I am not sure whether allocating and de-allocating the temporary character array repeatedly will slow down performance or not. Since, gpuErrChk is called every time a CUDA API function is called.

Common/src/linear_algebra/CSysVector.cpp
areenraj: Done, please let me know if my handling of null pointers was accurate

areenraj:
> Can you fix the merge conflicts so the regression tests can run in this PR?

It shows that merging is blocked (I guess a review is required?) Also please let me know how you wish for me to keep the fork synced with changes in the original - either through merge or rebase.

Common/include/linear_algebra/CMatrixVectorProduct.hpp
areenraj: Just realized that this code snippet got removed when I removed the virtual classes. It won't affect anything but I'll add it in the next iteration that I submit.

Common/include/linear_algebra/CMatrixVectorProduct.hpp
```diff
@@ -94,6 +101,14 @@ class CSysMatrixVectorProduct final : public CMatrixVectorProduct<ScalarType> {
...
   inline void operator()(const CSysVector<ScalarType>& u, CSysVector<ScalarType>& v) const override {
+#ifdef HAVE_CUDA
+    if (config->GetCUDA()) {
+      matrix.GPUMatrixVectorProduct(u, v, geometry, config);
+    } else {
+      matrix.MatrixVectorProduct(u, v, geometry, config);
+    }
+#else
     matrix.MatrixVectorProduct(u, v, geometry, config);
+#endif
```
pcarruscag:
```suggestion
      return;
    }
#endif
    matrix.MatrixVectorProduct(u, v, geometry, config);
```

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,53 @@ class CSysMatrix {
...
                            const CConfig* config) const;
 
+  /*!
+   * \brief Performs the product of a sparse matrix by a CSysVector.
+   * \param[in] vec - CSysVector to be multiplied by the sparse matrix A.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[out] prod - Result of the product.
+   */
+
```
pcarruscag:
```suggestion
```

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,53 @@ class CSysMatrix {
...
+  void GPUMatrixVectorProduct(const CSysVector<ScalarType>& vec, CSysVector<ScalarType>& prod, CGeometry* geometry,
+                              const CConfig* config) const;
+
```
pcarruscag, line 862:
```suggestion
```

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,53 @@ class CSysMatrix {
...
+  /*!
+   * \brief Performs first step of the LU_SGS Preconditioner building
+   * \param[in] vec - CSysVector to be multiplied by the sparse matrix A.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[out] prod - Result of the product.
+   */
+
```
pcarruscag:
```suggestion
```

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,53 @@ class CSysMatrix {
...
+  void GPUFirstSymmetricIteration(ScalarType& vec, ScalarType& prod, CGeometry* geometry, const CConfig* config) const;
+
+  /*!
+   * \brief Performs second step of the LU_SGS Preconditioner building
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[out] prod - Result of the product.
+   */
+
```
pcarruscag:
```suggestion
```

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -838,6 +848,53 @@ class CSysMatrix {
...
+  void GPUSecondSymmetricIteration(ScalarType& prod, CGeometry* geometry, const CConfig* config) const;
+
+  /*!
+   * \brief Performs Gaussian Elimination between diagional blocks of the matrix and the prod vector
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] config - Definition of the particular problem.
+   * \param[out] prod - Result of the product.
+   */
+
```
pcarruscag:
```suggestion
```

Common/include/linear_algebra/CSysVector.hpp
```diff
@@ -195,6 +198,29 @@ class CSysVector : public VecExpr::CVecExpr<CSysVector<ScalarType>, ScalarType>
...
+
+  /*!
+   * \brief Sets all the elements of the GPU vector to a certain value
+   * \param[in] trigger - boolean value that decides whether to conduct the transfer or not. True by default.
+   */
+  void GPUSetVal(ScalarType val, bool trigger = true) const;
+
+  /*!
+   * \brief return the number of local elements in the CSysVector
+   */
```
pcarruscag: Wrong comment

Common/include/linear_algebra/GPUComms.cuh
```diff
@@ -0,0 +1,55 @@
...
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include<cuda_runtime.h>
+#include"iostream"
```
pcarruscag:
```suggestion
#include <cuda_runtime.h>
#include <iostream>
```

Common/include/linear_algebra/GPUComms.cuh
```diff
@@ -0,0 +1,55 @@
...
+
+namespace KernelParameters{
+
+  inline constexpr int round_up_division(const int multiple, int x) { return ((x + multiple - 1) / multiple); }
+
+  const int MVP_BLOCK_SIZE = 1024;
+  const int MVP_WARP_SIZE = 32;
```
pcarruscag:
```suggestion
inline constexpr int round_up_division(const int multiple, int x) { return ((x + multiple - 1) / multiple); }

static constexpr int MVP_BLOCK_SIZE = 1024;
static constexpr int MVP_WARP_SIZE = 32;
```

Common/include/linear_algebra/GPUComms.cuh
```diff
@@ -0,0 +1,55 @@
...
+}
+/*!
+  * \brief assert style function that reads return codes after intercepting CUDA API calls.
+  *        It returns the result code and its location if the call is unsuccessful.
+  * \param[in] code - result code of CUDA function
+  * \param[in] file - name of file holding the function
+  * \param[in] line - line containing the function
+  */
+
```
pcarruscag:
```suggestion
 * \brief assert style function that reads return codes after intercepting CUDA API calls.
 *        It returns the result code and its location if the call is unsuccessful.
 * \param[in] code - result code of CUDA function
 * \param[in] file - name of file holding the function
 * \param[in] line - line containing the function
 */
```

Common/include/linear_algebra/GPUComms.cuh
```diff
@@ -0,0 +1,55 @@
...
+inline void gpuAssert(cudaError_t code, const char *file, int line, bool abort=true)
+{
+  if (code != cudaSuccess)
+  {
```
pcarruscag:
```suggestion
inline void gpuAssert(cudaError_t code, const char *file, int line, bool abort=true) {
  if (code != cudaSuccess) {
```

Common/include/toolboxes/allocation_toolbox.hpp
```diff
@@ -90,3 +94,55 @@ inline void aligned_free(T* ptr) noexcept {
...
+ * \return Pointer to memory, always use gpu_free to deallocate.
+ */
+template <class T>
+inline T* gpu_alloc_cpy(T* src_ptr, size_t size) noexcept {
+  void* ptr = nullptr;
+
+#ifdef HAVE_CUDA
+  gpuErrChk(cudaMalloc((void**)(&ptr), size));
+  gpuErrChk(cudaMemcpy((void*)(ptr), (void*)src_ptr, size, cudaMemcpyHostToDevice));
+  ;
```
pcarruscag:
```suggestion
```

Common/include/toolboxes/allocation_toolbox.hpp
```diff
@@ -90,3 +94,55 @@ inline void aligned_free(T* ptr) noexcept {
...
+#else
+  return 0;
```
pcarruscag:
```suggestion
```

Common/include/toolboxes/allocation_toolbox.hpp
```diff
@@ -90,3 +94,55 @@ inline void aligned_free(T* ptr) noexcept {
...
+  gpuErrChk(cudaFree((void*)ptr));
+#endif
+}
+/*!
+ * \brief Memory allocation for variables on the GPU along with initialization from a source host array.
+ * \param[in] size in bytes.
+ * \return Pointer to memory, always use gpu_free to deallocate.
+ */
+template <class T>
+inline T* gpu_alloc_cpy(T* src_ptr, size_t size) noexcept {
```
pcarruscag:
```suggestion
inline T* gpu_alloc_cpy(const T* src_ptr, size_t size) noexcept {
```

TestCases/gpu/flatplate/lam_flatplate.cfg
pcarruscag, line 1: Let's remove this since it's not used for any test

pcarruscag (approved):
:+1: address the minor comments + the version updates we just talked about and let's merge :partying_face:

Common/include/linear_algebra/CMatrixVectorProduct.hpp
areenraj: made some slight tweaks and fixed this

aladdin-afk:
damn , a fellow bitsian everywhere

PR #2352, opened 2024-09-06
Corrections to SA implementation

rois1995:
## Proposed Changes
The CRot value was set to 1.0 for the HLPW5 but the standard value is 2.0 as suggested in TMR. Plus, there may be some inconsistencies with the negative branch of SA when used with the R variant. $\hat{S}$ should be used in the computation of the production term when $\hat{\nu}>0$. When $\hat{\nu}<0$, instead, the vorticity $\Omega$ should be used together with the $c_{t3}$ constant. Citing TMR:

"This model is the same as the "standard" version (SA), except that the (SA-R) production term becomes: 

$c_{b1}(1-f_{t2})[\hat{S}+C_{rot}min(0,S-\Omega)]\hat{\nu}$

...

However, a negative production term is not acceptable for the negative branch ($\hat{\nu} < 0$) of the (SA-neg) model. If coding (SA-neg-R), a possible C0 continuous production term in the negative (SA-neg-R) branch is the following: 

$c_{b1}(1-c_{t3})[\hat{\Omega}+C_{rot}min(0,S-\Omega)]\hat{\nu}$

"

I don't understand if the second production term has to be used everytime  SA-neg is considered or only at the points where $\hat{\nu}<0$.

If it is already taken into account, then I'll modify the PR.

## PR Checklist

### DIFF
--- a/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
+++ b/SU2_CFD/include/numerics/turbulent/turb_sources.hpp
@@ -52,7 +52,7 @@ struct CSAVariables {
   const su2double c2 = 0.7, c3 = 0.9;
 
   /*--- List of auxiliary functions ---*/
-  su2double ft2, d_ft2, r, d_r, g, d_g, glim, fw, d_fw, Ji, d_Ji, S, Shat, d_Shat, fv1, d_fv1, fv2, d_fv2;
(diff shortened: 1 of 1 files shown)


bigfooted:
No regression tests are affected, so we have no test that uses this model? Can you add a test for this as well?

pcarruscag (approved):

rois1995:
I will try and add a new test case as soon as possible, although it might be in October.

PR #2362, opened 2024-10-09
Feature CFL adaptation initial iteration

Lisa-Bachmann:
A spark initialization was not possible with CFL adaptation, as the CFL number would increase too far before the spark initiates the combustion. This led to immediate flashback or blow-off. An additional parameter in the CFL adaptation defines the iteration, at which the CFL adaption should start. This allows the combustion to stabilize before the CFL number increases. Before the initial iterations has not started, the CFL number does not get adapted in either direction.

### DIFF
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1777,11 +1777,11 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Activate The adaptive CFL number. */
   addBoolOption("CFL_ADAPT", CFL_Adapt, false);
   /* !\brief CFL_ADAPT_PARAM
-   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max), acceptable linear residual )
+   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max)[, acceptable linear residual][, starting iteration]).
    * Factor down generally <1.0, factor up generally > 1.0 to cause the CFL to increase when the under-relaxation parameter is 1.0
    * and to decrease when the under-relaxation parameter is less than 0.1. Factor is multiplicative. \ingroup Config*/
   default_cfl_adapt[0] = 1.0; default_cfl_adapt[1] = 1.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 100.0;
-  default_cfl_adapt[4] = 0.001;
+  default_cfl_adapt[4] = 0.001; default_cfl_adapt[5] = 0.0;
   addDoubleListOption("CFL_ADAPT_PARAM", nCFL_AdaptParam, CFL_AdaptParam);
   /* DESCRIPTION: Reduction factor of the CFL coefficient in the adjoint problem */
   addDoubleOption("CFL_REDUCTION_ADJFLOW", CFLRedCoeff_AdjFlow, 0.8);
@@ -7230,7 +7230,8 @@ void CConfig::SetOutput(SU2_COMPONENT val_software, unsigned short val_izone) {
       if (!CFL_Adapt) cout << "No CFL adaptation." << endl;
       else cout << "CFL adaptation. Factor down: "<< CFL_AdaptParam[0] <<", factor up: "<< CFL_AdaptParam[1]
         <<",\n                lower limit: "<< CFL_AdaptParam[2] <<", upper limit: " << CFL_AdaptParam[3]
-        <<",\n                acceptable linear residual: "<< CFL_AdaptParam[4] << "." << endl;
+        <<",\n                acceptable linear residual: "<< CFL_AdaptParam[4]
+        <<"'\n                starting iteration: "<< CFL_AdaptParam[5] << "." << endl;
 
       if (nMGLevels !=0) {
         PrintingToolbox::CTablePrinter MGTable(&std::cout);
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1781,7 +1781,7 @@ void CConfig::SetConfig_Options() {
    * Parameters in square brackets are optional, parameter "starting iteration" only valid with parameter "acceptable linear residual".
    * Factor down generally <1.0, factor up generally > 1.0 to cause the CFL to increase when the under-relaxation parameter is 1.0
    * and to decrease when the under-relaxation parameter is less than 0.1. Factor is multiplicative. \ingroup Config*/
-  default_cfl_adapt[0] = 1.0; default_cfl_adapt[1] = 1.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 100.0;
+  default_cfl_adapt[0] = 0.1; default_cfl_adapt[1] = 2.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 1e10;
   default_cfl_adapt[4] = 0.001; default_cfl_adapt[5] = 0.0;
   addDoubleListOption("CFL_ADAPT_PARAM", nCFL_AdaptParam, CFL_AdaptParam);
   /* DESCRIPTION: Reduction factor of the CFL coefficient in the adjoint problem */
--- a/Common/src/CConfig.cpp
+++ b/Common/src/CConfig.cpp
@@ -1781,8 +1781,12 @@ void CConfig::SetConfig_Options() {
    * Parameters in square brackets are optional, parameter "starting iteration" only valid with parameter "acceptable linear residual".
    * Factor down generally <1.0, factor up generally > 1.0 to cause the CFL to increase when the under-relaxation parameter is 1.0
    * and to decrease when the under-relaxation parameter is less than 0.1. Factor is multiplicative. \ingroup Config*/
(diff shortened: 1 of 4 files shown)


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1764,8 +1765,11 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
 
     /* Check if we should decrease or if we can increase, the 20% is to avoid flip-flopping. */
     resetCFL = linRes > 0.99;
-    reduceCFL = linRes > 1.2*linTol;
-    canIncrease = linRes < linTol;
+    
+    /* only change CFL number when larger than starting iteration */
+    reduceCFL = linRes > 1.2*linTol && (config->GetOuterIter() > startingIter);
+
+    canIncrease = (linRes < linTol) && (config->GetOuterIter() > startingIter);
```
pcarruscag, line 1772: using outer iter only works for multizone right?

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1764,8 +1765,11 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
 
     /* Check if we should decrease or if we can increase, the 20% is to avoid flip-flopping. */
     resetCFL = linRes > 0.99;
-    reduceCFL = linRes > 1.2*linTol;
-    canIncrease = linRes < linTol;
+
```
bigfooted, line 1768:
```suggestion
unsigned long iter = config->GetMultizone_Problem() ? config->GetOuterIter() : config->GetInnerIter();

```

SU2_CFD/src/solvers/CSolver.cpp
bigfooted: Yes, it is, my bad. Let's introduce the more general iteration counter here.

Common/src/CConfig.cpp
```diff
@@ -1777,11 +1777,11 @@ void CConfig::SetConfig_Options() {
   /* DESCRIPTION: Activate The adaptive CFL number. */
   addBoolOption("CFL_ADAPT", CFL_Adapt, false);
   /* !\brief CFL_ADAPT_PARAM
-   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max), acceptable linear residual )
+   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max)[, acceptable linear residual][, starting iteration]).
```
TobiKattmann, line 1780:
```suggestion
   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max), acceptable linear residual, starting iteration).
```
why square braces here?

Common/src/CConfig.cpp
```diff
@@ -1777,11 +1777,11 @@ void CConfig::SetConfig_Options() {
...
    * Factor down generally <1.0, factor up generally > 1.0 to cause the CFL to increase when the under-relaxation parameter is 1.0
    * and to decrease when the under-relaxation parameter is less than 0.1. Factor is multiplicative. \ingroup Config*/
   default_cfl_adapt[0] = 1.0; default_cfl_adapt[1] = 1.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 100.0;
```
TobiKattmann, line 1783:
```suggestion
  default_cfl_adapt[0] = 0.1; default_cfl_adapt[1] = 2.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 1e10;
```
I think it would be nice to have the default values be with the config_template(?), i.e. 0.1, 2.0, 10.0, 1e10, 0.001, 1 (github does not allow me to suggest that over the two lines)

Common/src/CConfig.cpp
```diff
@@ -1777,11 +1777,11 @@ void CConfig::SetConfig_Options() {
...
-  default_cfl_adapt[4] = 0.001;
+  default_cfl_adapt[4] = 0.001; default_cfl_adapt[5] = 0.0;
```
TobiKattmann, line 1784:
```suggestion
  default_cfl_adapt[4] = 0.001; default_cfl_adapt[5] = 1.0;
```

TobiKattmann (commented):
Generally lgtm, is there maybe an existing testcase where you can reflect this(?) 

Also, please grep the cfg files and add a 1 (or 0, I guess it does not matter because this would kick in only in the 2nd iteration anyway 🤔  ) to the CFL_ADAPT_PARAM field

pcarruscag:
Don't modify the configs to add default options. Just the template is enough

Common/src/CConfig.cpp
bigfooted: > why square braces here?
to indicate that it is an optional argument.

Common/src/CConfig.cpp
bigfooted: We should start at iteration 0 by default or else all adaptive cfl cases will fail.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1764,8 +1765,12 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
...
     /* Check if we should decrease or if we can increase, the 20% is to avoid flip-flopping. */
     resetCFL = linRes > 0.99;
-    reduceCFL = linRes > 1.2*linTol;
-    canIncrease = linRes < linTol;
+    unsigned long iter = config->GetMultizone_Problem() ? config->GetOuterIter() : config->GetInnerIter();
+
+    /* only change CFL number when larger than starting iteration */
+    reduceCFL = linRes > 1.2*linTol && (iter > startingIter);
+
+    canIncrease = (linRes < linTol) && (iter > startingIter);
```
bigfooted, line 1773: We should use by default start at iteration 0 or else all existing adaptive CFL will have different residuals.

```suggestion
    reduceCFL = linRes > 1.2*linTol && (iter >= startingIter);

    canIncrease = (linRes < linTol) && (iter >= startingIter);
```

bigfooted:
> Don't modify the configs to add default options. Just the template is enough

In that case I think we're done?

Common/src/CConfig.cpp
```diff
@@ -1777,11 +1777,12 @@ void CConfig::SetConfig_Options() {
...
   addBoolOption("CFL_ADAPT", CFL_Adapt, false);
   /* !\brief CFL_ADAPT_PARAM
-   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max), acceptable linear residual )
+   * DESCRIPTION: Parameters of the adaptive CFL number (factor down, factor up, CFL limit (min and max)[, acceptable linear residual][, starting iteration]).
+   * Parameters in square brackets are optional, parameter "starting iteration" only valid with parameter "acceptable linear residual".
    * Factor down generally <1.0, factor up generally > 1.0 to cause the CFL to increase when the under-relaxation parameter is 1.0
    * and to decrease when the under-relaxation parameter is less than 0.1. Factor is multiplicative. \ingroup Config*/
-  default_cfl_adapt[0] = 1.0; default_cfl_adapt[1] = 1.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 100.0;
-  default_cfl_adapt[4] = 0.001;
+  default_cfl_adapt[0] = 0.1; default_cfl_adapt[1] = 2.0; default_cfl_adapt[2] = 10.0; default_cfl_adapt[3] = 1e10;
```
pcarruscag, line 1784:
```suggestion
  default_cfl_adapt[0] = 0.1;
  default_cfl_adapt[1] = 1.2;
  default_cfl_adapt[2] = 10.0;
  default_cfl_adapt[3] = 100.0;
```

Common/src/CConfig.cpp
```diff
@@ -1777,11 +1777,12 @@ void CConfig::SetConfig_Options() {
...
+  default_cfl_adapt[4] = 0.001; default_cfl_adapt[5] = 0.0;
```
pcarruscag, line 1785:
```suggestion
  default_cfl_adapt[4] = 0.001;
  default_cfl_adapt[5] = 0.0;
```

config_template.cfg
```diff
@@ -1389,12 +1389,12 @@ CFL_NUMBER= 15.0
 CFL_ADAPT= NO
 %
 % Parameters of the adaptive CFL number (factor-down, factor-up, CFL min value,
-%                                        CFL max value, acceptable linear solver convergence)
-% Local CFL increases by factor-up until max if the solution rate of change is not limited,
+%                                        CFL max value, acceptable linear solver convergence, starting iteration)
+% After the starting iteration has passed, local CFL increases by factor-up until max if the solution rate of change is not limited,
 % and acceptable linear convergence is achieved. It is reduced if rate is limited, or if there
```
pcarruscag, line 1394:
```suggestion
% After the starting iteration has passed, local CFL increases by factor-up until max if the solution rate of
% change is not limited, and acceptable linear convergence is achieved. It is reduced if rate is limited, or if there
```

pcarruscag (approved):

bigfooted:
Thanks, Pedro!

PR #2365, opened 2024-10-12
Fixed issue AttributeError: 'tuple' object has no attribute 'keys'

huynguyen2909:
I received this message when I am trying to run Dynamic FSI tutorial case 
This issue is derived from line 314 of file FSIInterface.py
What I received in terminal was like this:
File "home/huy/SU2/bin/FSI_tools/FSIInterface.py", line 314, in connect
    if allMovingMarkersTags[0] in allMarkersID.keys():
AttributeError: 'tuple' object has no attribute 'keys'

To fix this I add this line above the error line:
allMarkersID = {marker: idx for idx, marker in enumerate(allMarkersID)}


bigfooted:
Why is there an empty file called FSIInterface.py:Zone.Identifier ? Can you remove it?

huynguyen2909:
> Why is there an empty file called FSIInterface.py:Zone.Identifier ? Can you remove it?

Oh it's just one of those zone identifier files in Ubuntu WSL. They are created every time new files are moved into a directory and they are empty. I have removed it.

pcarruscag:
@bigfooted you fixed this another way right?

bigfooted:
> @bigfooted you fixed this another way right?


Yes, this is fixed now and in develop. I will close it. @huynguyen2909 Thanks for your help!

PR #2367, opened 2024-10-17
Fix aachen testcase

alecappiello:
## Proposed Changes
I discovered that the Aachen regression test was actually on its way to divergence. I think this explains why the regression test was behaving inconsistently, although I'm not sure about what caused this problem.

I made a new mesh, introduced wall functions as suggested by @joshkellyjak, and switched to second order directly. 
The convergence history is good. See the attached convergence histories obtained with JST starting from scratch with outlet pressure ramp.
conv_hist_rotor1
conv_hist_stator1
conv_hist_stator2

I have finally updated the restart files accordingly.

## Related Work
I think these small changes could fix the issues encountered in the past with this test case.

## PR Checklist

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -209,7 +209,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c fix_aachen_testcase -s ${{matrix.testscript}}
       - name: Cleanup
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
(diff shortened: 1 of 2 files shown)


pcarruscag (approved):

joshkellyjak:
@pcarruscag I think we may have spoken about this before but my memory does not serve me well, do you have any idea as to why the residual values change when we remove the entropy fix coeff?

pcarruscag:
hmm maybe boundaries, we still use the Roe scheme on boundaries even with JST, I stand corrected.

.github/workflows/regression.yml
```diff
@@ -209,7 +209,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c fix_aachen_test_case -s ${{matrix.testscript}}
```
pcarruscag, line 212: Fix this please @alecappiello

alecappiello: Yes! Though, before I can do that we need to push new mesh and restart files into TestCases/develop branch. Could you please accept that PR (https://github.com/su2code/TestCases/pull/155)?

Issue #2118, opened 2023-08-26
Restart files should only contain necessary fields


bigfooted:
Restart files currently contain fields that are not necessary for restarts, like residual information, pressure coefficient, etc. 
I propose to store only the fields that are necessary for restarts in the restart file. This will keep the size of the restart files to a minimum. 

I think we only need to store the transported variables, or do we need to store more information for perfect restarts, or adjoints?

PR #2372, opened 2024-10-31
Write compact restart files

bigfooted:
## Proposed Changes
We aim to implement 2 things:
1. Restart files contain variables that are not necessary for restarts. with the keyword WRT_RESTART_COMPACT= YES we can write compact restart files containing only necessary fields.
2. Other output contains variables that are not requested (all SOLUTION variables are always written). When using VOLUME_OUTPUT=COMPACT, we do not automatically write the SOLUTION fields to e.g. paraview.
4. additionally, we write compact surface .csv files.

## Related Work
Mentioned in Issue (request) #2118 

## PR Checklist


SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,85 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
...
-                                          requestedVolumeFields.end(), "SOLUTION");
-  if (itSol == requestedVolumeFields.end()){
+
+  /*--- If no volume fields were requested, we add the entire SOLUTION field.
+   *    We also add the solution field if we are not using the compact formulation.
+   *    This is for backwards compatibility. ---*/
+  if ((config->GetWrt_Restart_Compact() == false) || (nRequestedVolumeFields == 1)) {
     requestedVolumeFields.emplace_back("SOLUTION");
     nRequestedVolumeFields++;
   }
```
pcarruscag, line 1543: This logic is not correct, the idea was to add the solution if it was not present in the list of quantities, what you have is just if the list is empty which is generally not the case because derived classes add their own default.

I recommend keeping the old implementation because certain important quantities are part of the "solution", for example density.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -293,6 +294,11 @@ class COutput {
   /*! \brief Number of requested volume field names in the config file. */
   unsigned short nRequestedVolumeFields;
 
+  /*! \brief Minimum required volume fields for restart file. */
+  std::vector<string> requiredVolumeFields = {"COORDINATES", "SOLUTION"};
```
bigfooted, line 298:
```suggestion
  std::vector<string> requiredVolumeFields = {"COORDINATES", "SOLUTION", "GRID_VELOCITY"};
```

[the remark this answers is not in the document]
bigfooted: Yes, I missed that, thanks!

SU2_CFD/src/output/COutput.cpp
bigfooted: My intention was that the user can choose to write only a single field to e.g. a paraview file, previously the SOLUTION was always written to file.
The other derived classes actually already do a check on _(nRequestedVolumeFields == 0)_ , this one used to specifically check for the existence of the SOLUTION field. I think it is now more compatible with the other implementation as well, we can also remove the condition that we only do this if we ask for compact restarts. It is just my attempt to make it a bit more backward compatible.

bigfooted: OK I see your point regarding SOLUTION vs PRIMITIVE, but that is why I had the extra condition that this behavior is only active when a user is using the compact_restart keyword, nothing changes for the user and he can only get this behavior when he actively sets the keyword COMPACT_RESTART.
But maybe we can think of a better way to get this behavior later in another PR? At least I would like to have the ability to save paraview files that only contain a single field.

pcarruscag: Yep but I liked this idea of compact restarts now that we don't rely on SU2_SOL for anything, so making it the default makes sense.
Maybe we can add some special keyword to the output field option, for example NO_DEFAULTS, and then you get exclusively what you ask for?

bigfooted: OK, so remove the `((config->GetWrt_Restart_Compact() == false)` here and replace it with checking for RequestedVolumeField=="NO_DEFAULTS"?

pcarruscag: Yes, if that field is present, do not add the solution by default, otherwise keep what we had

bigfooted: OK, I added the keyword COMPACT.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -293,6 +294,11 @@ class COutput {
   /*! \brief Number of requested volume field names in the config file. */
   unsigned short nRequestedVolumeFields;
 
+  /*! \brief Minimum required volume fields for restart file. */
+  std::vector<string> requiredVolumeFields = {"COORDINATES", "SOLUTION", "GRID_VELOCITY"};
+  /*! \brief Minimum required volume fields for restart file. */
+  unsigned short nRequiredVolumeFields = requiredVolumeFields.size();
```
pcarruscag, line 300:
```suggestion
  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "GRID_VELOCITY"};
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,91 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
...
+  for (size_t iField_Output = 0; iField_Output < volumeOutput_List.size(); iField_Output++) {
+
+    const string &fieldReference = volumeOutput_List[iField_Output];
 
-  for (unsigned short iField_Output = 0; iField_Output < volumeOutput_List.size(); iField_Output++){
+    if (volumeOutput_Map.count(fieldReference) > 0) {
+
+      VolumeOutputField &Field = volumeOutput_Map.at(fieldReference);
+      /*--- Loop through all fields specified in the config ---*/
+      for (size_t iReqField = 0; iReqField < nRequiredVolumeFields; iReqField++) {
```
pcarruscag, line 1568:
```suggestion
      for (size_t iReqField = 0; iReqField < requiredVolumeFields.size(); iReqField++) {
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,91 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
...
-  if (itSol == requestedVolumeFields.end()){
-    requestedVolumeFields.emplace_back("SOLUTION");
-    nRequestedVolumeFields++;
-  }
 
-  nVolumeFields = 0;
+  /*--- If no volume fields were requested, we add the entire SOLUTION field.
+   *    We also add the solution field if we are not using the compact formulation.
+   *    This is for backwards compatibility. ---*/
+  itCoord = std::find(requestedVolumeFields.begin(),
```
pcarruscag, line 1540:
```suggestion
  auto itSol = std::find(requestedVolumeFields.begin(),
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,91 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
...
+                                     requestedVolumeFields.end(), "SOLUTION");
+  if (itCoord == requestedVolumeFields.end()) {
+    auto itCoordCompact = std::find(requestedVolumeFields.begin(),
```
pcarruscag, line 1543:
```suggestion
    auto itCompact = std::find(requestedVolumeFields.begin(),
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,91 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
 
   SetVolumeOutputFields(config);
 
-  /*---Coordinates and solution groups must be always in the output.
-   * If they are not requested, add them here. ---*/
-
+  /*--- Coordinates must be always in the output.
+   * If they are not requested, add them here.
+   * SOLUTION is only required for restart files. ---*/
```
pcarruscag, line 1529:
```suggestion
   * If they are not requested, add them here. ---*/
```

SU2_CFD/src/output/COutput.cpp
```diff
@@ -1504,66 +1524,91 @@ void COutput::PreprocessVolumeOutput(CConfig *config){
...
-                                          requestedVolumeFields.end(), "SOLUTION");
-  if (itSol == requestedVolumeFields.end()){
-    requestedVolumeFields.emplace_back("SOLUTION");
-    nRequestedVolumeFields++;
-  }
 
-  nVolumeFields = 0;
+  /*--- If no volume fields were requested, we add the entire SOLUTION field.
+   *    We also add the solution field if we are not using the compact formulation.
+   *    This is for backwards compatibility. ---*/
```
pcarruscag, line 1539:
```suggestion
  /*--- Add the solution if it was not requested for backwards compatibility, unless the COMPACT keyword was used to request exclusively the specified fields. ---*/
```

config_template.cfg
```diff
@@ -60,6 +60,9 @@ GRAVITY_FORCE= NO
 % Restart solution (NO, YES)
 RESTART_SOL= NO
 %
+% Only save minimum required variables for restart to minimize the restart file size.
+WRT_RESTART_COMPACT= YES
```
pcarruscag, line 64: Add info about the COMPACT option please

pcarruscag (approved):

TestCases/flamelet/02_laminar_premixed_ch4_flame_hx_ad/lam_prem_ch4_hx_ad.cfg
```diff
@@ -128,7 +128,7 @@ MARKER_ANALYZE_AVERAGE= MASSFLUX
 OUTPUT_FILES= RESTART, RESTART_ASCII, PARAVIEW
 VOLUME_OUTPUT= RESIDUAL, PRIMITIVE
 VOLUME_ADJ_FILENAME= ch4_flame_hx_ad
-
+WRT_RESTART_COMPACT = NO
```
pcarruscag, line 131: Why are these needed to pass regressions? Are they using a filediff?

bigfooted: Yes. Should I update the restart files instead?

bigfooted:
Not sure yet why the stl file writing is failing. It is just SU2_SOL writing an stl, no solution is involved.

bigfooted:
The stl file writing (rans/oneram6) is working fine on my machine and produces the same stl as the regression result. I'll see if I can get it to fail inside a docker later.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -293,6 +294,9 @@ class COutput {
   /*! \brief Number of requested volume field names in the config file. */
   unsigned short nRequestedVolumeFields;
 
+  /*! \brief Minimum required volume fields for restart file. */
+  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY"};
```
bigfooted, line 298: Some cases also needed the SENSITIVITY added to the restart file

TestCases/flamelet/02_laminar_premixed_ch4_flame_hx_ad/lam_prem_ch4_hx_ad.cfg
pcarruscag: No this is fine.

TestCases/cont_adj_euler/naca0012/inv_NACA0012.cfg
```diff
@@ -127,6 +127,7 @@ SURFACE_ADJ_FILENAME= surface_adjoint
 OUTPUT_WRT_FREQ= 250
 SCREEN_OUTPUT = (INNER_ITER, RMS_ADJ_DENSITY, RMS_ADJ_ENERGY, SENS_GEO, SENS_AOA, RMS_DENSITY, RMS_ENERGY, LIFT, DRAG)
 HISTORY_OUTPUT= (ITER, RMS_RES, AERO_COEFF, SENSITIVITY)
+WRT_RESTART_COMPACT = NO
```
pcarruscag, line 130: Can you revert some of these now to see if the tests pass again?

bigfooted:
thanks, Pedro!

PR #2375, opened 2024-11-08
[WIP] fix multigrid agglomeration

bigfooted:
## Proposed Changes
Implement multigrid agglomeration rules according to Nishikawa et al paper:

https://www.researchgate.net/publication/267557097_Development_and_Application_of_Parallel_Agglomerated_Multigrid_Method_for_Complex_Geometries

These rules were not consistently implemented. Issues:
1. Ridge -> ridge agglomeration was not possible.
2. Interior points were allowed to be agglomerated with boundary points.
3. The symmetry plane / euler wall agglomeration rules are too restrictive.

TO DO: properly agglomerate nodes that are on mpi interfaces (SEND_RECEIVE markers)

## PR Checklist

### DIFF
--- a/Common/include/geometry/CMultiGridGeometry.hpp
+++ b/Common/include/geometry/CMultiGridGeometry.hpp
@@ -45,7 +45,7 @@ class CMultiGridGeometry final : public CGeometry {
    * \param[in] config - Definition of the particular problem.
    * \return <code>TRUE</code> or <code>FALSE</code> depending if the control volume can be agglomerated.
    */
-  bool SetBoundAgglomeration(unsigned long CVPoint, short marker_seed, const CGeometry* fine_grid,
+  bool SetBoundAgglomeration(unsigned long CVPoint, vector<short> marker_seed, const CGeometry* fine_grid,
                              const CConfig* config) const;
 
   /*!
--- a/Common/src/geometry/CMultiGridGeometry.cpp
+++ b/Common/src/geometry/CMultiGridGeometry.cpp
@@ -31,6 +31,10 @@
 #include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
 
 CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, unsigned short iMesh) : CGeometry() {
+
+  vector<short> marker_seed;
+
+
   nDim = fine_grid->GetnDim();  // Write the number of dimensions of the coarse grid.
 
   /*--- Create a queue system to do the agglomeration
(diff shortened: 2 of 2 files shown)


Common/include/geometry/CMultiGridGeometry.hpp
```diff
@@ -45,7 +45,7 @@ class CMultiGridGeometry final : public CGeometry {
    * \param[in] config - Definition of the particular problem.
    * \return <code>TRUE</code> or <code>FALSE</code> depending if the control volume can be agglomerated.
    */
-  bool SetBoundAgglomeration(unsigned long CVPoint, short marker_seed, const CGeometry* fine_grid,
+  bool SetBoundAgglomeration(unsigned long CVPoint, vector<short> marker_seed, const CGeometry* fine_grid,
```
bigfooted, line 48: We need to pass information about all markers on the seed node (number of markers and index of marker)

YairMO:
Hi,

It would be great to make the MG work better. I would like to take this opportunity to improve my understanding of the MG inputs in the cfg file. Here is an example of MG input:

MGLEVEL=3
MGCYCLE= V_CYCLE
MG_PRE_SMOOTH= ( 5, 5, 5, 10 )
MG_POST_SMOOTH= ( 5, 5, 5, 5 )
MG_CORRECTION_SMOOTH= ( 25, 20, 15, 10 )

I was expecting to obtain the same report in the SU2 output file. However, there seems to be a gap between the input and the output reports. I attached the corresponding report output (of the above input):
MG-Output

Also, could you tell me how a turbulence model is handled within the MG? Is it projected or solved using the MG method?

Thank you

YairMO:
Thank you.
I understand why pre- and post-smoothing of the coarsest and finest levels may be indistinguishable. Therefore, I understand why the post-smoothing of these levels is zero, but I expect instead to control the pre-smoothing relaxation of the finest level. Fixing the pre-smoothing of the finest level to 1 relaxation may be insufficient. 

We wish to contribute our knowledge about MG for scalar transport equations; if an exciting group is on this topic, we will be happy to collaborate.

bigfooted:
Nice to hear there is a larger interest in getting the multigrid method to a higher level. If you would like to enable multigrid for species transport, the first thing to do is change the integration type from SINGLEGRID to MULTIGRID for CreateSpeciesSolver in CSolverFactory.cpp. Then in CMultigridIntegration.cpp, there might be some solver specific things that need to be added. 
Another thing that seems important for viscous applications is to agglomerate along implicit lines normal to viscous walls.

YairMO:
Thank you for pointing this out. Does the current agglomeration (for the mean-flow equations) consider lines normal to the surface? I think that the Hiroaki paper (MG) considred implicit line?

bigfooted:
Hi, No we currently do not do agglomeration along lines. I suspect this is one reason that we do not have good multigrid performance for a number of viscous testcases. @EvertBunschoten was looking into this as well, but he was not yet able to reproduce the good performance reported by the papers of Diskin, Nishikawa and others.

SU2_CFD/src/integration/CFEM_DG_Integration.cpp
```diff
@@ -60,12 +60,8 @@ void CFEM_DG_Integration::SingleGrid_Iteration(CGeometry ****geometry,
...
         local time stepping.  Note that we are currently hard-coding
         the classical RK4 scheme. ---*/
-  bool useADER = false;
-  switch (config[iZone]->GetKind_TimeIntScheme()) {
-    case RUNGE_KUTTA_EXPLICIT: iLimit = config[iZone]->GetnRKStep(); break;
-    case CLASSICAL_RK4_EXPLICIT: iLimit = 4; break;
-    case ADER_DG: iLimit = 1; useADER = true; break;
-    case EULER_EXPLICIT: case EULER_IMPLICIT: iLimit = 1; break; }
+  unsigned short iLimit = config[iZone]->GetnRKStep();
+
```
bigfooted, line 64: all that stuff was not used.

PR #2391, opened 2024-12-06
Improve unit tests

guusbertens:
## Proposed Changes

- Upgrade to Meson 0.64.1. This minor version introduces `fs.copyfile`, which is needed for copying files for unit tests (see next point). The latest patch level within this minor version is 0.64.1.
- Fix the build system so it copies files that are needed for unit tests to the build dir.
- Do not install unit tests. Unit tests depend on files that are not currently installed, so if the tests were installed, they could not run anyway.

## Related Work

- No currently open issues mention "unittests"
- No currently open PRs mention "unittests"

## PR Checklist

Githooks man page says that the pre-commit hook must have run, but I don't know how to verify that it did.


UnitTests/meson.build
```diff
@@ -29,23 +31,31 @@ su2_cfd_tests_dd = files(['Common/simple_directdiff_test.cpp'])
 
 if get_option('enable-tests')
   if get_option('enable-normal')
+    unit_test_deps = [
+        fs.copyfile('SU2_CFD/fluid/multicomponent_cantera.cfg', 'multicomponent_cantera.cfg'),
```
Cristopher-Morales, line 35: Hi Guus, it is Cristopher from TU Eindhoven, Thanks for this pull request. Yesterday, I mentioned that Cantera is still not available in SU2, so I think the multicomponent_cantera.cfg file should be removed for now. When the Cantera implementation is ready, I will add it. Thanks for your help !!!

guusbertens: Good point; that shouldn't be here for this branch. Thanks!

guusbertens:
The develop branch currently downloads Meson 0.61.1, but `fs.copyutils` is introduced in version 0.64.0.

UnitTests/meson.build
```diff
@@ -29,23 +31,30 @@ su2_cfd_tests_dd = files(['Common/simple_directdiff_test.cpp'])
...
+        workdir: meson.current_build_dir(),
+    )
   endif
 
   if get_option('enable-autodiff')
     unit_test_files_ad = su2_cfd_tests_ad + files(['test_driver.cpp'])
     test_driver_AD = executable(
         'test_driver_AD',
         unit_test_files_ad,
-        install : true,
```
pcarruscag, line 48: We would need to change the regression workflow to find the unit test binaries somewhere else. It's simpler if you add a meson option to skip installing them by default, and then we can install them in the regression test script.

guusbertens: Thanks! I'll have a look!

guusbertens: But... The documentation indicates the unit tests should be run from within the build dir (https://su2code.github.io/docs_v7/Running-Unit-Tests/). Shouldn't the regressions test script do the same, then?

(I'm not an SU2 pro... Maybe the regression test script tests for more and/or other things, which would invalidate the above.)

guusbertens: Option 2 would require a change to the `compileSU2.sh` script. I doubt that that will be reflected anytime soon, which results in the tests failing until the container is updated.

@pcarruscag I think options 3 above is what you suggested, but would still either break the documented way to run unit tests (`ninja -C builddir test`), or require one to specify a workdir in `regression.yml`, which seems impossible.

What is the intention of tests? Should they be run before installing, to ensure one doesn't install a broken binary, or should they be run after installing?

bigfooted: Hi, Guus,
All tests that are not unit tests are run after installation/compilation, and are there mainly to make sure that developers do not break existing functionality during development. The tests all represent very small 'real' testcases that are solved, and we compare the output with the known and stored output.
The unit tests are special because they involve testing for correctness of specific c++ functions that we have, so they involve the compilation of a small cpp file that calls the specific function.
So the unit tests have to know where the header files are, the rest of the tests only have to know where the executable is. Hope that helps.

Issue #2383, opened 2024-11-20
SST omega is wrong in sharp corners


bigfooted:
**Describe the bug**

In a 3D channel with square cross section and inflation layers, omega can be wrong in the corner. Below is a cross section of such a channel. Omega is computed using: 
$$\omega_{wall} = 60 \nu / (\beta d^2)$$, 
with d the "distance to the next point away from the wall" [Menter, 1994]. We take the point that is most normal to the wall, computed using FindNormal_Neighbor. However, in the case below, the returned points are the points on the wall. Since they are very far away, the corner omega is very small, order 1e3 instead of 1e6 in the points surrounding the corner point, which causes convergence issues. 

Menter mentions that the interior point should actually have y+ < 3. So we should favor nodes that are closer. Maybe prefer nodes based on a weight factor using the distance and the angle? 

![image](https://github.com/user-attachments/assets/21156570-086b-44ea-bbdd-2381917c8031)

bigfooted:
In any case since the boundary condition is based on an y+ value, I think points on the wall should always be excluded.

pcarruscag:
There are two versions of the w boundary condition, the "60" version at the face/wall and the "6" version on the first wall CV. We can switch to the latter, it's consistent with wall functions.

rois1995:
Is the problem related to the normal of the corner point being computed wrong?

pcarruscag:
Of course, all of this is brought about by the quirks of vertex centered discretizations :)

bigfooted:
@pcarruscag do you mean like the approach of Menter and Esch (What Fluent uses):
https://www.researchgate.net/file.PostFileLoader.html?id=58500561b0366d121d756004&assetKey=AS:438841040478208@1481639265395

This is the paper where they introduce a blending function for the wall bc $\omega(y^+) = (\omega_{visc}(y^+)^2 + \omega_{log}(y^+)^2)^{0.5}$
and they use $\omega_{visc}(y) = \frac{6 \nu}{\beta y^2}$

pcarruscag:
They are both cell centers for us, but one coincides with the wall and the other is the first node of the wall, there are some testing benefits to having a consistent treatment for no-slip and wall-modelled walls. But I don't know if the "60" version at the wall is more robust.

rois1995:
Maybe I am missing something, but in this particular case, if the wall normal on a boundary point is taken as the mean of the boundary edge's normal that are connected to that point, shouldn't the first normal point be the correct one?

[...]

PR #2392, opened 2024-12-06
Fix cornernode normal

bigfooted:
## Proposed Changes
Quick fix to issue #2383 
The normal node is the **interior** node that is most normal to the edge. If the node is on the wall, it is only considered as normal node if no previous normal nodes were detected (just for the case that there are no normal nodes). A normal node that is on the wall can be overwritten by any interior node.

## PR Checklist


pcarruscag (commented):
I don't think this is the full fix.
First issue, at vertices shared by two markers we compute two normals for the same point, potentially resulting in different "normal neighbors". Then the result we get depends on whether markers are merged or not, their order, etc.
Then the distance we should use is the wall distance and not just the straight line between points.

So if you want to do something that doesn't break all cases you have the solution I mentioned in the issue
> smallest non-zero wall distance of a neighbor of the wall point.

bigfooted:
OK, so the 'normal angle' does not matter anymore in that case, we only have edge length (to an interior node) as requirement?

stevosg99:
I am very interested in a solution to this issue, as this particular bug has caused convergence issues for a long time (especially with the SST Models). Here is a comparison of 8.1.0 (with the corner of issues circled in red) and the fix_cornernode_normal version just created. 

image

It seems to be causing all of the wall distances to be miscalculated now.

pcarruscag:
We only have wall distance as a requirement, not distance between nodes.

bigfooted:
OK, I changed it to wall distance. 
For the flow over the backward facing step, with a structured mesh as shown, the corner node iPoint does not have any connecting nodes jPoint that are in the interior. In this case, You can loop over the nodes jPoint and check those points for interior points.

But still, the corner node has zero shear stress, so y+ = 0 in that case.

image
image

pcarruscag:
Good point, in such cases we can probably use cbrt of the volume at that point to avoid a 0.
My suggestion is to change just the SST wall boundary condition and avoid touching the normal neighbor all together because it impacts too many things in the code.

bigfooted:
I can introduce FindClosest_Neighbor, and use that for the walls in TurbSSTSolver. We can then phase out FindNearest_Neighbor over time.

pcarruscag:
Just do what you need to do in the SST wall boundary condition for now

rois1995:
Y plus value of zero just means that the skin friction is zero there, right? which is expected since it is a stagnation point. Thus I don't understand where a division by zero should be occurring.

bigfooted:
@pcarruscag do you think this is the way to go? If so, I can clean it up and fix the regressions, that's still a whole bunch.

pcarruscag:
I don't think so.
- We need the distance projected in the normal direction, not the Euclidean distance.
- The second search you are using for vertices without an interior neighbor is not valid for MPI, if the vertex is close to an MPI boundary, the second-degree neighbors do not exist.
- I would rather not add anything to CVertex for now, I'll think about a better way of determining the normal neighbor.

So, in the SST solver, and nowhere else:
- Loop over the neighbors of the point and find the smallest non-zero wall distance.
- If that is zero, use 2 * Volume / (sum of vertex areas) as an estimate of the normal distance.

bigfooted:
> Loop over the neighbors of the point and find the smallest non-zero wall distance.
But we still take interior neighbors only and discard wall points, we just do not do the second search?
> If that is zero, use 2 * Volume / (sum of vertex areas) as an estimate of the normal distance.

pcarruscag:
Those will have wall distance = 0

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -1024,3 +1032,47 @@ void CTurbSSTSolver::SetUniformInlet(const CConfig* config, unsigned short iMark
...
+          cout << "no point found at i = " << iPoint << endl;
+          su2double Area = 0.0;
+          su2double TwoVol = 2.0* (geometry->nodes->GetVolume(iPoint) + geometry->nodes->GetPeriodicVolume(iPoint));
+          for (size_t iNeigh = 0; iNeigh < geometry->nodes->GetnPoint(iPoint); ++iNeigh) {
+            size_t iEdge = geometry->nodes->GetEdge(iPoint, iNeigh);
+            size_t jPoint = geometry->nodes->GetPoint(iPoint, iNeigh);
+            const su2double* Normal = geometry->edges->GetNormal(iEdge);
+            Area = GeometryToolbox::Norm(nDim, Normal);
+
+            dist_min += TwoVol / Area;
```
bigfooted, line 1071: @pcarruscag I now did this. If I compare this length scale in the corner node to the ones next to it, it is much larger.

pcarruscag: Sure but have you tried what I suggested? Namely, wall distance (not Euclidean), and volume divided by the sum of vertex areas (vertices of viscous markers).
Vertices are on boundaries, edges are internal.

bigfooted: Sorry, I'm getting old and am a bit hard of hearing :-) 
OK, so in the code above we only take the vertex areas, we check if jPoint is actually on the wall. using GetViscousBoundary
For the 'regular' routine we can do x.n/||n||
I have modified it, length scales seem to be fine now.

pcarruscag: Cool thanks for testing. I think it's simpler than that, I'll push something in the weekend, don't start updating tests yet.

bigfooted:
I updated the regression results. I did not update the restart files, that can be done after the next cornernode PR.

pcarruscag:
What is the next corner node PR?

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -2419,6 +2419,14 @@ class CFVMFlowSolverBase : public CSolver {
...
   }
 
+  /*!
+   * \brief Set the y plus.
+   * \param[in] val_yplus - new value of yplus
+   */
+  inline void SetYPlus(su2double val_yplus, unsigned short val_marker, unsigned long val_vertex) final {
+    YPlus[val_marker][val_vertex] = val_yplus;
+  }
+
```
pcarruscag, line 2429: remove?

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2498,10 +2498,15 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
...
-        // for CMultiGridGeometry, the normal neighbor of halo nodes in not set
+        // for CMultiGridGeometry, the normal neighbor of halo nodes is not set
         iPointNormal = geometry->vertex[iMarker][iVertex]->GetNormal_Neighbor();
         Coord_Normal = geometry->nodes->GetCoord(iPointNormal);
         WallDistMod = GeometryToolbox::Distance(nDim, Coord, Coord_Normal);
+        //nijso: we can modifiy y+ as well now?
+
+        
+
+
```
pcarruscag, line 2509: remove?

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -2989,6 +2989,14 @@ class CSolver {
...
   inline virtual su2double GetYPlus(unsigned short val_marker, unsigned long val_vertex) const { return 0; }
 
+  /*!
+   * \brief A virtual member.
+   * \param[in] val_marker - Surface marker where the coefficient is computed.
+   * \param[in] val_vertex - Vertex of the marker <i>val_marker</i> where the coefficient is evaluated.
+   * \return Value of the y plus.
+   */
+  inline virtual void SetYPlus(su2double val_yplus, unsigned short val_marker, unsigned long val_vertex) {};
+
```
pcarruscag, line 2999: remove?

bigfooted:
> What is the next corner node PR?
We can maybe discuss in the developer meeting what the next step is, what's the best approach to generalize?

Common/src/geometry/CGeometry.cpp
```diff
@@ -3946,26 +3976,53 @@ void CGeometry::ComputeWallDistance(const CConfig* const* config_container, CGeo
...
+            const auto iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
+            su2double dist = 0;
+            if (viscous && geometry->nodes->GetDomain(iPoint)) {
+              dist = NearestNeighborDistance(geometry, config, iPoint);
+            } else {
+              dist = geometry->nodes->GetWall_Distance(iPoint);
+            }
+            geometry->vertex[iMarker][iVertex]->SetNearestNeighborDistance(dist);
+          }
+          END_SU2_OMP_FOR
```
pcarruscag, line 4025:
```suggestion
      if (!wallDistanceNeeded[iZone] || kindSolver == MAIN_SOLVER::FEM_LES || kindSolver == MAIN_SOLVER::FEM_RANS) {
        continue;
      }
      const auto* config = config_container[iZone];
      auto* geometry = geometry_container[iZone][iInst][MESH_0];

      for (unsigned short iMarker = 0; iMarker < config->GetnMarker_All(); ++iMarker) {
        const auto viscous = config->GetViscous_Wall(iMarker);

        SU2_OMP_FOR_STAT(OMP_MIN_SIZE)
        for (auto iVertex = 0u; iVertex < geometry->nVertex[iMarker]; iVertex++) {
          const auto iPoint = geometry->vertex[iMarker][iVertex]->GetNode();
          su2double dist = 0;
          if (viscous && geometry->nodes->GetDomain(iPoint)) {
            dist = NearestNeighborDistance(geometry, config, iPoint);
          } else {
            dist = geometry->nodes->GetWall_Distance(iPoint);
          }
          geometry->vertex[iMarker][iVertex]->SetNearestNeighborDistance(dist);
        }
        END_SU2_OMP_FOR
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -3946,26 +3976,53 @@ void CGeometry::ComputeWallDistance(const CConfig* const* config_container, CGeo
...
         }
```
pcarruscag, line 4026:
```suggestion
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -2497,11 +2497,9 @@ void CFVMFlowSolverBase<V, FlowRegime>::Friction_Forces(const CGeometry* geometr
 
       FrictionVel = sqrt(fabs(WallShearStress[iMarker][iVertex]) / Density);
 
-      if (!wallfunctions && (MGLevel == MESH_0 || geometry->nodes->GetDomain(iPoint))) {
-        // for CMultiGridGeometry, the normal neighbor of halo nodes in not set
-        iPointNormal = geometry->vertex[iMarker][iVertex]->GetNormal_Neighbor();
-        Coord_Normal = geometry->nodes->GetCoord(iPointNormal);
-        WallDistMod = GeometryToolbox::Distance(nDim, Coord, Coord_Normal);
+      if (!wallfunctions && MGLevel == MESH_0 && geometry->nodes->GetDomain(iPoint)) {
+        // for CMultiGridGeometry, the normal neighbor of halo nodes is not set
```
pcarruscag, line 2501:
```suggestion
        // for CMultiGridGeometry and halos, the nearest neighbor distance is not set
```

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -684,12 +684,9 @@ void CIncNSSolver::SetTau_Wall_WF(CGeometry *geometry, CSolver **solver_containe
...
       const su2double VelTangMod = GeometryToolbox::Norm(int(MAXNDIM), VelTang);
 
-      /*--- Compute normal distance of the interior point from the wall ---*/
+      /*--- Compute nearest-neighbor distance of the interior point from the wall ---*/
 
-      su2double WallDist[MAXNDIM] = {0.0};
-      GeometryToolbox::Distance(nDim, Coord, Coord_Normal, WallDist);
-
-      su2double WallDistMod = GeometryToolbox::Norm(int(MAXNDIM), WallDist);
+      su2double WallDistMod = geometry->vertex[iMarker][iVertex]->GetNearestNeighborDistance();
```
pcarruscag, line 689: revert

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -863,12 +863,9 @@ void CNSSolver::SetTau_Wall_WF(CGeometry *geometry, CSolver **solver_container,
...
       const su2double VelTangMod = GeometryToolbox::Norm(int(MAXNDIM), VelTang);
 
-      /*--- Compute normal distance of the interior point from the wall ---*/
+      /*--- Compute nearest-neighbor distance of the interior point from the wall ---*/
 
-      su2double WallDist[MAXNDIM] = {0.0};
-      GeometryToolbox::Distance(nDim, Coord, Coord_Normal, WallDist);
-
-      const su2double WallDistMod = GeometryToolbox::Norm(int(MAXNDIM), WallDist);
+      const su2double WallDistMod = geometry->vertex[iMarker][iVertex]->GetNearestNeighborDistance();
```
pcarruscag, line 868: revert

SU2_CFD/src/solvers/CTurbSASolver.cpp
```diff
@@ -275,7 +275,7 @@ void CTurbSASolver::Postprocessing(CGeometry *geometry, CSolver **solver_contain
               FrictionVelocity = sqrt(flowNodes->GetLaminarViscosity(iPoint)*VorticityMag);
             }
 
-            const su2double wall_dist = geometry->nodes->GetWall_Distance(jPoint);
+            const su2double wall_dist = geometry->vertex[iMarker][iVertex]->GetNearestNeighborDistance();
```
pcarruscag, line 278: revert

SU2_CFD/src/solvers/CIncNSSolver.cpp
pcarruscag: Because the normal neighbor approach is being mixed with this and what we talked about was to replace where it was straightforward.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -459,20 +459,17 @@ void CTurbSSTSolver::BC_HeatFlux_Wall(CGeometry *geometry, CSolver **solver_cont
...
+        su2double wall_dist = geometry->vertex[val_marker][iVertex]->GetNearestNeighborDistance();
 
-        su2double distance2 = GeometryToolbox::SquaredDistance(nDim,
-                                                             geometry->nodes->GetCoord(iPoint),
-                                                             geometry->nodes->GetCoord(jPoint));
         /*--- Set wall values ---*/
-
+        const auto jPoint = geometry->vertex[val_marker][iVertex]->GetNormal_Neighbor();
         su2double density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(jPoint);
         su2double laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(jPoint);
```
pcarruscag, line 467:
```suggestion
        su2double density = solver_container[FLOW_SOL]->GetNodes()->GetDensity(iPoint);
        su2double laminar_viscosity = solver_container[FLOW_SOL]->GetNodes()->GetLaminarViscosity(iPoint);
```

pcarruscag:
I'll help update the tests.

bigfooted:
@pcarruscag regression tests are passing.

.github/workflows/regression.yml
```diff
@@ -209,7 +209,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t corner_node -c corner_node -s ${{matrix.testscript}}
```
pcarruscag, line 212:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```

.github/workflows/regression.yml
```diff
@@ -255,7 +255,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2-tsan:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tsan"
+          args: -b ${{github.ref}} -t corner_node -c corner_node -s ${{matrix.testscript}} -a "--tsan"
```
pcarruscag, line 258:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tsan"
```

.github/workflows/regression.yml
```diff
@@ -300,7 +300,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2-asan:240320-1536
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--asan"
+          args: -b ${{github.ref}} -t corner_node -c corner_node -s ${{matrix.testscript}} -a "--asan"
```
pcarruscag, line 303:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--asan"
```

TestCases/parallel_regression.py
```diff
@@ -1233,7 +1233,7 @@ def main():
     # For a thin disk with the inner and outer radius of this geometry, from
     # "Formulas for Stress, Strain, and Structural Matrices", 2nd Edition, figure 19-4,
     # the maximum stress is 165.6MPa, we get a Von Misses stress very close to that.
-    rotating_cylinder_fea.test_vals = [-6.861940, -6.835545, -6.895500, 22, -8.313847, 1.6502e+08]
+    rotating_cylinder_fea.test_vals = [-6.005467, -5.615543, -5.615529, 38.000000, -8.126591, 164570000.000000]
```
pcarruscag, line 1236:
```suggestion
    rotating_cylinder_fea.test_vals = [-6.861940, -6.835550, -6.895497, 22.000000, -8.313847, 1.6502e+08]
```

PR #2418, opened 2025-01-13
Fix output names of unsteady simulation

rois1995:
## Proposed Changes
It solves two main issues in unsteady simulations:
1- if the history file is not updated during the simulation (ex. setting inner_iter = 0) the time iteration is not updated in the file output names.
2- mesh file output does not have the time iteration at the end.

## PR Checklist


rois1995:
I would also try to add the writing of the mesh in the CGNS format, although this can be done in another PR.

bigfooted:
LGTM

SU2_CFD/src/output/COutput.cpp
```diff
@@ -476,7 +479,7 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, OUTPUT_TYPE form
       extension = CSU2MeshFileWriter::fileExt;
 
       if (fileName.empty())
-        fileName = volumeFilename;
+        fileName = config->GetFilename(volumeFilename, "", curTimeIter);
```
bigfooted, line 482: So if the filename is empty, the time iteration is added, but if the filename is not empty,  the iterations were always already added before to `fileName`, is that it?

rois1995: Essentially I have followed all the other cases. Without appending the time iter at the end of the file name it just overwrites the same file at each unsteady iteration.

SU2_CFD/src/output/COutput.cpp
```diff
@@ -389,6 +389,9 @@ void COutput::WriteToFile(CConfig *config, CGeometry *geometry, OUTPUT_TYPE form
   /*--- File writer that will later be used to write the file to disk. Created below in the "switch" ---*/
   CFileWriter *fileWriter = nullptr;
 
+  /*--- Set current time iter even if history file is not written ---*/
+  curTimeIter = config->GetTimeIter();
```
pcarruscag, line 393: If curTimeIter is a member variable, why are you having to set its value here?
It was already used in line 418/421.
So do we have the same problem with innerIter and outerIter? Can you fix this closer to the current place where these iteration counters are set? (history file judging from the comment)

rois1995: The problem only arises when I set the inner iterations to be 0 in an unsteady dual time-stepping simulation. I guess that, since in this case the history file is not updated, the curTimeIter is not updated and remains 0 when writing the output files. In all other cases where the history file is written it is not necessary. I guess I could use an if-statement for that but I don't think that it is necessary. I don't know if the problem persist with the outer iter, I will check (I guess it only matters during a steady simulation when (WRT_*_OVERWRITE is set to YES in the config file).

pcarruscag: Yes but please look for a single place where to put setting these variables, potentially moving where they are set at the moment.

rois1995: Sure, I missed that request before. I have created a setCurTimeIter function in the COutput class and I call it when the Solve function is called. I have checked also for multi-zone problems and it works fine.

SU2_CFD/src/iteration/CFluidIteration.cpp
```diff
@@ -376,6 +376,8 @@ void CFluidIteration::Solve(COutput* output, CIntegration**** integration, CGeom
   /*--- Boolean to determine if we are running a static or dynamic case ---*/
   bool steady = !config[val_iZone]->GetTime_Domain();
 
+  if (!steady) output->SetCurTimeIter(config[val_iZone]->GetTimeIter());
+
```
pcarruscag, line 380: This is worse, go with what you had initially.

rois1995:
Can you try a case with supersonic inlet and check if the mach number is computed correctly at the inlet?

SU2_CFD/src/iteration/CFluidIteration.cpp
pcarruscag: Because you are increasing the preconditions necessary to use the output classes correctly. I.e. an additional function must be called.
Moreover, you added changes to one iteration class, but what about the others?
The principle is really simple, minimize the "surface to volume ratio" of abstract classes. Smallest interface possible, and don't duplicate anything, especially logic.

pcarruscag: Sorry that it's so much work to fix a seemingly simple bug, but most SU2 bugs are because principles like this were  not followed.

rois1995: That's true. Then, I will revert to the previous implementation.

rois1995: > Sorry that it's so much work to fix a seemingly simple bug, but most SU2 bugs are because principles like this were not followed.

No worries, let me know if anything else has to be changed.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -418,6 +418,12 @@ class COutput {
    */
   void SetMultizoneHistoryOutput(COutput** output, CConfig **config, CConfig *driver_config,
                                   unsigned long TimeIter, unsigned long OuterIter);
+  
+  /*!
+   * \brief Sets the current time iteration
+   * \param[in] timeIter - the current time iteration
+   */
+  inline void SetCurTimeIter(unsigned long val_curTimeIter) {curTimeIter = val_curTimeIter;}
```
pcarruscag, line 426:
```suggestion
```

SU2_CFD/src/drivers/CMultizoneDriver.cpp
```diff
@@ -397,6 +397,8 @@ void CMultizoneDriver::Corrector(unsigned short val_iZone) {
 
 bool CMultizoneDriver::OuterConvergence(unsigned long OuterIter) {
 
+  // Set the current time iter for the multizone problem
+
```
pcarruscag, line 401:
```suggestion
```

pcarruscag (approved):

PR #2422, opened 2025-01-15
[WIP] Implementation of Amplification Factor Transport(AFT) 2019b tranistion model

sun5k:
## Proposed Changes

Hi, all.

This PR is for implementing well known subsonic transition model [(SA-AFT2019b)](https://turbmodels.larc.nasa.gov/aft_transition_3eqn.html).

Coder's SA-AFT2019b model is composed of the amplification factor and the logarithmic intermittency transport equation as shown below.
image

## Current State
The primary implementation of the model has been completed and is now in the validation. The validation problems are refer : https://doi.org/10.2514/6.2023-3530 and https://doi.org/10.2514/6.2019-0039.

The transition onset location and the amplification factor field are slightly different. I'm still debugging and checking various things.

## Validation Problem of Current State
Firstly, I'm trying the validation problem for 2-D problems (flat plate and NLF-0416 airfoil). 
The current results are shown below. 

**S&K Flat plate** from https://doi.org/10.2514/6.2023-3530

image

**NLF-0416 airfoil** from https://doi.org/10.2514/6.2023-3530
image

## PR Checklist


SU2_CFD/include/numerics/turbulent/transition/trans_diffusion.hpp
```diff
@@ -103,3 +103,78 @@ class CAvgGrad_TransLM final : public CAvgGrad_Scalar<FlowIndices> {
...
+    if (implicit) {
+      const su2double proj_on_rho_i = proj_vector_ij/Density_i;
+      Jacobian_i[0][0] = -diff_AF*proj_on_rho_i;  Jacobian_i[0][1] = 0.0;
+      Jacobian_i[1][0] = 0.0;                        Jacobian_i[1][1] = -diff_Gamma*proj_on_rho_i;
+
+      const su2double proj_on_rho_j = proj_vector_ij/Density_j;
+      Jacobian_j[0][0] = diff_AF*proj_on_rho_j;   Jacobian_j[0][1] = 0.0;
+      Jacobian_j[1][0] = 0.0;                     Jacobian_j[1][1] = diff_Gamma*proj_on_rho_j;
+    }
+  }
```
pcarruscag, line 166: This only differs from LM by some constant factors, refactor it to use the same class.

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -307,23 +317,28 @@ struct Edw {
...
 };
 
 /*! \brief Non-zero ft2 term according to the literature. */
 struct Nonzero {
-  static void get(CSAVariables& var) {
+  static void get(CSAVariables& var, bool transition_AFT) {
     const su2double xsi2 = pow(var.Ji, 2);
     var.ft2 = var.ct3 * exp(-var.ct4 * xsi2);
     var.d_ft2 = -2.0 * var.ct4 * var.Ji * var.ft2 * var.d_Ji;
+    if(transition_AFT) {
```
pcarruscag, line 338: if else

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
```diff
@@ -192,7 +196,13 @@ class CSourceBase_TurbSA : public CNumerics {
       }
 
       /*--- Compute ft2 term ---*/
-      ft2::get(var);
+      if(transition_AFT) {
+        var.lnintermittecncy = exp(intermittency_i);
```
pcarruscag, line 200: it's not lnintermittecncy because you take exp of it right? Can you use the `intermittency` that is already in the variables?

SU2_CFD/src/solvers/CEulerSolver.cpp
```diff
@@ -1073,6 +1074,19 @@ void CEulerSolver::SetNondimensionalization(CConfig *config, unsigned short iMes
...
+      N_crit = -8.43 - 2.4 * log( (2.5 * tanh( config->GetTurbulenceIntensity_FreeStream() *100 / 2.5)) / 100.0);
+    }
+    else {
+      N_crit = 9.0;
+    }
+  } 
+  else {
+    N_crit = config->GetN_Critical();
+  }
+  config->SetN_Crtical(N_crit);
```
pcarruscag, line 1088: Can this be done somewhere that lets be applicable to all flow solvers? Incompressible, NEMO, etc.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
pcarruscag, line 1: You need to find a way to share the common parts between the LM solver and this one, this way there is too much duplication.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
```diff
@@ -0,0 +1,598 @@
...
+
+  switch (config->GetKind_Gradient_Method()) {
+    case GREEN_GAUSS:
+      SetAuxVar_Gradient_GG(geometry, config);
+      break;
+    case WEIGHTED_LEAST_SQUARES:
+      SetAuxVar_Gradient_LS(geometry, config);
+    default:
+      break;
+    }
```
pcarruscag, line 263: Why are you computing the AuxVarGradient after you use it above instead of before?

SU2_CFD/src/solvers/CTransAFTSolver.cpp
```diff
@@ -0,0 +1,598 @@
...
+      nodes->SetAuxVar(iPoint, 1, Temp3 + Temp4 + flowNodes->GetVelocity(iPoint, 2) * nodes->GetAuxVarGradient(iPoint, 0, 2));
+    }
+
+    su2double VorticityMag = GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint));
+    su2double HLGradTerm = 0.0;
+    HLGradTerm = nodes->GetAuxVarGradient(iPoint, 1, 0) * nodes->GetAuxVarGradient(iPoint, 0, 0) + nodes->GetAuxVarGradient(iPoint, 1, 1) * nodes->GetAuxVarGradient(iPoint, 0, 1);
+
+    if(nDim == 3) {
+      HLGradTerm += nodes->GetAuxVarGradient(iPoint, 1, 2) * nodes->GetAuxVarGradient(iPoint, 0, 2);
+    }
```
pcarruscag, line 219: GeometryToolbox SquaredNorm

SU2_CFD/src/solvers/CTransAFTSolver.cpp
```diff
@@ -0,0 +1,598 @@
...
+    const su2double dist_i = geometry->nodes->GetWall_Distance(iPoint);
+    const su2double StrainMag_i = flowNodes->GetStrainMag(iPoint);
+    const su2double Eddy_Viscosity_i = turbNodes->GetmuT(iPoint);
+    const su2double Laminar_Viscosity_i = flowNodes->GetLaminarViscosity(iPoint);
+    const su2double Density_i = flowNodes->GetDensity(iPoint);
+    const su2double c_1 = 100.0;
+    const su2double c_2 = 0.06;
+    const su2double c_3 = 50.0;
+    su2double Temp3 = flowNodes->GetVelocity(iPoint, 0) * nodes->GetAuxVarGradient(iPoint, 0, 0);
+    su2double Temp4 = flowNodes->GetVelocity(iPoint, 1) * nodes->GetAuxVarGradient(iPoint, 0, 1);
```
pcarruscag, line 204: Add some documentation please, if the initial aux var grad is 0 then you set the aux var to 0, then its gradient is 0 and this remains 0 forever? Looks like a bug.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
```diff
@@ -0,0 +1,598 @@
...
+    const su2double F_onset1 = AF / Critical_N_Factor;
+    const su2double F_onset2 = min(F_onset1, 2.0);
+    const su2double F_onset3 = max(1.0 - pow(RT / 3.5, 3), 0.0);
+    const su2double F_onset = max(F_onset2 - F_onset3, 0.0);
+    const su2double F_turb = exp(-pow(RT / 2.0,4));
+    const su2double Pg = c_1 * Density_i * StrainMag_i * F_onset * (1.0 - exp(lnIntermittency));
+
+    /*-- destruction term of Intermeittency(Gamma) --*/
+    const su2double Dg = c_2 * Density_i * VorticityMag * F_turb * (c_3 * exp(lnIntermittency) - 1.0);
+    nodes -> SetAFT_Wonder_Func(iPoint, HL, H12, dNdRet, Ret0, D_H12, l_H12, m_H12, kv, Rev, Rev0, F_growth, F_crit, PAF, Pg, Dg);
```
pcarruscag, line 249: This is a lot of memory and these variables are only used for post processing (output) right?
If so these can be computed in the output class when setting the output values.
That saves doing this work every iteration and storing all these variables.

pcarruscag (commented):
Thanks :+1: Looks good but we can re-use a lot of code from the LM solver and see the question about the aux var gradient which looks suspicious.

sun5k:
> Thanks 👍 Looks good but we can re-use a lot of code from the LM solver and see the question about the aux var gradient which looks suspicious.

Thank you for comment @pcarruscag. I'll try to address each of your comments one by one.

For now, the SA-AFT2017b model is for testing purposes, so I'll delete the relevant parts.

Common/include/option_structure.hpp
```diff
@@ -1324,6 +1326,73 @@ inline LM_ParsedOptions ParseLMOptions(const LM_OPTIONS *LM_Options, unsigned sh
   return LMParsedOptions;
 }
 
+/*!
+ * \brief AFT Options
+ */
+enum class AFT_OPTIONS {
+  NONE,         /*!< \brief No option / default. */
+  AFT2017b,       /*!< \brief using AFT2017b model. */
+  AFT2019b       /*!< \brief using AFT2019b model. */
```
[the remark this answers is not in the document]
sun5k: Sure! 

The reference paper is "https://doi.org/10.2514/6.2019-0039." 
I'll add the above link.

Common/include/option_structure.hpp
```diff
@@ -1324,6 +1326,73 @@ inline LM_ParsedOptions ParseLMOptions(const LM_OPTIONS *LM_Options, unsigned sh
...
+};
+
+/*!
+ * \brief Function to parse AFT options.
+ * \param[in] AFT_Options - Selected AFT option from config.
+ * \param[in] nAFT_Options - Number of options selected.
+ * \param[in] rank - MPI rank.
+ * \return Struct with AFT options.
+ */
+inline AFT_ParsedOptions ParseAFTOptions(const AFT_OPTIONS *AFT_Options, unsigned short nAFT_Options, int rank) {
```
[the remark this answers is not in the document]
sun5k: Okay. I remove the rank both AFT and LM.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
sun5k: Thank you for your advise.
Current class, this function is for debugging. Later, I'll change to compute in the output class.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
sun5k: I wanted to discuss this with someone. 

On the SA-AFT2019b model, HL is defined as follows : 

image


where, d is the distance from the wall.

The AuxVar[0] is for the d and AuxVarGradient[0] is for the image. So, the d and image are the constant value when the geometry and mesh are not deformed.

And, the Auxvar[1] is for the image and AuxVarGradient[1] is for the image. 

The velocity values change with each iteration. So, the Auxvar[1] and it's gradient must be redefined and calculated each iteration. Therefore, I decided to do the gradient calculation in  the Postprocessing. 

On second thought, the Auxvar assign in "trans_sources.hpp", or would you have any other ideas?

pcarruscag: In the paper it is mentioned that grad d is the wall-normal unit vector, don't we have that stored already? I remember seeing it in another PR, maybe it was not merged yet.

sun5k: > In the paper it is mentioned that grad d is the wall-normal unit vector, don't we have that stored already? I remember seeing it in another PR, maybe it was not merged yet.

OK. I'll try to find it. :D

pcarruscag: I think it is in the simplified 2015 PR

rois1995: If I understood correctly then yes, it is in the simplified transition model PR. The branch is still under development (currently working on the validation on T3 series flat plates and Eppler airfoil) but that part has already been implemented and verified.

sun5k:
The original plan was to finish it before the new version (v8.2.0) was released, but I was busy and couldn't manage it. Now that I have some spare time, I'll start working on it again next week!

SU2_CFD/include/numerics/turbulent/turb_sources.hpp
sun5k: You are right. In the AFT model, the transported variable is γ̃ = ln(γ), so intermittency_i corresponds to γ̃ and exp(intermittency_i) is needed to recover γ. I will rename the variable to avoid confusion.
In addition, using var.intermittency directly is not appropriate because intermittency in the LM model and exp(intermittency_i) in the AFT model are applied at different locations. Therefore, instead of introducing an if statement inside ComputeProduction(), I believe it is better to declare a variable for the AFT intermittency.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
sun5k: @rois1995

Hi, if I understood correctly, the modifications in #1901 were introduced to obtain the unit normal vector at nodes. Is that right?

Commit:
https://github.com/su2code/SU2/commit/b1edf61fca342c17a93f4daca8701474f15de5e5

rois1995: Yes, it sets at each point in the mesh the wall normal of the associated point on a viscous surface.

SU2_CFD/src/solvers/CEulerSolver.cpp
sun5k: > Can this be done somewhere that lets be applicable to all flow solvers? Incompressible, NEMO, etc.

The AFT model was developed to predict subsonic boundary layer transitions (mainly caused by T-S wave instability). While it can be applied to all solvers, it appears to be effective only with incompressible solvers.

SU2_CFD/src/solvers/CTransAFTSolver.cpp
sun5k: Hi, @rois1995. 
I implemented your code, but I found some issues.

As mentioned above, the ∇(d) of the AFT model corresponds to wall-normal unit vector of Menter's transition model.  

I moved the part of wall-normal unit vector calculation into the my repository (feature_AFT_2019b_Model). However, when using wall-normal unit vector, I observed slight differences compared to the previous version  (using ∇(d)) for both the flat plate and the airfoil. So I performed several additional checks. 

It turns out that there is a minor bug in the wall-normal unit vector calculation. I initially thought it might be due to an incorrect implementation or mistake, so I installed and tested your code (feature_Trans_SLM) directly.   For both the flat plate (zero pressure gradient from TMR) and the airfoil (NACA-0012), the wall-normal unit vectors obtained from your code are shown below. For visualization purposes, I adjusted the spacing of the vectors.

Flat plate : 
image

For the flat plate, the wall-normal unit vector on the symmetry boundary has a value of (0, 1).



Airfoil : 

image

image

For the airfoil, reasonable results are obtained near the leading edge. In the wake region downstream of the trailing edge, the wall-normal unit vector is expected to be (1, 0). Instead, it appears to retain the trailing-edge wall-normal direction. In regions farther away, the y-component of the wall-normal vector is reversed.

PR #2420, opened 2025-01-14
Enable the use of conductivity models different from CONSTANT_PRANDTL for compressible flows.

Cristopher-Morales:
## Proposed Changes

Enable the use other conductivity models for compressible flow. Updating thermal conductivity which has the turbulence contribution from MakeThermalConductivityModel function in CFluidModel.cpp.

## Related Work

This is the first of small pull requests that are needed for enable FLUID_MIXTURE model for compressible flows.

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [X] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [ ] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

PR #2426, opened 2025-01-18
[WIP] Preconditioning fro multicomponent flows

Cristopher-Morales:
## Proposed Changes

This pull request aims to extend the preconditioning for incompressible multicomponent flows, specifically when FLUID_MIXTURE model is used and energy equation is solved. Changes introduced in this pull request aim to solve directly for the sensible enthalpy instead of temperature. Specifically, the energy equation that is going to be solved for multicomponent flows is given as follows:

image

and the sensible enthalpy is: 
image

Currently $C_{p}$ does not depend on temperature for FLUID_MIXTURE, but it depends on the mixture composition. So the sensible enthalpy simplifies to $C_{p}*(T-T_{0})$. For problems where $C_{p}$ strongly depends on the temperature, Cantera library will be coupled to SU2. After this pull request is completed, a pull request with a fluid model using Cantera will be done where an energy equation for the total enthalpy will be solved.

## Related Work

Currently fluid mixture does not work correctly when the energy equation is on, this is mainly due to the fact that the energy equation that must be solved for multicomponent flows is the one mentioned above for the sensible enthalpy and not the energy equation for a perfect gas that is currently solved in SU2.

## PR Checklist


SU2_CFD/include/fluid/CFluidModel.hpp
```diff
@@ -339,6 +355,13 @@ class CFluidModel {
    */
   virtual void SetTDState_T(su2double val_Temperature, const su2double* val_scalars = nullptr) {}
 
+  /*!
+   * \brief Virtual member.
+   * \param[in] val_enthalpy - Enthalpy value at the point.
+   */
+  virtual void ComputeTempFromEnthalpy(su2double val_enthalpy, su2double* val_temperature,
+                                       const su2double* val_scalars = nullptr) {}
```
pcarruscag, line 363: This seems inconsistent with the SetTDState_* functions.
Should this be SetTDState_h and then you use GetTemperature to get the temperature?

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -91,6 +91,11 @@ class CFluidScalar final : public CFluidModel {
    */
   su2double ComputeMeanSpecificHeatCp(const su2double* val_scalars);
 
+  /*!
+   * \brief Compute Enthalpy given the temperature and scalars.
+   */
+  su2double ComputeEnthalpyFromT(const su2double val_temperature, const su2double* val_scalars);
```
pcarruscag, line 97: Why doesn't this use TDState_T

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -434,7 +434,7 @@ void CFVMFlowSolverBase<V, R>::Viscous_Residual_impl(unsigned long iEdge, CGeome
   const bool implicit  = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
   const bool tkeNeeded = (config->GetKind_Turb_Model() == TURB_MODEL::SST);
 
-  CVariable* turbNodes = nullptr;
+  CVariable* turbNodes = nullptr;
```
pcarruscag, line 437: Please setup your IDE to remove trailing spaces.

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -212,13 +212,43 @@ su2double CFluidScalar::ComputeMeanSpecificHeatCp(const su2double* val_scalars)
   return mean_cp;
 }
 
+su2double CFluidScalar::ComputeEnthalpyFromT(const su2double val_temperature, const su2double* val_scalars){
+  su2double val_Enthalpy = Cp * (val_temperature - 298.15);
```
pcarruscag, line 216: Isn't T0 a config option? It is bad practice to hard code constants like this.

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -278,10 +279,20 @@ void CNumerics::GetInviscidIncProjJac(const su2double *val_density, const su2dou
...
-    val_Proj_Jac_Tensor[3][0] = val_scale*((*val_cp)*(*val_temperature)*proj_vel/(*val_betainc2));
-    val_Proj_Jac_Tensor[3][1] = val_scale*((*val_cp)*(*val_temperature)*val_normal[0]*(*val_density));
-    val_Proj_Jac_Tensor[3][2] = val_scale*((*val_cp)*(*val_temperature)*val_normal[1]*(*val_density));
-    val_Proj_Jac_Tensor[3][3] = val_scale*((*val_cp)*((*val_temperature)*(*val_dRhodT) + (*val_density))*proj_vel);
+    if (energy_multicomponent) {
+      val_Proj_Jac_Tensor[3][0] = val_scale * ((*val_temperature) * proj_vel / (*val_betainc2));
+      val_Proj_Jac_Tensor[3][1] = val_scale * ((*val_temperature) * val_normal[0] * (*val_density));
+      val_Proj_Jac_Tensor[3][2] = val_scale * ((*val_temperature) * val_normal[1] * (*val_density));
+      val_Proj_Jac_Tensor[3][3] =
+          val_scale * ((*val_temperature) * (*val_dRhodT) + (*val_density)) * proj_vel;
```
pcarruscag, line 287: What is happening here? Is temperature being used for enthalpy?

SU2_CFD/src/numerics/flow/convection/fds.cpp
```diff
@@ -157,9 +163,15 @@ CNumerics::ResidualType<> CUpwFDSInc_Flow::ComputeResidual(const CConfig *config
...
-    dRhodT_i   = -DensityInc_i/Temperature_i;
-    dRhodT_j   = -DensityInc_j/Temperature_j;
+    if (energy_multicomponent) {
+      MeandRhodT = -MeanDensity / (MeanCp * MeanTemperature);
+      dRhodT_i = -DensityInc_i / (Cp_i * Temperature_i);
+      dRhodT_j = -DensityInc_j / (Cp_j * Temperature_j);
+    } else {
+      MeandRhodT = -MeanDensity / MeanTemperature;
+      dRhodT_i = -DensityInc_i / Temperature_i;
+      dRhodT_j = -DensityInc_j / Temperature_j;
```
pcarruscag, line 173: Right, I'm going to stop reviewing now until you document your code properly

pcarruscag (changes requested):

SU2_CFD/src/numerics/flow/convection/fds.cpp
Cristopher-Morales: that is why I put [WIP] in the pull request.

SU2_CFD/src/fluid/CFluidScalar.cpp
Cristopher-Morales: https://ansyshelp.ansys.com/public/account/secured?returnurl=////Views/Secured/corp/v242/en/flu_th/flu_th_sec_hxfer_theory.html this is the theory related to the enthalpy equation for multicomponent gases. T_ref is always 298.15 K for this equation

SU2_CFD/src/numerics/CNumerics.cpp
Cristopher-Morales: the energy can be $C_{p}T$ or enthalpy. The enthalpy is used not the temperature.

SU2_CFD/include/fluid/CFluidScalar.hpp
Cristopher-Morales: We compute the enthalpy from the temperature and the species mass fractions.

SU2_CFD/include/fluid/CFluidModel.hpp
Cristopher-Morales: In the pull request made by evert regarding preferential diffusion, he also used the same function for the flamelet solver. The only difference is that  that function is in the CSpeciesFlameletSolver.cpp and I cannot access to that function where I set the primitive in SetPrimVar in CIncNSVariable.cpp

SU2_CFD/include/fluid/CFluidScalar.hpp
pcarruscag: And why don't you compute the enthalpy in that function too instead of adding another function?
The idea with generic classes like CFluidModel is to have a concise abstract interface. Having specific functions that only a few subclasses implement defeats the purpose of a generic interface.

SU2_CFD/src/fluid/CFluidScalar.cpp
pcarruscag: Then make it a constant, not a hard-coded value everywhere.

SU2_CFD/src/numerics/flow/convection/fds.cpp
pcarruscag: The purpose of WIP is to get early feedback to avoid re-doing work later on. For people to give feedback some amount of documentation is necessary.

Cristopher-Morales: OK, I am going to close the pull request

SU2_CFD/include/fluid/CFluidModel.hpp
pcarruscag: My point is again to keep an abstract interface.
Evert's function is a private function of a specific solver which is perfectly fine, and it uses the fluid model... If you need a similar function, move it somewhere higher up in the hierarchy to make it accessible. For example the scalar solver. Or even the fluid model but implemented in a way that keeps it useful for all fluids.

SU2_CFD/src/numerics/CNumerics.cpp
pcarruscag: then some renaming is in order and you need to come up with a better solution to switch between formulations that to have an if statement everytime Cp is used (for example making Cp=1, or encapsulating the computation of enthalpy in a function).

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -212,13 +213,40 @@ su2double CFluidScalar::ComputeMeanSpecificHeatCp(const su2double* val_scalars)
...
+}
+
+void CFluidScalar::GetGradEnthalpyDiffusivity(su2double* grad_enthalpy_diffusions){
+  for (int iVar = 0; iVar < n_species_mixture - 1; iVar++) {
+    grad_enthalpy_diffusions[iVar] = Density *
+                               (specificHeat[iVar] * massDiffusivity[iVar] -
+                                specificHeat[n_species_mixture - 1] * massDiffusivity[n_species_mixture - 1]);
+  }
+}
+
```
Cristopher-Morales: @bigfooted Hi Nijso, do you think these functions related to the diffusion of species due to different enthalpies  should be put together in a single one called like ComputeEnthDiffTerms?

bigfooted:
Hi Cristopher, when you are finished, just remove the WIP, you still need to update the restart file for one of the species cases that is now failing.

SU2_CFD/include/fluid/CFluidModel.hpp
```diff
@@ -186,6 +192,16 @@ class CFluidModel {
     return mass_diffusivity;
   }
 
+  /*!
+   * \brief Get heat diffusivity terms.
+   */
+  virtual void GetEnthalpyDiffusivity(su2double* enthalpy_diffusions = nullptr) {}
+
+  /*!
+   * \brief Get gradient heat diffusivity terms.
```
pcarruscag: These are gradients with respect to what variable?

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -130,6 +133,10 @@ class CNumerics {
   const su2double
   *ScalarVar_i,   /*!< \brief Vector of scalar variables at point i. */
   *ScalarVar_j;   /*!< \brief Vector of scalar variables at point j. */
+  su2double
+  HeatFluxDiffusion;   /*!< \brief Heat flux due to enthalpy diffusion for multicomponent. */
+  su2double
+  Jac_HeatFluxDiffusion;   /*!< \brief Heat flux jacobian due to enthalpy diffusion for multicomponent. */
```
pcarruscag:
```suggestion
  JacHeatFluxDiffusion;   /*!< \brief Heat flux jacobian due to enthalpy diffusion for multicomponent. */
```

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -751,6 +760,20 @@ class CNumerics {
...
+   * \brief Set the heat flux due to enthalpy diffusion
+   * \param[in] val_heatfluxdiffusion - Value of the heat flux due to enthalpy diffusion.
+   */
+  inline void SetHeatFluxDiffusion(su2double val_heatfluxdiffusion) { HeatFluxDiffusion = val_heatfluxdiffusion; }
+
+  /*!
+   * \brief Set Jacobian of the heat flux due to enthalpy diffusion
+   * \param[in] val_jacheatfluxdiffusion - Value of the heat flux jacobian due to enthalpy diffusion.
+   */
+  inline void SetJacHeatFluxDiffusion(su2double val_jac_heatfluxdiffusion) {
```
pcarruscag:
```suggestion
  inline void SetJacHeatFluxDiffusion(su2double val_jacheatfluxdiffusion) {
```

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -203,6 +203,19 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
+   *        in a more thermodynamically consistent way.
+   * \note This method is static to improve the chances of it being used in a
+   *       thread-safe manner.
+   * \param[in,out] fluidModel - The fluid model.
+   * \param[in] nDim - Number of physical dimensions.
+   * \param[in,out] primitive - Primitive variables.
+   * \param[in] scalar - scalar variable.
+   */
+  static void ComputeConsistentExtrapolation(CFluidModel* fluidModel, unsigned short nDim, su2double* primitive,
+                                             const su2double* scalar);
```
pcarruscag: put the primitives last since they are the main result of the function

SU2_CFD/include/solvers/CIncNSSolver.hpp
```diff
@@ -94,8 +94,19 @@ class CIncNSSolver final : public CIncEulerSolver {
...
+   * \brief Compute the enthalpy due to species diffusion contribution for multicomponent and reacting flows.
+   * \param[in] iEdge - Edge for which the flux and Jacobians are to be computed.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] numerics - Description of the numerical method.
+   * \param[in] n_species - Number of species in the mixture that a transport equation is solved.
+   * \param[in] implicit - Boolean for implicit iterations.
+   */
+  void ComputeEnthalpyDiffusion(unsigned long iEdge, CGeometry* geometry, CSolver** solver_container, CNumerics* numerics,
+                        const int n_species, const bool implicit) ;
```
pcarruscag: format

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -212,13 +213,40 @@ su2double CFluidScalar::ComputeMeanSpecificHeatCp(const su2double* val_scalars)
   return mean_cp;
 }
 
+su2double CFluidScalar::ComputeEnthalpyFromT(const su2double val_temperature, const su2double* val_scalars){
+  /* In future implementations, enthalpy should be computed using numerical integration. For now, as Cp does not
+   * depend on temperature, but it does depend on mixture composition, ehtalpy is directly computed from
```
pcarruscag:
```suggestion
   * depend on temperature, but it does depend on mixture composition, enthalpy is directly computed from
```

SU2_CFD/src/numerics/CNumerics.cpp
```diff
@@ -54,6 +54,7 @@ CNumerics::CNumerics(unsigned short val_nDim, unsigned short val_nVar,
   Prandtl_Lam = config->GetPrandtl_Lam();
   Prandtl_Turb = config->GetPrandtl_Turb();
   Gas_Constant = config->GetGas_ConstantND();
+  energy_multicomponent = ((config->GetKind_FluidModel() == FLUID_MIXTURE) && (config->GetEnergy_Equation()));
```
pcarruscag:
```suggestion
  energy_multicomponent = config->GetKind_FluidModel() == FLUID_MIXTURE && config->GetEnergy_Equation();
```

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -97,7 +97,17 @@ CNumerics::ResidualType<> CCentLaxInc_Flow::ComputeResidual(const CConfig* confi
   DensityInc_i  = V_i[nDim+2];        DensityInc_j  = V_j[nDim+2];
   BetaInc2_i    = V_i[nDim+3];        BetaInc2_j    = V_j[nDim+3];
   Cp_i          = V_i[nDim+7];        Cp_j          = V_j[nDim+7];
-  Enthalpy_i    = Cp_i*Temperature_i; Enthalpy_j    = Cp_j*Temperature_j;
+  if (energy_multicomponent) {
+    Enthalpy_i = V_i[nDim + 9];
+    Enthalpy_j = V_j[nDim + 9];
+    WorkingVariable_i = Enthalpy_i;
+    WorkingVariable_j = Enthalpy_j;
```
pcarruscag: Don't make these intermediate variables members of the numerics class if you can just use local variables in ComputeResidual

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -130,6 +141,12 @@ CNumerics::ResidualType<> CCentLaxInc_Flow::ComputeResidual(const CConfig* confi
   if (variable_density) {
     MeandRhodT = -MeanDensity/MeanTemperature;
   }
+  if (energy_multicomponent) {
+    MeandRhodT /= MeanCp ;
+    MeanCp = 1.0;
+    Cp_i = 1.0;
+    Cp_j = 1.0;
+  }
```
pcarruscag: Documentation

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1308,6 +1321,33 @@ void CIncEulerSolver::Upwind_Residual(CGeometry *geometry, CSolver **solver_cont
...
         Primitive_j[iVar] = V_j[iVar];
       }
+      if (energy_multicomponent) {
+        su2double Project_Grad_Enthalpy_i = 0.0;
+        su2double Project_Grad_Enthalpy_j = 0.0;
+
+        for (iDim = 0; iDim < nDim; iDim++) {
+          Project_Grad_Enthalpy_i += Vector_ij[iDim] * nodes->GetAuxVarGradient(iPoint, 1, iDim);
+          Project_Grad_Enthalpy_j -= Vector_ij[iDim] * nodes->GetAuxVarGradient(jPoint, 1, iDim);
+        }
```
pcarruscag: Why is it important to reconstruct enthalpy instead of computing it from the reconstructed temperature?

SU2_CFD/src/numerics/flow/convection/centered.cpp
Cristopher-Morales: Hi! 
Thank you so much for the review,
Here you can find the documentation [link](https://github.com/Cristopher-Morales/Computational_Fluid_Dynamics_notes/blob/main/preconditioning_general_approach.pdf)
I added it the pull request description.
Thank in advance.

SU2_CFD/include/solvers/CIncNSSolver.hpp
Cristopher-Morales: Hi, 
Thanks for the comment.
I have a doubt, this format that you mention is about the way of this function is writing or is about the specific format used in SU2? 
I had a problem with my linux system months ago, so I had to create a new one. I remember that time ago I added a package or similar to VSCode for formatting according to SU2. So now I am ussing the format option that comes in VSCode. 
Could you help me about this comment? 
Thank you so much in advance.

SU2_CFD/include/solvers/CIncEulerSolver.hpp
Cristopher-Morales: Thanks!! I will do it.

SU2_CFD/src/numerics/flow/convection/centered.cpp
Cristopher-Morales: Ok, thanks!!!

SU2_CFD/src/solvers/CIncEulerSolver.cpp
Cristopher-Morales: Thanks for the comment!!
With this extension of the preconditioning for multicomponent and reacting flows (next pull request is about that), we solve for the enthalpy (sensible and chemical enthalpy), so for these cases the temperature is a primitive, no longer the variable that we solve. 
image
As in the MUSCL reconstruction, the working variables are reconstructed, then enthalpy-temperature relation above and  the ideal gas law could be eventually broken. 
So if we reconstruct the enthalpy, then temperature and density must be also updated, otherwise the equation above and density (the ideal gas law) are not longer satisfied, and we see unbounded temperature and density. These unbounded issues are stronger when we simulate reacting flow using Cantera.
If we do not use MUSCL, then we do not see these issues.
That is the main reason why we reconstruct the variables. 
Please let me know if you have other recommendations/suggestions how this could be done in a better way.
Thanks!!!

SU2_CFD/include/fluid/CFluidModel.hpp
Cristopher-Morales: image
Thanks for the comment!
This gradient (I think I should call this in a different way) is for computing the Jacobian of the heat flux term due to species diffusion in the energy equation. In the picture attached, it is explained.
Thanks in advance!!!

SU2_CFD/include/fluid/CFluidScalar.hpp
bigfooted: @Cristopher-Morales can you address this comment please? I think you could generalize the existing setTDState_T.

SU2_CFD/include/fluid/CFluidModel.hpp
bigfooted: @Cristopher-Morales please just improve the \brief comment here.

SU2_CFD/include/fluid/CFluidScalar.hpp
Cristopher-Morales: @bigfooted  Hi Nijso, Thanks for your comment, that function is only declared the CFluidScalar.hpp, it is not used for other fluid models. In Cantera, the enthalpy is obtained using this NASA polynomials, as it is explained here in this link [cantera](https://cantera.org/dev/reference/thermo/species-thermo.html). Then in Cantera, when the SetTDState_T is used, then the enthalpy can be retrieved directly from Cantera because the enthalpy is computed internally evaluating these polynomials, this is an example how it is done in cantera:
`// Create a new Solution object
    auto sol = newSolution(file, phase);
    auto gas = sol->thermo();

    gas->setState_TPX(1500.0, 2.0*OneAtm, "O2:1.0, H2:3.0, AR:1.0");

    // temperature, pressure, and density
    std::cout << gas->temperature() << std::endl;
    std::cout << gas->pressure() << std::endl;
    std::cout << gas->density() << std::endl;

    // molar thermodynamic properties
    std::cout << gas->enthalpy_mole() << std::endl;
    std::cout << gas->entropy_mole() << std::endl;

    // specific (per unit mass) thermodynamic properties
    std::cout << gas->enthalpy_mass() << std::endl;
    std::cout << gas->entropy_mass() << std::endl;`

What I thought when I added this function was that Cp depending on temperature is not currently supported for FLUID_MIXTURE, but later when someone will have to add them (I hope I have time in the near future for addressing this) a function like this should be added for that purpose. Please let me know what you think about this. I can also delete this function for this pull request and compute directly the enthalpy inside the SetTDState_T using Cp * (T - T_ref).

```
void CFluidScalar::SetTDState_T(const su2double val_temperature, const su2double* val_scalars) {
  MassToMoleFractions(val_scalars);
  ComputeGasConstant();
  Temperature = val_temperature;
  Density = Pressure_Thermodynamic / (Temperature * Gas_Constant);
  Cp = ComputeMeanSpecificHeatCp(val_scalars);
  Cv = Cp - Gas_Constant;
  Enthalpy = ComputeEnthalpyFromT(Temperature, val_scalars);
}
```

SU2_CFD/src/numerics/flow/convection/centered.cpp
bigfooted: can you also add some comment above this block of code, explaining why cp=1 here?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
pcarruscag: Sorry if it was not clear, but my question is why you need to solve for enthalpy instead of temperature.
What is the advantage of reconstructing enthalpy and then using the fluid model to recover temperature, versus reconstructing temperature and then using the fluid model to evaluate enthalpy?
Who else does this in literature? Do you have any comparisons between the two approaches?

This change is very intrusive and makes the code harder for everyone, which is why I want to know if it's worth it or if you just hypothesized it would be better to do it this way.

bigfooted: Using total enthalpy is the common approach in combustion, even for incompressible flows. When solving total enthalpy, you do not have the very large reaction source term present.

pcarruscag: What is the reaction source term?

Cristopher-Morales: Hi Pedro,

Thank you so much for your comment, 
we need to solve for the enthalpy for multicomponent and when chemistry is involved because solving for temperature is not suitable for our model. 
In the past year we added a verification and validation test case with a bachelor student [VandV](https://su2code.github.io/vandv/SANDIA_jet/) and a [tutorial](https://su2code.github.io/tutorials/Inc_Species_Transport_Composition_Dependent_Model/) for multicomponent.
If you observe both cases, you can notice that the energy equation is not solved, the reason is that if we solve the energy equation for temperature as it is in the current status of develop, it does not converge even if we reduce the CFL number, use the CFL_ADAPT, reduce the mesh or many other good techniques already available in SU2. If we use this preconditioning for enthalpy for the [tutorial](https://su2code.github.io/tutorials/Inc_Species_Transport_Composition_Dependent_Model/) case, we can make a big improvement, the case converges quite fast and using the CFL_ADAPT we can even accelerate the convergence as you can see in the picture.
image

I totally agree with you that this change is very intrusive and I acknowledge that this implementation can be improved a lot, any guideline/suggestion that could make this implementation clearer and easier to everyone to follow would be very welcome. 

This is the energy equation that we should solve for the temperature for multicomponent and combustion:

image

where the source term mentioned by nijso is this one:

image

Thank you so much in advance for your help!!!

pcarruscag: I think the notation is different from the document you attached to the description. 
But the difference between that w' and the one in the PR description is the summation involving hsk? delta H^0 is the enthalpy of formation?
What is hsk?

pcarruscag: Can you also explain how this term is treated in the solution:
image

Is is treated explicitly or implicitly? Because the hsi are function of temperature but not functions of hs right?
image

Cristopher-Morales: > I think the notation is different from the document you attached to the description. But the difference between that w' and the one in the PR description is the summation involving hsk? delta H^0 is the enthalpy of formation? What is hsk?

Hi! 

Thanks for your comment,

Yes, the notation is slighly different. $h_{sk} = h_{s,k}$ to be consistent with $C_{p,k}$. it is similar to the notation used in [ansys fluent](https://www.afs.enea.it/project/neptunius/docs/fluent/html/th/node107.htm), the s is to highlight that is sensible enthalpy.
 In summary: 
$h_{s,k}$ : sensible enthalpy of species k
$h_{k}$: total enthalpy (sensible + enthalpy of formation),
$\Delta h_{f,k}^{0}$: enthalpy of formation.

Cristopher-Morales: > Can you also explain how this term is treated in the solution: image
> 
> Is is treated explicitly or implicitly? Because the hsi are function of temperature but not functions of hs right? image

Hi!!!,

Thanks for your comment,

Yes, the sensible enthalpy and total enthalpy can be computed using temperature and mass fraction. 
In the solver, we solve for $h_{s}$ or $h$ depending if we use FLUID_MIXTURE model or FLUID_CANTERA(next pull request), then we use the `SetTDState_h ` where we compute the temperature and other thermochemical properties that depends on enthalpy and species mass fractions. This is only done in SetPrimVar in the CIncNSVariable.cpp
```
if(Energy_Multicomponent){
    su2double Enthalpy = Solution(iPoint, nDim +1);
    FluidModel->SetTDState_h(Enthalpy, scalar);
    Temperature = FluidModel->GetTemperature();
  } else {
    /*--- Set the value of the temperature directly ---*/
    Temperature = Solution(iPoint, indices.Temperature());
  }
```
, once we compute the temperature, we can use SetTDState_T in other parts of the solver where we compute these sensible enthalpies of species k (diffusivities for the species solver as well) and we add them as viscous residuals, similar to the heat flux due to temperature gradients in the flow_diffusion.cpp.

Cristopher-Morales: > I think the notation is different from the document you attached to the description. But the difference between that w' and the one in the PR description is the summation involving hsk? delta H^0 is the enthalpy of formation? What is hsk?

As an additional clarification, the temperature equation attached before is taken from the well-known book [Theoretical and numerical combustion](https://elearning.cerfacs.fr/combustion/onlinePoinsotBook/buythirdedition/index.php), Poinsot and Veynante. In this [link](https://cefrc.princeton.edu/sites/g/files/toruqf1071/files/Files/2015%20Lecture%20Notes/Poinsot/2-Basics-elements.pdf), from pages 31 to 40, you can see a better explanation given by the author of that book about the different existing energy equations for combustion.

pcarruscag: Come to the next developer meeting and we'll talk in person there.
I think our ideas of "solving for temperature" are different. The transport quantity and solution quantity need not be the same, you can transport enthalpy and linearize with respect to temperature, as Tom mentioned when we talked about this in an SU2 conference a couple of years ago.

Cristopher-Morales: Hi!!

Thanks for your comment,

Yes, I will attend the next developer meeting. 
See you next week on Tuesday at 19:00!!
Thank you so much in advance!!

SU2_CFD/src/numerics/flow/convection/fds.cpp
```diff
@@ -118,7 +119,17 @@ CNumerics::ResidualType<> CUpwFDSInc_Flow::ComputeResidual(const CConfig *config
...
+    Enthalpy_i = V_i[nDim + 9];
+    Enthalpy_j = V_j[nDim + 9];
+    WorkingVariable_i = Enthalpy_i;
+    WorkingVariable_j = Enthalpy_j;
+  } else {
+    Enthalpy_i = Cp_i * Temperature_i;
+    Enthalpy_j = Cp_j * Temperature_j;
+    WorkingVariable_i = Temperature_i;
+    WorkingVariable_j = Temperature_j;
+  }
```
bigfooted: if we do not have energy_multicomponent, what is then stored at V_i[nDim+9]? Just always use V_i[nDim+9].

Cristopher-Morales: Hi @bigfooted , 
Thank you so much for this comment.
I think I have just addressed it. I moved it outside the if statement.
Please let me know what you think!!

SU2_CFD/include/numerics/flow/convection/centered.hpp
```diff
@@ -49,7 +49,7 @@ class CCentLaxInc_Flow final : public CNumerics {
   MeanDensity, MeanPressure,
   MeanBetaInc2, MeanEnthalpy,
   MeanCp, MeanTemperature,         /*!< \brief Mean values of primitive variables. */
-  MeandRhodT,                      /*!< \brief Derivative of density w.r.t. temperature (variable density flows). */
+  MeandRhodh,                      /*!< \brief Derivative of density w.r.t. temperature (variable density flows). */
```
bigfooted:
```suggestion
  MeandRhodh,                      /*!< \brief Derivative of density w.r.t. enthalpy (variable density flows). */
```

SU2_CFD/include/numerics/flow/convection/centered.hpp
```diff
@@ -109,7 +109,7 @@ class CCentJSTInc_Flow final : public CNumerics {
   MeanDensity, MeanPressure,
   MeanBetaInc2, MeanEnthalpy,
   MeanCp, MeanTemperature,        /*!< \brief Mean values of primitive variables. */
-  MeandRhodT,                     /*!< \brief Derivative of density w.r.t. temperature (variable density flows). */
+  MeandRhodh,                     /*!< \brief Derivative of density w.r.t. temperature (variable density flows). */
```
bigfooted:
```suggestion
  MeandRhodh,                     /*!< \brief Derivative of density w.r.t. enthalpy (variable density flows). */
```

SU2_CFD/include/fluid/CFluidModel.hpp
pcarruscag: unresolving because the comment was not improved

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -42,8 +42,10 @@ class CFluidScalar final : public CFluidModel {
   const int n_species_mixture;            /*!< \brief Number of species in mixture. */
   su2double Gas_Constant;           /*!< \brief Specific gas constant. */
   const su2double Pressure_Thermodynamic; /*!< \brief Constant pressure thermodynamic. */
+  const su2double Ref_Temperature;        /*!< \brief Reference temperature. */
   const su2double GasConstant_Ref;        /*!< \brief Gas constant reference needed for Nondimensional problems. */
   const su2double Prandtl_Number;         /*!< \brief Prandlt number.*/
+  const su2double Schmidt_Turb_Number;         /*!< \brief Prandlt number.*/
```
pcarruscag: Wrong doc

SU2_CFD/include/numerics/flow/flow_sources.hpp
```diff
@@ -126,7 +126,8 @@ class CSourceGeneralAxisymmetric_Flow final : public CSourceAxisymmetric_Flow {
 class CSourceIncAxisymmetric_Flow final : public CSourceBase_Flow {
   bool implicit, /*!< \brief Implicit calculation. */
   viscous,       /*!< \brief Viscous incompressible flows. */
-  energy;        /*!< \brief computation with the energy equation. */
+  energy,        /*!< \brief computation with the energy equation. */
+  multicomponent; /*!< \brief multicomponent incompressible flows. */
```
pcarruscag, line 130: Not needed anymore?

SU2_CFD/include/numerics/flow/flow_diffusion.hpp
```diff
@@ -289,6 +289,7 @@ class CAvgGrad_Flow final : public CAvgGrad_Base {
 class CAvgGradInc_Flow final : public CAvgGrad_Base {
 private:
   su2double Mean_Thermal_Conductivity; /*!< \brief Mean value of the effective thermal conductivity. */
+  su2double Mean_Heat_Capacity;        /*!< \brief Mean value of the heat capacity. */
```
pcarruscag: Can we avoid intermediate su2doubles as class members?

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -203,6 +203,19 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
 
+  /*!
+   * \brief Recompute the extrapolated quantities, after MUSCL reconstruction,
+   *        in a more thermodynamically consistent way.
+   * \note This method is static to improve the chances of it being used in a
+   *       thread-safe manner.
+   * \param[in,out] fluidModel - The fluid model.
+   * \param[in] nDim - Number of physical dimensions.
+   * \param[in,out] primitive - Primitive variables.
+   * \param[in] scalar - scalar variable.
```
pcarruscag: order doesn't match function signature

SU2_CFD/src/fluid/CFluidFlamelet.cpp
```diff
@@ -117,12 +117,17 @@ CFluidFlamelet::~CFluidFlamelet() {
 #endif
 }
 
+void CFluidFlamelet::SetTDState_h(su2double val_enthalpy, const su2double* val_scalars) {
+  SetTDState_T(val_enthalpy, val_scalars);
+}
+
```
pcarruscag, line 130: This looks comment-worthy, why can you pass enthalpy as temperature?

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -212,13 +214,43 @@ su2double CFluidScalar::ComputeMeanSpecificHeatCp(const su2double* val_scalars)
   return mean_cp;
 }
 
+su2double CFluidScalar::ComputeEnthalpyFromT(const su2double val_temperature, const su2double* val_scalars){
+  /* In future implementations, enthalpy should be computed using numerical integration. For now, as Cp does not
```
pcarruscag, line 218: You should try to come up with a way of directly computing enthalpy instead of computing an integral everytime this is called.

SU2_CFD/src/fluid/CFluidScalar.cpp
```diff
@@ -212,13 +214,43 @@ su2double CFluidScalar::ComputeMeanSpecificHeatCp(const su2double* val_scalars)
...
+   * depend on temperature, but it does depend on mixture composition, enthalpy is directly computed from
+   * the expression h_s = Cp(T - T_ref).
+   */
+  su2double val_Enthalpy = Cp * (val_temperature - Ref_Temperature);
+  return val_Enthalpy;
+}
+
+void CFluidScalar::GetEnthalpyDiffusivity(su2double* enthalpy_diffusions) {
```
pcarruscag: Can these be const?
```suggestion
void CFluidScalar::GetEnthalpyDiffusivity(su2double* enthalpy_diffusions) const {
```

SU2_CFD/include/fluid/CFluidScalar.hpp
```diff
@@ -137,9 +144,26 @@ class CFluidScalar final : public CFluidModel {
    */
   inline su2double GetMassDiffusivity(int ivar) override { return massDiffusivity[ivar]; }
 
+  /*!
+   * \brief Get enthalpy diffusivity terms.
+   */
```
pcarruscag, line 156: document the arguments, what size is the array expected to have, etc.

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -33,14 +33,14 @@
 /*!
  * \class CIncEulerVariable
  * \brief Class for defining the variables of the incompressible Euler solver.
- * \note Primitive variables (P, vx, vy, vz, T, rho, beta, lamMu, EddyMu, Kt_eff, Cp, Cv)
- * \note Gradients of primitives (P, vx, vy, vz, T, rho, beta)
+ * \note Primitive variables (P, vx, vy, vz, T, rho,h, beta, lamMu, EddyMu, Kt_eff, Cp, Cv)
```
```suggestion
 * \note Primitive variables (P, vx, vy, vz, T, rho, h, beta, lamMu, EddyMu, Kt_eff, Cp, Cv)
```

SU2_CFD/src/numerics/flow/convection/centered.cpp
```diff
@@ -126,9 +126,12 @@ CNumerics::ResidualType<> CCentLaxInc_Flow::ComputeResidual(const CConfig* confi
...
    law, but in the future, dRhodT should be in the fluid model. ---*/
 
-  MeandRhodT = 0.0;
+  MeandRhodh = 0.0;
   if (variable_density) {
-    MeandRhodT = -MeanDensity/MeanTemperature;
+    MeandRhodh = -MeanDensity/(MeanTemperature * MeanCp);
+    MeanCp = 1.0;
+    Cp_i = 1.0;
+    Cp_j = 1.0;
```
pcarruscag: why?

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -275,10 +275,90 @@ void CIncNSSolver::Compute_Streamwise_Periodic_Recovered_Values(CConfig *config,
...
+  if (implicit) FluidModel->GetGradEnthalpyDiffusivity(GradEnthalpyDiffusion_i);
+
+  /*--- Repeat the above computations for jPoint. ---*/
+
+  su2double EnthalpyDiffusion_j[MAXNVAR_SPECIES]{0.0};
+  su2double GradEnthalpyDiffusion_j[MAXNVAR_SPECIES]{0.0};
+  FluidModel->SetTDState_T(nodes->GetPrimitive(jPoint)[prim_idx.Temperature()], Species_j);
+  FluidModel->SetEddyViscosity(nodes->GetPrimitive(jPoint)[prim_idx.EddyViscosity()]);
+  FluidModel->GetEnthalpyDiffusivity(EnthalpyDiffusion_j);
+  if (implicit) FluidModel->GetGradEnthalpyDiffusivity(GradEnthalpyDiffusion_j);
```
pcarruscag: use a lambda to avoid the copy paste

SU2_CFD/src/solvers/CIncNSSolver.cpp
```diff
@@ -275,10 +275,90 @@ void CIncNSSolver::Compute_Streamwise_Periodic_Recovered_Values(CConfig *config,
...
+  CMatrixView<const su2double> Species_Grad_i = speciesNodes->GetGradient(iPoint);
+  CMatrixView<const su2double> Species_Grad_j = speciesNodes->GetGradient(jPoint);
+
+  /*--- Compute Projected gradient for species variables ---*/
+  su2double ProjGradScalarVarNoCorr[MAXNVAR_SPECIES]{0.0};
+  su2double Proj_Mean_GradScalarVar[MAXNVAR_SPECIES]{0.0};
+  su2double proj_vector_ij =
+      numerics->ComputeProjectedGradient(nDim, n_species, Normal, Coord_i, Coord_j, Species_Grad_i, Species_Grad_j,
+                                         true, Species_i, Species_j, ProjGradScalarVarNoCorr, Proj_Mean_GradScalarVar);
+  (void)proj_vector_ij;
```
pcarruscag:
```suggestion
      numerics->ComputeProjectedGradient(nDim, n_species, Normal, Coord_i, Coord_j, Species_Grad_i, Species_Grad_j,
                                         true, Species_i, Species_j, ProjGradScalarVarNoCorr, Proj_Mean_GradScalarVar);
 
```

SU2_CFD/src/numerics/flow/convection/centered.cpp
Cristopher-Morales: Hi!!

Thank you so much for your comment,

I forgot to remove those ones 😅

SU2_CFD/include/solvers/CIncEulerSolver.hpp
Cristopher-Morales: HI! 

Thanks for the comment!

I switched the order of them according to the function.

Please let me know if that was needed to address this comment.

Thanks in advance!!!

SU2_CFD/src/solvers/CIncNSSolver.cpp
Cristopher-Morales: Hi!

Thank you so much for your comment,

I have added the lambda function for this part.

Please let me know what you think about it.

Thanks!

SU2_CFD/include/numerics/flow/flow_diffusion.hpp
Cristopher-Morales: Hi!

Thanks for the comment! 

I have removed this line because with the previous pull request about conductivity for compressible flows #2420, the mean heat capacity is now a member of the CAvgGrad_Base in the flow_diffussion.hpp, so there was no need to re-declare it in the CAvgGradInc_Flow. 
Please let me know any additional comment about it.

Thanks in advance!!

SU2_CFD/include/fluid/CFluidScalar.hpp
Cristopher-Morales: Hi!

Thanks for your comment,

I have added the right documentation,

please let me know any additional comment/suggestion about it.

Thanks in advance!!

SU2_CFD/include/numerics/flow/flow_sources.hpp
Cristopher-Morales: Hi!

thanks for you comment,

I added that boolean multicomponent because for axisymmetric flows, there is an additional term that comes from the enthalpy diffusion which is only needed for multicomponent flows (**term in blue**)

image

I thought that It should only be computed and added when it is needed because when the FLUID_MODEL is different than FLUID_MIXTURE, the species mass fractions become passive scalars, so the term in blue becomes zero.

Please let me know what you think about it.

Thank you so much in advance!!!

SU2_CFD/src/fluid/CFluidFlamelet.cpp
Cristopher-Morales: Hi!

thanks for the comment!!,

when the SPECIES_FLAMELET solver is used, one of the scalar equation is for total enthalpy, specifically the one with index 1.
As you can see from the picture in SetTDState_T added by Nijso and Evert, if I am not mistaken, the temperature (which is the current variable that is being solved in SU2) is a passive value, it is not used in that function because with the val_scalars (which are the variables used in the manifold for the Flamelet generated Manifold (FGM) approach) the temperature and any other thermochemical property are retrieved from the lookup table as you can see from lines 131.

image

So, as you might remember, in a previous developer meeting I mentioned that if we solve for the enthalpy in the preconditioning, we can eventually move the enthalpy equation solved in the SPECIES FLAMELET solver to the INC_EULER_SOLVER. This is not done in the current pull request because then the pull request will become even much larger as it is.
Please let me know if I should add a comment above these two functions explaining why enthalpy and temperature are passive values when the SPECIES_FLAMELET solver is used. 

In a future pull request, we plan to move the enthalpy equation to the INC_EULER_SOLVER, so the SPECIES_FLAMELET solver can be reduced and used only for solving the additional transport equations needed for the FGM approach.

Thank you so much in advance!!!

SU2_CFD/include/fluid/CFluidScalar.hpp
Cristopher-Morales: Hi!

Thank you so much for your comment,

I have added a documentation of arguments and the size of the array expected by this function.

Please let me know what you think about it.

Thanks!

SU2_CFD/include/fluid/CFluidModel.hpp
Cristopher-Morales: Hi!

Thank you so much for your comment,

I have improved the documentation of this function.

Please let me know what you think about it.

Thanks in advance!

SU2_CFD/src/fluid/CFluidScalar.cpp
Cristopher-Morales: Hi!

thanks for your comment!

I discussed this with @bigfooted and we agreed that, at some point, the option of computing Cp based on the NASA polynomials should be implemented in SU2, in that way the enthalpy could be computed directly from them.
Please let me know what you think about this.

Thanks!

Cristopher-Morales:
> Hi @Cristopher-Morales , do the pywrapper testcases work correctly or is there still some issue?

Hi @bigfooted, 

Thanks for your comment!!

the test case pywrapper_zimont does not work correctly, the other one regarding custom source buoyancy works fine. 
Could you share with me information about this test case? 
so I can see the energy equation that is used for this test case. 
It might need some additional adaption this test case.

Thank you so much in advance!!!

SU2_CFD/include/fluid/CConstantDensity.hpp
```diff
@@ -56,5 +56,15 @@ class CConstantDensity final : public CFluidModel {
        decoupled equation. Hence, we update the value.
        Note Cp = Cv, (gamma = 1).*/
     Temperature = t;
+    Enthalpy = Cp * Temperature;
+  }
+
+  /*!
+   * \brief Set the Dimensionless State using Enthalpy.
+   * \param[in] val_enthalpy - Enthalpy value at the point.
+   */
```
bigfooted, line 66:
```suggestion
   * \param[in] val_enthalpy - Enthalpy value at the point.
   * \param[in] val_scalars - not used here. 
   
   */
```

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
```diff
@@ -72,18 +77,68 @@ class CIncIdealGasPolynomial final : public CFluidModel {
...
+        Enthalpy_iter += coeffs_[i] * t_i * temp_iter / (i + 1);
+      }
+
+      delta_enthalpy_iter = Enthalpy - Enthalpy_iter;
+
+      delta_temp_iter = delta_enthalpy_iter / Cp_iter;
+
+      temp_iter += delta_temp_iter;
+      if (temp_iter < 0.0) {
+        cout << "Warning: Negative temperature has been found during Newton-Raphson" << endl;
```
bigfooted:
```suggestion
      if (temp_iter < 0.0) {
        cout << "Warning: Negative temperature has been found during Newton-Raphson" << endl;
        temp_iter = 0.0;
```

Cristopher-Morales: Hi!!

thank you so much for the comment, 
with the user option of using temperature limits for the incompressible solver, 
Would it be more consistent with that option to return the minimum temperature limit? 
the if statement would be: 
if (temp_iter < temp_limit_minimum){
     temp_iter = temp_limit_minimum
}
the default minimum temperature limit is 0.0

please let me know what you think about this.

Thank you so much in advance!!!

Cristopher-Morales: In the CIncNSVariable.cpp, SetPrimVar function, the temperature and the density are checked, and if their values are outside of the allowed values, the solution is not updated, as it can be seen here:

```
if (check_dens || check_temp) {

    /*--- Copy the old solution ---*/

    for (auto iVar = 0ul; iVar < nVar; iVar++)
      Solution(iPoint,iVar) = Solution_Old(iPoint,iVar);

    /*--- Recompute the primitive variables ---*/

    Temperature = Solution(iPoint, indices.Temperature());
    SetTemperature(iPoint, Temperature, TemperatureLimits);
    FluidModel->SetTDState_T(Temperature, scalar);
    Enthalpy = Solution(iPoint, nDim + 1);
    FluidModel->SetTDState_h(Enthalpy, scalar);
    SetTemperature(iPoint, FluidModel->GetTemperature(), TemperatureLimits);

    SetDensity(iPoint, FluidModel->GetDensity());

    /*--- Flag this point as non-physical. ---*/
```

bigfooted: > Hi!!
> 
> thank you so much for the comment, with the user option of using temperature limits for the incompressible solver, 
> Would it be more consistent with that option to return the minimum temperature limit? 

Yes, if there is a minimum temperature limit that you can use it would be better, by the time you reach T=0, you are probably already too far gone.

SU2_CFD/include/fluid/CConstantDensity.hpp
```diff
@@ -56,5 +56,17 @@ class CConstantDensity final : public CFluidModel {
...
        Note Cp = Cv, (gamma = 1).*/
     Temperature = t;
+    Enthalpy = Cp * Temperature;
+  }
+
+  /*!
+   * \brief Set the Dimensionless State using Enthalpy.
+   * \param[in] val_enthalpy - Enthalpy value at the point.
+   * \param[in] val_scalars - not used here. 
+
```
pcarruscag:
```suggestion
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -72,6 +72,8 @@ class CIncEulerVariable : public CFlowVariable {
   VectorType Streamwise_Periodic_RecoveredPressure,    /*!< \brief Recovered/Physical pressure [Pa] for streamwise periodic flow. */
              Streamwise_Periodic_RecoveredTemperature; /*!< \brief Recovered/Physical temperature [K] for streamwise periodic flow. */
   su2double TemperatureLimits[2];                      /*!< \brief Temperature limits [K]. */
+  su2double TemperatureInc = 0.0;                      /*!< \brief Temperature [K] imposed when energy equation is switch off. */
+  bool Energy;                                         /*!< \brief Flag for Energy equation in incompressible flows. */
```
pcarruscag: Probably doesn't need to be a member
```suggestion
```

SU2_CFD/src/output/CAdjFlowIncOutput.cpp
```diff
@@ -213,8 +225,10 @@ void CAdjFlowIncOutput::LoadHistoryData(CConfig *config, CGeometry *geometry, CS
     SetHistoryOutputValue("RMS_ADJ_TEMPERATURE",         log10(adjheat_solver->GetRes_RMS(0)));
   }
   if (heat) {
-    if (nDim == 3) SetHistoryOutputValue("RMS_ADJ_TEMPERATURE",         log10(adjflow_solver->GetRes_RMS(4)));
-    else           SetHistoryOutputValue("RMS_ADJ_TEMPERATURE",         log10(adjflow_solver->GetRes_RMS(3)));
+    if (nDim == 3)
+      SetHistoryOutputValue("RMS_ADJ_ENTHALPY", log10(adjflow_solver->GetRes_RMS(4)));
+    else
+      SetHistoryOutputValue("RMS_ADJ_ENTHALPY", log10(adjflow_solver->GetRes_RMS(3)));
```
pcarruscag: It was there already, but easy opportunity to simplify.
```suggestion
    SetHistoryOutputValue("RMS_ADJ_ENTHALPY", log10(adjflow_solver->GetRes_RMS(nDim + 1)));
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -241,7 +241,7 @@ void CFlowOutput::SetAnalyzeSurface(const CSolver* const*solver, const CGeometry
               sqrt(config->GetBulk_Modulus()/(flow_nodes->GetDensity(iPoint)));
             }
             Temperature       = flow_nodes->GetTemperature(iPoint);
-            Enthalpy          = flow_nodes->GetSpecificHeatCp(iPoint)*Temperature;
+            Enthalpy = flow_nodes->GetEnthalpy(iPoint);
```
pcarruscag:
```suggestion
            Enthalpy          = flow_nodes->GetEnthalpy(iPoint);
```

TestCases/radiation/p1adjoint/of_grad_cd.csv.ref
```diff
@@ -1,51 +1,51 @@
...
-47              , 0.00752232      , 0.001
-48              , 0.0039072       , 0.001
-49              , 0.00131195      , 0.001
+0               , -0.000805612    , 0.001
+1               , -0.00295734     , 0.001
+2               , -0.00745938     , 0.001
+3               , -0.0149199      , 0.001
+4               , -0.0249982      , 0.001
+5               , -0.0369863      , 0.001
+6               , -0.050081       , 0.001
```
pcarruscag: These changes look significant, did you compare the results for converged adjoint and primal solver?

SU2_CFD/include/variables/CIncEulerVariable.hpp
Cristopher-Morales: Hi!!

Thank you so much for your comment,

I added this flag because for multicomponent and reacting flows, the temperature can be constant but the enthalpy can change due to changes in the species mass fraction composition. So, as in the incompressible solver, the energy equation can be switched-off,  we decide to recompute the enthalpy based on the INC_TEMPERATURE_INIT and the current species mass fraction in the CIncNSVariable.cpp, in the SetPrimVar implementation, as it can be seen here:
```
if (Energy) {
    Enthalpy = Solution(iPoint, nDim + 1);
    FluidModel->SetTDState_h(Enthalpy, scalar);
    Temperature = FluidModel->GetTemperature();
  } else {
    /*--- When energy equation is switch off, a constant temperature is imposed, and enthalpy is recomputed based on
     * this temperature. As in the fluid flamelet model, the temperature is retrieved from a look-up table, then the
     * temperature is obtained directly from the fluidmodel. For the other fluid models, GetTemperature provides the
     * same value as TemperatureInc ---*/
    FluidModel->SetTDState_T(TemperatureInc, scalar);
    Enthalpy = Solution(iPoint, nDim + 1) = FluidModel->GetEnthalpy();
    Temperature = FluidModel->GetTemperature();
  }
```
This could be further improved,
I would really appreciate any advice/suggestion about this.

Thank you so much in advance!!

TestCases/radiation/p1adjoint/of_grad_cd.csv.ref
Cristopher-Morales: Hi! 

thanks for your comment,

Yes, I did some comparison, 
during the day I will post some plots about this specific case.

Thanks!

SU2_CFD/include/variables/CIncEulerVariable.hpp
pcarruscag: But it's used in NS variable not Euler variable, right? Then move to NS variable and initialize in the initializer list so it can be const
```suggestion
  const bool Energy = false;                                         /*!< \brief Flag for Energy equation in incompressible flows. */
```

Cristopher-Morales: Hi!!

thank you so much for your suggestion, I will do what you suggest. Thanks again!

Cristopher-Morales: Hi!

I moved Energy flag to CIncNSVariable and I made it const:

In CIncNSVariable.hpp:

`const bool Energy;          /*!< \brief Flag for Energy equation in incompressible flows. */`

in CIncNSVariable.cpp:


```
CIncNSVariable::CIncNSVariable(su2double pressure, const su2double *velocity, su2double enthalpy,
                               unsigned long npoint, unsigned long ndim, unsigned long nvar, const CConfig *config) :
                               CIncEulerVariable(pressure, velocity, enthalpy, npoint, ndim, nvar, config) {
                               CIncEulerVariable(pressure, velocity, enthalpy, npoint, ndim, nvar, config),
                               Energy(config->GetEnergy_Equation()) {
```
Please let me know if this is a right implementation.

Thank you so much in advance!!!!

Cristopher-Morales: HI!!

I have replied to this in the other comment related to it.

Thanks!!!!

TestCases/radiation/p1adjoint/of_grad_cd.csv.ref
Cristopher-Morales: Hi!!

Thank you so much for pointing out this difference, I checked this case and I realized that I did not update correctly the jacobian contribution in the energy equation due to radiation. My apologies for this mistake.

Here some comparison:

**Develop:**
comparison_temperature_develop

**Current branch:**
comparison_temperature

**Convergence adjoint solver:**
convergence plot

The adjoint residuals of pressure for develop and the current branch overlap, that is the reason why there is only one line. This can also be confirmed observing the residuals of the test case radiative heat transfer p1model, where after the update of the jacobian, the residuals do not need to be modified, also the file of_grad_cd.csv.ref related to this test case.

Please let me know your comments about this.

Thank you so much in advance!!

SU2_CFD/include/variables/CIncEulerVariable.hpp
pcarruscag: Yep, that's it.

TestCases/radiation/p1adjoint/of_grad_cd.csv.ref
pcarruscag: Great, so no changes to this case :+1:

PR #2432, opened 2025-01-25
Add FADO to installation

pcarruscag:
## Proposed Changes
I don't think `import FADO` will work out of the box, let's see...


TestCases/serial_regression.py
```diff
@@ -30,6 +30,10 @@
 from TestCase import TestCase
 from TestCase import parse_args
 
+import FADO
+FADO.printDocumentation()
+
+
```
pcarruscag, line 36:
```suggestion
```

PR #2442, opened 2025-02-11
Tape/Recording debug mode

oleburghardt:
## Proposed Changes
Add feature to record a tape for debug purposes (as a tool for AD / discrete adjoint / .. developers), using CoDiPack's `codi::RealReverseTag`.

This is a (cleaned-up) continuation of #2343.

## Related Work
Supposed to help finding remaining errors in #2317.

## PR Checklist


Common/include/CConfig.hpp
```diff
@@ -9209,6 +9230,16 @@ class CConfig {
...
 
+  /*!
+   * \brief Get the type of tape that will be checked in a tape debug run.
+   */
+  bool GetAD_CheckTapeType(void) const { return AD_CheckTapeType; }
+
+  /*!
+   * \brief Get the type of variables that will be checked for in a tape debug run.
+   */
+  bool GetAD_CheckTapeVariables(void) const { return AD_CheckTapeVariables; }
```
pcarruscag: These are unsigned short but returned as bool?

Common/include/basic_types/ad_structure.hpp
```diff
@@ -547,17 +594,15 @@ FORCEINLINE void SetPreaccOut() {}
...
 }
 
 template <class T>
 FORCEINLINE void SetPreaccOut(T&& data, const int size) {
   if (PreaccActive) {
     for (int i = 0; i < size; i++) {
-      if (IsIdentifierActive(data[i])) {
-        PreaccHelper.addOutput(data[i]);
-      }
+      PreaccHelper.addOutput(data[i]);
```
pcarruscag, line 620: Any efficiency impact from this change? Or codi is already doing something equivalent internally?

Common/include/basic_types/ad_structure.hpp
```diff
@@ -676,6 +719,42 @@ FORCEINLINE void ResumePreaccumulation(bool wasActive) {
   SU2_OMP_SAFE_GLOBAL_ACCESS(PreaccEnabled = true;)
 }
 
+struct ErrorReport {
+  unsigned long ErrorCounter;
+  std::ofstream* out;
+};
+
+FORCEINLINE void ResetErrorCounter(struct ErrorReport* report) { report->ErrorCounter = 0; }
```
pcarruscag: Doesn't seem like there is a reason to pass these by pointer since the functions don't do any checking on whether the argument is valid.
```suggestion
FORCEINLINE void ResetErrorCounter(ErrorReport& report) { report.ErrorCounter = 0; }
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -676,6 +719,42 @@ FORCEINLINE void ResumePreaccumulation(bool wasActive) {
...
+
+FORCEINLINE void SetDebugReportFile(struct ErrorReport* report, std::ofstream* output_file) {
+  report->out = output_file;
+}
+
+FORCEINLINE unsigned long* GetErrorCount(struct ErrorReport* report) { return &(report->ErrorCounter); }
```
pcarruscag:
```suggestion
FORCEINLINE unsigned long& GetErrorCount(ErrorReport& report) { return report->ErrorCounter; }
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -676,6 +719,42 @@ FORCEINLINE void ResumePreaccumulation(bool wasActive) {
   SU2_OMP_SAFE_GLOBAL_ACCESS(PreaccEnabled = true;)
 }
 
+struct ErrorReport {
+  unsigned long ErrorCounter;
+  std::ofstream* out;
```
pcarruscag: Worth having default initialization?
```suggestion
  unsigned long ErrorCounter = 0;
  std::ofstream* out = &std::cout;
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -676,6 +719,42 @@ FORCEINLINE void ResumePreaccumulation(bool wasActive) {
...
+
+FORCEINLINE unsigned long* GetErrorCount(struct ErrorReport* report) { return &(report->ErrorCounter); }
+
+#ifdef CODI_TAG_TAPE
+
+FORCEINLINE void SetTag(int tag) { AD::getTape().setCurTag(tag); }
+FORCEINLINE void ClearTagOnVariable(su2double& v) { AD::getTape().clearTagOnVariable(v); }
+
+static void tagErrorCallback(int const& correctTag, int const& wrongTag, void* userData) {
+  ErrorReport* report = (ErrorReport*)userData;
```
pcarruscag: or reinterpret _cast if really necessary
```suggestion
  auto* report = static_cast<ErrorReport*>(userData);
```

Common/include/option_structure.hpp
```diff
@@ -2543,13 +2543,40 @@ static const MapType<std::string, ENUM_DIRECTDIFF_VAR> DirectDiff_Var_Map = {
   MakePair("ELECTRIC_FIELD", D_EFIELD)
 };
 
+/*!
+ * \brief Types of tapes that can be checked in a tape debug run.
+ */
+enum ENUM_CHECK_TAPE_TYPE {
```
pcarruscag:
```suggestion
enum class CHECK_TAPE_TYPE {
```

Common/include/option_structure.hpp
```diff
@@ -2543,13 +2543,40 @@ static const MapType<std::string, ENUM_DIRECTDIFF_VAR> DirectDiff_Var_Map = {
...
+  OBJECTIVE_FUNCTION_TAPE = 0,    /*!< \brief Tape that only includes dependencies and objective function calculation. */
+  FULL_SOLVER_TAPE = 1            /*!< \brief Tape that includes dependencies and all solvers. */
```
pcarruscag:
```suggestion
  OBJECTIVE_FUNCTION = 0,    /*!< \brief Tape that only includes dependencies and objective function calculation. */
  FULL_SOLVER = 1            /*!< \brief Tape that includes dependencies and all solvers. */
```

Common/include/option_structure.hpp
```diff
@@ -2543,13 +2543,40 @@ static const MapType<std::string, ENUM_DIRECTDIFF_VAR> DirectDiff_Var_Map = {
...
+};
+static const MapType<std::string, ENUM_CHECK_TAPE_TYPE> CheckTapeType_Map = {
+    MakePair("OBJECTIVE_FUNCTION_TAPE", OBJECTIVE_FUNCTION_TAPE)
+    MakePair("FULL_SOLVER_TAPE", FULL_SOLVER_TAPE)
+};
+
+/*!
+ * \brief Types of variables that can be checked for in a tape debug run.
+ */
+enum ENUM_CHECK_TAPE_VARIABLES {
```
pcarruscag:
```suggestion
enum class CHECK_TAPE_VARIABLES {
```

Common/include/parallelization/vectorization.hpp
```diff
@@ -90,11 +90,11 @@ class Array : public CVecExpr<Array<Scalar_t, N>, Scalar_t> {
  public:
   using Scalar = Scalar_t;
   enum : size_t { Size = N };
-  enum : size_t { Align = Size * sizeof(Scalar) };
+  enum : size_t { Align = Size * 32 };
```
pcarruscag:
```suggestion
  enum : size_t { Align = Size * alignof(Scalar) };
```

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -738,15 +855,17 @@ void CDiscAdjMultizoneDriver::SetObjFunction(RECORDING kind_recording) {
     }
   }
 
+
+
```
pcarruscag:
```suggestion
```

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -178,17 +178,21 @@ void CDiscAdjSolver::RegisterVariables(CGeometry *geometry, CConfig *config, boo
...
 
-  if((config->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE) && (KindDirect_Solver == RUNTIME_FLOW_SYS && !config->GetBoolTurbomachinery())) {
+  if((config->GetKind_Regime() == ENUM_REGIME::COMPRESSIBLE) && (KindDirect_Solver == RUNTIME_FLOW_SYS && !config->GetBoolTurbomachinery())){
 
     su2double Velocity_Ref = config->GetVelocity_Ref();
     Alpha                  = config->GetAoA()*PI_NUMBER/180.0;
     Beta                   = config->GetAoS()*PI_NUMBER/180.0;
     Mach                   = config->GetMach();
-    Pressure               = config->GetPressure_FreeStreamND();
-    Temperature            = config->GetTemperature_FreeStreamND();
```
pcarruscag, line 188: These can be removed from solver members now?

Common/include/basic_types/ad_structure.hpp
oleburghardt: But what is your preferred way to initialize an o**f**stream?

pcarruscag: The one I suggested

Common/include/basic_types/ad_structure.hpp
oleburghardt: What about the non-AD default implementation? It would have to return some reference, whereas I can return `NULL` in this.

Common/include/basic_types/ad_structure.hpp
oleburghardt: std::cout is a std::ostream

Common/src/CConfig.cpp
```diff
@@ -3411,7 +3423,12 @@ void CConfig::SetPostprocessing(SU2_COMPONENT val_software, unsigned short val_i
 
     /*---  Using default frequency of 250 for all files when steady, and 1 for unsteady. ---*/
     for (auto iVolumeFreq = 0; iVolumeFreq < nVolumeOutputFrequencies; iVolumeFreq++){
-      VolumeOutputFrequencies[iVolumeFreq] = Time_Domain ? 1 : 250;
+      if (Multizone_Problem && DiscreteAdjoint) {
+        VolumeOutputFrequencies[iVolumeFreq] = Time_Domain ? 1 : nOuterIter;
+      }
+      else {
+        VolumeOutputFrequencies[iVolumeFreq] = Time_Domain ? 1 : 250;
+      }
```
pcarruscag, line 3435: Is this change important for this PR?

Common/include/basic_types/ad_structure.hpp
pcarruscag: Then use the base type for cout, cerr, ofstream, etc.

Common/include/basic_types/ad_structure.hpp
pcarruscag: I guess the real question is why you need to return a reference or pointer at all.
If the objective is just to read the value you can return by value

Common/src/CConfig.cpp
oleburghardt: No, I would have pointed that out as soon as WIP is removed. There seems to be an issue with the re-recording after having written out adjoint restart files; so that's the only connection with the actual content of this PR.

Common/include/basic_types/ad_structure.hpp
oleburghardt: Probably because I wanted to pass it directly, by pointer or through reference, to some MPI function. Just returning the value is much simpler though.

Common/include/CConfig.hpp
oleburghardt: Thanks!

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
oleburghardt: Thanks! At least Pressure can be removed. Temperature is still used in other cases (this is just for compressible && !turbo).

Those cases probably need some maintenance as well (if not RegisterVariables in general) but I'd prefer to leave that for a separate branch and PR.

Common/include/basic_types/ad_structure.hpp
oleburghardt: Yes it is. It's no longer needed :)

Common/include/basic_types/ad_structure.hpp
oleburghardt: Why do you prefer `auto` over `ErrorReport`?

pcarruscag: Just for not typing the type twice.

Common/src/CConfig.cpp
oleburghardt: Up :)

Common/include/basic_types/ad_structure.hpp
```diff
@@ -259,7 +268,50 @@ inline void SetExtFuncOut(T&& data, const int size_x, const int size_y) {}
...
+ * \brief Struct to store information about errors during a tag debug run.
+ */
+struct ErrorReport {};
+
+/*!
+ * \brief Set a reference to the output file of an ErrorReport.
+ * \param[in] report - the ErrorReport whose output file is set.
+ * \param[in] output_file - pointer to the output file.
+ */
+inline void SetDebugReportFile(struct ErrorReport* report, std::ofstream* output_file) {}
```
pcarruscag:
```suggestion
inline void SetDebugReportFile(ErrorReport& report, std::ostream* output_stream) {}
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -259,7 +268,50 @@ inline void SetExtFuncOut(T&& data, const int size_x, const int size_y) {}
...
+
+/*!
+ * \brief Set the ErrorReport to which error information from a tag debug recording is written.
+ * \param[in] report - the ErrorReport to which error information is written.
+ */
+inline void SetTagErrorCallback(struct ErrorReport* report) {}
```
pcarruscag:
```suggestion
inline void SetTagErrorCallback(ErrorReport& report) {}
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -259,7 +268,50 @@ inline void SetExtFuncOut(T&& data, const int size_x, const int size_y) {}
...
+
+/*!
+ * \brief Reset the error counter in an ErrorReport.
+ * \param[in] report - the ErrorReport whose error counter is resetted.
+ */
+inline void ResetErrorCounter(struct ErrorReport& report) {}
```
pcarruscag:
```suggestion
inline void ResetErrorCounter(ErrorReport& report) {}
```

Common/include/basic_types/ad_structure.hpp
```diff
@@ -259,7 +268,50 @@ inline void SetExtFuncOut(T&& data, const int size_x, const int size_y) {}
...
+
+/*!
+ * \brief Get the error count of an ErrorReport.
+ * \param[in] report - the ErrorReport whose pointer to its error counter is returned.
+ * \return Value of the error counter.
+ */
+inline unsigned long GetErrorCount(ErrorReport& report) { return 0; }
```
pcarruscag:
```suggestion
inline unsigned long GetErrorCount(const ErrorReport& report) { return 0; }
```

Common/include/parallelization/vectorization.hpp
```diff
@@ -90,11 +90,11 @@ class Array : public CVecExpr<Array<Scalar_t, N>, Scalar_t> {
  public:
   using Scalar = Scalar_t;
   enum : size_t { Size = N };
-  enum : size_t { Align = Size * sizeof(Scalar) };
+  enum : size_t { Align = Size * alignof(Scalar) };
   static constexpr bool StoreAsRef = true;
 
  private:
-  alignas(Size * sizeof(Scalar)) Scalar x_[N];
+  alignas(Size * 32) Scalar x_[N];
```
pcarruscag: missed this 32

Common/src/CConfig.cpp
pcarruscag: There is an FSI regression where we output files in the middle and the results do not change.
Your change only changes the default value but does not override the user specified frequency right?

Common/src/CConfig.cpp
```diff
@@ -1184,6 +1190,8 @@ void CConfig::SetConfig_Options() {
   addBoolOption("WRT_VOLUME_OVERWRITE", Wrt_Volume_Overwrite, true);
   /*!\brief SYSTEM_MEASUREMENTS \n DESCRIPTION: System of measurements \n OPTIONS: see \link Measurements_Map \endlink \n DEFAULT: SI \ingroup Config*/
   addEnumOption("SYSTEM_MEASUREMENTS", SystemMeasurements, Measurements_Map, SI);
+  /*!\brief MULTIZONE_ADAPT_FILENAME \n DESCRIPTION: Append zone numer to restart and solution filenames. \ingroup Config*/
```
pcarruscag:
```suggestion
  /*!\brief MULTIZONE_ADAPT_FILENAME \n DESCRIPTION: Append zone number to restart and solution filenames. \ingroup Config*/
```

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
```diff
@@ -141,6 +141,16 @@ class CDiscAdjMultizoneDriver : public CMultizoneDriver {
...
 
+  /*!
+   * \brief [Overload] Launch the tape test mode for the discrete adjoint multizone solver.
+   */
+  void TapeTest();
+
+  /*!
+   * \brief [Overload] Get error numbers after a tape test run of the discrete adjoint multizone solver.
+   */
+  int TapeTest_GatherErrors(struct AD::ErrorReport& error_report);
```
pcarruscag: Can all these things be const? You don't need to add struct to function arguments (I didn't even know this was allowed).
```suggestion
  int TapeTestGatherErrors(const AD::ErrorReport& error_report) const;
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -671,12 +671,12 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
 
     StrainMag(iPoint) = sqrt(2.0*StrainMag(iPoint));
     AD::SetPreaccOut(StrainMag(iPoint));
+    AD::EndPreacc();
 
     /*--- Max is not differentiable, so we not register them for preacc. ---*/
     strainMax = max(strainMax, StrainMag(iPoint));
     omegaMax = max(omegaMax, GeometryToolbox::Norm(3, Vorticity));
 
-    AD::EndPreacc();
```
pcarruscag: Won't this increase the memory used by the tape? Since it's now inside the main recording, it gets pushed onto the tape no?

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -217,6 +229,96 @@ void CDiscAdjMultizoneDriver::StartSolver() {
...
+int CDiscAdjMultizoneDriver::TapeTest_GatherErrors(struct AD::ErrorReport& error_report) {
+
+  int num_errors = AD::GetErrorCount(error_report);
+  int total_errors = 0;
+  std::vector<int> process_error(size);
+  SU2_MPI::Allreduce(&num_errors, &total_errors, 1, MPI_INT, MPI_SUM, SU2_MPI::GetComm());
+  SU2_MPI::Gather(&num_errors, 1, MPI_INT, process_error.data(), 1, MPI_INT, 0, SU2_MPI::GetComm());
+
+  if (rank == MASTER_NODE) {
+    std::cout << "\nNumber of detected tape inconsistencies: " << total_errors << std::endl;
```
pcarruscag:
```suggestion
    std::cout << "\nTotal number of detected tape inconsistencies: " << total_errors << std::endl;
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
oleburghardt: So what happens is that e.g. `omegaMax` depends on `Vorticity` which is not registered as input. I think this is probably done by intention, at least in case I understand the comment above correctly. However the tape test/debug mode detects that we've cut a dependency (through e.g. `Vorticity[2]` carrying a non-zero tag that appears inside a preaccumulation region where only tags "1337" are allowed).
What we can do is to either leave it this way (increasing memory, yes), to register `Vorticity` as input (increasing memory, too), or to enforce the correct tag alongside the intentional(?) omission of the max-dependencies.

Common/src/CConfig.cpp
oleburghardt: Good to know! Yes (I've just double checked).

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
oleburghardt: Which other functions are you thinking of?

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
oleburghardt: I thought that excluding the max computations was the clean solution that we are looking for but it made a regression test fail. So whatever the remaining dependency is, there seems to be one. So I chose to fix the preaccumulation region by adding `Vorticity` as an input.

Common/include/basic_types/ad_structure.hpp
```diff
@@ -259,7 +279,50 @@ inline void SetExtFuncOut(T&& data, const int size_x, const int size_y) {}
...
+ * \param[in] report - the ErrorReport whose error counter is resetted.
+ */
+inline void ResetErrorCounter(ErrorReport& report) {}
+
+/*!
+ * \brief Get the error count of an ErrorReport.
+ * \param[in] report - the ErrorReport whose pointer to its error counter is returned.
+ * \return Value of the error counter.
+ */
+inline unsigned long GetErrorCount(ErrorReport& report) { return 0; }
```
pcarruscag:
```suggestion
inline unsigned long GetErrorCount(const ErrorReport& report) { return 0; }
```

SU2_CFD/include/drivers/CDiscAdjMultizoneDriver.hpp
pcarruscag: Everything that can be const, should be const.

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -646,7 +646,10 @@ void CFVMFlowSolverBase<V, R>::ComputeVorticityAndStrainMag(const CConfig& confi
     const su2double vy = nodes->GetVelocity(iPoint, 1);
     const su2double y = geometry->nodes->GetCoord(iPoint, 1);
     AD::StartPreacc();
+    AD::SetPreaccIn(omegaMax);
+    AD::SetPreaccIn(strainMax);
     AD::SetPreaccIn(VelocityGradient, nDim, nDim);
+    AD::SetPreaccIn(Vorticity, 3);
```
pcarruscag: The tape check detects an issue, but is it really an issue in normal AD modes? Can we put the extra preaccumulation behind some #ifdef ADJOINT_DEBUG so that we still get memory savings when possible?

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -217,6 +229,96 @@ void CDiscAdjMultizoneDriver::StartSolver() {
...
 }
 
+void CDiscAdjMultizoneDriver::TapeTest() {
+
+  if (rank == MASTER_NODE) {
+    cout <<"\n---------------------------- Start Debug Run ----------------------------" << endl;
+  }
+
+  int total_errors = 0;
+  struct AD::ErrorReport error_report;
```
pcarruscag: Please look for all the unnecessary `struct`
```suggestion
  AD::ErrorReport error_report;
```

SU2_CFD/src/drivers/CDiscAdjMultizoneDriver.cpp
```diff
@@ -217,6 +229,96 @@ void CDiscAdjMultizoneDriver::StartSolver() {
...
+  }
+}
+
+int CDiscAdjMultizoneDriver::TapeTest_GatherErrors(struct AD::ErrorReport& error_report) {
+
+  int num_errors = AD::GetErrorCount(error_report);
+  int total_errors = 0;
+  std::vector<int> process_error(size);
+  SU2_MPI::Allreduce(&num_errors, &total_errors, 1, MPI_INT, MPI_SUM, SU2_MPI::GetComm());
+  SU2_MPI::Gather(&num_errors, 1, MPI_INT, process_error.data(), 1, MPI_INT, 0, SU2_MPI::GetComm());
```
pcarruscag:
```suggestion
  SU2_MPI::Gather(&num_errors, 1, MPI_INT, process_error.data(), 1, MPI_INT, MASTER_NODE, SU2_MPI::GetComm());
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
oleburghardt: I think it's really an issue;  either the computation of `omegaMax` and `strainMax` shall be on the tape or not. If we exclude their computation from the taping, a regression test fails, as I said. If we want them in the recording, and the in preaccumulation, we have do it correctly, I think. So, what do we want?

pcarruscag (approved):

PR #2453, opened 2025-02-25
Update CTurbSSTSolver.cpp to fix bug on calculating MuT

ShiheJia:
## Proposed Changes
When calculating the turb viscosity MuT for the SST 2003 version, the variable StrainMag is needed. However, in the current code, we use the wrong CVariable class to get the StrainMag so that we only get 0. The turb variable class should be replaced by the flow variable class.
CurrentCode

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
+++ b/SU2_CFD/src/solvers/CTurbSSTSolver.cpp
@@ -230,7 +230,7 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
     const su2double dist = geometry->nodes->GetWall_Distance(iPoint);
 
     const su2double VorticityMag = max(GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint)), 1e-12);
-    const su2double StrainMag = max(nodes->GetStrainMag(iPoint), 1e-12);
+    const su2double StrainMag = max(flowNodes->GetStrainMag(iPoint), 1e-12);
     nodes->SetBlendingFunc(iPoint, mu, dist, rho, config->GetKind_Trans_Model());
 
     const su2double F2 = nodes->GetF2blending(iPoint);


bigfooted:
Yep, setStrainMag updates the flownodes only. Can you update the regression tests?

ShiheJia:
I run a flatplate case and the results are shown below.
It seems that the fixed code can accelerate the convergence very slightly, but the results deviates from SA a lot. I'm not sure how this happened and whether it was harmful.
github-1
github-2
github-3
github-4

ShiheJia:
> Yep, setStrainMag updates the flownodes only. Can you update the regression tests?

A flatplate case is listed below.

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
```diff
@@ -230,7 +230,7 @@ void CTurbSSTSolver::Postprocessing(CGeometry *geometry, CSolver **solver_contai
     const su2double dist = geometry->nodes->GetWall_Distance(iPoint);
 
     const su2double VorticityMag = max(GeometryToolbox::Norm(3, flowNodes->GetVorticity(iPoint)), 1e-12);
-    const su2double StrainMag = max(nodes->GetStrainMag(iPoint), 1e-12);
+    const su2double StrainMag = max(flowNodes->GetStrainMag(iPoint), 1e-12);
```
pcarruscag, line 233: Have you looked for other uses of GetStrainMag to check if they are ok?

pcarruscag (approved):

SU2_CFD/src/solvers/CTurbSSTSolver.cpp
ShiheJia: I have checked the other part of code and they are all ok.

bigfooted:
> > Yep, setStrainMag updates the flownodes only. Can you update the regression tests?
> 
> A flatplate case is listed below.
Here are the results from the nasa website. It looks like they also see a difference between SA and SST, maybe you can check if this corresponds to your results 
https://turbmodels.larc.nasa.gov/flatplate.html

Our validation case is here:
https://su2code.github.io/tutorials/Turbulent_Flat_Plate/
is this the same as your setup?

ShiheJia:
> > > Yep, setStrainMag updates the flownodes only. Can you update the regression tests?
> > 
> > 
> > A flatplate case is listed below.
> > Here are the results from the nasa website. It looks like they also see a difference between SA and SST, maybe you can check if this corresponds to your results
> > https://turbmodels.larc.nasa.gov/flatplate.html
> 
> Our validation case is here: https://su2code.github.io/tutorials/Turbulent_Flat_Plate/ is this the same as your setup?

The case above is just the SU2 validation case, with a finer mesh.
The results from NASA do show that SST will get a lower cf, however, it seems that the difference between SA and SST is larger in SU2.

ShiheJia:
github-5

joshkellyjak:
> I run a flatplate case and the results are shown below. It seems that the fixed code can accelerate the convergence very slightly, but the results deviates from SA a lot. I'm not sure how this happened and whether it was harmful. 

Are you confident that this is converging? Two orders of magnitude seems rather small, maybe this could explain the discrepancy you are observing?

bigfooted:
The case is not converging but stalling, this can also be a reason that drag is not going down. maybe reduce the cfl.

pcarruscag:
Converging well or not, this was certainly a bug.
@ShiheJia do you know what is the process to update the regression tests?

ShiheJia:
> Converging well or not, this was certainly a bug. @ShiheJia do you know what is the process to update the regression tests?

Actually I know a little about that.

ShiheJia:
> > From the CD, I think the case has converged. I will try the mesh from NASA, and we can exclude the grid effects.
> 
> Cool, I eagerly await the results!

It seems the new code is alright. The new SST in SU2 is closer to CFL3D and FUN3D, although there is still small difference.
github-1
github-2
github-3
github-4

bigfooted:
ok, those results, especially Cf_x, look pretty convincing. What did you change?

bigfooted:
For updating regression tests, you will se a lot of cases that use the SST model that now have different results:

> test_vals (stored): -4.029282, -2.181911, -7.734686, 0.000000, -0.939944
> sim_vals (computed): -3.597708, -2.983823, -8.354806, 0.000000, -0.916525

So you will need to change the 'stored' values in Testcases/parallel_regression.py, serial_regression.py, etc. with the 'computed' values. You can download the logfiles from github and use them to update the regression files.

ShiheJia:
I use the config and mesh files in https://github.com/su2code/VandV/tree/master/rans/flatplate.
The provided mesh is finer in the normal direction than the initial mesh.

joshkellyjak:
> It seems the new code is alright. The new SST in SU2 is closer to CFL3D and FUN3D, although there is still small difference.

Beautiful!

ShiheJia:
> For updating regression tests, you will se a lot of cases that use the SST model that now have different results:
> 
> > test_vals (stored): -4.029282, -2.181911, -7.734686, 0.000000, -0.939944
> > sim_vals (computed): -3.597708, -2.983823, -8.354806, 0.000000, -0.916525
> 
> So you will need to change the 'stored' values in Testcases/parallel_regression.py, serial_regression.py, etc. with the 'computed' values. You can download the logfiles from github and use them to update the regression files.

I have tried a lot but I failed to do that. I am not very familiar with the Github process. Maybe someone more skilled can help with updating the regression files.

pcarruscag:
I'll merge into another branch where we need to update tests for SST

Issue #2390, opened 2024-12-04
Symmetry BC not working (perfectly)


rois1995:
Hi all,

I am simulating a 2D channel with two triangular bumps in a Mach = 2.0 inviscid flow. When slicing the solution along x-normal planes, I can see that the symmetry condition does not lead to zero-gradients in the normal direction at the euler wall. Only the pressure variable has a zero-gradient at the wall. The situations doesn't change if I try with Roe, AUSMPLUSUP or JST schemes, nor if I change the way gradients are computed. Also refining the mesh does not improve the solution. I tried running the same case reverting SU2 to the commit previous to the implementation of the new symmetry BC and it works fine. The euler walls are the top and the bottom ones.

Here is the general flow solution colored by the Mach number
![Cuts](https://github.com/user-attachments/assets/351a869e-a05e-4263-9d79-418dc06880ae)

Here the solution with the new symmetry BC
![Density_Roe](https://github.com/user-attachments/assets/9ae1374d-d174-4010-8d10-9d31864f6e36)

Here the solution with the old symmetry BC
![Density_Roe](https://github.com/user-attachments/assets/71d90899-157a-4e64-853e-621888d2f8bb)

I have also tried with a subsonic flow and the result is the same. Notice how before the triangular bump the gradients are zero at the euler walls (red line). 

All the simulations have converged to a RMS of the residual density of $10^{-13}$.

Here you can find the mesh and the config file.
https://polimi365-my.sharepoint.com/:f:/g/personal/10507725_polimi_it/Em_r77HJAJpOiceuw5YSeAMB2CDPDoI5RJT_4CTe9fj7Aw?e=zz2jBN

I think that the problem is that when setting the normal gradients to 0 in the code then we are unable to reconstruct the primitive variables at the mid-point of the wall normal edges.  

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - C++ compiler and version: gcc 9.4.0
 - MPI implementation and version: 4.0.3
 - SU2 Version: v8, commit 77fed34

bigfooted:
It looks like the energy also has nonzero normal gradients at the wall...
Gradients are corrected in the correctGradientsSymmetry routine, called from green-gauss or least-squares. We get the local normal and then make a call to correctGradient, where we remove the normal component. All solver variables are corrected here, so not sure what is going on here. Any ideas, @pcarruscag ?

rois1995:
[...]

PR #2460, opened 2025-03-05
Fix curved symmetries

bigfooted:
## Proposed Changes
For nodes that are shared by 2 or more curved symmetries (usually Euler walls) we do not need to apply Gram Schmidt orthogonalization to correct the face normals. Instead, we need to add the normals to retrieve the same normals as if these faces were merged. For straight symmetries (could be euler walls as well but usually symmetry planes) we always perform  GSO when nodes are shared with another symmetry (straight or not).

## Related Work

Reported on slack:
"Hi all. I am running a 3D euler simulation on a vehicle. I build the mesh in gmsh. the vehicle surface is composed by several gmsh surfaces. initially I had 1 boundary condition for the whole vehicle surface, with marker name "wall", in which I applied a marker Euler.for mesh adaptation purposes, I need one of the gmsh surfaces to have a different marker name, called "no_adapt", in which I also apply a marker Euler.So I have MARKER_EULER = (wall) vs. MARKER_EULER = (wall, no_adapt)with 1 marker, the surface solution is smooth in the region of interest. with 2 markers, I would expect the same outcome, since I use the same BC, simple divide it with 2 different names. However, the solution at the interface between the two markers goes really weird.I already checked the mesh generation several times and I cannot find any issue on the mesh generation end.So I'm left with wondering if the problem could be on the SU2 side ... Anyone has any feedback on this?"

## PR Checklist


rois1995:
Is this related to the problems I was encountering in #2390?

bigfooted:
> Is this related to the problems I was encountering in #2390?

No, this is a fix for nodes that are shared by 2 curved Euler walls. For your issue, I am not sure what the root cause is. Did you work more on this?

rois1995:
Unfortunately not, I am trying to close some other PR that I have open from long ago. I guess it is still a problem with normals in the point where the slope changes direction. I will try to have a look into it whenever I can.

SU2_CFD/include/gradients/correctGradientsSymmetry.hpp
```diff
@@ -130,6 +130,7 @@ void correctGradientsSymmetry(CGeometry& geometry, const CConfig& config, const
 
       if (it != geometry.symmetryNormals[iMarker].end()) {
         for (auto iDim = 0u; iDim < nDim; iDim++) unitNormal[iDim] = it->second[iDim];
+
```
bigfooted, line 133:
```suggestion
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1136,6 +1136,7 @@ void CFVMFlowSolverBase<V, FlowRegime>::BC_Sym_Plane(CGeometry* geometry, CSolve
 
     if (it != geometry->symmetryNormals[val_marker].end()) {
       for (auto iDim = 0u; iDim < nDim; iDim++) UnitNormal[iDim] = it->second[iDim];
+
```
bigfooted, line 1139:
```suggestion
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -2491,15 +2560,22 @@ void CGeometry::ComputeModifiedSymmetryNormals(const CConfig* config) {
...
+          isShared = false;
+          continue;
+        }
+
         std::array<su2double, MAXNDIM> jNormal = {};
         const auto it = symmetryNormals[jMarker].find(jVertex);
 
         if (it != symmetryNormals[jMarker].end()) {
           jNormal = it->second;
+
```
bigfooted, line 2574:
```suggestion
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -2456,16 +2456,73 @@ su2double CGeometry::GetSurfaceArea(const CConfig* config, unsigned short val_ma
...
+        /*--- Average the normals on the shared nodes. Also normalize them again. ---*/
+        for (auto iDim = 0u; iDim < nDim; iDim++) kNormal[iDim] = iNormal[iDim] + jNormal[iDim];
+        const su2double karea = GeometryToolbox::Norm(nDim, kNormal.data());
+        for (auto iDim = 0u; iDim < nDim; iDim++) kNormal[iDim] /= karea;
+
+        for (auto iDim = 0u; iDim < nDim; iDim++) {
+          symmetryNormals[iMarker][iVertex][iDim] += kNormal[iDim];
+          symmetryNormals[jMarker][jVertex][iDim] += kNormal[iDim];
+        }
```
pcarruscag, line 2519: +=? not just =? What if there are more than 2 markers?

Common/src/geometry/CGeometry.cpp
```diff
@@ -2491,15 +2550,21 @@ void CGeometry::ComputeModifiedSymmetryNormals(const CConfig* config) {
 
         isShared = true;
 
+        /*--- Both symmetries are curved, no further correction. ---*/
+        if ((bound_is_straight[iMarker]) == false && (bound_is_straight[jMarker]) == false) {
```
pcarruscag, line 2554:
```suggestion
        /*--- Both symmetries are curved, no further correction. ---*/
        if (!boundIsStraight[iMarker] && !boundIsStraight[jMarker]) {
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -2480,6 +2537,8 @@ void CGeometry::ComputeModifiedSymmetryNormals(const CConfig* config) {
       /*--- Get the vertex normal on the current symmetry. ---*/
       std::array<su2double, MAXNDIM> iNormal = {};
       vertex[iMarker][iVertex]->GetNormal(iNormal.data());
+      const su2double iarea = GeometryToolbox::Norm(nDim, iNormal.data());
+      for (auto iDim = 0u; iDim < nDim; iDim++) iNormal[iDim] /= iarea;
```
pcarruscag, line 2541: Probably not needed?

Common/src/geometry/CGeometry.cpp
```diff
@@ -2456,16 +2456,73 @@ su2double CGeometry::GetSurfaceArea(const CConfig* config, unsigned short val_ma
...
+        } else {
+          vertex[jMarker][jVertex]->GetNormal(jNormal.data());
+          const su2double jarea = GeometryToolbox::Norm(nDim, jNormal.data());
+          for (auto iDim = 0u; iDim < nDim; iDim++) jNormal[iDim] /= jarea;
+        }
+
+        /*--- Average the normals on the shared nodes. Also normalize them again. ---*/
+        for (auto iDim = 0u; iDim < nDim; iDim++) kNormal[iDim] = iNormal[iDim] + jNormal[iDim];
+        const su2double karea = GeometryToolbox::Norm(nDim, kNormal.data());
+        for (auto iDim = 0u; iDim < nDim; iDim++) kNormal[iDim] /= karea;
```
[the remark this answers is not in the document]
bigfooted: Ah yes, thanks!

Common/src/geometry/CGeometry.cpp
bigfooted: ok, I will now first add all normals (n_i + n_j) for all markers here, then += is correct. I then normalize after this routine.

Common/src/geometry/CGeometry.cpp
```diff
@@ -2456,16 +2456,78 @@ su2double CGeometry::GetSurfaceArea(const CConfig* config, unsigned short val_ma
...
+          vertex[jMarker][jVertex]->GetNormal(jNormal.data());
+        }
+
+        /*--- Average the normals on the shared nodes. ---*/
+        for (auto iDim = 0u; iDim < nDim; iDim++) kNormal[iDim] = iNormal[iDim] + jNormal[iDim];
+
+        for (auto iDim = 0u; iDim < nDim; iDim++) {
+          symmetryNormals[iMarker][iVertex][iDim] += kNormal[iDim];
+          symmetryNormals[jMarker][jVertex][iDim] += kNormal[iDim];
+        }
```
pcarruscag, line 2513: I still don't think this is correct, I'll write it someway that expresses what we want better.

bigfooted: Thanks. 
I thought what we should do is for a vertex shared by markers m0,m1,m2 we find the shared markers (m1,m0), (m2,m0) and (m2,m1) and we sum (n1+n0) + (n2+n0) + (n2+n1) = 2(n0+n1+n2)  (and then we normalize).

pcarruscag:
We may want to fabricate a case that has all these things, 3 merged symmetries, intersecting a plane.

bigfooted:
> We may want to fabricate a case that has all these things, 3 merged symmetries, intersecting a plane.

Like the flow around a wall mounted cube, with a vertical symmetry, and then split the square cube surfaces diagonally.
image

pcarruscag:
Maybe a cylinder otherwise all surfaces are flat.

bigfooted:
3D flow over wall-mounted cylinder (Re=1)

image

pcarruscag:
How many shared symmetries does it have?

bigfooted:
I changed the 'flow over wall mounted cylinder'  such that the leading edge lower corner now has 3 Euler walls and 1 symmetry. So basically every side is a symmetry or an Euler wall except inlet and outlet.
The trailing edge has 2 Euler walls and 1 symmetry
image

bigfooted:
The solution looks  bit wobbly because of the course mesh, but there is now jump over the interfaces.

bigfooted:
During convergence, there are sometimes some peaks in the residual plot, caused by the 4-symmetry points:
image
But the residuals recover quickly and in the end it converges.

pcarruscag:
Let's hope it's good enough for some months without symmetry issues. Can you add the case? I can update the residuals again.

pcarruscag (approved):

PR #2446, opened 2025-02-18
Multizone adjoints for turbomachinery

joshkellyjak:
## Proposed Changes

This is a cleaned up PR of the fixes needed for multizone adjoints for turbomachinery from the previous PR of @oleburghardt and I's work. This PR refactors the mixing plane implementation and some of the turbo specific features to make them tape consistent.

## Related Work
Now closed PR #2317 

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [ X ] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [x] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [x] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [x] I have added a test case that demonstrates my contribution, if necessary.
- [x] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

PR #2472, opened 2025-03-21
Fix preaccumulation error, improve recording of RANS solvers

oleburghardt:
## Proposed Changes
This contributes two algorithmic-differentiation-based fixes;

- Add a missing preaccumulation input to the recording of `SetRoe_Dissipation_FD`. This may change derivatives as the current recording of this function is likely to be wrong if `PREACC = YES`.
- Include `muT` (eddy viscosity) as input/output to the recording of RANS solvers. Since both muT and conservative variables go _"on the same level"_  into the flow solver's `Preprocessing` (i.e. we cannot think of muT as a derived value from the conservative ones), this should be reflected in the recording. (Also, the debug mode developed in #2442 will complain otherwise ;-)).
	This change does not affect derivative values – especially it does nothing about the circumstance that Preprocessing is based on an initial value of muT, whereas the flow solvers depend on an updated one – but it may clarify things and remove a source of future problems.

## Related Work

- Both fixes are a by-product of working on #2446
- The tape recording debug mode that can detect such kind of issues (#2442) needs some more features before being ready to be merged.

## PR Checklist


SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -137,6 +137,20 @@ class CVariable {
     RegisterContainer(input, variable, &ad_index);
   }
 
+  void RegisterContainer(bool input, su2activevector& variable, su2vector<int>* ad_index = nullptr) {
```
oleburghardt, line 140: I don't like this almost-duplicate but I couldn't decide on the best alternative, too.

SU2_CFD/src/variables/CNSVariable.cpp
```diff
@@ -114,6 +114,8 @@ void CNSVariable::SetRoe_Dissipation_FD(unsigned long iPoint, su2double val_wall
   AD::SetPreaccIn(Primitive(iPoint, indices.EddyViscosity()));
   /*--- Laminar viscosity --- */
   AD::SetPreaccIn(Primitive(iPoint, indices.LaminarViscosity()));
+  /*--- Density ---*/
+  AD::SetPreaccIn(Solution(iPoint, 0));
```
pcarruscag, line 118:
```suggestion
  AD::SetPreaccIn(Solution(iPoint, indices.Density()));
```

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2133,6 +2147,11 @@ class CVariable {
    */
   inline virtual void RegisterFlowTraction(bool reset) { }
 
+  /*!
+   * \brief A virtual member.
+   */
+  inline virtual void RegisterEddyViscosity(bool input) { }
+
```
pcarruscag, line 2154: Probably doesn't need to be virtual
```suggestion
```

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -164,6 +164,9 @@ void CDiscAdjSolver::RegisterSolution(CGeometry *geometry, CConfig *config) {
 
   direct_solver->RegisterSolutionExtra(true, config);
 
+  /*--- Register quantities that are no solver variables but further inputs/outputs of the (outer) iteration. ---*/
+  direct_solver->RegisterComplementary(true, config);
```
pcarruscag, line 168: Can't you put this in solution extra?

SU2_CFD/include/variables/CVariable.hpp
pcarruscag: Templating the other one didn't work?

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
oleburghardt: Yes! Actually that's what I did originally but RegisterSolutionExtra is also used in the constructor to determine an extra allocation of adjoint variables. I let you decide, I've tested both solutions..

pcarruscag: Yep move it there please, we already have quite a variety of things to register, solution, extra, variables, geometry, :rofl:

SU2_CFD/src/variables/CNSVariable.cpp
oleburghardt: indices.Density() is for the position in `Primitive`, I think..

pcarruscag: what a pain this code :( then please mention in the comment that this is for `GetDensity` below, because if we change that to get density from somewhere else it breaks.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2133,6 +2147,11 @@ class CVariable {
    */
   inline virtual void RegisterFlowTraction(bool reset) { }
 
+  /*!
+   * \brief A virtual member.
+   */
+  inline void RegisterEddyViscosity(bool input) { }
+
```
pcarruscag, line 2154: Does it compile without this?
```suggestion
```

pcarruscag (approved):

pcarruscag:
@oleburghardt if everything is addressed can you resolve the conversations and merge please?

oleburghardt:
> @oleburghardt if everything is addressed can you resolve the conversations and merge please?

Sure. CodeFactor is complaining about CSolver.hpp (complex code) which I've touched in a commit but which eventually remains unchanged. Forgot to ask during last dev meeting how we should handle it?

pcarruscag:
You can ignore that warning, It doesn't say what to do so we can't do anything.

PR #2484, opened 2025-03-26
Fix wall distance computation for points on viscous wall

rois1995:
## Proposed Changes
This PR fixes the problems encountered in #2474, where for some points on a viscous wall the wall distance was higher than EPS, thus the normal distance introduced in #2392 will be wrong for any point connected to the incriminated ones. This will cause a large Omega at the wall when the SST model is considered.

Develop:
Problem

Current branch:
NoProblem

Here I initialize the wall distance on all the points belonging to a viscous boundary to 0, thus it won't be overwritten by the wall distance computed by the ADT tree. 

## Related Work
#2474 #2392 

## PR Checklist


Common/src/geometry/CPhysicalGeometry.cpp
```diff
@@ -10559,7 +10559,9 @@ void CPhysicalGeometry::SetWallDistance(CADTElemClass* WallADT, const CConfig* c
 
         WallADT->DetermineNearestElement(nodes->GetCoord(iPoint), dist, markerID, elemID, rankID);
 
-        if (dist < nodes->GetWall_Distance(iPoint)) {
+        if (nodes->GetWall_Distance(iPoint) == 0) dist = 0;
+
+        if (dist < nodes->GetWall_Distance(iPoint) || dist == 0) {
```
rois1995, line 10564: Here I have forced the assignment of the wall distance even in case the point is on a viscous boundary just to be sure that the other fields are correctly set (markerID etc..). I am not sure if this is necessary or not, I have added it for safety. I have tried removing it and it didn't change, although I have to try with roughness.

rois1995:
> This makes sense but I think there is an easier way, change
> 
> ```
>   for (const auto jPoint : geometry->nodes->GetPoints(iPoint)) {
>     const su2double dist = geometry->nodes->GetWall_Distance(jPoint);
>     if (dist > EPS) distance = fmin(distance, dist);
>   }
> ```
> 
> to skip points on viscous walls (essentially what you are doing here).

I also was thinking about it but how do I know if the iPoint is on a viscous wall?

pcarruscag:
There is something in `geometry->nodes->` I think.

rois1995:
> There is something in `geometry->nodes->` I think.

Found it, geometry->nodes->GetViscousBoundary(iPoint) I guess.

Isn't the proposed fix more general though? As the wall distance on the points on a viscous domain is exactly zero, so we don't have this kind of problems if it is ever needed anywhere else in the same way. Maybe I can do it within the CPhysicalGeometry function, something like

```
void SetWallDistance(su2double val, bool init=false) override {
    su2double valHere = 0.0;
    for (unsigned long iPoint = 0; iPoint < GetnPoint(); iPoint++) {
      valHere = val;
      if (init && nodes->GetViscousBoundary(iPoint)) valHere = 0;
      nodes->SetWall_Distance(iPoint, valHere);
    }
  }
```

In either cases I think it is a larger overhead than just cycling on all the viscous points and setting it to 0 after the first uniform initialization.

pcarruscag:
No I don't think it is because there is nothing wrong with the wall distance computation, the issue is in this normal neighbor distance function that is using an indirect way of detecting boundaries. Fix it there please.

Common/include/geometry/CGeometry.hpp
```diff
@@ -1843,6 +1843,13 @@ class CGeometry {
    */
   virtual void SetWallDistance(su2double val) {}
 
+  /*!
+   * \brief Set wall distances a specific value and at a specific point
+   *  \param[in] iPoint - i-th point to set the wall distance.
+   *  \param[in] val - new value for the wall distance at the input point.
+   */
+  virtual void SetWallDistance(unsigned long iPoint, su2double val) {}
+
```
pcarruscag, line 1852:
```suggestion
```

Common/include/geometry/CPhysicalGeometry.hpp
```diff
@@ -811,6 +811,11 @@ class CPhysicalGeometry final : public CGeometry {
     }
   }
 
+  /*!
+   * \brief Set wall distance of i-th point to a specific value
+   */
+  void SetWallDistance(unsigned long iPoint, su2double val) override { nodes->SetWall_Distance(iPoint, val); }
+
```
pcarruscag, line 818:
```suggestion
```

pcarruscag (approved):

YairMO:
Hi,

Would this guarantee zero wall distance at the wall? Or in other words, would this cover all possible cases of geomtries?
Why is guaranteeing a zero-wall distance not sufficient? Is it expensive in some way, or will it fail?

YairMO:
Agreed. 
I may have completely misunderstood the issue, but I thought that it was found that a node on the wall has a very small value of wall distance (for some reason). This value "passed" the if statement. As a result, the assigned distance for this node was its own (false) distance. Am I wrong?

Thank you, and sorry for the trouble

pcarruscag:
The small but nonzero value was interfering with the new way of computing the distance to the nearest node.

PR #2506, opened 2025-05-14
Consistent names for the variables of the inlet profile

vdweide:
## Proposed Changes
Small PR, which changes the names of the variables of the template inlet profile file to the correct names.
The names that were written to the template profile file for inlets were not consistent with the actual variables for some boundary conditions. E.g. for a supersonic inlet the velocity components are written, but the names used were NORMAL_X, NORMAL_Y and NORMAL_Z, which is confusing. Something similar for the total inlet conditions for a subsonic inlet for which the total temperature and total pressure are written, but the names were TEMPERATURE and PRESSURE. This has been corrected. 

## Related Work
None

## PR Checklist

### DIFF
--- a/SU2_CFD/src/solvers/CSolver.cpp
+++ b/SU2_CFD/src/solvers/CSolver.cpp
@@ -3689,20 +3689,17 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
       case TURB_FAMILY::SA:
         /*--- 1-equation turbulence model: SA ---*/
         columnName << "NU_TILDE         " << setw(24);
-        columnValue << config->GetNuFactor_FreeStream() * config->GetViscosity_FreeStream() /
-                           config->GetDensity_FreeStream()
-                    << "\t";
+        columnValue << config->GetNuFactor_FreeStream() * config->GetViscosity_FreeStream() / config->GetDensity_FreeStream() <<"\t";
         break;
       case TURB_FAMILY::KW:
         /*--- 2-equation turbulence model (SST) ---*/
         columnName << "TKE              " << setw(24) << "DISSIPATION      " << setw(24);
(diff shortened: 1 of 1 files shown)


vdweide:
Don't look at it yet. I accidentally did a reformatting in VSCode of the file. Let me undo that first.

vdweide:
> Don't look at it yet. I accidentally did a reformatting in VSCode of the file. Let me undo that first.

Should be good now.

SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -3619,38 +3619,42 @@ void CSolver::LoadInletProfile(CGeometry **geometry,
...
+    columnValue << t_value << "\t" << p_value << "\t";
     for (unsigned short iDim = 0; iDim < nDim; iDim++) {
-      columnValue << flow_dir[iDim] <<"\t";
+      columnValue << flow_dir_or_vel[iDim] << "\t";
     }
 
-    columnName << "# COORD-X  " << setw(24) << "COORD-Y    " << setw(24);
-    if(nDim==3) columnName << "COORD-Z    " << setw(24);
+    columnName << "# COORD-X        " << setw(24) << "COORD-Y          " << setw(24);
+    if (nDim == 3) columnName << "COORD-Z          " << setw(24);
```
pcarruscag, line 3646: We should not need the blank spaces if were using `setw` maybe the `setw`s need to be before the string instead of after.

pcarruscag (approved):

SU2_CFD/src/solvers/CSolver.cpp
vdweide: > We should not need the blank spaces if were using `setw` maybe the `setw`s need to be before the string instead of after.

You're right. But in order to have a left alignment you should use the following construction
`columnName << left << setw(24) << "VARNAME"`
It looks neater now.

PR #2516, opened 2025-06-16
Fix damping mass outlet

bigfooted:
## Proposed Changes
Only use the damping factor on the target massflow when dP is relatively large compared to P. This greatly improves the convergence rate and prevents stalling of the massflow rate.

## PR Checklist


bigfooted:
our entire incompressible ideal gas:
image

pcarruscag:
Then two equivalent incompressible cases with just a different pressure level will behave differently with this change, right?
Perhaps a criterion based on the range of pressures would be more invariant.

bigfooted:
I checked and we never actually add a reference pressure anywhere, and the thermodynamic pressure is only used to compute the (incompressible ideal gas) density. So I think it's fine the way it is currently done.

pcarruscag:
If you set up a case where you put all your pressures as absolute, and another where you put them as relative (to ambient, or to one of the inlets, or outlets) they will not behave the same with this logic.
One solution is to use a reference based on the difference between min and max pressure in the domain, this quantity should be the same regardless of whether you use relative or absolute pressures.
This is what the Venkat-Wang limiter does to scale the limiter constant.

bigfooted:
OK, thanks, I understand. I can loop over all inlets and outlets and get the P_max/P_min ratio

bigfooted:
Is it sufficient here to subtract the minimum pressure in the domain and use that as the reference pressure, so we compare the update pressure dp to the local pressure (P-P_min_global)

pcarruscag:
From the invariance perspective, yes, whether it is a good enough way of deriving a reference value, I don't know.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1972,6 +1976,45 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
 
 }
 
+void CIncEulerSolver::SetRange_Pressure(CGeometry *geometry, CSolver **solver_container,
+                                        CConfig *config, unsigned short iMesh) {
+  static su2double MinP,MaxP;
```
pcarruscag:
```suggestion
  static su2double MinP, MaxP;
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1972,6 +1976,45 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
...
+
+  if (iMesh == MESH_0) {
+    SU2_OMP_MASTER
+    MinP = 0.0;
+    MaxP = 0.0;
```
pcarruscag: Init min to a large value and max to a large negative value (we have them defined somewhere I think) otherwise 0 may become the min or max

Common/include/CConfig.hpp
```diff
@@ -9215,6 +9216,18 @@ class CConfig {
...
+   * \brief Get the maximum pressure (pmax - pmin) in the domain.
+   * \return Value of the maximum pressure in the domain.
+   */
+  su2double GetRange_Pressure(int minmax) const { return Range_Pressure[minmax]; }
+
+  /*!
+   * \brief Set the maximum pressure in the domain.
+   * \param[in] Value of the maximum pressure in the domain.
+   */
+  void SetRange_Pressure(su2double val_dp_min,su2double val_dp_max) { Range_Pressure[0] = val_dp_min;Range_Pressure[1]=val_dp_max; }
```
pcarruscag: No more underscores in CamelCase variables please :) I should look for an AI to cleanup the entire code of this :rofl: 
```suggestion
  su2double GetRangePressure(int minmax) const { return RangePressure[minmax]; }

  /*!
   * \brief Set the maximum pressure in the domain.
   * \param[in] Value of the maximum pressure in the domain.
   */
  void SetRangePressure(su2double val_dp_min,su2double val_dp_max) { RangePressure[0] = val_dp_min;RangePressure[1]=val_dp_max; }
```

pcarruscag (approved):
Do we have a test that uses this?

Common/include/CConfig.hpp
bigfooted: lol yes, it is so common that I just take over the existing naming style all the time instead of thinking for myself :-)

bigfooted:
> Do we have a test that uses this?
Yes, the nozzle case that is failing in the regression tests.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1974,6 +1978,45 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
...
 
+void CIncEulerSolver::SetRangePressure(CGeometry *geometry, CSolver **solver_container,
+                                        CConfig *config, unsigned short iMesh) {
+  static su2double MinP, MaxP;
+
+  if (iMesh == MESH_0) {
+    SU2_OMP_MASTER
+    MinP = +1.0e12;
+    MaxP = -1.0e12;
+    END_SU2_OMP_MASTER
```
pcarruscag:
```suggestion
    SU2_OMP_MASTER {
      MinP = std::numeric_limits<su2double>::max();
      MaxP = std::numeric_limits<su2double>::lowest();
    }
    END_SU2_OMP_MASTER
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1974,6 +1978,45 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
...
+    su2double minP = +1.0e12;
+    su2double maxP = -1.0e12;
```
pcarruscag:
```suggestion
    su2double minP = std::numeric_limits<su2double>::max();
    su2double maxP = std::numeric_limits<su2double>::lowest();
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1974,6 +1978,46 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
...
+    for (auto iPoint = 0ul; iPoint < nPoint; iPoint++) {
+      minP = min(minP, nodes->GetPressure(iPoint));
+      maxP = max(maxP, nodes->GetPressure(iPoint));
+    }
+    END_SU2_OMP_FOR
+
+    SU2_OMP_CRITICAL
+    MinP = min(MinP, minP);
+    MaxP = max(MaxP, maxP);
+    END_SU2_OMP_CRITICAL
```
pcarruscag:
```suggestion
    SU2_OMP_CRITICAL {
      MinP = min(MinP, minP);
      MaxP = max(MaxP, maxP);
    }
    END_SU2_OMP_CRITICAL
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1974,6 +1978,47 @@ void CIncEulerSolver::SetBeta_Parameter(CGeometry *geometry, CSolver **solver_co
...
+
+    BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS
+    {
+      minP = MinP;
+      SU2_MPI::Allreduce(&minP, &MinP, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+      maxP = MaxP;
+      SU2_MPI::Allreduce(&maxP, &MaxP, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+
+      config->SetRangePressure(MinP,MaxP);
+
```
pcarruscag:
```suggestion
```

PR #2520, opened 2025-06-25
Fix periodic parallel preprocessing

### LINKED ISSUE #2518: Segfaults in periodic boundary processing
**Describe the bug**
I frequently get a segfault when running a simple problem with periodic boundaries (inviscid vortex on square mesh). It only happens with some meshes (e.g. the attached `TriAdapt1.su2`, but not `TriInit.su2`). `CPhysicalGeometry::MatchPeriodic()` seems to give the correct number of matches, but then the segfault occurs in `CGeometry::PreprocessPeriodicComms()`:

> ==99060== Invalid write of size 8
==99060==    at 0xD26DC8: CGeometry::PreprocessPeriodicComms(CGeometry*, CConfig*) (CGeometry.cpp:973)
...

(and in 3 other locations). It only occurs when running in parallel.

I've attached the full valgrind outputs (serial and parallel), as well as the meshes and config I used.

**Bug report checklist**

 - OS: Linux (Fedora 42 and Ubuntu 22.04)
 - C++ compiler and version: GCC 15.1.1 and 11.4.0
 - MPI implementation and version: MPICH 4.2.2 and 4.0
 - SU2 Version: 8.2.0 (develop)

[config.zip](https://github.com/user-attachments/files/20871208/config.zip)
[meshes.zip](https://github.com/user-attachments/files/20871207/meshes.zip)
[valgrind_out.zip](https://github.com/user-attachments/files/20871209/valgrind_out.zip)

bmunguia:
## Proposed Changes
Fixes #2518.

Modify the logic in `CGeometry::PreprocessPeriodicComms()` to use unordered sets of point-marker pairs when determining number of points to communicate to each destination rank.

When counting points, the old logic just checked that the previous point was not the same as the current point, and when generating the list of points to be communicated, there was no check in place. In some cases, the old logic would lead to a segfault during preprocessing, such as the old result shown below for the isentropic vortex.

### Before
vortex_parallel_old

### After
vortex_parallel_new

## PR Checklist


bigfooted:
Hi Brian, can you add this testcase to the regression tests as well?

Common/src/geometry/CGeometry.cpp
```diff
@@ -969,14 +976,22 @@ void CGeometry::PreprocessPeriodicComms(CGeometry* geometry, CConfig* config) {
...
-              Local_Marker_PeriodicSend[ii] = static_cast<unsigned long>(iMarker);
-              jj = ii * nPackets;
-              idSend[jj] = geometry->vertex[iMarker][iVertex]->GetDonorPoint();
-              jj++;
-              idSend[jj] = static_cast<unsigned long>(iPeriodic);
-              ii++;
+            if (iRank == destRank) {
+              /*--- Check if we have already added this (point, marker) pair
+               for this destination rank. ---*/
+              auto pointMarkerPair = std::make_pair(iPoint, static_cast<unsigned long>(iMarker));
```
pcarruscag:
```suggestion
              const auto pointMarkerPair = std::make_pair(iPoint, static_cast<unsigned long>(iMarker));
```

Common/src/geometry/CGeometry.cpp
```diff
@@ -969,14 +976,22 @@ void CGeometry::PreprocessPeriodicComms(CGeometry* geometry, CConfig* config) {
...
+                Local_Point_PeriodicSend[ii] = iPoint;
+                Local_Marker_PeriodicSend[ii] = static_cast<unsigned long>(iMarker);
+                jj = ii * nPackets;
+                idSend[jj] = geometry->vertex[iMarker][iVertex]->GetDonorPoint();
+                jj++;
+                idSend[jj] = static_cast<unsigned long>(iPeriodic);
+                ii++;
+
+                /*--- Store the (point, marker) pair in the set for the destination rank. ---*/
+                Points_Send_All[destRank].insert(pointMarkerPair);
```
pcarruscag: You can always call insert and use the result of insert to determine whether this is the first time the pair is found. This saves the double work of calling find and then insert.

pcarruscag (approved):

Issue #2409, opened 2024-12-30
Minor 8.1.0 compilation issues on linux


yuriy-yarosh:
**Describe the bug**

Building `8.1.0` from source code on Archlinux with LLVM and stock (non su2) meson/ninja.
I've installed python 3.13 with asdf and bundled `meson.py` script does not resolve the ninja `PATH` correctly.

While trying to build the code I've stumbled upon couple issues

1. Posix definition for strict `fileno` usage
`externals/cgns/adf/ADF_internals.c` requires `#define _POSIX_SOURCE` before including `stdio.h` for `fileno` implicit function definition to actually be resolved, [over here tbe](https://github.com/CGNS/CGNS/blob/develop/src/adf/ADF_internals.c#L169)

![pic](https://github.com/user-attachments/assets/f840d79a-b5b5-4b6e-8ed9-34949c09e7f7)

2. Missing `strings.h` include in Metis `externals/metis/GKlib/gkregex.c` for `strcasecmp` to actually be resolved, [over here tbe](https://github.com/MITIBMxGraph/METIS-GKlib/blob/main/GKlib/gkregex.c#L92)

![pic](https://github.com/user-attachments/assets/4a1f701a-dc65-47d0-b85c-6f6b0b4e9154)

**Desktop (please complete the following information):**
 - OS: Arch
 - C++ compiler and version: Python 3.13.1 ninja 1.12.1 meson 1.6.1 with asdf
 - MPI implementation and version: arch OpenMPI 5.0.6 
 - SU2 Version: v8.1.0

pcarruscag:
Thanks, let us know when the CGNS and METIS developers give you an update.

yuriy-yarosh:
@pcarruscag not sure if I've opened the issue in the correct metis repo, but CGNS folks had fixed that already and there's a PR due to C API cleanup effort.

pcarruscag:
The METIS change is very simple so I would just open a PR to Prof. Karypis repo and another one to SU2 with the same update, we distribute METIS and ParMETIS with SU2 and the Apache 2 license allows modifying the source.

PR #2531, opened 2025-07-07
Upgrade to CGNS v4.5

bmunguia:
## Proposed Changes
Upgrade external CGNS library to v4.5, fixing a compilation issue on Linux.

## Related Work
#2409: Fixes compilation issue 1 (undefined `_POSIX_SOURCE` in CGNS library)

## PR Checklist

### DIFF
--- a/externals/cgns/README
+++ b/externals/cgns/README
@@ -1,6 +1,6 @@
 **********************************************************************
-https://github.com/CGNS/CGNS/releases, 
-the source files for CGNS v4.2.0 are bundled and build directly into SU2
+https://github.com/CGNS/CGNS/releases,
+the source files for CGNS v4.5.0 are bundled and built directly into SU2
 
 From the CGNS webpage:
 
--- a/externals/cgns/adf/ADF_interface.c
+++ b/externals/cgns/adf/ADF_interface.c
@@ -96,6 +96,9 @@ static char ADF_A_identification[] = "\300\250\243\251ADF Database Version A0201
 /***********************************************************************
     Includes
 ***********************************************************************/
+#ifndef _WIN32
+  #define _POSIX_C_SOURCE 200112L
+#endif
 #include <stdio.h>
 #include <errno.h>
 #include <string.h>
--- a/externals/cgns/adf/ADF_internals.c
+++ b/externals/cgns/adf/ADF_internals.c
@@ -166,6 +166,9 @@ bytes   start   end   description      range / format
 /***********************************************************************
  	Includes
 ***********************************************************************/
+#ifndef _WIN32
+  #define _POSIX_C_SOURCE 200112L
+#endif
 #include <sys/types.h>
 #include <time.h>
 #include <stdio.h>
--- a/externals/cgns/adf/ADF_internals.h
+++ b/externals/cgns/adf/ADF_internals.h
@@ -525,7 +525,7 @@ extern  void    ADFI_increment_array(
             cgulong_t *element_offset,
             int *error_return ) ;
 
-extern  void    ADFI_is_block_in_core() ;
+extern  void    ADFI_is_block_in_core(void) ;
 
 extern  void    ADFI_little_endian_to_cray(
 	    const char from_format,
@@ -591,7 +591,7 @@ extern  void    ADFI_read_data_translated(
             char *data,
             int *error_return ) ;
 
-extern  void    ADFI_read_disk_block() ;
+extern  void    ADFI_read_disk_block(void) ;
 
 extern  void    ADFI_read_disk_pointer_from_disk(
             const unsigned int file_index,
@@ -707,7 +707,7 @@ extern  void    ADFI_write_data_translated(
             const char *data,
             int *error_return ) ;
 
-extern  void    ADFI_write_disk_block() ;
+extern  void    ADFI_write_disk_block(void) ;
 
 extern  void    ADFI_write_disk_pointer_2_disk(
             const unsigned int file_index,
--- a/externals/cgns/adfh/ADFH.c
+++ b/externals/cgns/adfh/ADFH.c
@@ -21,7 +21,9 @@ freely, subject to the following restrictions:
 /*-------------------------------------------------------------------
  * HDF5 interface to ADF
  *-------------------------------------------------------------------*/
-
+#ifndef _WIN32
+  #define _POSIX_C_SOURCE 200112L
+#endif
 #include <stdio.h>
 #include <string.h>
 #include <math.h>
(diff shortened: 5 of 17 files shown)


pcarruscag:
Yep if they require those flags then using the old values we had sounds reasonable

bmunguia:
While I'm at it, I get an error building ninja with my system's python (3.13)
```
ninja executable not found. Building ...
Traceback (most recent call last):
  File "/home/brian/SU2/externals/ninja/configure.py", line 26, in <module>
    import pipes
ModuleNotFoundError: No module named 'pipes'
```
I've been installing in a virtual environment with python 3.12 which builds successfully but gives a deprecation warning
```
ninja executable not found. Building ...
/home/brian/SU2/externals/ninja/configure.py:26: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
  import pipes
```
This seems to have been fixed by ninja-build/ninja#2340. Should I go ahead and update meson/ninja?

pcarruscag:
Go for it, we also changed to c++17 recently, we might make someone unhappy :rofl:

bmunguia:
Some of the regressions are failing after updating meson/ninja even though they all passed after updating CGNS. The tests that failed are

- Tutorials: propeller_variable_load
- Serial Regression: naca0012_geo,  shape_opt_euler_py
- Parallel Regression: fsi2d
- Address Sanitizer Tests: naca0012_geo 

propeller_variable_load and fsi2d both produce out of bounds access errors:

> /usr/include/c++/15/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = double; _Alloc = std::allocator<double>; reference = double&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
/usr/include/c++/15/bits/stl_vector.h:1263: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = double; _Alloc = std::allocator<double>; reference = double&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

The other 2 tests fail running SU2_GEO

pcarruscag (commented):
We should not be seeing asserts in release builds, so something fishy is happening.

bmunguia:
It looks like [meson changed how assertions are controlled in v1.4](https://mesonbuild.com/Release-notes-for-1-4-0.html#ndebug-setting-now-controls-c-stdlib-assertions). If `b_ndebug==false` (default false), it enables debugging assertions:
> For GCC, -D_GLIBCXX_ASSERTIONS=1 is set.
>
> For Clang, -D_GLIBCXX_ASSERTIONS=1 is set to cover libstdc++ usage, and -D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_EXTENSIVE or -D_LIBCPP_ENABLE_ASSERTIONS=1 is used depending on the Clang version.

So now I set `b_ndebug=true` by default in meson.build (which sets `-DNDEBUG`) and undefine `NDEBUG` for debug builds, which should result in the pre-upgrade behavior (and seemed to fix the regressions).

meson.build
```diff
@@ -51,6 +52,13 @@ if build_machine.system() != 'windows'
   endif
 endif
 
+# Handle assertions: default is b_ndebug=true (assertions disabled)
+# but for debug builds, we want to enable assertions
+if get_option('buildtype') == 'debug'
+  # Undefine NDEBUG
+  su2_cpp_args += '-UNDEBUG'
+endif
+
```
pcarruscag, line 61: Are you sure that this is not compiler-dependent? Did you get this info from the meson docs?

bmunguia: It shouldn't be compiler-dependent as-is. I referenced the source code to see why `-D_GLIBCXX_ASSERTIONS` was getting set when the regressions were failing. Assert args are added in `compilers.py` [here](https://github.com/mesonbuild/meson/blob/5a82ea0501736a666ca9cc003ea0774f8219fd65/mesonbuild/compilers/compilers.py#L313)
```
args += compiler.get_assert_args(are_asserts_disabled(target, env), env)
```
where `are_asserts_disabled()` is defined [here](https://github.com/mesonbuild/meson/blob/5a82ea0501736a666ca9cc003ea0774f8219fd65/mesonbuild/compilers/compilers.py#L249-L258) as
```
def are_asserts_disabled(target: 'BuildTarget', env: 'Environment') -> bool:
    """Should debug assertions be disabled

    :param target: a target to check for
    :param env: the environment
    :return: whether to disable assertions or not
    """
    return (env.coredata.get_option_for_target(target, 'b_ndebug') == 'true' or
            (env.coredata.get_option_for_target(target, 'b_ndebug') == 'if-release' and
             env.coredata.get_option_for_target(target, 'buildtype') in {'release', 'plain'}))
```
and `get_assert_args()` is defined in `cpp.py` [here](https://github.com/mesonbuild/meson/blob/5a82ea0501736a666ca9cc003ea0774f8219fd65/mesonbuild/compilers/cpp.py#L314-L331) for clang and [here](https://github.com/mesonbuild/meson/blob/5a82ea0501736a666ca9cc003ea0774f8219fd65/mesonbuild/compilers/cpp.py#L524-L541) for gcc as (the function definitions are the same)
```
    def get_assert_args(self, disable: bool, env: 'Environment') -> T.List[str]:
        if disable:
            return ['-DNDEBUG']

        # Don't inject the macro if the compiler already has it pre-defined.
        for macro in ['_GLIBCXX_ASSERTIONS', '_LIBCPP_HARDENING_MODE', '_LIBCPP_ENABLE_ASSERTIONS']:
            if self.defines.get(macro) is not None:
                return []

        if self.language_stdlib_provider(env) == 'stdc++':
            return ['-D_GLIBCXX_ASSERTIONS=1']
        else:
            if version_compare(self.version, '>=18'):
                return ['-D_LIBCPP_HARDENING_MODE=_LIBCPP_HARDENING_MODE_FAST']
            elif version_compare(self.version, '>=15'):
                return ['-D_LIBCPP_ENABLE_ASSERTIONS=1']

        return []
```
So I added `b_ndebug=true` to the default project options, which makes `are_asserts_disabled()` return `True`, which makes `get_assert_args()` return `[-DNDEBUG]`. If `buildtype=='debug'`, then we just need to unset that flag like I do in the lines you're reviewing.

If you wanted the option to generate a configuration that gets past the `if disable: return ['-DNDEBUG']` and adds the cpp assertions, then I would need to modify the logic and probably add another meson option.

pcarruscag: Cool, thanks for checking, it's good like this.

pcarruscag (approved):

pcarruscag:
Are you working on fixing this @bmunguia? -> https://github.com/su2code/SU2/actions/runs/16277520538

PR #2535, opened 2025-07-11
SU2 Binary grid format

vdweide:
## Proposed Changes
A binary version of the SU2 grid format has been defined, see below, and a very first version of the reader has been implemented. The binary version of this SU2 format enables a much faster reading of the grid file, especially for very large cases. E.g., I am running a case, which consists of 430 million hexahedra (435 million points). When using the ASCII version of the SU2 file it took 45 minutes on 3,000 cores to read the grid. With the new binary format it takes less than 2 seconds. I tried to create a CGNS version of this grid, but I got errors in the CGNS library (I tried multiple versions) and therefore created this solution.

At the moment the format described below is used for the binary grid file (extension .su2b). However, at this stage this can be changed easily. It is a bit less flexible than the ASCII format, but it is still so similar that the same base classes can be used for the actual reading. The current format is as follows, where conn_type = uint32_t for 4 byte types and conn_type = uint64_t for 8 byte types

``int32_t size_conn_type           ``       // Type used for the connectivities. Either 4 or 8 bytes. 
``                                 ``         // This also serves as a check if little or big endian format is used.
``int32_t number_of_zones          ``  // Number of zones. Must be present, in contrast to the ASCII format
``for all zones                    ``       // Loop over all the zones in the grid
``  int32_t zone_id                  ``         // Zone ID. Must be present, in contrast to the ASCII format
``  int32_t n_dim                    ``         // Number of dimensions of the zone
``  conn_type n_elem                 ``   // Number of volume elements of the zone.
``  conn_type offset_elem            ``  // Offsets of the volume element connectivity, size is n_elem+1
``  conn_type elem_conn              ``  // Connectivity of the volume elements. Same as in ASCII format.
``                                 ``            //   Size is offset_elem[nElem].
``  conn_type n_points               ``   // Number of points in the grid
`` double coor + conn_type point ID`` // Coordinates and point ID. Same as in ASCII format.
``                                 ``                // Size n_dim*n_points coordinates and n_points ID's
``  int32_t n_markers                ``  // Number of markers for this zone
``  for all markers                  ``  // Loop over the markers of this zone
``     char[33] marker_name          `` // Name of the marker
``     conn_type n_bound_elem        `` // Number of boundary elements for this marker.
``     conn_type offset_bound_elem   `` // Offsets of the surface connectivity, size is n_bound_elem+1
``     conn_type bound_elem_conn     `` // Connectivity of the boundary elements. Same as in ASCII format. 

The addition of the offset allows for fast navigating through the file, which is beneficial for large number of MPI ranks. The conn_type is also used for the offset a uint32_t can be used for grids of roughly 300 to 400 million points, depending a bit on the element types used. For even bigger grids the uint64_t is necessary.

This is still a draft PR, so there is no support yet for actuator disks. Also the corresponding reading of the FEM part is not implemented yet. This will be done if people think this is a useful addition. Then also a converter from ASCII to the binary format will be made available.

## Related Work
None

## PR Checklist


Common/src/toolboxes/SwapBytes.cpp
```diff
@@ -0,0 +1,53 @@
+/*!
+ * \file SwapBytes.cpp
+ * \brief Function to swap bytes of primitive data types
+ * \author P. Gomes
```
pcarruscag, line 4: I don't remember writing this function, but I'll take it :grin:

Common/src/geometry/meshreader/CSU2BinaryMeshReaderFVM.cpp
```diff
@@ -0,0 +1,64 @@
...
+  ReadPointCoordinates();
+  ReadSurfaceElementConnectivity();
+
+  fclose(mesh_file);
+}
+
+CSU2BinaryMeshReaderFVM::~CSU2BinaryMeshReaderFVM() = default;
+
+void CSU2BinaryMeshReaderFVM::SplitActuatorDiskSurface() {
+  SU2_MPI::Error(string("Not implemented yet"), CURRENT_FUNCTION);
```
pcarruscag:
```suggestion
  SU2_MPI::Error("Not implemented yet", CURRENT_FUNCTION);
```

Common/src/CConfig.cpp
```diff
@@ -629,6 +630,45 @@ unsigned short CConfig::GetnZone(const string& val_mesh_filename, unsigned short
 
     }
 
+    case SU2_BIN: {
+
+      /*--- Check if the mesh file can be opened for binary reading. ---*/
+      FILE *mesh_file = fopen(val_mesh_filename.c_str(), "rb");
+      if ( !mesh_file )
+        SU2_MPI::Error(string("There is no geometry file called ") + val_mesh_filename,
```
pcarruscag: I don't think you need to cast to string explicitly in these places
```suggestion
        SU2_MPI::Error("There is no geometry file called " + val_mesh_filename,
```

Common/include/geometry/meshreader/CSU2BinaryMeshReaderFEM.hpp
```diff
@@ -0,0 +1,66 @@
...
+ public:
+  /*!
+   * \brief Constructor of the CSU2BinaryMeshReaderFEM class.
+   */
+  CSU2BinaryMeshReaderFEM(CConfig* val_config, unsigned short val_iZone, unsigned short val_nZone);
+
+  /*!
+   * \brief Destructor of the CSU2BinaryMeshReaderFEM class.
+   */
+  ~CSU2BinaryMeshReaderFEM(void) override;
```
pcarruscag, line 65: Omit destructors if you don't need them to do something custom.

Common/include/geometry/meshreader/CSU2BinaryMeshReaderBase.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include "../../../include/toolboxes/SwapBytes.hpp"
+
+/*!
+ * \class CSU2BinaryMeshReaderBase
+ * \brief Base class for the reading of a native SU2 binary grid.
+ * \author T. Economon, E. van der Weide
+ */
+class CSU2BinaryMeshReaderBase : public CSU2MeshReaderBase {
+ protected:
+  constexpr static int SU2_STRING_SIZE = 33; /*!< \brief Size of the strings in the SU2 binary mesh file. */
```
pcarruscag: Is there a reason behind the small-ish limit? Can it be relaxed? I remember some of the CRM meshes having quite long names padded with spaces.

Common/src/CConfig.cpp
```diff
@@ -757,6 +797,51 @@ unsigned short CConfig::GetnDim(const string& val_mesh_filename, unsigned short
       break;
     }
 
+    case SU2_BIN: {
+
+      /*--- Check if the mesh file can be opened for binary reading. ---*/
+      FILE *mesh_file = fopen(val_mesh_filename.c_str(), "rb");
+      if ( !mesh_file )
+        SU2_MPI::Error(string("There is no geometry file called ") + val_mesh_filename,
+                              CURRENT_FUNCTION);
```
pcarruscag: Helper function for the common parts between this and GetnZone please

Common/src/geometry/meshreader/CSU2BinaryMeshReaderBase.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+  for (unsigned long i = 0; i < numberOfLocalElements; ++i) {
+    auto ind = offset[i];
+    auto size_this_elem = static_cast<unsigned short>(offset[i + 1] - offset[i]);
+    auto VTK_Type = conn_buff[ind++];
+    const auto nPointsElem = nPointsOfElementType(static_cast<unsigned short>(VTK_Type));
+    if (size_this_elem < (nPointsElem + 2)) SU2_MPI::Error("Not enough items in volume connectivity", CURRENT_FUNCTION);
+
+    for (unsigned short j = 0; j < nPointsElem; ++j, ++ind) connectivity[j] = conn_buff[ind];
+    auto GlobalIndex = conn_buff[ind];
```
pcarruscag, line 334: Is the global index used? I have the feeling we always end up ignoring the file and basing things on the order of elements in the file

pcarruscag (commented):
Looks good.
Would it be a lot of work to handle the writing too? I suppose you have a simple converter but missing all the mpi stuff?

vdweide:
> Looks good. Would it be a lot of work to handle the writing too? I suppose you have a simple converter but missing all the mpi stuff?

Yes, I do have a sequential converter. I think it should be doable to add a write option as well. I'll have a look at it.

Common/src/geometry/meshreader/CSU2BinaryMeshReaderFVM.cpp
vdweide: Done.

Common/src/CConfig.cpp
vdweide: Done.

Common/include/geometry/meshreader/CSU2BinaryMeshReaderBase.hpp
vdweide: Same value as CGNS uses. It can definitely be larger. What size do you suggest? 129 or even 257?

pcarruscag: Maybe 65, that would already be a pretty long line in the config file

Common/src/CConfig.cpp
vdweide: Done

Common/src/geometry/meshreader/CSU2BinaryMeshReaderBase.cpp
vdweide: Not that I know of. We typically assume that the position of the element is the same as the global index. I kept it in for consistency reasons.

Common/include/geometry/meshreader/CSU2BinaryMeshReaderBase.hpp
vdweide: OK, I changed it to 65.

PR #2566, opened 2025-09-08
Add nested preconditioner: FGMRES with BiCGSTAB

eorbay-metu:
## Proposed Changes
This draft PR introduces a nested Krylov solver setup in SU2.
The main contribution is enabling FGMRES to use BiCGSTAB as an inner preconditioner.
The aim is to improve convergence robustness and efficiency for stiff linear systems, especially in high-Re/low-Mach cases.

Changes include:

Modifications to the linear solver routines (CSysSolve.cpp, CSysSolve.hpp), CConfig.cpp and option_structure.hpp

Added new functions to allow BiCGSTAB calls inside FGMRES iterations.

Extended configuration parsing to accept LINEAR_SOLVER= FGMRESandBCGSTAB2

## Related Work
Builds upon existing Krylov-based solvers already implemented in SU2 (FGMRES, BiCGSTAB).

Complements discussions on improving solver performance for stiff CFD problems.

## PR Checklist


eorbay-metu:
@pcarruscag Thank you for the suggestion! I understand your point about wrapping the inner solver into the generic preconditioner interface for broader reuse.  

In my initial implementation I integrated BiCGSTAB directly inside FGMRES rather than as a separate preconditioner function, because my primary goal was to quickly evaluate the nested Krylov approach in SU2 and confirm its effect on convergence. This way I could test the functionality without restructuring the existing preconditioner hierarchy.  

I see the benefit of a cleaner design by wrapping it into the preconditioner interface, and I will look into how to move in that direction. I would be glad to discuss this further during one of the weekly developer meetings.

bigfooted:
Hi, great work, can you also add a small testcase that uses the new function? Looking forward to hearing more about it at the conference.

eorbay-metu:
Hi @bigfooted , thanks for checking in! I'm currently working through the remaining CI issues. The compilation and regression tests are green, but the Code Style workflow is still reporting formatting errors. I’m planning to run clang-format/pre‑commit locally and push a cleaned-up commit to address those shortly.

To show that the new nested FGMRES/BiCGSTAB preconditioner works, I’ve prepared a modified version of an existing RANS/NACA test that uses this nested solver.

Once the formatting is fixed and the new test case is integrated, I’ll take the PR out of draft and open it up for full review. Thanks for your patience and feedback — if you have a particular scenario you’d like me to cover, please let me know!

TestCases/TestCases/.travis.yml
pcarruscag, line 1: I think you copy pasted TestCases into TestCases and now there are a lot of new files

Common/src/linear_algebra/CSysSolve.cpp
```diff
@@ -172,6 +172,112 @@ void CSysSolve<ScalarType>::ModGramSchmidt(bool shared_hsbg, int i, su2matrix<Sc
   w[i + 1] /= nrm;
 }
 
+/*--- Inner Krylov solver for nested FGMRES ---*/
+
+template <class ScalarType>
+void BCGSTABpre_parallel(const CSysVector<ScalarType>& a, CSysVector<ScalarType>& b_in,
+                         const CMatrixVectorProduct<ScalarType>& mat_vec, const CPreconditioner<ScalarType>& precond,
+                         const CConfig* config) {
```
pcarruscag, line 180: So is this the final version you propose?

TestCases/TestCases/.travis.yml
eorbay-metu: Oh, you’re right – that was a mistake on my side. I accidentally copied the TestCases directory into TestCases/, which created a duplicated tree and a lot of unrelated files in this PR.

I’ll clean this up and. I’ll push the updated version accordingly.

Common/src/linear_algebra/CSysSolve.cpp
eorbay-metu: Yes, this is the version I am proposing at this stage.

The idea is to keep the inner solver logic inside `CSysSolve.cpp` and reuse the existing BCGSTAB implementation as an inner linear solver. In the current design, FGMRES behaves as before unless the config explicitly enables nesting, i.e.

- `LINEAR_SOLVER_NESTED = YES`
- `LINEAR_SOLVER_INNER  = BCGSTAB`

In that case, FGMRES calls the inner BCGSTAB helper instead of the usual preconditioner application. I kept the inner solver out of `CPreconditioner.hpp` on purpose, to avoid mixing “preconditioner” and “inner Krylov solver” concepts and to prevent confusion since the inner solver itself uses the existing preconditioner infrastructure. There is already a `BCGSTAB_LinSolver` in `CSysSolve` which returns an `unsigned long`. For the nested case, I introduced `BCGSTABpre_parallel` as a `void` helper, because I wanted a simpler interface that only takes `(A, b, mat_vec, precond, config)` and updates `x` in-place, without changing the existing outer solver interfaces or return types. The nested FGMRES code only needs the updated solution vector, so the helper keeps the call site straightforward. If you prefer, I can refactor this further to reduce duplication.

Right now, the inner solver uses a fixed tolerance (`1e-5`) hard-coded in the implementation; there are no separate config options yet for the inner solver (e.g. inner tolerance / inner max iterations). My plan is to expose these as dedicated config options in a follow-up patch, once the overall nested FGMRES + BiCGSTAB structure is agreed upon.

pcarruscag:
It will be a squash merge because of the intermediate commits of many files.

PR #2569, opened 2025-09-11
Adding temperature limits for incompressible solver

Cristopher-Morales:
## Proposed Changes

The aim of this pull request is to add temperature limits to prevent unphysical values for incompressible flows. Controlling temperature values is especially important for multicomponent and reacting flows, where low and high values of temperature can lead to unbounded density values, which might result in unphysical species mass fractions values (outside range 0.0-1.0).

Test H2-Air mixture adapted from test case rectangle passive transport validation:

top: develop
bottom: using temperature limits 

**Species mass fraction H2**:
  
image
 
image

**Density:**

image

image

## PR Checklist


Cristopher-Morales:
> Hi Cristopher, good to add this in! I would not add the underrelaxation now if it is still unclear how to make it work in a robust way. But good to also keep that on the to-do list (give it to a student project maybe). Why are so many regression tests affected, do they all hit the boundary of the temperature limits? Does it make sense to choose as lower limit not zero but a positive value, maybe 250K or something?

Hi Nijso,
Thanks for your comment!
Yes, that could also be an option to propose  as a student project to investigate how to mae the automatic underrelation approach more robust, which could eventually be very useful for the species solver as well.
The regression test that are failing are due to enable the automatic underrelaxation where the option MAX_UPDATE_FLOW has a default value 0.2. I think if I add to the failing test cases the option MAX_UPDATE_FLOW=1.0, they should not fail. When I disable the automatic underrelaxation, the regression tests do not fail.
It does make sense to choose a lower limit 250K, in the figure of species and density that I obtained using **relaxation with temperature limits (bottom figure)**, the temperature limits are 250,400.

Cristopher-Morales:
> We don't know how many people out there are using the incompressible solver for heat transfer, so I would advocate for default values that keep the default behavior. Thanks for checking the underrelaxation, if most of the improvements still come from the limits I think it's fine to leave it for a project.

Thanks for the comment.

Yes, the most of the improvements come from the limits. I am going to disable the ComputeUnderRelaxationFactor in this pull request and let its implementation for a future project.

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -82,6 +82,8 @@ class CIncEulerVariable : public CFlowVariable {
    * \param[in] nvar - Number of variables of the problem.
    * \param[in] config - Definition of the particular problem.
    */
+  su2double Temperature_Limits[2];
+
```
pcarruscag: Move to protected variables.
```suggestion
  su2double TemperatureLimits[2];

```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -116,9 +118,9 @@ class CIncEulerVariable : public CFlowVariable {
    * \brief Set the value of the temperature for incompressible flows with energy equation.
    * \param[in] iPoint - Point index.
    */
-  inline bool SetTemperature(unsigned long iPoint, su2double val_temperature) final {
+  inline bool SetTemperature(unsigned long iPoint, su2double val_temperature, su2double *val_temp_limits) {
```
```suggestion
  inline bool SetTemperature(unsigned long iPoint, su2double val_temperature, const su2double *val_temp_limits) {
```

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -37,6 +37,8 @@ CIncNSVariable::CIncNSVariable(su2double pressure, const su2double *velocity, su
   Tau_Wall.resize(nPoint) = su2double(-1.0);
   DES_LengthScale.resize(nPoint) = su2double(0.0);
   Max_Lambda_Visc.resize(nPoint);
+  Temperature_Limits[0] = config->GetTemperature_Limits(0);
+  Temperature_Limits[1] = config->GetTemperature_Limits(1);
```
pcarruscag: Not needed, the constructor of CIncEulerVariable already does it.
```suggestion

```

Common/include/CConfig.hpp
```diff
@@ -2155,6 +2155,12 @@ class CConfig {
    */
   su2double GetInc_Temperature_Init(void) const { return Inc_Temperature_Init; }
 
+  /*!
+   * \brief Get Temperature limits for incompressible flows.
+   * \return Temperature limits minimum and maximum values.
+   */
+  const su2double GetTemperature_Limits(int iVar) const { return TemperatureLimits[iVar]; }
```
```suggestion
  su2double GetTemperatureLimits(int iVar) const { return TemperatureLimits[iVar]; }
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
Cristopher-Morales: I moved them to protected variables, Thanks!!

bigfooted:
Hi Cristopher, can you add this testcase as well?

Common/src/CConfig.cpp
```diff
@@ -1368,6 +1368,9 @@ void CConfig::SetConfig_Options() {
   addEnumOption("INC_DENSITY_MODEL", Kind_DensityModel, DensityModel_Map, INC_DENSITYMODEL::CONSTANT);
     /*!\brief ENERGY_EQUATION \n DESCRIPTION: Solve the energy equation in the incompressible flow solver. \ingroup Config*/
   addBoolOption("INC_ENERGY_EQUATION", Energy_Equation, false);
+  /*!\brief TEMPERATURE_LIMITS \n DESCRIPTION: Temperature limits for incompressible flows (0.0, 5000 K by default) \ingroup Config*/
+  TemperatureLimits[0] = 0.0; TemperatureLimits[1] = 5000.0;
```
pcarruscag, line 1372: It's possible to use negative temperatures with the incompressible solver right?
```suggestion
  /*!\brief TEMPERATURE_LIMITS \n DESCRIPTION: Temperature limits for incompressible flows (0.0, 5000 K by default) \ingroup Config*/
  TemperatureLimits[0] = -1e9; TemperatureLimits[1] = 1e9;
```

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -76,7 +78,7 @@ bool CIncEulerVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel
   /*--- Set the value of the temperature directly ---*/
 
   su2double Temperature = Solution(iPoint, nDim+1);
-  const auto check_temp = SetTemperature(iPoint, Temperature);
+  const auto check_temp = SetTemperature(iPoint, Temperature, TemperatureLimits);
```
pcarruscag, line 81: Do we get the warnings about the number of unphysical points when the bounds are active?

Common/src/CConfig.cpp
Cristopher-Morales: Hi!!

I have a doubt about this comment, specifically about having negatives temperature. In the config.cpp file I see this

Cristopher-Morales: Hi!!

I have a doubt about this comment. In the config.cpp file I see the following check:

```
/*--- Incompressible solver currently limited to SI units. ---*/

  if ((Kind_Solver == MAIN_SOLVER::INC_EULER || Kind_Solver == MAIN_SOLVER::INC_NAVIER_STOKES || Kind_Solver == MAIN_SOLVER::INC_RANS) && (SystemMeasurements == US)) {
    SU2_MPI::Error("Must use SI units for incompressible solver.", CURRENT_FUNCTION);
  }
```
So the incompressible solver can only work with SI units, and the temperature is in Kelvin in the SI units, so If I am not mistaken, the temperature cannot be negative in this units system. Before in SetTemperature function, there was a check if the temperature was lower than zero, and I think it was because of the reason explained before. Please let me know what you think about this.
Thank you so much in advance!!!!.

SU2_CFD/src/variables/CIncEulerVariable.cpp
Cristopher-Morales: Hi!!

I think we do not get the warnings, should we print a message when the temperature goes outside the bounds?
Thank you so much in advance!!!

Common/src/CConfig.cpp
pcarruscag: Alright sounds good

SU2_CFD/src/variables/CIncEulerVariable.cpp
pcarruscag: I think the return of this function is used by the "non physical counter" maybe you can try a case where you set the bounds tightly. The counter is printed either at solver exit or when we write solution files.

Cristopher-Morales: Thanks! I will try a case with tighter limits to check it.

PR #2581, opened 2025-09-29
Newton Krylov improvements

pcarruscag:
## Proposed Changes
Described in the config template and the Oneram M6 with NK example.
It was also possible to improve performance by not computing the gradients of density and enthalpy for the Roe scheme.

## PR Checklist

### DIFF
--- a/SU2_CFD/src/integration/CNewtonIntegration.cpp
+++ b/SU2_CFD/src/integration/CNewtonIntegration.cpp
@@ -277,9 +277,10 @@ void CNewtonIntegration::MultiGrid_Iteration(CGeometry ****geometry_, CSolver **
     solvers[FLOW_SOL]->SetResLinSolver(eps);
 
     if (!startupPeriod && autoRelaxation) {
-      if (eps > config->GetLinear_Solver_Error()) {
+      const su2double adaptTol = config->GetCFL_Adapt() ? config->GetCFL_AdaptParam(4) : 0;
+      if (eps > fmax(config->GetLinear_Solver_Error(), adaptTol)) {
         nkRelaxation *= 0.9;
-      } else if (iter < config->GetLinear_Solver_Iter()) {
+      } else if (eps < 0.9 * fmax(config->GetLinear_Solver_Error(), adaptTol)) {
         nkRelaxation = fmin(nkRelaxation * 1.05, 1);
       }
     }


pcarruscag:
I need to re run those cases. When there are no major convergence issues it's better to have high limit in the adaptive CFL and this relaxation in "auto mode".

rois1995:
I have tried comparing the NK implementation on version 8.1, 8.3, and 8.3 on this branch. The comparison is made on the Level B mesh of the Pointwise family (1.r.09). Version 8.1 is included since the implicit discretization of the diffusion terms was not implemented yet. On the left figure, the functional is plotted against the number of iterations, whereas on the right figure, it is plotted against the computational time (cumulative sum of the time-per-iteration).

Density residual:
rms_Rho_


$\tilde{\nu}$ residual:
rms_nu_

Pitching moment coefficient:
CMy


The NK options for the SU2 versions 8.1.0 and 8.3.0 develop are the following:
NEWTON_KRYLOV_IPARAM= (10, 5, 2) 
NEWTON_KRYLOV_DPARAM= (1.0, 0.01, -6, 1e-5) 

The NK options for the SU2 version 8.3.0 on this branch are the following:
NEWTON_KRYLOV_IPARAM= (200, 0, 0)
NEWTON_KRYLOV_DPARAM= (0, 0.0, 0, 1e-5, -1)

Other options common to all three simulations:
NUM_METHOD_GRAD= GREEN_GAUSS
LINEAR_SOLVER= RESTARTED_FGMRES
LINEAR_SOLVER_PREC= ILU
LINEAR_SOLVER_ERROR= 1e-2
LINEAR_SOLVER_ITER= 60
CFL_ADAPT= YES
CFL_NUMBER= 10.0
CFL_REDUCTION_TURB= 1.0
CFL_ADAPT_PARAM= ( 0.1, 1.1, 10.0, 100.0, 0.01)

The simulation starts for 500 iterations at constant CFL = 1 without NK, then I activate NK with the settings above.

The computational time should not be that different, since the number of cores and the machine on which they are running is the same (at least the first 500 iterations should be the same). Maybe it is related to the fact that the simulation in yellow has access to more cache.

Nevertheless, the simulation with the changes here proposed seems to reach low residual levels much faster, although it then plateaus on the same levels as the standard implementation, which is kind of strange.

The implicit discretization of the diffusion term (red vs blue curve) does not seem to have much influence.

Should I try the same NK settings used for this branch on the standard implementation?

pcarruscag:
I think that case is not ideal to compare NK settings because the CFL eventually settles at the minimum value, which is very small (10). @Bot-Enigma-0 saw the same behavior. We have to figure out where and why the CFL becomes so limited.

pcarruscag:
In general to make NK as cost effective as possible we need loosely converged linear systems.
This seems a lot, but maybe it's necessary
LINEAR_SOLVER_ERROR= 1e-2
LINEAR_SOLVER_ITER= 60

You looked into this case a lot more than me in recent times, did you try other linear solver settings?
Sometimes it's also helpful to use a higher linear tolerance in the CFL_ADAPT so that the CFL doesn't become limited by the linear solver, e.g. CFL_ADAPT_PARAM= ( 0.1, 1.1, 10.0, 100.0, **0.1**)

pcarruscag:
It would be nice if https://github.com/su2code/SU2/pull/2570 could point out the source of the relatively low convergence rate, at least the coefficients are converged by 2k iterations (I know it gets a lot slower as the meshes get larger).

rois1995:
> I tried it on the first HLCRM-WBHV mesh, with these settings:
> 
> ```
> NUM_METHOD_GRAD= GREEN_GAUSS
> CFL_NUMBER= 10.0
> CFL_ADAPT= YES
> CFL_ADAPT_PARAM= ( 1.0, 1.05, 10.0, 50.0 )
> %
> 
> % -------------------- FLOW NUMERICAL METHOD DEFINITION -----------------------%
> %
> CONV_NUM_METHOD_FLOW= ROE
> ENTROPY_FIX_COEFF= 0.001
> TIME_DISCRE_FLOW= EULER_IMPLICIT
> %
> % ----------- SLOPE LIMITER AND DISSIPATION SENSOR DEFINITION -----------------%
> %
> MUSCL_FLOW= YES
> SLOPE_LIMITER_FLOW= VAN_ALBADA_EDGE
> MUSCL_TURB= NO
> SLOPE_LIMITER_TURB= VENKATAKRISHNAN_WANG
> VENKAT_LIMITER_COEFF= 0.1
> %
> % -------------------- TURBULENT NUMERICAL METHOD DEFINITION ------------------%
> %
> % Convective numerical method (SCALAR_UPWIND, BOUNDED_SCALAR)
> CONV_NUM_METHOD_TURB= SCALAR_UPWIND
> TIME_DISCRE_TURB= EULER_IMPLICIT
> CFL_REDUCTION_TURB= 1
> %
> % ------------------------ LINEAR SOLVER DEFINITION ---------------------------%
> %
> LINEAR_SOLVER= FGMRES
> LINEAR_SOLVER_PREC= ILU
> LINEAR_SOLVER_ERROR= 0.2
> LINEAR_SOLVER_ITER= 12
> 
> NEWTON_KRYLOV= YES
> NEWTON_KRYLOV_IPARAM= (400, 0, 1)
> NEWTON_KRYLOV_DPARAM= (0.0, 0.0, 0.0, 1e-4, -1)
> ```
> 
> The convergence looks like this image

Isn't 1e-3 the standard value for the CFL increase tolerance? Plus, what was your rationale behind these settings?

TestCases/rans/oneram6/turb_ONERAM6_nk.cfg
```diff
@@ -18,22 +18,29 @@ NEWTON_KRYLOV= YES
...
+% can be solved.
+% If "n0">0, and RESTART_SOL=NO, the NK relaxation is applied to the residual (not
+% just to the NK system) to ramp from MUSCL=OFF to ON during "n0" iterations.
+NEWTON_KRYLOV_IPARAM= (200, 0, 1) % n0, np, ft
+NEWTON_KRYLOV_DPARAM= (0, 0, 0, 1e-5, -1) % r0, tp, rf, e, rnk
 
 CFL_ADAPT= YES % it's needed
 CFL_NUMBER= 10
-CFL_ADAPT_PARAM= ( 0.8, 1.1, 5, 1000 ) % no point using NK with low CFL values
+CFL_ADAPT_PARAM= ( 1, 1.05, 10, 400 ) % no point using NK with low CFL values
```
bigfooted, line 43: so when the linear solver convergence stalls, we don't reduce cfl but keep it at the current level, and that is sufficient? Often/in general?

pcarruscag:
1e-3 is the default, but that setting only matters if it's larger than the linear solver tolerance option.
The motivation there was that sometimes you want to keep the linear solver going to a fixed number of iterations (low tolerance), but base the CFL adaptation on a different target.
When you are aiming for a high tolerance (to keep cost low) the convergence can be dominated by a few bad eigenvalues of the Jacobian, and so the linear solver hits the target tolerance quickly, but it doesn't do any smoothing on the rest of the spectrum.
Some solvers are more sensitive than others, BiCGSTAB tends to produce terrible solutions at high tolerance, compared to GMRES.

The rationale is getting away with the cheapest linear solve you can while getting a reasonable convergence rate, between approximating the Jacobian (or Jacobian-vector product in the NK case) by making it less second order and/or limiting CFL, and only converging the linear system a little.

I know this is hand-wavy, but it's a very non-linear problem... I do know the "optimum" solution time is not going to be below CFL of 25 or above ~15 linear iterations. If we need that kind of settings for some cases for stability we need to understand what's responsible for it.

TestCases/rans/oneram6/turb_ONERAM6_nk.cfg
pcarruscag: Stalling resets back to the minimum even if the down factor is 1.
The downfactor is annoying for many problems because of the reduction we do based on the residuals oscillating. However, Rutvik mentioned this was still saving some of the cases from diverging...

pcarruscag:
For completeness, something I never got to try were subspace recycling techniques, each linear system starts from scratch, if we can carry some information from the previous iterations, we can potentially get to a higher CFL at the same cost.
https://personal.math.vt.edu/sturler/publications/SISC_KrylovRecycling_2006.pdf

The other known challenge of Krylov methods is scaling, with more unknowns, you also need larger subspaces because there are more "slow modes". In theory, the solution for that is a preconditioner that can take care of those modes (multigrid).

rois1995:
> I tried it on the first HLCRM-WBHV mesh, with these settings:
> 
> ```
> NUM_METHOD_GRAD= GREEN_GAUSS
> CFL_NUMBER= 10.0
> CFL_ADAPT= YES
> CFL_ADAPT_PARAM= ( 1.0, 1.05, 10.0, 50.0 )
> %
> 
> % -------------------- FLOW NUMERICAL METHOD DEFINITION -----------------------%
> %
> CONV_NUM_METHOD_FLOW= ROE
> ENTROPY_FIX_COEFF= 0.001
> TIME_DISCRE_FLOW= EULER_IMPLICIT
> %
> % ----------- SLOPE LIMITER AND DISSIPATION SENSOR DEFINITION -----------------%
> %
> MUSCL_FLOW= YES
> SLOPE_LIMITER_FLOW= VAN_ALBADA_EDGE
> MUSCL_TURB= NO
> SLOPE_LIMITER_TURB= VENKATAKRISHNAN_WANG
> VENKAT_LIMITER_COEFF= 0.1
> %
> % -------------------- TURBULENT NUMERICAL METHOD DEFINITION ------------------%
> %
> % Convective numerical method (SCALAR_UPWIND, BOUNDED_SCALAR)
> CONV_NUM_METHOD_TURB= SCALAR_UPWIND
> TIME_DISCRE_TURB= EULER_IMPLICIT
> CFL_REDUCTION_TURB= 1
> %
> % ------------------------ LINEAR SOLVER DEFINITION ---------------------------%
> %
> LINEAR_SOLVER= FGMRES
> LINEAR_SOLVER_PREC= ILU
> LINEAR_SOLVER_ERROR= 0.2
> LINEAR_SOLVER_ITER= 12
> 
> NEWTON_KRYLOV= YES
> NEWTON_KRYLOV_IPARAM= (400, 0, 1)
> NEWTON_KRYLOV_DPARAM= (0.0, 0.0, 0.0, 1e-4, -1)
> ```
> 
> The convergence looks like this image

With these settings I am going towards a completely different solution than the other ones, although it is still far from being converged.

RMS density
NKComp_Settings_rms_Rho_

Pitching coefficient
NKComp_Settings_CMy

Linear solver residual
NKComp_Settings_LinSolRes

Linear solver turbulence residual
NKComp_Settings_LinSolResTurb

pcarruscag:
Yeah when the convergence is all spiky like that it's usually not a good sign, in any case I think I gave good-enough disclaimers :sweat_smile: 
> When there are no major convergence issues it's better to have high limit in the adaptive CFL and this relaxation in "auto mode".
> The rationale is getting away with the cheapest linear solve you can while getting a reasonable convergence rate

If tighter tolerances are necessary to achieve convergence then we can try to understand why
> When you are aiming for a high tolerance (to keep cost low) the convergence can be dominated by a few bad eigenvalues of the Jacobian, and so the linear solver hits the target tolerance quickly, but it doesn't do any smoothing on the rest of the spectrum.

Some of the NK parameters allow ramping the linear solver tolerance (third IPARAM and third DPARAM).
In any case, I'm not changing the default behavior here, but some cases can converge faster and we have one more degree of freedom to tune besides CFL and linear solver.
So can I get an approval? :smile:

PR #2607, opened 2025-11-01
Fix Linux arm64 regression workflow

bmunguia:
## Proposed Changes
The workflow "[Regression tests running on Linux ARM64](https://github.com/su2code/SU2/actions/workflows/regression-arm64.yml)" seems to have always timed out after 24 hours of looking for a runner, and has attempted to run on every merge into `develop`. 

This PR:
- Fixes the runner name according to [the available runners for public repos](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/)
- Adds a `workflow_dispatch` trigger to allow manually triggering on either of the available runners
- Updates the test values for aarch64 regressions
- Adds a `TestCase.tol_aarch64` in the regression Python script for cases where a larger tolerance is needed than in the x86_64 regressions

## PR Checklist

### DIFF
--- a/.github/workflows/regression-arm64.yml
+++ b/.github/workflows/regression-arm64.yml
@@ -3,21 +3,11 @@ name: Regression tests running on Linux ARM64
 on:
   push:
     branches:
-      - 'fix_regression_arm'
-  workflow_dispatch:
-    inputs:
-      runner:
-        description: 'ARM runner to use'
-        default: 'ubuntu-24.04-arm'
-        options:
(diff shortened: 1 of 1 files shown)


pcarruscag (commented):
This was using private runners that someone was donating, I was not able to get in touch with the original developer to fix it. So, thanks for the fix.
If there are standard github runners now we can keep it on merge with develop.

PR #2633, opened 2025-11-30
fix: Replace bare except clauses with specific exception types in Python utilities

shbhmexe:
Fixes remaining bare except clauses to improve PEP 8 compliance and code robustness:

- **SU2/__init__.py**: Use `ImportError` for optional readline import.
- **SU2/io/config.py**: Use `Exception` for config file error handling.
- **SU2/util/ordered_bunch.py**: Use `Exception` and [(KeyError, TypeError)](cci:1://file:///s:/Gsoc%20organisation/SU2/SU2_PY/topology_optimization.py:335:4-345:59).
- **SU2/util/bunch.py**: Use `Exception` and `(KeyError, TypeError)`.

### DIFF
--- a/SU2_PY/SU2/__init__.py
+++ b/SU2_PY/SU2/__init__.py
@@ -28,5 +28,5 @@ try:
         readline.parse_and_bind("bind ^I rl_complete")
     else:
         readline.parse_and_bind("tab: complete")
-except ImportError:
+except Exception:
     pass  # readline is optional, continue without it


SU2_PY/SU2/__init__.py
```diff
@@ -28,5 +28,5 @@ class DivergenceFailure(EvaluationFailure):
         readline.parse_and_bind("bind ^I rl_complete")
     else:
         readline.parse_and_bind("tab: complete")
-except:
-    pass
+except ImportError:
+    pass  # readline is optional, continue without it
```
pcarruscag: This change looks incorrect, what if there is an exception in parse_and_bind?

shbhmexe: Thanks for pointing this out. You're right  parse_and_bind() may raise exceptions other than ImportError. I'll update the code to catch those as well while still avoiding a bare except. Appreciate the feedback

pcarruscag:
Maintaining SU2 is not a full-time job for any of us. I don't need daily reminders to merge 7 line changes.

shbhmexe:
@pcarruscag  Got it, thanks for the clarification. I understand maintaining SU2 isn’t a full-time job and I’ll avoid sending reminders. Next time, I’ll focus on more substantial/core fixes when I get . Thanks for reviewing .

PR #2639, opened 2025-12-07
MUSCL Ramp options

joshkellyjak:
## Proposed Changes
Added abilitiy to ramp MUSCL scheme from first to second order across specified number of iterations. Two methods are implemented, ITERATION and SMOOTH_FUNCTION. ITERATION varies based on the current iteration from starting to end iteration, SMOOTH_FUNCTION enables a smooth S-Curve according to a cosine function. The option for power controls how aggresively the ramp is applied.

For ITERATION:
 Power < 1 results in quicker ramp up early then slower as it approaches final value
 Power = 1 results in linear (ITERATION) or cosine (SMOOTH_FUNCTION) ramp
 Power > 1 results in slow initial ramping and then faster towards end

## Related Work
N/A

## PR Checklist


bigfooted:
Cool, do you have a testcase as well? I guess ramping is always good for convergence when you want second order MUSCL?

joshkellyjak:
> Cool, do you have a testcase as well? I guess ramping is always good for convergence when you want second order MUSCL?

Not something coarse enough to be uploaded to GitHub but I will adapt one of the available testcases for this purposes. I mainly implemented this for running cases in an optimisation loop without having to implement some method to restart if you want to go from first to second order Roe scheme. In my experience, if your case can handle second order from the start it is already going to converge pretty well but ramping may improve the the convergence - as long as you don't set too large a ramp. At the minute my determination of the coefficients for the ramp are more of an art rather than a science.

Common/include/CConfig.hpp
```diff
@@ -5189,6 +5201,37 @@ class CConfig {
...
+   * \brief Get MUSCL ramp power.
+   * \return Ramp MUSCL power
+  */
+  su2double GetMUSCLRampPower(void) const { return RampMUSCLPower; }
+
+  /*!
+   * \brief Get MUSCL ramp kind.
+   * \return Ramp MUSCL kind
+  */
+  MUSCL_RAMP_TYPE GetKind_MUSCLRamp(void) const { return Kind_MUSCLRamp; }
```
pcarruscag: With _ this one and not the others?
```suggestion
  MUSCL_RAMP_TYPE GetKindMUSCLRamp(void) const { return Kind_MUSCLRamp; }
```
Can you return the ramp options as a struct please?

Common/src/CConfig.cpp
```diff
@@ -1986,6 +1986,18 @@ void CConfig::SetConfig_Options() {
...
   /*!\brief MUSCL_KAPPA_FLOW \n DESCRIPTION: Blending coefficient for the U-MUSCL scheme \ingroup Config*/
   addDoubleOption("MUSCL_KAPPA_FLOW", MUSCL_Kappa_Flow, 0.0);
+  /*!\brief RAMP_MUSCL \n DESCRIPTION: Enable ramping of the MUSCL scheme from 1st to 2nd order using specified method*/
+  addBoolOption("RAMP_MUSCL", RampMUSCL, false);
+  /*! brief RAMP_OUTLET_COEFF \n DESCRIPTION: the 1st coeff is the ramp start iteration,
+   * the 2nd coeff is the iteration update frequenct, 3rd coeff is the total number of iterations */
+  unsigned short nMUSCLCoeffs = 3;
+  rampMUSCLCoeff = new unsigned long [nMUSCLCoeffs];
+  rampMUSCLCoeff[0] = 0.0; rampMUSCLCoeff[1] = 1.0; rampMUSCLCoeff[2] = 500.0;
+  addULongListOption("RAMP_MUSCL_COEFF", nMUSCLCoeffs, rampMUSCLCoeff);
```
pcarruscag: Use the option for fixed-size arrays

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -251,8 +251,8 @@ void CScalarSolver<VariableType>::Upwind_Residual(CGeometry* geometry, CSolver**
           for (auto iVar = 0u; iVar < nVar; iVar++) {
             const su2double U_ij = Scalar_j[iVar] - Scalar_i[iVar];
 
-            su2double Project_Grad_i = MUSCL_Reconstruction(Gradient_i[iVar], Vector_ij, U_ij, kappa);
-            su2double Project_Grad_j = MUSCL_Reconstruction(Gradient_j[iVar], Vector_ij, U_ij, kappa);
+            su2double Project_Grad_i = config->GetMUSCLRampValue() * MUSCL_Reconstruction(Gradient_i[iVar], Vector_ij, U_ij, kappa);
+            su2double Project_Grad_j = config->GetMUSCLRampValue() * MUSCL_Reconstruction(Gradient_j[iVar], Vector_ij, U_ij, kappa);
```
pcarruscag: It's more consistent to make this an argument of MUSCL_Reconstruction, and also safer from the point of view of not forgetting a solver.
You need to change the SIMD numerics too.

Common/include/CConfig.hpp
joshkellyjak: Must've been before and after cup of tea. Just to clarify, how do mean return ramp options as a struct? As in similar to this

`struct CRBFParam {
    bool DataReduction;               /*!< \brief Determines use of data reduction methods for RBF mesh deformation. */
    su2double GreedyTolerance;        /*!< \brief Tolerance used in the greedy data reduction for RBF mesh deformation. */
  }`
  
  including ramp bool and coeffs in one?

Common/src/CConfig.cpp
joshkellyjak: Apparently we are missing one for u longs, added in latest commit

TestCases/hybrid_regression.py
```diff
@@ -62,6 +62,14 @@ def main():
     naca0012.test_vals = [-4.766168, -4.287699, 0.326688, 0.022661]
     test_list.append(naca0012)
 
+    # NACA0012 w/ MUSCL ramp
+    naca0012_MUSCL_ramp           = TestCase('naca0012_MUSCL_ramp')
+    naca0012_MUSCL_ramp.cfg_dir   = "euler/naca0012"
+    naca0012_MUSCL_ramp.cfg_file  = "inv_NACA0012_Roe_MUSCL_ramp.cfg"
+    naca0012_MUSCL_ramp.test_iter = 30
+    naca0012_MUSCL_ramp.test_vals = [-4.766168, -4.287699, 0.326688, 0.022661]
+    test_list.append(naca0012_MUSCL_ramp)
```
pcarruscag: Just modify some of the existing cases please, especially if there was one for which this was particularly helpful.

Common/include/CConfig.hpp
pcarruscag: Yep something like that. Just for the properties of the ramp, the value of the ramp can still be its own function.

TestCases/hybrid_regression.py
joshkellyjak: There probably isn't a suitable testcase for this, I have used this method for large turbomachinery cases so I updated the Axial_stage2D case.

joshkellyjak:
One thing I have noticed is the CFL should be ramped down during this MUSCL ramp procedure, and then reincreased after it is finished. There can be some convergence issues if not. Here I ramp the MUSCL scheme at iteration 5000 for 2000 iterations and retain a reasonably high CFL for the case (radial turbine). There is some divergence followed by stalled residuals if I keep the CFL high throughout the ramp. Will add some functionality to do this at some point this week.

image

PR #2641, opened 2025-12-10
Implementing variable density for unsteady incompressible flow

tkiymaz:
## Proposed Changes
Implements variable density treatment for unsteady incompressible flow simulations. Currently, SU2 uses constant density in transient simulations, which is inaccurate for combustion cases where density varies significantly due to heat release and species composition changes. This contribution enables proper density updates during time-stepping for flamelet-based combustion modeling. For now, only 1st order time marching is implemented

## Related Work
Related to incompressible flow solver and flamelet combustion modeling. No specific issue linked yet.

## PR Checklist


Cristopher-Morales:
> @Cristopher-Morales can you review this pr please?

Hi @pcarruscag ! 

Yes, I can help reviewing this PR.

Please let me know if you need something else from my side

bigfooted:
@Cristopher-Morales I guess the changes to the preconditioner can be removed from this PR since we now have your implementation

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -70,6 +68,8 @@ class CVariable {
   MatrixType Solution_time_n;    /*!< \brief Solution of the problem at time n for dual-time stepping technique. */
   MatrixType Solution_time_n1;   /*!< \brief Solution of the problem at time n-1 for dual-time stepping technique. */
   VectorType Delta_Time;         /*!< \brief Time step. */
+  VectorType Density_unsteady;
+  VectorType Density_time_n;         /*!< \brief density at time n for dual-time stepping technique. */
```
Cristopher-Morales:
```suggestion
  VectorType Density_unsteady;       /*!< \brief density for unsteady flows. */
  VectorType Density_time_n;         /*!< \brief density at time n for dual-time stepping technique. */
```
Hi Berk!! I think you need to add a documentation here. Like the one suggested

SU2_CFD/include/output/COutput.hpp
```diff
@@ -302,7 +300,7 @@ class COutput {
   unsigned short nRequestedVolumeFields;
 
   /*! \brief Minimum required volume fields for restart file. */
-  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY"};
+  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY","DENSITY_TIME_N"};
```
Cristopher-Morales:
```suggestion
  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY", "DENSITY_TIME_N"};
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -101,7 +99,19 @@ class CIncEulerVariable : public CFlowVariable {
     Primitive(iPoint, indices.Density()) = val_density;
     return val_density <= 0.0;
   }
+  
+  inline void Set_Density_time_n(unsigned long iPoint, su2double val) {
+    Density_time_n[iPoint] = val;
+  }
```
Cristopher-Morales: Documentation must be added, similar to the function above SetDensity, Same for the two functions below Set_Density_unsteady and GetDensity_time_n

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -101,7 +99,19 @@ class CIncEulerVariable : public CFlowVariable {
...
 
+  inline void Set_Density_unsteady(unsigned long iPoint, su2double val) {
+    Density_unsteady[iPoint] = val;
+  }
```
Cristopher-Morales:
```suggestion
  inline void SetDensity_Unsteady(unsigned long iPoint, su2double val) {
    Density_unsteady[iPoint] = val;
  }
```
to keep consistency with the other function names, like **S**et**V**elocity_**O**ld

SU2_CFD/include/variables/CEulerVariable.hpp
```diff
@@ -189,6 +184,14 @@ class CEulerVariable : public CFlowVariable {
...
   }
 
+  inline void Set_Density_time_n(unsigned long iPoint, su2double val) {
+    Density_time_n[iPoint] = val;
+  }
+
+  inline void Set_Density_unsteady(unsigned long iPoint, su2double val) {
+    Density_unsteady[iPoint] = val;
+  }
+
```
Cristopher-Morales:
```suggestion
  inline void SetDensity_Time_n(unsigned long iPoint, su2double val_density_time_n) {
    Density_time_n[iPoint] = val_density_time_n;
  }

  inline void SetDensity_Unsteady(unsigned long iPoint, su2double val_density_unsteady) {
    Density_unsteady[iPoint] = val_density_unsteady;
  }

```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -101,7 +99,19 @@ class CIncEulerVariable : public CFlowVariable {
...
+  
+  inline su2double GetDensity_time_n(unsigned long iPoint) const {
+    return Density_time_n[iPoint];
+  }
+
```
Cristopher-Morales:
```suggestion
  
  inline su2double GetDensity_Time_n(unsigned long iPoint) const {
    return Density_time_n[iPoint];
  }
  
```

documentation must be added

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2753,13 +2753,15 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
 
       /*--- Access the density and Cp at this node (constant for now). ---*/
 
+      Density_time_n = nodes->GetDensity_time_n(iPoint);
+      Density_unsteady = nodes->GetDensity_unsteady(iPoint);
```
Cristopher-Morales:
```suggestion
```
Hi Berk, 

One of the regression tests fails because the Density_unsteady seems that is not used in this routine. You could check if you can remove it in this routine.

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -55,14 +58,18 @@ CVariable::CVariable(unsigned long npoint, unsigned long ndim, unsigned long nva
...
     Solution_time_n.resize(nPoint,nVar) = su2double(0.0);
 
+  if (config->GetTime_Domain()) {
+    Density_unsteady.resize(nPoint);
+    Density_time_n.resize(nPoint);
+  }
+
   if (config->GetTime_Marching() != TIME_MARCHING::STEADY)
-    Solution_time_n1.resize(nPoint,nVar) = su2double(0.0);
+    Solution_time_n1.resize(nPoint,nVar) = su2double(0.0);
```
Cristopher-Morales:
```suggestion
    Solution_time_n1.resize(nPoint,nVar) = su2double(0.0);
```

SU2_CFD/src/variables/CVariable.cpp
```diff
@@ -55,14 +58,18 @@ CVariable::CVariable(unsigned long npoint, unsigned long ndim, unsigned long nva
...
   Solution_Old.resize(nPoint,nVar) = su2double(0.0);
 
   if (config->GetTime_Domain())
     Solution_time_n.resize(nPoint,nVar) = su2double(0.0);
 
+  if (config->GetTime_Domain()) {
+    Density_unsteady.resize(nPoint);
+    Density_time_n.resize(nPoint);
+  }
+
```
Cristopher-Morales:
```suggestion
  if (config->GetTime_Domain()) {
    Solution_time_n.resize(nPoint,nVar) = su2double(0.0);
    Density_unsteady.resize(nPoint);
    Density_time_n.resize(nPoint);
  }

```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -83,8 +83,7 @@ class CIncEulerVariable : public CFlowVariable {
    * \param[in] nvar - Number of variables of the problem.
    * \param[in] config - Definition of the particular problem.
    */
-
-  CIncEulerVariable(su2double pressure, const su2double *velocity, su2double temperature,
+  CIncEulerVariable(su2double density, su2double pressure, const su2double *velocity, su2double temperature,
```
Cristopher-Morales:
```suggestion
  CIncEulerVariable(su2double pressure, const su2double *velocity, su2double enthalpy, su2double density,
```
Hi Berk!

as the pull request related to enthalpy for incompressible flows has been merged to develop, 
I suggest to change the other, just to avoid future misunderstanding of what variables are solved in the incompressible solver.
The same applies to the CIncNSVariable.hpp
Let me know what you think about this!!!

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -78,6 +78,7 @@ bool CIncNSVariable::SetPrimVar(unsigned long iPoint, su2double eddy_visc, su2do
   /*--- Set the value of the density ---*/
 
   const auto check_dens = SetDensity(iPoint, FluidModel->GetDensity());
+  Density_unsteady[iPoint] = FluidModel->GetDensity();
```
Cristopher-Morales:
```suggestion

```

I think it should be set after the block where the checks for Non-physical solution are done.
After line 104, in case that a non-physical density is found.

SU2_CFD/src/variables/CIncNSVariable.cpp
Cristopher-Morales, line 96:
```suggestion

/*--- Set density for unsteady problems ---*/
Density_unsteady[iPoint] = FluidModel->GetDensity();

```

here it should be set.

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -67,8 +68,7 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
   }
 }
 
-bool CIncEulerVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel) {
-
+bool CIncEulerVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)  {
```
Cristopher-Morales:
```suggestion
bool CIncEulerVariable::SetPrimVar(unsigned long iPoint, CFluidModel *FluidModel)  {

```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1072,6 +1072,7 @@ void CFVMFlowSolverBase<V, R>::PushSolutionBackInTime(unsigned long TimeIter, bo
   for (unsigned short iMesh = 0; iMesh <= config->GetnMGLevels(); iMesh++) {
     solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Solution_time_n();
     solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Solution_time_n1();
+    solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n();
```
bigfooted:
```suggestion
    solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n();
    solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n1();
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.inl
```diff
@@ -1102,6 +1103,7 @@ void CFVMFlowSolverBase<V, R>::PushSolutionBackInTime(unsigned long TimeIter, bo
 
     for (unsigned short iMesh = 0; iMesh <= config->GetnMGLevels(); iMesh++) {
       solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Solution_time_n();
+      solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n();
```
bigfooted:
```suggestion
      solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n();
      solver_container[iMesh][FLOW_SOL]->GetNodes()->Set_Density_time_n1();
```

SU2_CFD/include/output/COutput.hpp
bigfooted: only add when incompressible + unsteady + energy

SU2_CFD/src/solvers/CIncEulerSolver.cpp
bigfooted: density_unsteady is only used to copy into density_n and density_n1, so no need to use it anywhere else.

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -58,6 +58,7 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
   if (dual_time) {
     Solution_time_n = Solution;
     Solution_time_n1 = Solution;
+    Density_unsteady = density;
```
bigfooted:
```suggestion
    Density_time_n = density;
    Density_time_n1 = density;
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
bigfooted: time is never capitalized for some reason. let's go through this later for proper variable naming...

SU2_CFD/src/solvers/CIncEulerSolver.cpp
Cristopher-Morales:
```suggestion
      Density = nodes->GetDensity(iPoint);
      Density_time_nM1 = nodes->GetDensity_time_n1(iPoint);
      Density_time_n = nodes->GetDensity_time_n(iPoint);
```

I think these densities should be added

Cristopher-Morales:
```suggestion
V2U(Density_time_nM1, V_time_nM1, U_time_nM1);
V2U(Density_time_n, V_time_n, U_time_n);
V2U(Density, V_time_nP1, U_time_nP1);
```
I think these should be changed

Cristopher-Morales:
```suggestion
Density_time_n = nodes->GetDensity_time_n(iPoint);
V2U(Density_time_n, V_time_n, U_time_n);
```

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -102,6 +102,9 @@ bool CIncNSVariable::SetPrimVar(unsigned long iPoint, su2double eddy_visc, su2do
 
   }
 
+/*--- Set density for unsteady problems ---*/
+Density_unsteady[iPoint] = FluidModel->GetDensity();
+
```
Cristopher-Morales:
```suggestion
/*--- Set density for unsteady problems ---*/
 if (Unsteady) {
    SetDensity_unsteady(iPoint, FluidModel->GetDensity();)
  }
```

the test cases such as the sandia test case are failing because when the problem is steady, there is no memory allocated for Density_unsteady. So if Density_unsteady is needed, it should be set only for unsteady problems. Then a flag similar to Energy in the SetPrimVar should be added, and its initialization should go in the list of like this:

```
CIncNSVariable::CIncNSVariable(su2double density, su2double pressure, const su2double *velocity, su2double enthalpy,
                               unsigned long npoint, unsigned long ndim, unsigned long nvar, const CConfig *config) :
                               CIncEulerVariable(density, pressure, velocity, enthalpy, npoint, ndim, nvar, config),
                               Energy(config->GetEnergy_Equation()), Unsteady(config->GetTime_Domain()){
```

SU2_CFD/include/output/COutput.hpp
```diff
@@ -302,7 +302,7 @@ class COutput {
   unsigned short nRequestedVolumeFields;
 
   /*! \brief Minimum required volume fields for restart file. */
-  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY"};
+  const std::vector<string> restartVolumeFields = {"COORDINATES", "SOLUTION", "SENSITIVITY", "GRID_VELOCITY", "DENSITY_TIME_N", "DENSITY_TIME_N1"};
```
bigfooted: add density only with unsteady

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2897,7 +2899,8 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
 
       V_time_n = nodes->GetSolution_time_n(iPoint);
       Density = nodes->GetDensity(iPoint);
-      V2U(Density, V_time_n, U_time_n);
+Density_time_n = nodes->GetDensity_time_n(iPoint);
+V2U(Density_time_n, V_time_n, U_time_n);
```
bigfooted: alignment

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2981,12 +2984,16 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
...
+      Density_time_n = nodes->GetDensity_time_n(iPoint);
 
       /*--- Compute the conservative variable vector for all time levels. ---*/
 
       V2U(Density, V_time_nM1, U_time_nM1);
       V2U(Density, V_time_n, U_time_n);
-      V2U(Density, V_time_nP1, U_time_nP1);
+V2U(Density_time_nM1, V_time_nM1, U_time_nM1);
+V2U(Density_time_n, V_time_n, U_time_n);
+V2U(Density, V_time_nP1, U_time_nP1);
```
bigfooted: alignment

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -221,6 +223,7 @@ void CSpeciesFlameletSolver::SetInitialCondition(CGeometry** geometry, CSolver**
 
       prog_unburnt = config->GetSpecies_Init()[I_PROGVAR];
       SU2_OMP_FOR_STAT(omp_chunk_size)
+      //for (unsigned long i_point = 0; i_point < geometry[i_mesh]->GetnPointDomain(); i_point++) {
```
bigfooted:
```suggestion
```

SU2_CFD/src/variables/CIncNSVariable.cpp
```diff
@@ -47,6 +47,15 @@ CIncNSVariable::CIncNSVariable(su2double pressure, const su2double *velocity, su
     AuxVar.resize(nPoint,nAuxVar) = su2double(0.0);
     Grad_AuxVar.resize(nPoint,nAuxVar,nDim);
   }
+
+  if (config->GetTime_Marching() != TIME_MARCHING::STEADY) {
+    Unsteady = true;
+    //Density_unsteady.resize(nPoint);
+    //Density_time_n.resize(nPoint);
+    //Density_time_n1.resize(nPoint);
```
bigfooted:
```suggestion
```

bigfooted:
@tkiymaz I updated the code with the changes that we discussed in the past weeks. There's also the bug fix for the combustion restart. Can you add the testcase with a restart file? The su2 mesh and restart file go in the testcases repository, the cfg file can go in this PR.

bigfooted:
@pcarruscag can you have a quick look at this PR? The primal works as expected, the adjoint unsteady will be finalized in another PR.

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -686,13 +684,26 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
...
+          unsteady_term = (Density_nP1 * U_time_nP1[iVar] - Density_n * U_time_n[iVar]) * Volume_nP1 / TimeStep;
         if (second_order)
-          LinSysRes(iPoint, iVar) += (3.0 * Density_nP1 * U_time_nP1[iVar] - 4.0 * Density_n * U_time_n[iVar] +
+          unsteady_term = (3.0 * Density_nP1 * U_time_nP1[iVar] - 4.0 * Density_n * U_time_n[iVar] +
                                       1.0 * Density_nM1 * U_time_nM1[iVar]) *
                                      Volume_nP1 / (2.0 * TimeStep);
+
+        bool bounded_scalar = false;
+        if (config->GetKind_Solver() == MAIN_SOLVER::INC_EULER || config->GetKind_Solver() == MAIN_SOLVER::INC_NAVIER_STOKES ||
+            config->GetKind_Solver() == MAIN_SOLVER::INC_RANS) {
```
pcarruscag: There is an "incompressible regime" bool function in CConfig

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -833,20 +840,49 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
...
+	  if (IsSpeciesSolver()) bounded_scalar = config->GetBounded_Species();
+	  else if (IsTurbSolver()) bounded_scalar = config->GetBounded_Turb();
+	}
+	su2double diag_factor = 1.0;
+	if (Conservative) {
+	  if (bounded_scalar) {
+           if (first_order)  diag_factor = Density_n;
+           if (second_order) diag_factor = (4.0 * Density_n - Density_nM1) / 3.0;
+	 } else {
+	  diag_factor = Density_nP1;
```
pcarruscag: formating

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -833,20 +840,49 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
...
       }
 
       /*--- Compute the Jacobian contribution due to the dual time source term. ---*/
       if (implicit) {
-        if (first_order) Jacobian.AddVal2Diag(iPoint, Volume_nP1 / TimeStep);
-        if (second_order) Jacobian.AddVal2Diag(iPoint, (Volume_nP1 * 3.0) / (2.0 * TimeStep));
-      }
+	bool bounded_scalar = false;
+	if (config->GetKind_Solver() == MAIN_SOLVER::INC_EULER || config->GetKind_Solver() == MAIN_SOLVER::INC_NAVIER_STOKES ||
+            config->GetKind_Solver() == MAIN_SOLVER::INC_RANS) {
```
pcarruscag: same about compressible

SU2_CFD/include/solvers/CSpeciesSolver.hpp
```diff
@@ -69,6 +69,11 @@ class CSpeciesSolver : public CScalarSolver<CSpeciesVariable> {
 
   void Initialize(CGeometry* geometry, CConfig* config, unsigned short iMesh, unsigned short nVar);
 
+  /*!
+   * \brief Virtual function returning whether this is the species solver.
+   */
+  inline bool IsSpeciesSolver() const override { return true; }
```
pcarruscag:
```suggestion
  inline bool IsSpeciesSolver() const final { return true; }
```

SU2_CFD/include/solvers/CTurbSolver.hpp
```diff
@@ -55,6 +55,11 @@ class CTurbSolver : public CScalarSolver<CTurbVariable> {
    */
   CTurbSolver(CGeometry* geometry, CConfig *config, bool conservative);
 
+  /*!
+   * \brief Virtual function returning whether this is the turbulent solver.
+   */
+  inline bool IsTurbSolver() const override { return true; }
```
pcarruscag:
```suggestion
  inline bool IsTurbSolver() const final { return true; }
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -71,6 +71,8 @@ class CIncEulerVariable : public CFlowVariable {
 
   VectorType Streamwise_Periodic_RecoveredPressure,    /*!< \brief Recovered/Physical pressure [Pa] for streamwise periodic flow. */
              Streamwise_Periodic_RecoveredTemperature; /*!< \brief Recovered/Physical temperature [K] for streamwise periodic flow. */
+  VectorType Density_time_n,                           /*!< \brief Density at time n for dual-time stepping. */
+             Density_time_n1;                          /*!< \brief Density at time n-1 for dual-time stepping. */
```
pcarruscag, line 75: Is it possible to recompute the old densities on the fly from the old solution variables instead of storing?
This would avoid the issues with restart files, adjoints, etc.

tkiymaz: Dear Pedro, thanks for the advice, but I'd keep storing it here due to runtime. In the flamelet path ρ isn't a closed-form EoS; it's a multi-dimensional LUT interpolation (one of the most expensive per-node ops in the solver). ρⁿ/ρⁿ⁻¹ are frozen over the whole physical time step, but the dual-time residual is assembled every inner iteration (and it gets even more computationally costly with high inner iteration number), so storing evaluates each old density once per step while recomputing from the old state would redo the identical LUT lookup. It also mirrors the existing scheme: we already store Solution_time_n/_n1 rather than reconstructing old states on the fly. There is no problem with the restart right now. Regarding adjoints, I can do a future work on it.

bigfooted: @pcarruscag : do you mean only recompute the density when doing a restart or always recompute from the EoS? Because at restart that would be beneficial in the sense that it saves (or actually it does not save) 2 restart fields. 
@tkiymaz we have the possibility to use the ideal gas law instead of lookup for the steady state flamelet solver. Do we re-evaluate the EoS every inner iteration with this approach?

SU2_CFD/include/solvers/CScalarSolver.inl
pcarruscag: We should use CFlowVariable to abstract this away.
Declare the `GetDensity_time_n` functions in CFlowVariable as virtual, the default implementation grabs density from the solution containers, whereas CIncEulerVariable overrides and uses its own containers for density.
Then the scalar solver does not need to carry logic about whether the flow is compressible or incompressible.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2227,6 +2227,48 @@ class CVariable {
...
+
+  /*!
+   * \brief Register the density at time n as an AD input (no-op for non-incompressible nodes).
+   */
+  inline virtual void RegisterDensity_time_n() {}
+
+  /*!
+   * \brief Register the density at time n-1 as an AD input (no-op for non-incompressible nodes).
+   */
+  inline virtual void RegisterDensity_time_n1() {}
```
pcarruscag: Can't this be part of registering the time_n solutions?

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -439,6 +449,14 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
   SetVolumeOutputValue("PRESSURE_COEFF", iPoint, (Node_Flow->GetPressure(iPoint) - solver[FLOW_SOL]->GetPressure_Inf())/factor);
   SetVolumeOutputValue("DENSITY", iPoint, Node_Flow->GetDensity(iPoint));
 
+  // Load density at previous timesteps for restart (unsteady with non-constant density)
+  if (config->GetTime_Domain() && config->GetKind_FluidModel() != CONSTANT_DENSITY) {
+    SetVolumeOutputValue("DENSITY_TIME_N", iPoint, Node_Flow_Inc->GetDensity_time_n(iPoint));
+    if (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_2ND) {
+      SetVolumeOutputValue("DENSITY_TIME_N1", iPoint, Node_Flow_Inc->GetDensity_time_n1(iPoint));
+    }
+  }
```
pcarruscag: We have density in the restart file, we load more than one restart file for transient to get the previous solutions.
Why do we need each of the multiple files we write and load to also carry old information? This way we are storing and loading repeated information.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -168,11 +168,18 @@ void CDiscAdjSolver::RegisterSolution(CGeometry *geometry, CConfig *config) {
   /*--- Register quantities that are no solver variables but further inputs/outputs of the (outer) iteration. ---*/
   direct_solver->RegisterSolutionExtra(true, config);
 
-  if (time_n_needed)
+  if (time_n_needed) {
     direct_solver->GetNodes()->RegisterSolution_time_n();
+    /*--- Density is stored separately from the solution vector for incompressible flows. ---*/
+    if (config->GetKind_Regime() == ENUM_REGIME::INCOMPRESSIBLE)
+      direct_solver->GetNodes()->RegisterDensity_time_n();
+  }
```
pcarruscag, line 173: So make RegisterSolution_time_n virtual and specialize it for the incompressible variable.
The purpose of abstractions like CSolver, CVariable, etc. is to keep physics/regime specific aspects contained in the "leaf classes" to allow the bigger ones like iteration, driver, etc. to deal with more general orchestration aspects

SU2_CFD/include/variables/CIncEulerVariable.hpp
pcarruscag: I meant always, but if it's too expensive it's ok

SU2_CFD/src/output/CFlowIncOutput.cpp
bigfooted: you are right, we do not need this.

SU2_CFD/include/variables/CVariable.hpp
bigfooted: yes.

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
bigfooted: no problem

SU2_CFD/include/variables/CIncEulerVariable.hpp
bigfooted: obsolete...

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -624,9 +624,13 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
   const bool second_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_2ND);
   const bool incompressible = (config->GetKind_Regime() == ENUM_REGIME::INCOMPRESSIBLE);
 
+  /*--- The bounded-scalar correction is currently only used by the incompressible solvers. ---*/
+  const bool bounded_scalar = incompressible && ((IsSpeciesSolver() && config->GetBounded_Species()) ||
+                                                 (IsTurbSolver()    && config->GetBounded_Turb()));
+
```
bigfooted: needs to be fixed still

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -686,13 +680,20 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
...
-              (Density_nP1 * U_time_nP1[iVar] - Density_n * U_time_n[iVar]) * Volume_nP1 / TimeStep;
+          unsteady_term = (Density_nP1 * U_time_nP1[iVar] - Density_n * U_time_n[iVar]) * Volume_nP1 / TimeStep;
         if (second_order)
-          LinSysRes(iPoint, iVar) += (3.0 * Density_nP1 * U_time_nP1[iVar] - 4.0 * Density_n * U_time_n[iVar] +
+          unsteady_term = (3.0 * Density_nP1 * U_time_nP1[iVar] - 4.0 * Density_n * U_time_n[iVar] +
                                       1.0 * Density_nM1 * U_time_nM1[iVar]) *
                                      Volume_nP1 / (2.0 * TimeStep);
+
+        if (bounded_scalar) {
+          if (first_order) unsteady_term -= U_time_nP1[iVar] * (Density_nP1 - Density_n) * Volume_nP1 / TimeStep;
```
bigfooted, line 705: check again why only for bounded scalar?

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -270,4 +270,26 @@ class CFlowVariable : public CVariable {
    * \return Vector of magnitudes.
    */
   inline su2activevector& GetStrainMag() { return StrainMag; }
+
+  /*!
+   * \brief Get the density at time level n for dual-time stepping (compressible flows).
+   * For compressible flows, density is extracted from the solution container.
+   * For incompressible flows, this is overridden to use dedicated storage.
```
bigfooted:
```suggestion
   * For incompressible flows, this is overridden.
```

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -270,4 +270,26 @@ class CFlowVariable : public CVariable {
...
+   * \return Density at time level n.
+   */
+  inline virtual su2double GetDensity_time_n(unsigned long iPoint) const {
+    return GetSolution_time_n(iPoint, 0);  // Density is the first component of Solution
+  }
+
+  /*!
+   * \brief Get the density at time level n-1 for dual-time stepping (compressible flows).
+   * For compressible flows, density is extracted from the solution container.
+   * For incompressible flows, this is overridden to use dedicated storage.
```
bigfooted:
```suggestion
   * For incompressible flows, this is overridden.
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -291,4 +303,68 @@ class CIncEulerVariable : public CFlowVariable {
...
+  void RegisterSolution_time_n() override;
+
+  /*!
+   * \brief Register solution at time n-1, including the density container for incompressible flows.
+   */
+  void RegisterSolution_time_n1() override;
+
+  /*!
+   * \brief Get the adjoint of density at time n.
+   *        Simplified version: directly extracts the AD derivative of Density_time_n,
```
bigfooted: check what this was about again...

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -291,4 +303,74 @@ class CIncEulerVariable : public CFlowVariable {
...
   }
 
+  /*!
+   * \brief Get the density at time level n for dual-time stepping.
+   * \param[in] iPoint - Point index.
+   * \return Density at time level n.
+   */
+  inline su2double GetDensity_time_n(unsigned long iPoint) const override {
+    return Density_time_n.size() > 0 ? Density_time_n(iPoint) : GetDensity(iPoint);
+  }
```
bigfooted: @pcarruscag like this?

bigfooted:
@tkiymaz  @pcarruscag 
Done from my side, please have a look...

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -974,6 +976,20 @@ void CIncEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_
   SU2_OMP_ATOMIC
   ErrorCounter += SetPrimitive_Variables(solver_container, config);
 
+  /*--- Provide a valid initial Density_time_n for restart or step 0 since
+        it was 0.0 before SetPrimitive_Variables evaluated the exact field.
+        Checking GetDensity_time_n(0) == 0.0 ensures we only evaluate this
+        once when the history arrays are uninitialized. ---*/
+  if (dual_time && nPoint > 0 && nodes->GetDensity_time_n(0) == 0.0 && iMesh == MESH_0) {
```
pcarruscag: This is not OpenMP-safe: `CommonPreprocessing` runs inside the solver's parallel region, and this worksharing loop sits behind a branch whose condition is invalidated by the loop itself. A fast thread can enter the `omp for` and write point 0 while a slower thread has not evaluated the condition yet; the slow thread then sees a nonzero value and skips the construct. A worksharing construct encountered by only some threads of a team is undefined behavior (in practice a hang, and at minimum a TSan report in the CI). The decision to initialize must be made in a single-threaded context (`BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS`, or a flag set once outside the parallel region), not by probing the data from all threads.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -974,6 +976,20 @@ void CIncEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_
   SU2_OMP_ATOMIC
   ErrorCounter += SetPrimitive_Variables(solver_container, config);
 
+  /*--- Provide a valid initial Density_time_n for restart or step 0 since
```
pcarruscag: Three related problems with this initialization strategy:

1. **Multigrid**: the `iMesh == MESH_0` guard means the coarse-grid variable objects (which also allocate `Density_time_n = 0`) are never initialized, and `SetResidual_DualTime` runs on all MG levels — so `MGLEVEL > 0` + variable density + dual time uses rho_n = 0 on the coarse levels for the first time step. Either initialize all levels or error out for this combination.
2. **Restart consistency**: on restart, rho_n and rho_n-1 are set to the *current* density rather than densities consistent with `Solution_time_n` / `Solution_time_n1`, so the first step after restart is not a true continuation (and the 2nd-order history is wrong — the PR says only 1st order is implemented, but the 2nd-order branches exist and nothing guards `DUAL_TIME_STEPPING-2ND_ORDER`).
3. **The `== 0.0` sentinel** probed every inner iteration is fragile even once the race is fixed.

All three disappear with one change: recompute `Density_time_n{,1}` from the stored solution histories via the fluid model once per time step (after the push-back), instead of lazily cloning the current density. That is correct after restart by construction, needs no sentinel, works on every MG level it runs on, and — if the recomputation sits in the recorded path — the dual-time residual's dependence on rho_n flows back to the already-registered `Solution_time_n` inputs on the tape, closing the adjoint chain rule that is currently left as a TODO, with no manual extraction needed.

SU2_CFD/include/solvers/CSolver.hpp
```diff
@@ -347,6 +347,16 @@ class CSolver {
    */
   inline virtual CFluidModel* GetFluidModel(void) const { return nullptr;}
 
+  /*!
+   * \brief Virtual function returning whether this is the species solver.
+   */
+  inline virtual bool IsSpeciesSolver() const { return false; }
```
pcarruscag: These two virtuals don't belong on the `CSolver` base: their only call site is `CScalarSolver::SetResidual_DualTime`, called on `this`. A protected `bounded_scalar` member of `CScalarSolver`, set by the derived constructors from `GetBounded_Species()` / `GetBounded_Turb()` (same pattern as the existing `Conservative` flag), does the job with zero base-class footprint and no per-call config queries.

SU2_CFD/include/variables/CFlowVariable.hpp
```diff
@@ -270,4 +270,26 @@ class CFlowVariable : public CVariable {
...
   inline su2activevector& GetStrainMag() { return StrainMag; }
+
+  /*!
+   * \brief Get the density at time level n for dual-time stepping (compressible flows).
+   * For compressible flows, density is extracted from the solution container.
+   * For incompressible flows, this is overridden.
+   * \param[in] iPoint - Point index.
+   * \return Density at time level n.
+   */
+  inline virtual su2double GetDensity_time_n(unsigned long iPoint) const {
```
pcarruscag: Two things here:

1. These override the `CVariable` versions, so per the style guide they should be marked `override` (and then don't need `virtual`).
2. More fundamentally, the base declarations on `CVariable` are not needed at all. The only call site that goes through `CVariable*` is `CScalarSolver::SetResidual_DualTime`, and the idiom for exactly this situation is already used in the same file (CScalarSolver.inl:156): `su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes())`. With that cast, these getters can be declared here on `CFlowVariable` (compressible default) with the `CIncEulerVariable` override, and the dangerous silent-`0.0` default on `CVariable` disappears — which is exactly what makes the uninitialized coarse-MG case fail silently instead of loudly.

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2214,19 +2214,53 @@ class CVariable {
...
    * \brief Register the variables in the user defined source array as input/output variable.
    */
   void RegisterUserDefinedSource();
 
+  /*!
+   * \brief Get the stored density at time n (incompressible only).
+   * \param[in] iPoint - Point index.
+   * \return Density at time n; 0 for compressible nodes.
+   */
+  inline virtual su2double GetDensity_time_n(unsigned long /*iPoint*/) const { return 0.0; }
```
pcarruscag: See the comment on `CFlowVariable` — these declarations can be removed from `CVariable` entirely by casting to `CFlowVariable*` at the scalar-solver call site. A base default that silently returns 0.0 for the density turns any missed override or missed initialization into a silently zeroed unsteady term. The doc comments here ("incompressible only; 0 for compressible nodes") also contradict the `CFlowVariable` behavior.

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -291,4 +303,74 @@ class CIncEulerVariable : public CFlowVariable {
...
+
+  /*!
+   * \brief Get the adjoint of density at time n.
+   *        Simplified version: directly extracts the AD derivative of Density_time_n,
+   *        analogous to GetAdjointSolution_time_n for the solution vector.
+   *        The chain-rule contribution d(rho)/d(h) is currently not applied (TODO).
+   * \param[in] iPoint - Point index.
+   * \return Adjoint of the density at time n.
+   */
+  inline su2double GetAdjointDensity_time_n(unsigned long iPoint) const override {
```
pcarruscag: `GetAdjointDensity_time_n{,1}` are never called anywhere — the only references are the TODO comments in `CDiscAdjSolver.cpp`. Please remove them (here and the `CVariable` defaults), along with the `RegisterSolution_time_n{,1}` overrides that register inputs whose adjoints are never extracted, and add them together with the actual chain-rule extraction in a follow-up (or better, avoid the manual chain rule entirely by recomputing the density histories on tape — see the comment in `CIncEulerSolver.cpp`). Until the adjoint is closed, an error for unsteady discrete adjoint + non-constant `INC_DENSITY_MODEL` would prevent silently wrong gradients.

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include "../../../Common/include/geometry/CGeometry.hpp"
 #include "../../include/solvers/CSolver.hpp"
+#include "../../include/variables/CIncEulerVariable.hpp"
```
pcarruscag: This include appears to be unused — nothing else in this file changed and no `CIncEulerVariable` symbol is referenced.

tkiymaz: Done in 12fd1f6.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
tkiymaz: Folded into the fix in abe9e19. The initialization is now guarded by InnerIter == 0, which is a uniform predicate, so the worksharing loop inside is entered by every thread of the team.

SU2_CFD/include/solvers/CSolver.hpp
tkiymaz: Done in de8eb53. Both virtuals pulled off CSolver. CScalarSolver now stores a const bool BoundedScalar set from the derived constructors, same pattern as Conservative — CSpeciesSolver passes GetBounded_Species(), CTurbSolver passes GetBounded_Turb(), CHeatSolver passes false.

SU2_CFD/include/variables/CFlowVariable.hpp
tkiymaz: Done in 8120bda. GetDensity_time_n[,1] now live on CFlowVariable with the CIncEulerVariable override, and CScalarSolver::SetResidual_DualTime casts to CFlowVariable* using the same su2staticcast_p<> idiom. Thanks for pointing at the existing pattern.

SU2_CFD/include/variables/CVariable.hpp
tkiymaz: Fixed in same commit (8120bda). Silent-0 defaults are gone — a missed override would now fail to compile instead of quietly zeroing the unsteady term.

SU2_CFD/include/variables/CIncEulerVariable.hpp
tkiymaz: Done in 324d94d — GetAdjointDensity_time_n[,1], the RegisterSolution_time_n[,1] density overrides, and the CVariable defaults are all gone, along with the two TODOs in CDiscAdjSolver::ExtractAdjoint_Solution. Added an SU2_MPI::Error in the discrete-adjoint constructor for unsteady + non-CONSTANT INC_DENSITY_MODEL.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
tkiymaz: Done in abe9e19. Density_time_n[,1] are now recomputed from Solution_time_n[,1] (plus the species histories on the flamelet path) with the fluid model. It happens once per physical time step in CommonPreprocessing. 
I also updated the flamelet regression test_vals because the pre-fix values were tuned with the previous runs.

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -86,6 +86,19 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
   virtual unsigned long SetPrimitive_Variables(CSolver **solver_container, const CConfig *config);
 
+  /*!
+   * \brief Recompute the dual-time density history (rho_n, rho_n-1) from the stored
+   *        solution histories (Solution_time_n[,1], and the species histories for
+   *        flamelet/variable-density) via the fluid model.
+   * \details Called once per physical time step (InnerIter == 0), after the solution
+   *          push-back. This keeps the density history consistent with the primitive
+   *          state at time n (also after restart), on every MG level where the inputs
+   *          exist, without the fragile per-iteration sentinel that used to seed it.
```
pcarruscag:
```suggestion
   *          state at time n (also after restart).
```

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -622,11 +622,16 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
   const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
   const bool first_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_1ST);
   const bool second_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_2ND);
-  const bool incompressible = (config->GetKind_Regime() == ENUM_REGIME::INCOMPRESSIBLE);
 
-  /*--- Flow solution, needed to get density. ---*/
+  /*--- The bounded-scalar correction (set from the derived solver constructor). ---*/
```
pcarruscag: This type of AI comment is utterly useless (the strong words are for Claude to pay attention).
```suggestion
```

SU2_CFD/include/solvers/CScalarSolver.inl
```diff
@@ -622,11 +622,16 @@ void CScalarSolver<VariableType>::SetResidual_DualTime(CGeometry* geometry, CSol
...
+  const bool bounded_scalar = BoundedScalar;
 
-  CVariable* flowNodes = solver_container[FLOW_SOL]->GetNodes();
+  /*--- Flow solution, needed to get density. The static cast to CFlowVariable*
+        exposes GetDensity_time_n[,1] (declared on CFlowVariable, overridden by
+        CIncEulerVariable) without a silent 0.0-returning base on CVariable.
+        Same idiom as Upwind_Residual above. ---*/
+
```
pcarruscag: This is not the place to document another class hierarchy.
```suggestion
  /*--- Flow solution, needed to get density. ---*/
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -291,4 +293,36 @@ class CIncEulerVariable : public CFlowVariable {
     for (unsigned long iDim = 0; iDim < nDim; iDim++) Solution(iPoint, iDim+1) = val_vector[iDim];
   }
 
+  /*!
+   * \brief Get the density at time level n for dual-time stepping.
+   * \param[in] iPoint - Point index.
+   * \return Density at time level n.
+   */
+  inline su2double GetDensity_time_n(unsigned long iPoint) const override {
```
```suggestion
  inline su2double GetDensity_time_n(unsigned long iPoint) const final {
```

SU2_CFD/include/variables/CIncEulerVariable.hpp
```diff
@@ -291,4 +293,36 @@ class CIncEulerVariable : public CFlowVariable {
...
+    return Density_time_n.size() > 0 ? Density_time_n(iPoint) : GetDensity(iPoint);
+  }
+
+  /*!
+   * \brief Get the density at time level n-1 for dual-time stepping.
+   * \param[in] iPoint - Point index.
+   * \return Density at time level n-1.
+   */
+  inline su2double GetDensity_time_n1(unsigned long iPoint) const override {
```
```suggestion
  inline su2double GetDensity_time_n1(unsigned long iPoint) const final {
```

SU2_CFD/include/variables/CFlowVariable.hpp
pcarruscag: Make them pure virtual in this class and override (final) in CEulerVariable and CNemoEulerVariable such that the flow solver can optimize the virtual away. The way it is now, the compiler cannot prove that these are final for the compressible solvers.

SU2_CFD/include/variables/CVariable.hpp
pcarruscag, line 1: changes in this file are no longer needed, revert

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
```diff
@@ -111,6 +111,16 @@ CDiscAdjSolver::CDiscAdjSolver(CGeometry *geometry, CConfig *config, CSolver *di
...
+
+  /*--- The discrete adjoint chain rule through the time-history of the incompressible density
+        is not yet implemented. Bail out until it is, to avoid silently wrong gradients. ---*/
+  if (KindDirect_Solver == RUNTIME_FLOW_SYS &&
+      config->GetKind_Regime() == ENUM_REGIME::INCOMPRESSIBLE &&
+      config->GetTime_Marching() != TIME_MARCHING::STEADY &&
+      config->GetKind_DensityModel() != INC_DENSITYMODEL::CONSTANT) {
+    SU2_MPI::Error("Unsteady discrete adjoint with non-constant INC_DENSITY_MODEL is not yet supported.",
+                   CURRENT_FUNCTION);
+  }
```
pcarruscag: Probably not needed anymore since you are recomputing densities like I suggested in the first place.
Either way settings validation is done in CConfig not in the solvers so that dry run mode can catch issues like these.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -974,6 +976,18 @@ void CIncEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_
...
   ErrorCounter += SetPrimitive_Variables(solver_container, config);
 
+  /*--- Recompute the dual-time density history from the stored solution history
+        via the fluid model, once per physical time step after the push-back.
+        This is consistent with the primitive state at time n by construction
+        (also after restart, since Solution_time_n[,1] are the loaded histories)
+        and needs no sentinel. The InnerIter == 0 guard is a uniform, non
+        data-dependent condition, so the worksharing loop inside is entered by
+        every thread of the team -- OpenMP-safe, unlike the old probe of
+        GetDensity_time_n(0) == 0.0. ---*/
```
pcarruscag: What is the purpose of repeating the function documentation?
What is the purpose of making a reference to an iteration of the code that will not be kept after merging into develop?
```suggestion
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1077,6 +1091,48 @@ unsigned long CIncEulerSolver::SetPrimitive_Variables(CSolver **solver_container
...
+  /*--- For flamelet / multicomponent density the fluid-model lookup needs the
+        scalar state; those come from the species solver, which only exists on
+        the fine grid. Where it is absent (coarse MG levels, or non-species
+        variable-density cases) the density depends on the flow enthalpy alone. ---*/
+  CVariable* speciesNodes = (solver_container[SPECIES_SOL] != nullptr)
+                              ? solver_container[SPECIES_SOL]->GetNodes() : nullptr;
+
+  /*--- If the density model requires scalars but the species solver is not
+        available on this level, leave the history untouched (it is maintained
+        by MG restriction of the primitive field). ---*/
```
pcarruscag: no it's not

SU2_CFD/src/variables/CIncEulerVariable.cpp
```diff
@@ -58,6 +59,16 @@ CIncEulerVariable::CIncEulerVariable(su2double pressure, const su2double *veloci
...
+
+    /*--- For constant-density flows, save memory by using GetDensity() on demand. ---*/
+    const bool variable_density = (config->GetKind_DensityModel() != INC_DENSITYMODEL::CONSTANT);
+
+    if (variable_density) {
+      /*--- Allocate density storage for time levels.
+            Note: Actual density values will be set after SetPrimVar is called in Preprocessing. ---*/
+      Density_time_n.resize(nPoint) = su2double(0.0);
+      Density_time_n1.resize(nPoint) = su2double(0.0);
+    }
```
pcarruscag:
```suggestion
    if (config->GetKind_DensityModel() != INC_DENSITYMODEL::CONSTANT) {
      Density_time_n.resize(nPoint) = su2double(0.0);
      Density_time_n1.resize(nPoint) = su2double(0.0);
    }
```

SU2_CFD/include/solvers/CScalarSolver.inl
tkiymaz: Thanks Pedro, it slipped from my attention, will be more careful in the future

SU2_CFD/include/variables/CVariable.hpp
tkiymaz: Reverted, it is identical to develop again

SU2_CFD/include/variables/CFlowVariable.hpp
tkiymaz: Done as you asked

SU2_CFD/src/solvers/CDiscAdjSolver.cpp
tkiymaz: Removed

SU2_CFD/src/solvers/CIncEulerSolver.cpp
tkiymaz: You are right, nothing maintains it.

tkiymaz:
The two remaining CI failures are da_sp_pinArray_cht_2d_dp_hf and da_unsteadyCHT_cylinder. @bigfooted can you check that?  I tried but could not find the issue.

pcarruscag (approved):

bigfooted:
> The two remaining CI failures are da_sp_pinArray_cht_2d_dp_hf and da_unsteadyCHT_cylinder. @bigfooted can you check that? I tried but could not find the issue.

The residuals are barely above the threshold so I would not worry about it. They are both related to a change to the adjoint implementation, but that (unsteady adjoint) needs to  be tested in more depth anyway for the variable density case. So I suggest to update the residuals now and create a verification test using finite differences later.

tkiymaz:
> > The two remaining CI failures are da_sp_pinArray_cht_2d_dp_hf and da_unsteadyCHT_cylinder. @bigfooted can you check that? I tried but could not find the issue.
> 
> The residuals are barely above the threshold so I would not worry about it. They are both related to a change to the adjoint implementation, but that (unsteady adjoint) needs to be tested in more depth anyway for the variable density case. So I suggest to update the residuals now and create a verification test using finite differences later.

Then let me know if there is anything i can do to finalize the PR.

bigfooted:
> > > The two remaining CI failures are da_sp_pinArray_cht_2d_dp_hf and da_unsteadyCHT_cylinder. @bigfooted can you check that? I tried but could not find the issue.
> > 
> > 
> > The residuals are barely above the threshold so I would not worry about it. They are both related to a change to the adjoint implementation, but that (unsteady adjoint) needs to be tested in more depth anyway for the variable density case. So I suggest to update the residuals now and create a verification test using finite differences later.
> 
> Then let me know if there is anything i can do to finalize the PR.

just go to the failed testcases overview and copy the line with computed results and update the parallel_regression_AD.py with these values.

tkiymaz:
> > > > The two remaining CI failures are da_sp_pinArray_cht_2d_dp_hf and da_unsteadyCHT_cylinder. @bigfooted can you check that? I tried but could not find the issue.
> > > 
> > > 
> > > The residuals are barely above the threshold so I would not worry about it. They are both related to a change to the adjoint implementation, but that (unsteady adjoint) needs to be tested in more depth anyway for the variable density case. So I suggest to update the residuals now and create a verification test using finite differences later.
> > 
> > 
> > Then let me know if there is anything i can do to finalize the PR.
> 
> just go to the failed testcases overview and copy the line with computed results and update the parallel_regression_AD.py with these values.

Done! I believe it is ready now

.github/workflows/regression.yml
```diff
@@ -233,7 +233,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:260405-0054
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c fgm_unsteady -s ${{matrix.testscript}}
```
bigfooted: @tkiymaz I've merged the testcases repository so you can now change this back to develop

bigfooted:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
```

.github/workflows/regression.yml
```diff
@@ -282,7 +282,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:260405-0054
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tapetests"
+          args: -b ${{github.ref}} -t develop -c fgm_unsteady -s ${{matrix.testscript}} -a "--tapetests"
```
bigfooted:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tapetests"
```

.github/workflows/regression.yml
```diff
@@ -330,7 +330,7 @@ jobs:
           PMIX_MCA_gds: hash
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tsan"
+          args: -b ${{github.ref}} -t develop -c fgm_unsteady -s ${{matrix.testscript}} -a "--tsan"
```
bigfooted:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--tsan"
```

.github/workflows/regression.yml
```diff
@@ -375,7 +375,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2-asan:260405-0054
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--asan"
+          args: -b ${{github.ref}} -t develop -c fgm_unsteady -s ${{matrix.testscript}} -a "--asan"
```
bigfooted:
```suggestion
          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}} -a "--asan"
```

.github/workflows/regression.yml
bigfooted: I've changed. should be ready now.

PR #2645, opened 2025-12-12
[WIP] Adding ENERGY  in CPrimitiveIndices to get index of energy equation

Cristopher-Morales:
## Proposed Changes

with @bigfooted , we were discussing about getting the index of the energy equation in order to overwrite the solution independent of the variable that is being actually solved. For Incompressible is the temperature, with pull request #2426 , the enthalpy is going to be solved, and for compressible solvers the total energy is solved. 
However, if a user would like to overwrite the energy solution in the compressible solver, it might encounter an issue in identifying the index of the Total energy, that user would probably have to do something like:

```
# get solution index for energy equation
    iEnergy = solindex.get("PRESSURE")
```

because pressure has index nDim+1 in the primitives, which also corresponds to the index of the energy in the solution vector.
The turbulent_premixed_psi test case using the python wrapper for incompressible flows is a good example of this.

with this pull request we aim to make this distinction between the index of primitives and the index of the solution, in particular, the index of the energy equation.

Obviously, a better solution to this could be implemented.

## Related Work
related with pull request #2426

## PR Checklist

- [ X I am submitting my contribution to the develop branch.

### DIFF
--- a/SU2_CFD/src/solvers/CIncEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CIncEulerSolver.cpp
@@ -2799,18 +2799,17 @@ void CIncEulerSolver::SetResidual_DualTime(CGeometry *geometry, CSolver **solver
   su2double U_time_nM1[MAXNVAR], U_time_n[MAXNVAR], U_time_nP1[MAXNVAR];
   su2double Volume_nM1, Volume_nP1, TimeStep;
   const su2double *Normal = nullptr, *GridVel_i = nullptr, *GridVel_j = nullptr;
-  su2double Density, Cp;
+  su2double Density;
 
   const bool implicit = (config->GetKind_TimeIntScheme() == EULER_IMPLICIT);
   const bool first_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_1ST);
   const bool second_order = (config->GetTime_Marching() == TIME_MARCHING::DT_STEPPING_2ND);
   const bool energy = config->GetEnergy_Equation();
 
-  const int ndim = nDim;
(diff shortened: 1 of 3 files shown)


SU2_CFD/include/variables/CPrimitiveIndices.hpp
```diff
@@ -124,6 +124,9 @@ std::map<std::string, IndexType> PrimitiveNameToIndexMap(const CPrimitiveIndices
   nameToIndex["VELOCITY_Y"] = idx.Velocity() + 1;
   if (idx.NDim() == 3) {
     nameToIndex["VELOCITY_Z"] = idx.Velocity() + 2;
+    nameToIndex["ENERGY"] = idx.Velocity() + 3;
+  } else {
+    nameToIndex["ENERGY"] = idx.Velocity() + 2;
```
pcarruscag, line 129: energy is not stored in the primitive variables

Cristopher-Morales: Hi!

Thank you for your comment!

Yes, It is not stored. 

What  we were discussing in the morning was about how the energy solution index should be obtained without having to know that the temperature is the one that matches the index. In the python wrapper case mentioned, the index was called using "TEMPERATURE", is there a way of defining indexes for the solution and used like "ENERGY_SOLUTION"?

I would really appreciate any suggestion/advice about this

Thank you so much!!

pcarruscag: We should define "solution" or "conservative" indices instead of aliasing the ones in primitives.

Cristopher-Morales: Thanks for suggestion!!

PR #2648, opened 2025-12-17
fix: make preconfigure download paths consistent

shbhmexe:
### Proposed Changes
This PR fixes path handling in the Meson/Ninja preconfigure tooling so downloads and extracted artifacts are placed relative to the script directory (not the current working directory). This makes `preconfigure.py`/`meson.py` more reliable when invoked from outside the repo root.

Files Modified:

| Module | File | Changes |
| --- | --- | --- |
| SU2_BUILD | `preconfigure.py` | Download `ninja-win.zip` into the script directory; fix undefined exception variable in the download error path; ensure built `ninja` is copied into the script directory; close zip after extraction. |
| SU2_BUILD | `meson_scripts/init.py` | Save downloaded module archives into the script directory (consistent with later open/remove); close zip after extraction. |

### Issues Fixed
Preconfigure Tooling:

| Issue | Impact | Fix |
| --- | --- | --- |
| `ninja-win.zip` saved to CWD but later opened/removed from script dir | `FileNotFoundError` when running `preconfigure.py` from outside repo root (Windows) | Download to `sys.path[0]/ninja-win.zip` and extract from the same path. |
| Undefined exception variable in download failure path | `NameError` that hides the real download failure | Use `except Exception as e:` and print the actual exception. |
| Module archives saved to CWD but later opened/removed from script dir | `FileNotFoundError` for non-git source distributions or when using URL downloads from outside repo root | Download to `sys.path[0]` (expected `filepath`). |
| Zip files not closed after extraction | Potential file-handle leaks / Windows file lock issues | Use context managers (`with ZipFile(...) as zipf`). |
| Built `ninja` copied to `.` instead of script dir (non-Windows) | `meson.py` expects `sys.path[0]/ninja`, which can break if invoked from another working directory | Copy to `sys.path[0]/ninja`. |

### Related Work
This PR improves the build/preconfiguration workflow described in `README.md` (Meson + Ninja via `meson.py`) by making downloads path-safe.

### PR Checklist
- [x] My contribution generates no new compiler warnings.
- [ ] I used the pre-commit hook to prevent dirty commits.
- [ ] I have updated appropriate documentation, if necessary.

### DIFF
--- a/meson_scripts/init.py
+++ b/meson_scripts/init.py
@@ -289,7 +289,7 @@ def download_module(name, alt_name, git_repo, commit_sha):
 
             if not os.path.exists(filepath) and not os.path.exists(alt_filepath):
                 try:
-                    urllib.request.urlretrieve(url, filepath)
+                    urllib.request.urlretrieve(url, filename)
                 except Exception as e:
                     print(e)
                     print("Download of module " + name + " failed.")
--- a/preconfigure.py
+++ b/preconfigure.py
@@ -77,7 +77,8 @@ def build_ninja():
                 ["python3", "configure.py", "--bootstrap"], cwd=ninjapath, env=env
             )
             shutil.copy(
-                ninjapath + os.path.sep + "ninja", sys.path[0] + os.path.sep + "ninja"
+                os.path.join(ninjapath, "ninja"),
+                os.path.join(sys.path[0], "ninja"),
             )
 
 


preconfigure.py
```diff
@@ -38,6 +38,7 @@ def build_ninja():
     # If we are on windows, we don't need to compile ninja, we just download the executable
     if os.name == "nt":
         ninja_exe_url = "https://github.com/ninja-build/ninja/releases/download/v1.13.0/ninja-win.zip"
+        ninja_zip_path = sys.path[0] + os.path.sep + "ninja-win.zip"
```
pcarruscag: isn't `path.Join` the pythonic way of doing it?

shbhmexe: os.path.join() / Path is the more idiomatic and safer way to build paths (cleaner, avoids manual separator handling, and stays cross-platform). I’ll switch this to os.path.join(sys.path[0], "ninja-win.zip") for consistency.  
Does that work for your expectations here, or would you prefer I keep the current sys.path[0] + os.path.sep + ... style? If it’s not suitable, I can revert/adjust accordingly.

pcarruscag: Change to join please

shbhmexe: Done , updated it to use os.path.join(sys.path[0], "ninja-win.zip")

preconfigure.py
```diff
@@ -75,7 +76,9 @@ def build_ninja():
             subprocess.run(
                 ["python3", "configure.py", "--bootstrap"], cwd=ninjapath, env=env
             )
-            shutil.copy(ninjapath + os.path.sep + "ninja", ".")
+            shutil.copy(
+                ninjapath + os.path.sep + "ninja", sys.path[0] + os.path.sep + "ninja"
```
pcarruscag: here too

meson_scripts/init.py
```diff
@@ -289,7 +289,7 @@ def _extract_member(self, member, targetpath, pwd):
 
             if not os.path.exists(filepath) and not os.path.exists(alt_filepath):
                 try:
-                    urllib.request.urlretrieve(url, commit_sha + ".zip")
+                    urllib.request.urlretrieve(url, filepath)
```
pcarruscag: Doesn't look equivalent.

preconfigure.py
shbhmexe: Done — updated the shutil.copy() call to use os.path.join(...) for both src/dst, so the ninja binary is copied to the repo root (sys.path[0]) without relying on the current working directory.

meson_scripts/init.py
shbhmexe: urlretrieve() now writes to filename again (commit_sha + ".zip"), which matches the previous behavior. We still compute filepath/alt_filepath for the later existence check/unzip. Verified locally with python -m compileall preconfigure.py meson_scripts/init.py.

PR #2652, opened 2025-12-22
Fix enthalpy related issues

bigfooted:
## Proposed Changes
- use reference temperature for enthalpy consistenly
- Write temperature to paraview when energy=off, being consistent with other thermodynamic quantities like viscosity.
- Fixed a bug when restarting with species.
- When energy=off, enthalpy field will still be used to compute temperature (necessary for python wrapper)
- Fixed an MPI  bug where viscosity was out of sync when computing non-constant mass diffusivity
- update python wrapper case. No need to put ENERGY=ON anymore. temperature field is written to paraview 

## Related Work
small fixes for #2426 

## PR Checklist

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -211,7 +211,7 @@ jobs:
         uses: docker://ghcr.io/su2code/su2/test-su2:250717-1402
         with:
           # -t <Tutorials-branch> -c <Testcases-branch>
-          args: -b ${{github.ref}} -t fix_enthalpy -c fix_enthalpy -s ${{matrix.testscript}}
+          args: -b ${{github.ref}} -t develop -c develop -s ${{matrix.testscript}}
       - name: Cleanup
         uses: docker://ghcr.io/su2code/su2/test-su2:250717-1402
         with:


SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -330,7 +330,7 @@ void CFlowIncOutput::SetVolumeOutputFields(CConfig *config){
     AddVolumeOutput("LAMINAR_VISCOSITY", "Laminar_Viscosity", "PRIMITIVE", "Laminar viscosity");
     AddVolumeOutput("HEAT_CAPACITY", "Heat_Capacity", "PRIMITIVE", "Heat capacity");
     AddVolumeOutput("THERMAL_CONDUCTIVITY", "Thermal_Conductivity", "PRIMITIVE", "Thermal conductivity");
-    if (heat || flamelet) AddVolumeOutput("TEMPERATURE", "Temperature", "PRIMITIVE", "Temperature");
+    AddVolumeOutput("TEMPERATURE", "Temperature", "PRIMITIVE", "Temperature");
```
Cristopher-Morales: Hi!

This was kept in this way because in the heat solver, the temperature is defined as "SOLUTION", then when the INC Solver was used with the Heat Solver, the solution crashed. Maybe this pull request solves that issue. Did you test one of the test cases that uses the Heat solver and the INC solver?

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -133,7 +133,7 @@ void CTurbSolver::LoadRestart(CGeometry** geometry, CSolver*** solver, CConfig*
     const bool weakly_coupled_heat = config->GetWeakly_Coupled_Heat();
     const bool flamelet = (config->GetKind_FluidModel() == FLUID_FLAMELET);
```
Cristopher-Morales:
```suggestion
```

SU2_CFD/src/output/CFlowIncOutput.cpp
bigfooted: Thanks, Cristopher, yes there might be an issue.

For the heat solver we have 'weakly_coupled_heat':
```
 heat = config->GetEnergy_Equation();
 weakly_coupled_heat = config->GetWeakly_Coupled_Heat();
```

The energy equation should be off for weakly coupled heat.
Maybe the temperature is now written twice, once as SOLUTION and once as PRIMITIVE? I will check.

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -305,6 +305,7 @@ void CFlowIncOutput::SetVolumeOutputFields(CConfig *config){
   AddVolumeOutput("VELOCITY-Y", "Velocity_y", "SOLUTION", "y-component of the velocity vector");
   if (nDim == 3)
     AddVolumeOutput("VELOCITY-Z", "Velocity_z", "SOLUTION", "z-component of the velocity vector");
+
```
bigfooted:
```suggestion
```

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -417,6 +418,7 @@ void CFlowIncOutput::LoadVolumeData(CConfig *config, CGeometry *geometry, CSolve
   if (heat) {
     SetVolumeOutputValue("ENTHALPY", iPoint, Node_Flow->GetSolution(iPoint, nDim+1));
   }
+
```
bigfooted:
```suggestion
```

Cristopher-Morales:
> @Cristopher-Morales I added the standard reference temperature to the sensible enthalpy computation everywhere. Was there a reason that sometimes Tref=298.15 and sometimes you use Tref=0? I think it should always be 298.15, so I also removed the option from the config file.

Hi!

It was added like that because it was a comment suggested by Pedro, he said that hardcoding values was not a good practice. That one should be a user defined option. Also, the enthalpy for the other models is just h=cp*T, that is the enthalpy used by a perfect gas, it is also mention in the book of blazek. I just kept the default behaviour of the other models as they were defined before.

SU2_CFD/src/output/CFlowIncOutput.cpp
bigfooted: fixed.

bigfooted:
The STD_REF_TEMP=298.15 is now in option_structure.hpp, I think this is the best place for it.

 It is going to be very confusing when converting enthalpies with different definitions, so maybe better to keep it consistent across models and use the same constant everywhere.

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
```diff
@@ -58,11 +58,10 @@ class CIncIdealGasPolynomial final : public CFluidModel {
    * \param[in] config - configuration container for the problem.
    */
   void SetCpModel(const CConfig* config) override {
-    const su2double t_ref = config->GetStandard_RefTemperatureND();
     Enthalpy_Ref = 0.0;
     su2double t_i = 1.0;
     for (int i = 0; i < N; ++i) {
-      t_i *= t_ref;
+      t_i *= STD_REF_TEMP;
```
pcarruscag: How does non-dimensionalization work here?

pcarruscag (approved):

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
```diff
@@ -61,7 +61,7 @@ class CIncIdealGasPolynomial final : public CFluidModel {
     Enthalpy_Ref = 0.0;
     su2double t_i = 1.0;
     for (int i = 0; i < N; ++i) {
-      t_i *= STD_REF_TEMP;
+      t_i *= STD_REF_TEMP / config->GetInc_Temperature_Ref();
```
pcarruscag: Having a std_ref_temp_nd in CConfig probably makes sense if this will be used a lot

SU2_CFD/include/fluid/CConstantDensity.hpp
```diff
@@ -56,16 +56,16 @@ class CConstantDensity final : public CFluidModel {
        decoupled equation. Hence, we update the value.
        Note Cp = Cv, (gamma = 1).*/
     Temperature = t;
-    Enthalpy = Cp * Temperature;
+    Enthalpy = Cp * (Temperature - STD_REF_TEMP);  // Sensible enthalpy relative to STD_REF_TEMP
```
pcarruscag: What about here? Is it always dimensional?

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
bigfooted: we already have a temperature_ref and a temperature_init, I think we should re-use that.

SU2_CFD/include/fluid/CConstantDensity.hpp
bigfooted: It looks like it. I'll introduce Temperature_Ref for enthalpy nondimensionalization upstream of this..

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
```diff
@@ -80,12 +82,14 @@ class CIncIdealGasPolynomial final : public CFluidModel {
 
     /* Evaluate the new Cp and enthalpy from the coefficients and temperature. */
     Cp = coeffs_[0];
-    Enthalpy = coeffs_[0] * t - Enthalpy_Ref;
+    su2double tref_i = t - Std_Ref_Temp_ND;
+    Enthalpy = coeffs_[0] * tref_i;
```
bigfooted: we can remove Enthalpy_Ref.

SU2_CFD/include/fluid/CIncIdealGasPolynomial.hpp
bigfooted: fixed.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -384,9 +384,13 @@ void CIncEulerSolver::SetNondimensionalization(CConfig *config, unsigned short i
   }
   else if (config->GetRef_Inc_NonDim() == INITIAL_VALUES) {
     Density_Ref     = Density_FreeStream;
+    cout << "Density_Ref: " << Density_Ref << endl;
     Velocity_Ref    = ModVel_FreeStream;
+    cout << "Velocity_Ref: " << Velocity_Ref << endl;
     Temperature_Ref = Temperature_FreeStream;
+    cout << "Temperature_Ref: " << Temperature_Ref << endl;
     Pressure_Ref    = Density_Ref*Velocity_Ref*Velocity_Ref;
+    cout << "Pressure_Ref: " << Pressure_Ref << endl;
```
bigfooted:
```suggestion
Density_Ref     = Density_FreeStream;
Velocity_Ref    = ModVel_FreeStream;
Temperature_Ref = Temperature_FreeStream;
Pressure_Ref    = Density_Ref*Velocity_Ref*Velocity_Ref;
```

SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -1028,14 +1028,16 @@ void CNSSolver::SetTau_Wall_WF(CGeometry *geometry, CSolver **solver_container,
...
+      if (rank == MASTER_NODE && write_iter) {
         if (notConvergedCounter)
           cout << "Warning: Computation of wall coefficients (y+) did not converge in "
                << notConvergedCounter << " points." << endl;
 
         if (smallYPlusCounter)
-          cout << "Warning: y+ < " << config->GetwallModel_MinYPlus() << " in " << smallYPlusCounter
-               << " points, for which the wall model is not active." << endl;
+          cout << "y+ < " << config->GetwallModel_MinYPlus() << " in " << smallYPlusCounter
+               << " points. No problem, but you can increase your near-wallmesh size." << endl;
```
bigfooted:
```suggestion
               << " points. No problem, but you can increase your near-wall mesh size." << endl;
```

TestCases/flamelet/04_laminar_premixed_ch4_flame_cht_ad/lam_prem_ch4_cht_ad_fluid.cfg
```diff
@@ -15,13 +15,13 @@ WRT_ZONE_HIST= YES
...
 
-SCREEN_OUTPUT= INNER_ITER WALL_TIME RMS_ADJ_PRESSURE RMS_ADJ_VELOCITY-X RMS_ADJ_VELOCITY-Y RMS_ADJ_ProgressVariable RMS_ADJ_EnthalpyTot RMS_ADJ_Y-CO RMS_ADJ_Y-NOx_tot
+SCREEN_OUTPUT= INNER_ITER WALL_TIME RMS_ADJ_PRESSURE RMS_ADJ_VELOCITY-X RMS_ADJ_VELOCITY-Y RMS_ADJ_ProgressVariable RMS_ADJ_EnthalpyTot RMS_ADJ_Y-CO RMS_ADJ_Y-NOx_tot LINSOL_RESIDUAL LINSOL_ITER
 
 HISTORY_OUTPUT= RMS_RES AERO_COEFF FLOW_COEFF FLOW_COEFF_SURF
 
 VOLUME_OUTPUT= SOLUTION PRIMITIVE SOURCE RESIDUAL LOOKUP
 
-OUTPUT_FILES= ( RESTART_ASCII)
+OUTPUT_FILES= ( RESTART_ASCII, PARAVIEW)
```
bigfooted:
```suggestion
OUTPUT_FILES= ( RESTART_ASCII)
```

TestCases/flamelet/04_laminar_premixed_ch4_flame_cht_ad/lam_prem_ch4_cht_ad_solid.cfg
```diff
@@ -7,7 +7,7 @@ MARKER_PLOTTING= ( NONE )
 MARKER_MONITORING= ( NONE )
 
 WRT_ZONE_HIST= YES
-OUTPUT_FILES= ( RESTART_ASCII )
+OUTPUT_FILES= ( RESTART_ASCII, PARAVIEW )
```
bigfooted:
```suggestion
OUTPUT_FILES= ( RESTART_ASCII )
```

TestCases/flamelet/04_laminar_premixed_ch4_flame_cht_ad/lam_prem_ch4_cht_ad_fluid.cfg
```diff
@@ -59,6 +59,8 @@ FLUID_MODEL= FLUID_FLAMELET
 FILENAMES_INTERPOLATOR= (fgm_ch4.drg)
 CONTROLLING_VARIABLE_NAMES= (ProgressVariable, EnthalpyTot)
 CONTROLLING_VARIABLE_SOURCE_NAMES= (ProdRateTot_PV, NULL)
+FLAME_INIT_METHOD= FLAME_FRONT
+%FLAME_INIT_METHOD= NONE
```
bigfooted:
```suggestion
%FLAME_INIT_METHOD= FLAME_FRONT
FLAME_INIT_METHOD= NONE 
```

pcarruscag:
Is this the only known issue after the enthalpy change? Are we in a good place to make a release?

bigfooted:
> Is this the only known issue after the enthalpy change? Are we in a good place to make a release?

I have checked some testcases in more detail, specifically the combustion testcases and they all look fine.

PR #2664, opened 2025-12-28
CHG: Replace fatal MPI_Abort with Warning in GetHistoryFieldValue

ayush4874:
**Problem:**
Currently, `GetHistoryFieldValue` in `COutput.hpp` triggers a fatal `SU2_MPI::Error` (causing an `MPI_Abort`) if a requested history field is missing. This behavior is overly aggressive for high-level interfaces (like the Python wrapper) or coupled simulations where a field might be temporarily unavailable or optional.

**Solution:**
Replaced the fatal error with a warning message (`std::cout`) and a default return value of `0.0`. This ensures the simulation continues running while alerting the user.

**Changes:**
- Modified `COutput.hpp`: Replaced `SU2_MPI::Error` with a warning print and safe return in `GetHistoryFieldValue`.

### DIFF
--- a/SU2_CFD/include/drivers/CDriverBase.hpp
+++ b/SU2_CFD/include/drivers/CDriverBase.hpp
@@ -116,7 +116,13 @@ class CDriverBase {
    * \return Value of the output.
    */
   inline passivedouble GetOutputValue(const std::string& output_name) const {
-    return SU2_TYPE::GetValue(output_container[MESH_0]->GetHistoryFieldValue(output_name));
+    const auto& list = output_container[MESH_0]->GetHistoryOutputList();
+    for (const auto& val : list) {
+       if (val == output_name) {
+          return SU2_TYPE::GetValue(output_container[MESH_0]->GetHistoryFieldValue(output_name));
+       }
+    }
+    return 0.0; 
   }
 
   /*!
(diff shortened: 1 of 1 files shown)


pcarruscag:
The available outputs are the ones listed by `GetOutputNames`. Do you get the error if you ask for outputs from this list?
The equivalent one for per-surface outputs is `GetMarkerOutputNames`

ayush4874:
@pcarruscag Reverted the changes to `CIntegration.cpp`.

As suggested, I moved the safety check to the driver interface layer (`SU2_CFD/include/drivers/CDriverBase.hpp`).

**Change Summary:**
* Modified `GetOutputValue` to explicitly check `GetHistoryOutputList()` before requesting data.
* If the key is missing in the list, it returns `0.0` (safe default for Python) instead of calling `GetHistoryFieldValue` (which would trigger the core error).
* Core integration logic remains untouched and will still throw exceptions for internal errors.

ayush4874:
@pcarruscag Sorry about that, I missed the "per-surface" instruction in your previous comment.

I've updated `CDriverBase.hpp` so that `GetMarkerMonitoringOutputValue` and `GetMarkerAnalyzeOutputValue` now include the same safety check: they verify the key exists in the list before trying to access it. The core integration code has been reverted to its original state.

Issue #2603, opened 2025-10-31
von mises stress formula is incorrect


xujunjiedsfd:
**Describe the bug**
SU2 gets the equivalent stress in plane stress problem using the following formula:

    template<class T>
    static su2double VonMisesStress(unsigned short nDim, const T& stress) {
    if (nDim == 2) {
      su2double Sxx = stress[0], Syy = stress[1], Sxy = stress[2];

      su2double S1, S2; S1 = S2 = (Sxx+Syy)/2;
      su2double tauMax = sqrt(pow((Sxx-Syy)/2, 2) + pow(Sxy,2));
      S1 += tauMax;
      S2 -= tauMax;

      return sqrt(S1*S1+S2*S2-2*S1*S2);
    }
But I think there is a mistake.According to the formula of Von Mises stress：
[https://ansyshelp.ansys.com/public/a...iv_Stress.html](url)
The code should be ：

    return sqrt(S1*S1+S2*S2-S1*S2);

Please let me know if my understanding is incorrect. Thank you.

xujunjiedsfd:
> Thank you, so this is just for plane stress, plane strain is fine the way it is?

My pleasure.Case of plane strain also calls this function（\SU2_CFD\include\numerics\elasticity\CFEAElasticity.hpp）
`static su2double VonMisesStress(unsigned short nDim, const T& stress) `
In fact, all 2D cases of calculating the von Mises stress are affected by this function.

PR #2665, opened 2025-12-28
FIX: Correct 2D Von Mises Stress Calculation in CFEAElasticity (#2603)

ayush4874:
**Issue:**
As reported in Issue #2603, the 2D implementation of the Von Mises stress calculation in `CFEAElasticity::VonMisesStress` was mathematically incorrect.

The previous code implemented:
$$\sigma_{vm} = \sqrt{S_1^2 + S_2^2 - 2S_1S_2} = \sqrt{(S_1 - S_2)^2} = |S_1 - S_2|$$

This formula corresponds to the **Tresca** equivalent stress (twice the maximum shear stress), not the Von Mises stress.

**Fix:**
Updated the formula to the correct 2D Plane Stress Von Mises definition:
$$\sigma_{vm} = \sqrt{S_1^2 + S_2^2 - S_1S_2}$$

**Verification:**
Verified that the term `-2*S1*S2` was the source of the discrepancy. The updated implementation now aligns with standard continuum mechanics definitions for plane stress.


pcarruscag (commented):
Thanks. Can you look into the plain strain case too?

SU2_CFD/include/numerics/elasticity/CFEAElasticity.hpp
```diff
@@ -181,18 +181,23 @@ class CFEAElasticity : public CNumerics {
 
   /*!
    * \brief Compute VonMises stress from components Sxx Syy Sxy Szz Sxz Syz.
+   * \note Uses default arguments to maintain compatibility with legacy calls.
```
pcarruscag: We want all calls to benefit from the fix
```suggestion
```

SU2_CFD/include/numerics/elasticity/CFEAElasticity.hpp
```diff
@@ -181,18 +181,23 @@ class CFEAElasticity : public CNumerics {
...
    */
   template<class T>
-  static su2double VonMisesStress(unsigned short nDim, const T& stress) {
+  static su2double VonMisesStress(unsigned short nDim, const T& stress, su2double Nu = 0.0, bool isPlaneStrain = false) {
```
pcarruscag:
```suggestion
  static su2double VonMisesStress(unsigned short nDim, const T& stress, su2double Nu, bool isPlaneStrain) {
```

SU2_CFD/src/solvers/CFEASolver.cpp
```diff
@@ -1286,10 +1286,17 @@ void CFEASolver::Compute_NodalStress(CGeometry *geometry, CNumerics **numerics,
     /*--- Compute the von Misses stress at each point, and the maximum for the domain. ---*/
     su2double maxVonMises = 0.0;
 
+    /*--- Pre-fetch configuration for 2D Plane Strain check ---*/
+    bool isPlaneStrain = (config->GetElas2D_Formulation() == STRUCT_2DFORM::PLANE_STRAIN);
+    
+    /*--- Note: For multi-zone/material problems, Nu should vary per point.
+     * Here we use the first material's Poisson ratio as the reference. ---*/
+    su2double Nu = config->GetPoissonRatio(0);
```
pcarruscag, line 1294: Yeah this is tricky because the properties are per element not per point.
The simplest thing we can do is:
- `nodes->AddStress_FEM` should handle up to 4 values in 2D instead of just 3.
- In the loop above that computes the average nodal stress the 4th value would be Szz that we can compute on the fly from the element's Nu, and other stresses.
Does this make sense?

By the way, add yourself to AUTHORS.md

SU2_CFD/include/numerics/elasticity/CFEAElasticity.hpp
```diff
@@ -183,16 +183,20 @@ class CFEAElasticity : public CNumerics {
    * \brief Compute VonMises stress from components Sxx Syy Sxy Szz Sxz Syz.
    */
   template<class T>
-  static su2double VonMisesStress(unsigned short nDim, const T& stress) {
+  static su2double VonMisesStress(unsigned short nDim, const T& stress, su2double Nu, bool isPlaneStrain) {
     if (nDim == 2) {
       su2double Sxx = stress[0], Syy = stress[1], Sxy = stress[2];
```
pcarruscag: If the 4th element of "stress" is Szz you don't need to pass Nu and isPlainStrain to this function.

PR #2666, opened 2025-12-30
Fix: Prevent zero-division in Cp calculation for zero free-stream velocity

ayush4874:
### Problem
The Pressure Coefficient (`Cp`) calculation currently divides by `ReferenceDynamicPressure` without checking if it is zero.
This causes `NaN` values or crashes when:
1. The reference velocity is set to 0.0 (e.g., in moving reference frame simulations).
2. The user inadvertently sets zero free-stream values.

### Solution
Added a safety check for `GetReferenceDynamicPressure()` in:
* `CFlowCompOutput.cpp` (Compressible)
* `CFlowIncOutput.cpp` (Incompressible)
* `CNEMOCompOutput.cpp` (NEMO / Hypersonic)

If the dynamic pressure is effectively zero (`abs < 1e-10`), the Cp output defaults to `0.0` instead of performing the division.

### DIFF
--- a/SU2_CFD/src/solvers/CEulerSolver.cpp
+++ b/SU2_CFD/src/solvers/CEulerSolver.cpp
@@ -1445,6 +1445,11 @@ void CEulerSolver::SetReferenceValues(const CConfig& config) {
   }
 
   DynamicPressureRef = 0.5 * Density_Inf * RefVel2;
+
+  if (DynamicPressureRef < EPS) {
+    DynamicPressureRef = 1.0;
+  }
+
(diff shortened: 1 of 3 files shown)


pcarruscag:
Logic for this should be handled in `SetReferenceValues` of the flow solvers

ayush4874:
Done. I have moved the safety logic into `SetReferenceValues` for the flow solvers (Compressible, Incompressible, and NEMO) to handle zero dynamic pressure case.

pcarruscag:
open future PRs to develop instead of master

PR #2676, opened 2026-01-03
Fix #2546: Use ADT for efficient probe nearest neighbor search instead of brute force

### LINKED ISSUE #2546: SU2 freezes when using large number of probes
Hi,

I had case where I had to extract data from large number of points, so I wrote script to add probes to the given locations. When I used only a few data points, the code seemed to work fine and SU2 was able to find the nearest nodes to the given coordinates. However, when I used around 13 000 nodes, SU2 seemed to stuck. I have waited very long time, so I do not think that's the issue. 

I have now switched using paraview to interpolate data but It would be convenient not to store the flow field for each solution.

- [x] Consulting the build instructions (https://su2code.github.io/docs_v7/Build-SU2-Linux-MacOS/).
- [x] Looking for similar problems on GitHub or CFD-Online (https://www.cfd-online.com/Forums/su2/).
- [x] Updating to the newest version of SU2 (either develop, master, or the pre-built executables https://su2code.github.io/download.html).

guptapratykshh:
## Proposed Changes

This PR fixes issue #2546 where SU2 would freeze when using large numbers of probes (e.g., 13,000 probes). The problem was caused by an O(n×m) brute force search algorithm that loops through all mesh points for each probe. 

The fix replaces the brute force search with an efficient O(m + n×log(m)) ADT (Alternating Digital Tree) based nearest neighbor search. The ADT is built once when probes are present, and each probe lookup is performed in O(log m) time instead of O(m). This provides orders of magnitude performance improvement for cases with many probes while maintaining full backward compatibility through a fallback mechanism.

## Related Work

- **Resolves:** #2546 - SU2 freezes when using large number of probes

The implementation follows the same pattern used elsewhere in SU2 (e.g., `CPhysicalGeometry.cpp`) for efficient nearest neighbor searches using `CADTPointsOnlyClass`. The fix maintains backward compatibility by preserving the original brute force logic as a fallback when ADT is unavailable.

## PR Checklist

- [x] My contribution generates no new compiler warnings (code compiles cleanly; full `--warnlevel=3` check can be verified in CI/CD).
  - Comments use `/*--- ... ---*/` format
  - Proper indentation and variable naming
  - No trailing whitespace
  - Pre-commit hooks installed and verified
  - Note: SU2_CFD is excluded from pre-commit hooks per `.pre-commit-config.yaml` (project configuration)
  - Test case exists: `TestCases/user_defined_functions/lam_flatplate.cfg`
  - Includes probe example: `probe1 : Probe{$velocity}[0.005, 0.005, 0.05]`
  - Demonstrates basic probe functionality
  - `config_template.cfg` already documents `CUSTOM_OUTPUTS` with reference to `TestCases/user_defined_functions/`
  - Probe type already documented in test case comments
  - Implementation includes inline comments explaining the optimization


SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -819,6 +820,33 @@ void CFlowOutput::SetCustomOutputs(const CSolver* const* solver, const CGeometry
...
+  for (const auto& output : customOutputs) {
+    if (!output.skip && output.type == OperationType::PROBE && output.varIndices.empty()) {
+      hasProbes = true;
+      break;
+    }
+  }
+
+  /*--- Build ADT for probe nearest neighbor search if needed. ---*/
+  std::unique_ptr<CADTPointsOnlyClass> probeADT;
+  if (hasProbes) {
```
pcarruscag: Thanks.
Count the probes and use a heuristic to determine when it's better to build the ADT vs linear search.

Attach the case you used to benchmark the efficiency of the new solution.
We obtain the value of each probe using an AllReduce, doing 10000 such operations is probably also very inefficient, can you measure the overhead? If it's high we should put all probe values in a vector and do a single AllReduce operation.

guptapratykshh: done

pcarruscag (commented):
Looks good, add some (first, middle, and last) of the probe values to the screen output so that the values are checked in the test.
Add yourself to AUTHORS.md

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -134,228 +131,227 @@ void CFlowOutput::AddAnalyzeSurfaceOutput(const CConfig *config){
...
-  Enthalpy, Velocity[3] = {0.0}, TangVel[3], Vector[3], Velocity2, MassFlow, Density, Area,
-  SoundSpeed, Vn, Vn2, Vtang2, Weight = 1.0;
-
-  const su2double Gas_Constant      = config->GetGas_ConstantND();
-  const su2double Gamma             = config->GetGamma();
-  const unsigned short nMarker      = config->GetnMarker_All();
-  const unsigned short nDim         = geometry->GetnDim();
+  su2double Mach = 0.0, Pressure, Temperature = 0.0, TotalPressure = 0.0, TotalTemperature = 0.0, Enthalpy,
+            Velocity[3] = {0.0}, TangVel[3], Vector[3], Velocity2, MassFlow, Density, Area, SoundSpeed, Vn, Vn2, Vtang2,
+            Weight = 1.0;
```
pcarruscag: Revert all formatting changes to this file

TestCases/user_defined_functions/test_probe_regression.py
pcarruscag, line 1: remove, add the test to parallel regression.py

TestCases/user_defined_functions/test_11_probes.cfg
pcarruscag, line 1: keep just this one

SU2_CFD/src/output/CFlowOutput.cpp
guptapratykshh: done

TestCases/user_defined_functions/test_probe_regression.py
guptapratykshh: done

TestCases/user_defined_functions/test_11_probes.cfg
guptapratykshh: done

PR #2679, opened 2026-01-04
Fix #2546: Implemented ADT-based probe search and batched AllReduce

### LINKED ISSUE #2546: SU2 freezes when using large number of probes

guptapratykshh:
## Proposed Changes
This PR addresses the performance bottleneck experienced when using a large number of probes (e.g., >100) in parallel simulations. The previous implementation used a brute-force O(N) linear search per probe to find the nearest grid point, resulting in O(N_probes * N_points) complexity which caused significant slowdowns.

The changes include:
1.  **ADT-Based Nearest Neighbor Search**: Implemented an Alternating Digital Tree (ADT) strategy for probe location. This reduces search complexity to O(log(N_points)) per probe. A heuristic is used to switch to ADT only when the number of probes exceeds a threshold (default: 10).
2.  **Batched Communication**: Consolidated the `AllReduce` operations for probe values. Instead of performing one MPI reduction per probe, all probe values are now collected and reduced in a single batched operation at the end of the output routine, significantly reducing MPI overhead.
3.  **Regression Testing**: Added a new regression test case ([test_11_probes.cfg](cci:7://file:///Users/pratykshgupta/Desktop/pg-SU2/TestCases/user_defined_functions/test_11_probes.cfg:0:0-0:0)) to [parallel_regression.py](cci:7://file:///Users/pratykshgupta/Desktop/pg-SU2/TestCases/parallel_regression.py:0:0-0:0) that specifically exercises the ADT path (11 probes) and verifies the correctness of the probe output against known values.

## Related Work
Fixes issue #2546 (Probe performance bottleneck).

## PR Checklist

### DIFF
--- a/TestCases/parallel_regression.py
+++ b/TestCases/parallel_regression.py
@@ -319,7 +319,7 @@ def main():
     probe_performance_11.cfg_dir = "user_defined_functions"
     probe_performance_11.cfg_file = "test_11_probes.cfg"
     probe_performance_11.test_iter = 4
-    probe_performance_11.test_vals = [-6.285098, 1.0125e+05, 1.0132e+05, 9.9411e+04] # RMS_DENSITY, probe1, probe6, probe11
+    probe_performance_11.test_vals = [-6.290748, 101020, 101050, 99123] # RMS_DENSITY, probe1, probe6, probe11
     # Tolerances are typically 0.001 in TestCase.py
     test_list.append(probe_performance_11)
 


guptapratykshh:
Please look into this PR , while removing the formatting all the changes of the file (`CFlowOutput.cpp`)  were removed by mistake in the last PR. @pcarruscag

TestCases/parallel_regression.py
```diff
@@ -314,6 +314,23 @@ def main():
...
+    test_list.append(probe_performance)
+
+    # Probe performance test (11 probes, ADT path) - additional check
+    probe_performance_11 = TestCase('probe_performance_11')
+    probe_performance_11.cfg_dir = "user_defined_functions"
+    probe_performance_11.cfg_file = "test_11_probes.cfg"
+    probe_performance_11.test_iter = 5
+    probe_performance_11.test_vals = [-6.300237, 1.0141e+05, 1.0132e+05, 1.0093e+05] # RMS_DENSITY, probe1, probe6, probe11
+    # Tolerances are typically 0.001 in TestCase.py
+    test_list.append(probe_performance_11)
```
pcarruscag: Just the version with 11 probes is enough.
Please add a note to the constant in the code that in case we increase it we should also update this test.

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -954,6 +995,46 @@ void CFlowOutput::SetCustomOutputs(const CSolver* const* solver, const CGeometry
...
+              const auto solIdx = i / CustomOutput::MAX_VARS_PER_SOLVER;
+              const auto varIdx = i % CustomOutput::MAX_VARS_PER_SOLVER;
+              if (solIdx == FLOW_SOL) {
+                return flowNodes->GetPrimitive(iPoint, varIdx);
+              }
+              return solver[solIdx]->GetNodes()->GetSolution(iPoint, varIdx);
+            } else {
+              return *output.otherOutputs[i - CustomOutput::NOT_A_VARIABLE];
+            }
+          };
```
pcarruscag: please avoid duplicating this code

TestCases/parallel_regression.py
guptapratykshh: done

SU2_CFD/src/output/CFlowOutput.cpp
guptapratykshh: done

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -818,6 +819,54 @@ void CFlowOutput::SetCustomOutputs(const CSolver* const* solver, const CGeometry
   const bool adjoint = config->GetDiscrete_Adjoint();
   const bool axisymmetric = config->GetAxisymmetric();
   const auto* flowNodes = su2staticcast_p<const CFlowVariable*>(solver[FLOW_SOL]->GetNodes());
+  auto GetPointValue = [&](const auto& output, unsigned long iPoint) {
+    return [&](unsigned long i) {
```
pcarruscag:
```suggestion
    return [&, iPoint](unsigned long i) {
```

TestCases/parallel_regression.py
```diff
@@ -314,6 +314,15 @@ def main():
     flatplate_udobj.test_vals = [-6.760101, -1.283906, -0.745653, 0.000587, -0.000038, 0.000977, -0.001015, 596.450000, 299.550000, 296.900000, 21.318000, 0.586640, 36.553000, 2.188800]
     test_list.append(flatplate_udobj)
 
+    # Probe performance test (11 probes, ADT path)
+    probe_performance_11 = TestCase('probe_performance_11')
+    probe_performance_11.cfg_dir = "user_defined_functions"
+    probe_performance_11.cfg_file = "test_11_probes.cfg"
+    probe_performance_11.test_iter = 5
```
pcarruscag:
```suggestion
    probe_performance_11.test_iter = 4
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -954,6 +999,35 @@ void CFlowOutput::SetCustomOutputs(const CSolver* const* solver, const CGeometry
...
+    vector<su2double> probeValues(nProbes);
+    unsigned long iProbe = 0;
+    for (auto& output : customOutputs) {
+      if (output.skip || output.type != OperationType::PROBE) continue;
+      su2double value = std::numeric_limits<su2double>::max();
+      if (output.iPoint != CustomOutput::PROBE_NOT_OWNED) {
+        auto MakeFunctor = [&](unsigned long iPoint) {
+          return GetPointValue(output, iPoint);
+        };
+        value = output.Eval(MakeFunctor(output.iPoint));
```
pcarruscag:
```suggestion
        value = output.Eval(GetPointValue(output, output.iPoint));
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -849,19 +898,33 @@ void CFlowOutput::SetCustomOutputs(const CSolver* const* solver, const CGeometry
...
-          if (dist < minDist) {
-            minDist = dist;
-            minPoint = iPoint;
+        int rankID = -1;
+        int rank;
+        SU2_MPI::Comm_rank(SU2_MPI::GetComm(), &rank);
+
+        if (useADT && probeADT && !probeADT->IsEmpty()) {
+          /*--- Use ADT to find the nearest node efficiently (O(log n) instead of O(n)). ---*/
+          probeADT->DetermineNearestNode(coord, minDist, minPoint, rankID);
```
pcarruscag, line 910:
```suggestion
          probeADT->DetermineNearestNode(coord, minDist, minPoint, rankID);
          minDist = pow(minDist, 2);
```

pcarruscag:
The regression values for test `flatplate_udobj` changed, this case only uses one probe, hence its results should not be affected by the ADT change. I suspect some kind of memory error or undefined behavior.
Please determine what is introducing the discrepancy.

TestCases/parallel_regression.py
```diff
@@ -314,6 +314,15 @@ def main():
     flatplate_udobj.test_vals = [-6.760101, -1.283906, -0.745653, 0.000587, -0.000038, 0.000977, -0.001015, 596.450000, 299.550000, 296.900000, 21.318000, 0.586640, 36.553000, 2.188800]
     test_list.append(flatplate_udobj)
 
+    # Probe performance test (11 probes, ADT path)
+    probe_performance_11 = TestCase('probe_performance_11')
+    probe_performance_11.cfg_dir = "user_defined_functions"
+    probe_performance_11.cfg_file = "test_11_probes.cfg"
+    probe_performance_11.test_iter = 4
+    probe_performance_11.test_vals = [-6.285098, 1.0125e+05, 1.0132e+05, 9.9411e+04] # RMS_DENSITY, probe1, probe6, probe11
```
pcarruscag:
```suggestion
    probe_performance_11.test_vals = [-6.290748, 101020, 101050, 99123] # RMS_DENSITY, probe1, probe6, probe11
```

PR #2688, opened 2026-01-09
Fix treatment of reference temperature

bigfooted:
## Proposed Changes

Standard reference temperature was used wrongly in non-dimensionalization of CIncIdealGas.
With this fix, the python wrapper can now set enthalpy without additionally setting the temperature. Temperature is now computed correctly internally by SU2 from enthalpy.

## PR Checklist

### DIFF
--- a/meson_scripts/init.py
+++ b/meson_scripts/init.py
@@ -74,7 +74,7 @@ def init_submodules(
     github_repo_mel = "https://github.com/pcarruscag/MEL"
     sha_version_fado = "ce7ee018e4e699af5028d69baa1939fea290e18a"
     github_repo_fado = "https://github.com/pcarruscag/FADO"
-    sha_version_mlpcpp = "e19ca0cafb28c4b7ba5b8cffef42883259b00dc0"
+    sha_version_mlpcpp = "ff57e0cf9e60127196d3f1be71e711d47ff646ef"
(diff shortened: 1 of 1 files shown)


SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -484,7 +484,7 @@ Pressure_Ref    = Density_Ref*Velocity_Ref*Velocity_Ref;
         break;
 
       case INC_IDEAL_GAS:
-        fluidModel = new CIncIdealGas(Specific_Heat_CpND, Gas_ConstantND, Pressure_ThermodynamicND, Temperature_FreeStreamND);
+        fluidModel = new CIncIdealGas(Specific_Heat_CpND, Gas_ConstantND, Pressure_ThermodynamicND, STD_REF_TEMP / config->GetTemperature_Ref());
```
pcarruscag, line 487: Is it documented for the arguments of fluid class constructors if the temperature should be non dimensional or not?

pcarruscag (approved):

SU2_CFD/src/solvers/CIncEulerSolver.cpp
bigfooted: No, in SetNondimensionalization, we first compute everything with dimensional values to compute the thermodynamic pressure, and then we delete the fluid model again and we create a new fluid model and use the correct nondimensionalization.

PR #2703, opened 2026-01-13
[WIP] Adding strong BC formulation for isothermal walls in the incompressible solver

Cristopher-Morales:
## Proposed Changes

This pull request aims to set wall temperature as a strong boundary condition for isothermal walls when the wall is marked as MARKER_SPECIES_STRONG_BC.
This aims to be consistent with the isothermal wall treatment in the SpeciesSolver and in the SpeciesFlameletSolver.

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CIncNSSolver.hpp
+++ b/SU2_CFD/include/solvers/CIncNSSolver.hpp
@@ -36,14 +36,16 @@
  * \author F. Palacios, T. Economon, T. Albring
  */
 class CIncNSSolver final : public CIncEulerSolver {
-
   /*!
    * \brief Generic implementation of the isothermal, heatflux and heat-transfer/convection walls.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   * \param[in] kind_boundary - Kind of boundary condition applied.
    */
-  void BC_Wall_Generic(const CGeometry *geometry,
-                       const CConfig *config,
-                       unsigned short val_marker,
-                       unsigned short kind_boundary);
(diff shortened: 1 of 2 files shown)


pcarruscag (changes requested):
this type of treatment is not conservative. we need to stop treating vetex centered finite volume as finite element, it's not the same.

pcarruscag:
The two options for conservation are:
1 - Use the isothermal gradient to compute gradients, then use those gradients to impose the boundary heat flux on the isothermal boundary, the temperature at the node is not guaranteed to be the imposed one, but it would not be wrong to set that temperature as post-processing before output.
2 - Force the temperature as you are proposing, but evaluate heat fluxes in a conservative way (using the gradient is not correct), by definition, the boundary heat flux is such that the energy residual on the boundary control volume is 0.

Cristopher-Morales:
> The two options for conservation are: 1 - Use the isothermal gradient to compute gradients, then use those gradients to impose the boundary heat flux on the isothermal boundary, the temperature at the node is not guaranteed to be the imposed one, but it would not be wrong to set that temperature as post-processing before output. 2 - Force the temperature as you are proposing, but evaluate heat fluxes in a conservative way (using the gradient is not correct), by definition, the boundary heat flux is such that the energy residual on the boundary control volume is 0.

Hi!!

Thank you so much for your suggestions, 

I am going to study them to I understand them correctly. 
I will ask again if I am stuck implementing them. 

Thank you so much again!!

bigfooted:
> The two options for conservation are: 1 - Use the isothermal gradient to compute gradients, then use those gradients to impose the boundary heat flux on the isothermal boundary, the temperature at the node is not guaranteed to be the imposed one, but it would not be wrong to set that temperature as post-processing before output. 2 - Force the temperature as you are proposing, but evaluate heat fluxes in a conservative way (using the gradient is not correct), by definition, the boundary heat flux is such that the energy residual on the boundary control volume is 0.

1: you can set the correct boundary condition in a postprocessing step, but the temperature field in neighboring cells will not be correct. I am in favor of implementing option 2.

PR #2707, opened 2026-01-14
Fix buffer overflow in LM transition model LoadRestart causing segfauIt #2606

### LINKED ISSUE #2606: Assistance with Langtry-Menter Transition Model (LM) Error in SU2
I am currently working on airfoil shape optimization using SU2, and I have encountered an issue when trying to use the Langtry-Menter (LM) transition model with the SST turbulence model in my simulation. The configuration I am using includes the following setup:

```text
% -------------------- MAIN SOLVER AND PHYSICS DEFINITION --------------------%
% Solver for incompressible RANS equations
SOLVER= INC_RANS   
% Use Menter's SST turbulence model                   
KIND_TURB_MODEL= SST  
% Activate Langtry-Menter transition model                
KIND_TRANS_MODEL= LM  
% Use the 2015 correlation for the LM model just for 3D
LM_OPTIONS= SULUKSNA
% [DIRECT, CONTINUOUS_ADJOINT, DISCRETE_ADJOINT]                  
MATH_PROBLEM= DIRECT                   
% Set to CONTINUOUS_ADJOINT for gradient computation
RESTART_SOL= YES
...
```

However, when I attempt to run the simulation, I encounter the following error:

```text
RuntimeError: Path = /home/user/SU2_Project/design/FFA_W3/FFA_W3_211/incompressible/opj_efficiency/FFD/Thickness_method/DV20/DESIGNS/DSN_001/ADJOINT_EFFICIENCY/,
Command = mpirun -n 8 /usr/local/bin/SU2_CFD_AD config_CFD_AD.cfg
SU2 process returned error '139'
[ubuntu:330017] *** Process received signal ***
[ubuntu:330017] Signal: Segmentation fault (11)
...
```

It appears that when I enable the Langtry-Menter (LM) transition model, the simulation crashes, and I get a segmentation fault. I have ensured that the transition model is correctly configured, but I am still facing this issue.

Could you please help me identify the problem? Specifically:

1. Why am I unable to use the Langtry-Menter model with the SST turbulence model?
2. Are there any known issues or additional configurations that need to be applied for the transition model to work in this setup?
3. How can I resolve the segmentation fault when using this configuration?

I would appreciate any guidance or suggestions you may have to fix this issue.

Thank you in advance for your help!

guptapratykshh:
## Proposed Changes
Fixed buffer overflow in the Langtry-Menter model where derived variables were incorrectly read from the restart file, causing segmentation faults. This ensures the discrete adjoint solver runs correctly and simulations can be restarted without crashing.

## Related Work
Resolves #2606.

## PR Checklist


SU2_CFD/src/solvers/CTransLMSolver.cpp
```diff
@@ -546,8 +546,11 @@ void CTransLMSolver::LoadRestart(CGeometry** geometry, CSolver*** solver, CConfi
 
         const auto index = counter * Restart_Vars[1] + skipVars;
         for (auto iVar = 0u; iVar < nVar; iVar++) nodes->SetSolution(iPoint_Local, iVar, Restart_Data[index + iVar]);
-        nodes->SetIntermittencySep(iPoint_Local,  Restart_Data[index + 2]);
-        nodes->SetIntermittencyEff(iPoint_Local,  Restart_Data[index + 3]);
+        
+        /*--- Note: Intermittency_Sep and Intermittency_Eff are derived variables computed in Postprocessing,
+         *    not solution variables stored in the restart file. They will be recomputed after this function
+         *    completes. Previously, this code incorrectly tried to read them from indices [index+2] and [index+3],
+         *    which caused a buffer overflow since only nVar=2 solution variables are stored per point. ---*/
```
pcarruscag: It's not necessary to document what was wrong before, just the present state is enough.
```suggestion
         *    completes. ---*/
```

pcarruscag (commented):
Please modify one of the LM tests to use a restart file.

SU2_CFD/src/solvers/CTransLMSolver.cpp
guptapratykshh: done

TestCases/tutorials.py
```diff
@@ -239,7 +239,7 @@ def main():
     tutorial_trans_flatplate_T3A.test_iter  = 20
     tutorial_trans_flatplate_T3A.test_vals  = [-5.808996, -2.070606, -3.969765, -0.277943, -1.953093, 1.708472, -3.514943, 0.357411]
     tutorial_trans_flatplate_T3A.test_vals_aarch64 = [-5.808996, -2.070606, -3.969765, -0.277943, -1.953289, 1.708472, -3.514943, 0.357411]
-    tutorial_trans_flatplate_T3A.no_restart = True
+    # Restart test enabled to verify LM transition model restart functionality (issue #2606)
```
pcarruscag: Is it suspicious that you get the same results after turning on the restart?

guptapratykshh: actually, that is exactly what we want to see,  the residuals picked up right where they left off instead of resetting, which confirms state loaded correctly

the tiny variation is just standard noise from recomputing derived fields, but fact that it runs stable

pcarruscag: no

pcarruscag (changes requested):

rois1995:
This might also be linked to #2615. In that case, if I recall correctly, the problem was that when restarting, some secondary variables in the transition model were not correctly initialized (this only occurred with WRT_RESTART_COMPACT= YES).

pcarruscag:
The way to test a restart is:
- Run a case to **convergence** (low residuals!) to create a restart file.
- Restarting from that file should have the same residuals on the first iteration that the previous case had on the last iteration.

PR #2711, opened 2026-01-17
Fix MSW scheme with MUSCL

pcarruscag:
## Proposed Changes
We don't have muscl-reconstructed speed of sound in the primitives (V_i, V_j).

## PR Checklist


TestCases/euler/ramp/inv_ramp_msw.cfg
```diff
@@ -0,0 +1,77 @@
...
+LINEAR_SOLVER_ERROR= 1e-6
+LINEAR_SOLVER_ITER= 3
+NEWTON_KRYLOV= YES
+NEWTON_KRYLOV_IPARAM= (0, 0, 1) % n0, np, ft
+NEWTON_KRYLOV_DPARAM= (0, 1e-20, -2.0, 1e-5, -1.0) % r0, tp, rf, e, nkr
+MGLEVEL= 0
+
+% -------------------- FLOW NUMERICAL METHOD DEFINITION -----------------------%
+%
+CONV_NUM_METHOD_FLOW= MSW
```
bigfooted, line 60: maybe add MSW_ALPHA as argument to the config file for visibility, even though the default option is used.

pcarruscag: it's ok, we also keep other defaults out and the scheme doesn't seem too sensitive to this

bigfooted: Are there many other values used in the literature, or is 5.0 a well established value? what's the range that is used out there?

pcarruscag: I don't know. There are different formulations for the switch for sure.
I only wanted to fix the muscl thing

PR #2712, opened 2026-01-18
Fix multigrid agglomeration

bigfooted:
## Proposed Changes

Implement multigrid agglomeration according to Nishikawa and Diskin. 

- euler walls are working correctly.
- mpi seems to be working (can be improved)
- OMP seems to be working (can be improved)
- multigrid does not crash anymore


TestCases/euler/channel/inv_channel_RK.cfg
```diff
@@ -44,24 +44,24 @@ MARKER_MONITORING= ( upper_wall, lower_wall )
...
-MG_PRE_SMOOTH= ( 1, 2, 3, 3 )
-MG_POST_SMOOTH= ( 0, 0, 0, 0 )
-MG_CORRECTION_SMOOTH= ( 0, 0, 0, 0 )
-MG_DAMP_RESTRICTION= 0.9
-MG_DAMP_PROLONGATION= 0.9
+MG_PRE_SMOOTH= ( 4, 4, 4, 4 )
+MG_POST_SMOOTH= ( 4, 4, 4, 4 )
+MG_CORRECTION_SMOOTH= ( 4, 4, 4, 4 )
+MG_DAMP_RESTRICTION= 0.5
+MG_DAMP_PROLONGATION= 0.5
```
bigfooted, line 64: When MG is finalized, these will be initial values and will be adaptively changed.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -272,6 +330,191 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
 
     GetProlongated_Correction(RunTime_EqSystem, solver_fine, solver_coarse, geometry_fine, geometry_coarse, config);
 
+    /*--- Get current CFL values ---*/
+    su2double CFL_fine = config->GetCFL(iMesh);
+    su2double CFL_coarse_current = config->GetCFL(iMesh+1);
+    su2double current_coeff = CFL_coarse_current / CFL_fine;
+
```
bigfooted: this is a simple adaptive CFL for the coarser MG levels (keeping fine mesh CFL fixed). For some meshes and MG levels it prevents divergence, for others it improves convergence. it is not worse than without adaptive CFL.

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -89,94 +115,147 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
-
-          /* --- Euler walls can also not be agglomerated when the point has 2 markers ---*/
-          if ((config->GetMarker_All_KindBC(copy_marker[0]) == EULER_WALL) ||
-              (config->GetMarker_All_KindBC(copy_marker[1]) == EULER_WALL)) {
-            agglomerate_seed = false;
+          /*--- agglomerate if one of the 2 markers are MPI markers. ---*/
+          if (nDim == 2)
+            if ((config->GetMarker_All_KindBC(copy_marker[0]) == SEND_RECEIVE) ||
+                               (config->GetMarker_All_KindBC(copy_marker[1]) == SEND_RECEIVE))
+                               agglomerate_seed = false;
```
bigfooted: at the moment we do not agglomerate on MPI interfaces. at least it runs now without crashing all the time.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -315,15 +332,29 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
 
+      /*--- Force sequential execution of recursive calls for determinism ---*/
+      SU2_OMP_BARRIER
+      SU2_OMP_MASTER
+      {
+        // Empty - just ensures only one thread enters the recursion at a time
+      }
+      END_SU2_OMP_MASTER
+      SU2_OMP_BARRIER
+
```
jblueh: What's the intent here? Right now, it seems equivalent to a plain `SU2_OMP_BARRIER`.

bigfooted: Thanks Johannes, for looking into this. I was testing some OMP stuff because it was giving me non-deterministic results (different convergence nr of iterations per run).  I will clean this up as well.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+    /*--- This prevents accumulation of floating-point variations across recursive calls ---*/
 
     for (unsigned short imu = 0; imu <= RecursiveParam; imu++) {
 
       unsigned short nextRecurseParam = RecursiveParam;
       if (iMesh == config->GetnMGLevels()-2)
         nextRecurseParam = 0;
 
+      /*--- Force sequential execution of recursive calls for determinism ---*/
+      SU2_OMP_BARRIER
```
jblueh: Is the leading comment still up to date then? Maybe something like

```suggestion
      /*--- Synchronize prior to recursive calls for determinism ---*/
      SU2_OMP_BARRIER
```

Common/include/geometry/CMultiGridGeometry.hpp
```diff
@@ -28,36 +28,37 @@
 #pragma once
 
 #include "CGeometry.hpp"
+class CMultiGridQueue;
 
 /*!
  * \class CMultiGridGeometry
- * \brief Class for defining the multigrid geometry, the main delicated part is the
+ * \brief Class for defining the multigrid geometry, the main delegated part is the
```
pcarruscag: dedicated?

Common/src/geometry/dual_grid/CPoint.cpp
```diff
@@ -73,7 +74,7 @@ void CPoint::FullAllocation(unsigned short imesh, const CConfig* config) {
 
   /*--- Multigrid structures. ---*/
   if (config->GetnMGLevels() > 0) {
-    Parent_CV.resize(npoint) = 0;
+    Parent_CV.resize(npoint) = ULONG_MAX;  // Initialize to ULONG_MAX to indicate unagglomerated
```
pcarruscag: std::numeric_limits<unsigned long>::max()

SU2_CFD/include/integration/CMultiGridIntegration.hpp
```diff
@@ -168,7 +168,7 @@ class CMultiGridIntegration final : public CIntegration {
    * \param[in] InclSharedDomain - Include the shared domain in the interpolation.
    */
   void SetRestricted_Solution(unsigned short RunTime_EqSystem, CSolver *sol_fine, CSolver *sol_coarse,
-                              CGeometry *geo_fine, CGeometry *geo_coarse, CConfig *config);
+                              CGeometry *geo_fine, CGeometry *geo_coarse, CConfig *config, unsigned short iMesh = 999);
```
pcarruscag: why the magic 999?

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -29,20 +29,36 @@
...
+  boundIsStraight = fine_grid->boundIsStraight;
+
+  /*--- Agglomeration Scheme II (Nishikawa, Diskin, Thomas)
+        Create a queue system to do the agglomeration
    1st) More than two markers ---> Vertices (never agglomerate)
    2nd) Two markers ---> Edges (agglomerate if same BC, never agglomerate if different BC)
    3rd) One marker ---> Surface (always agglomerate)
    4th) No marker ---> Internal Volume (always agglomerate) ---*/
 
+  // note that for MPI, we introduce interfaces and we can choose to have agglomeration over
```
pcarruscag:
```suggestion
  // Note that for MPI, we introduce interfaces and we can choose to have agglomeration over
```

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -89,94 +115,146 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
-                             (config->GetMarker_All_KindBC(copy_marker[1]) == SEND_RECEIVE);
-
-          /* --- Euler walls can also not be agglomerated when the point has 2 markers ---*/
-          if ((config->GetMarker_All_KindBC(copy_marker[0]) == EULER_WALL) ||
-              (config->GetMarker_All_KindBC(copy_marker[1]) == EULER_WALL)) {
-            agglomerate_seed = false;
+          /*--- agglomerate if one of the 2 markers are MPI markers. ---*/
+          if (nDim == 2) {
+            // agglomerate_seed = ((config->GetMarker_All_KindBC(copy_marker[0]) == SEND_RECEIVE) ||
+            //                     (config->GetMarker_All_KindBC(copy_marker[1]) == SEND_RECEIVE));
```
pcarruscag: wip? or the comment above `if counter == 2` needs to be updated?

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -424,64 +551,180 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
         const auto iPoint_Coarse = Parent_Local[iVertex];
         const auto iPoint_Fine = Children_Local[iVertex];
 
-        /*--- Be careful, it is possible that a node changes the agglomeration configuration,
-         the priority is always when receiving the information. ---*/
+        /*--- Debug: Check for out-of-bounds access ---*/
+        if (iPoint_Coarse >= Index_CoarseCV) {
+          cout << "ERROR [Rank " << rank << ", Marker " << MarkerS << "/" << MarkerR
+               << "]: Out-of-bounds coarse CV index " << iPoint_Coarse << " >= " << Index_CoarseCV << " (vertex "
+               << iVertex << ", fine point " << iPoint_Fine << ")" << endl;
```
pcarruscag: Is it a real error or just a warning?

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -424,64 +551,180 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
+          nOutOfBounds++;
+          continue;
+        }
+
+        /*--- Solution 1: Skip if this halo point was already agglomerated ---*/
+        auto existing_parent = fine_grid->nodes->GetParent_CV(iPoint_Fine);
+        if (existing_parent != ULONG_MAX) {
```
pcarruscag: numeric_limits

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -424,64 +551,180 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
   SU2_MPI::Allreduce(&nPointFine, &Global_nPointFine, 1, MPI_UNSIGNED_LONG, MPI_SUM, SU2_MPI::GetComm());
 
   SetGlobal_nPointDomain(Global_nPointCoarse);
 
   if (iMesh != MESH_0) {
-    const su2double factor = 1.5;
-    const su2double Coeff = pow(su2double(Global_nPointFine) / Global_nPointCoarse, 1.0 / nDim);
+    /*--- Note: CFL at the coarse levels have a large impact on convergence,
+          this should be rewritten to use adaptive CFL. ---*/
+    const su2double factor = 1.0;
```
pcarruscag:
```suggestion
```

Common/src/geometry/CMultiGridGeometry.cpp
```diff
@@ -503,11 +746,68 @@ CMultiGridGeometry::CMultiGridGeometry(CGeometry* fine_grid, CConfig* config, un
...
+  bool Volume = true;
+  su2double ratio = pow(fine_grid->nodes->GetVolume(iPoint), 1.0 / su2double(nDim)) * max_dimension;
+  su2double limit = pow(config->GetDomainVolume(), 1.0 / su2double(nDim));
+  if (ratio > limit) {
+    Volume = false;
+    cout << "Volume limit reached!" << endl;
+  }
+  /*--- Evaluate the stretching of the element ---*/
+
+  bool Stretching = true;
```
pcarruscag: Is it worth keeping? Do you know what it's supposed to do?
```suggestion
```

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -28,6 +28,60 @@
...
+ * \brief Helper function to enforce Euler wall BC by projecting momentum to tangent plane.
+ * \param[in] geo_coarse - Coarse grid geometry.
+ * \param[in] config - Problem configuration.
+ * \param[in,out] sol_coarse - Coarse grid solver (to access and modify solution/correction).
+ * \param[in] use_solution_old - If true, project Solution_Old (corrections); if false, project Solution.
+ */
+void ProjectEulerWallToTangentPlane(CGeometry* geo_coarse, const CConfig* config, CSolver* sol_coarse,
+                                    bool use_solution_old) {
+  for (auto iMarker = 0u; iMarker < config->GetnMarker_All(); iMarker++) {
+    if (config->GetMarker_All_KindBC(iMarker) == EULER_WALL) {
```
pcarruscag: What about symmetry? Also in the rules for agglomeration, isn't it equivalent?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -65,6 +119,17 @@ void CMultiGridIntegration::MultiGrid_Iteration(CGeometry ****geometry,
...
+
+#ifdef _OPENMP
+  /*--- Store original thread affinity settings ---*/
+  static bool affinity_set = false;
+  if (!affinity_set) {
+    // This ensures threads are bound consistently across runs
+    omp_set_schedule(omp_sched_static, 0);
+    affinity_set = true;
+  }
+#endif
```
pcarruscag: I don't think this has any effect because we don't use the "runtime" type of schedule, we always specify whether we want static or dynamic.
Do you have a particularly non-deterministic example case?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
pcarruscag:
```suggestion
    if (config->GetMarker_All_KindBC(iMarker) != EULER_WALL) continue;
```

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -211,6 +279,13 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
 
     }
 
+    /*--- MPI sync after RK stage to ensure halos have updated solution for next smoothing iteration ---*/
+    solver_fine->InitiateComms(geometry_fine, config, MPI_QUANTITIES::SOLUTION);
+    solver_fine->CompleteComms(geometry_fine, config, MPI_QUANTITIES::SOLUTION);
+
+    /*--- OpenMP memory fence to ensure all threads see consistent state ---*/
+    SU2_OMP_BARRIER
```
pcarruscag: You should not need a barrier both at the start and at the end of the loop.
Also, I think CompleteComms already has a barrier.
So maybe you only need a barrier before entering the loop.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+#ifdef HAVE_MPI
+      /*--- Ensure all ranks use the same CFL value (broadcast from rank 0) ---*/
+      SU2_MPI::Bcast(&CFL_coarse_new, 1, MPI_DOUBLE, 0, SU2_MPI::GetComm());
+#endif
+
+      /*--- Update the shared config object ---*/
+      config->SetCFL(iMesh+1, CFL_coarse_new);
+    }
+    END_SU2_OMP_MASTER
+    /*--- Implicit barrier at end of master region ensures all threads see updated CFL ---*/
```
pcarruscag: Hmmm there is no implied barrier in master regions, so maybe that's where the issues come from

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+      su2double rms_res_coarse_local = 0.0;
+      for (unsigned short iVar = 0; iVar < solver_coarse->GetnVar(); iVar++) {
+        rms_res_coarse_local += solver_coarse->GetRes_RMS(iVar);
+      }
+
+      /*--- MPI synchronization: ensure all ranks use the same global residual value ---*/
+      /*--- This is critical for consistent CFL adaptation across all ranks ---*/
+      su2double rms_res_coarse = rms_res_coarse_local;
+
+#ifdef HAVE_MPI
```
pcarruscag: Not necessary, allreduce works properly in non mpi builds.
```suggestion
```

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+    su2double current_coeff = CFL_coarse_current / CFL_fine;
+
+    /*--- Adaptive CFL using Exponential Moving Average (EMA) ---*/
+    /*--- All operations performed in master thread for determinism ---*/
+    constexpr int AVG_WINDOW = 5;
+    constexpr int MAX_MG_LEVELS = 10;
+
+    /*--- Use static storage shared across all threads for deterministic CFL adaptation ---*/
+    /*--- Each MPI rank maintains its own independent adaptive CFL state ---*/
+    /*--- Note: MPI ranks do not share memory, so static is safe for MPI parallelism ---*/
```
pcarruscag: This confused me at first, it sounded like each rank would have its own value of CFL, I think you can remove the comment.
MPI implies distributed memory.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+    static su2double current_avg[MAX_MG_LEVELS] = {};
+    static su2double prev_avg[MAX_MG_LEVELS] = {};
+    static su2double last_res[MAX_MG_LEVELS] = {};
+    static bool last_was_increase[MAX_MG_LEVELS] = {};
+    static int oscillation_count[MAX_MG_LEVELS] = {};
+    static unsigned long last_check_iter[MAX_MG_LEVELS] = {};
+    static unsigned long last_update_iter[MAX_MG_LEVELS] = {};
```
pcarruscag: Maybe make them class members so that `static` is not abused too much

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+      /*--- Get global iteration count first ---*/
+      unsigned long current_iter;
+      if (config->GetTime_Domain())
+        current_iter = config->GetTimeIter();
+      else
+        current_iter = config->GetInnerIter();
+
+      /*--- Reset state at the beginning of a new solve (iter 0 or 1) ---*/
+      /*--- This ensures deterministic behavior across multiple runs ---*/
+      static unsigned long last_reset_iter = ULONG_MAX;
```
pcarruscag: yeah this gets a bit out of hand xD

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +332,224 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+      /*--- Update the shared config object ---*/
+      config->SetCFL(iMesh+1, CFL_coarse_new);
+    }
+    END_SU2_OMP_MASTER
+    /*--- Implicit barrier at end of master region ensures all threads see updated CFL ---*/
+
+    /*--- Update LocalCFL at each coarse grid point ---*/
+    SU2_OMP_FOR_STAT(roundUpDiv(geometry_coarse->GetnPoint(), omp_get_num_threads()))
+    for (auto iPoint = 0ul; iPoint < geometry_coarse->GetnPoint(); iPoint++) {
+      solver_coarse->GetNodes()->SetLocalCFL(iPoint, CFL_coarse_new);
```
pcarruscag, line 373: Make a function for the adaptation, please.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -300,58 +581,69 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
         solver_fine->Postprocessing(geometry_fine, solver_container_fine, config, iMesh);
 
       }
+
+      /*--- MPI sync after RK stage to ensure halos have updated solution for next smoothing iteration ---*/
```
pcarruscag: I don't see why this would be necessary, in the whole code we always communicate everything after it is computed, so this is either missing somewhere else, or masking another issue.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -300,58 +581,69 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
 
-    for (iVar = 0; iVar < nVar; iVar++)
+    for (auto iVar = 0u; iVar < nVar; iVar++)
       sol_coarse->GetNodes()->SetSolution_Old(Point_Coarse,Solution);
   }
   END_SU2_OMP_FOR
 
   delete [] Solution;
 
+  /*--- Ensure all threads complete before MPI communication ---*/
```
pcarruscag: Not needed, for loops have an implied barrier at the end.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -369,10 +661,13 @@ void CMultiGridIntegration::GetProlongated_Correction(unsigned short RunTime_EqS
   sol_coarse->InitiateComms(geo_coarse, config, MPI_QUANTITIES::SOLUTION_OLD);
   sol_coarse->CompleteComms(geo_coarse, config, MPI_QUANTITIES::SOLUTION_OLD);
 
+  /*--- Ensure MPI completion visible to all threads ---*/
+  SU2_OMP_BARRIER
```
pcarruscag: Same comment about CompleteComms.
We use funneled communications, so I'm pretty sure we have a barrier somewhere in the comms.

Common/src/geometry/dual_grid/CPoint.cpp
bigfooted: Thanks for the correction, I had to update myself on the difference.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: ok, I'll check if this can now be removed, it seemed to be useful before.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: ah ok, I guess this shows my lack of knowledge regarding OMP implementation.

pcarruscag: That's what the standard says, it's not our implementation doing funny bussiness.
If you want a loop that doesn't sync, you need to explicitly use the "no wait" directive.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: removed...

Common/src/geometry/CMultiGridGeometry.cpp
bigfooted: Stretching: maybe something akin to implicit lines, or prevent mesh quality deterioration. I can remove it because I have an implicit lines implementation.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: You are right, I am assuming that a symmetry plane is always straight. Should we actually assume that? We currently don't in ComputeSurfStraightness. Not sure if curved symmetry planes make sense?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: it was my original idea but I decided against it. Not sure if it would give a lot of benefit to have a local CFL per rank, and probably leads to all kinds of issues on the boundaries where you have to match CFL.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: is using static like this bad for something?

pcarruscag: Besides reasoning about their initialization, they force the compiler to not optimize a lot of things because static is almost like a global variable.

SU2_CFD/include/integration/CMultiGridIntegration.hpp
bigfooted: thanks, removed it.

Common/src/geometry/CMultiGridGeometry.cpp
bigfooted: it actually never happens, so I will remove this.

Common/src/geometry/CMultiGridGeometry.cpp
bigfooted: I rechecked this. We can allow agglomeration on a marker node when it is shared with an mpi interface. It does not seem to do any (convergence) harm. It agglomerates a couple of additional cells, so it also does not help much.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: all cases seem to be nondeterministic, it increases with nr of multigrid levels and compiler optimization level. I tried out a bunch of stuff and even let Claude write some stuff, but most of that did not seem to help in the end. I have removed this part.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: thanks, I checked and removed this.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: modified the statement...

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -488,12 +488,25 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
 
-    /*--- Compute adaptive CFL for coarse grid ---*/
-    su2double CFL_coarse_new = computeMultigridCFL(config, solver_coarse, geometry_coarse, iMesh, CFL_fine, CFL_coarse_current);
+    SU2_OMP_MASTER
+    {
+      CFL_fine = config->GetCFL(iMesh);
+      CFL_coarse_current = config->GetCFL(iMesh+1);
+
+      /*--- Compute adaptive CFL for coarse grid (master thread only) ---*/
+      CFL_coarse_new = computeMultigridCFL(config, solver_coarse, geometry_coarse, iMesh, CFL_fine, CFL_coarse_current);
```
pcarruscag: I can help you with the OMP stuff, just point me to a simple case with issues.
What you're doing here doesn't work because CFL_fine is a thread-local variable, so only the master thread will have a value here.

bigfooted: Thank, pedro, I am still (a bit)  lost here.

bigfooted: Hmmm I reverted this again, but this specific fix seems to solve the 'regular' thread sanitizer tests'. The AD thread sanitizer tests are still failing though.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,21 +466,62 @@
...
+          Extract passive values to avoid AD tape recording in parallel region. ---*/
+    // passivedouble CFL_fine_passive = 0.0;
+    // passivedouble CFL_coarse_current_passive = 0.0;
+
+    // SU2_OMP_MASTER
+    // {
+    //   CFL_fine_passive = SU2_TYPE::GetValue(config->GetCFL(iMesh));
+    //   CFL_coarse_current_passive = SU2_TYPE::GetValue(config->GetCFL(iMesh+1));
+    // }
+    // END_SU2_OMP_MASTER
```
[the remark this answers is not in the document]
bigfooted: @pcarruscag ok, it is definitely this CFL_coarse_new that is giving issues, I put it to constant and the thread sanitizer tests are passing now - no more race conditions.

Common/include/geometry/CMultiGridGeometry.hpp
```diff
@@ -28,36 +28,37 @@
 #pragma once
 
 #include "CGeometry.hpp"
+class CMultiGridQueue;
```
pcarruscag: unused?

Common/include/CConfig.hpp
```diff
@@ -3057,7 +3057,27 @@ class CConfig {
...
+        break;
+      case EULER_EXPLICIT:
+      case EULER_IMPLICIT:
+        iRKLimit = 1;
+        break;
+      default:
+        iRKLimit = 1;
+        break;
+    }
+  return iRKLimit;
```
```suggestion
    return iRKLimit;
```

SU2_CFD/include/integration/CMultiGridIntegration.hpp
```diff
@@ -133,6 +133,55 @@ class CMultiGridIntegration final : public CIntegration {
...
+   * \param[in] iRKLimit - Number of Runge-Kutta steps.
+   */
+  void PostSmoothing(unsigned short RunTime_EqSystem,
+  CSolver* solver_fine,
+  CNumerics** numerics_fine,
+  CGeometry* geometry_fine,
+  CSolver** solver_container_fine,
+  CConfig *config,
+  unsigned short iMesh,
+  unsigned short iRKLimit);
```
pcarruscag: looks a bit unusual

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -28,6 +28,225 @@
...
+
+      /*--- Compute normal component of momentum.
+       *    For grid movement (e.g. rotating frame) and when projecting the solution
+       *    (not corrections), compute the normal component of the RELATIVE momentum
+       *    (rho*v - rho*v_grid).n to enforce (v - v_grid).n = 0.
+       *    This matches the fine-grid Euler/symmetry BC (BC_Sym_Plane). ---*/
+      su2double momentum_n = 0.0;
+      for (auto iDim = 0u; iDim < nDim; iDim++) {
+        momentum_n += solution_coarse[iDim + 1] * UnitNormal[iDim];
+      }
```
pcarruscag: Can we get the velocity indices from the solver? Or simply move this entire function to the base flow solver where that info is readily available.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -28,6 +28,225 @@
...
+ * \param[in] config - Problem configuration.
+ * \param[in,out] sol_coarse - Coarse grid solver (to access and modify solution/correction).
+ * \param[in] use_solution_old - If true, project Solution_Old (corrections); if false, project Solution.
+ */
+void ProjectEulerWallToTangentPlane(CGeometry* geo_coarse, const CConfig* config, CSolver* sol_coarse,
+                                    bool use_solution_old) {
+  const bool grid_movement = config->GetGrid_Movement();
+
+  for (auto iMarker = 0u; iMarker < config->GetnMarker_All(); iMarker++) {
+    if (config->GetMarker_All_KindBC(iMarker) != EULER_WALL) continue;
```
pcarruscag: I think I asked but maybe forgot. What about symmetry?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -28,6 +28,225 @@
...
+    /*--- Update coarse grid CFL ---*/
+    CFL_coarse_new = max(0.5 * CFL_fine, min(CFL_fine, CFL_fine * new_coeff));
+
+#ifdef HAVE_MPI
+    /*--- Ensure all ranks use the same CFL value (broadcast from rank 0) ---*/
+    /*--- Use su2double buffer for MeDiPack compatibility. ---*/
+    su2double CFL_bcast = CFL_coarse_new;
+    SU2_MPI::Bcast(&CFL_bcast, 1, MPI_DOUBLE, 0, SU2_MPI::GetComm());
+    CFL_coarse_new = SU2_TYPE::GetValue(CFL_bcast);
+#endif
```
pcarruscag: Aren't all ranks doing the same calculations? If not, Bcast doesn't seem like the robust way to get concensus among ranks
```suggestion
```

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -64,8 +283,6 @@ void CMultiGridIntegration::MultiGrid_Iteration(CGeometry ****geometry,
 
   const unsigned short Solver_Position = config[iZone]->GetContainerPosition(RunTime_EqSystem);
 
-  /*--- Start an OpenMP parallel region covering the entire MG iteration, if the solver supports it. ---*/
-
```
pcarruscag, line 68: revert

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,101 +418,192 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+CSolver***** solver_container,
+CConfig **config_container,
+CSolver* solver_fine,
+CNumerics** numerics_fine,
+CGeometry* geometry_fine,
+CSolver** solver_container_fine,
+CConfig *config,
+unsigned short iMesh,
+unsigned short iZone,
+unsigned short iRKLimit) {
```
pcarruscag: formatting

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,101 +418,192 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+  const bool classical_rk4 = (config->GetKind_TimeIntScheme() == CLASSICAL_RK4_EXPLICIT);
+  const unsigned short nPostSmooth = config->GetMG_PostSmooth(iMesh);
+  const unsigned long timeIter = config->GetTimeIter();
 
+  /*--- Do a postsmoothing on the grid iMesh after prolongation from the grid iMesh+1 ---*/
+  for (unsigned short iPostSmooth = 0; iPostSmooth < nPostSmooth; iPostSmooth++) {
+      /*--- Synchronize before each post-smoothing iteration ---*/
+      SU2_OMP_BARRIER
+      for (unsigned short iRKStep = 0; iRKStep < iRKLimit; iRKStep++) {
+      solver_fine->Preprocessing(geometry_fine, solver_container_fine, config, iMesh, iRKStep, RunTime_EqSystem, false);
```
pcarruscag: formatting

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,101 +418,192 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+      Time_Integration(geometry_fine, solver_container_fine, config, iRKStep, RunTime_EqSystem);
+      /*--- Send-Receive boundary conditions, and postprocessing ---*/
+      solver_fine->Postprocessing(geometry_fine, solver_container_fine, config, iMesh);
     }
-  }
 
+    /*--- MPI sync after RK stage to ensure halos have updated solution for next smoothing iteration ---*/
+    solver_fine->InitiateComms(geometry_fine, config, MPI_QUANTITIES::SOLUTION);
+    solver_fine->CompleteComms(geometry_fine, config, MPI_QUANTITIES::SOLUTION);
+  }
```
pcarruscag: I don't understand why all the extra comms are necessary.
You only changed the agglomeration and CFL, where is this coming in?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -257,101 +418,192 @@ void CMultiGridIntegration::MultiGrid_Cycle(CGeometry ****geometry,
...
+void CMultiGridIntegration::PostSmoothing(unsigned short RunTime_EqSystem, CSolver* solver_fine,CNumerics** numerics_fine, CGeometry* geometry_fine, CSolver** solver_container_fine, CConfig *config, unsigned short iMesh,unsigned short iRKLimit)
+{
+  const bool classical_rk4 = (config->GetKind_TimeIntScheme() == CLASSICAL_RK4_EXPLICIT);
+  const unsigned short nPostSmooth = config->GetMG_PostSmooth(iMesh);
+  const unsigned long timeIter = config->GetTimeIter();
 
+  /*--- Do a postsmoothing on the grid iMesh after prolongation from the grid iMesh+1 ---*/
+  for (unsigned short iPostSmooth = 0; iPostSmooth < nPostSmooth; iPostSmooth++) {
+      /*--- Synchronize before each post-smoothing iteration ---*/
+      SU2_OMP_BARRIER
```
pcarruscag: All the extra barriers also don't make sense to me.
What changed in the solver operations that made them sensitive to this?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -28,6 +28,225 @@
...
+#ifdef HAVE_MPI
+    /*--- Ensure all ranks use the same CFL value (broadcast from rank 0) ---*/
+    /*--- Use su2double buffer for MeDiPack compatibility. ---*/
+    su2double CFL_bcast = CFL_coarse_new;
+    SU2_MPI::Bcast(&CFL_bcast, 1, MPI_DOUBLE, 0, SU2_MPI::GetComm());
+    CFL_coarse_new = SU2_TYPE::GetValue(CFL_bcast);
+#endif
+
+    /*--- Update the shared config object (wrapped to prevent tape recording) ---*/
+    const bool wasActive = AD::BeginPassive();
```
pcarruscag: Better wrap the entire function?

Common/include/geometry/CMultiGridGeometry.hpp
bigfooted: yep

SU2_CFD/include/integration/CMultiGridIntegration.hpp
bigfooted: fixed alignment.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: Only curved surfaced had issues, so symmetry is fine. Actually, I don't think we check curvature of symmetry planes. I will add SYMMETRY here as well. Not sure if there is a use case for curved symmetry.

pcarruscag: Ok if it works it works, just a leave a comment why only Euler is handled

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: move to CFVMFlowSolverBase.inl?

pcarruscag: yep

bigfooted: Computer says yes.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: you mean just move it to the start of the function 
```
passivedouble CMultiGridIntegration::computeMultigridCFL(CConfig* config, CSolver* solver_coarse, CGeometry* geometry_coarse,  unsigned short iMesh, passivedouble CFL_fine, passivedouble CFL_coarse_current) {

  const bool wasActive = AD::BeginPassive();
```

pcarruscag: yes, because there are other su2doubles in it.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: I checked and they seem to be redundant, I will remove them.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
bigfooted: Thanks, there was an issue with the allreduce that is in the same routine, it was not doing it for the finest mesh. After fixing that, I could remove this.

pcarruscag (approved):

PR #2716, opened 2026-01-23
Fix clang warnings in CNSSolver.cpp

Nesar976:
This PR fixes clang and clang-tidy warnings in CNSSolver.cpp.

Unused includes were removed where applicable, and std::endl usages were
replaced to address performance warnings. The inclusion of
CFVMFlowSolverBase.inl is intentionally preserved for explicit template
instantiation, with a local suppression added to handle the clang
unused-includes warning.

No solver logic or numerical behavior is changed.


SU2_CFD/src/solvers/CNSSolver.cpp
```diff
@@ -681,34 +626,47 @@ void CNSSolver::BC_Isothermal_Wall_Generic(CGeometry *geometry, CSolver **solver
...
+    su2double eddy_viscosity = nodes->GetEddyViscosity(iPoint);
     su2double thermal_conductivity = nodes->GetThermalConductivity(iPoint) + Cp * eddy_viscosity / Prandtl_Turb;
 
-    // work in progress on real-gases...
-    //thermal_conductivity = nodes->GetThermalConductivity(iPoint);
-    //Cp = nodes->GetSpecificHeatCp(iPoint);
-    //thermal_conductivity += Cp*eddy_viscosity/Prandtl_Turb;
+    /*--- Safeguard for CONSTANT_PRANDTL_TURB on coarse meshes.
+          If the first point off the wall is in the turbulent region (mu_t/mu_l > 0.1),
+          the linear temperature gradient assumption is inaccurate. ---*/
```
bigfooted, line 640: Please cite a paper that states this.

Nesar976: Good point, thanks for flagging this.

This is a known limitation of wall-function-based heat transfer models when the first off-wall point lies in the fully turbulent region. For example, see Kader (1981) and Launder & Spalding (1974), which discuss that linear temperature profiles are only valid within the near-wall region and break down when y+ is too large.

I can add an explicit reference in the comment or rephrase it to make the assumption more precise if you prefer.

pcarruscag:
This looks like more AI slop. Join the slack channel and come a developer meeting introduce yourself if you are actually interested in contributing.

Nesar976:
Understood — thanks for the feedback. I’ll join the Slack and take some time to understand the project workflow and expectations before contributing further.

Issue #2594, opened 2025-10-16
Issue running  Streamwise periodic flow tutorial


Cristopher-Morales:
Hi!!

I hope you are doing well,

I tried to run the tutorial case [Stream periodic flow](https://su2code.github.io/tutorials/Inc_Streamwise_Periodic/) and it seems that it does not converge. I am using the current SU2 develop.
Also the solution seems to have some unphysical behaviour 

<img width="933" height="465" alt="Image" src="https://github.com/user-attachments/assets/79c95f1e-46e0-4b91-a089-438336f37ab8" />

the residuals:

<img width="1178" height="740" alt="Image" src="https://github.com/user-attachments/assets/eddc65c7-a4f5-42ec-9f7c-90c7a930d639" />

additionally, I observed that the solutions files were updated 3 months ago, when I run the adjoint solver using SU2_CFD_AD DA_configMaster.cfg
the simulation diverges:

<img width="1052" height="690" alt="Image" src="https://github.com/user-attachments/assets/4a647634-16b4-43cb-87b4-d9a05736e9f7" />

Could someone have a look about this tutorial case and guide me how to make it work again?

Thank you so much in advance!!!

bigfooted:
Problem started to occur after #2533 

> 
> commit 0b6e0ad10d5c874ac61b09254e2b583089a4ff32 (HEAD)
> Author: Rutvik Khedkar <140829696+Bot-Enigma-0@users.noreply.github.com>
> Date:   Tue Jul 29 21:15:44 2025 +0100
> 
>     Updating the Cross-Diffusion Term Discretisation in k-w SST Model (#2533)
>

bigfooted:
The underrelaxations don't help, or switching to accurate jacobians.

bigfooted:
Before commit https://github.com/su2code/SU2/commit/0b6e0ad10d5c874ac61b09254e2b583089a4ff32
<img width="995" height="649" alt="Image" src="https://github.com/user-attachments/assets/a510bbfb-8470-4983-8eae-49488f50197f" />
after commit:

<img width="833" height="615" alt="Image" src="https://github.com/user-attachments/assets/7eddfb9b-24ef-4d18-9968-f3de8cb3ddfc" />

C-Perron:
Hey everyone. I stumbled onto this issue randomly, but I think I have found the issue with the SST turbulence model. If you set `MAX_UPDATE_SST` to a large value, say 1e8, you will find that the simulation will then converge as it used to.

[...]

PR #2728, opened 2026-02-09
Fixing relaxation factor for conserved turbulent variables in SST model

C-Perron:
## Proposed Changes

This fixes an issue introduced in #2525 where the under-relaxation factors for $k$ and $\omega$ were calculated incorrectly with the SST model.

Since the update for $k$ and $\omega$ is in terms of conserved transport quantities, their stored value must be multiplied by the local flow density when compared to the non-linear update. This was not the case previously, and the solver could overestimate the relative change and compute a much stricter relaxation factor.

This is particularly relevant to incompressible flow, such as the [steam periodic flow](https://su2code.github.io/tutorials/Inc_Streamwise_Periodic/) tutorial case, where the density is several orders of magnitude. The consequence of this issue was a stalled convergence of the turbulence variables as discussed in #2594.

This issue did not affect problems using the SA model, which doesn't use a conserved quantity in its formulation, or where the normalized density is near unity.

## Related Work
Related to #2525 and #2594

## PR Checklist

- [ ] ~~I have added a test case that demonstrates my contribution, if necessary.~~
- [ ] ~~I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.~~


pcarruscag (approved):
LGTM

bigfooted:
Thanks! Can you update the residual values in the Testcases/*.py regression files with the new values as reported above in the failed regression tests?

PR #1237, opened 2021-03-18
Fix restart file writing for time convergence and 2nd order time stepping

ScSteffen:
## Proposed Changes
Single and Multizone driver now write restart files for last and second last iteration, if windowed time convergence happend and 2nd order time marching is active.
 

## Related Work
Solve issue #1211

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [x] I am submitting my contribution to the develop branch.
- [x] My contribution generates no new compiler warnings (try with the '-Wall -Wextra -Wno-unused-parameter -Wno-empty-body' compiler flags, or simply --warnlevel=2 when using meson).
- [x] My contribution is commented and consistent with SU2 style.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp) , if necessary.

PR #2730, opened 2026-02-13
Add functionality for unsteady restart handling in CSinglezoneDriver …

### LINKED ISSUE #2493: saving unsteady restart files mismatch
When max_time is reached, the unsteady restart for second order time stepping only saves the last timestep N and not timestep N-1.

Soumyadipta-Banerjee:
## Proposed Changes
This PR fixes an issue where unsteady simulations with second-order time stepping would only save the current timestep (N) to restart files when max_time was reached, leaving out the previous timestep (N-1). For second-order time marching schemes, we need both timesteps to correctly restart the simulation.

The fix adds logic to detect when we're stopping due to max_time being reached and automatically writes the previous timestep before writing the current one. This way, the restart file contains all the data needed for a proper restart.
## Related Work
*Resolves #2493 - Saving unsteady restart files mismatch

This addresses the issue where users had to work around missing N-1 timestep data by ensuring max_time aligns with output frequency, which isn't always practical or desired.

## PR Checklist


pcarruscag (commented):
Thanks. I think the  cleaner way of doing this is to detect that we are about to stop due to max time, we already have similar logic for stopping due to max iterations.
I don't think what you implemented around the old solution generalizes to all types of problems.

pcarruscag:
Here is the PR with a similar fix for max iterations instead of max time. Do it this way because it's simpler, and swapping Solution and SolutionOld is not safe. https://github.com/su2code/SU2/pull/1237/changes
Either detect that the current iteration is the penultimate and force the writing of output files, or make the solver run one more iteration after max time is reached (like the PR above).

Soumyadipta-Banerjee:
Hi @pcarruscag, I have made the changes and the new code detects when max_time is reached and delays convergence by one iteration (modeled after PR #1237), which ensures both the current and previous timesteps are saved to the restart file.

pcarruscag:
Thanks, I will test your changes when I get some time, maybe this weekend.

SU2_CFD/include/output/COutput.hpp
```diff
@@ -591,6 +592,7 @@ class COutput {
    * \param[in] solver_container - Container vector with all the solutions.
    * \param[in] iter - The current time, outer or inner iteration index.
    * \param[in] force_writing - If <TRUE>, writing of output files is forced without checking the output frequency.
+   * \param[in] write_restart_only - If <TRUE>, only restart (and CSV restart) volume files are written.
```
pcarruscag:
```suggestion
```

PR #2733, opened 2026-02-17
Improved  combustion solver robustness for coarse grids

EvertBunschoten:
Simulations of flames in large domains is difficult because volumetric source terms blow up in large cells, causing the species solver to diverge. I implemented a damping term which scales the species source terms in cells with a length scale above a user-defined threshold. This allows for the flame to propagate through regions of the domain with large cells (such as in the far-field) without diverging, while the source terms in refined regions remain unaffected.

To illustrate how this feature improves robustness, I ran a 2D simulation of a circular flame propagating from a refined region in the middle domain towards the edges where the mesh is significantly more coarse.

grid

The image below shows the temperature field after 30 iterations as computed with the previous implementation (left), and the new implementation with a flame length scale specified as 1e-4 m (right). The solution is unaffected while the flame is within the refined region of the mesh.

iter30

However, the instance the flame reaches the coarser cells, the solver quickly diverges without damping. With the damping active, the flame propagates through the entire domain without an issue. 
iter40

## PR Checklist


pcarruscag (commented):
Hmmm but then you are not solving the same equations anymore.
If the issue is "propagation speed" it sounds like something should be done about the CFL.

EvertBunschoten:
> Hmmm but then you are not solving the same equations anymore. If the issue is "propagation speed" it sounds like something should be done about the CFL.

It is not about accurately modeling the propagation of the flame, it is just to allow the flame to propagate through the coarse regions of the domain without the solver diverging. For cells larger than the flame length scale, we can't accurately resolve the propagation of the flame front anyway because the diffusive time scales and reaction source time scales differ at those length scales.

pcarruscag:
So is it possible to adjust the diffusion coefficients in those regions?
Numerical diffusion as a way of dealing with instabilities is a more established practice than disabling source terms.

bigfooted:
I agree with Pedro, it is great that it doesn't blow up, but the solution is not physical anymore. You can look into 'thickened flame' models. In these models, the propagation speed is physically correct, but the flame is artificially thickened by increasing the diffusion coefficient.

bigfooted:
It's mentioned in Poinsot & Veynante btw.

EvertBunschoten:
I was not clear when initially describing the problem. The images I showed were of a **steady** simulation, whereas the thickened flame models described in literature were developed for **unsteady** problems. 
For unsteady flames, adding the correction factor to only the diffusivity seems to improve the propagation of flames in coarse cells very nicely. 
The image below shows the comparison between unsteady simulations of an expanding flame front with (right) and without (left) scaling the diffusivity according to the cell volume. The mesh is refined within the purple circle and is very coarse beyond it.

comparison_t001

The flame propagates at the same speed for both cases within the refined region, but changes once the flame goes into the coarse region. 
comparison_t0024

With the artificially thickened flame correction (diffusion only), the flame propagates at approximately the same speed through the coarse cells, whereas the flame speed drops significantly without the correction. 

comparison_t003

EvertBunschoten:
For the steady case (constant CFL number=10), it seems that suppressing the source term and enhancing diffusivity in the course cells improves the robustness of the simulations. The image below shows the results of a steady simulation on the same grid as the unsteady simulation without any correction term (left) and with diffusion increased and source terms decreased (right). 

steady_comparison_60

At around 80 iterations, the flame front reaches the coarse cells, where the uncorrected simulation quickly diverges.

steady_comparison_80

With the correction terms, the flame front propagates rapidly through the coarse cells, which is expected for a steady simulation, given that larger cells have a larger time step.

steady_comparison_100

bigfooted:
looks good. Note that the thickened flame model of the paper we discussed https://share.google/Ia0hcPhfoOKvHypb3 recovers the flame speed by also modifying the diffusivity and source term in the energy equation, and therefore modifies the flame temperature. 
If you plan to silently do this by default, maybe make the lengthscale depend on the local estimate of the flame thickness (in all regions with non-negligible source terms) and do this when the cell size is > 5 times the flame thickness or so. And start giving warnings to the user on screen about it.

EvertBunschoten:
> looks good. Note that the thickened flame model of the paper we discussed https://share.google/Ia0hcPhfoOKvHypb3 recovers the flame speed by also modifying the diffusivity and source term in the energy equation, and therefore modifies the flame temperature. If you plan to silently do this by default, maybe make the lengthscale depend on the local estimate of the flame thickness (in all regions with non-negligible source terms) and do this when the cell size is > 5 times the flame thickness or so. And start giving warnings to the user on screen about it.

It is possible to calculate the flame thickness during simulations with Equation 4 of the paper you linked. However, the flame speed correction model will not work when the correction factor depends on the local value of the flame thickness.
As the flame propagates through coarse cells, its thickness will increase due to the artificially increased diffusivity. Therefore, the flame will grow in thickness relative to the local cell size, which will reduce the value of the  correction factor, leading to a reduction in flame propagation speed.

bigfooted:
is it possible to compute an estimate of the original flame thickness locally?

EvertBunschoten:
> is it possible to compute an estimate of the original flame thickness locally?

I have tried the following method to try to obtain a measure for the thickness of the flame on a sufficiently refined grid. 
Each iteration, the flame thickness is calculated with
$$t=\frac{\max(T) - \min(T)}{\max(\nabla_c T)},$$
where 
$$\nabla_c T = \frac{\nabla c \cdot \nabla T}{|\nabla c|}.$$
To account for the thickening of the flame when propagating through coarse cells, the flame thickness value used to calculate the correction factor is only updated if the flame thickness **decreases**.
$$t_{i+1} = \min(t_i, t)$$
This approach seems to work well for **adiabatic** flames when the flame starts in a refined section of the domain. However, if there are **isothermal or CHT BCs**, the flame thickness drastically reduces when the flame front propagates through the thermal boundary layer. 
The decrease in flame thickness increases the value of the correction factor, causing the flame front to stagnate or the solver to diverge. 
I have also tried to only use the gradient of the progress variable to calculate the flame thickness, but I encounter the same issue.

bigfooted:
So the issue I have is: how much are we going to accommodate a bad setup (a mesh that is too coarse) for a user? 
If it is a matter of adding a 'coarse_mesh_correction = ON' and that's it, then that's fine by me. Preferably when it can be on by default and does not do anything when the mesh is fine enough. 
If the adiabatic correction reproduces the correct flamespeed (steady and unsteady) without any other magic constants, I think we should definitely put that in. 
Since your testcase is non-adiabatic, more correction is needed. Is there some additional correction possible that takes into account the change in enthalpy? In this heat loss region, can you scale the correction to get it to converge, so have an additional correction when the enthalpy has changed compared to a reference enthalpy? If yes, then maybe we can convert this 'magic scaling factor' into an local_enthalpy/adiabatic_enthalpy ratio or something.

EvertBunschoten:
> So the issue I have is: how much are we going to accommodate a bad setup (a mesh that is too coarse) for a user? If it is a matter of adding a 'coarse_mesh_correction = ON' and that's it, then that's fine by me. Preferably when it can be on by default and does not do anything when the mesh is fine enough. If the adiabatic correction reproduces the correct flamespeed (steady and unsteady) without any other magic constants, I think we should definitely put that in. Since your testcase is non-adiabatic, more correction is needed. Is there some additional correction possible that takes into account the change in enthalpy? In this heat loss region, can you scale the correction to get it to converge, so have an additional correction when the enthalpy has changed compared to a reference enthalpy? If yes, then maybe we can convert this 'magic scaling factor' into an local_enthalpy/adiabatic_enthalpy ratio or something.

The purpose of this feature is not to accommodate a bad setup where the mesh in the flame zone is too coarse to accurately resolve the flame front. For complicated 3D laminar test cases and future support for tubulent flames, the flow field needs to develop before igniting the flame. Consequently, the flame front will propagate  through all regions of the computational domain downstream of the flame, including regions close to the outlet where the mesh is less refined. 
Modifying the correction factor based on the change in the total enthalpy w.r.t. the initial value is something I'm working on, and by using the suggestion of Josh to have it be dependent on the wall distance.

bigfooted:
ok, so for this PR, do you just want to implement the steady case, where any unsteady propagation is nonphysical anyway? There is still some kind of length scale involved then?

pcarruscag (commented):
Perhaps we can move forward with a default that does not compromise results, and it is then an option to set an adequate lenght scale?

EvertBunschoten:
> Perhaps we can move forward with a default that does not compromise results, and it is then an option to set an adequate lenght scale?

I developed a new way to approximate the flame thickness:

$t_\mathrm{flame} = \frac{\max(c) - \min(c)}{\max(|\nabla_{Tu}c|)},$
where
$\nabla_{Tu}c = \frac{\nabla_u c\cdot\nabla T}{\nabla T \cdot \nabla T}\nabla T,$
and 
$\nabla_u c = \frac{\nabla c\cdot u}{u\cdot u} u.$
This formulation works better in non-adiabatic regions. 

The flame thickness value is now also included in the history output when the option is enabled in the configuration so users are better informed on whether the mesh is sufficiently refined. 

Finally, the global flame thickness value is now adjusted upwards or downwards depending on whether the flame thickness increases or decreases. This allows for the flame thickness value to become more accurate as the solution converges, while providing sufficient dampening while the flame propagates through coarse cells.

bigfooted:
Nice work! Is there also no correction when the mesh is sufficiently fine?

Common/include/option_structure.hpp
```diff
@@ -1446,6 +1446,7 @@ struct FluidFlamelet_ParsedOptions {
   su2double* spark_reaction_rates; /*!< \brief Source terms for flamelet spark ignition option. */
   unsigned short nspark;           /*!< \brief Number of source terms for spark initialization. */
   bool preferential_diffusion = false;  /*!< \brief Preferential diffusion physics for flamelet solver.*/
+  bool thickenedflame_correction{true}; /*!< \brief Thickened flame correction. */
```
bigfooted, line 1449: The default=true is OK if the model does not do a correction for sufficiently fine meshes.

Common/src/CConfig.cpp
```diff
@@ -1429,6 +1429,9 @@ void CConfig::SetConfig_Options() {
   /*!\brief SPARK_REACTION_RATES \n DESCRIPTION: Net source term values applied to species within spark area during spark ignition. \ingroup Config*/
   addDoubleListOption("SPARK_REACTION_RATES", flamelet_ParsedOptions.nspark, flamelet_ParsedOptions.spark_reaction_rates);
 
+  /*!\brief THICKENED_FLAME_CORRECTION \n DESCRIPTION: Dampen source terms and enhance diffusion based on the flame length scale. \ingroup Config*/
```
bigfooted: from the description it is not clear that this is a coarse mesh correction.

Common/include/option_structure.hpp
EvertBunschoten: The dampening factor is automatically set to 1.0 if the cell length scale is lower than the calculated flame length scale.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -383,7 +418,7 @@ void CSpeciesFlameletSolver::SetPreconditioner(CGeometry* geometry, CSolver** so
 
 void CSpeciesFlameletSolver::Source_Residual(CGeometry* geometry, CSolver** solver_container,
                                              CNumerics** numerics_container, CConfig* config, unsigned short iMesh) {
-
+
```
bigfooted:
```suggestion

```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@
   su2double pv_burnt = scalars[I_PROGVAR] - delta;
   return pv_burnt;
 }
+
+
+su2double CSpeciesFlameletSolver::ThickenedFlameCorrection(CGeometry const * geometry, const unsigned long iPoint) const {
+  su2double F{1.0};
+  if (global_flame_thickness != default_flame_thickness) {
```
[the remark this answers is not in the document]
bigfooted: I agree, @EvertBunschoten can you modify this?

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+  END_SU2_OMP_FOR
+  SU2_MPI::Barrier(SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&gradpv_local, &gradpv_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&pvmin_local, &pvmin_global, 1, MPI_DOUBLE, MPI_MIN, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&pvmax_local, &pvmax_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&Tmax_local, &Tmax_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+
+  /* Update flame thickness value. */
+  su2double flame_thickness{default_flame_thickness};
+  if (Tmax_global > 800.0) flame_thickness = (pvmax_global - pvmin_global) / (gradpv_global+1e-5);
```
bigfooted: is EPS not good enough?

bigfooted:
@EvertBunschoten can you add a testcase?

SU2_CFD/include/solvers/CSpeciesFlameletSolver.hpp
```diff
@@ -105,6 +109,22 @@ class CSpeciesFlameletSolver final : public CSpeciesSolver {
...
   unsigned long SetPreferentialDiffusionScalars(CFluidModel* fluid_model_local,
                                                 unsigned long iPoint, const vector<su2double>& scalars);
+  
+  /*!
+   * \brief Calculate correction factor for flame propagation on coarse grids.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] iPoint - node ID.
+   * \return - flame thickness correction factor.
+   */                                              
+  su2double ThickenedFlameCorrection(CGeometry const * geometry, const unsigned long iPoint) const;
```
pcarruscag:
```suggestion
  su2double ThickenedFlameCorrection(const CGeometry* geometry, unsigned long iPoint) const;   
```

SU2_CFD/include/solvers/CSpeciesFlameletSolver.hpp
```diff
@@ -105,6 +109,22 @@ class CSpeciesFlameletSolver final : public CSpeciesSolver {
...
+   */                                              
+  su2double ThickenedFlameCorrection(CGeometry const * geometry, const unsigned long iPoint) const;   
+
+  /*!
+   * \brief Approximate the minimum flame thickness value used for the thickened flame model.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \return - approximate flame thickness value.
+   */
+  su2double GetOverallFlameThickness(CGeometry * geometry,  CSolver **solver_container) const;
```
pcarruscag:
```suggestion
  su2double GetOverallFlameThickness(CGeometry* geometry,  CSolver** solver_container) const;
```

SU2_CFD/include/solvers/CSpeciesFlameletSolver.hpp
```diff
@@ -195,4 +215,10 @@ class CSpeciesFlameletSolver final : public CSpeciesSolver {
    */
   void Viscous_Residual(const unsigned long iEdge, const CGeometry* geometry, CSolver** solver_container, CNumerics* numerics,
                         const CConfig* config) final;
+
+  /*!
+   * \brief Obtain the overall flame thickness value.
+   * \return flame thickness value.
+   */
+  virtual su2double GetFlameThickness() const {return global_flame_thickness;}
```
pcarruscag:
```suggestion
  su2double GetFlameThickness() const override {return global_flame_thickness;}
```

SU2_CFD/src/output/CFlowOutput.cpp
```diff
@@ -1069,6 +1069,8 @@ void CFlowOutput::AddHistoryOutputFields_ScalarRMS_RES(const CConfig* config) {
         const auto& CV_name = flamelet_config_options.controlling_variable_names[iCV];
         AddHistoryOutput("RMS_"+CV_name, "rms["+CV_name+"]",ScreenOutputFormat::FIXED, "RMS_RES", "Root-mean squared residual of " + CV_name + " controlling variable equation.", HistoryFieldType::RESIDUAL);
       }
+      if (flamelet_config_options.thickenedflame_correction)
+        AddHistoryOutput("THICKNESS","flamethickness",ScreenOutputFormat::FIXED, "RMS_RES", "Flame thickness used for thickened flame correction model.", HistoryFieldType::RESIDUAL);
```
pcarruscag: It's not a residual

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+  su2double F{1.0};
+  if (global_flame_thickness != default_flame_thickness) {
+    su2double max_flame_vol = pow(global_flame_thickness, nDim);
+    F = max(1.0, geometry->nodes->GetVolume(iPoint) / max_flame_vol);
+  }
+  return F;
+}
+
+su2double CSpeciesFlameletSolver::GetOverallFlameThickness(CGeometry * geometry, CSolver **solver_container) const {
+  CFlowVariable const * flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
```
pcarruscag:
```suggestion
  const CFlowVariable* flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
   su2double pv_burnt = scalars[I_PROGVAR] - delta;
   return pv_burnt;
 }
+
+
+su2double CSpeciesFlameletSolver::ThickenedFlameCorrection(CGeometry const * geometry, const unsigned long iPoint) const {
```
pcarruscag: You need to open PRs more often so you don't forge what style we follow in the code 😉

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+  }
+  return F;
+}
+
+su2double CSpeciesFlameletSolver::GetOverallFlameThickness(CGeometry * geometry, CSolver **solver_container) const {
+  CFlowVariable const * flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
+  su2double pvmax_local{-1e3}, pvmin_local{1e3}, pvmax_global{0.0},pvmin_global{0.0}, gradpv_local{0.0}, gradpv_global{0.0}, Tmax_local{-1e6},Tmax_global{0.0};
+  
+  SU2_OMP_FOR_STAT(omp_chunk_size)
+  for (auto iPoint = 0u; iPoint < nPoint; iPoint++) {
```
pcarruscag:
```suggestion
  for (auto iPoint = 0u; iPoint < nPointDomain; iPoint++) {
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+      proj_grad_T_u *= mag_gradT;
+
+      /* Update minimum and maximum values. */
+      gradpv_local = max(gradpv_local, proj_grad_T_u);
+      pvmax_local = max(pvmax_local, pv_local);
+      pvmin_local = min(pvmin_local, pv_local);
+      Tmax_local = max(Tmax_local, T_local);
+  }
+  END_SU2_OMP_FOR
+  SU2_MPI::Barrier(SU2_MPI::GetComm());
```
pcarruscag:
```suggestion
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
pcarruscag: Is this temperature a tunable parameter?

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+        proj_grad_pv_u[iDim] = gradpv * val_u * val_u / (flowNodes->GetVelocity2(iPoint) + 1e-5);
+      }
+
+      /* Parallel projection of temperature gradient against projected progress variable gradient */
+      su2double proj_grad_T_u = 0, gradT2{0}, mag_gradT{0};
+      for (auto iDim=0u; iDim < nDim; iDim++) {
+        su2double gradT = flowNodes->GetGradient_Primitive(iPoint, prim_idx.Temperature(), iDim);
+        proj_grad_T_u += gradT * proj_grad_pv_u[iDim];
+        gradT2 += gradT*gradT;
+      }
```
pcarruscag: GeometryToolbox

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+    F = max(1.0, geometry->nodes->GetVolume(iPoint) / max_flame_vol);
+  }
+  return F;
+}
+
+su2double CSpeciesFlameletSolver::GetOverallFlameThickness(CGeometry * geometry, CSolver **solver_container) const {
+  CFlowVariable const * flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
+  su2double pvmax_local{-1e3}, pvmin_local{1e3}, pvmax_global{0.0},pvmin_global{0.0}, gradpv_local{0.0}, gradpv_global{0.0}, Tmax_local{-1e6},Tmax_global{0.0};
+  
+  SU2_OMP_FOR_STAT(omp_chunk_size)
```
pcarruscag: This does not work with OpenMP because the local variables are also thread-local, you can take a look at how we compute time steps for the pattern you need to use.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -804,3 +840,61 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+      pvmax_local = max(pvmax_local, pv_local);
+      pvmin_local = min(pvmin_local, pv_local);
+      Tmax_local = max(Tmax_local, T_local);
+  }
+  END_SU2_OMP_FOR
+  SU2_MPI::Barrier(SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&gradpv_local, &gradpv_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&pvmin_local, &pvmin_global, 1, MPI_DOUBLE, MPI_MIN, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&pvmax_local, &pvmax_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&Tmax_local, &Tmax_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
```
pcarruscag: A single reduction over 4 values would be more performant than 4 separate reductions.

EvertBunschoten:
> Nice work! Is there also no correction when the mesh is sufficiently fine?

No correction is applied if the length scale of the cell is lower than the calculated flame length scale.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
EvertBunschoten: More PRs coming in! You refer to the const and pointer placement?

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
EvertBunschoten: Now it is

SU2_CFD/src/output/CFlowOutput.cpp
EvertBunschoten: I have placed the flame thickness under its own category as a user-defined coefficient.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
EvertBunschoten: turns out it is!

bigfooted:
@EvertBunschoten Is the OMP comment of @pcarruscag the only one that's still open?

EvertBunschoten:
> @EvertBunschoten Is the OMP comment of @pcarruscag the only one that's still open?

Yes and I still need to create a regression test.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
EvertBunschoten: I have replaced the pattern with `SU2_OMP_FOR_(schedule(static,omp_chunk_size) SU2_NOWAIT)` like how it is used in the `SetTime_Step` functions.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -886,15 +909,24 @@ su2double CSpeciesFlameletSolver::GetOverallFlameThickness(CGeometry * geometry,
       Tmax_local = max(Tmax_local, T_local);
   }
   END_SU2_OMP_FOR
-  SU2_MPI::Barrier(SU2_MPI::GetComm());
-  SU2_MPI::Allreduce(&gradpv_local, &gradpv_global, 1, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  su2double* MyFlameThickness = new su2double[3];
+  su2double* TotalFlameThickness = new su2double[3]; 
+  MyFlameThickness[0] = gradpv_local;
+  MyFlameThickness[1] = pvmax_local;
+  MyFlameThickness[2] = Tmax_local;
```
pcarruscag: Use static size arrays, you know the count,  no use using `new/delete`.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -530,8 +563,7 @@ void CSpeciesFlameletSolver::BC_ConjugateHeat_Interface(CGeometry* geometry, CSo
 }
 
 unsigned long CSpeciesFlameletSolver::SetScalarSources(const CConfig* config, CFluidModel* fluid_model_local,
-                                                       unsigned long iPoint, const vector<su2double>& scalars) {
-  SU2_ZONE_SCOPED
```
pcarruscag, line 617: Add this back please

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -830,3 +863,70 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+  if (fabs(global_flame_thickness - default_flame_thickness) > EPS * max(1.0, fabs(default_flame_thickness))) {
+    su2double max_flame_vol = pow(global_flame_thickness, nDim);
+    F = max(1.0, geometry->nodes->GetVolume(iPoint) / max_flame_vol);
+  }
+  return F;
+}
+
+su2double CSpeciesFlameletSolver::GetOverallFlameThickness(CGeometry* geometry, CSolver** solver_container) const {
+  const CFlowVariable* flowNodes = su2staticcast_p<CFlowVariable*>(solver_container[FLOW_SOL]->GetNodes());
+  su2double pvmax_local{-1e3}, pvmin_local{1e3}, pvmax_global{0.0},pvmin_global{1e3}, gradpv_local{0.0}, gradpv_global{0.0}, Tmax_local{-1e6},Tmax_global{0.0};
```
pcarruscag: The "global" variables are not global by OpenMP standards unless they are static or class members.
I think in the time step example, they are class members.
```suggestion
  su2double pvmax_local{-1e3}, pvmin_local{1e3}, gradpv_local{0.0}, Tmax_local{-1e6};
  
  BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS {
    pvmax_global = -1e3;
    ...
  }
  END_SU2_OMP_SAFE_GLOBAL_ACCESS
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -830,3 +863,70 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+      /* Parallel projection of temperature gradient against projected progress variable gradient */
+      su2double gradT[MAXNDIM]={0};
+      for (auto iDim=0u; iDim < nDim; iDim++) 
+        gradT[iDim] = flowNodes->GetGradient_Primitive(iPoint, prim_idx.Temperature(), iDim);
+      
+      su2double proj_grad_T_u = GeometryToolbox::DotProduct(nDim, gradT, proj_grad_pv_u);
+      su2double mag_gradT = GeometryToolbox::Norm(nDim, gradT);
+      
+      proj_grad_T_u /= (pow(mag_gradT,2) +  EPS);
+      proj_grad_T_u *= mag_gradT;
```
pcarruscag:
```suggestion
      proj_grad_T_u /= max(mag_gradT,  EPS);
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -830,3 +863,70 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+
+      /* Update minimum and maximum values. */
+      gradpv_local = max(gradpv_local, proj_grad_T_u);
+      pvmax_local = max(pvmax_local, pv_local);
+      pvmin_local = min(pvmin_local, pv_local);
+      Tmax_local = max(Tmax_local, T_local);
+  }
+  END_SU2_OMP_FOR
```
pcarruscag, line 1002:
```suggestion
  END_SU2_OMP_FOR
  
  atomicMax(gradpv_local, gradpv_global);
  ...
```

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -830,3 +863,70 @@ su2double CSpeciesFlameletSolver::GetBurntProgressVariable(CFluidModel* fluid_mo
...
+  su2double* MyFlameThickness = new su2double[3];
+  su2double* TotalFlameThickness = new su2double[3]; 
+  MyFlameThickness[0] = gradpv_local;
+  MyFlameThickness[1] = pvmax_local;
+  MyFlameThickness[2] = Tmax_local;
+  
+  SU2_MPI::Allreduce(MyFlameThickness, TotalFlameThickness, 3, MPI_DOUBLE, MPI_MAX, SU2_MPI::GetComm());
+  SU2_MPI::Allreduce(&pvmin_local, &pvmin_global, 1, MPI_DOUBLE, MPI_MIN, SU2_MPI::GetComm());
```
pcarruscag: Then this is done just by the master thread like in the time step example.

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
pcarruscag: And add SU2_ZONE_SCOPED to the new function please

bigfooted:
@EvertBunschoten I think this is basically done besides some comments from Pedro. Can it be finalized?

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
EvertBunschoten: I tried making them class members, but that did not compile. I set them to static variables and that worked.

EvertBunschoten:
> @EvertBunschoten I think this is basically done besides some comments from Pedro. Can it be finalized?

I implemented all the suggestions, but two AD regression tests are failing because the 6th digit is wrong :(

bigfooted:
That's interesting, they are the same cases that fail for Berk's PR.

bigfooted:
@EvertBunschoten  can you fix the regression tests and the conflict? I think we can then merge this, @pcarruscag ?

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
pcarruscag: Needs to be inside BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS

EvertBunschoten: I put these two lines inside the same BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS scope where the global maximum values for the temperature, progress variable, and gradient are assigned.

pcarruscag:
Omp stuff looks ok.

bigfooted:
Can you add a testcase? I think it's complete then.

EvertBunschoten:
> Can you add a testcase? I think it's complete then.

There is a test case under flamelet/08_flame_init_method which demonstrates initialization of the flame front with the SPARK and FLAME_FRONT methods.

bigfooted (approved):
LGTM

PR #2735, opened 2026-02-22
Fix cli11 cstdint

kanakaero:
## Proposed Changes
Add a missing <cstdint> include to the vendored CLI11 header. 

GCC 14+ no longer permits transitive inclusion of fixed-width integer types. CLI11.hpp uses uint64_t without including \<cstdint\>, which causes build failures with GCC 15.

## Related Work
None

## PR Checklist

### DIFF
--- a/externals/CLI11/CLI11.hpp
+++ b/externals/CLI11/CLI11.hpp
@@ -62,6 +62,7 @@
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include <cstdint>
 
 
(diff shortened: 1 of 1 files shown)


pcarruscag:
Thanks for the fix.
We should try to use the current version of CLI11 and integrate it as a submodule like we do with other github-based dependencies. If you are looking for simple things to do in SU2, this would be one.

kanakaero:
Yeah, that would be a better long term approach, I will look into it.

PR #2749, opened 2026-03-07
Fix: patch bundled Boost hash.hpp to remove std::unary_function for C++17 compatibility

### LINKED ISSUE #2583: C++17 and Boost generates an error for unary_function in external submodule
Reporting here in case anyone encounters this issue in future. In C++ 17 the function std::unary_function was discontinued, there is one line in the boost library of tecio that uses this and causes a compilation error (at least on MacOS). The solution is very simple and this can be resolved by changing the function to std::__unary_function. I have observed this and it has also been encountered by at least one other.

**Bug report checklist**
*Please make sure that you have followed the checklist below, many common problems can be solved by:*

 - OS: MacOS
 - C++ compiler and version: XCode/clang
 - MPI implementation and version: 
 - SU2 Version: v8.3.0

Sahilll10:
## Root Cause

`boost/container_hash/hash.hpp` uses a guard `#if defined(_HAS_AUTO_PTR_ETC)`
which is MSVC-only. On MacOS/clang and GCC with `-std=c++17`, this guard never
activates, so `hash_base` inherits from `std::unary_function<T, std::size_t>`
which was removed in C++17 — causing a build failure.

## Fix

Patched the `#else` branch of `hash_base` in the bundled Boost headers inside
`externals/tecio/boost.tar.gz` to use explicit `typedef` declarations instead
of `std::unary_function` inheritance. This is identical to what the `#if` branch
already does correctly for MSVC. No behaviour change whatsoever.

## Verification

After repacking, confirmed zero occurrences of `unary_function` remain in
`boost/container_hash/hash.hpp`.

## Files Changed
- `externals/tecio/boost.tar.gz` — one struct patched in `boost/container_hash/hash.hpp`

Fixes #2583

Please let me know if there are further adjustments required.


pcarruscag:
You opened the PR against the wrong branch, I advise you to read the PR checklist and the contributing guidelines.

PR #2750, opened 2026-03-08
Fix: patch bundled Boost hash.hpp for C++17 std::unary_function removal

### LINKED ISSUE #2583: C++17 and Boost generates an error for unary_function in external submodule

Sahilll10:
## Root Cause
`boost/container_hash/hash.hpp` has a guard `#if defined(_HAS_AUTO_PTR_ETC)`
which is MSVC-only. On MacOS/clang and GCC with C++17, this guard never
activates, so `hash_base` inherits from `std::unary_function` which was
removed in C++17, causing a build failure.

## Fix
Patched the `#else` branch of `hash_base` in `externals/tecio/boost.tar.gz`
to use explicit `typedef` declarations instead of `std::unary_function`
inheritance. This matches what the `#if` branch already does correctly for
MSVC. No behaviour change.

## Verification
Zero occurrences of `unary_function` remain in
`boost/container_hash/hash.hpp` after repacking.

## Files Changed
- `externals/tecio/boost.tar.gz` — one struct patched in `boost/container_hash/hash.hpp`

PR Checklist
[ * ] I am submitting my contribution to the develop branch.
[ * ] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
[ * ] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
[  ] I used the pre-commit hook to prevent dirty commits and used pre-commit run --all to format old commits.
[  ] I have added a test case that demonstrates my contribution, if necessary.
[  ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

Fixes #2583


joshkellyjak:
~~You can still use unary_functions, you just need to declare them as std::__unary_function. I think this a better way to do this as I suggest in the original issue. This way we are consistent with the original implementation in tecio's boost library.~~

Actually, I changed my mind, if this is fully deprecated in C++ 17 then the changes are probably better in the long run. LGTM

Sahilll10:
Thanks for the review @joshkellyjakcom 
I've also merged the latest develop into this branch to resolve any conflicts before review. The only file touched remains externals/tecio/boost.tar.gz — one struct patched, nothing else.

I believe this is ready — the fix is minimal, targeted, and consistent with how the #if branch already handles this. But I want to make sure I'm not missing anything before this gets merged.

Let me know if there's anything else needed to get this across the line? 

I want to make sure this meets the project's standards.

pcarruscag (changes requested):
I'm afraid I can't accept an opaque change to a binary file from a first-time contributor.
If there is no other way to apply the patch, I'll wait for Josh to look at the changes **thoroughly**.

joshkellyjak:
> I'm afraid I can't accept an opaque change to a binary file from a first-time contributor. If there is no other way to apply the patch, I'll wait for Josh to look at the changes **thoroughly**.

Hi Pedro, I have looked through the changes, these are the only differences between the two versions of boost as from this branch and from current develop. These two are functionally equivalent and so, in my opinion, this is good to go, assuming the reg tests complete.

`struct hash_base : std::__unary_function<T, std::size_t> {};`

`struct hash_base
{
       typedef T argument_type;
       typedef std::size_t result_type;
};`

EDIT: Formatting

pcarruscag (approved):

PR #2774, opened 2026-03-31
Feature: Add Iteration Wall-Clock Timing (ITER_TIME)

Soumyadipta-Banerjee:
## Proposed Changes
This pull request introduces the `ITER_TIME` field, which prints the physical wall-clock time elapsed during the inner iteration step as an optional diagnostic column.
- Uses `SU2_MPI::Wtime()` synchronously across ranks to capture precise delta times between iteration steps.
- Designed as an opt-in metric via `SCREEN_OUTPUT` or `HISTORY_OUTPUT` configuration properties. It does not clutter the default console arrays, protecting the `TestCases` baseline reference setups in the CI pipeline.
- Handles padding organically aligned to `ScreenOutputFormat::FIXED` to retain the tight aesthetic tracking without jitter.

## Related Work
N/A

## PR Checklist

### DIFF
--- a/SU2_CFD/src/output/COutput.cpp
+++ b/SU2_CFD/src/output/COutput.cpp
@@ -2088,8 +2088,6 @@ void COutput::SetCommonHistoryFields() {
   AddHistoryOutput("OUTER_ITER", "Outer_Iter", ScreenOutputFormat::INTEGER, "ITER", "Outer iteration index");
   /// DESCRIPTION: The inner iteration index.
   AddHistoryOutput("INNER_ITER", "Inner_Iter", ScreenOutputFormat::INTEGER,  "ITER", "Inner iteration index");
-  /// DESCRIPTION: The current iteration calculation time.
-  AddHistoryOutput("ITER_TIME", "Time(s)", ScreenOutputFormat::FIXED, "ITER", "Time per iteration (s)");
   /// END_GROUP
 
   /// BEGIN_GROUP: TIME_DOMAIN, DESCRIPTION: Time integration information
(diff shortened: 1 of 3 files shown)


SU2_CFD/src/output/COutput.cpp
```diff
@@ -2087,6 +2089,8 @@ void COutput::SetCommonHistoryFields() {
   AddHistoryOutput("OUTER_ITER", "Outer_Iter", ScreenOutputFormat::INTEGER, "ITER", "Outer iteration index");
   /// DESCRIPTION: The inner iteration index.
   AddHistoryOutput("INNER_ITER", "Inner_Iter", ScreenOutputFormat::INTEGER,  "ITER", "Inner iteration index");
+  /// DESCRIPTION: The current iteration calculation time.
+  AddHistoryOutput("ITER_TIME", "Time(s)", ScreenOutputFormat::FIXED, "ITER", "Time per iteration (s)");
```
pcarruscag: Put it together, and in the same group as WALL_TIME

SU2_CFD/src/output/COutput.cpp
```diff
@@ -237,6 +238,7 @@ void COutput::SetHistoryOutput(CGeometry ****geometry, CSolver *****solver, CCon
     Iter = OuterIter;
 
   /*--- Turbomachinery Performance Screen summary output---*/
+  LoadCommonHistoryData(config[val_iZone]);
```
pcarruscag: Why was this added?

Soumyadipta-Banerjee: I initially added it because the timer wasn't appearing in the specialized turbo screen summary tables, so I thought it was missing for turbo cases. I've since realized the base class handles the history data correctly for every zone, so this call was redundant. I've removed it.

pcarruscag (approved):

PR #2782, opened 2026-04-10
Use FGCRODR instead of GMRES for "Newton-Krylov" adjoint

pcarruscag:
## Proposed Changes
It allows using a smaller restart period than GMRES, especially for more challenging problems.
It's been used, or at least researched, in Tau, and in https://arc.aiaa.org/doi/abs/10.2514/6.2026-2015

### TLDR

These are the relevant options:
```
% Adjoint GMRES settings.
MULTIZONE= YES
NEWTON_KRYLOV= YES
QUASI_NEWTON_NUM_SAMPLES= 999
% Acts as the restart frequency.
INNER_ITER= 50
% This should be ~1/6 to ~1/4 of the restart frequency. After the first outer
% iteration, the solver does "frequency - deflation" inner iterations.
LINEAR_SOLVER_RESTART_DEFLATION= 12
```
FGCRODR with 0 deflation vectors (`LINEAR_SOLVER_RESTART_DEFLATION= 0`) is equivalent to GMRES.

### Comparison

With a restart period N, and M deflation vectors, FGCRODR does about the same number of matrix-vector products as GMRES(N - M) and uses the same memory as GMRES(N + 2M).
**Note** that the current implementation is not fully optimized for memory use, the memory equivalence can be lowered to N + M.
In any case, I'll compare apples to apples as it is now (N + 2M). We can look at convergence performance, **residual vs wall time**, for the "same matrix-vector products" and "same memory usage" bounds. The former also corresponds to about the same wall time per outer iteration.

**30P30N (single zone RANS)**
With rough best practices and without tuning, the performance is approximately equivalent at the "memory usage" bound.
image

Note that convergence stalled for GMRES(30) above. If we try to use the smallest restart period that allows GMRES to converge well (which is important for larger problems to limit memory usage). We get a time-to-solution benefit at the "memory usage" bound.
image

**Small FSI case ("Airfoil 2D" regression, multizone, GMRES only on fluid)**
Here, there is a clear benefit, perhaps because the overall convergence rate is also limited by the FSI coupling, and thus any additional convergence from GMRES(60) is "wasted".
image

## PR Checklist


pcarruscag:
Updated the description.

PR #2786, opened 2026-04-12
Avoid recomputing residual norms for MG early exit

pcarruscag:
I think the norm should include the MG part of the residual, because it is the sum of the 2 that the solver is trying to converge.


SU2_CFD/src/solvers/CSolver.cpp
```diff
@@ -1963,7 +1963,7 @@ void CSolver::AdaptCFLNumber(CGeometry **geometry,
 void CSolver::SetResidual_RMS(const CGeometry *geometry, const CConfig *config) {
   SU2_ZONE_SCOPED
 
-  if (geometry->GetMGLevel() != MESH_0) return;
+  if (geometry->GetMGLevel() != MESH_0 && !config->GetMGOptions().MG_Smooth_EarlyExit) return;
```
bigfooted, line 1966: We do not need to set the RMS residual on the fine mesh when we do an early exit because we already call computelinsysresRMS?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -596,33 +602,28 @@ void CMultiGridIntegration::PreSmoothing(unsigned short RunTime_EqSystem,
...
       /*--- Capture initial RMS after the very first residual evaluation.
-       *    This is the earliest point where LinSysRes = R(u_current) (not stale).
-       *    ComputeLinSysResRMS must be called by all threads (uses parallel dot). ---*/
+       *    This is the earliest point where LinSysRes = R(u_current) (not stale). ---*/
       if (iPreSmooth == 0 && iRKStep == 0) {
-        const passivedouble initial_rms = ComputeLinSysResRMS(solver_fine);
-        BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS
-        {
+        BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS {
+          const passivedouble initial_rms = ComputeLinSysResRMS(solver_fine);
```
bigfooted, line 612: ok that makes sense, so space+time gives complete update, then we compute residual of this

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -596,33 +602,28 @@ void CMultiGridIntegration::PreSmoothing(unsigned short RunTime_EqSystem,
...
-    /*--- Early exit: check if RMS has dropped sufficiently.
-     *    ComputeLinSysResRMS must be called by all threads.
-     *    only master uses the result inside the safe block. ---*/
+    /*--- Early exit: check if RMS has dropped sufficiently. ---*/
     if (early_exit) {
-      const passivedouble current_rms = ComputeLinSysResRMS(solver_fine);
-      BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS
-      {
+      BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS {
+        const passivedouble current_rms = ComputeLinSysResRMS(solver_fine);
```
bigfooted, line 626: ok, we need to compute this again because of the update with the boundary conditions in postprocessing?

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -904,17 +893,16 @@ void CMultiGridIntegration::SmoothProlongated_Correction(unsigned short RunTime_
...
   }
 
-  /*--- Record final correction norm for debugging output.
-   *    ComputeLinSysResRMS must be called by all threads (uses parallel dot). ---*/
-    const passivedouble final_corr_rms = ComputeLinSysResRMS(solver);
-    SU2_OMP_SAFE_GLOBAL_ACCESS(lastCorrecSmoothRMS[iMesh][1] = final_corr_rms;)
-
+  /*--- Record final correction norm for debugging output. ---*/
+  if (config->GetMGOptions().MG_Smooth_Output) {
+    const su2double res = sqrt(solver->LinSysRes.squaredNorm() / (nVar * geometry->GetGlobal_nPointDomain()));
```
bigfooted, line 898: this is not the same as linsysres?

SU2_CFD/src/solvers/CSolver.cpp
pcarruscag: What this does is compute the residual for all mesh levels, if MG early exit is on.
Before it was only doing the fine mesh.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
pcarruscag: This is as before, to check the convergence of the smoothing iterations.

SU2_CFD/src/integration/CMultiGridIntegration.cpp
pcarruscag: It is equivalent, but here it needs to be computed from scratch instead of grabbing the result from the solver because this function manipulates LinSysRes directly, not through the solver (Space + Time integration).

SU2_CFD/src/integration/CMultiGridIntegration.cpp
pcarruscag: Yeah I think it has to be like this, otherwise the residual you were using to check the convergence is not consistent.
Before your initial residual was just from space integration but the one to check convergence was space+time (time adds the MG part) that inconsistency is what is causing the regressions.

Can you run the test you were using to develop this to check if the convergence is ok?

TestCases/parallel_regression.py
```diff
@@ -539,7 +539,7 @@ def main():
     turb_naca0012_sst_expliciteuler.cfg_dir   = "rans/naca0012"
     turb_naca0012_sst_expliciteuler.cfg_file  = "turb_NACA0012_sst_expliciteuler.cfg"
     turb_naca0012_sst_expliciteuler.test_iter = 10
-    turb_naca0012_sst_expliciteuler.test_vals = [-3.532365, -3.157224, 3.743381, 1.124798, 0.501715, -float("inf")]
+    turb_naca0012_sst_expliciteuler.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -inf]
```
bigfooted:
```suggestion
    turb_naca0012_sst_expliciteuler.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -float("inf")]
```

TestCases/parallel_regression.py
```diff
@@ -586,7 +586,7 @@ def main():
     turb_naca0012_sst_restart_mg.cfg_file  = "turb_NACA0012_sst_multigrid_restart.cfg"
     turb_naca0012_sst_restart_mg.test_iter = 20
     turb_naca0012_sst_restart_mg.ntest_vals = 5
-    turb_naca0012_sst_restart_mg.test_vals = [-6.538846, -5.057149, 0.830238, -0.008740, 0.078171]
+    turb_naca0012_sst_restart_mg.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -inf]
```
bigfooted:
```suggestion
    turb_naca0012_sst_restart_mg.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -float("inf")]
```

TestCases/hybrid_regression.py
```diff
@@ -224,15 +224,15 @@ def main():
...
+    turb_naca0012_sst_fixedvalues.test_vals = [-10.448081, 0.773965, 1.022534, 0.040529, -2.383403]
     test_list.append(turb_naca0012_sst_fixedvalues)
 
     # NACA0012 (SST, explicit Euler for flow and turbulence equations)
     turb_naca0012_sst_expliciteuler           = TestCase('turb_naca0012_sst_expliciteuler')
     turb_naca0012_sst_expliciteuler.cfg_dir   = "rans/naca0012"
     turb_naca0012_sst_expliciteuler.cfg_file  = "turb_NACA0012_sst_expliciteuler.cfg"
     turb_naca0012_sst_expliciteuler.test_iter = 10
-    turb_naca0012_sst_expliciteuler.test_vals = [-3.532365, -3.157224, 3.743381, 1.124798, 0.501715, -float("inf")]
+    turb_naca0012_sst_expliciteuler.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -inf]
```
bigfooted:
```suggestion
    turb_naca0012_sst_expliciteuler.test_vals = [-3.157224, 3.743381, 1.124798, 0.501715, -float("inf")]
```

TestCases/parallel_regression.py
```diff
@@ -220,31 +220,31 @@ def main():
...
+    channel.test_vals = [-2.305865, 3.236424, 0.124945, 0.153985]
     test_list.append(channel)
 
     # NACA0012
     naca0012           = TestCase('naca0012')
     naca0012.cfg_dir   = "euler/naca0012"
     naca0012.cfg_file  = "inv_NACA0012_Roe.cfg"
     naca0012.test_iter = 20
-    naca0012.test_vals = [-4.314738, -3.845428, 0.307738, 0.024670]
+    naca0012.test_vals = [-4.427649, -3.952898, 0.319961, 0.022649]
```
bigfooted:
```suggestion
    naca0012.test_vals = [-8.373603, -7.829504, 0.335767, 0.023275]
```

TestCases/hybrid_regression.py
```diff
@@ -481,7 +481,7 @@ def main():
     gust_mesh_defo.cfg_dir   = "gust"
     gust_mesh_defo.cfg_file  = "gust_with_mesh_deformation.cfg"
     gust_mesh_defo.test_iter = 6
-    gust_mesh_defo.test_vals = [-1.844761, 0.001077, -0.000263]
+    gust_mesh_defo.test_vals = [-2.418805, 0.001653, -0.001246, 0.000403, -0.000593]
```
bigfooted:
```suggestion
    gust_mesh_defo.test_vals = [6.000000, 0.000000, -1.844761, 0.000990, -0.000275]
```

bigfooted:
A couple of testcases have very large differences in residuals, and one case is diverging. I will have to check these in more detail to see if they need different settings and produce reasonable results.

bigfooted:
OK, this compiles properly again, and it actually skips smoothing iterations again.

Common/src/CConfig.cpp
```diff
@@ -1986,15 +1986,21 @@ void CConfig::SetConfig_Options() {
   /*!\brief MG_SMOOTH_EARLY_EXIT\n DESCRIPTION: Enable early exit for MG smoothing when RMS drops below threshold. DEFAULT: NO \ingroup Config*/
   addBoolOption("MG_SMOOTH_EARLY_EXIT", MGOptions.MG_Smooth_EarlyExit, true);
   /*!\brief MG_SMOOTH_RES_THRESHOLD\n DESCRIPTION: Smoothing stops when current_rms < threshold * initial_rms. DEFAULT: 0.1 \ingroup Config*/
-  addDoubleOption("MG_SMOOTH_RES_THRESHOLD", MGOptions.MG_Smooth_Res_Threshold, 0.5);
+  addDoubleOption("MG_SMOOTH_RES_THRESHOLD", MGOptions.MG_Smooth_Res_Threshold, 0.9);
   /*!\brief MG_SMOOTH_OUTPUT\n DESCRIPTION: Print compact per-cycle smoothing iteration summary. DEFAULT: NO \ingroup Config*/
   addBoolOption("MG_SMOOTH_OUTPUT", MGOptions.MG_Smooth_Output, false);
+  /*!\brief MG_SMOOTH_STAGNATION_TOL\n DESCRIPTION: Early exit smoothing when step-to-step RMS ratio >= this value (0 = disabled). DEFAULT: 0.0 \ingroup Config*/
+  addDoubleOption("MG_SMOOTH_STAGNATION_TOL", MGOptions.MG_Smooth_StagnationTol, 0.99);
```
pcarruscag, line 1993: Commented default doesn't match the code

SU2_CFD/src/integration/CMultiGridIntegration.cpp
```diff
@@ -688,102 +599,99 @@ void CMultiGridIntegration::PostSmoothing(unsigned short RunTime_EqSystem,
...
+                mg_early_exit_flag = true;
+              } else if (defect >= mg_prev_smooth_rms * stag_tol) {
+                /*--- 'A' = amplifying-stagnation (defect grew vs previous step).
+                 *    'S' = clean stagnation (defect is not improving but also not growing). ---*/
+                lastPostSmoothExitReason[iMesh] = (defect > mg_prev_smooth_rms) ? 'A' : 'S';
+                mg_early_exit_flag = true;
+              }
+            }
+          }
+          mg_prev_smooth_rms = defect;
```
pcarruscag, line 633: Can you add a helper to share this between Pre and Post?

bigfooted:
taken over by another PR

PR #2791, opened 2026-04-20
Use Classical Gram–Schmidt twice instead of Modified G-S with conditional re-orthogonalization.

pcarruscag:
## Proposed Changes
Doing CGS twice is faster (if not implemented naively) and at least as robust.
+ other small linear algebra improvements.

## PR Checklist


bigfooted:
Nice! So it's faster per iteration, not faster because it converges in less iterations I guess? How much faster re we talking about?

pcarruscag:
Yes per iteration. Maybe 1.4x for the orthogonalization part, which is only significant for many linear iterations since it scales quadratically.
So this is mostly for Adjoints with GMRES and FEA/Heat problems.

PR #2210, opened 2024-02-14
[WIP] Pressure Based Flow Solver V8

NAnand-TUD:
## Proposed Changes
*Give a brief overview of your contribution here in a few sentences.*
Reimplemented pressure-based solver from feature_Pressure_based into the development branch.  

## Related Work
branch: feature_Pressure_based

## PR Checklist
*Put an X by all that apply. You can fill this out after submitting the PR. If you have any questions, don't hesitate to ask! We want to help. These are a guide for you to know what the reviewers will be looking for in your contribution.*

- [X ] I am submitting my contribution to the develop branch.
- [ ] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [ X] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [ ] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

Euler Solver: Incompressible cylinder case Pressure based (PB) vs Density Based (DB)
incomp_cylinder
incomp_cylinder_history

Navier Stokes Solver: Incompressible Hydrofoil case Pressure-based (PB) vs Density Based (DB)
NACA0012_hydrofoil
NACA0012_hydrofoil_history

Still TODOs
* Fixing RANS: Current CIncPBVariable inherits CVariable instead of CFlowVariable, if this is fixed RANS should work.
* Fixing MPI: The problem is currently unknown
* Fixing ND: non-dimensionalization is different in the two solvers. It should be harmonized.
* Multigrid: A PB multigrid is available but has not been tested.
* Small style changes.

PR #2812, opened 2026-05-07
Pressure Based Solver

thijsaalbers:
## Proposed Changes 

The current work (part of GSoC) provides a working version of a pressure-based algorithm for the incompressible flow solver as an alternative to the existing Density-based solver. Below, the reader may find the algorithm which has been implemented, as well as the current progress of the code and challenges. All the way at the bottom one can find performance comparisons between the DB and PB solvers for some test cases.

### Algorithm

A lot of versions of pressure-based algorithms exist, and many different versions can be implemented. Here, we opt for versions of the original SIMPLE/PISO algorithm, it is briefly defined here for clarity.

First, the momentum equations are solved, starting from the previous time step's velocity $\vec{u}^{(0)}$, pressure $p^{(0)}$, and face velocity $\vec{u}_f^{(0)}$. The resulting momentum is the predicted momentum, here its discretized form is shown, as its coefficients are used in the subsequent equations

$A_p(\rho\vec{u})_p^{(1)}+\sum_n A_n (\rho\vec{u})_n^{(1)}=-V\nabla p^{(0)}+S_m$

The subsequent momentum is not necessarily incompressible, the pressure correction equation can be derived by rewriting it as follows, using a term often called H by A

$(\rho\vec{u})_p^{(1)}=-\frac{\sum_n A_n (\rho\vec{u})_n^{(1)}}{A_p}-\frac{V}{A_p}\nabla p+S_m=\frac{H((\rho\vec{u})^{(1)})}{A}-\frac{V}{A_p}\nabla p^{(0)}+S_m$

$(\rho\vec{u})_p^{(2)}=\frac{H((\rho\vec{u})^{(1)})}{A}-\frac{V}{A_p}\nabla p^{(1)}+S_m$

Note how a simplification is used here where the HbyA term is neglected. Subtracting these two equations yields the first pressure correction equation for $p'$ as

$\nabla \cdot \left( \frac{V}{A_p}\nabla p'\right)=\nabla \cdot \left( \rho \vec{u}^{(1)}\right)$

Make note that for the divergence here, we require the face mass fluxes, which are computed using Rhie-Chow interpolation to avoid odd-even decoupling. After the equation is solved, using the pressure correction $p'$, the pressure and momentum are corrected according to

$p^{(1)} = p^{(0)} + p' ,\quad (\rho u)^{(2)} = (\rho u)^{(1)}+(\rho u)'.\quad (\rho u)'= -\frac{V}{A_p}\nabla p'$

So far, this is equal to a pseudo-transient version of the SIMPLE algorithm. This algorithm however suffers from a very tight stability condition on the time-step size. Therefore, multiple pressure corrections can be applied, which for two corrections is originally called the PISO algorithm.

The second pressure correction does not neglect the HbyA term, which then results in the equation

$\nabla \cdot \left( \frac{V}{A_p}\nabla p'\right)=\nabla \cdot \left( \rho \vec{u}^{(2)}\right)+\nabla\cdot\left(\frac{H((\rho u)')}{A}\right)$

And the new correction equations are defined as

$p^{(2)} = p^{(1)} + p' ,\quad (\rho u)^{(3)} = (\rho u)^{(2)}+(\rho u)'.\quad (\rho u)'= \frac{H((\rho u)')}{A}-\frac{V}{A_p}\nabla p'$

Note that HbyA here uses the previous velocity correction and is thus the same quantity as the one used in the second pressure correction equation. Later pressure correction equations follow analogously.

### Progress:

- Pressure-based solver added as alternative to density-based solver for the incompressible flow equations.
- The pressure-based solver has only been tested for constant density cases for basic Navier-Stokes and Euler flow.
- The pressure-based solver is implemented based on a pseudo time-stepping approach to remain consistent with the other solvers in SU2. The pressure-based solver is currently set to the SIMPLE algorithm by default, with options for SIMPLEC and PISO available. 
- The Poisson solver is a major bottleneck in the computation speed. To account for this, an option was added to use a different linear solver and preconditioner for the poisson solver. 
- For details on the implementation of the algorithm and the responsibility distribution please see the file CPBFluidIteration.cpp.
- A tutorial showcasing the different options of the PB solver for the lid driven cavity flow problem has been added, please see su2code/Tutorials#86 and su2code/su2code.github.io#218
- Regression tests have been  added.

#### Performance:

- The Poisson solver can sometimes struggle a lot due to high Reynolds numbers and fine meshes, and thus require a ridiculous number of iterations to converge reasonably. Possible fixes include adding multigrid support or a DIC preconditioner (far less efficient). Multigrid support is tricky as SU2 currently only considers multigrid for the main (flow) solver and not for auxiliary solvers.

- Convergence issues with RANS (SA and SST) on fine meshes with high Reynolds numbers. Tests have shown that cases such as flow over a flat plate converges fine. However, external aerodynamic cases such as the naca0012 RANS test case do not converge well at all. The convergence does slightly improve when we switch out the mesh for a more uniform unstructured mesh without large aspect ratio cells in the wake of the airfoil, although this only slightly helps. The flat plate turbulence test case also uses large aspect ratio cells so this is not the sole issue. The RANS solver also often requires many iterations of the Poisson solver to converge reasonably, this is however not the reason for the lack of convergence.

- Periodic boundary conditions have not been implemented/tested at all as of yet. 
- Any code related to adjoints has not been considered at all either.

#### Code:

- Parallelization with OMP gives wrong results due to a unknown issue in the Poisson solver, MPI however does work as expected.
- The restart solution currently does not write the edge mass fluxes which are used by the pressure-based solver. Therefore restarted solutions have to estimate these mass fluxes based on the average of the nodal solutions. This results in the restarted solutions starting from slightly different residuals. A fix for this issue can be to let the edge mass fluxes be stored in the restart file.
- Support for the energy equation is added trough both the enthalpy equation as well as trough the weakly coupled heat solver. This support has however not yet been verified trough actual test cases.
- The numerics class for the convective residuals write the continuity parts of the flux and jacobians. This will allow for a coupled solver in the future, but as it stands now this code is _not_ used.

### TODO list 

- Fix issues mentioned above (left for future work)

## Related Work

This work is based on earlier attempts by Nitish Anand (2024) and Akshay Koodly (2021), see feature branches feature_PBFlow_V8 and feature_Pressure_based respectively. Also see PR #2210

## PR Checklist

### Inviscid Hydrofoil

*Convergence history of the inviscid flow around a hydrofoil at a 5 degree aoa.*

*The pressure coefficient along the surface of the hydrofoil at a 5 degree aoa and the corresponding lift coefficients, X-FOIL predicts C_L=0.6.*

### Lid Driven Cavity 

*Convergence history of the lid driven cavity problem, note that CFL=60 is the highest stable CFL for the PB solver, whereas the DB solver does not have this CFL related stability issue.*

### Flatplate RANS

*The skin friction coefficient for turbulent flow over a (rough) flat plate with SA.*


SU2_CFD/src/numerics/poisson.cpp
pcarruscag, line 1: Please think about ways of reusing the numerics used by the scalar solvers, not duplicating code is part of the criteria for merging PRs.

SU2_CFD/src/numerics/pbflow.cpp
```diff
@@ -0,0 +1,573 @@
...
+  ProjFlux_i = new su2double [nVar];
+  ProjFlux_j = new su2double [nVar];
+  Jacobian_i = new su2double* [nVar];
+  Jacobian_j = new su2double* [nVar];
+  Jacobian_upw = new su2double* [nVar];
+  
+  for (iVar = 0; iVar < nVar; iVar++) {
+    Jacobian_i[iVar] = new su2double [nVar];
+    Jacobian_j[iVar] = new su2double [nVar];
+    Jacobian_upw[iVar] = new su2double [nVar];
```
pcarruscag: Please look at more recent examples of introducing new numerics classes or cleaning old ones, this is not the way for new code.

SU2_CFD/src/numerics/pbflow.cpp
```diff
@@ -0,0 +1,573 @@
...
+  Area = sqrt(Area);
+  
+  /*--- (rho*u_i) ---*/
+  Face_Flux = 0.0;
+  for (iDim = 0; iDim < nDim; iDim++) {
+    Velocity_i[iDim] = V_i[iDim+1];
+    Velocity_j[iDim] = V_j[iDim+1];
+    MeanVelocity[iDim] =  0.5*(Velocity_i[iDim] + Velocity_j[iDim]);
+    Face_Flux += MeanDensity*MeanVelocity[iDim]*Normal[iDim];
+  }
```
pcarruscag: This is not a correct implementation of Rhie-Chow momentum interpolation.
Please consider attending a developer meeting to explain your plans with this PR, the way I see it going, it's not going to make the cut to merge into develop.

thijsaalbers: Thanks for your feedback. I will let you know when it is ready for review.

pcarruscag: Ok. Keep in mind that we prefer many small PRs over one big one. This way, lessons learned from one review carry over to the next small PR.
Very large PRs tend to take a long time to review because the authors tend to run out of time rather than finish the work.
https://su2code.github.io/docs_v7/Code-Review/

bigfooted: @thijsaalbers please discuss with @vdweide if there can be a 'minimum example' PR of the implementation, one that can be delivered half-way in this GSOC project. Then add 'nice to have' features later.
For announcements on developer meetings and also for quick questions, please check slack.

pcarruscag: We should discuss architecture at a developers meeting... I'm getting dejavu :sweat_smile: 
image

thijsaalbers:
I will have to review the code myself first, it is not ready for review as of yet.

Common/include/CConfig.hpp
```diff
@@ -649,6 +660,8 @@ class CConfig {
   su2double SemiSpan;                   /*!< \brief Wing Semi span. */
   su2double MSW_Alpha;                  /*!< \brief Coefficient for blending states in the MSW scheme. */
   su2double Roe_Kappa;                  /*!< \brief Relaxation of the Roe scheme. */
+  su2double RCFactor;                   /*!< \brief Relaxation for the Rhie-Chow interpolation contribution. */
```
thijsaalbers: Should this variable still be named RCfactor? This factor is named RC to remain consistent with the thesis of Akshay, it however has no relation to Rhie-Chow at all and instead functions as a means of controlling the time dependent term on the momentum coefficients for the poisson equation... I do not know why he called it RC in his thesis, nor if we should keep this same name or change it.

SU2_CFD/include/solvers/CPoissonSolver.hpp
```diff
@@ -0,0 +1,288 @@
...
+   * \param[in] ExtIter - External iteration.
+   */
+  void SetInitialCondition(CGeometry **geometry,
+                           CSolver ***solver_container,
+                           CConfig *config,
+                           unsigned long TimeIter) override {}
+
+  /*!
+   * \brief Compute the coefficients for the pressure correction equation based
+   *        on the residuals from the solution of the momentum equation.
```
thijsaalbers, line 166: Still need to add \param[in] comments

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -2093,6 +2146,22 @@ void CDriver::InitializeNumerics(CConfig *config, CGeometry **geometry, CSolver
     }
   }
 
+  /*--- Solver definition for the poisson/pressure correction problem ---*/
+  if (poisson) {
+    //TODO: currently heat is reused which is wrong as the actual laplacian operator is anisotropic for the poisson problem
```
thijsaalbers:
```suggestion

```

SU2_CFD/src/numerics/flow/convection/pressure_based.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+CCentLinearPB_Flow::CCentLinearPB_Flow(unsigned short val_nDim, unsigned short val_nVar, CConfig *config) : CNumerics(val_nDim, val_nVar, config) {
+  
+  implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+  dynamic_grid = config->GetDynamic_Grid();
+  
+  Velocity_i = new su2double [nDim];
+  Velocity_j = new su2double [nDim];
+  MeanMassFlux = new su2double [nDim];
+  Flux = new su2double [nDim];
```
thijsaalbers:
```suggestion
  Velocity_i = new su2double [nVar];
  Velocity_j = new su2double [nVar];
  MeanMassFlux = new su2double [nVar];
  Flux = new su2double [nVar];
```

SU2_CFD/src/numerics/flow/convection/pressure_based.cpp
```diff
@@ -0,0 +1,241 @@
...
+CUpwPB_Flow::CUpwPB_Flow(unsigned short val_nDim, unsigned short val_nVar, CConfig *config) : CNumerics(val_nDim, val_nVar, config) {
+  
+  implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+  dynamic_grid = config->GetDynamic_Grid();
+  
+  Velocity_i = new su2double [nDim];
+  Velocity_j = new su2double [nDim];
+  Velocity_upw = new su2double [nDim];
+  MeanMassFlux = new su2double [nDim];
+  Flux = new su2double [nDim];
```
thijsaalbers:
```suggestion
  Velocity_i = new su2double [nVar];
  Velocity_j = new su2double [nVar];
  Velocity_upw = new su2double [nVar];
  MeanMassFlux = new su2double [nVar];
  Flux = new su2double [nVar];
```

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+#include "../../include/fluid/CIncIdealGas.hpp"
+#include "../../include/fluid/CIncIdealGasPolynomial.hpp"
+#include "../../include/variables/CPBIncNSVariable.hpp"
+#include "../../../Common/include/toolboxes/geometry_toolbox.hpp"
+#include "../../include/fluid/CFluidScalar.hpp"
+#include "../../include/fluid/CFluidFlamelet.hpp"
+#include "../../include/fluid/CFluidModel.hpp"
+
+// TODO: Temporarily directly copied from CIncEulerSolver's constructor to adhere to necessary
+// initializations, only real difference is the definition of the nVar,nprimVar etc
```
thijsaalbers: I do not know if this should be changed and if so what would be the best course of action.

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+}
+
+//TODO: this funciton is largely if not entirely the same as the one from cinceulersolver
+void CPBIncEulerSolver::SetNondimensionalization(CConfig *config, unsigned short iMesh) {
+  SU2_ZONE_SCOPED
+  
+  su2double Temperature_FreeStream = 0.0,  ModVel_FreeStream = 0.0,Energy_FreeStream = 0.0,
+  ModVel_FreeStreamND = 0.0, Omega_FreeStream = 0.0, Omega_FreeStreamND = 0.0, Viscosity_FreeStream = 0.0,
+  Thermal_Conductivity_FreeStream = 0.0, SpecificHeat_Cp_FreeStream = 0.0,
+  Density_FreeStream = 0.0, Pressure_FreeStream = 0.0, Pressure_Thermodynamic = 0.0, Tke_FreeStream = 0.0,
```
thijsaalbers: The same as i mentioned before, I do not know what would be the desirable method of unifying this with CIncEuler, as of now the only 'real' difference is the amount of fluid models available in either as the pressure based solver only considers constant density.

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+
+  if (config->GetKind_Gradient_Method() == GREEN_GAUSS) {
+    SetPrimitive_Gradient_GG(geometry, config);
+  }
+  if (config->GetKind_Gradient_Method() == WEIGHTED_LEAST_SQUARES) {
+    SetPrimitive_Gradient_LS(geometry, config);
+  }
+}
+
+//TODO: exact copy of CIncEulerSolver
```
thijsaalbers: See comment about nondimensionalization function

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+  END_SU2_OMP_FOR
+
+  AD::EndNoSharedReading();
+
+  return nonPhysicalPoints;
+
+}
+
+
+// TODO: exact copy of CIncEulerSolver
```
thijsaalbers: See comment about nondimensionalization function

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+    /*--- Compute the residual based on the pressure gradient. ---*/
+    for (iVar = 0; iVar < nVar; iVar++)
+      Vdpdx[iVar] = geometry->nodes->GetVolume(iPoint)*nodes->GetGradient_Primitive(iPoint,0,iVar);
+
+    auto residual = CNumerics::ResidualType<>(Vdpdx, nullptr, nullptr);
+
+    /*--- Add Residual to the total ---*/
+    LinSysRes.AddBlock(iPoint, residual);
+  }
+  END_SU2_OMP_FOR
```
thijsaalbers: Speaks for itself I think. Should this become a separate Source class similar to radiation or boussinesq source term or should something else be done?

SU2_CFD/src/solvers/CPBIncEulerSolver.cpp
```diff
@@ -0,0 +1,1661 @@
...
+
+  // TODO: Write all other possible source terms e.g. body force, axisymmetry etc. etc.
+
+}
+
+// TODO: The given time step works although it gives different values then the old version from PBFlowV8
```
thijsaalbers:
```suggestion
```

SU2_CFD/src/solvers/CPBIncNSSolver.cpp
```diff
@@ -0,0 +1,276 @@
...
+ * License along with SU2. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "../../include/solvers/CPBIncNSSolver.hpp"
+#include "../../include/variables/CPBIncNSVariable.hpp"
+#include "../../../Common/include/toolboxes/printing_toolbox.hpp"
+#include "../../include/solvers/CFVMFlowSolverBase.inl"
+
+//TODO: almost everything in this file is a direct or indirect copy of the code in 
+// the CIncNSSolver class. this will need to be refactored.
```
thijsaalbers: The only really different thing in this file from the CIncNSSolver is the wall boundary condition defintion (which is just massively simplified). I again do not know how this should be unified with the CIncNSSolver class as the C..IncEuler classes in between are massively distinct.

SU2_CFD/src/solvers/CTurbSolver.cpp
```diff
@@ -113,6 +113,8 @@ void CTurbSolver::LoadRestart(CGeometry** geometry, CSolver*** solver, CConfig*
 
     unsigned short skipVars = nDim + solver[MESH_0][FLOW_SOL]->GetnVar();
 
+    if (config->GetKind_Incomp_System() == ENUM_INCOMP_SYSTEM::PRESSURE_BASED) skipVars+=2; // skip pressure
```
thijsaalbers: This works, although i do not know why +2 and not +1...

SU2_CFD/src/solvers/CPoissonSolver.cpp
```diff
@@ -0,0 +1,764 @@
...
+  unsigned long iPoint, jPoint, iNeigh;
+  su2double Mom_Coeff, Mom_Coeff_nb, Vol, delT;
+  bool simplec = (config->GetKind_PBIter() == ENUM_PBITER::SIMPLEC);
+  bool implicit = (config->GetKind_TimeIntScheme_Flow() == EULER_IMPLICIT);
+
+  const CSolver* flow_solution = solver_container[FLOW_SOL];
+  const CVariable* flow_nodes = flow_solution->GetNodes();
+  
+
+  if (!periodic)  {
```
thijsaalbers:
```suggestion
```

SU2_CFD/src/solvers/CPoissonSolver.cpp
```diff
@@ -0,0 +1,764 @@
...
+        Mom_Coeff = delT;
+
+        nodes->SetMomCoeff(iPoint, Mom_Coeff);
+      }
+    }
+
+    /*--- Insert MPI call here. ---*/
+    InitiateComms(geometry, config, MPI_QUANTITIES::MOM_COEFF);
+    CompleteComms(geometry, config, MPI_QUANTITIES::MOM_COEFF); 
+  }
```
thijsaalbers, line 216:
```suggestion
```

thijsaalbers (commented):
The current version of the solver is ready for review. It is however not yet finished and there is a list of known issues that I am still working through. I have updated the PR description with the current status and known limitations; please read that before reviewing the implementation.

pcarruscag (commented):
There are some errors in the implementation, which I made myself when I first implemented SIMPLE, and they are easy to make because certain simplifications look equivalent but are not.
For professional reasons I should discuss this only in terms of my prior work.
Here is an implementation that I think is mostly correct: https://github.com/pcarruscag/Flow-Solver-Experiments/blob/main/lib/flow/src/flow_simple.cpp
It's written in a funny way because it was my first exercise into porting something to GPUs, but follow the operations, and their order, not the code itself.
The main error here is that you do not carry nor correct the face velocities (or face mass flows), nor use them to discretize momentum and scalars.
You correct the nodal velocities (equivalent to cell centers) but that is not the same thing.
I was not nice enough to my future self to document why in my old code, but I'm somewhat confident an AI can explain the nuance.

thijsaalbers:
@pcarruscag Thanks for the very usefull explanation of the issue. I think I now more or less fixed it. I also rewrote CPBFluidIteration.cpp, and I think the flow of the algorithm (and the changes I now made to it) should be clear in there.

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -687,10 +687,11 @@ class CSysMatrix {
    *            blocks of the matrix. Solvers that perform multiple updates would lose too much information, so
    *            that pattern is not supported with quantization (the code will hit null pointers). It is up to
    *            the solver to declare whether it will "set and forget".
+   * \param[in] poisson - If preconditioner should use settings defined for the poisson solver.
    */
   void Initialize(unsigned long npoint, unsigned long npointdomain, unsigned short nvar, unsigned short neqn,
                   bool EdgeConnect, CGeometry* geometry, const CConfig* config, bool needTranspPtr = false,
-                  bool grad_mode = false, bool allow_quant = false);
+                  bool grad_mode = false, bool allow_quant = false, bool poisson = false);
```
pcarruscag: Instead of grad_mode and poisson, we should just pass the enum value that we want to override.
Using a std::optional<Preconditioner>.

Common/include/CConfig.hpp
```diff
@@ -135,7 +135,8 @@ class CConfig {
   Hold_GridFixed,           /*!< \brief Flag hold fixed some part of the mesh during the deformation. */
   Axisymmetric,             /*!< \brief Flag for axisymmetric calculations */
   Enable_Cuda,              /*!< \brief Flag for switching GPU computing*/
-  Integrated_HeatFlux;      /*!< \brief Flag for heat flux BC whether it deals with integrated values.*/
+  Integrated_HeatFlux,      /*!< \brief Flag for heat flux BC whether it deals with integrated values.*/
+  Pressure_based;           /*!< \brief FLag to check if we are using a pressure-based system.*/
```
```suggestion
  Pressure_Based;           /*!< \brief Flag to check if we are using a pressure-based system.*/
```

Common/include/CConfig.hpp
```diff
@@ -4420,6 +4477,30 @@ class CConfig {
...
+   * \brief Get the coefficient for the removal of the transient term in the poisson solver coefficients.
+   * \return coefficient for the removal of the transient term in the poisson solver coefficients.
+   */
+  su2double GetTransient_Term_Removal_Factor(void) const  { return Transient_Term_Removal_Factor; }
+
+  /*!
+   * \brief Verify if there is mixing plane interface specified from config file.
+   * \return boolean.
+   */
+  bool GetBoolAutomaticRelaxationFactors(void) const { return AutomaticRelaxationFactors; }
```
pcarruscag: Bundle the SIMPLE-specific options in a struct like we do for turbulence models and ILU options please, that way there is a singe funtion to return them.

Common/include/option_structure.hpp
```diff
@@ -339,6 +340,31 @@ static const MapType<std::string, STRUCT_COMPRESS> MatComp_Map = {
...
+ * \brief Type of incompressible solver
+ */
+enum class ENUM_INCOMP_SYSTEM {
+  DENSITY_BASED,        /*!< \brief Density-based. */
+  PRESSURE_BASED,       /*!< \brief Pressure-based. */
+};
+static const MapType<std::string, ENUM_INCOMP_SYSTEM> Incomp_Map = {
+ MakePair("DENSITY_BASED", ENUM_INCOMP_SYSTEM::DENSITY_BASED)
+ MakePair("PRESSURE_BASED", ENUM_INCOMP_SYSTEM::PRESSURE_BASED)
+};
```
pcarruscag: This classification is not correct, the incompressible solver solves for pressure, not density, hence it is not a density-based solver, it also does not use artificial compressibility.

Rhie-Chow is a spatial discretization scheme, make that an option for the flow convective it scheme, and then that triggers the use of the segregated solver.

Common/include/option_structure.hpp
```diff
@@ -2769,6 +2802,9 @@ enum class MPI_QUANTITIES {
   MESH_DISPLACEMENTS   ,  /*!< \brief Mesh displacements at the interface. */
   SOLUTION_TIME_N      ,  /*!< \brief Solution at time n. */
   SOLUTION_TIME_N1     ,  /*!< \brief Solution at time n-1. */
+  MOM_COEFF            ,  /*!< \brief Momentum coefficient for the pressure-based poisson solver. */
```
pcarruscag:
```suggestion
  MOM_COEFF            ,  /*!< \brief Momentum coefficient for the Rhie-Chow scheme. */
```

SU2_CFD/include/numerics/CNumerics.hpp
```diff
@@ -1857,6 +1859,12 @@ class CNumerics {
    */
   inline void SetMassFlux(const su2double val_MassFlux) { MassFlux = val_MassFlux; }
 
+  /*!
+   * \brief SetEdgeVelocity
+   * \param[in] val_EdgeVelocity: Velocity across the edge
+   */
+  inline void SetEdgeVelocity(const su2double* val_EdgeVelocity) { EdgeVelocity = val_EdgeVelocity; }
```
pcarruscag: Can you re-use MassFlux please?

SU2_CFD/include/solvers/CPoissonSolver.hpp
```diff
@@ -0,0 +1,288 @@
...
+   * \brief No dual time stepping as there is no time stepping at all.
+   * \param[in] geometry - Geometric definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] iRKStep - Current step of the Runge-Kutta iteration.
+   * \param[in] iMesh - Index of the mesh in multigrid computations.
+   * \param[in] RunTime_EqSystem - System of equations which is going to be solved.
+   */
+  void SetResidual_DualTime(CGeometry* geometry, CSolver** solver_container, CConfig* config, unsigned short iRKStep,
+                            unsigned short iMesh, unsigned short RunTime_EqSystem) override {}
```
pcarruscag, line 162: I don't thing you need to declare these if they are empty?
If you need to, don't bother with the argument comments since it's a no-op anyway

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -218,6 +219,33 @@ CIncEulerSolver::CIncEulerSolver(CGeometry *geometry, CConfig *config, unsigned
...
+    for (auto color : EdgeColoring) {
+      /*--- Chunk size is at least OMP_MIN_SIZE and a multiple of the color group size. ---*/
+      SU2_OMP_FOR_DYN(nextMultiple(OMP_MIN_SIZE, color.groupSize))
+      for(auto k = 0ul; k < color.size; ++k) {
+        auto iEdge = color.indices[k];
+        for (unsigned short iDim = 0; iDim < nDim; ++iDim) 
+          EdgeVelocity[iEdge][iDim] = Velocity_Inf[iDim];
+      }
+      END_SU2_OMP_FOR
+    } // end color loop
```
pcarruscag: You don't need the coloring for this, looping plainly over edges is fine,

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1015,6 +1043,22 @@ void CIncEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_
...
+  if (pressure_based || config->GetViscous()) {
+    /*--- Compute gradient of the primitive variables (required for pressure source term) ---*/
+
+    if (config->GetKind_Gradient_Method() == GREEN_GAUSS) {
+      SetPrimitive_Gradient_GG(geometry, config);
+    }
+    else if (config->GetKind_Gradient_Method() == WEIGHTED_LEAST_SQUARES) {
+      SetPrimitive_Gradient_LS(geometry, config);
+    }
+  }
```
pcarruscag: Gradients are computed in ::Preprocessing, why is this needed here too?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1124,54 +1168,98 @@ void CIncEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contain
...
+
+      FORCEINLINE su2double operator() (const CIncEulerVariable& nodes, unsigned long iPoint) const {
+        su2double thermalCond = nodes.GetThermalConductivity(iPoint);
+        su2double laminarVisc = nodes.GetLaminarViscosity(iPoint);
+        su2double eddyVisc = nodes.GetEddyViscosity(iPoint);
+        su2double density = nodes.GetDensity(iPoint);
+        return lambda(laminarVisc, eddyVisc, density);
+      }
+
+    } lambdaVisc;
```
pcarruscag: No duplication please, define outside of the if/else

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1477,6 +1575,36 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+        pressureGradientSource[iDim+1] = geometry->nodes->GetVolume(iPoint)*nodes->GetGradient_Primitive(iPoint,0,iDim) / Density;
+      
+
+      auto residual = CNumerics::ResidualType<>(pressureGradientSource, nullptr, nullptr);
+
+      /*--- Add Residual to the total ---*/
+      LinSysRes.AddBlock(iPoint, residual);
+
+    }
+    END_SU2_OMP_FOR
```
pcarruscag: Make it part of the convective scheme (numerics), pressures are available there in the primitive variables.
That consistency might also help with boundary conditions.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1984,28 +2112,31 @@ template<ENUM_TIME_INT IntegrationType>
 FORCEINLINE void CIncEulerSolver::Explicit_Iteration(CGeometry *geometry, CSolver **solver_container,
                                                      CConfig *config, unsigned short iRKStep) {
   SU2_ZONE_SCOPED
-  struct Precond {
-    const CIncEulerSolver* solver;
-    su2activematrix matrix;
-    unsigned short nVar;
+  if (pressure_based) CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME::INCOMPRESSIBLE>::Explicit_Iteration<IntegrationType>(geometry, solver_container, config, iRKStep);
+  else {
+    struct Precond {
```
pcarruscag: Instead of if/else just do a short if for "pressure_based" (please rename to segregated or simple or rhie-chow everywhere) and then return early in that if.
That saves all the changes to else branch.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2292,27 +2425,33 @@ void CIncEulerSolver::BC_Far_Field(CGeometry *geometry, CSolver **solver_contain
     V_infty[prim_idx.Pressure()] = GetPressure_Inf();
 
     /*--- Dirichlet condition for temperature at far-field (if energy is active). ---*/
+    if (!pressure_based) {
 
-    V_infty[prim_idx.Temperature()] = GetTemperature_Inf();
+      V_infty[prim_idx.Temperature()] = GetTemperature_Inf();
```
pcarruscag: Same comment about just one if with `continue` in this case. Makes it a lot easier to review what is going on.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -3408,3 +3626,351 @@ void CIncEulerSolver::ExtractAdjoint_SolutionExtra(su2activevector& adj_sol, con
...
+  su2double GradP_f[MAXNDIM], GradP_in[MAXNDIM], GradP_proj, Coeff_Mom;
+
+  CSolver* poisson_solver = solver_container[POISSON_SOL];
+  CVariable* poisson_nodes = poisson_solver->GetNodes();
+
+  /*--- Mass flux is computed over all edges ---*/
+  for (auto color : EdgeColoring) {
+    SU2_OMP_FOR_DYN(nextMultiple(OMP_MIN_SIZE, color.groupSize))
+    for (auto k = 0ul; k < color.size; ++k) {
+      auto iEdge = color.indices[k];
```
pcarruscag: You also don't need the coloring here since you write to the edges directly.
Coloring is for writing to the points of the edge.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -3408,3 +3626,351 @@ void CIncEulerSolver::ExtractAdjoint_SolutionExtra(su2activevector& adj_sol, con
...
+      Coord_i = geometry->nodes->GetCoord(iPoint);
+      Coord_j = geometry->nodes->GetCoord(jPoint);
+      dist_ij_2 = 0.0;
+      for (iDim = 0; iDim < nDim; iDim++) {
+        Edge_Vector[iDim] = Coord_j[iDim]-Coord_i[iDim];
+        dist_ij_2 += Edge_Vector[iDim]*Edge_Vector[iDim];
+      }
+      /*--- 1. Interpolate the pressure gradient based on node values ---*/
+      for (iDim = 0; iDim < nDim; iDim++) {
+        Grad_Avg = 0.5*(nodes->GetGradient_Primitive(iPoint,0,iDim) + nodes->GetGradient_Primitive(jPoint,0,iDim));
```
pcarruscag: use the primitive indices class of the solver instead of hardcoding the index (0) please

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -3408,3 +3626,351 @@ void CIncEulerSolver::ExtractAdjoint_SolutionExtra(su2activevector& adj_sol, con
...
+      /*--- Eq 15.62 F Moukalled, L Mangani M. Darwish OpenFOAM and uFVM book. ---*/
+      GradP_proj = 0.0;
+      for (iDim = 0; iDim < nDim; iDim++) {
+        GradP_proj += GradP_in[iDim]*Edge_Vector[iDim];
+      }
+      if (dist_ij_2 != 0.0) {
+        for (iDim = 0; iDim < nDim; iDim++) {
+          GradP_f[iDim] = GradP_in[iDim] - (GradP_proj - (nodes->GetPressure(jPoint) - nodes->GetPressure(iPoint)))*Edge_Vector[iDim]/ dist_ij_2;
+        }
+      }
```
pcarruscag: Can you extract a common helper from the numerics classes? this is similar to the corrected gradient calculation.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -3408,3 +3626,351 @@ void CIncEulerSolver::ExtractAdjoint_SolutionExtra(su2activevector& adj_sol, con
...
+
+    /*--- Pressure corrections ---*/
+
+    Current_Pressure = nodes->GetPressure(iPoint);
+    Current_Pressure += alpha_p[iPoint] * (pressureCorrection[iPoint] - PCorr_Ref);
+    nodes->SetSolution(iPoint, 0, Current_Pressure);
+
+    /*--- Velocity corrections ---*/
+    for (iDim = 0; iDim < nDim; ++iDim) {
+      nodes->SetSolution(iPoint, iDim + 1, nodes->GetSolution(iPoint,iDim + 1) + alpha_u * velocityCorrection[iPoint][iDim]);
```
pcarruscag: This isn't how alpha_u is typically used in SIMPLE

Common/include/option_structure.hpp
thijsaalbers: I agree that density based vs pressure based is ambiguous right now. I however dont think that using Rhie-Chow as the trigger for the segregated solver is very clear as Rhie-Chow is just the method we use to avoid odd-even decoupling and is not inherently related to the solver method at all.

I would prefer to keep the current enum, but then to name the options ARTIFICIAL_COMPRESSIBILITY and PRESSURE_CORRECTION, which i think are the most clear. This also allows easily for future addition of a COUPLED method or PRESSURE_PROJECTION etc. 

Why do you say that the default method for incompressible flow is also not artificial compressibility? I thought it was based on the artificial speed of sound beta which is normally called the artificial compressibility method i think? What would it then be called?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
thijsaalbers: Thats correct. Only as we currently use only pseudo-time stepping, there is not really a direct method for the user to influence the velocity corrections. It's not entirely necessary either indeed as the user can alter the CFL, which indirectly influences the time step size, which in turn indirectly functions as a form of alpha_u. So it is removable if you prefer, it doesn't matter that much. 

We could also add an option to change the time step size based on the value of alpha_u and the jacobian matrix, which is mathematically equivalent to using a steady-state SIMPLE algorithm (with alpha_u as the underrelaxation value). This would however completely side-step the CFL parameter. So would this even be desired?

pcarruscag: If CFL works ok and alpha doesn't do much, I would say remove alpha.
Traditional implementations do not relax the velocity and mass flow corrections directly.

Common/include/option_structure.hpp
pcarruscag: Ok let's go with artificial compressibility and pressure correction.
The reference for the incompressible preconditioning is more modern than the artificial compressibility idea by a few decades, which is why I don't like the term very much, but it does introduce a pseudo time derivative to the continuity

pcarruscag: Actually. I think the appropriate term is pressure-velocity coupling, with options: Weiss-Smith (preconditioning) and Rhie-Chow (momentum interpolation).
You already have another enum for the solution method (SIMPLE, SIMPLEC, and maybe COUPLED one day) that can be used with Rhie-Chow.

thijsaalbers: I dont particularly like coupling the Rhie-Chow interpolation to the solution algorithm as they're not inherently related, and in the future we might use different methods depending on the algorithm. I looked up the original work of Economon and he refers to his method himself as a density-based method. I think this is because any non-segregated algorithm is unofficially referred to as density based in much of the literature as it mimics the compressible solver. So I would prefer either

KIND_INCOMP {
  DENSITYBASED,
  PRESSUREBASED,
}

With

KIND_PB {
  SIMPLE,
  SIMPLEC,
  SIMPLER,
  COUPLED,
  PROJECTION,
  Etc.
}

Or

KIND_INCOMP {
  PreconditionedArtificialcompressibility/customartificialcompressibility/economon_method/somethinglikethat,
  PRESSURECORRECTION,
  COUPLED,
  PROJECTION,
  Etc.
}

With

KIND_PRESSURECORRECTION {
 SIMPLE,
 SIMPLEC,
 etc.
}

pcarruscag: ok keep it like it is. Most people don't understand the methods well enough to see it's all the same

SU2_CFD/src/solvers/CIncEulerSolver.cpp
thijsaalbers: I tried it but it results in the residuals stagnating when outlet boundary conditions are used. This is likely a result of the pressure-based solver directly enforcing a dirichlet BC for pressure at the outlet, in contrast to the density-based solver. I could not find a solution for this problem, maybe someone can figure this out in the future.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
thijsaalbers: I could not directly use a common helper as some intermediate variables are used in the Rhie-Chow interpolation, I would thus not know how to unify this with the other existing corrected gradient functions. For now I've added a single helper function specific to the pressure gradient. This could be improved in the future but I think this suffices for now.

Common/include/CConfig.hpp
```diff
@@ -432,6 +433,7 @@ class CConfig {
   su2double *WeightsIntegrationADER_DG;     /*!< \brief The weights of the ADER-DG time integration points on the interval [-1,1]. */
   unsigned short nRKStep;                   /*!< \brief Number of steps of the explicit Runge-Kutta method. */
   su2double *RK_Alpha_Step;                 /*!< \brief Runge-Kutta beta coefficients. */
+  unsigned short nCorrections_PISO;         /*!< \brief Number of corrections used in PISO algorithm. */
```
pcarruscag: Put this one under the SIMPLE options too

Common/include/CConfig.hpp
```diff
@@ -587,6 +591,11 @@ class CConfig {
   Kind_Upwind_Heat,             /*!< \brief Upwind scheme for the heat transfer model. */
   Kind_Upwind_Template;         /*!< \brief Upwind scheme for the template model. */
 
+  ENUM_PBITER
+  Kind_PBIter;             /*< \brief Kind of pressure-based algorithm that is used. */
+  ENUM_INCOMP_SYSTEM
+  Kind_Incomp_System;      /*< \brief Kind of incompressible solver. */
```
```suggestion
  ENUM_PBITER Kind_PBIter;             /*< \brief Kind of pressure-based algorithm that is used. */
  ENUM_INCOMP_SYSTEM Kind_Incomp_System;      /*< \brief Kind of incompressible solver. */
```

Common/include/option_structure.hpp
```diff
@@ -339,6 +340,31 @@ static const MapType<std::string, STRUCT_COMPRESS> MatComp_Map = {
   MakePair("NEARLY_INCOMPRESSIBLE", STRUCT_COMPRESS::NEARLY_INCOMP)
 };
 
+/*!
+ * \brief Type of incompressible solver
+ */
+enum class ENUM_INCOMP_SYSTEM {
```
pcarruscag:
```suggestion
enum class INCOMP_SYSTEM {
```

Common/include/option_structure.hpp
```diff
@@ -339,6 +340,31 @@ static const MapType<std::string, STRUCT_COMPRESS> MatComp_Map = {
...
+};
+static const MapType<std::string, ENUM_INCOMP_SYSTEM> Incomp_Map = {
+ MakePair("DENSITY_BASED", ENUM_INCOMP_SYSTEM::DENSITY_BASED)
+ MakePair("PRESSURE_BASED", ENUM_INCOMP_SYSTEM::PRESSURE_BASED)
+};
+
+/*!
+ * \brief Type of iteration
+ */
+enum class ENUM_PBITER {
```
pcarruscag:
```suggestion
enum class PBITER {
```

Common/include/option_structure.hpp
```diff
@@ -2781,6 +2814,9 @@ enum class MPI_QUANTITIES {
   MESH_DISPLACEMENTS   ,  /*!< \brief Mesh displacements at the interface. */
   SOLUTION_TIME_N      ,  /*!< \brief Solution at time n. */
   SOLUTION_TIME_N1     ,  /*!< \brief Solution at time n-1. */
+  MOM_COEFF            ,  /*!< \brief Momentum coefficient for the Rhie-Chow scheme. */
+  VEL_CORRECTION       ,  /*!< \brief Velocity correction for the pressure-based poisson solver (HbyA). */
+  HBYA_CORRECTION      ,  /*!< \brief HbyA correction for the pressure-based poisson solver. */
```
pcarruscag: Both of them claim to be HbyA?

SU2_CFD/include/iteration/CPBFluidIteration.hpp
```diff
@@ -0,0 +1,75 @@
...
+public:
+  /*!
+   * \brief Constructor of the class.
+   * \param[in] config - Definition of the particular problem.
+   */
+  explicit CPBFluidIteration(const CConfig* config) : CFluidIteration(config) {}
+
+  /*!
+   * \brief Preprocessing to prepare for an iteration of the physics.
+   * \param[in] ??? - Description here.
```
pcarruscag:
```suggestion
```

SU2_CFD/include/solvers/CFVMFlowSolverBase.hpp
```diff
@@ -995,6 +995,17 @@ class CFVMFlowSolverBase : public CSolver {
...
+    /*--- Delete pressure rows for segregated solver type. ---*/
+    if (config->GetKind_Incomp_System() == ENUM_INCOMP_SYSTEM::PRESSURE_BASED) {
+      SU2_OMP_FOR_(schedule(static,omp_chunk_size) SU2_NOWAIT)
+      for (unsigned long iPoint = 0; iPoint < nPoint; iPoint++) {
+        Jacobian.DeleteValsRowi(iPoint, 0);
+        LinSysRes(iPoint,0) = 0.0;
+        LinSysSol(iPoint,0) = 0.0;
+      }
+      END_SU2_OMP_FOR
+    }
```
pcarruscag: Can this be done in the incompressible solver before it calls this generic implementation?
The idea for this class it to contain the common parts and have the others specialize the behavior.

SU2_CFD/include/solvers/CIncEulerSolver.hpp
```diff
@@ -451,4 +457,36 @@ class CIncEulerSolver : public CFVMFlowSolverBase<CIncEulerVariable, ENUM_REGIME
...
+   * \param[in] val_edge_vector - The vector between points i and j
+   * \param[in] val_dist_ij_2 - The distance between points i and j, squared
+   */
+  void CorrectPressureGradient(su2double* corrected_grad_pressure,
+                               const su2double* avg_grad_pressure,
+                               const su2double val_pressure_i,
+                               const su2double val_pressure_j,
+                               const su2double* val_edge_vector,
+                               const su2double val_dist_ij_2);
+
```
pcarruscag: Can it be in the private part of the solver?

SU2_CFD/include/solvers/CPoissonSolver.hpp
```diff
@@ -0,0 +1,264 @@
...
+   * \param[in] solver - Container vector with all of the solvers.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_iter - Current external iteration number.
+   * \param[in] val_update_geo - Flag for updating coords and grid velocity.
+   */
+  void LoadRestart(CGeometry **geometry,
+                   CSolver ***solver,
+                   CConfig *config,
+                   int val_iter,
+                   bool val_update_geo) override {}
```
pcarruscag: I think we can inherit the empty implementations from the parent classes.
```suggestion
```

SU2_CFD/include/solvers/CPoissonSolver.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+  /*!
+   * \brief No upwind residual for poisson equation.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] numerics_container - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] iMesh - Index of the mesh in multigrid computations.
+   */
+  void Upwind_Residual(CGeometry* geometry, CSolver** solver_container, CNumerics** numerics_container,
```
pcarruscag, line 138: same

SU2_CFD/include/solvers/CPoissonSolver.hpp
```diff
@@ -0,0 +1,264 @@
...
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] config - Definition of the particular problem.
+   */
+  void ImplicitEuler_Iteration(CGeometry *geometry, CSolver **solver_container, CConfig *config) override;
+
+  /*!
+   * \brief No time step as it is a stationary problem
+   */
+  void SetTime_Step(CGeometry *geometry,
```
pcarruscag, line 152: same

SU2_CFD/include/variables/CVariable.hpp
```diff
@@ -2416,4 +2416,16 @@ class CVariable {
...
+
+  inline virtual void SetMomCoeff(unsigned long iPoint, su2double val_Mom_Coeff) { }
+
+  inline virtual su2double GetVelocityCorrection(unsigned long iPoint, unsigned short iDim) { return 0.0; }
+
+  inline virtual void SetVelocityCorrection(unsigned long iPoint, unsigned short iDim, su2double val_mom) { }
+
+  inline virtual su2double GetHbyACorrection(unsigned long iPoint, unsigned short iDim) { return 0.0; }
+
+  inline virtual void SetHbyACorrection(unsigned long iPoint, unsigned short iDim, su2double val_HbyAcorrection) { }
```
pcarruscag, line 2430: Do these need to be in the public interface? If they are only used in the solvers that know the exact type of the variable, you don't need them here. Maybe these are used for communications? please check

SU2_CFD/src/drivers/CDriver.cpp
```diff
@@ -1567,7 +1570,7 @@ void CDriver::InitializeNumerics(CConfig *config, CGeometry **geometry, CSolver
   if (fem_ns)       nVar_Flow = solver[MESH_0][FLOW_SOL]->GetnVar();
 
   if (fem)          nVar_FEM = solver[MESH_0][FEA_SOL]->GetnVar();
-
+
```
pcarruscag, line 1573:
```suggestion

```

SU2_CFD/src/iteration/CPBFluidIteration.cpp
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+#include "../../include/iteration/CPBFluidIteration.hpp"
+#include "../../include/output/COutput.hpp"
+#include "../../include/integration/CIntegration.hpp"
+
+void CPBFluidIteration::Preprocess(COutput* output, CIntegration**** integration, CGeometry**** geometry,
+                                 CSolver***** solver, CNumerics****** numerics, CConfig** config,
+                                 CSurfaceMovement** surface_movement, CVolumetricMovement*** grid_movement,
+                                 CFreeFormDefBox*** FFDBox, unsigned short val_iZone, unsigned short val_iInst) { }
```
pcarruscag: No empty functions if possible.
```suggestion
```

SU2_CFD/src/numerics/flow/convection/pressure_based.cpp
```diff
@@ -0,0 +1,164 @@
...
+  AdvectedVelocity = new su2double [MAXNDIM];
+  Flux = new su2double [nVar];
+  Jacobian_i = new su2double* [nVar];
+  Jacobian_j = new su2double* [nVar];
+  
+  for (iVar = 0; iVar < nVar; iVar++) {
+    Jacobian_i[iVar] = new su2double [nVar];
+    Jacobian_j[iVar] = new su2double [nVar];
+  }
+}
```
pcarruscag, line 46: Why do you need a new base class? Can't you build this out of the ScalarNumerics? There is at least upwinding there already.

SU2_CFD/src/output/CFlowIncOutput.cpp
```diff
@@ -91,8 +93,12 @@ CFlowIncOutput::CFlowIncOutput(CConfig *config, unsigned short nDim) : CFlowOutp
   restartFilename = config->GetRestart_FileName();
 
   /*--- Set the default convergence field --- */
+  /*--- The default field for the pressure-based solver is velocity, as the pressure is only 
+  solved as a correction and it is thus possible to run the pb solver without corrections. to 
+  ensure the default residual is always defined we use velocity instead ---*/
 
-  if (convFields.empty()) convFields.emplace_back("RMS_PRESSURE");
+  if (convFields.empty() && !pressure_based) convFields.emplace_back("RMS_PRESSURE");
+  if (convFields.empty() && pressure_based) convFields.emplace_back("RMS_VELOCITY-X");
```
pcarruscag, line 101: Isn't there a notion of a continuity residual in the new solver?

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1015,6 +1038,11 @@ void CIncEulerSolver::CommonPreprocessing(CGeometry *geometry, CSolver **solver_
     SU2_OMP_SAFE_GLOBAL_ACCESS(GetOutlet_Properties(geometry, config, iMesh, Output);)
   }
 
+  /*--- Reset flag for strong BCs. ---*/
+  if (pressure_based)
+    for (unsigned long iPoint = 0; iPoint < nPointDomain; iPoint++)
```
pcarruscag, line 1044: omp_for

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1169,9 +1185,42 @@ void CIncEulerSolver::SetTime_Step(CGeometry *geometry, CSolver **solver_contain
...
+    to be zero such that a regular advective time step is computed */
+    struct SoundSpeed {
+      FORCEINLINE su2double operator() (const CIncEulerVariable& nodes, unsigned long iPoint, unsigned long jPoint) const {
+        return 0.0;
+      }
+
+      FORCEINLINE su2double operator() (const CIncEulerVariable& nodes, unsigned long iPoint) const {
+        return 0.0;
+      }
+
```
pcarruscag:
```suggestion
      FORCEINLINE su2double operator() (const CIncEulerVariable& nodes, unsigned long iPoint, unsigned long jPoint = 0) const {
        return 0.0;
      }
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1477,6 +1534,34 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+  if (pressure_based) {
+
+    /*--- Add pressure source term ---*/
+
+    SU2_OMP_FOR_STAT(omp_chunk_size)
+    for (auto iPoint = 0ul; iPoint < nPointDomain; iPoint++) {
+
+      su2double Density = nodes->GetDensity(iPoint);
+
+      su2double *pressureGradientSource = new su2double[nVar];
```
pcarruscag: this is leaking memory, use a C array of size MAXNVAR or similar

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1477,6 +1534,34 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+      pressureGradientSource[0] = 0.0;
+      pressureGradientSource[nDim+1]=0.0;
+
+      /*--- Compute the residual (V / rho * gradp) based on the pressure gradient. ---*/
+
+      for (unsigned short iDim = 0; iDim < nDim; iDim++)
+        pressureGradientSource[iDim+1] = geometry->nodes->GetVolume(iPoint) / Density * nodes->GetGradient_Primitive(iPoint,prim_idx.Pressure(),iDim);
+
+      auto residual = CNumerics::ResidualType<>(pressureGradientSource, nullptr, nullptr);
```
pcarruscag: You can pass the array directly.
Or use LinSysRes(iPoint, iDim + 1) += ... directly.
```suggestion
```

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -1477,6 +1534,34 @@ void CIncEulerSolver::Source_Residual(CGeometry *geometry, CSolver **solver_cont
...
+      su2double Density = nodes->GetDensity(iPoint);
+
+      su2double *pressureGradientSource = new su2double[nVar];
+      pressureGradientSource[0] = 0.0;
+      pressureGradientSource[nDim+1]=0.0;
+
+      /*--- Compute the residual (V / rho * gradp) based on the pressure gradient. ---*/
+
+      for (unsigned short iDim = 0; iDim < nDim; iDim++)
+        pressureGradientSource[iDim+1] = geometry->nodes->GetVolume(iPoint) / Density * nodes->GetGradient_Primitive(iPoint,prim_idx.Pressure(),iDim);
```
pcarruscag: Why is the momentum equation divided by density?

TestCases/incomp_euler/naca0012/incomp_pb_NACA0012.cfg
pcarruscag, line 1: I don't see changes to the regression scripts, can you make sure the new configs have regression tests.
And at least a couple should be V&V quality.

SU2_CFD/include/solvers/CPoissonSolver.hpp
thijsaalbers: The parent class here (for unknown reason to me) forces the solver to have this implementation (with =0), can I change this then change this to {}?

SU2_CFD/include/solvers/CPoissonSolver.hpp
thijsaalbers: The upwind as well as the set time step functions are not empty in the parent class

SU2_CFD/include/variables/CVariable.hpp
thijsaalbers: Yes these are used for communication

SU2_CFD/src/numerics/flow/convection/pressure_based.cpp
thijsaalbers: The scalar upwind class only really implements the upwinding (and adjoint stuff). This would not really be usefull  (and even akward) for the centered scheme. I think it would be cleaner as it is now where the centered and upwind scheme share a base class specific to the pressure based flow equations.

SU2_CFD/src/output/CFlowIncOutput.cpp
thijsaalbers: Yes there is. Only as it is right now, it is possible to run the pressure-based solver without the pressure-corrections which then just solves the momentum equations without constraints. This is (at least in my experience) useful for debugging. Without pressure correctiosn the continuity residual is thus empty.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
thijsaalbers: The density based solver uses the solution variables (u,v,w) and the preconditioning matrix basically makes it an equation for (rho u, rho v, rhow). Because we now use the same solution variables for the pressure-based solver, which does not use this preconditioning matrix, the equations are/should be divided by the density. Other implementations sometimes neglect this division by density in the pressure which then represents kinematic pressure but I think that creates confusion with the other solvers in SU2.

SU2_CFD/include/solvers/CPoissonSolver.hpp
pcarruscag: ok then in the comment just put the reason for overriding the base with an empty implementation and no need to document the arguments again

SU2_CFD/src/solvers/CIncEulerSolver.cpp
pcarruscag: Hmm the "division" appears through the Jacobian matrix, dividing everything by density is only correct for constant density.
The residual should reflect the conservative form, the Jacobian can take a derivative with respect to rho u or just u, just that the solution of the linear system is an update for u.

PR #2818, opened 2026-05-21
Thermochemically consistent backflow treatment for FGM

joshkellyjak:
## Proposed Changes
Added thermochemically consistent backflow treatment for FGM calculations to stabilise early convergence in the presence of reversed-velocity at the outlet. In the flow solver, reversed-velocity ghost states at outlet faces are corrected by reflecting the normal velocity component. In the flamelet scalar solver, a new BC_Outlet override blends the enthalpy ghost state between the interior value and a thermodynamically consistent inlet enthalpy, computed via Newton iteration on the FGM LUT, while clamping the progress variable to it's unburnt value and preventing burned scalars from re-entering the domain and pushing FGM lookups outside manifold bounds. The backflow prevention deactivates automatically after a user-specified number of outer iterations, allowing physically real backflow to develop once the solution has stabilised.

## Related Work
N/A

## PR Checklist


Common/include/parallelization/mpi_structure.cpp
```diff
@@ -101,12 +101,12 @@ void CBaseMPIWrapper::Error(const std::string& ErrorMsg, const std::string& Func
...
-    std::cout << "-------------------------------------------------------------------------" << std::endl;
-    std::cout << ErrorMsg << std::endl;
-    std::cout << "------------------------------ Error Exit -------------------------------" << std::endl;
-    std::cout << std::endl << std::endl;
+    std::cerr << std::endl << std::endl;
+    std::cerr << "Error in \"" << FunctionName << "\": " << std::endl;
+    std::cerr << "-------------------------------------------------------------------------" << std::endl;
+    std::cerr << ErrorMsg << std::endl;
+    std::cerr << "------------------------------ Error Exit -------------------------------" << std::endl;
+    std::cerr << std::endl << std::endl;
```
pcarruscag, line 109: revert please

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2830,6 +2863,16 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
 
+  /*--- Print a warning if backflow was detected on this outlet marker. ---*/
+  BEGIN_SU2_OMP_SAFE_GLOBAL_ACCESS {
+    if (nBackflow_loc > 0) {
+      cout << "WARNING [Rank " << rank << "]: Backflow detected at outlet marker \""
+           << Marker_Tag << "\": " << static_cast<unsigned long>(nBackflow_loc)
+           << " face(s) have reversed normal velocity." << endl;
+    }
+  }
+  END_SU2_OMP_SAFE_GLOBAL_ACCESS
```
pcarruscag, line 2874: I don't like this type of debug output we have in wall functions too and it's a pain when it goes off on multiple ranks, you come back to your terminal and can even scroll back to find where the problem started.
The right way to do this is to store the count as a solver member, then the output class can do an MPI reduction and report it as an output, which you can choose to have on screen or not. Then if you want to print a concise warning when files are written out, like we do for non physical points, that's also not a bad idea.

SU2_CFD/src/solvers/CIncEulerSolver.cpp
```diff
@@ -2674,6 +2677,16 @@ void CIncEulerSolver::BC_Outlet(CGeometry *geometry, CSolver **solver_container,
...
     P_domain = nodes->GetPressure(iPoint);
 
+    /*--- Compute the face area and normal velocity (positive = outflow, negative = backflow). ---*/
+
+    const su2double Area = GeometryToolbox::Norm(nDim, Normal);
+    const su2double Vn   = GeometryToolbox::DotProduct(nDim, &V_domain[prim_idx.Velocity()], Normal) / Area;
+
+    if (Vn < 0.0) {
+      SU2_OMP_ATOMIC
+      nBackflow_loc += 1;
```
pcarruscag, line 2687: This variable is thread-local because all threads already enter this function and thus get their own copy of all local variables.
You need to make it static, or if you make it a class member like I mention in the other commend it also works (because the solver is not created by all threads so solver members are shared).

SU2_CFD/src/solvers/CSpeciesFlameletSolver.cpp
```diff
@@ -496,6 +496,142 @@ void CSpeciesFlameletSolver::BC_Inlet(CGeometry* geometry, CSolver** solver_cont
...
+    }
+  }
+  if (!found_inlet) {
+    /*--- No inlet boundary defined — nothing to blend against; fall back to Neumann. ---*/
+    CSpeciesSolver::BC_Outlet(geometry, solver_container, conv_numerics, visc_numerics, config, val_marker);
+    return;
+  }
+
+  const su2double T_inlet = config->GetInletTtotal(Inlet_Tag);
+  const su2double V_ref   = config->GetInletPtotal(Inlet_Tag) / config->GetVelocity_Ref();
```
pcarruscag, line 538: This needs hardening. What if you have multiple inlets? What about the type of inlet?
The robust thing to do is to either have the backflow values specified or derive them some other way (which is possible, but I can't say what I've done in other places).

Common/include/parallelization/mpi_structure.cpp
joshkellyjak: Oops this must've slipped through

Issue #1468, opened 2021-12-08
Issues with compiling SU2 on Windows 10


GabrieleMorvillo:
Good evening, I am currently struggling with compiling SU2 from source code. After I have built the program I try to compile it using ninja.exe, but during the process I get the following error:

```
[443/454] Linking target SU2_DEF/src/SU2_DEF.exe
FAILED: SU2_DEF/src/SU2_DEF.exe
c++ @SU2_DEF/src/SU2_DEF.exe.rsp
C:/Program Files/mingw-w64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lC:\Program Files (x86)\Microsoft SDKs\MPI\Lib\x64\msmpi
C:/Program Files/mingw-w64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lC:\Program Files (x86)\Microsoft SDKs\MPI\Lib\x64\msmpi
C:/Program Files/mingw-w64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lC:\Program Files (x86)\Microsoft SDKs\MPI\Lib\x64\msmpi
C:/Program Files/mingw-w64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lC:\Program Files (x86)\Microsoft SDKs\MPI\Lib\x64\msmpi
collect2.exe: error: ld returned 1 exit status
[448/454] Compiling C++ object SU2_CFD/src/libSU2core.a.p/.._include_numerics_simd_CNumericsSIMD.cpp.obj
ninja: build stopped: subcommand failed.
```

I do not understand what causes the problem since MPI should have been correctly installed and the msmpi.lib file is present in C:\Program Files (x86)\Microsoft SDKs\MPI\Lib\x64.
I imagine this is a trivial problem due to my inexperience but I am not able to understand how to fix it.

Any help would be greatly appreciated.

pcarruscag:
Hi, can you try the process mentioned in #1471

GabrieleMorvillo:
> Hi, can you try the process mentioned in #1471

Good morning and thanks for the reply.
[...]

PR #2821, opened 2026-05-25
Fix/windows portable version script

DEFAULTE-R:
## Proposed Changes

Replaces the Unix-specific shell `grep` and `rm` calls in `SU2_PY/change_version_number.py` with a pure Python cross-platform implementation using `pathlib`.

The new implementation:

* removes dependency on shell utilities,
* recursively scans files using `Path.rglob()`,
* skips `.git`, `.svn`, and `build` directories,
* preserves existing script behavior,
* improves Windows compatibility.

## Related Work

Related to #1468.

This addresses the TODO comment in `change_version_number.py` noting that the original implementation only worked on Unix systems due to shell command usage.

## PR Checklist

* [x] I am submitting my contribution to the develop branch.
* [x] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
* [ ] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I have added a test case that demonstrates my contribution, if necessary.
* [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.


pcarruscag:
Thank but we don't need this change because thus script is only used in CI and that is not going to be windows any time soon.

PR #2822, opened 2026-05-25
Harden host-side sparse index width and allocation safety

LwhJesse:
## Proposed Changes

Introduce `su2_index_t` as the host-side sparse index type:

```cpp
using su2_index_t = std::uint64_t;
```

Use it through the host sparse-pattern path, including graph-toolbox aliases, geometry sparse-pattern accessors, `CSysMatrix` LDU metadata, `CSysVector` dimensions, and PaStiX sparse input handling.

Add checked arithmetic for sparse storage sizes and checked casts at external boundaries. CUDA device sparse indices remain `su2uint`; host indices are checked before upload.

This prevents silent overflow or truncation in large host-side sparse matrices. On LLP64 platforms, `unsigned long` is 32-bit, so products such as `nnz_blocks * nVar * nEqn` can wrap before allocation. For example, a `5 x 5` double block matrix crosses the 32-bit unsigned limit at:

```text
171,798,692 * 25 = 4,294,967,300 entries
4,294,967,300 * 8 bytes = 34,359,738,400 bytes
```

The affected paths now fail explicitly instead of allocating with wrapped sizes.

## Related Work

Keeps the LDU/custom-kernel direction discussed in #2838. This does not replace the current CUDA sparse backend with cuSPARSE.

Validation:

- `git diff --check`
- `uvx --python 3.10 pre-commit run -a -v`
- `ninja -C build`
- `ninja -C build-verify-cuda`
- `ninja -C build-verify-pastix`
- `ninja -C build-harness-omp-cpu`
- `periodic2d` OpenMP regression, `SU2_CFD -t 2`, passed at iteration 1400 within `1e-4`
- `fea_topology/quick_start/settings.cfg`, `SU2_CFD`, completed with `Exit Success`

## PR Checklist

### DIFF
--- a/Common/include/code_config.hpp
+++ b/Common/include/code_config.hpp
@@ -111,7 +111,10 @@ FORCEINLINE Out su2staticcast_p(In ptr) {
 #undef USE_SINGLE_PRECISION
 #endif
 
-/*--- Default integer types. Currently used for rank-local sparse patterns. ---*/
+/*--- Default integer types. Host sparse metadata is explicit 64-bit so rank-local
+ * sparse patterns and allocation-size arithmetic do not depend on the platform
+ * width of unsigned long. Device sparse metadata keeps the compact su2uint type. ---*/
+using su2_index_t = uint64_t;
 using su2uint = uint32_t;
 using su2int = int32_t;
 
--- a/Common/include/geometry/CGeometry.hpp
+++ b/Common/include/geometry/CGeometry.hpp
@@ -205,10 +205,10 @@ class CGeometry {
   LDUSparsePattern finiteElementPatternFill0; /*!< \brief FEM sparsity with 0-fill (structural pattern). */
   LDUSparsePattern finiteElementPatternFillN; /*!< \brief FEM sparsity with N-fill (e.g. for ILU-N). */
 
-  su2vector<su2uint> finiteVolumeLToUTranspMap;  /*!< \brief FVM L-entry -> U-entry of its transpose. */
-  su2vector<su2uint> finiteVolumeUToLTranspMap;  /*!< \brief FVM U-entry -> L-entry of its transpose. */
-  su2vector<su2uint> finiteElementLToUTranspMap; /*!< \brief FEM L-entry -> U-entry of its transpose. */
-  su2vector<su2uint> finiteElementUToLTranspMap; /*!< \brief FEM U-entry -> L-entry of its transpose. */
+  su2vector<su2_index_t> finiteVolumeLToUTranspMap;  /*!< \brief FVM L-entry -> U-entry of its transpose. */
+  su2vector<su2_index_t> finiteVolumeUToLTranspMap;  /*!< \brief FVM U-entry -> L-entry of its transpose. */
+  su2vector<su2_index_t> finiteElementLToUTranspMap; /*!< \brief FEM L-entry -> U-entry of its transpose. */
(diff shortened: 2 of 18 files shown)


LwhJesse:
After [#2816][pr-2816], I think the follow-up direction needs a policy decision before I move [#2822][pr-2822] or the resource-cache draft further.

I currently have two related follow-ups:

* [#2822][pr-2822] is the host-side sparse-index / allocation-safety draft.
* The [#2816-based resource-cache draft][resource-cache-diff] is the cuSPARSE resource-lifetime follow-up.

The next steps depend on which CUDA sparse-backend direction the project prefers. I see four practical routes:

1. Modern cuSPARSE-only route: keep the generic `cusparseSpMV()` BSR path from [#2816][pr-2816], require CUDA/cuSPARSE 13.1+ for that backend, add a clear configure/build guard for unsupported CUDA/cuSPARSE 13.0 or older versions, and move both host and device sparse indices to fixed 64-bit.

2. Transitional host-side hardening route: keep [#2822][pr-2822] limited to host-side sparse-index hardening. Host indices become fixed 64-bit, while the current CUDA boundary keeps its existing index type with checked conversion.

3. Split-backend compatibility route: keep the modern generic cuSPARSE BSR path for CUDA/cuSPARSE 13.1+, but add a CUDA/cuSPARSE 13.0-or-older fallback. That fallback could be either the legacy int32 BSR API, or a corrected custom CUDA kernel.

4. Full custom-kernel route: move away from the generic cuSPARSE BSR path and maintain a corrected SU2 custom CUDA matvec kernel instead. This would avoid the CUDA/cuSPARSE 13.1+ requirement, but it would also change the direction of the [#2816-based resource-cache draft][resource-cache-diff].

My technical preference is route 1, because it gives the cleanest correctness story and the cleanest future GPU-resident Krylov path. If that is too large as the immediate next step, route 2 is a reasonable limited scope for [#2822][pr-2822].

Could the maintainers confirm which direction they prefer before I mark [#2822][pr-2822] ready or move the [#2816-based resource-cache draft][resource-cache-diff] into the main project?

[pr-2816]: https://github.com/su2code/SU2/pull/2816
[pr-2822]: https://github.com/su2code/SU2/pull/2822
[resource-cache-diff]: https://github.com/LwhJesse/SU2/pull/1/files

pcarruscag:
I think both are going too deep into the software engineering weeds.
I would see an end-to-end linear solve on GPUs, product, preconditioner, vector-vector operations, before we start worrying about these portability and caching aspects.

pcarruscag (commented):
I think that the su2uint I introduced addresses this, and we have custom kernels for almost everything now, since we switched to the LDU format.
So this work is probably not needed anymore? What do you think?

LwhJesse:
> I think that the su2uint I introduced addresses this, and we have custom kernels for almost everything now, since we switched to the LDU format. So this work is probably not needed anymore? What do you think?

I checked the current `develop` again. The specific overflow risk this PR addresses still exists independently of the CUDA sparse backend.

`su2uint` fixes the sparse pattern index type at 32 bits, but the host-side storage size is still computed from expressions such as `nnz_l * nVar * nEqn` and passed through `unsigned long`. On Windows LLP64, `unsigned long` is 32-bit, so the block pattern itself can fit in 32 bits while the scalar-entry count overflows.

For example, `171,798,692 * 5 * 5 = 4,294,967,300`, which is just above `UINT32_MAX`. That can wrap before allocation even though `nnz_l` itself is valid.

So the switch to LDU/custom CUDA kernels removes the old cuSPARSE-related concern, but it does not remove this host-side allocation/index-width issue.

PR #2825, opened 2026-06-01
Add initial end-to-end CUDA FGMRES solver path

LwhJesse:
## Proposed Changes

This PR adds an end-to-end CUDA linear solve path: the Krylov solvers keep their host control flow, and `CSysVector` operations, the SpMV and the Jacobi preconditioner are dispatched to CUDA kernels when `ENABLE_CUDA=YES`.

Transfers are explicit and owned by the object responsible for the data, with no coherency or dirty-flag tracking in `CSysVector` / `CSysMatrix`:
- `CSysMatrixVectorProduct` uploads the matrix on construction
- the preconditioner uploads its data in `Build()`
- `CSysSolve` uploads `b` and `x` and downloads `x` in `HandleTemporariesIn/Out`, which is also where device evaluation is switched on and off
- preconditioners without a device implementation (ILU, LU-SGS, Linelet, PaStiX) download the input, apply on the host, and upload the result

A solve is therefore fully device resident for Identity and Jacobi preconditioners: **2 uploads and 1 download per linear system**, independent of the Krylov subspace size.

Implementation notes:
- `cuBLAS` for `dot` / `norm`, custom kernels for the block-LDU SpMV, the Jacobi apply, and `CSysVector` expression assignment
- `CSysVector` operands are captured in expressions by value, so an arbitrary expression tree is trivially copyable into the assignment kernel; the required expression shapes are explicitly instantiated in `CSysVectorGPU.cu`, consistent with how `CSysMatrix` is instantiated
- device work is issued by a single thread with the OpenMP team synchronized around it, so the GPU path is usable from inside the existing parallel regions; this is internal to the linear algebra layer
- the CUDA translation units are only linked into the primal libraries, since they cannot be compiled with the CoDiPack defines; device dispatch is compiled out of the AD builds
- adds `LINEAR_SOLVER_PREC= NONE` (identity)

## Related Work

Follows the review direction in #2822 (show a working end-to-end GPU linear solve before splitting out infrastructure) and the implementation preferences in #2816.

## Validation

- CPU vs GPU on inviscid NACA0012, 25 iterations, RTX 4070 Ti SUPER (sm_89), for FGMRES + `JACOBI`, FGMRES + `NONE`, FGMRES + `ILU` (host preconditioner path) and BCGSTAB. Results agree to the printed precision in double, and to ~6 significant figures in mixed precision. BCGSTAB agrees exactly once the linear system is converged (`LINEAR_SOLVER_ERROR=1e-10`); at loose tolerances the two paths diverge through BCGSTAB's own sensitivity, not a difference in the algebra.
- Mixed, normal and single precision builds all compile and run; device dispatch confirmed live in each (kernel launch counts track the subspace size).
- `OMP_NUM_THREADS=1` and `4` give bit-identical results on the GPU path.
- Builds verified: primal, primal + AD + directdiff, `enable-cuda` + `with-omp`, mixed / normal / single precision.
- Transfer counts measured per solve: 2 H2D + 1 D2H + 1 matrix upload for Jacobi and `NONE`, plus one download/upload pair per preconditioner application for ILU.

Earlier validation of the original design (6 representative cases, `nsys` / `ncu` profiling) predates the rework of the transfer and dispatch model and should be repeated.

## PR Checklist


Common/src/linear_algebra/CSysSolveGPU.cu
```diff
@@ -0,0 +1,242 @@
...
+};
+
+}  // namespace
+
+template <class ScalarType>
+unsigned long CSysSolve<ScalarType>::FGMRES_LinSolver_GPU(const VectorType& b, VectorType& x,
+                                                          const ProductType& mat_vec, const PrecondType& precond,
+                                                          ScalarType tol, unsigned long m, ScalarType& residual,
+                                                          bool monitoring, const CConfig* config) const {
+  SU2_ZONE_SCOPED
```
pcarruscag: We don't need to duplicate anything, we just need the CSysVector expression templates to dispatch to GPU.

LwhJesse: Thanks, I reworked this so the FGMRES control flow is shared through `FGMRES_LinSolverImpl`.

`CSysSolveGPU.cu` now only provides the CUDA backend ops and a thin wrapper, and the CUDA vector updates go through `CSysVector` device expression-template views.

Common/include/linear_algebra/CSysVector.hpp
```diff
@@ -59,6 +63,116 @@
 #define END_CSYSVEC_PARFOR
 #endif
 
+namespace VecExpr {
+
+enum class DeviceAssignOp { Assign, Add, Subtract, Multiply, Divide };
+
+template <class Scalar>
+class CDeviceVectorView : public CVecExpr<CDeviceVectorView<Scalar>, Scalar> {
```
pcarruscag: This is a step in the right direction, but what I have in mind is to use CSysVector directly, so that CSysSolve can stay nearly identical and completely agnostic to cpu or gpu.
I think we can do this by specializing the `store_t` trait for CSysVector, so that the type stored by expressions becomes this view, note that we only need to capture the pointer, the size is defined by the vector on the left-hand side of the assignment or compound assignment.
This way CSysVector either does the CPU loop or launches the GPU kernel in the assignment, according to how the GPU boolean is set.

pcarruscag: For clarity, CSysVector stops being stored as a reference and is instead stored "as a view" (which is a value type).

LwhJesse: Reworked along those lines: CSysVector expressions now use a store_t value-view, and FGMRES shares FGMRES_LinSolverImpl. Does this match what you had in mind?

pcarruscag: Nearly there, nice work.
Is it possible to put the FGMRES implementation back in CSysSolve now? What is special about the implementation in the new header vs the old one? I'm hoping we can keep the linear solver implementations completely agnostic to CPU/GPU details.
The product, preconditioner, and vector abstractions should be enough to insulate CSysSolve from CPU/GPU details.

LwhJesse: Thanks, reworked again.

FGMRES is back in CSysSolve.cpp, and the separate CSysSolveFGMRES.inl / CSysSolveGPU.cu path is gone. CPU/GPU details are now handled by the product, preconditioner, and CSysVector abstractions, so CSysSolve should stay agnostic.

pcarruscag: Sorry, I missed your reply.

pcarruscag (approved):

PR #2830, opened 2026-06-15
Detect non-physical solutions in a statistic way instead of just P or T <= 0

pcarruscag:
## Proposed Changes
I went back to the HLCRM (everything but the Nacelle's version), and there are areas (sharp corners) behind the pylons where temperature becomes either very low or very high.
Eventually, these unstable points (about 100 out of ~4M) lead to divergence despite the automatic relaxation and CFL adaptation. Solution limiters are not enough, probably because within these regions the solution appears smooth relative to the large changes between points.
So the idea here is to find these points from the mean temperature and standard deviation and treat them as first order; we already had something like this for edges where the reconstructed states become non-physical.

Off by default. I don't have a small enough case that triggers this.
It may not work very well for flows with huge temperature ranges.

## PR Checklist


YairMO:
Hello,

Could you please share which turbulence model it is? 
If it is the SA model, I suggest checking the eddy viscosity values in these regions. It may happen (although not popular) that the eddy viscosity (of the SA) reaches unrealistic values and, in turn, affects the pressure/temparture such that the solution blows up. It is likely to occur in highly anisotropic grid elements and is CFL-dependent. However, once eddy viscosity grows unbounded, automatic relaxation and adaptive CFL will only delay solution divergence.

pcarruscag:
Hi Yair, it is the SA model, but in these regions where the temperature becomes very low, nu tilde is close to 0.
It is inside very sharp corners where the slat brackets meet the inside surface of the slat.
image
image
I don't think it is the viscosity affecting the flow but simply the geometry and mesh affecting both flow and turbulence.

bigfooted:
It's an outlier when a value is more than twice the standard deviation away from the mean?

pcarruscag:
The default is 5 times the standard deviation, and can be adjusted via the parameters.

PR #2838, opened 2026-06-29
Store matrices in LDU format to improve performance of LU-SGS and ILU preconditioners

### LINKED ISSUE #2837: LU-SGS cost can be twice what one would expect
For a 3D mesh with many tets and prisms (thus a dual mesh with many neighbors to each point) LU-SGS costs about 2 matrix-vector products.

Image

LU-SGS, as the name implies, does one pass over the lower entries of the matrix and then one over the upper entries.
So one would naively expect it to cost about the same as one product.
I think the issue is that since there are many neighbors, the CPU falsely detects (via the hardware prefetcher) that the entire matrix is being used and reads the whole thing from memory for each pass.

The ideal solution would probably be to switch to the LDU format (used by OpenFOAM), but I might try to confuse the prefetcher so it doesn't read the entire matrix to see if that can get close to the "speed of light" for this (1 product).

pcarruscag:
## Proposed Changes
Store matrices in LDU format
On a representative 3D case (one of the HLCRM meshes):
Time per iteration with ILU: 2.65s -> 2.4s
Time per iteration with LU_SGS: 2.05s -> 1.85s

I was not very careful in making sure these profiles were 100% apples to apples by covering the same number of iterations but they show the preconditioner time is not much closer to the matrix-vector product.
ILU develop:
ilu_dev
ILU using LDU format
ilu_new
SGS develop:
lu_sgs_dev
SGS using LDU format:
lu_sgs_new

## Related Work
Closes #2837 

## PR Checklist


TestCases/euler/oneram6/inv_ONERAM6.cfg
```diff
@@ -113,7 +113,7 @@ GRAD_OBJFUNC_FILENAME= of_grad
 SURFACE_FILENAME= surface_flow
 SURFACE_ADJ_FILENAME= surface_adjoint
 OUTPUT_WRT_FREQ= 100
-SCREEN_OUTPUT = (INNER_ITER, RMS_DENSITY, RMS_ENERGY, LIFT, DRAG)
+SCREEN_OUTPUT = (INNER_ITER, RMS_DENSITY, RMS_ENERGY, LIFT, DRAG, LINSOL)
```
pcarruscag:
```suggestion
SCREEN_OUTPUT = (INNER_ITER, RMS_DENSITY, RMS_ENERGY, LIFT, DRAG)
```

PR #2843, opened 2026-07-10
[GSoC] Feature: MultiDot  with cublas<t>gemm (Originally: Jacobi preconditioner on GPU and CUDA Unified preconditiong matrix)

ddg93:
# Google Summer of Code: final evaluation
## Goals of the project
This PR includes large part of my GSoC 2026 activities. The goal of the project was to extend the section of the FGMRES solver running on GPU by porting it to CUDA, and to benchmark the CUDA Unified Memory approach for the Krylov vectors processed by the FGMRES solver.

## What was done
At the beginning of the project, the FGMRES solver had only a matrix-vector multiplication-add operation ported to GPU. Note that a concurrent Pull-Request #2825 had already started to work on the end-to-end porting.
The goals were tackled in the following way:
 - first, the Unified Memory allocation was benchmarked on the preconditioner matrix. This is an array that migrates to device only once per iteration, differently from the Krylov vectors which were moving between Host and Device several times. This required also the porting of a small kernel to apply such preconditioner.
 - secondly, the Unified Memory allocation was extended to the Krylov vectors. This allowed to incrementally extend the section of the FGMRES solver ported to GPU. Custom kernels were developed for specific operations such as the multiDot and Dot products and the linear combination, while operator overloading was used to port the rest of the vector operations.
 - Finally, the Unified Memory approach was benchmarked against the Managed Memory approach.

## Current state
Despite PR #2825 porting the whole FGMRES solver was merged before the end of this GSoC PR, the custom multiDot approach developed in this PR during the GSoC was still kept and merged into SU2 given its good performance. Instead, the Unified Memory approach was deprecated following the results of the benchmarks, which highlighted no performance gain against the Managed Memory, and the design choices of keeping only one memory management approach.

Following a discussion with the maintainers, the rest of the GSoC effort has been dedicated to the porting of another preconditioner to GPU: the LU-SGS preconditioner in PR #2872. This work is ongoing but already in an advanced state, and might potentially be merged in the next days.

## Challenges and important things learned during the project
This GSoC project was an important opportunity to learn how to work on an advanced Open Source project such as SU2, how to interact with the OS community and how to effectively adapt the scope of my efforts to produce an useful contribution.

The original history of the PR is kept below:

## Proposed Changes
This PR introduces improves the GPU implementation of the multiDot product between Krylov vectors used in the FGMRES solver. It moves from a looped instantiation of pairwise dot products to a batched one-time kernel launch. A custom CUDA kernel and the batched gemm using cublas<t>gemm are benchmarked, the latter being eventually selected.

## Original proposal for the GSoC
This PR introduces CUDA Unified memory and Managed memory allocation, and memory management for the CSysVector class and the preconditioning matrix inside the Jacobi preconditioner. This allows for a benchmark between the two memory strategies within the FGMRES solver.

Also, this draft PR extends the section of GPU execution inside the FGMRES solver
Custom CUDA kernels are implemented for the preconditioning matrix, the multi dot product and the linear combination (inside the Modified Gram-Schmidt orthogonalization), and the vector norm calculation. Unary vector-scalar operations based on templates are also offloaded to GPU through a generic kernel. Moreover, abstract Syntax Tree are deployed to offload vector-vector binary operations to the GPU through a generic kernel based on a runtime evaluation of the tree.

The solver logic is not modified, and the GPU path is hidden inside the specific methods. 

### Unified memory approach https://github.com/su2code/SU2/pull/2843/commits/5fe25b814ff1c9b256a00a568f4c5e41be5c96ee

A custom data() method recovers the CSysVector Unified pointer inside the CUDA logic, cleaning the logic from the double host/device pointers. All memory explicit memory copies are also removed from the CUDA logic, but explicit device synchronizations are introduced around MPI calls and at the end of the CUDA section. 

### Managed memory approach https://github.com/su2code/SU2/pull/2843/commits/4b43fa0d365475d5e500ff50a3fa1c73c0d13976

Operators accessing the CSysVector on Host after GPU operations demand synchronization, which is introduced explicitly in each.

## Performance evaluation (on-going)
For the considered test-case (rae2822), the CPU execution expresses an **Avg. s/iter:    0.198681**. 
The GPU execution with Unified memory expresses an **Avg. s/iter:    0.238953**.        
The GPU execution with Managed memory expresses an **Avg. s/iter:    0.372046**.
The CPU is a Intel(R) Xeon(R) E-2276M  CPU @ 2.80GHz with 12 cores. The GPU is a Quadro P620. Tests on more advanced hardware are ongoing.
Profiling results comparing the Unified Memory (left column) against the Managed Memory (right column) are available in the attached pdf: 
[SU2_ra2822_GPU_MA_vs_UM_nsys_prof.pdf](https://github.com/user-attachments/files/30520807/SU2_ra2822_GPU_MA_vs_UM_nsys_prof.pdf)

## Asynchronous pre-fetching:
The Jacobi preconditioner calculations are performed on GPU through a new custom CUDA kernel under the preconditioner abstraction. 
The preconditioning matrix is selected to test CUDA Unified Memory asynchronous prefetching to the GPU.
For simplicity, the double CPU/GPU pointer is still maintained in the current logic, although the device pointer reduces to an alias for the Unified Memory pointer when this kind of allocation is adopted.
 
**This strategy introduces a simple context to test the CUDA Unified Memory usage and study the possibility of overlapping memory transfers and calculations without the need to introduce CUDA streams.**

Concretely, this PR:
- introduces new CUDA Unified Memory allocation methods and asynchronous prefetching;
- introduces the GPU logic for the Jacobi preconditioner;
- introduces the GPU logic for the multi dot product and the linear combination (Modified Gram-Schmidt orthogonalization);
- introduces the GPU logic for the vector norm operation;
- introduces the GPU logic for generic scalar-vector unary operations through templates and generic vector/scalar-vector binary operations through Abstract Syntax Tree evaluated at runtime;
- finally falls back to CUDA Managed memory as highlighted in the following discussion. This introduces the need for explicity synchronization in all the custom setter/getter methods of the CSysVector.

This work is part of my ongoing contribution during the Google Summer of Code 2026 program. 

## Validation
Validated locally with:
- serial CUDA build compilation
- serial CPU build compilation
- CPU/GPU numerical comparison on 1 representative case (rae2822) tested with LINEAR_SOLVER_PREC=JACOBI with both CUDA Unified and Managed memory approaches.

Nsys profiling was performed to confirm the asynchronous prefetching of the CUDA Unified Memory preconditioning matrix on my local GPU. Partial prefetching is observed, although page faults were reported during the preconditioner CUDA kernel, indicating that the calculations were slowed down by the prefetching matrix still being transferred to the GPU. This overlap is expected to largely improve on more modern hardware; tests are ongoing in the cloud.

## Related Work
The Jacobi preconditioner kernels come from the PR #2825.

## Observed Issues
Some issues were observed during this first period of GSoC:
- PR #2825 compiles with CUDA 13.3 but CUSPARSE calls raise an unknown operation at the first matrix-vector product.
- the build.meson file has a hard-coded CUDA arch.
-  the HAVE_MPI flag is not being passed to the nvcc compiler in the develop branch. This raises a linking error if MPI operations are included within the *.cu files. That is not the case in the master branch.

## Next steps
I propose to continue working on the following steps:
- [X] extend the CUDA Unified Memory allocation to the CSysVector class
- [X] evaluate if it might be of interest to extend the CUDA Unified Memory allocation to the CSysMatrix class; Update: discussion with mantainer indicates preference for Managed Memory approach;
- [X] benchmark the multiDot approached: looped vs custom kernel batched vs cublas<t>gemm batched: batched approaches are convenient due to just one kernel launch overhead, cublas implementation shows moderate speed-up (0.01s) against the custom CUDA kernel on local P620 but ensures less code complexity, thus it is selected.

## PR Checklist


bigfooted:
That's great!
Can you run pre-commit to fix the code style/formatting?

pcarruscag (commented):
I think Areen looked into unified memory a bit.
Do you have some performance measurements? I imagine we save on the data transfer but suffer a little on the kernel performance?
Does unified memory introduce a significant restriction in terms of the required compute capability?

ddg93:
> That's great! Can you run pre-commit to fix the code style/formatting?

Thanks for the heads-up. Updated that commit and the following ones.

ddg93:
> I think Areen looked into unified memory a bit. Do you have some performance measurements? I imagine we save on the data transfer but suffer a little on the kernel performance? Does unified memory introduce a significant restriction in terms of the required compute capability?

For the moment, I've only one performance measurement on my local GPU Quadro P620: 
- GPU version based on Unified memory: Avg. s/iter:    0.256129
- CPU version: Avg. s/iter:    0.198681
This result is clearly not sufficient but shows a decent performance of the GPU version based on Unified memory on a small test-case and on a very modest GPU.
I intend to measure performance for a GPU version without Unified memory, and also consider larger tests and more advanced hardware (through cloud services).

Preliminary profiling shows indeed that the transfer of the preconditioning matrix is asynchronously launched but then the preconditioning kernel takes longer due to page fault. I intend to better profile what is going on with the CSysVector Unified memory implementation.

Unified memory is available for the compute capability 6.0 or higher. Some of its advanced features seem to depend on the system.

Is there some report by Areen? It could be interesting for me to read it.

pcarruscag:
Thanks I was wondering about unified vs explicit transfers not CPU vs GPU.
Not sure if Areen registered those tests somewhere.
Based on what nvidia says about unified memory, and the our goal with the linear solver, we should transfer explicitly and avoid complexity of keeping both strategies

ddg93:
Yes, I understand. I have further developed my PR to cover the whole FGMRES solver so that operations with CSysVectors run on GPU, operators too. In the last commit, I also fell back from Unified to Managed CUDA memory for the CSysVector. In this way, I could profile as you suggested. For the moment I could do this only on my local Quadro P620, I'll be trying to get better GPUs on the cloud. Please find here [SU2_ra2822_GPU_MA_vs_UM_nsys_prof.pdf](https://github.com/user-attachments/files/30518043/SU2_ra2822_GPU_MA_vs_UM_nsys_prof.pdf)
 a pdf report showing the differences between the nsys profiling output for Unified Memory (left column) (second last commit 5fe25b8) and Managed Memory (right column) (last commit 4b43fa0) for one run. UM is surprisingly competitive (should average on multiple runs and run on better GPUs to get more realistic results) and clearly facilitates the development as I did not have to safeguard the CPU manipulation of CSysVectors still on GPU.

Common/include/linear_algebra/gpu_ast.hpp
pcarruscag, line 1: Thanks but I want the compile-time version of this that is being worked in https://github.com/su2code/SU2/pull/2825/changes

ddg93: Ok. What do you mean by that, exactly? Should I adopt the specific implementation from https://github.com/su2code/SU2/pull/2825/changes or just follow the same approach? Besides the operators kernels, my implementation differs in other parts of the FGMRES solver, especially I introduced custom CUDA kernels rather than CUBLAS.

pcarruscag: I will verify that pr does what it should this weekend and merge it since I had already reviewed a few iterations of it.
The multi dot and other custom kernels you have here would be nice additions on top.
Unfortunately both PRs port the jacobi preconditioner, if you have time left in your project and want to take a look at completing the lu-sgs port, that would also be high value.

pcarruscag: PR did what it should after I got an LLM to clean it up.
I have a weekly budget left to spend, so I'm getting the AI to try and do the ILU preconditioner.
Let me know if you want to take on the tasks I suggested above, and sorry for the merge conflicts, we can't choose who/what/when contributes to SU2.

ddg93: Ok to add the multi dot and other custom kernels as additions on top. I assume this meas updating my PR over the current version of the develop branch. Then, I could propose the CUBLAS or Custom kernels path for the solver based on a pre-processor flag that checks if CUBLAS is available, what do you think?
 
Ok to work on the lu-sgs porting, I still have some time left with the GSoC so I am happy to continue working. Besides, I am enjoying this so I'll try to stay involved even after the project has ended.
No worries about the conflicts.

pcarruscag: yes bring this PR up to date with develop and we can merge the multiDot part, then the lu-sgs can be a second PR

ddg93: Ok, I've merged with develop and included the multiDot and the Linear Combination. Local benchmarks show that the multiDot offers a performance gain (lower time per timestep) compared to the current approach. This is less important for the Linear Combination. I'll run the benchmark on more advanced hardware to get a more complete picture.

ddg93:
> Thanks I was wondering about unified vs explicit transfers not CPU vs GPU. Not sure if Areen registered those tests somewhere. Based on what nvidia says about unified memory, and the our goal with the linear solver, we should transfer explicitly and avoid complexity of keeping both strategies

For the sake of completeness, I report a benchmark between Unified Memory (left column, commit https://github.com/su2code/SU2/pull/2843/commits/5fe25b814ff1c9b256a00a568f4c5e41be5c96ee ) and Managed Memory (right column, commit https://github.com/su2code/SU2/pull/2843/commits/4b43fa0d365475d5e500ff50a3fa1c73c0d13976 ) here:  : [SU2_rae2822_UM_vs_MA_NVIDIARTX2000AdaGene.pdf](https://github.com/user-attachments/files/30837151/SU2_rae2822_UM_vs_MA_NVIDIARTX2000AdaGene.pdf). I was able to run this on an RTX 2000 for the rae 2822 grid with 24576 elements.

Common/include/linear_algebra/gpu_ast.hpp
ddg93: Profiled for the rae2822 testcase, 24578 elements, averaged on 100 iterations, on an Nvidia RTX2000 Ada Gene. The average time per time-step is 0.457684 s/iter without custom kernels, stays to a similar value of 0.457895 s/iter when I introduce the custom LinearCombinationKernel only, drops to 0.272235 s/iter when I introduce the custom GPUmultiDot product only. 
Attached here the comparison between the two profiling stats: on the left the current develop implementation, on the right this PR latest implementation introducing the two custom kernels for multiDot and Linear Combination 
[SU2_rae2822_Nocustom_vs_multiDot_LinComb_custom_RTX2000.pdf](https://github.com/user-attachments/files/30855251/SU2_rae2822_Nocustom_vs_multiDot_LinComb_custom_RTX2000.pdf)
When implemented, the custom multiDot reduces the expired time per iteration; the custom linear Combination is not so relevant.

pcarruscag:
Good direction — fusing the `n*m` dots into one kernel removes develop's launch-bound loop of separate cuBLAS `dot` calls.

**Suggest dropping `LinearCombinationGPU`** and the `CSysSolve.cpp` hunk. You measured no speedup, which is expected since the expression templates already fuse the combination into a single kernel. It also writes `v.nElmDomain` where the expression path writes `nElm`, so the halos of `x`/`r` go stale and the FGMRES restart is wrong on 2+ ranks, and it skips the `n == 0` zeroing that `LinearCombinationImpl` does at the top. Both fixable, but with no perf win, removing is simpler.

**Blockers in `multiDot`:**

1. The `if (VecExpr::UseDeviceExpressions())` guard is gone and the host path moved under `#else`. A CUDA build running `ENABLE_CUDA= NO` now reads device pointers that were never uploaded, with no host fallback compiled at all. The same removal makes the `su2_gpu_capable_v` else-branch fire unconditionally, so CUDA+AD builds abort. Please keep the runtime guard and leave the host path compiled, as in `CSysMatrix.cpp:818` and `ApplyPreconditionerOnHost`.

2. `multiDotGPU` adds an `SU2_OMP_MASTER` and a trailing `SU2_OMP_BARRIER` inside the caller's `BEGIN_SU2_DEVICE_REGION`, which is already `barrier + omp master`. A barrier inside `master` is non-conforming and hangs at `OMP_NUM_THREADS > 1` — the macro's own comment says these regions must not be nested. Drop both; only the master thread runs that body, so the statics don't need protecting either.

3. `template class CSysVector<su2double>;` is unguarded (develop uses `#ifdef CODI_REVERSE_TYPE`). In a mixed-precision build `su2double == passivedouble`, so it duplicates the explicit member instantiations below it. Instantiate just `multiDotGPU` with the neighbouring guards — the new `USE_MIXED_PRECISION` guard can then go away.

**Determinism:** `atomicAdd` across blocks makes `V^T W` non-reproducible; the host code reduces in a fixed order on purpose. Since this is a tall-skinny GEMM, worth benchmarking `cublas<t>gemmBatched` before committing to the custom kernel — likely faster *and* deterministic.

**Minor:** 1024-thread blocks mean 8 KB shared memory per block for a reduction (128-256 is more typical), and the tree reduction assumes a power-of-two size — worth a `static_assert`. The device statics are never freed, `su2matrix local(n, m)` heap-allocates per call, and `multiDot`/`multiDotGPU` keep two separate `shared` statics with a copy between them. The `#ifndef SWIG` change is unrelated and hides `PassiveValue` from `pySU2.i` — please split it out or explain it. Note that `.cu` files aren't covered by the clang-format hook (`types_or: [c++, c]`), so the Allman braces / 4-space indent have to be matched by hand.

**Testing before this leaves draft:** `OMP_NUM_THREADS > 1`, a CUDA build running `ENABLE_CUDA= NO`, a mixed-precision build, and 2+ MPI ranks.

pcarruscag:
Sorry for the previous huge comment from Claude, I had it clean up, the findings look valid.

ddg93:
Thanks for the comment. I've removed the linear combination custom kernel and switched to the batched cublas<t>gemm for the multi dot product. I've just profiled on my local machine and there is just a sligth speed-up (0.01 s on average). 

I've cleaned the blockers 1,2,3 of the previous comment and addressed the minor points. 
I've tested the GPU build with OpenMP (4 threads) finding matching residuals along 100 iterations with the CPU build. 
The mixed precision build test shows just a slight difference in the last digits of the residuals against the CPU. 
Running with 2 MPI processes shows matching residuals for the first iteration only, then differences with the CPU build accumulate (I verified that this was not introduced by this PR contribution). 

Finally, I've tested the non-batched gemm operation (not included in the last commit). This requires a custom kernel to store the vectors in coalesced arrays. Ultimately, there is no speed-up against the custom multiDot cuda kernel, so I think the batched gemm is the best approach: simpler and faster.

pcarruscag (approved):

pcarruscag:
The custom kernel was actually the way to go, I think for the small case you tested on, the overhead of copying the pointers to device buffers lead to the wrong conclusion, or maybe the GPU you used does not have enough SMs to show the benefit vs the cublas gemm, which according to Claude parallelizes along the "wrong axis" for what multiDot needs. See #2869.
Let me know if you are going to look into the LU-SGS preconditioner, I can try to make this week's developer meeting if you want to talk about it there.

ddg93:
Very well, thanks for the update.
Yes, I've started looking into the LU-SGS preconditioner. We can surely discuss it on the next meeting or in the next PR that I'll open about it. My first approach will be to follow the ILU preconditioner logic.

PR #2846, opened 2026-07-16
Add support for Riemann boundary condition to all scalar solvers

aureliegrebe:
## Proposed Changes
Add support for Riemann boundary conditions to all scalar solvers by moving BC_Riemann definition from CTurbSolver to CScalarSolver.

I've only tested this for for CSpeciesSolver; I put the definition in CScalarSolver mainly just to avoid code duplication between CTurbSolver and CSpeciesSolver. Please let me know if there is a reason not to do this.

## PR Checklist

### DIFF
--- a/SU2_CFD/include/solvers/CScalarSolver.hpp
+++ b/SU2_CFD/include/solvers/CScalarSolver.hpp
@@ -494,6 +494,22 @@ class CScalarSolver : public CSolver {
   }
 
   /*!
+   * \brief Impose the boundary condition using characteristic recostruction.
+   * \param[in] geometry - Geometrical definition of the problem.
+   * \param[in] solver_container - Container vector with all the solutions.
+   * \param[in] numerics - Description of the numerical method.
+   * \param[in] config - Definition of the particular problem.
+   * \param[in] val_marker - Surface marker where the boundary condition is applied.
+   */
+  void BC_Riemann(CGeometry *geometry,
+                  CSolver **solver_container,
+                  CNumerics *conv_numerics,
+                  CNumerics *visc_numerics,
+                  CConfig *config,
+                  unsigned short val_marker) final;
+
+
+  /*!
    * \brief Impose the supersonic inlet boundary condition (same as inlet, see BC_Inlet).
    */
   void BC_Supersonic_Inlet(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics,
--- a/SU2_CFD/include/solvers/CScalarSolver.inl
+++ b/SU2_CFD/include/solvers/CScalarSolver.inl
@@ -371,6 +371,23 @@ void CScalarSolver<VariableType>::SumEdgeFluxes(CGeometry* geometry) {
   END_SU2_OMP_FOR
 }
 
+template<class VariableType>
+void CScalarSolver<VariableType>::BC_Riemann(CGeometry *geometry, CSolver **solver_container, CNumerics *conv_numerics, CNumerics *visc_numerics, CConfig *config, unsigned short val_marker) {
+  SU2_ZONE_SCOPED
+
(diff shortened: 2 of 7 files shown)


pcarruscag:
Thanks for this — moving `BC_Riemann` up to `CScalarSolver` to share it across turbulence and species is a nice simplification and the change itself looks correct as a relocation (dispatch to `BC_Inlet`/`BC_Outlet` is virtual, so it resolves properly for each solver).

While reviewing, we ran into a pre-existing gap that your change surfaces more visibly now that species goes through this path too: **scalar inlet values (turbulence/species) are never actually populated for a `MARKER_RIEMANN` boundary**, so `BC_Riemann` silently falls back to freestream turbulence/species values at that marker regardless of what's imposed on the flow side.

The root cause is in the inlet-filling logic, which only populates the per-marker scalar arrays when the marker's `KindBC` is `INLET_FLOW` (or `SUPERSONIC_INLET` for species) — a `RIEMANN_BOUNDARY` marker never qualifies:

- `CSpeciesSolver::SetUniformInlet` — `SU2_CFD/src/solvers/CSpeciesSolver.cpp:521`
- `CTurbSASolver::SetUniformInlet` — `SU2_CFD/src/solvers/CTurbSASolver.cpp:1954`

For species this means there's currently no way to specify the composition entering through a Riemann boundary at all — it's silently freestream regardless of `MARKER_INLET_SPECIES`. Turbulence has the same issue (pre-existing, not introduced by you) since `CConfig::GetInlet_TurbVal` falls back to freestream when the tag isn't found under `MARKER_INLET_TURBULENT`, which it never will be for a Riemann-tagged marker.

Could you extend the fix to make these `SetUniformInlet` (and equivalent) checks also apply to `RIEMANN_BOUNDARY` markers, so `MARKER_INLET_TURBULENT`/`MARKER_INLET_SPECIES` entries are actually honored there?

Lastly, could you add a test case (or adapt an existing regression case) that uses `MARKER_RIEMANN` with the species/turbulence solver, so this path gets exercised in CI going forward?

bigfooted:
@joshkellyjak can you have a look as well?

aureliegrebe:
> While reviewing, we ran into a pre-existing gap that your change surfaces more visibly now that species goes through this path too: **scalar inlet values (turbulence/species) are never actually populated for a `MARKER_RIEMANN` boundary**, so `BC_Riemann` silently falls back to freestream turbulence/species values at that marker regardless of what's imposed on the flow side.

I've added logic to the `CSpeciesSolver::SetUniformInlet` that sets the inlet values according to `MARKER_INLET_SPECIES` if the Riemann boundary is treated as an inlet.

> For species this means there's currently no way to specify the composition entering through a Riemann boundary at all — it's silently freestream regardless of `MARKER_INLET_SPECIES`. Turbulence has the same issue (pre-existing, not introduced by you) since `CConfig::GetInlet_TurbVal` falls back to freestream when the tag isn't found under `MARKER_INLET_TURBULENT`, which it never will be for a Riemann-tagged marker.

I'm not entirely convinced there was actually a bug in the turbulent solvers causing `MARKER_INLET_TURBULENT` to be ignored. This value is never read in `SetUniformInlet` under any case, and is read in the `BC_Inlet`, which `BC_Riemann` calls. As far as I can tell, the value set in `SetUniformInlet` is only used when inlet-files are used, which is not supported for Riemann boundary conditions.

> Lastly, could you add a test case (or adapt an existing regression case) that uses `MARKER_RIEMANN` with the species/turbulence solver, so this path gets exercised in CI going forward?

I've copied and modified the `axi_rans_air_nozzle_restart` to create `axi_rans_air_nozzle_species`.

pcarruscag (approved):

bigfooted:
@aureliegrebe Thank you for this contribution!

PR #2849, opened 2026-07-21
Fix MG species restart crash

aureliegrebe:
## Proposed Changes
Currently, the species solver crashes when restarting a multi-grid simulation due to mismatched grid levels in two calls in `CSpeciesSolver::LoadRestart`. This PR fixes that, and modifies a test to exercise this code path.

It'd be great if someone could also double-check whether this code makes sense. Given this bug, I suspect these lines have never been run since they were committed 5 years ago. 

## Related Work
I'll send an accompanying PR for the new restart file needed for the test case.

## PR Checklist


bigfooted:
Thanks! Note that multigrid is currently only implemented for the flow equations, so species and turbulence are running the single grid solver. 
And multigrid is still not doing a very good job in many cases. Did you get good convergence for this testcase? Because the current settings are not very good.

bigfooted:
also needed for turbulence, I'll pick that up.

bigfooted (approved):

aureliegrebe:
> Did you get good convergence for this testcase? Because the current settings are not very good.

To be honest, I didn't really check convergence in the test regression test that I modified, I was more focused on removing the segfault. In my internal test cases (a Mach 3 nozzle), I only use multi-grid during the flow initialization with uniform species, although I have regularly had good convergence with MG + the SST solver on a variety of cases.

Issue #2353, opened 2024-09-06
Non-Restart Unsteady `Time`, `TimeIter`, and IC Outputs Not Correct?


j-signorelli:
There seems to potentially be a bug with doing an unsteady dual-timestepping simulation without a provided initial condition (restart file):

When no restart file is provided, the first inner iterations have Time_Iter=0 and Cur_Time=0 (the state **before** these inner iterations). After completion of this time iteration, a file `flow_00000.vtu` and all others are outputted. These files are _not_ the initial condition as specified by the configuration file -- they seem to actually be the solution after the first time step. This can be reproduced easily in the config file I have attached to this issue (an updated version of TestCases/plunging_naca0012 for SU2 v8). The `flow_00000.vtu` file has non-uniform flowfields, which doesn't make sense.

For simulations where a restart file like `restart_00000.dat` is provided, the Time_Iter=1 and Cur_Time=dt (the state **after** these inner iterations), and the subsequently outputted files have the correct time iteration appended to it.

I think that for non-restart unsteady simulations, the IC should be outputted as solely just the initialization before any inner iterations are completed, Time_Iter++ and Cur_Time+=dt, and THEN inner iterations performed.

[plunging_NACA0012.cfg.txt](https://github.com/user-attachments/files/16912625/plunging_NACA0012.cfg.txt)

**Bug report checklist**
*Please make sure that you have followed the checklist below, many common problems can be solved by:*

- [X] Consulting the build instructions (https://su2code.github.io/docs_v7/Build-SU2-Linux-MacOS/).
- [X] Looking for similar problems on GitHub or CFD-Online (https://www.cfd-online.com/Forums/su2/).
- [X] Updating to the newest version of SU2 (either develop, master, or the pre-built executables https://su2code.github.io/download.html).

**Desktop (please complete the following information):**
 - OS: Linux (Ubuntu 22.04)
 - C++ compiler and version: GNU
 - MPI implementation and version: MPICH
 - SU2 Version: v8.0.1

pcarruscag:
The solution is always at the end of time steps, to restart at time iter 2 you need 0 and 1

j-signorelli:
[...]

PR #2857, opened 2026-07-31
Fix implicit target-time evaluation

bigboateng:
## What changed

- centralize projection from zero-based time-iteration labels to physical solver time
- evaluate direct dual-time target states at `(TimeIter + 1) * deltaT`
- use the projected time for grid and surface motion, gusts, solver globals, and history output
- preserve zero-based output/restart labels and existing adjoint and harmonic-balance interpretations
- add configuration and first-step rigid-motion regression coverage

## Why

In implicit dual-time schemes, `TimeIter` is a zero-based artifact/restart label, while the state being solved represents the end of that physical step. Several consumers reconstructed physical time directly as `TimeIter * deltaT`, causing the first implicit state to use time zero for time-varying motion and forcing terms.

This separates artifact identity from target-state time without renumbering restart or output files.

Related to su2code/SU2#2353.

## Validation

- built `SU2_CFD/src/SU2_CFD`
- built `UnitTests/test_driver`
- passed `[CConfig][time]`: 2 test cases, 12 assertions
- bounded dual-time integration case retained `Time_Iter = 0` and `flow_00000.vtk` while reporting `Cur_Time = 0.01`

The full SU2 regression matrix and complete adjoint regression suites were not run.

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -3401,6 +3401,30 @@ public:
   su2double GetPhysicalTime(void) const { return PhysicalTime; }
 
   /*!
+   * \brief Get the physical time represented by the completed state of the current time iteration.
+   * \return Nondimensional physical time at the end of the current time step.
+   */
+  su2double GetPhysicalTimeAtEndOfTimeStep(void) const {
+    if (!ContinuousAdjoint && !DiscreteAdjoint && TimeMarching == TIME_MARCHING::TIME_STEPPING)
+      return PhysicalTime + Delta_UnstTimeND;
+    return PhysicalTime;
+  }
+
+  /*!
+   * \brief Project a zero-based time iteration label onto the physical time represented by the solver state.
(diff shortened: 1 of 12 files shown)


bigboateng:
@pcarruscag I worked on this a while ago based on #2353, but I’ve since lost some of the context. Would you mind taking a quick look and letting me know whether the approach is directionally correct? If it looks useful, I can rebase the branch and resume the validation work; otherwise I’m happy to close the draft.

pcarruscag:
To be honest I would rather not make changes here because it's a large blast radius that would affect any existing restarts for moving mesh cases, and it's hard to reason about the logic for time accurate discrete adjoints...
To my knowledge no one is complaining that this is broken... so I would rather no get complaints because we fixed it lol.

bigboateng:
Thanks for the context. That makes sense given the blast radius, the restart and discrete-adjoint uncertainty, and the lack of reports that the current behavior is causing problems. I’ve closed this.

Issue #2313, opened 2024-06-27
FSI - Issue with `pysu2_nastran.py` when Reading SET1 from Complex Nastran .f06 Files


safadumanli:
Description:

There is a problem in the `pysu2_nastran.py` script when reading the SET1 section from Nastran .f06 output files. The current implementation of the function that reads the SET1 section is insufficient for more complex, such as 3D, mesh structures.

In the tutorial, a simple airfoil case is used with a mesh consisting of 123 surface nodes, which works fine. However, for more complex mesh structures, the SET1 section in the Nastran .f06 file includes additional lines with dates, headers, and the Nastran program's name. These extra lines cause issues while reading the SET1 section, leading to errors in processing the file.

https://github.com/su2code/Tutorials/blob/master/multiphysics/unsteady_fsi_python/Ma01/modal.f06

line 376 starts with SET1. There is no additional line between the SET1 ID's. But for more complex mesh, additional lines look like this:

"              191305-        SET1    1       1       2       3       4       5       6       7       +^M
               191306-        +       8       9       10      11      12      13      14      15      +^M
               191307-        +       16      17      18      19      20      21      22      23      +^M
               191308-        +       24      25      26      27      28      29      30      31      +^M
               191309-        +       32      33      34      35      36      37      38      39      +^M
               191310-        +       40      41      42      43      44      45      46      47      +^M
               191311-        +       48      49      50      51      52      53      54      55      +^M
               191312-        +       56      57      58      59      60      61      62      63      +^M
               191313-        +       64      65      66      67      68      69      70      71      +^M
               191314-        +       72      73      74      75      76      77      78      79      +^M
               191315-        +       80      81      82      83      84      85      86      87      +^M
               191316-        +       88      89      90      91      92      93      94      95      +^M
               191317-        +       96      97      98      99      100     101     102     103     +^M
               191318-        +       104     105     106     107     108     109     110     111     +^M
               191319-        +       112     113     114     115     116     117     118     119     +^M
               191320-        +       120     121     122     123     124     125     126     127     +^M
               191321-        +       128     129     130     131     132     133     134     135     +^M
               191322-        +       136     137     138     139     140     141     142     143     +^M
               191323-        +       144     145     146     147     148     149     150     151     +^M
               191324-        +       152     153     154     155     156     157     158     159     +^M
               191325-        +       160     161     162     163     164     165     166     167     +^M
               191326-        +       168     169     170     171     172     173     174     175     +^M
[...]

PR #2859, opened 2026-08-04
Fix parsing of Nastran small field reals in the modal structural solver

binggao1230:
## Proposed Changes

`SU2_PY/SU2_Nastran/pysu2_nastran.py` reads the GRID and CORD2R entries of a Nastran sorted bulk data echo by slicing eight character fields and then interpreting each slice *positionally*. Two helpers do that interpretation, and both are only correct when the value happens to be left flush and to fill all eight columns:

```python
def nastran_float(s):
    if s.find("E") == -1:
        s = s.replace("-", "e-")
        s = s.replace("+", "e+")
        if s[0] == "e":        # only removes the spurious "e" if it lands on index 0
            s = s[1:]
    return float(s)

def __checkBlankField(self, string):
    if string == " " * 8:      # only recognises a blank field exactly eight characters wide
        return int(0)
    return int(string)
```

Neither assumption is guaranteed. The Quick Reference Guide, *Format of Bulk Data Entries -> Small Field Format*, says:

> Fields 1 and 10 must be left justified.
> Fields 2 through 9 do not need to be either right or left justified, although aligning the data fields is good practice.

and on the same page:

> Real numbers may be entered in a variety of ways. For example, the following are all acceptable versions of the real number seven:
> `7.0    .7E1    0.7+1    .70+1    7.E+0    70.-1`

So three families of legal input abort the FSI run before it starts:

1. **Any negative real that is not left flush.** `nastran_float("    -5.2")` builds `"    e-5.2"` and raises `ValueError`. This does not need the E-less exponent form at all; a plain right justified negative coordinate is enough. pyNastran, for one, writes every small field with `'%8s'`/`'%8d'`/`'%8.Nf'`, i.e. right justified.
2. **Lower case `e` and `D` exponents.** `"700.e-2"` becomes `"700.ee-2"` and `".7D1"` is never converted, in *any* justification. Both are accepted real formats (`.1d-5` is in the OptiStruct bulk data guidelines, and the QRG notes that "a double precision specification requires a 'D' type exponent").
3. **A GRID entry without the optional CD field.** Nastran trims the trailing blanks of each echo line, so the entry ends after X3 and `line[48:56]` is `''` -> `int('')` raises. In the tutorial mesh (`Tutorials/multiphysics/unsteady_fsi_python/Ma01/modal.f06`) CD is written explicitly, which is why this has not shown up; one column shorter and it is an empty slice.

The three symptoms are one root cause, so the fix is at the two helpers rather than at the twelve GRID/CORD2R call sites, which are unchanged. `nastran_float` normalises the field first and then inserts the omitted exponent letter at the first sign after the mantissa sign, which is what the established readers do; `__checkBlankField` treats any whitespace-only or absent field as blank. `nastran_float` moved to module level so it can be tested — it was a closure inside `__readNastranMesh`.

## Verification

I enumerated the QRG spellings of a real crossed with sign and with left/right/centre justification in an eight column field: **108 fields, 49 of them rejected or miscomputed before, 0 after**. The failures group as:

| family | left | right | centre |
|---|---|---|---|
| QRG spellings of seven | 0 | 8 | 8 |
| lower case `e` / `D` exponent | 9 | 13 | 11 |

`SU2_PY/SU2_Nastran/test_pysu2_nastran.py` is new: it drives that table through `nastran_float`, and then builds the same little model (a CORD2R plus three GRIDs with negative coordinates, an E-less exponent, a blank CP and an omitted CD) in each justification and reads it through `__readNastranMesh`, asserting the parsed geometry does not depend on how the fields were written. 10 tests, all 10 fail on `develop` and pass here.

There is no Python test harness in the repository at the moment, so this is plain `unittest` with no new dependency, run with:

```
python -m unittest discover -s SU2_PY/SU2_Nastran
```

Happy to drop the file if you would rather not start a Python test directory here.

As a no-regression check I read the real tutorial `modal.f06` (124 grid points, 41 negative coordinate values) with `develop` and with this branch: the parsed coordinates, IDs, CP, CD and marker sets are bit-for-bit identical. `pre-commit run` is clean (black 22.6.0). I did not build the C++ side; nothing outside this Python module is touched.

## Related Work

No related PR that I can find. #2313 is a different problem in the same file (page headers interrupting a SET1 continuation in the echo) and is not addressed here.

## PR Checklist

- [x] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson). Nothing compiled changes.

### DIFF
--- a/SU2_PY/SU2_Nastran/pysu2_nastran.py
+++ b/SU2_PY/SU2_Nastran/pysu2_nastran.py
@@ -34,6 +34,30 @@ import scipy.linalg as linalg
 from math import *
 
 # ----------------------------------------------------------------------
+#  Bulk data field parsing
+# ----------------------------------------------------------------------
+
+
+def nastran_float(s):
+    """
+    This method converts one small field of a bulk data entry into a float.
+
+    Fields 2 through 9 do not need to be either right or left justified, and
+    the exponent may be written with an "E", with a "D", or with no letter at
+    all: "7.0", ".7E1", "0.7+1", ".70+1", "7.E+0" and "70.-1" all mean seven.
+    """
+
+    s = s.strip().upper().replace("D", "E")
+    if "E" not in s:
+        # The exponent letter is omitted, so the exponent starts at the first
+        # sign which is not the sign of the mantissa.
+        signs = [i for i in (s.find("+", 1), s.find("-", 1)) if i > 0]
+        if signs:
+            s = s[: min(signs)] + "E" + s[min(signs) :]
+    return float(s)
+
+
+# ----------------------------------------------------------------------
 #  Config class
 # ----------------------------------------------------------------------
 
@@ -426,14 +450,6 @@ class Solver:
         This method reads the nastran 3D mesh.
         """
 
-        def nastran_float(s):
-            if s.find("E") == -1:
-                s = s.replace("-", "e-")
-                s = s.replace("+", "e+")
-                if s[0] == "e":
-                    s = s[1:]
-            return float(s)
-
         self.nMarker = 0
         self.nPoint = 0
         self.nRefSys = 0
@@ -573,7 +589,8 @@ class Solver:
         This method considers that Nastran apply 0 when the reference system is not specified
         """
 
-        if string == " " * 8:
+        string = string.strip()
+        if not string:
             return int(0)
         return int(string)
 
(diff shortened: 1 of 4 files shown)


pcarruscag:
Thank you. Can you call your test script as part of our serial_regression so that it runs under CI?

binggao1230:
Done — `serial_regression.py` now runs `test_pysu2_nastran.py` with the active Python interpreter and includes its result in the serial-test summary.

Local checks: 10/10 parser tests pass, both Python files compile, and pre-commit passes on the touched files. Pushed as `9d1d7f9`.

pcarruscag:
seems that the file is not found in CI
```/usr/bin/python: can't open file '/github/workspace/tests/SU2_PY/SU2_Nastran/test_pysu2_nastran.py': [Errno 2] No such file or directory```

pcarruscag (approved):

PR #2865, opened 2026-08-09
Fix biparabolic regression

### LINKED ISSUE #2864: biparabolic airfoil not in regression tests
**Is your feature request related to a problem? Please describe.**
The biparabolic airfoil in Testcases/euler/biparabolic 
is not in a regression test. What is needed:
1. update the config file
2. test with current develop version
3. optimize settings for best convergence
4. add to regression test

Syphonicc:
## Proposed Changes

Updates the biparabolic supersonic Euler case (`TestCases/euler/biparabolic`) so it runs on current `develop`, and adds it to `serial_regression.py`.

Config changes:
- `EXT_ITER` → `ITER` (deprecated option prevented the config from parsing at all)
- Fixed `CFL_ADAPT_PARAM`: factor-down and factor-up were swapped (`1.5, 0.5` → `0.5, 1.5`). SU2 rejects factor-down ≥ 1.0, but this was never caught because `CFL_ADAPT` was set to `NO`.
- Switched `CONV_NUM_METHOD_FLOW` from JST to ROE with `MUSCL_FLOW` and the Venkatakrishnan limiter. The case diverged with JST at CFL 25, 5 and 1, so it wasn't a timestep issue — the central scheme wasn't robust enough for the shocks at Mach 1.7.
- `MGLEVEL` 3 → 0
- Removed unused adjoint, design-variable, and equivalent-area settings

**Note on scope:** the original config was set up for an adjoint-based sonic-boom (equivalent-area) design study rather than a direct solve, and the `MARKER_NEARFIELD` / `EQUIV_AREA` setup appears to be what was causing the divergence. I reduced it to a minimal direct Euler run so it works as a regression case  happy to restore the design-related settings if you'd prefer the case keep that capability.

Regression entry uses `test_iter = 25`, with values verified reproducible across repeated runs. Residuals stall around -4 rather than reaching the -6 criterion at longer horizons  let me know if you'd prefer different settings.

Tested locally with a serial build (MPI disabled). I'm new to SU2, and worked through this with AI assistance.

## Related Work

Resolves #2864.

## PR Checklist


bigfooted:
Thanks! Please keep the adjoint stuff so the file can then be re-used for an adjoint setup as well. With these settings, does it converge completely, can you post a plot of the residuals as function of iterations here?
The residual values reported by github are slightly different from the ones you computed yourself, you can just copy-paste the github values and update them.

Syphonicc:
> Thanks! Please keep the adjoint stuff so the file can then be re-used for an adjoint setup as well. With these settings, does it converge completely, can you post a plot of the residuals as function of iterations here? The residual values reported by github are slightly different from the ones you computed yourself, you can just copy-paste the github values and update them.



Adjoint settings restored put back the adjoint, DV, and adjoint I/O options. They're inert in a DIRECT run and the regression values are unchanged with them present.

EQUIV_AREA / MARKER_NEARFIELD I left commented out, with a note in the file explaining why. I tested this specifically: with everything else identical, re-enabling them gives Warning: there are 760 non-physical points in the solution and rms[Rho] sits at +3.43 at iteration 500, versus −3.7 without them. Happy to re-enable if you'd rather the direct case carry them, but it doesn't produce a usable solution as-is.

Convergence no, it doesn't converge completely. Residuals drop quickly for ~150 iterations, then plateau around −4 and stay there. I ran it out to 3000 iterations to check: rms[Rho] was −4.05 at 2000 and −3.93 at 3000, so it's a genuine stall rather than slow progress. Plot below.
biparabolic_residuals

bigfooted:
That convergence doesn't look good. Also, the equivalent area and nearfield conditions are important parts of this setup so I think it is best to keep those, since we already have a bunch of testcases that test the standard euler flow with a farfield BC. That probably means tinkering a bit with the settings to find out how to make it converge well.

Syphonicc:
> That convergence doesn't look good. Also, the equivalent area and nearfield conditions are important parts of this setup so I think it is best to keep those, since we already have a bunch of testcases that test the standard euler flow with a farfield BC. That probably means tinkering a bit with the settings to find out how to make it converge well.

I dug into the nearfield setup and I think the problem is deeper than config settings. Findings, all tested with everything else held constant:

MARKER_NEARFIELD alone (with EQUIV_AREA= NO) gives 760 non-physical points and rms[Rho] = 3.4319. Enabling EQUIV_AREA= YES on top gives bit-identical results, so the EA post-processing isn't involved  it's the nearfield BC itself.
Without the nearfield markers, rms[Rho] reaches −3.9 with no warnings.
I also corrected EA_INT_LIMIT from (-100, 100, 2.0) to (1.562, 4.562, 2.0) to match the actual nearfield extent in the mesh no effect on the solution.
In the mesh, upper_nearfield and lower_nearfield are coincident duplicate nodes, both on the line y = −2 with bit-identical x-coordinates (e.g. node 8 and node 56581 are both at x = 1.562149913, y = −2.0). So this is an internal interface construction rather than two surfaces above and below the airfoil.
The one working EA case in the suite, ea_naca64206, is 3D with a circumferential nearfield surface, which matches what config_template.cfg describes as required.

So this looks like the 2D coincident-node nearfield path being broken rather than a settings issue  probably needs a look at the nearfield BC implementation. Happy to keep going if you can point me at the relevant code, or to close this if you'd rather handle it separately.

PR #2871, opened 2026-08-17
[WIP] cache Least Squares matrix for gradient computation

tbellosta:
## Proposed Changes

  This PR ~adds an opt-in option, `CACHE_LSQ_METRICS` (default `NO`),~  caches the factorized metric terms of the (weighted or unweighted) least-squares gradients. The LSQ normal matrix A depends only on the node coordinates and on the weighting, yet it is currently re-assembled and re-factorized on every gradient evaluation, every nonlinear iteration. With the option enabled:

  - On the first evaluation the normal matrix is assembled and factorized as usual, and its inverse S = A^-1 (6 unique entries per node in 3D, 3 in 2D) is stored in `CGeometry`.
  - Every subsequent evaluation only accumulates the right-hand side b = sum_k w*d*(u_k - u_i) in a loop over the edges (using the edge coloring; the contribution of an edge is identical for both of its end points, so each edge is visited once) followed by one small symmetric matrix-vector product S*b per node.
  - The cache is stored per grid level with one slot per weighting, shared by all solvers (flow primitive and reconstruction gradients, turbulence/species/scalar solution gradients, auxiliary variables). The common practice of unweighted LSQ for MUSCL reconstruction plus inverse-distance-weighted LSQ for viscous gradients is fully supported: both slots are built once and reused; e.g. the turbulence solver directly reuses the metrics built by the flow solver.
  - Moving/deforming grids are supported: the cache is invalidated in `SetControlVolume(..., UPDATE)` (the common point of all mesh motion/deformation paths, on fine and coarse MG levels) and rebuilt on the next evaluation, i.e. once per mesh update, so the savings scale with the number of inner iterations per time step.
  - The option is automatically disabled (with a warning) for periodic boundaries (the periodic LSQ communication fuses the matrix and RHS accumulations; an RHS-only exchange would be needed) and for the discrete adjoint (the coordinate dependence of the metrics must remain on the tape).

  #### Saved operations (per node per gradient evaluation, 3D, k edge-neighbors, N variables)

  Eliminated from every evaluation after the first:
  - normal-matrix assembly: ~21k flops (~300 flops per node on tetrahedra, k=14),
  - per-node Cholesky factorization + inversion: ~35 flops and 3 sqrt (no sqrt/div remain in the iteration path),
  - half of the geometric edge work (distance vectors and solution differences are computed once per edge instead of once per node-neighbor pair).

  What remains is the RHS accumulation (~4N flops per node per edge) and one S*b product (18N flops per node). For the N=6 compressible primitives this roughly halves the gradient-kernel flops; for scalar solvers (N=1-2, e.g. SA) where the assembly dominates it is a ~4-5x reduction. Measured end-to-end (serial, linear-solver work held fixed): ~3.3% of total wall time on an inviscid ONERA M6 (57.5k nodes, tets, one gradient set per iteration) and ~6% on a 2D RANS-SA case (three LSQ gradient sets per iteration). Cases dominated by the linear solver will see proportionally less.

  #### Memory footprint
  
  The cache adds nDim*(nDim+1)/2 su2doubles per node per weighting used (48 B/node per weighting in 3D, 24 B/node in 2D; at most two weightings), allocated lazily per grid level and shared by all solvers. For reference, every solver already allocates an nDim*nDim `Rmatrix` scratch (72 B/node in 3D), so for a typical RANS case the addition is small compared to the existing gradient machinery and negligible compared to overall solver memory.

  #### Validation

  With the option enabled, results are identical to `develop` (to output precision) for:
  - inviscid NACA0012, unweighted LSQ reconstruction, 200 iterations (serial and 2 MPI ranks),
  - RANS-SA NACA0012 with `NUM_METHOD_GRAD= WEIGHTED_LEAST_SQUARES` and `NUM_METHOD_GRAD_RECON= LEAST_SQUARES` (all residuals including the turbulence variable),
  - an unsteady pitching NACA0012 with `GRID_MOVEMENT= RIGID_MOTION` (dual time stepping), exercising the invalidation/rebuild path,
  - hybrid-parallel runs (OpenMP edge-colored loop, bit-reproducible run-to-run; falls back to a race-free node loop if the natural coloring is in use).

  ~`config_template.cfg` documents the new option. Default behavior (`CACHE_LSQ_METRICS= NO`) is bit-identical to current develop.~
  
- [ ] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cfg), if necessary.

### DIFF
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -135,7 +135,8 @@ private:
   Hold_GridFixed,           /*!< \brief Flag hold fixed some part of the mesh during the deformation. */
   Axisymmetric,             /*!< \brief Flag for axisymmetric calculations */
   Enable_Cuda,              /*!< \brief Flag for switching GPU computing*/
-  Integrated_HeatFlux;      /*!< \brief Flag for heat flux BC whether it deals with integrated values.*/
+  Integrated_HeatFlux,      /*!< \brief Flag for heat flux BC whether it deals with integrated values.*/
+  Cache_LSQ_Metrics;        /*!< \brief Cache the factorized least-squares metric terms (rebuilt after mesh motion). */
   su2double Buffet_k;       /*!< \brief Sharpness coefficient for buffet sensor.*/
   su2double Buffet_lambda;  /*!< \brief Offset parameter for buffet sensor.*/
   su2double Damp_Engine_Inflow;   /*!< \brief Damping factor for the engine inlet. */
@@ -9732,6 +9733,11 @@ public:
    * \return YES if the passed values is the integrated heat flux over the marker's surface.
    */
   bool GetIntegrated_HeatFlux() const { return Integrated_HeatFlux; }
+
+  /*!
+   * \brief Reuse the factorized least-squares metric terms across gradient evaluations.
+   */
+  bool GetCache_LSQ_Metrics() const { return Cache_LSQ_Metrics; }
 
   /*!
    * \brief Get Compute Average.
--- a/Common/include/CConfig.hpp
+++ b/Common/include/CConfig.hpp
@@ -135,8 +135,7 @@ private:
   Hold_GridFixed,           /*!< \brief Flag hold fixed some part of the mesh during the deformation. */
   Axisymmetric,             /*!< \brief Flag for axisymmetric calculations */
   Enable_Cuda,              /*!< \brief Flag for switching GPU computing*/
-  Integrated_HeatFlux,      /*!< \brief Flag for heat flux BC whether it deals with integrated values.*/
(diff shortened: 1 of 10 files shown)


pcarruscag:
Nice. Make it the default, compute in geometry preprocessing when necessary, no opt-in

pcarruscag:
The failed tests seem to be just due to floating-point ordering. Is the PR ready for review?

tbellosta:
> The failed tests seem to be just due to floating-point ordering. Is the PR ready for review?

Kind of. I wanted to run some more detailed measurements on larger runs and add the caching for periodic boundaries as well. I will have some spare time next week to do that. I also have to run the pre-commit hook

PR #2858, opened 2026-08-03
Port ILU to GPU

pcarruscag:
## PR Checklist
- [X] I am submitting my contribution to the develop branch.
- [X] My contribution generates no new compiler warnings (try with --warnlevel=3 when using meson).
- [X] My contribution is commented and consistent with SU2 style (https://su2code.github.io/docs_v7/Style-Guide/).
- [X] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [ ] I have added a test case that demonstrates my contribution, if necessary.
- [X] I have updated appropriate documentation (Tutorials, Docs Page, config_template.cpp), if necessary.

PR #2872, opened 2026-08-18
[GSoC] Feature: LU-SGS preconditioner to GPU

ddg93:
This PR ports the LU-SGS preconditioner to GPU

## Proposed Changes
This PR introduces a build phase for the LU-SGS preconditioner where the inverse of the diagonal D is calculated to facilitate the compute phase. Then, forward and backward sweeps are implemented through custom CUDA kernels matching the logic of the CPU section (except for the inverse of D, which is directly multiplied in the GPU logic). These kernels are managed by the compute method on GPU which maintains the MPI communications.

## Related Work
This PR follows the discussion of #2843 . Kernel level dependencies are managed as for the ILU preconditioner proposed in #2858 .

## Validation
CPU and GPU implementations of the LU-SGS preconditioner are compared for validation of this PR on the rae2822 test-case. After one iteration, the first, preconditioned Krylov vector is compared finding matching results down to machine precision. Residuals are compared after 100 iterations finding matching results. 

## TO-DO
- [X] complete with the quantized_mode option path;
- [ ] run benchmarks on modern GPUs. Current evaluations on a P620 are inconclusive;
- [ ] profiling of the custom kernels;

## PR Checklist


Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -405,6 +405,15 @@ class CSysMatrix {
...
   mutable struct CUevent_st* htd_event = nullptr;
 
+  vector<su2uint> lusgs_level_ptr;      /*!< \brief Start of each level in d_lusgs_level_idx, size nLevels+1. */
+  su2uint* d_lusgs_level_idx = nullptr; /*!< \brief Row indices, grouped by level. */
+  mutable struct CUgraphExec_st* lusgs_fwd_graph_exec = nullptr;
+  mutable struct CUgraphExec_st* lusgs_bwd_graph_exec = nullptr;
+  mutable const ScalarType* lusgs_fwd_graph_vec = nullptr;
+  mutable const ScalarType* lusgs_bwd_graph_vec = nullptr;
+  mutable ScalarType* lusgs_fwd_graph_prod = nullptr;
+  mutable ScalarType* lusgs_bwd_graph_prod = nullptr;
```
pcarruscag: Some of this can be re-used from ILU, like the aux_stream that used to be just for ILU.

Common/src/linear_algebra/CSysMatrix.cpp
```diff
@@ -416,6 +420,31 @@ void CSysMatrix<ScalarType>::Initialize(unsigned long npoint, unsigned long npoi
...
+    level_idx.reserve(nPointDomain);
+    lusgs_level_ptr.clear();
+    lusgs_level_ptr.push_back(0);
+    for (auto level = 0ul; level < levels_lusgs.getOuterSize(); ++level) {
+      for (auto k = 0ul; k < levels_lusgs.getNumNonZeros(level); ++k) {
+        level_idx.push_back(static_cast<su2uint>(levels_lusgs.getInnerIdx(level, k)));
+      }
+      lusgs_level_ptr.push_back(static_cast<su2uint>(level_idx.size()));
+    }
+    d_lusgs_level_idx = GPUMemoryAllocation::gpu_alloc_cpy(level_idx.data(), level_idx.size() * sizeof(su2uint));
```
pcarruscag: The logic can be shared with the ILU for example use lambda that takes the sparse pattern as argument.

Common/src/linear_algebra/CSysMatrixGPU.cu
```diff
@@ -694,6 +694,228 @@ void CSysMatrix<ScalarType>::ComputeILUPreconditionerGPU(const CSysVector<Scalar
...
+/*!
+ * \brief Exact forward substitution for the rows of one level, x* = D^{-1}.(b-Lx*)
+ * \note See notes in IluForwardKernel for more details.
+ */
+template <class ScalarType>
+__global__ void LU_SGS_ForwardKernel(const su2uint* __restrict__ level_idx, unsigned long level_begin,
+                                     unsigned long level_size, unsigned long nVar, DeviceLDU<ScalarType> M,
+                                     const ScalarType* __restrict__ invD, const ScalarType* __restrict__ vec,
+                                     ScalarType* __restrict__ prod) {
+  if (blockIdx.x >= level_size) return;
```
pcarruscag, line 751: Does it have to be a different kernel? can we share some parts of the forward/backward with the ILU counterparts?

Common/src/linear_algebra/CSysMatrixGPU.cu
```diff
@@ -694,6 +694,228 @@ void CSysMatrix<ScalarType>::ComputeILUPreconditionerGPU(const CSysVector<Scalar
...
+void CSysMatrix<ScalarType>::ComputeLU_SGSPreconditionerGPU(const CSysVector<ScalarType>& vec,
+                                                            CSysVector<ScalarType>& prod, CGeometry* geometry,
+                                                            const CConfig* config) const {
+  SU2_ZONE_SCOPED
+
+  if (d_invM == nullptr) {
+    SU2_MPI::Error("CUDA LU-SGS preconditioner used without device storage.", CURRENT_FUNCTION);
+  }
+  if (nPointDomain == 0) return;
+
```
pcarruscag, line 852: Can we split it into 2 parts and have the CPU version call each one? similar to what it does now for normal vs quantized?
Then the MPI parts stay outside, if it gets too complicated with the OpenMP directives, don't worry.

pcarruscag:
Oh and add yourself to AUTHORS.md please

Common/include/linear_algebra/CSysMatrix.hpp
ddg93: yes, I have unified the variables where possible.

Common/src/linear_algebra/CSysMatrix.cpp
ddg93: yes, I've formatted the preconditioners preparation so that common procedures are shared. That logic is now a lambda.

Common/src/linear_algebra/CSysMatrixGPU.cu
ddg93: The ILU forward kernel could be extended to get the LUSGS forward, but that would imply two kernel calls. Instead, it is possible to introduce common operations as device kernels and build all the preconditioner kernels accordingly. This latter approach is now implemented.

Common/src/linear_algebra/CSysMatrixGPU.cu
ddg93: Thanks for this suggestion. Indeed, I've introduced the forward and backward methods for the LU-SGS preconditioner, inside which the CPU/GPU switching is managed. Like that, the MPI part stays outside and stays unique.

ddg93:
The pointed issues have been addressed for the GPU LU-SGS. Also, I've added the Quantized mode through a bool runtime control passed to the kernels.

I've performed some tests on my CPU with the RAE2822 test-case: 
- serial execution: CPU and GPU residuals perfectly match;
- OPENMP, 4 threads: CPU and GPU residuals match down to machine precision;
- MPI, 2 and 4 tasks: residuals diverge rapidly after few iterations, but remain comparable.

I am wondering if this divergence between CPU and GPU when running with multiple MPI tasks is expected?

To be sure, I run the same test with the ILU preconditioner and the version of the HEAD of the develop branch: matched residuals in serial and 4 OPENMP threads, non matching residuals when running with 4 MPI threads. Moreover, the GPU residuals drop and become positive. For this reason, I though that the issue might not be in the modifications introduced in this PR and pushed.

Common/include/linear_algebra/CSysMatrix.hpp
```diff
@@ -1236,6 +1259,16 @@ class CSysMatrix {
...
+  /*!
+   * \brief Apply the forward pass of the LU-SGS preconditioner
+   */
+  void ComputeLU_SGSPreconditionerForward(const CSysVector<ScalarType>& vec, CSysVector<ScalarType>& prod) const;
+
+  /*!
+   * \brief Apply the backward pass of the LU-SGS preconditioner
+   */
+  void ComputeLU_SGSPreconditionerBackward(CSysVector<ScalarType>& prod) const;
+
```
pcarruscag: Move to the private part of the class members

Common/src/linear_algebra/CSysMatrixGPU.cu
```diff
@@ -664,36 +707,249 @@ void CSysMatrix<ScalarType>::ComputeILUPreconditionerGPU(const CSysVector<Scalar
...
+  const auto sharedForward = (threads + nVar) * sizeof(ScalarType);
+
+  if (aux_stream == nullptr) gpuErrChk(cudaStreamCreate(&aux_stream));
+
+  /*--- First part of the symmetric iteration: (D+L).x* = b ---*/
+  if (precond_fwd_graph_exec == nullptr || precond_fwd_graph_vec != d_vec || precond_fwd_graph_prod != d_prod) {
+    if (precond_fwd_graph_exec != nullptr) {
+      gpuErrChk(cudaGraphExecDestroy(precond_fwd_graph_exec));
+      precond_fwd_graph_exec = nullptr;
+    }
```
pcarruscag, line 870: Is it worth using the graph when the vectors are always changing? Does that only happen with FGMRES?

pcarruscag:
Looks good, yes MPI is expected not to work, the communications are only using the CPU values, we would have to download and reupload vectors, which would be inefficient, so the best bet for now is OpenMP

Common/include/linear_algebra/CSysMatrix.hpp
ddg93: updated, thanks.

Common/src/linear_algebra/CSysMatrixGPU.cu
ddg93: I've investigated by examining the pointers at runtime. Indeed with the FGMRES I observe a long rotation among the vectors, therefore capturing the graphs with a single slot is not so efficient. If removed, the solver runs more smoothly on a low-end GPU. I tried another solver with the LU-SGS graph, the BCGSTAB, for which the vector change between two addresses. This suggests the implementation of a cache which gets populated with the graphs as the solver progresses. It would be quite a substantial change, whereas removing the graphs would simplify the logic greatly.

pcarruscag: I see, let's keep them for now, I can do some profiling on my machine as well.
The alternative is to use some buffer vectors into which the input/output are copied each time.
One graph per GMRES iteration seems like a lot.

pcarruscag (approved):

ddg93:
Ok, glad to hear that. I brought it up to date. Maybe CUDA_AWARE MPI could be an interesting option to make GPU comms more efficient.

pcarruscag:
yes but for that the entire communication logic would need to be compatible with GPUs

